PR #406, opened 2016-07-24
[#8663] Fix last test on AppVeyor

hawkowl:
https://twistedmatrix.com/trac/ticket/8663

### DIFF
--- a/twisted/names/test/test_tap.py
+++ b/twisted/names/test/test_tap.py
@@ -105,7 +105,7 @@ class OptionsTests(SynchronousTestCase):
         # lying about that we can't easily clean up without reaching into the
         # reactor and cancelling them. We only cancel the cleanup functions, as
         # there should be no others (and it leaving a callLater lying about
-        # should rightly cause the test to fail.
+        # should rightly cause the test to fail).
         if platform.getType() != 'posix':
 
             # We want the delayed calls on the reactor, which should be all of
--- a/twisted/names/test/test_tap.py
+++ b/twisted/names/test/test_tap.py
@@ -5,15 +5,20 @@
 Tests for L{twisted.names.tap}.
 """
 
-from twisted.trial.unittest import TestCase
-from twisted.python.usage import UsageError
-from twisted.names.tap import Options, _buildResolvers
-from twisted.names.dns import PORT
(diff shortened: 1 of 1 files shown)


twisted/names/test/test_tap.py
```diff
@@ -96,4 +101,21 @@ def test_recursiveConfiguration(self):
...
+            # ours from the threaded resolver cleanup
+            from twisted.internet import reactor
+            for x in reactor._newTimedCalls:
+                if _PY3:
+                    self.assertEqual(x.func.__func__,
+                                     ThreadedResolver._cleanup)
+                else:
+                    self.assertEqual(x.func.__func__,
+                                     ThreadedResolver._cleanup.__func__)
+                x.cancel()
```
adiroiban, line 121: it is not clear why the _newTimedCalls are cancelled ... I guess that they are cancelled because we can not clean them up or we can not wait for them to be done before checking the reactor state ... but it is not  clear

hawkowl, line 121: We explicitly cancel them so that Trial doesn't fail the test saying that there is an unclean reactor.

adiroiban:
Please consider adding OpenSSH-Win64 to the cached folders and only install it if missing + remove the appveyor.yml config which instructs appveyor to ignore the tests results

rodrigc:
Looks good, please merge.
appveyor.yml changes can be done in a separate ticket.

twisted/names/test/test_tap.py
```diff
@@ -96,4 +101,21 @@ def test_recursiveConfiguration(self):
                 if isinstance(recurser, Resolver):
                     recurser._parseCall.cancel()
 
+        # On Windows, we need to use a threaded resolver, which leaves trash
+        # lying about that we can't easily clean up. This isn't a good bit of
+        # code. Sorry, abstraction gods.
```
adiroiban: Maybe, instead of `This isn't a good bit of  code. Sorry, abstraction gods.` we can have:

Deferreds scheduled using `callLater` are manually stopped to have a clean reactor.

---

I think that it would be good to have a follow up tech-debt ticket in case someone has time to look into this issue and in case this will hide a nasty bug.

PR #409, opened 2016-07-25
Fix documentation for twisted.test.test_jelly

acabhishek942:
See https://twistedmatrix.com/trac/ticket/8671

### DIFF
--- a/twisted/test/test_jelly.py
+++ b/twisted/test/test_jelly.py
@@ -169,6 +169,10 @@ class JellyTests(unittest.TestCase):
 
 
     def test_newStyle(self):
+        """
+        Test that a new style class can be jellied and unjellied with its
+        objects and attribute values preserved.
+        """
         n = D()
         n.x = 1
         n2 = D()
--- a/twisted/test/test_jelly.py
+++ b/twisted/test/test_jelly.py
@@ -118,7 +118,7 @@ class JellyTests(unittest.TestCase):
     Testcases for L{jelly} module serialization.
 
     @cvar decimalData: serialized version of decimal data, to be used in tests.
-    @type decimalData: C{list}
+    @type decimalData: L{list}
     """
 
     def _testSecurity(self, inputList, atom):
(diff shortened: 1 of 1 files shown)


rodrigc:
Since you are fixing up the docstring in this file, please fix more stuff.

In this document, https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

it mentions that basic Python data types like int, str, list, tuple, frozenset
should be encapsulated as L{int}, L{str}, L{list}, L{tuple}, L{frozenset}.

Since you are cleaning, you might as well try to clean up whitespace as well.

Before a new method, there should be 2 newlines.
Before a new class, there should be 3 newlines.

If you get the twistedchecker tool by doing:

```
pip install twistedchecker
```

and then run twistedchecker on this file, then you will see the
whitespace issues that can be cleaned up.

twisted/test/test_jelly.py
```diff
@@ -224,13 +232,18 @@ def test_typeBuiltin(self):
...
+        """
+        Jellying L{datetime.timedelta} instances and then unjellying the result
+        should produce objects which represent the values of the original
+        inputs.
+        """
         dtn = datetime.datetime.now()
         dtd = datetime.datetime.now() - dtn
-        input = [dtn, dtd]
-        c = jelly.jelly(input)
+        inputList = [dtn, dtd]
```
adiroiban, line 242: this feels like Hungarian notation... https://en.wikipedia.org/wiki/Hungarian_notation

don't use it :)

what is wrong with `input` ?

acabhishek942, line 242: I used inputList to keep it up with the variable naming style of other tests. For example : https://github.com/twisted/twisted/blob/trunk/twisted/test/test_jelly.py#L242, https://github.com/twisted/twisted/blob/trunk/twisted/test/test_jelly.py#L282,
https://github.com/twisted/twisted/blob/trunk/twisted/test/test_jelly.py#L304.
Besides input is also a keyword in Python3 not that it causes any problem.

PR #410, opened 2016-07-25
[#8664] Make appveyor required to pass

hawkowl:
https://twistedmatrix.com/trac/ticket/8664

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,7 +32,7 @@ environment:
   # See: http://www.appveyor.com/docs/installed-software#python
   matrix:
     - PYTHON: C:\Python27-x64
-      TOXENV: py27-coverage-windows,py27-nomodulescov-windows,codecov-publish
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -32,13 +32,8 @@ environment:
   # See: http://www.appveyor.com/docs/installed-software#python
   matrix:
     - PYTHON: C:\Python27-x64
-      TOXENV: py27-coverage-windows,codecov-publish
+      TOXENV: py27-coverage-windows,py27-nomodulescov-windows,codecov-publish
```
adiroiban: Why not named it `py27-nomodules-coverage-windows`  ?

adiroiban: Also, maybe it better to have it as a separate job (builder) and not run it together with `py27-coverage-windows`

when we run 2 different tests set in the same job, we end up with logs which are hard to read... and this is one of the reason why we moved away from tox-travis

PR #411, opened 2016-07-25
8675 einarfd ssh filetransfer

einarfd:
This merge request ports conch.ssh.filetransfer to Python 3 and is a fix for https://twistedmatrix.com/trac/ticket/8675 .

### DIFF
--- a/twisted/protocols/loopback.py
+++ b/twisted/protocols/loopback.py
@@ -271,7 +271,7 @@ def _loopbackAsyncContinue(ignored, server, serverToClient, client,
 
 @implementer(interfaces.ITransport, interfaces.IConsumer)
 class LoopbackRelay:
-    buffer = ''
+    buffer = b''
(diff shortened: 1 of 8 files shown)


twisted/protocols/loopback.py
```diff
@@ -286,7 +286,7 @@ def write(self, data):
             self.logFile.write("loopback writing %s\n" % repr(data))
 
     def writeSequence(self, iovec):
-        self.write(b"".join(iovec))
+        self.write("".join(iovec))
```
rodrigc: Instead of taking this change out, can you leave it in,
and extend  twisted/conch/test/test_recvline.py
to additionally call Loopback.writeSequence() ?

That way this line will have test coverage.

Your original change is good, and I would hate to lose this change.

rodrigc:
Merged in https://github.com/twisted/twisted/pull/414

PR #412, opened 2016-07-25
[#8677] Run select and iocp reactor jobs on Appveyor.

adiroiban:
See https://twistedmatrix.com/trac/ticket/8677


admin/tox-build-extensions.py
```diff
@@ -0,0 +1,16 @@
+"""
```
rodrigc: This script is not needed at all.  
By default, tox calls **setup.py bdist_wheel** which builds the extensions.

appveyor.yml
```diff
@@ -26,13 +26,25 @@ cache:
 environment:
   global:
     TRIAL_REPORTER: text
+    TWISTED_REACTOR: default
```
rodrigc: Not needed.  If **TWISTED_REACTOR** is not in the environment, tox.ini uses **default**

appveyor.yml
```diff
@@ -41,7 +53,8 @@ install:
   # how to run such command on Windows :)
   # - ps: "ls \"C:/\""
   # - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
-  # - ps: echo $env:PYTHON $env:TOXENV
+  # See the environment variables used for this build.
+  - ps: echo $env:PYTHON $env:TOXENV $env:TWISTED_REACTOR
```
rodrigc: Not needed.  The PYTHON version is printed out elsewhere, and we can
already see which reactor is being used in the --reactor argument to trial

rodrigc:
This was fixed in https://github.com/twisted/twisted/pull/485

PR #415, opened 2016-07-26
Use bytes instead of string in twisted.internet._sslverify

acabhishek942:
See https://twistedmatrix.com/trac/ticket/8672

### DIFF
--- a/twisted/internet/_sslverify.py
+++ b/twisted/internet/_sslverify.py
@@ -452,7 +452,7 @@ class CertBase:
         @type interface: L{zope.interface.interfaces.IInterface}
 
         @return: an L{IOpenSSLTrustRoot} provider or L{NotImplemented}
-        @rtype: C{interface} or L{NotImplemented}
+        @rtype: L{IOpenSSLTrustRoot} or L{NotImplemented}
         """
         if interface is IOpenSSLTrustRoot:
             return OpenSSLCertificateAuthorities([self.original])
@@ -532,11 +532,12 @@ class Certificate(CertBase):
         """
         Get the certificate for the remote end of the given transport.
 
-        @type: L{ISystemHandle}
+        @param transport: an L{ISystemHandle} provider
+
         @rtype: C{Class}
 
         @raise: L{CertificateError}, if the given transport does not have a peer
-        certificate.
+            certificate.
         """
         return _handleattrhelper(Class, transport, 'peer')
     peerFromTransport = classmethod(peerFromTransport)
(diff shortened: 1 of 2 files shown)


rodrigc:
This is good, but we need a few more changes.
(1)  In all the places that you changed, use .encode("ascii") and not .encode("utf-8").  That is what
      the old code was doing based on twisted.python.compat.nativeString(), so we want to use 
      the same encoding for now, so we want to keep the same behavior.

(2)  Some of the docstring is wrong in this file.  It would be good if you could go through
       it and verify that the types returned by the pyOpenSSL library match what is 
       in the docstring for this file.

```
   For example, on this line: https://github.com/twisted/twisted/blob/trunk/twisted/internet/_sslverify.py#L588 it says
    that it returns C{str}.  But if I look at this documentation:
     https://pyopenssl.readthedocs.io/en/stable/api/crypto.html#x509-objects
     the OpenSSL.crypto.X509.digest() method returns L{bytes}.
```

It's important to get this stuff right, especially for Python 3.

Thanks for working on this.

twisted/internet/_sslverify.py
```diff
@@ -1645,11 +1648,11 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
             name = "%s-%d" % (reflect.qual(self.__class__), _sessionCounter())
             sessionName = md5(networkString(name)).hexdigest()
 
-            ctx.set_session_id(sessionName)
+            ctx.set_session_id(sessionName.encode('ascii'))
```
markrwilliams, line 1651: On Python 2 this will allow users to pass in `unicode` objects that can be encoded with the `ascii` codec.  But the docstring for this method says that it wants a `list` of `bytes`.  I don't think the docstring should be changed; instead, I think you should add a helper method that ensures a given object is of type `bytes`, and optionally that ensures it contains only ASCII.  Here's what that might look like

``` python
def isBytes(self, obj):
    if not isinstance(obj, bytes):
        raise TypeError("bytes required, not {!r}".format(obj))
    obj.decode('ascii')
    return obj
```

And in use:

``` python
ctx.set_session_id(self.isBytes(sessionName))
```

I know this will be the second time you've been told to change your code :(  But implicit encoding concerns me when it's used in a security-sensitive API.

acabhishek942, line 1651: As far as I can see this method doesn't have any docstring. Also sessionName is not a passed argument rather calculated using call to hexdigest(), which returns a native `str` object on both python2 and 3. https://docs.python.org/3/library/hashlib.html, https://docs.python.org/2/library/hashlib.html

rodrigc, line 1651: @acabhishek942 your change is fine.  In the hashlib library, the hexdigest() method returns a **str**:
https://github.com/python/cpython/blob/master/Modules/clinic/md5module.c.h#L41

rodrigc:
Merged in https://github.com/twisted/twisted/pull/425

PR #416, opened 2016-07-26
8683 ssh channel name bytestring

einarfd:
This is a fix for bug https://twistedmatrix.com/trac/ticket/8683, and changes the name field in SSHChannel to a bytestring. 

I am a bit conflicted about the changes to **str** and logPrefix.  I am not sure they are the best way of solving the failing unittest when changing the name field. It might be better to either change the unittest to expect differences between Python 3 and 2, or modify the result from them to be common between the two variants but different from what it is today.

### DIFF
--- a/twisted/conch/ssh/channel.py
+++ b/twisted/conch/ssh/channel.py
@@ -14,7 +14,7 @@ from __future__ import division, absolute_import
 from zope.interface import implementer
 
 from twisted.python import log
-from twisted.python.compat import nativeString
+from twisted.python.compat import nativeString, networkString
 from twisted.internet import interfaces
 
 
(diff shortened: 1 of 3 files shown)


twisted/conch/ssh/channel.py
```diff
@@ -77,13 +77,19 @@ def __init__(self, localWindow = 0, localMaxPacket = 0,
 
 
     def __str__(self):
```
rodrigc, line 79: Can you leave the **str**(self) method the way it is,
and introduce a new method called **bytes**(self), and explicitly call that method
instead of calling str()?

I have some patches to conch which do similar things.

twisted/conch/test/test_channel.py
```diff
@@ -151,8 +151,22 @@ def test_str(self):
...
+                str(self.channel), '<SSHChannel channel (lw 131072 rw 0)>')
+
+    def test_bytes(self):
+        """
+        Test that bytes(SSHChannel) works, gives the channel name and
+        local and remote windows at a glance..
+
+        """
+        self.assertEqual(
+            bytes(self.channel), b'<SSHChannel channel (lw 131072 rw 0)>')
```
rodrigc: bytes(self.channel) behaves differently on Python 2 and Python 3.
On Python 3, bytes(self.channel) will call the self.channel.**bytes**() method if it exists.
On Python 2, it does not.

```
class A:

    def __bytes__(self):
        print("I called __bytes__()")
        return b"Hi"


a = A()

b = bytes(a)
```

Until Twisted drops support for Python 2, I advise to explictly call **bytes**() to have consistent behavior.

rodrigc:
@einarfd I made a slight change to your patch so there is no need for if _PY3, and the testcases
are the same on Python2 and Python3.  I merged it here: https://github.com/twisted/twisted/pull/435

Thanks for your hard work, and thanks for your contributions.

PR #418, opened 2016-07-26
Don't explode by sending headers on a reset stream.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8685


twisted/web/topfiles/8685.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web._http2.H2Connection now swallows header writes on reset streams, rather than erroring out.
```
adiroiban: hm... we are referencing a private API. Can this be redone so that we only talk about public API in the release notes?

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
         return a._streamCleanupCallbacks[1].addCallback(validate)
 
 
+    def test_resetAfterBody(self):
+        """
+        A client that immediately resets after sending the body causes Twisted
+        to send no response.
+        """
+        f = FrameFactory()
+        b = StringTransport()
```
adiroiban: can we have it names `transport` instead of `b` ?

twisted/web/topfiles/8685.bugfix
Lukasa: Done.

twisted/web/test/test_http2.py
Lukasa: Done.

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+        to send no response.
+        """
+        f = FrameFactory()
+        transport = StringTransport()
+        a = H2Connection()
+        a.requestFactory = DummyHTTPHandler
+
+        requestBytes = f.preamble()
+        requestBytes += buildRequestBytes(self.getRequestHeaders, [], f)
+        requestBytes += hyperframe.frame.RstStreamFrame(
```
adiroiban: I assume that hyperframe.frame.RstStreamFrame is building a reset message, associate it with a stream and pack it as a frame :) 

I was expecting to have the following DSL  `message = h2.makeResetMessage(stream=some_id)`  and then I can do `a.dataReceived(hyperframe.packForWire(message))`

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+        """
+        A client that immediately resets after sending the body causes Twisted
+        to send no response.
+        """
+        f = FrameFactory()
+        transport = StringTransport()
+        a = H2Connection()
+        a.requestFactory = DummyHTTPHandler
+
+        requestBytes = f.preamble()
```
adiroiban: the preamble name is very generic. what exact is happening here and why do we need it? :)

twisted/web/test/test_http2.py
Lukasa: That API is a different spelling for what is happening here.

Your `h2.makeResetMessage(stream=some_id)` is `hyperframe.frame.RstStreamFrame(stream_id=some_id)`. Your `hyperframe.packForWire(message)` is `frame.serialize()`.

twisted/web/test/test_http2.py
Lukasa: You need it because all HTTP/2 connections begin with the clients sending the client preamble as defined in [RFC 7540](https://tools.ietf.org/html/rfc7540#section-3.5). This is a fixed byte stream (`b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"`) that is required to begin speaking HTTP/2.

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
         return a._streamCleanupCallbacks[1].addCallback(validate)
 
 
+    def test_resetAfterBody(self):
+        """
+        A client that immediately resets after sending the body causes Twisted
+        to send no response.
+        """
+        f = FrameFactory()
```
adiroiban: is the frame factory important for this test?
can we just go with the  default one?

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+        transport = StringTransport()
+        a = H2Connection()
+        a.requestFactory = DummyHTTPHandler
+
+        requestBytes = f.preamble()
+        requestBytes += buildRequestBytes(self.getRequestHeaders, [], f)
```
adiroiban: I now see the DSL :)  ...

This is easier to read

```
buildRequestBytes(header=self.getRequestHeaders, data=[])
```

but maybe 

```
packFramesForHeadersAndPayload(headers=self.getRequestHeaders, payload=[])
or
packFramesForHeadersAndData(headers=self.getRequestHeaders, data=[])
```

twisted/web/test/test_http2.py
Lukasa: The frame factory is a state container which includes, amongst other things, header compression state. This means that we can only use one in a given test.

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+            stream_id=1
+        ).serialize()
+        a.makeConnection(transport)
+        a.dataReceived(requestBytes)
+
+        buffer = FrameBuffer()
+        buffer.receiveData(transport.value())
+        frames = list(buffer)
+
+        self.assertEqual(len(frames), 1)
```
adiroiban, line 1426: the docstring says it send no response, but here I can see that one frame is send.
Why is this frame not a response ? :)

Lukasa, line 1426: It's a SETTINGS frame. The server has to send one at the start of the connection: RFC 7540 also says this.

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+        self.assertNotIn(1, a._streamCleanupCallbacks)
```
adiroiban, line 1427: do we need to use this private member in the test?

maybe we can wrap it as a helper assertion... something like
`self.assertConnectionDoesNotHaveStream(a, 1)`

twisted/web/test/test_http2.py
```diff
@@ -1403,6 +1403,32 @@ def validate(streamID):
...
+        requestBytes += buildRequestBytes(self.getRequestHeaders, [], f)
+        requestBytes += hyperframe.frame.RstStreamFrame(
+            stream_id=1
+        ).serialize()
+        a.makeConnection(transport)
+        a.dataReceived(requestBytes)
+
+        buffer = FrameBuffer()
+        buffer.receiveData(transport.value())
+        frames = list(buffer)
```
adiroiban: maybe this code can also be moved into a helper like

```
frames = get_frames_from_wire(transport.value())
```

twisted/web/test/test_http2.py
adiroiban: My problem with this DSL is that the RFC does not contain the 'preamble' name.
So if I got ot the RFC and the search for preamble I got nothing :)

From what I can see, in the RFC this is called `the connection preface` and `f.preamble()` looks like `f.clientConnectionPreface()`

For a DSL I am expecting to see the same names as in the RFC, so that someone can search the RFC and see what is going on there

twisted/web/test/test_http2.py
adiroiban, line 1426: Reading the RFC this looks like the `server connection preface`,

```
The server connection preface consists of a potentially empty
SETTINGS frame (Section 6.5) that MUST be the first frame the server
sends in the HTTP/2 connection.
```

So to have a DSL, maybe this can be implemented as `self.assertServerConnectionPreface(frames[1])`

twisted/web/test/test_http2.py
adiroiban: Understood.

minor comment: f is a very generic name ... it can be f for frame, frameFactory httpFactory :)

PR #420, opened 2016-07-26
8690 einarfd conch ssh

einarfd:
This is a fix for https://twistedmatrix.com/trac/ticket/8690 .  It ports what is left of twisted.conch.ssh to Python 3 and get the last tests for that module up and running.

### DIFF
--- a/twisted/conch/test/loopback.py
+++ b/twisted/conch/test/loopback.py
@@ -1,0 +1,28 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+"""
+Loopback helper used in test_ssh and test_recvline
+"""
+
+from __future__ import division, absolute_import
+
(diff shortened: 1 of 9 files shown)


twisted/conch/test/loopback.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Twisted Matrix Laboratories.
```
adiroiban, line 1: can we have it twisted.test_proto_helpers ? it looks like a non conch specific helper

adiroiban:
Changes look very good :)

Please consider my comment about moving the helper to proto_helpers.

Moving to proto_helper will put it into the an almost public API zone so it will help to move it with docstring + details about how this helper could be used.

If you are not in the mood for cleanup up this code that is fine and we can leave it as private code in the conch only area.

Thanks!

PR #422, opened 2016-07-26
8370 the0id ssh ec

the0id:
Trying again to create a pull request.


the0id:
It looks like I'm having the same problem.

Appveyor immediately errors with:  AppVeyor was unable to build non-mergeable pull request and travis-ci is waiting for status to be reported.
The only two things I can click are Hide all checks and the appveyor icon to go to the Appveyor homepage.

And when I clicked Comment Github crashed.  I knew my code was bad, but I didn't think it was that bad.

glyph:
@the0id: "This branch has conflicts that must be resolved".  You have to get it up-to-date with trunk.  Do `git merge origin/trunk` locally, fix conflicts, commit and push.

glyph:
Actually, this branch _also_ needs to have some commits pruned out of it.  Normally I'm a favor of preserving full history, but in this case, we have a 16kb file in history which just shouldn't exist at all. You also need to set your author metadata correctly - you can tell it's still wrong because your avatar icon is not showing up next to your commits.

the0id:
Yeah, I think it may be best to start this over from scratch.  I'll create a new branch from the most recent trunk.

the0id:
Though before I go and do something stupid, should I go ahead and delete the fork I created?  It still has all of the old history.

It seems like it should delete it (after backing it up locally of course), and create a new fork. Correct?

glyph:
Creating a new fork won't really do anything. You should possibly create a new branch, though.

PR #423, opened 2016-07-26
Adding ECDH and the requested changes.

the0id:
Let's hope 3rd time's the charm.

Additional notes for this request:
It seems that some of the upstream commits created a few more errors in my code that I needed to address.  Most of them involved putting b in front of strings, but I did need to make a possibly significant change on line 1282 and 1625 of transport.py.

I changed
`sharedSecret = MP(int(self.ecPriv.exchange(ec.ECDH(), self.theirECPub).encode('hex')), 16))`
to
`sharedSecret = MP(int(binascii.hexlify(self.ecPriv.exchange(ec.ECDH(), self.theirECPub)), 16))`
to work with python 3.

I still feel like this transformation could be done better, but it's the best I could come up with.

Also as I mentioned before, but will reiterate this code does not pass all trials.  It still fails on the twisted.conch.test.test_endpoints.NewConnectionTests.test_mismatchedHostKey check.

While making changes to the code I found that it appears to have to do with the addition of  `or _kex.isEllipticCurve(kexAlgorithm)]` to factory.py.  Why I'm not exactly sure, and I'm not sure how to fix it, but hasn't caused any errors when I use Twisted as a client or a server.

I can add a ticket for this to be fixed.

### DIFF
--- a/twisted/conch/ssh/_kex.py
+++ b/twisted/conch/ssh/_kex.py
@@ -6,7 +6,9 @@
 SSH key exchange handling.
 """
 
-from hashlib import sha1, sha256
+from __future__ import absolute_import, division
+
+from hashlib import sha1, sha256, sha384, sha512
 
 from zope.interface import Attribute, implementer, Interface
 
@@ -46,6 +48,12 @@ class _IFixedGroupKexAlgorithm(_IKexAlgorithm):
         "Python generator functions.)")
 
 
+class _IEllipticCurveExchangeKexAlgorithm(_IKexAlgorithm):
+    """
+    An L{_IEllipticCurveExchangeKexAlgorithm} describes a key exchange algorithm
+    that uses an elliptic curve exchange between the client and server.
+    """
+
 
 class _IGroupExchangeKexAlgorithm(_IKexAlgorithm):
     """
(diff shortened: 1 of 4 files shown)


the0id:
_/facepalm_

I did a merge with trunk locally and fixed the conflict it found, and it's saying my branch is up to date.  Though to do so I had to switch to my local trunk branch, merge with upstream/trunk, then merge this branch with my local trunk.  Maybe this is where I went wrong.

My guess/fear is that to complete the merge I had to commit the many,many files that were changed up stream to this branch.  I didn't want to push those changes because it seemed like a bad idea.

Currently when I do git merge origin/trunk or upstream/trunk or just trunk, it says I'm up to date, though git status says I'm ahead of origin/8370-the0id-sshEC-redux by 285 commits.

rodrigc:
@the0id I don't know if you have the cycles to look into this, but if you could help, that would be
great.

While working with @einarfd on porting conch to Python 3, I found the following.

I ran:

```
 python   -bb  -Wall -m unittest twisted.conch.test.test_default
```

I saw a lot of warnings like this:

```
/Users/crodrigues/venv/lib/python3.6/site-packages/pyasn1/codec/ber/decoder.py:95: BytesWarning: Comparison between bytes and string
```

I could figure out where in the Twisted code was causing these warnings, so I put an assert on line 94 of decoder.py.
I found the stack came to this line: https://github.com/twisted/twisted/blob/trunk/twisted/conch/ssh/keys.py#L321

It looks like for pyasn1.codec.ber.decoder(), you aren't supposed to pass it
a bytes directly, but are supposed to wrap the bytes in something else such as a univ.OctetString

http://pyasn1.sourceforge.net/codecs.html#2.2

I am not familiar with the pyasn1 API.  Do you have any ideas how to fix this?

the0id:
Hi, sorry I didn't get back to you sooner, I don't really check this
address often. :)

The short answer is I'm not really sure what the problem is.

My guess would be that the pyasn1 module was written for python2 and has
the same issues most python2 strings have being run under python3, and that
the 'fix' would be finding a python3 version pf pyasn1. I have worked very
little with the pyasn1 API, and I did not enjoy it.  I had similar troubles
you're having in trying to figure out what kind of data it wants (and
gives) and ended up just giving up because I couldn't figure it out.

I'm sorry I can't be of more help, but as I said above, I wouldn't worry
about it unless it's causing you problems, and that it'll likely be sorted
out with a python3 version pf pyasn1.

-Michael

the0id:
If you do figure it out, please let me know.  I'd like to understand the
module more as I'll likely be using it in the future.

PR #429, opened 2016-07-27
Handle WINDOW_UPDATE frames when we have no data to send.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8695


twisted/web/_http2.py
```diff
@@ -625,7 +625,8 @@ def _handleWindowUpdate(self, event):
                 # unnecessary but benign. We'll ignore it.
                 return
 
-            self.priority.unblock(streamID)
+            if self._outboundStreamQueues.get(streamID):
```
adiroiban, line 632: minor comment

rather than this `if: do` , maybe we can go with `try: do: catch:`

```
try:
   self.priority.unblock(streamID)
except IndexError:
  pass
```

I don't know if catching this generic IndexError is a good idea ... maybe prority.unblock() can be changed to raise a more specific error 

again, just a minor comment

Lukasa, line 632: We cannot. `priority.unblock()` doesn't raise an `IndexError` at this point in the code, because the stream _is_ in the priority tree. Instead it raises an `IndexError` much later in the prioritised data sending loop, where we attempt to send some data we don't actually have.

twisted/web/test/test_http2.py
```diff
@@ -1971,6 +1972,30 @@ def validate(streamID):
...
+        f = FrameFactory()
+        transport = StringTransport()
+        conn = H2Connection()
+        conn.requestFactory = DummyHTTPHandler
+
+        # Send the request.
+        frames = buildRequestFrames(
+            self.postRequestHeaders, self.postRequestData, f
+        )
+        frames.insert(1, f.buildWindowUpdateFrame(streamID=1, increment=5))
```
adiroiban: maybe is better to have

```
frames = []
frames.append(buildHeadersFrame(self.postRequestHeaders))
frames.append(buildWindowUpdateFrame(streamID=1, increment=5))
frames.append(buildDataFrame(self.postRequestData))

conn.dataReceived(serializeFrames(frames))
```

I feel that rather working around the limitation of our own helper methods, we should create better helper methods :)

twisted/web/test/test_http2.py
```diff
@@ -1971,6 +1972,30 @@ def validate(streamID):
...
+        firstBytes = f.preamble()
+        firstBytes += b''.join(f.serialize() for f in frames[:2])
+
+        conn.makeConnection(transport)
+        conn.dataReceived(firstBytes)
+
+        self.assertRaises(priority.DeadlockError, next, conn.priority)
```
adiroiban: I don't undertand how this relates to `nothing exciting happens` from the docstring .

I think that the docstring can be improved to be more specific. Since it is not `nothing happens` I assume that something will happen...just that it is not exciting :)

and with a better docstring, maybe this assertion will make sense

twisted/web/test/test_http2.py
```diff
@@ -1971,6 +1972,30 @@ def validate(streamID):
...
+        conn = H2Connection()
+        conn.requestFactory = DummyHTTPHandler
+
+        # Send the request.
+        frames = buildRequestFrames(
+            self.postRequestHeaders, self.postRequestData, f
+        )
+        frames.insert(1, f.buildWindowUpdateFrame(streamID=1, increment=5))
+        firstBytes = f.preamble()
+        firstBytes += b''.join(f.serialize() for f in frames[:2])
```
adiroiban: why are we not serializing all the frames?

twisted/web/test/test_http2.py
```diff
@@ -1971,6 +1972,30 @@ def validate(streamID):
...
+
+        conn.makeConnection(transport)
+        conn.dataReceived(firstBytes)
```
adiroiban: why is this named `firstBytes` .. are we going to have `secondBytes` ?

I assume that this is related to `no data is currently waiting` part of the docstring, but I can't find the connection  and I can't see how the stream with no waiting data is created.

twisted/web/_http2.py
adiroiban, line 632: Thanks! Maybe it is worth leaving a comment for future generations which might want to refactor this based on EAFP

twisted/web/test/test_http2.py
Lukasa: So this high-level idea isn't bad, but it turns out to be a bit tricky in practice.

Firstly, at a low-level we already have those methods. hyperframe basically exposes those using the frame constructors.

The bigger problem is that there is connection state that needs to be managed. For example, HTTP/2 header compression is stateful, which means that we need somewhere to store a H2 header compressor that lasts for the duration of the test, and then pass it to all these helper methods (or, more likely, `serializeFrames`). The next problem is that both HEADERS and DATA frames also have stream IDs, and flags, all of which need to conditionally be populated depending on exactly how the stream state is being controlled.

We can certainly do this, but it doesn't vastly improve the state of affairs. Regardless, if we want to do this we should do it as part of a separate ticket, because all the tests in this file would want to be changed.

adiroiban: I guess that we can introduce the new DSL helpers in this ticket and if they work out well we can roll them to other tests.

For me is important to have tests which are simple and easy to read as we don't have any automated code to test the test code. The only way to check that a test is valid is by reading the code.

Here we create 2 frames, then we insert a new frame between the newly created 2 frames and then we send only the first 2 frames.

So the code would be

```
frames = []
frames.append(f.buildHeadersFrame(streamID=1, self.postRequestHeaders))
frames.append(f.buildWindowUpdateFrame(streamID=1, increment=5))

conn.dataReceived(serializeFrames(frames))
```

adiroiban: And with the above code, I think that the current big  comment block can just go away :)

twisted/web/test/test_http2.py
```diff
@@ -1971,6 +1972,36 @@ def validate(streamID):
...
+        )
+        frames.insert(1, f.buildWindowUpdateFrame(streamID=1, increment=5))
+        data = f.preamble()
+        data += b''.join(f.serialize() for f in frames[:2])
+
+        conn.makeConnection(transport)
+        conn.dataReceived(data)
+
+        # All the streams should still be blocked.
+        self.assertRaises(priority.DeadlockError, next, conn.priority)
```
adiroiban: with a helper assertion like `self.assertConnectionWithAllStreamBlocked(conn)` we would no longer need the comment :)

rodrigc:
@Lukasa as you are taking a pass over the HTTP code, can you see if you can fix
these errors generated during the api documentation build:

```
twisted.web._http2.H2Connection:65 invalid ref to 'priority.PriorityTree' resolved as 'PriorityTree'
twisted.web._http2.H2Connection.priority:0 invalid ref to 'priority.PriorityTree' resolved as 'PriorityTree'
```

If we can fix that, we will have zero errors during the documentation build.

See:
https://buildbot.twistedmatrix.com/builders/documentation/builds/1202/steps/api-documentation/logs/stdio

Lukasa:
I'm not sure I understand the problem. `priority` is in the namespace, and `priority.PriorityTree` is used in the code itself. The `priority` module provides an intersphinx URL which is correct and responding.

I'm willing to fix the problem, but I honest-to-goodness don't know what it is.

adiroiban:
This is a bug in pydoctor https://github.com/twisted/pydoctor/issues/124

it was introduced since pydoctor automated test coverage is low... but recently we are also running pydoctor on twisted-trunk ... so in the future we should see less regressions of these type

rodrigc:
On this line https://github.com/twisted/twisted/blob/trunk/twisted/python/_release.py#L40
the build process is supposed to fetch the objects.inv file for the priority library and use
that during the API documentation build to resolve all the references.

What I have done to debug this before is the following:

(1)  in a Python 2.7 virtualenv, make sure that pydoctor and sphinx are installed

(2)  Grab this script https://gist.github.com/epc/4118456 , and hack it as per https://gist.github.com/epc/4118456#gistcomment-1218679

(3)  Get the objects.inv file for priority and run:

```
python /tmp/inventory.py  --file objects.inv --terse
```

Output should be:

```
:std:label:`priority:modindex`
:std:label:`priority:genindex`
:std:label:`priority:search`
:py:method:`priority.PriorityTree.remove_stream <priority:priority.PriorityTree.remove_stream>`
:py:method:`priority.PriorityTree.block <priority:priority.PriorityTree.block>`
:py:method:`priority.PriorityTree.insert_stream <priority:priority.PriorityTree.insert_stream>`
:py:method:`priority.PriorityTree.reprioritize <priority:priority.PriorityTree.reprioritize>`
:py:method:`priority.PriorityTree.unblock <priority:priority.PriorityTree.unblock>`
:py:class:`priority.MissingStreamError <priority:priority.MissingStreamError>`
:py:class:`priority.PriorityLoop <priority:priority.PriorityLoop>`
:py:class:`priority.DeadlockError <priority:priority.DeadlockError>`
:py:class:`priority.DuplicateStreamError <priority:priority.DuplicateStreamError>`
:py:class:`priority.PriorityTree <priority:priority.PriorityTree>`
```

After that, I am not sure what is wrong.

PR #431, opened 2016-07-27
8697 einarfd py3 conch client knownhosts

einarfd:
This ports conch.client.knownhosts to Python 3 and is a fix for https://twistedmatrix.com/trac/ticket/8697 .


rodrigc:
Merged in https://github.com/twisted/twisted/pull/433

PR #432, opened 2016-07-27
[#8325] Adding in ECDH for SSH.

the0id:
I think I'm learning how to Git.

It seems that some of the upstream commits created a few more errors in my code that I needed to address. Most of them involved putting b in front of strings, but I did need to make a possibly significant change on line 1282 and 1625 of transport.py.

I changed
sharedSecret = MP(int(self.ecPriv.exchange(ec.ECDH(), self.theirECPub).encode('hex')), 16))
to
sharedSecret = MP(int(binascii.hexlify(self.ecPriv.exchange(ec.ECDH(), self.theirECPub)), 16))
to work with python 3.

I still feel like this transformation could be done better, but it's the best I could come up with.

Also as I mentioned before, but will reiterate this code does not pass all trials. It still fails on the twisted.conch.test.test_endpoints.NewConnectionTests.test_mismatchedHostKey check.

While making changes to the code I found that it appears to have to do with the addition of or _kex.isEllipticCurve(kexAlgorithm)] to factory.py. Why I'm not exactly sure, and I'm not sure how to fix it, but hasn't caused any errors when I use Twisted as a client or a server.

I can add a ticket for this to be fixed.


the0id:
Well there's no conflicts.  That's progress. :)

Not too surprised it failed, as mentioned before it's known that the mismatchedHostKey test fails.  Looking at the doc of the test it seems that it's expecting a certain host key in known_hosts but is getting a different one.  Why this is I don't know, in my testing the clients make use of known_hosts fine, but as the test was running I got an idea of what could be happening, so I'll take a look at it.

The other failure was because I didn't have a top file in my manifest.  I don't know what this means.  Is there another file I need to commit or add in addition to my changes?

the0id:
I guess this is what happens when you forget to verify the fingerprint of the host key.

Working on a patch.

the0id:
It appears I have too many diffs?  Am I reading the condecov report correctly?  What needs to be done about that?

admin/run-python3-tests
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3.3
+#!/usr/bin/env python3
```
glyph: This change is unrelated to this branch; would you mind making it in a separate PR?

twisted/conch/ssh/keys.py
```diff
@@ -4,6 +4,9 @@
 
 """
 Handling of RSA and DSA keys.
+
+@ivar curveTable: a look up table for translating NIST standard curve names to Cryptograpy instances.
```
glyph: This should be `@var` and not `@ivar`, since modules aren't classes, they don't have instances - `@ivar` stands for "instance variable".

The reference documentation for this sort of things is http://epydoc.sourceforge.net/epytext.html

twisted/conch/ssh/keys.py
```diff
@@ -40,7 +44,38 @@
 from twisted.python.deprecate import deprecated, getDeprecationWarningString
 from twisted.python.versions import Version
 
-
+# Curve lookup table
+curveTable = {
+        b'nistp256' : ec.SECP256R1(),
```
glyph: This is too much indentation; it should be 4 spaces.

twisted/conch/ssh/keys.py
```diff
@@ -640,27 +688,33 @@ def __repr__(self):
...
-                for c in iterbytes(m):
-                    o = o + '%02x:' % (ord(c),)
-                if len(m) < 15:
-                    o = o[:-1]
-                lines.append('\t' + o)
-        lines[-1] = lines[-1] + '>'
-        return '\n'.join(lines)
+        if self.type() == 'EC':
+            if self.isPublic():
+                return self._keyObject.public_bytes(serialization.Encoding.PEM, serialization.PublicFormat.SubjectPublicKeyInfo)
```
glyph: This line's too long; it should be wrapped before 80 characters. Sorry that our twistedchecker setup is a bit jacked up at the moment; tooling should be telling you this, not me :)

twisted/conch/ssh/keys.py
```diff
@@ -824,6 +878,12 @@ def type(self):
                 'unknown type of object: %r' % (self._keyObject,))
 
 
+    def getECKeyName(self):
```
glyph: This will always return `None`, since `ecKeyName` is not referenced anywhere else in this pull request.

twisted/conch/ssh/keys.py
```diff
@@ -1121,6 +1191,11 @@ def _toString_AGENTV3(self):
             elif self.type() == 'DSA':
                 values = (data['p'], data['q'], data['g'], data['y'],
                           data['x'])
+            # I assume this is what's needed
+            # But commented out because it hasn't been tested yet.
+            elif self.type() == 'EC':
+                raise UnsupportedAlgorithm("toString() does not support Elliptic Curves yet.")
+                #values = (data['curve'], data['curve'][-8:], data['p'], data['x'])
```
glyph: This commented-out line shouldn't be here.

twisted/conch/ssh/keys.py
```diff
@@ -1121,6 +1191,11 @@ def _toString_AGENTV3(self):
             elif self.type() == 'DSA':
                 values = (data['p'], data['q'], data['g'], data['y'],
                           data['x'])
+            # I assume this is what's needed
+            # But commented out because it hasn't been tested yet.
+            elif self.type() == 'EC':
+                raise UnsupportedAlgorithm("toString() does not support Elliptic Curves yet.")
```
glyph: Whenever you have an unsupported thing like this, you should have a comment that links to a ticket on twistedmatrix.com, so that when people comes along they know what to do about it :)

twisted/conch/ssh/transport.py
```diff
@@ -1498,6 +1593,68 @@ def ssh_KEXINIT(self, packet):
...
+            lambda unused: self.sendDisconnect(
+                DISCONNECT_HOST_KEY_NOT_VERIFIABLE, b'bad host key'))
+        return d
+
+    def _continue_KEX_ECDH_REPLY(self, ignored, hostKey, pktPub, signature):
+        #Save off the host public key.
+        self.theirECHost = hostKey
+
+        #Take the provided public key and transform it into a format for the cryptography module
+        self.theirECPub = ec.EllipticCurvePublicNumbers.from_encoded_point(self.curve, pktPub).public_key(default_backend())
```
glyph: This line's also a bit too long :)

twisted/conch/ssh/transport.py
```diff
@@ -1498,6 +1593,68 @@ def ssh_KEXINIT(self, packet):
...
+        d = self.verifyHostKey(hostKey, fingerprint)
+        d.addCallback(self._continue_KEX_ECDH_REPLY, hostKey, f, signature)
+        d.addErrback(
+            lambda unused: self.sendDisconnect(
+                DISCONNECT_HOST_KEY_NOT_VERIFIABLE, b'bad host key'))
+        return d
+
+    def _continue_KEX_ECDH_REPLY(self, ignored, hostKey, pktPub, signature):
+        #Save off the host public key.
+        self.theirECHost = hostKey
```
glyph: I could have sworn I made this comment before, but:
1. `theirECHost` is a public attribute based on its name; it doesn't need to be exposed to application programmers.
2. It actually doesn't need to be recorded at all in an attribute; it's used only in the local function.
3. Same for `theirECPub`.

Can you get rid of these attributes unless there's a reason to keep them?  And if there's a reason to keep them, add some test coverage that checks for them?

twisted/conch/ssh/transport.py
```diff
@@ -1498,6 +1593,68 @@ def ssh_KEXINIT(self, packet):
...
+        fingerprint = b':'.join([binascii.hexlify(ch) for ch in
+                                 iterbytes(md5(hostKey).digest())])
+        d = self.verifyHostKey(hostKey, fingerprint)
+        d.addCallback(self._continue_KEX_ECDH_REPLY, hostKey, f, signature)
+        d.addErrback(
+            lambda unused: self.sendDisconnect(
+                DISCONNECT_HOST_KEY_NOT_VERIFIABLE, b'bad host key'))
+        return d
+
+    def _continue_KEX_ECDH_REPLY(self, ignored, hostKey, pktPub, signature):
```
glyph: This method is called only once from `_ssh_KEX_ECDH_REPLY`.  Perhaps it should be an inner function so that other maintainers looking at conch don't mistakenly think it's a method they may need to call?

twisted/conch/ssh/transport.py
```diff
@@ -1229,6 +1233,71 @@ def ssh_KEXINIT(self, packet):
...
+        try:
+            shortKex = re.search(b"(nist[kpbt]\d{3})$", self.kexAlg).group(1)
+        except AttributeError:
+            raise UnsupportedAlgorithm(self.kexAlg)
+
+        #Get the curve instance
+        self.curve = keys.curveTable[shortKex]
+        
+        #Generate the private key
+        self.ecPriv = ec.generate_private_key(self.curve, default_backend())
```
glyph: Does `_ecPriv` need to be public?

twisted/conch/ssh/transport.py
```diff
@@ -1229,6 +1233,71 @@ def ssh_KEXINIT(self, packet):
...
+        privHostKey = self.factory.privateKeys[self.keyAlg]
+
+        #Get the base curve info
+        try:
+            shortKex = re.search(b"(nist[kpbt]\d{3})$", self.kexAlg).group(1)
+        except AttributeError:
+            raise UnsupportedAlgorithm(self.kexAlg)
+
+        #Get the curve instance
+        self.curve = keys.curveTable[shortKex]
```
glyph: Does `curve` need to be public?

twisted/conch/ssh/transport.py
```diff
@@ -1229,6 +1233,71 @@ def ssh_KEXINIT(self, packet):
...
+        
+        #Generate the private key
+        self.ecPriv = ec.generate_private_key(self.curve, default_backend())
+
+        #Get the public key
+        self.ecPub = self.ecPriv.public_key()
```
glyph: Does `ecPub` need to be public?

twisted/conch/ssh/transport.py
```diff
@@ -1229,6 +1233,71 @@ def ssh_KEXINIT(self, packet):
...
+        encPub = self.ecPub.public_numbers().encode_point()
+
+        #Take the provided public key and transform it into a format for the cryptography module
+        self.theirECPub = ec.EllipticCurvePublicNumbers.from_encoded_point(self.curve, pktPub).public_key(default_backend())
```
glyph: Does `theirECPub` need to be public?

admin/run-python3-tests
the0id: Yeah that accidentally got pushed, I wasn't sure if it's something you'd want or not.  I'll take it out. :)

twisted/conch/ssh/keys.py
the0id: I agree.  I was going to hold off on writing a ticket for this until the changes were accepted since it didn't make sense to have a ticket for something that's not in the baseline yet.

Once the PR goes through I'll put in a ticket.

twisted/conch/ssh/transport.py
the0id: Yeah it was brought up before, and I mentioned that I was following the convention elsewhere (or so I thought) to keep the public/private keys  exposed, even if they aren't currently being referenced elsewhere.

This applies to the comments below about ecPub and curve.

I don't remember the resolution of that discussion, but I'll make the changes. :)

twisted/conch/ssh/transport.py
the0id: I was following the format that _ssh_KEXDH_REPLY uses by splitting the exchange into _ssh_KEXDH_REPLY and _continue_KEXDH_REPLY.

I assumed that it was set to be a callback to handle the error if they verifyHostKey function failed, but mostly I did it this way just to keep convention.

twisted/conch/ssh/keys.py
the0id: Got it, thanks for the link. :)

the0id:
After looking at the code again, I decided to keep ecPub, ecPriv, and curve public.  They need to be self variables under SSHClientTransport, and it made sense to me to keep them the same way in SSHServerTransport even though they don't really need to be to mirror the behaviors.

But if this isn't desired I can remove self from them.

twisted/conch/ssh/keys.py
glyph: My point is that the PR needs to include the link, which means the ticket needs to be filed first :).  I realize this is a little awkward, but just include some context in the ticket's description, and if the PR doesn't land or the ticket becomes invalid, someone else can close it later.

twisted/conch/ssh/transport.py
glyph: Unfortunately, the extreme age of Twisted often means you'll find local conventions which are sub-optimal and even downright bad.  This convention - using bound private methods instead of inner functions - dates back to a time when Python _did not even have inner functions_; you had to use a bound method just to get "self" curried into the method object so you could share state with the previous step in the invocation chain.

So, in accordance with modern (i.e. post-1998) Python, we should avoid making the situation worse even if the code does look a little mismatched.  If you wanted to do an easy lay-up refactoring PR later, you could remove some of the other `_continue_` methods and make them callbacks later as well.

glyph:
Why do they "need to be self variables"?  If they need to, how can you "remove self from them"?

If we need to expose them to clients, we can expose them in a different PR.  We should generally make things private by default and only expose them as necessary.

the0id:
We may be talking about two different things.  I'm assuming that "self", "instance", and "public" are synonyms when talking about the variables.  That if I have self.foo, foo is a "self", "instance", and a "public" variable.

If this is the case, then ecPriv/ecPub variables in the SSHClientTransport instance, they have to be instance variables because they get generated in ssh_KEXINIT but have to be referenced later in _continue_KEX_ECDH_REPLY.

This isn't the case in SSHServerTransport, but I thought since they're instance variables in SSHServerTransport, and it appears other parts of the normal DH keys are saved as "self", then I should do the same with all of the ecPriv/ecPub variables.

glyph:
> We may be talking about two different things. I'm assuming that "self", "instance", and "public" are synonyms when talking about the variables. That if I have self.foo, foo is a "self", "instance", and a "public" variable.

The formally-correct terminology here is that, if you have `self.foo`
- `foo` is an _attribute_ on `self` ("self variable" is totally comprehensible but not quite the right lingo)
- `foo` is an "instance variable"
- `foo` is a "public attribute"

However, if you have `self._foo`, the first two statements would be true but not the third.  So big parts of this are me saying "it should have an underscore so that 3rd-party applications don't use it".  You can see http://twisted.readthedocs.io/en/twisted-16.3.0/core/development/policy/compatibility-policy.html for more information about why that's important.

> If this is the case, then ecPriv/ecPub variables in the SSHClientTransport instance, they have to be instance variables because they get generated in ssh_KEXINIT but have to be referenced later in _continue_KEX_ECDH_REPLY.

Right, and that's exactly why I'm suggesting `_continue_KEX_ECDH_REPLY` be made into an inner function :).  If these variables live only for the scope of a single `Deferred` transaction and then are unused from then on, keeping them on `self` is confusing for other maintainers debugging this code in the future.  They have to look around and see what else is touching those variables.  Whereas if it's a variable attached only to a closure, they know only that closure can possibly be using it and they can limit the amount of code they read and understand.

> This isn't the case in SSHServerTransport, but I thought since they're instance variables in SSHServerTransport, and it appears other parts of the normal DH keys are saved as "self", then I should do the same with all of the ecPriv/ecPub variables.

Every new public attribute is an additional maintenance burden for future maintainers.  The more functionality we can add without exposing implementation details to users, the more flexibility we have for changing the way our code works in the future.  If there's a reason applications need to be accessing those variables then by all means we can expose them, but:
1. even if that's the case, we can expose it later, in a separate PR, and
2. if it is the case, it should be clearly explained (ideally in a docstring) who would want to use it and why

Hopefully this clears up my concern in this area :)

the0id:
Thanks for explaining all of that, I think that does clear it up some. :)

But while I think I understand what you're saying, I don't understand how to do it.

When you say make _continue_KEX_ECDH_REPLY an in inner function, I assume you mean an inner function of _ssh_KEX_ECDH_REPLY, which makes sense.  But once inside _continue_KEX_ECDH_REPLY, I don't see how curve, ecPub, and ecPriv can be in scope if you remove self from them because they are instantiated in ssh_KEXINIT.

But for these variables in theSSHServerTransport class, I'll remove the self, because I see what you're saying about maintenance burden.

markrwilliams:
@the0id `self` is just another argument a method takes.  That means you can close over it!

``` python
class Something(object):
    def method(self, arg):
        def closure():
            return self, arg
        return closure


closure = Something().method(1)
print(closure())
```

(And the result: https://eval.in/616150)

So if you define `_continue_KEX_ECDH_REPLY` inside `_ssh_KEX_ECDH_REPLY`, it will close over `self`, through which you can access any instance variable you'd like.

Hope that helps!

the0id:
@markrwilliams thanks for the help, but I think the situation here is a little different.
It's more like this:

```
class Something(object):
    def method(self):
        def closure():
            arg = bar
            return self, arg
        return closure 

     def instance(self):
          bar = "Hello World!"

```

And this is why I think bar needs to be self.bar, otherwise closure() can't see bar.

But as always I may still be missing something. :)

the0id:
I went ahead and moved _continue_KEX_ECDH_REPLY into a inner function, but I still had to keep the self variables on the SSHClientTransport side (I removed them from SSHServerTransport).

Coverage is still failing, but I think it's better than the tool reports.

The coverage of the re search for the shortKexes in transport.py should be covered by the test_checkBadCurveName in test_transport.py.

the0id:
I've merged and did some minor updates, but the appveyor test seems to be failing. Looking at the results it seems like appveyor cannot install the required libraries to build the package.

Is this an issue with appveyor or something I did?

rodrigc:
@the0id not your fault.  It looks like python-cryptography just got updated now on pypi https://pypi.python.org/pypi/cryptography/1.5 and the Windows wheels haven't been uploaded to pypi properly.  It should not be trying to build things which need the openssl header files.  The wheel on Windows should be properly precompiled.

the0id:
Great, things are always better when they're not my fault. :)

I'll come back later and try again.

glyph:
We should really be pinning our dependencies in a manner similar to the Mimic project https://github.com/rackerlabs/mimic, testing against a defined exact set of versions in `requirements.txt`.

glyph:
If we can do that and use `requires.io` to stay current on our dependencies, then the PR that fails the builds will be one that changes nothing but upgrading.

glyph:
@adiroiban - is there any way to re-start a build on appveyor?

glyph:
@the0id - https://pypi.org/project/cryptography/#files now shows Windows wheels available, so if you push an empty commit you should be able to pass appveyor now.

adiroiban:
> @adiroiban - is there any way to re-start a build on appveyor?

To allow other people to restart, we can create a shared account with shared password ... there is a ticket for that, but I did not had time to set up the new acccount

---

regarding pinning, I think that we are already defining the test dependencies inside tox.ini ... it just that we need to pin then

AFAIK the tox.ini is only for development usage and not targeted to end users.

but yes, we can also move those dependencies definition outside of tox and into a requirements.txt file

glyph:
`requirements.txt` is easier to parse and manipulate than `tox.ini` so that would be a better place for them.

One of the other things about Mimic's specific organization of dependencies is that it doesn't install by doing `-r requirements.txt`; that would mean `setup.py`'s dependency declarations weren't being tested.  Instead, it pre-populates a directory of wheels using the relevant `.txt` (different environments have different packages and pins) and then uses `PIP_NO_INDEX` to tell `tox` not to talk to the network when it assembles the venv for testing; just to look at the pre-populated directory.

the0id:
I decided to check in and see what the status of the PR was.  Things have been getting busy again so I haven't checked in as much as I'd like.

It seems like code coverage is still an issue, but I still hold it's really better than what's reported.  So I don't know how you'd like to proceed with this, but I probably won't be able to get back to trying to fix the coverage any time soon.

glyph:
> I decided to check in and see what the status of the PR was. Things have been getting busy again so I haven't checked in as much as I'd like.

https://twistedmatrix.com/trac/ticket/8370 is not in review right now, so the ball is in your court.  Or rather, the author's court - if you want to try to prod someone else into responding to the review feedback that is totally legit :).

> It seems like code coverage is still an issue, but I still hold it's really better than what's reported. So I don't know how you'd like to proceed with this, but I probably won't be able to get back to trying to fix the coverage any time soon.

You can see the patch-coverage report [here](https://codecov.io/gh/twisted/twisted/compare/cc51187e5cde697df6ec408c799669a37d3bd130...b21510573fe279d68f5453f9d7e4febb650ec431).  If you load that page and hit <kbd>x</kbd>, it will jump down to the uncovered line.  There are only 3 uncovered lines that I see in that report - if you can fix those 3 lines I suspect that it will go to 100%.

adiroiban:
there are also these PRs by @acabhishek942 

 https://github.com/twisted/twisted/pull/533 and https://github.com/twisted/twisted/pull/539

the0id:
@glyph, oh I didn't know I needed to re-add review to the ticket.  I thought this was still in review and the tag was removed to avoid someone else picking up the ticket.

I made some comments on my ticket and put the tag back in.  This PR is still being built as I write this, but I need to get back so I'm just going to assume my code will be flawless as always. :)

@adiroiban I looked at the PR's you reference, and as I mention in my ticket comments I don't think there's a big difference between those PR's and my PR (though the tests @acabhishek942 added are very welcome), so handle it as you wish.

glyph:
@the0id Sorry about the process confusion!  Thanks for responding to take care of the paperwork here :).  I'm not totally sure which PR is going to "win" here, but @acabhishek942 is mostly just cleaning up and splitting up your code, so you will be listed as an author regardless :).

the0id:
@glyph No problem, part of the learning process.

I've giving one last try at getting full coverage.  From what I understand the two lines that are red should be covered by test_checkBad_KEX_ECDH_INIT, but maybe I'm just checking for UnsupportedAlgorithm being raised, and not AttributeError (though you can't have one without the other).

We'll see if this does it.  After all this time I would like to be able to complete my trials and throw my hands up in the air atop the Philadelphia Art Museum steps.  If not, I'll just tell Mickey to ring the bell.

the0id:
Looks like the first python2.7 test is causing failures for both me and acabhishek942.  I'll check back later and try again.

the0id:
So Mickey said he didn't hear no bell.  Which means I had to give this another shot.

Adding the keySizeZero test fixed the one coverage complaint codecov had, but then something else needed to be covered and I seem to be perpetually at 95% coverage. Which means either I'm not understanding what codecov is saying, or adding one test allows it to find another test that is required.

Regardless of which it is, I think I'm going to have to throw in the towel for now.  I can never really let these things go, and I hate being so close but not being able to finish it, so I'll probably be back eventually, but unless I find some more time (aside from hitting Update branch), this is probably where it's going to be left for awhile.

the0id:
At this point I'm going to do what Rocky should have done and throw in the towel.  Now that acabhishek942's changes have been merged in it's better to use the main trunk instead of trying to maintain a separate fork; which was the point all along. :)

Though I made some comments in my ticket [#8325 ](https://twistedmatrix.com/trac/ticket/8325#comment:23) saying that when I tried to incorporate acabhishek942's changes with my ECDH module I ran into some problems.

These may be problems I introduced while trying to merge, but in case you run into the same problems, I've updated this branch with what worked for me.

the0id:
And boy did that blow up.

I found some of the problems, and I'll take a look at the others later on.  But while I've never put a high priority on getting the tests passed, it's even less now that this fork has become a dead end. :)

the0id:
It looks like most of the issues were related to how the ECDSA key components were returned and the changes to use SHA256 fingerprints.

I rolled back the SHA256 fingerprints back to MD5 because that was easier than trying to fix all of the tests, but I left the code commented out because I believe that's a good direction to go.

The differences between how acabhishek942 and I were returning the ECDSA key components I think are mostly a stylistic preference, but there's value in his method and since it has already been merged in I incorporated that way into this fork.

There did seem to be an issue with how the key blob was being returned when using ECDSA.  It was missing some components, and the encoding seemed incorrect.  I was able to fix it by copying the code in cryptography's publickey.encode_point() method, which makes it a little more complicated, but that seemed to be the only way I could get it to work.

As before, acabhishek942 may find a better way to do what I've done, but this should be working again.

rodrigc:
Moved to https://github.com/twisted/twisted/pull/630

Thanks for all the work on this @the0id

PR #436, opened 2016-07-28
[#7598] Port Perspective Broker to Python 3

glyph:


twisted/python/dist3.py
```diff
@@ -299,6 +299,8 @@
     "twisted.scripts.twistd",
     "twisted.spread.__init__",
     "twisted.spread.banana",
```
hawkowl: I think we need 'flavors' here too...

glyph: yep, added

hawkowl:
Py3.3 doesn't seem to be happy...

glyph:
The only test I see failing on 3.3 is `twisted.trial.test.test_tests.UnhandledDeferredTests.test_doesntBleed` which is completely unrelated as far as I can tell.  It is a garbage collection thing; maybe Travis's infra is experiencing some memory pressure?

hawkowl:
Considering its failing on push _and_ pr builders (which are separate job
runs) I find that unlikely... Did the GC semantics change in 3.4? And do we
care if 3.3 doesn't work? (Apart from pypy 3 I guess)

glyph:
I have no idea. That test should be totally unrelated!

hawkowl:
Are the pb tests leaving a reference around, maybe? Does it happen on 3.4
if you run the tests in the same order as 3.3?

glyph:
I suppose I should run 3.3 locally first.  But if I were going to - how would I run them in the same order?

twisted/cred/checkers.py
```diff
@@ -76,7 +76,7 @@ class InMemoryUsernamePasswordDatabaseDontUse(object):
                             credentials.IUsernameHashedPassword)
 
     def __init__(self, **users):
-        self.users = users
+        self.users = {x.encode('ascii'):y for x, y in users.items()}
```
adiroiban: minor comment:

why not encode with utf-8 ? ... so that we can also use this code to test unicode related usernames / passwords ?

twisted/spread/flavors.py
```diff
@@ -16,33 +16,35 @@
...
 
-# twisted imports
 from twisted.python import log, reflect
-
-# sibling imports
-from .jelly import setUnjellyableForClass, setUnjellyableForClassTree, setUnjellyableFactoryForClass, unjellyableRegistry
-from .jelly import Jellyable, Unjellyable, _newDummyLike
-from .jelly import setInstanceState, getInstanceState
+from twisted.python.compat import _PY3, unicode, comparable, cmp
+from .jelly import (
```
adiroiban: are relative import a thing we want to use in Twisted?

just trying to understand the coding standard

twisted/spread/flavors.py
```diff
@@ -107,6 +109,10 @@ def remoteMessageReceived(self, broker, message, args, kw):
         """
         args = broker.unserialize(args)
         kw = broker.unserialize(kw)
+
+        if not isinstance(message, str):
+            message = message.decode('utf8')
```
adiroiban: this code is not covered ... and I find it hard to understand why we got it here.

I am expected that remote messages are always received as bytes... so they should always be decoded.

Can we have the `message` argument documented to make it clear what types are accepted ?

twisted/spread/jelly.py
```diff
@@ -88,12 +90,15 @@
         # sets module is deprecated in Python 2.6, and gone in
         # Python 3
         _sets = None
+    else:
+        _SetTypes.append(_sets.Set)
```
adiroiban: minor comment: can these code raise an ImportError, why not move it closer to the `import sets` ... I find it easier to read

twisted/spread/pb.py
```diff
@@ -981,14 +1005,16 @@ def proto_error(self, requestID, fail):
...
 
     def proto_decref(self, objectID):
         """(internal) Decrement the reference count of an object.
 
         If the reference count is zero, it will free the reference to this
         object.
         """
+        if isinstance(objectID, unicode):
+            objectID = objectID.encode('utf8')
         refs = self.localObjects[objectID].decref()
```
adiroiban: since localObjects is a public API, maybe it should be replaced with a dict like implementation which converts the objectID on the fly . I guess that external code might also want to access it and the forget to do the conversion..

or maybe it should be private and it needs to be deprecated as a public API

twisted/spread/pb.py
```diff
@@ -882,6 +903,9 @@ def _recvMessage(self, findObjMethod, requestID, objectID, message, answerRequir
         Look up message based on object, unserialize the arguments, and
         invoke it with args, and send an 'answer' or 'error' response.
         """
+        if not isinstance(message, str):
+            message = message.decode('utf8')
```
adiroiban: why decoding here since remoteMessageReceived will also do the decoding ?

twisted/spread/pb.py
adiroiban: I saw there are many other places in which we convert the key for localObjects. With that in mind I think that a dict like class is needed.

With manual conversion we might forget to convert at some place and then we end up with errors.

twisted/spread/pb.py
```diff
@@ -557,12 +564,17 @@ def expressionReceived(self, sexp):
         """
         if isinstance(sexp, list):
             command = sexp[0]
+
+            if not isinstance(command, str):
+                command = command.decode('utf8')
```
adiroiban: can expressionReceived be called with non-bytes commands ?

twisted/spread/jelly.py
```diff
@@ -889,18 +915,9 @@ def _unjelly_instance(self, rest):
...
-            state = self.unjelly(rest[1])
-            inst.__setstate__(state)
-        else:
-            state = self.unjelly(rest[1])
-            inst = _newInstance(clz, state)
-        if hasattr(clz, 'postUnjelly'):
-            self.postCallbacks.append(inst.postUnjelly)
-        return inst
+        if not _PY3 and type(clz) is not _OldStyleClass:
+            raise InsecureJelly("Legacy 'instance' found with new-style class")
```
adiroiban: I don't see any tests for this change ... if we are testing just for InsecureJelly I think that we should also assert the InsecureJelly's message

adiroiban:
I have re-run the 3.3 job on travis-ci and it still fails ...  while the tests seems unrelated, maybe the new spread tests are having some side effects .

I have re-run it on trunk on travis-ci and it pass.

I guess there is something wrong about the current changes :)

glyph:
Thanks for the review comments @adiroiban .  I will look into the py3.3 stuff...

rodrigc:
I can confirm that twisted.trial.test.test_tests.UnhandledDeferredTests.test_doesntBleed is consistently failing under Python 3.3.
I put a print() in the test to print out the value of **gc.garbage** inside the test and found:

```
[<twisted.spread.test.test_pb.DumbPub object at 0x108544650>, <twisted.spread.pb.RemoteReference object at 0x109f15110>, <twisted.spread.test.test_pb.DumbPub object at 0x109ec0290>, <twisted.spread.pb.RemoteReference object at 0x108533f90>, <twisted.spread.pb.RemoteReference object at 0x108e465d0>, <twisted.spread.pb.RemoteReference object at 0x10848ec10>, <twisted.spread.pb.RemoteReference object at 0x10848ed10>, <twisted.spread.pb.RemoteReference object at 0x10848ed50>, <twisted.spread.pb.RemoteReference object at 0x10848ebd0>, <twisted.spread.pb.RemoteReference object at 0x10851c150>, <twisted.spread.pb.RemoteReference object at 0x10851c5d0>, <twisted.spread.pb.RemoteReference object at 0x10851e850>, <twisted.spread.pb.RemoteReference object at 0x1082e1950>, <twisted.spread.pb.RemoteReference object at 0x1082e1e90>, <twisted.spread.pb.RemoteReference object at 0x1082e1250>, <twisted.spread.pb.RemoteReference object at 0x10850bc10>, <twisted.spread.pb.RemoteReference object at 0x10850bf10>, <twisted.spread.pb.RemoteReference object at 0x10850be90>, <twisted.spread.pb.RemoteReference object at 0x108f73490>, <twisted.spread.pb.RemoteReference object at 0x10842a290>, <twisted.spread.pb.RemoteReference object at 0x10842a250>, <twisted.spread.pb.RemoteReference object at 0x108422110>, <twisted.spread.pb.RemoteReference object at 0x108422710>, <twisted.spread.pb.RemoteReference object at 0x10a57f850>, <twisted.spread.pb.RemoteReference object at 0x10a57fbd0>, <twisted.spread.pb.RemoteReference object at 0x10a57fed0>, <twisted.spread.pb.RemoteReference object at 0x10a57f7d0>, <twisted.spread.pb.RemoteReference object at 0x10a57fb10>, <twisted.spread.pb.RemoteReference object at 0x10a57fad0>, <twisted.spread.pb.RemoteReference object at 0x10a57f9d0>, <twisted.spread.pb.RemoteReference object at 0x10a57fdd0>, <twisted.spread.pb.RemoteReference object at 0x10a57f4d0>, <twisted.spread.pb.RemoteReference object at 0x10a57f450>, <twisted.spread.pb.RemoteReference object at 0x10a57f910>, <twisted.spread.pb.RemoteReference object at 0x10a57f510>,
```

with over 1000 entries of twisted.spread.pb.RemoteReference object.

It looks like this is triggering behavior which was changed in Python 3.4 under https://www.python.org/dev/peps/pep-0442/

glyph:
@rodrigc Thanks for investigating and confirming.

glyph:
The link to the PEP was particularly useful :)

rodrigc:
👍 This looks good.  I tested it on my system with Python 3.3.6, and could not reproduce the previous test errors in garbage collection.  Please merge.

glyph:
I'm a _little_ unhappy with my own test coverage here, but I think that will actually be easier to address after landing this, so I'm going to go with @rodrigc's code review :-).  (Presuming required statuses pass...)

glyph:
stopping the OS X travis build because https://www.traviscistatus.com/incidents/4mvp857qx8bw

glyph:
Augh, cancelling the build, even though it's in `allow_failures`, causes the build to fail.  Everything else has passed though so I'm going to push the big _red_ button.

Ugh.

rodrigc:
@glyph thanks for porting PB to Py3.  That has opened the door to me submitting a lot of Py3 fixes upstream to buildbot: https://lists.buildbot.net/pipermail/devel/2017-January/012304.html

glyph:
@rodrigc -

> thanks for porting buildbot to Py3

I think you mean "thanks for porting _PB_ to Py3"? :)  If so - you're welcome, glad to help!

rodrigc:
Yeah I meant PB!

PR #437, opened 2016-07-28
Fix #8699 ascii encoding error in flattenEvent()

### LINKED ISSUE #8699: flattenEvent() fails with "'ascii' codec can't encode character" in Python 2.7
|[znerol's avatar](https://github.com/znerol)| @znerol reported|
|-|-|
|Trac ID|trac#8699|
|Type|defect|
|Created|2016-07-28 05:59:35Z|

backtrace ends in _flatten.py

```
twisted/logger/_flatten.py(119)flattenEvent()
-> flattenedValue = conversionFunction(fieldValue)
(Pdb) conversionFunction
<type 'str'>
```

I guess conversionFunction should be unicode in Python 2.7 and str in Python 3.

```
trac-id__8699 8699
type__defect defect
reporter__znerol znerol
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__closed closed
resolution__fixed fixed
component__core core
keywords__None None
time__1469685575533160 1469685575533160
changetime__1470085969851774 1470085969851774
version__None None
owner__Craig_Rodrigues__rodrigc_____ Craig Rodrigues <rodrigc@...>

```
</details>

znerol:
Fixes https://twistedmatrix.com/trac/ticket/8699

### DIFF
--- a/twisted/logger/_flatten.py
+++ b/twisted/logger/_flatten.py
@@ -111,7 +111,7 @@ def flattenEvent(event):
         if conversion == "r":
             conversionFunction = repr
         else:  # Above: if conversion is not "r", it's "s"
-            conversionFunction = str
+            conversionFunction = unicode
 
         if callit:
             fieldValue = fieldValue()
--- a/twisted/logger/test/test_flatten.py
+++ b/twisted/logger/test/test_flatten.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright (c) Twisted Matrix Laboratories.
 # See LICENSE for details.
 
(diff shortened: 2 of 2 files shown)


rodrigc:
What input did you use to trigger this problem on Python 2.7?
Can you add a testcase in twisted/logger/test/test_flatten.py ?

rodrigc:
Whoops!  Sorry, I saw you added the testcase to test_flatten.py!

rodrigc:
@wsanchez can you review this?  I think it looks OK, but just want to confirm.

wsanchez:
Yes, this looks correct to me.

rodrigc:
Merged in https://github.com/twisted/twisted/pull/452

glyph:
This definitely looks wrong to me.  Previously this would have invoked `__str__` on py2 and now it invokes `__unicode__`; also, `__repr__` can have the same problem.

I am not rolling back though because there is no other test, but I think we need to continue investigation here to see if this is really the right solution...

glyph:
Also not rolling back because it is quite possibly _less_ wrong than what was there before, so in that sense, it does look correct :)

PR #443, opened 2016-07-28
Do not use inspect.getargspec() in twisted/names/dns.py on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8703

### DIFF
--- a/twisted/names/dns.py
+++ b/twisted/names/dns.py
@@ -2073,16 +2073,25 @@ def _compactRepr(obj, alwaysShow=None, flagNames=None, fieldNames=None,
 
     out = ['<', obj.__class__.__name__]
 
-    # Get the argument names and values from the constructor.
-    argspec = inspect.getargspec(obj.__class__.__init__)
-    # Reverse the args and defaults to avoid mapping positional arguments
-    # which don't have a default.
-    defaults = dict(zip(reversed(argspec.args), reversed(argspec.defaults)))
-    for name in fieldNames:
-        defaultValue = defaults.get(name)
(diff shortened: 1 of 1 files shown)


twisted/names/dns.py
```diff
@@ -2073,16 +2073,25 @@ def _compactRepr(obj, alwaysShow=None, flagNames=None, fieldNames=None,
...
+        # Get the argument names and values from the constructor.
+        argspec = inspect.getargspec(obj.__class__.__init__)
+        # Reverse the args and defaults to avoid mapping positional arguments
+        # which don't have a default.
+        defaults = dict(zip(reversed(argspec.args), reversed(argspec.defaults)))
+        for name in fieldNames:
+            defaultValue = defaults.get(name)
+            fieldValue = getattr(obj, name, defaultValue)
+            if (name in alwaysShow) or (fieldValue != defaultValue):
+                out.append(' %s=%r' % (name, fieldValue))
```
Lukasa: So I appreciate this is the kind of feedback you don't really want for changes like this, but this code block is sufficiently large now that I think it's obscuring the intent of the function. Do you mind hoisting this out to a helper function?

rodrigc: I pulled it out into a helper function.

twisted/names/dns.py
```diff
@@ -2098,6 +2125,9 @@ def _compactRepr(obj, alwaysShow=None, flagNames=None, fieldNames=None,
 
 
 
+
+
+
```
Lukasa: Mind stripping all this bonus whitespace?

PR #444, opened 2016-07-29
Fix twisted.conch.test.test_recvline on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8709


rodrigc:
@einarfd @acabhishek942 please review this conch fix

acabhishek942:
looks good to me

rodrigc:
@acabhishek942 thanks for the review.  Today as I was working on conch,
I found a new trick for debugging with Twisted.

When I was debugging a test, I put the following somewhere at the very beginning
of the test's execution

```

debugFile = open("/tmp/log.txt", "w")
log.startLogging(debugFile)

```

After that, any print() statements that I put in the code would get logged to /tmp/log.txt.
What was even more useful is that any exceptions would also get logged to that file.

This was handy for debugging str vs. bytes exceptions thrown deep in the callback chain
in conch.  Conch makes heavy use of callbacks!!

PR #445, opened 2016-07-29
[#8710] Actually log asynchronous failures from listen.

mithrandi:
https://twistedmatrix.com/trac/ticket/8710


glyph:
I feel like the weird compatibility variable in question here could be better-documented, because if I didn't have the weird backwards-compatibility context I would not know what the test was actually testing.  But the NEWS file quite straightforwardly details the user-visible consequence of the fix, so I am 👍 .  Just waiting for a passing / merged build result.

glyph:
Our poor build infrastructure...

mithrandi:
The "documentation" build is failing due to errors unrelated to this branch; they seem to have been present in trunk for some time, but not flagged due to issues with the "new errors" detection? Anyhow, I'm going to ignore those and merge anyway.

PR #446, opened 2016-07-30
[#8367] Implement a Twisted reactor on top of asyncio

hawkowl:


twisted/internet/test/test_threads.py
```diff
@@ -181,8 +181,12 @@ def test_cleanUpThreadPoolEvenBeforeReactorIsRun(self):
         reactor = self.buildReactor()
         threadPoolRef = ref(reactor.getThreadPool())
         reactor.fireSystemEvent("shutdown")
-        gc.collect()
-        self.assertIsNone(threadPoolRef())
+
+        if reactor.__class__.__name__ == "AsyncioSelectorReactor":
+            self.assertIs(reactor.threadpool, None)
```
rodrigc: Better to use assertIsNone() here

twisted/internet/asyncioreactor.py
```diff
@@ -0,0 +1,261 @@
...
+           reader in self._continuousPolling._readers:
+            return
+
+        fd = reader.fileno()
+        try:
+            self._asyncioEventloop.add_reader(fd, callWithLogger, reader,
+                                              self._readOrWrite, reader,
+                                              True)
+            self._readers[reader] = fd
+        except IOError as e:
```
markrwilliams, line 152: There's a bug in the standard library's `selectors` module that prevents that stdio test from passing.  [I put the gory details into trac](https://twistedmatrix.com/trac/ticket/8367#comment:7), but you can get it to pass by adding the following at the beginning of this `IOError` block and the one in `addWriter`:

``` python
self._asyncioEventloop._selector.unregister(fd)
```

twisted/internet/asyncioreactor.py
```diff
@@ -0,0 +1,261 @@
...
+try:
+    from asyncio import new_event_loop
+except ImportError:
+    raise ImportError("Requires asyncio.")
+
+
+
+class _DCHandle(object):
+    """
+    Wrapper for L{asyncio.Handle} to be used by
```
markrwilliams: It looks like the point of this class is to provide a single reference the `DelayedCall`'s callbacks can close over that can resolve to arbitrary asyncio handles.  Maybe a better docstring would be something like:

```
    Wraps ephemeral L{asyncio.Handle} instances.  Callbacks can close
    over this and use it as a mutable reference to asyncio Handles.

    @ivar handle: The current L{asyncio.Handle}
```

twisted/internet/posixbase.py
```diff
@@ -611,6 +611,164 @@ def _doReadOrWrite(self, selectable, fd, event):
...
+            self._loop = None
+
+
+    def iterate(self):
+        """
+        Call C{doRead} and C{doWrite} on all readers and writers respectively.
+        """
+        for reader in list(self._readers):
+            self._doReadOrWrite(reader, reader, self._POLL_IN)
+        for reader in list(self._writers):
```
markrwilliams: Whoops -- this isn't a `reader`, it's a `writer`!

twisted/internet/asyncioreactor.py
```diff
@@ -0,0 +1,319 @@
...
+
+        * The second C{connection.write(intToByte(value))} call hits the
+        asyncio reactor's C{addWriter} method, which hits the C{add_writer}
+        method. But the loop's selector's get_key method now returns a
+        C{SelectorKey}! Now the asyncio reactor's C{addWriter} method thinks
+        the asyncio loop will watch the file descriptor, even though it won't.
+        """
+        try:
+            self._asyncioEventloop._selector.unregister(fd)
+        except:
```
markrwilliams, line 121: So this is a no-op, and all the exceptions should this might raise now or in the future should get caught.

However, I _think_ I prefer `KeyError` -- in the original bug, that was the only exception that could be raised if this referred to an fd that hadn't been monitored.  I guess I'm worried about `KeyboardInterrupt` and other `BaseException`s that somehow happen here.

I don't think this is a merge-blocker, though.

markrwilliams, line 121: Also you're very kind to mention me :)

twisted/test/test_internet.py
```diff
@@ -686,6 +686,7 @@ def test_callLaterUsesReactorSecondsInDelayedCall(self):
             self.assertEqual(call.getTime(), 105)
         finally:
             reactor.seconds = oseconds
+        call.cancel()
```
markrwilliams, line 689: Whoa, how long was this _not_ happening??  This seems like a worthwhile patch in and of itself, like a freebie!

markrwilliams:
Closing and reopening to re-trigger Buildbots.

twisted/internet/asyncioreactor.py
hawkowl, line 121: My concern is if asyncio renames _selector, we're talking to internal API after all :)

PR #448, opened 2016-07-31
"python -m twisted.scripts.trial" and "python -m twisted.trial should be" valid ways to invoke trial

rodrigc:
https://twistedmatrix.com/trac/ticket/8712


markrwilliams:
@hawkowl this looks good to me!  thanks (and thanks @rodrigc!)

hawkowl:
@markrwilliams don't forget to remove the tag on the ticket for those sweet Twisted Points ;)

PR #449, opened 2016-07-31
Fix twisted.python.filepath.setContent() on Windows with Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8610

### DIFF
--- a/twisted/python/filepath.py
+++ b/twisted/python/filepath.py
@@ -1530,7 +1530,7 @@ class FilePath(AbstractFilePath):
             f.write(content)
         if platform.isWindows() and exists(self.path):
             os.unlink(self.path)
-        os.rename(sib.path, self.path)
+        os.rename(sib.path, self.asBytesMode().path)
 
 
     def __cmp__(self, other):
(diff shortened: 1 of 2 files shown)


rodrigc:
@hawkowl please review, this is cherry-picked from https://github.com/twisted/twisted/blob/moar-windows-8025-7/twisted/python/filepath.py on your branch

hawkowl:
@rodrigc I can't review this, since I wrote the original code :)

paging @glyph

glyph:
This _looks_ correct to me, but is there a skipped test someplace that could be un-skipped, or should a new test be written?

rodrigc:
@glyph without this fix, many, many tests fail on Windows with Python 3
because they call FilePath.setContent().
This is one step towards getting a usable Windows Python 3 buildbot or appveyor.

glyph:
@rodrigc Gotcha.

glyph:
Should the ticket here be in review?

rodrigc:
I put it in review

twisted/python/filepath.py
```diff
@@ -1529,8 +1529,8 @@ def setContent(self, content, ext=b'.new'):
         with sib.open('w') as f:
             f.write(content)
         if platform.isWindows() and exists(self.path):
-            os.unlink(self.path)
-        os.rename(sib.path, self.path)
+            os.remove(self.path)
```
Lukasa: What's the reasoning for this change? It seems like needless churn, but it's also harmless, so I'm not going to block review on it. Generally speaking it'd be nice if we didn't make a change like this, so please consider removing it.

rodrigc: That's not a big deal, I'll change it back.  I took this patch from hawkowl's branch

PR #451, opened 2016-08-01
[#8714] Fix negative jittered reconnection delay

haphut:
https://twistedmatrix.com/trac/ticket/8714

### DIFF
--- a/twisted/test/test_factories.py
+++ b/twisted/test/test_factories.py
@@ -8,6 +8,7 @@ Test code for basic Factory classes.
 from __future__ import division, absolute_import
 
 import pickle
+import random
 
 from twisted.trial.unittest import TestCase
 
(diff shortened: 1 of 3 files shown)


twisted/test/test_factories.py
```diff
@@ -91,6 +92,34 @@ def connect(self):
...
+    def test_jitteredDelayIsNonNegative(self):
+        """
+        Reconnection delay should not be negative.
+        """
+        class FakeClock:
+            def callLater(self, _seconds, _f, *args, **kw):
+                pass
+
+        # Let's not mess with other tests involving randomness.
+        original = random.getstate()
```
adiroiban, line 104: well, we should call `self.addCleanup(random.setstate, original)` just after this call... in case the tests fails the random state is not reverted.

.... but the best way to have clean tests is to allow injecting the `random` dependecy ... just like we inject the fake clock

```
class ReconnectingClientFactory:
    _random = random

    def someMethod(self):
        self.delay = max(self._random())
```

twisted/test/test_factories.py
```diff
@@ -91,6 +92,34 @@ def connect(self):
...
+
+        # Values picked for demonstration.
+        seed = 405410
+        jitter = 0.21
+
+        random.seed(seed)
```
adiroiban, line 110: with an injectable random dependecy we can just use

```
f._random = Random(405410)
```

and don't worry about the cleanup

glyph:

This pull request hasn't been updated in quite some time, and has merge conflicts, so I'm going to close it to reduce the clutter in the PR queue.

This is not a rejection, just a clean-up; please do not hesitate to re-open a new PR with updated code when you have time.

PR #453, opened 2016-08-01
[#8088] Support async/await with Deferreds

hawkowl:
https://twistedmatrix.com/trac/ticket/8088


twisted/internet/defer.py
```diff
@@ -661,6 +661,66 @@ def __str__(self):
...
+
+    def _setResult(self, value):
+        self.result = value
+
+    def __next__(self):
+        if self.result is not _NO_RESULT:
+            raise StopIteration(self.result)
+        return self.d
+
+    def __await__(self):
```
1st1: Should also implement `__iter__` (also returning `self`).

twisted/internet/defer.py
```diff
@@ -661,6 +661,66 @@ def __str__(self):
     __repr__ = __str__
 
 
+    def __await__(self):
+        awaiter = _MakeDeferredAwaitable(self)
```
1st1: Maybe it's possible to implement the `_MakeDeferredAwaitable` class as a generator... The performance would be a bit better (although the code would be harder to read...)

twisted/internet/defer.py
hawkowl: Right. Should this actually need to implement `__await__` at all? Since Deferred is the thing with await, I just need to change it to call `__iter__` and rename the `__await__` on this object...

twisted/internet/defer.py
hawkowl: I'm unsure what would make it any more performant, could you go into a bit more detail on that?

twisted/internet/defer.py
1st1: Actually it should only implement `__iter__`, and no `__await__`:

``` python
class Deferred:
    def __await__(self):
        return _MakeDeferredAwaitable(self)

class _MakeDeferredAwaitable(object):
    def __iter__(self):
        return self

    def __next__(self):
         if self.result is not _NO_RESULT:
             raise StopIteration(self.result)
         return self.d
```

Per PEP 492 -- `__await__` should return an iterator (not an awaitable).  Also, you might want to implement `_MakeDeferredAwaitable.throw` if it's possible in Twisted to throw exceptions to the `await` points.

hawkowl: @1st1 right, hmm, that's fixable.

re throw: it's not, because the Deferred will fire with a `Failure`; which `inlineCallbacks` will then throw into it (https://github.com/twisted/twisted/pull/453/files/5ffe7c0aed2c3dcea26ab6c6af8cb1ac7bf5f216#diff-41cf94828242f95ae48c543d7db3309aR1183).

twisted/internet/defer.py
1st1: It's just that `__next__` and `__iter__` will be in C. Also, `YIELD_FROM` opcode has a fast-path for generator objects. This is still a micro-optimization, but you might want to write a micro-benchmark.

hawkowl: I think for now I'll keep it like this -- it's getting a bit complex to review for those that don't know it as well as either of us do, so I can do it once I have a benchmark :)

1st1: Yeah, that's the right thing to do. Again, it's a micro-optimization that might or might not be worth implementing.

twisted/internet/test/test_await.py
```diff
@@ -0,0 +1,21 @@
...
+    from twisted.python.filepath import FilePath
+
+    _path = FilePath(__file__).parent().child("_awaittests.py.3only")
+
+    _g = {}
+    execfile(_path.path, _g)
+
+    AwaitTests = _g["AwaitTests"]
+    __all__ = ["AwaitTests"]
+except:
```
glyph: drive-by comment!

This is too broad, in that weird misconfigurations will raise weird errors and these tests silently won't run.
1. Move the `compat` and `FilePath` imports, and the `_path` construction, out of the `try` block. Even the `_g = {}`.  Presumably the only thing that should fail is the `execfile`, so that's the only thing that should be in the `try`.
2. Catch only the exception that's the problem (`SyntaxError`, I'm guessing?) not bare-word except.
3. Even the silly obvious can't fail stuff after the `execfile` should be in the `else` suite of a `try`/`except`/`else`; you don't want to swallow any more errors than you have to.
4. Consider adding an `AwaitTests` class that just has a single, empty test method and a "skip" attribute, so there's at least visibility that the test was there, and it was intentionally not run.
5. Maybe don't even have a `try`.  This is _always_ supposed to work on 3.5, right?  So just compare `sys.version_info` and do it unconditionally.

This comment brought to you by literally dozens of saturday nights ruined by `NameError` or `AttributeErrors`s in some obscure `try` block that "would only fail in this one case".

glyph: I guess "conditionally", not "unconditionally", but still; what I mean is, "without a `try`".

docs/core/howto/defer-intro.rst
```diff
@@ -390,6 +390,67 @@ With ``inlineCallbacks``, we can rewrite this as:
...
+           log.failure("makeRequest failed due to connection error")
+           return []
+
+   def do():
+       d = ensureDeferred(getUsers())
+       d.addCallback(print)
+       return d
+
+
+When writing coroutines, you do not need to use :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>` when you are writing a coroutine which calls other coroutines which await on Deferreds; you can just ``await`` on it directly.
```
gsemet, line 432: there is repetition of "when writing"

PR #438, opened 2016-07-28
8700 einarfd py3 conch client default

einarfd:
This fixes https://twistedmatrix.com/trac/ticket/8700 and is a port of conch.client.default to Python 3, with some added tests and a very minor bug fix thrown in for good measure. 

Unfortunately stdin/stdout is marked different on the Python versions, so there are a couple of lines that are Python 3 specific, and the coverage of the patch will therefore not be a 100%.

PR #454, opened 2016-08-02
Port twisted.conch.client.default to Python 3

rodrigc:
This work was originally done in this pull request: https://github.com/twisted/twisted/pull/438

With one additional patch to skip tests that require an interactive tty
if we are not on an interactive tty.

Tickets addressed:
http://twistedmatrix.com/trac/ticket/8700
http://twistedmatrix.com/trac/ticket/8715

### DIFF
--- a/twisted/conch/client/default.py
+++ b/twisted/conch/client/default.py
@@ -189,7 +189,8 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         if prompt:
             prompt = nativeString(prompt)
         else:
-            prompt = "%s@%s's password: " % (nativeString(self.user), self.transport.transport.getPeer().host)
+            prompt = ("%s@%s's password: " %
+                (nativeString(self.user), self.transport.transport.getPeer().host))
         try:
             # We don't know the encoding the other side is using,
             # signaling that is not part of the SSH protocol. But
--- a/twisted/conch/client/default.py
+++ b/twisted/conch/client/default.py
@@ -295,7 +295,7 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         and return them.
 
         @return: File objects for reading and writing to /dev/tty,
-                 standard input and standard output.
+                 corresponding to standard input and standard output.
         @rtype: A L{tuple} of L{io.TextIOWrapper} on Python 3.
                 A L{tuple} of binary files on Python 2.
         """
--- a/twisted/conch/client/default.py
+++ b/twisted/conch/client/default.py
@@ -89,8 +89,10 @@ def verifyHostKey(transport, host, pubKey, fingerprint):
     return kh.verifyHostKey(ui, actualHost, host, actualKey)
 
 
+
 def isInKnownHosts(host, pubKey, options):
-    """checks to see if host is in the known_hosts file for the user.
+    """
+    checks to see if host is in the known_hosts file for the user.
     returns 0 if it isn't, 1 if it is and is the same, 2 if it's changed.
     """
     keyType = common.getNS(pubKey)[0]
@@ -136,6 +138,7 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         if not options.identitys:
             options.identitys = ['~/.ssh/id_rsa', '~/.ssh/id_dsa']
 
+
     def serviceStarted(self):
         if 'SSH_AUTH_SOCK' in os.environ and not self.options['noagent']:
             log.msg('using agent')
@@ -146,10 +149,12 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         else:
             userauth.SSHUserAuthClient.serviceStarted(self)
 
+
     def serviceStopped(self):
         if self.keyAgent:
             self.keyAgent.transport.loseConnection()
             self.keyAgent = None
+
 
     def _setAgent(self, a):
         self.keyAgent = a
@@ -157,8 +162,10 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         d.addBoth(self._ebSetAgent)
         return d
 
+
     def _ebSetAgent(self, f):
         userauth.SSHUserAuthClient.serviceStarted(self)
+
 
     def _getPassword(self, prompt):
         """
@@ -176,6 +183,7 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
             except (KeyboardInterrupt, IOError):
                 print()
                 raise ConchError('PEBKAC')
+
 
     def getPassword(self, prompt = None):
         if prompt:
--- a/twisted/conch/client/default.py
+++ b/twisted/conch/client/default.py
@@ -14,6 +14,8 @@ interact with a known_hosts database, use L{twisted.conch.client.knownhosts}.
 from __future__ import print_function
 
 from twisted.python import log
+from twisted.python.compat import (
+    nativeString, raw_input, _PY3, _b64decodebytes as decodebytes)
 from twisted.python.filepath import FilePath
 
 from twisted.conch.error import ConchError
@@ -24,7 +26,10 @@ from twisted.conch.client.knownhosts import KnownHostsFile, ConsoleUI
 
 from twisted.conch.client import agent
 
-import os, sys, base64, getpass
+import os, sys, getpass, contextlib
+
+if _PY3:
+    import io
 
 # The default location of the known hosts file (probably should be parsed out
 # of an ssh config file someday).
@@ -110,7 +115,7 @@ def isInKnownHosts(host, pubKey, options):
             if hostKeyType != keyType: # incorrect type of key
                 continue
             try:
-                decodedKey = base64.decodestring(encodedKey)
+                decodedKey = decodebytes(encodedKey)
             except:
                 continue
             if decodedKey == pubKey:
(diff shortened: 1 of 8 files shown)


twisted/conch/client/default.py
```diff
@@ -156,21 +161,25 @@ def _ebSetAgent(self, f):
...
+                print()
+                raise ConchError('PEBKAC')
 
     def getPassword(self, prompt = None):
-        if not prompt:
-            prompt = "%s@%s's password: " % (self.user, self.transport.transport.getPeer().host)
+        if prompt:
+            prompt = nativeString(prompt)
+        else:
+            prompt = "%s@%s's password: " % (nativeString(self.user), self.transport.transport.getPeer().host)
```
glyph: This line's 110 characters long, could you wrap it?

twisted/conch/client/default.py
```diff
@@ -156,21 +161,25 @@ def _ebSetAgent(self, f):
...
-            raise ConchError('PEBKAC')
+        with self._replaceStdoutStdin():
+            try:
+                p=getpass.getpass(prompt)
+                return p
+            except (KeyboardInterrupt, IOError):
+                print()
+                raise ConchError('PEBKAC')
 
     def getPassword(self, prompt = None):
```
glyph, line 190: The type of `prompt` seems pretty important now; would you mind adding a docstring that specifies its `@type`?  Looks like it's `native L{str}` if I understand properly?

twisted/conch/client/default.py
```diff
@@ -230,10 +239,10 @@ def getPrivateKey(self):
             return defer.succeed(keys.Key.fromFile(file))
         except keys.EncryptedKeyError:
             for i in range(3):
-                prompt = "Enter passphrase for key '%s': " % \
-                    self.usedFiles[-1]
+                prompt = "Enter passphrase for key '%s': " % self.usedFiles[-1]
                 try:
-                    p = self._getPassword(prompt)
+                    p = self._getPassword(prompt).encode(
+                        sys.getfilesystemencoding())
```
glyph, line 264: This is totally optional feedback, since it's a private method, but I think this belongs in `_getPassword`.

twisted/conch/client/default.py
```diff
@@ -246,18 +255,46 @@ def getPrivateKey(self):
 
     def getGenericAnswers(self, name, instruction, prompts):
         responses = []
-        try:
-            oldout, oldin = sys.stdout, sys.stdin
-            sys.stdin = sys.stdout = open('/dev/tty','r+')
+        with self._replaceStdoutStdin():
             if name:
-                print(name)
+                print(name.decode("UTF8"))
```
glyph: In this interests of consistency, this should be `"utf-8"` (117 occurrences in the codebase) not `"UTF8"` (24 occurrences, only one of which is in an `.encode` or `.decode` call.)  I also dimly recall that choice was intentional because there are some common misconfigurations of Python which will cause "UTF8" to break but "utf-8" works everywhere.

twisted/conch/client/default.py
```diff
@@ -156,21 +161,25 @@ def _ebSetAgent(self, f):
...
-            sys.stdin = sys.stdout = open('/dev/tty','r+')
-            p=getpass.getpass(prompt)
-            sys.stdout,sys.stdin=oldout,oldin
-            return p
-        except (KeyboardInterrupt, IOError):
-            print()
-            raise ConchError('PEBKAC')
+        with self._replaceStdoutStdin():
+            try:
+                p=getpass.getpass(prompt)
```
glyph: I wish twistedchecker were working; sorry for the nitpick.  There should be spaces around the equal sign.

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
         self.patch(client, '_getPassword', _getPassword)
         return client.getPrivateKey().addCallback(_cbGetPrivateKey)
+
+
+    def test_getPassword(self):
+        """
+        Verify that the getPassword function in L{SSHUserAuthClient}
+        works, when prompt is sent, and fetching the password from the
+        user succeeds.
+
```
glyph: (nitpick: spurious blank line)

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
             self.assertFalse(key.isPublic())
             self.assertEqual(key, rsaPrivate)
 
         self.patch(client, '_getPassword', _getPassword)
         return client.getPrivateKey().addCallback(_cbGetPrivateKey)
+
+
+    def test_getPassword(self):
+        """
+        Verify that the getPassword function in L{SSHUserAuthClient}
```
glyph: Test docstrings should not begin with "verify that", "test that", "correctly", etc.  See https://jml.io/pages/test-docstrings.html for the standard we should use.

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
+        d = client.getPassword()
+        d.addCallback(self.assertEqual, b'bad password')
+        return d
+
+    test_getPassword.skip = windowsSkip or ttySkip
+
+
+    def test_getPasswordPrompt(self):
+        """
+        Verify that the getPassword function in L{SSHUserAuthClient}
```
glyph: See above about "Verify"

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
+        d = client.getPassword(prompt)
+        d.addCallback(self.assertEqual, b'bad password')
+        return d
+
+    test_getPasswordPrompt.skip = windowsSkip or ttySkip
+
+
+    def test_getPasswordConchError(self):
+        """
+        Verify that the getPassword function in L{SSHUserAuthClient}
```
glyph: See above about "Verify"

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
+                [stdout, stdin], [sys.stdout, sys.stdin])
+            return fail
+        self.assertFailure(d, ConchError)
+
+    test_getPasswordConchError.skip = windowsSkip or ttySkip
+
+
+    def test_getGenericAnswers(self):
+        """
+        Verify that the getGenericAnswers function in L{SSHUserAuthClient}
```
glyph: See above about "Verify"

twisted/conch/test/test_default.py
```diff
@@ -147,23 +169,129 @@ def test_getPrivateKeyPassphrase(self):
...
+            return "getpass"
+
+        self.patch(default.getpass, 'getpass', getpass)
+
+        def raw_input(prompt):
+            self.assertEqual(prompt, "raw_input prompt")
+            return "raw_input"
+
+        self.patch(default, 'raw_input', raw_input)
+        d =client.getGenericAnswers(
```
glyph: twistedchecker, where art thou: there should be a second space after the equals sign

twisted/conch/client/default.py
```diff
@@ -246,18 +255,46 @@ def getPrivateKey(self):
...
-        finally:
-            sys.stdout,sys.stdin=oldout,oldin
         return defer.succeed(responses)
+
+
+    @classmethod
+    def _openTty(cls):
+        """
+        Open /dev/tty as two streams one in read, one in write mode,
+        and return them.
```
glyph, line 295: As per the doc standard, this needs a `@return` which describes its return value; it should be crisply documented that it returns a `TextIOWrapper` on py3 but a binary file on python 2, matching the behavior of `sys.stdin` and `sys.stdout`.  (You learn something every day, I guess!  I thought that `sys.stdin` et. al. were rather informally specified but checking up on this branch I learned that they are in fact always unicode on py3 and always bytes on py2, albeit sometimes bytes with an encoding that can accept `unicode` writes on py2.)

Speaking of accepting `unicode` writes; there's a weird nuance here that I'm not sure how to fully reproduce.  On python 2, `sys.stdin` normally has an `encoding` attribute.  Although there's a C API to set this (`PyFile_SetEncoding`) it seems like the attribute is just completely read-only from Python.  `io.open` doesn't support `encoding=` with binary mode.  It's probably fine not to care for now but this is something we should keep in mind for the future.

twisted/conch/client/default.py
```diff
@@ -246,18 +255,46 @@ def getPrivateKey(self):
...
+        stdout = open("/dev/tty", "wb")
+        if _PY3:
+            stdin = io.TextIOWrapper(stdin)
+            stdout = io.TextIOWrapper(stdout)
+        return stdin, stdout
+
+
+    @classmethod
+    @contextlib.contextmanager
+    def _replaceStdoutStdin(cls):
```
glyph, line 312: Another (optional) thought - maybe this should be somewhere more central?  there are a dozen other places in Twisted that set sys.stdout; some of them might benefit from the improved discipline of a contextmanager as well.

twisted/conch/client/default.py
```diff
@@ -246,18 +255,46 @@ def getPrivateKey(self):
...
+    @classmethod
+    def _openTty(cls):
+        """
+        Open /dev/tty as two streams one in read, one in write mode,
+        and return them.
+        """
+        stdin = open("/dev/tty", "rb")
+        stdout = open("/dev/tty", "wb")
+        if _PY3:
+            stdin = io.TextIOWrapper(stdin)
```
glyph, line 305: Should these be explicitly setting some encoding, or will this do the right thing?

twisted/test/test_compat.py
```diff
@@ -861,3 +861,27 @@ def test_unichr(self):
...
         self.assertEqual(unichr(0x2603), u"\N{SNOWMAN}")
+
+
+class RawInputTests(unittest.TestCase):
+    """
+    Tests for L{raw_input}
+    """
+    def test_raw_input(self):
+        """
+        Verify that raw_input exists and that it works.
```
glyph: Same comment about "verify".

twisted/conch/client/default.py
einarfd, line 305: By default the encoding is set to locale.getpreferredencoding(False), which imo. seems like a reasonable choice for us here as well.

twisted/conch/client/default.py
einarfd, line 264: _getPassword is used once where the result is passed on the wire, and once when it is used as a pass phrase when opening a ssh key. For the wire one it uses sys.getdefaultencoding() and for the file one it uses sys.getfilesystemencoding(). Since those two are different, I put the encode on the outside. But adding the encoding as a argument to _getPassword would work as well.

PR #455, opened 2016-08-02
Don't explode when the client resets a stream while it uploads data.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8682

### DIFF
--- a/twisted/web/_http2.py
+++ b/twisted/web/_http2.py
@@ -686,7 +686,18 @@ class H2Connection(Protocol, TimeoutMixin):
         @type increment: L{int}
         """
         # TODO: Consider whether we want some kind of consolidating logic here.
-        self.conn.increment_flow_control_window(increment, stream_id=streamID)
+        try:
+            self.conn.increment_flow_control_window(
+                increment, stream_id=streamID
+            )
+        except h2.exceptions.StreamClosedError:
+            # The stream got reset and we haven't worked it out yet. The stream
+            # window doesn't matter now. We still want to increment the
+            # connection window though.
+            # NOTE: This branch may become unneeded in the future if
+            # https://github.com/python-hyper/hyper-h2/issues/228 happens.
+            pass
+
         self.conn.increment_flow_control_window(increment, stream_id=None)
         self.transport.write(self.conn.data_to_send())
 
--- a/twisted/web/_http2.py
+++ b/twisted/web/_http2.py
@@ -692,8 +692,13 @@ class H2Connection(Protocol, TimeoutMixin):
             )
         except h2.exceptions.StreamClosedError:
             # The stream got reset and we haven't worked it out yet. The stream
-            # window doesn't matter now. We still want to increment the
-            # connection window though.
+            # window doesn't matter now as all the data that can possibly be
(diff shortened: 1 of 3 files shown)


twisted/web/test/test_http2.py
```diff
@@ -2079,6 +2079,64 @@ def validate(*args):
...
+        )
+        del frames[-1]
+        frames.append(
+            frameFactory.buildRstStreamFrame(
+                streamID=1, errorCode=h2.errors.INTERNAL_ERROR
+            )
+        )
+
+        requestBytes = frameFactory.clientConnectionPreface()
+        requestBytes += b''.join(f.serialize() for f in frames)
```
adiroiban, line 2107: minor comment. I have seen these 2 lines repeated multiple times. maybe they can be refactored as

```
requestBytes = frameFactory.serializedClientFrames(frames)
```

twisted/web/test/test_http2.py
```diff
@@ -2079,6 +2079,64 @@ def validate(*args):
...
+        frames = framesFromBytes(transport.value())
+
+        # Check that the only WINDOW_UPDATE frames came for the connection.
+        windowUpdateFrames = (
+            f for f in frames
+            if isinstance(f, hyperframe.frame.WindowUpdateFrame)
+        )
+        self.assertTrue(
+            all(frame.stream_id == 0 for frame in windowUpdateFrames)
+        )
```
adiroiban: minor comment

Are we receiving a random number of WINDOW_UPDATE frames, and we receive them in a random order?

Can we use something like this?

```
windowUpdateFramesID = (f.stream_id for f in frames if isinstance(f, hyperframe.frame.WindowUpdateFrame))
self.assertEqual([0], windowUpdateFramesID)
```

assertTrue will give as an `false is not true` message, but assertEqual should give a clue if we are not received the expected number or the expected stream id

twisted/web/_http2.py
```diff
@@ -686,7 +686,18 @@ def openStreamWindow(self, streamID, increment):
         @type increment: L{int}
         """
         # TODO: Consider whether we want some kind of consolidating logic here.
```
adiroiban, line 688: I think that is better to have a ticket for this todo and convert it into a FIXME

In the ticket we can add more information, notes and discuss this issue

twisted/web/_http2.py
```diff
@@ -686,7 +686,18 @@ def openStreamWindow(self, streamID, increment):
...
-        self.conn.increment_flow_control_window(increment, stream_id=streamID)
+        try:
+            self.conn.increment_flow_control_window(
+                increment, stream_id=streamID
+            )
+        except h2.exceptions.StreamClosedError:
+            # The stream got reset and we haven't worked it out yet. The stream
+            # window doesn't matter now. We still want to increment the
+            # connection window though.
```
adiroiban: `We still want to increment the connection window though.`

dumb question ahead. why ? :)

we want or we should ?

is this behaviour discussed in the h2 rfc?

Lukasa: I'll expand this comment.

PR #456, opened 2016-08-03
8722 einarfd py3 conch endpoints

einarfd:
This is a fix for bug https://twistedmatrix.com/trac/ticket/8722, and ports conch.endpoints to Python 3.

I'm not sure about keeping ConsoleUI as a byte based construct is a great idea. But changing it to string, became rather painful when Python 3 refused to open /dev/tty in string mode. It at least work similar on both versions now.


rodrigc:
@einarfd @acabhishek942 I think that large portions of conch now work under Python 3,
in large part to @einarfd 's patches.  There seems to be one item causing problems with using conch as an SSH client with Python 3.  If I try to ssh to localhost using conch with:

```
python -m  twisted.conch.scripts.conch localhost
```

I get:

```
Connection to localhost closed.
conch: exiting with error [Failure instance: Traceback (failure with no frames): <class 'twisted.conch.error.ConchError' at 0x7fbdc6a303d8>: (b'bad host key', 9)
```

Einar, do you have any patches you are working on to fix this problem on Python 3?  There seem to be only a few modules in conch that are not Python 3 clean.

einarfd:
@rodrigc I'm not working on anything on conch at the moment. I'm afraid it will be hard to spend much time on it just now. 
At the top of my list of things to have a look at in conch land would be: 
- Get conch.endpoints._UserAuth to have a getGenericAnswers so it works better with OpenSSH.
- Get conch.endpoints.SSHCommandClientEndpoint to support TCP6 connections in addition to TCP4. 
- Try to get getPassword and getGenericAnswers to both use the same type for their return values. At the moment getPassword returns a bytes, while getGenericAnswers returns a string. 

Our application does not use anything in conch.scripts, so it is unfortunately not a priority for me.

rodrigc:
@einarfd I hope you manage to eventually contribute your changes.  What exactly does your application which uses conch do?

einarfd:
Our app is a test automation library that is used to remotely automate and control video and audio conferencing systems. A lot of them are built by different groups inside Cisco, but some of it is also built by our competitors. These systems use a lot of different protocol to allow automation, HTTP and SSH based protocols are the most common, but there are some of the devices that use other protocols as well. We use Twisted for all the protocols and communication parts of the library, while the rest of the tests are on the other side of a crochet layer that mask the event driven nature of the network communication code.

rodrigc:
Thank you for the explanation of your application.  It's amazing where libraries like Twisted end up being used.
Out of curiousity, what is the force which is motivating you to port conch to Python 3?

I think Python 3 is nice, but in a company, it can seem like porting to Python 3 involves
a lot of work, but other than _code cleanliness_, I don't see any significant gains
which can make justifying the efforts to managers worthwhile.

einarfd:
The force pushing for Python 3 ports are the engineers, and the reason we are pushing for it is that we do not want to be stuck on Python 2.7, don't get me wrong. Python 2.7 is great, but it is never going to get any better. While Python 3 is steadily improving and imo. getting better and better with every release.  
The reason we are allowed to spend some time on getting our code based moved is that our management allow us to spend some time on improving our code base even if it doesn't lead to new features or clear returns on investment.  So since this was one of the things at the top of our wish list, we did get to spend some time on it. 
It also helps that our internal Python training material which you can find at https://github.com/JonJagger/two-day-courses, is for Python 3, and that my site has a history of eagerly keeping up with new versions of the tools and programming languages we use.

rodrigc:
You are very lucky to be in a work environment that permits you to improve the codebase without having to justify it via new features of bug fixes.  That is not always the case in companies!

rodrigc:
@einarfd @acabhishek942 I think I found one reason why the conch command-line client fails on Python 3: https://twistedmatrix.com/trac/ticket/8791
due to the call to _hmacedString() throwing an exception.  If you or Abhishek want to investigate that a bit, that would be great.

rodrigc:
@einarfd I've pushed more Python 3 fixes into conch.  The conch and cftp command-line utilities work on Python 3.  Please provide feedback when you have time.

PR #459, opened 2016-08-04
https://twistedmatrix.com/trac/ticket/3461

glyph:


twisted/web/server.py
```diff
@@ -387,27 +393,74 @@ def view_unregisterProducer(self, issuer):
...
                 except KeyError:
                     pass
             # if it still hasn't been set, fix it up.
-            if not self.session:
-                self.session = self.site.makeSession()
-                self.addCookie(cookiename, self.session.uid, path=b'/')
-        self.session.touch()
+            if not session:
+                session = self.site.makeSession()
+                self.addCookie(cookiename, session.uid, path='/',
```
Lukasa: You lost a `b`-prefixed string here: is that on purpose?

glyph: Oddly enough, this was the one place where the `b`-prefixed string _wasn't_ significant (`addCookie` does encoding leniently).

glyph: I restored it for consistency, but I also fixed the other types, starting with the tests.  Hopefully the python 3 tests will pass now.

glyph:
Re-ran the buildbot because I'm pretty sure that failure is unrelated and intermittent.

PR #461, opened 2016-08-04
[#8155] use bytes not str when working with pipes

evgeni:
see https://twistedmatrix.com/trac/ticket/8155#comment:2

### DIFF
--- a/twisted/topfiles/8155.bugfix
+++ b/twisted/topfiles/8155.bugfix
@@ -1,0 +1,1 @@
+twistd can now properly daemonize on Linux/Unix when run under Python3 and will not hang indefinitely.
--- a/twisted/scripts/_twistd_unix.py
+++ b/twisted/scripts/_twistd_unix.py
@@ -9,7 +9,7 @@ import os
 import sys
 
 from twisted.python import log, logfile, usage
-from twisted.python.compat import intToBytes
+from twisted.python.compat import (intToBytes, nativeString)
 from twisted.python.util import (
     switchUID, uidFromString, gidFromString, untilConcludes)
 from twisted.application import app, service
@@ -342,9 +342,10 @@ class UnixApplicationRunner(app.ApplicationRunner):
         @rtype: C{int}
         """
         data = untilConcludes(os.read, readPipe, 100)
+        datastr = nativeString(data)
         if data != b"0":
             msg = ("An error has occurred: '%s'\nPlease look at log "
-                   "file for more information.\n" % (data[2:],))
+                   "file for more information.\n" % (datastr[2:],))
             untilConcludes(sys.__stderr__.write, msg)
             return 1
         return 0
--- a/twisted/scripts/_twistd_unix.py
+++ b/twisted/scripts/_twistd_unix.py
@@ -218,7 +218,7 @@ class UnixApplicationRunner(app.ApplicationRunner):
         else:
             statusPipe = self.config.get("statusPipe", None)
             if statusPipe is not None:
-                untilConcludes(os.write, statusPipe, "0")
+                untilConcludes(os.write, statusPipe, b"0")
                 untilConcludes(os.close, statusPipe)
         self.startReactor(None, self.oldstdout, self.oldstderr)
         self.removePID(self.config['pidfile'])
@@ -342,7 +342,7 @@ class UnixApplicationRunner(app.ApplicationRunner):
         @rtype: C{int}
         """
         data = untilConcludes(os.read, readPipe, 100)
-        if data != "0":
+        if data != b"0":
             msg = ("An error has occurred: '%s'\nPlease look at log "
                    "file for more information.\n" % (data[2:],))
             untilConcludes(sys.__stderr__.write, msg)
--- a/twisted/test/test_twistd.py
+++ b/twisted/test/test_twistd.py
@@ -1767,7 +1767,7 @@ class DaemonizeTests(unittest.TestCase):
         with non-zero status code.
         """
         self.mockos.child = False
-        self.mockos.readData = "1: An identified error"
+        self.mockos.readData = b"1: An identified error"
         errorIO = NativeStringIO()
         self.patch(sys, '__stderr__', errorIO)
         with AlternateReactor(FakeDaemonizingReactor()):
--- a/twisted/test/test_twistd.py
+++ b/twisted/test/test_twistd.py
@@ -1668,7 +1668,7 @@ class DaemonizeTests(unittest.TestCase):
         self.assertEqual(
             self.mockos.actions,
             [('chdir', '.'), ('umask', 0o077), ('fork', True), 'setsid',
-             ('fork', True), ('write', -2, '0'), ('unlink', 'twistd.pid')])
+             ('fork', True), ('write', -2, b'0'), ('unlink', 'twistd.pid')])
         self.assertEqual(self.mockos.closed, [-3, -2])
 
 
@@ -1678,7 +1678,7 @@ class DaemonizeTests(unittest.TestCase):
         status pipe and then exit the process.
         """
         self.mockos.child = False
-        self.mockos.readData = "0"
+        self.mockos.readData = b"0"
         with AlternateReactor(FakeDaemonizingReactor()):
             self.assertRaises(SystemError, self.runner.postApplication)
         self.assertEqual(
@@ -1708,7 +1708,7 @@ class DaemonizeTests(unittest.TestCase):
             [('chdir', '.'), ('umask', 0o077), ('fork', True), 'setsid',
              ('fork', True), ('unlink', 'twistd.pid')])
         self.assertEqual(self.mockos.closed, [-3, -2])
-        self.assertEqual([(-2, '0'), (-2, '0')], written)
+        self.assertEqual([(-2, b'0'), (-2, b'0')], written)
 
 
     def test_successInParentEINTR(self):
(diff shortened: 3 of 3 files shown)


twisted/topfiles/8155.bugfix
```diff
@@ -0,0 +1 @@
+_twistd_unix.py used "0" when working with pipes, but in Python3 you need b"0" to get a bytes object
```
adiroiban: The text from here is provided as a help to the release manager for generating the release notes.

It should not be written as a commit message or as the technical fix resolution. 

Twisted users should not care about twistedscripts._twistd_unix as this is a private implementation details.

How is this issue observed to end users?
How can end use know if this change will fix one of their issues?

For example: twistd can now perform operation X when used on Linux/Unix under Python3.

Thanks!

evgeni:
@adiroiban thanks, fixed the topfile

markrwilliams:
@evgeni: this looks really good.  Looks like you might have missed something, though: [I think this should also be bytes](https://github.com/evgeni/twisted/blob/e3a43cc93eefebe26619dcef13e821d68b6840fc/twisted/test/test_twistd.py#L1755).

evgeni:
@markrwilliams: well, I could argue that the tests pass :D but yeah, lemme look into that.

evgeni:
@markrwilliams so yeah, looking at the tests, I agree that this should be a `bytes` too. however, changing this will fail the tests as if you `%s` a `bytes` in a Python3, you end up with a literal `b'foo'`:

``` python
>>> '%s' % (b'foo')
"b'foo'"
```

To fix this we'd have to amend the generation of the error message in (at least) `_waitForStart`:

``` patch
diff --git a/twisted/scripts/_twistd_unix.py b/twisted/scripts/_twistd_unix.py
index 178b633..29096c3 100644
--- a/twisted/scripts/_twistd_unix.py
+++ b/twisted/scripts/_twistd_unix.py
@@ -9,7 +9,7 @@ import os
 import sys

 from twisted.python import log, logfile, usage
-from twisted.python.compat import intToBytes
+from twisted.python.compat import (intToBytes, _PY3)
 from twisted.python.util import (
     switchUID, uidFromString, gidFromString, untilConcludes)
 from twisted.application import app, service
@@ -342,9 +342,13 @@ class UnixApplicationRunner(app.ApplicationRunner):
         @rtype: C{int}
         """
         data = untilConcludes(os.read, readPipe, 100)
+        if _PY3:
+            datastr = data.decode('ascii')
+        else:
+            datastr = data
         if data != b"0":
             msg = ("An error has occurred: '%s'\nPlease look at log "
-                   "file for more information.\n" % (data[2:],))
+                   "file for more information.\n" % (datastr[2:],))
             untilConcludes(sys.__stderr__.write, msg)
             return 1
         return 0
```

I don't really like that :/
ideas?

edit: we could also wrap `data` with `twisted.python.compat.nativeString` instead?

adiroiban:
why not always decode the data?

if we are decoding it anyway in python3 we should also decode it in python2... to get a similar behaviour

---

but instead of `%s` I think that we should have `%r` as we might get raw data and then an ascii decoding might fail

We can use a compat helper to return the future representation of bytes.

I remember that we hit the `repr(b'data')` in other places so I think that having support for this in compat might help

In twisted.python.compat we can have

```
def future_bytes_repr(data):
     if _PY3:
        return repr(data)
    return 'b' + repr(data)
```

we have compat to avoid using `if _PY3` as much as possible in the other code.

evgeni:
`data.decode('ascii')` will fail on Python2:

```
  File "/home/evgeni/Devel/twisted/twisted/test/test_twistd.py", line 1774, in test_errorInParent
    self.assertRaises(SystemError, self.runner.postApplication)
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 414, in assertRaises
    return context._handle(lambda: f(*args, **kwargs))
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 308, in _handle
    self._returnValue = obj()
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 346, in __exit__
    self._expectedName, reason.getTraceback()),
twisted.trial.unittest.FailTest: exceptions.TypeError raised instead of SystemError:
  File "/home/evgeni/Devel/twisted/twisted/internet/utils.py", line 197, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/home/evgeni/Devel/twisted/twisted/test/test_twistd.py", line 1774, in test_errorInParent
    self.assertRaises(SystemError, self.runner.postApplication)
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 414, in assertRaises
    return context._handle(lambda: f(*args, **kwargs))
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 308, in _handle
    self._returnValue = obj()
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 308, in _handle
    self._returnValue = obj()
  File "/home/evgeni/Devel/twisted/twisted/trial/_synctest.py", line 414, in <lambda>
    return context._handle(lambda: f(*args, **kwargs))
  File "/home/evgeni/Devel/twisted/twisted/scripts/_twistd_unix.py", line 208, in postApplication
    self.startApplication(self.application)
  File "/home/evgeni/Devel/twisted/twisted/scripts/_twistd_unix.py", line 394, in startApplication
    self.config['pidfile'])
  File "/home/evgeni/Devel/twisted/twisted/scripts/_twistd_unix.py", line 283, in setupEnvironment
    self.config["statusPipe"] = self.daemonize(reactor)
  File "/home/evgeni/Devel/twisted/twisted/scripts/_twistd_unix.py", line 313, in daemonize
    code = self._waitForStart(r)
  File "/home/evgeni/Devel/twisted/twisted/scripts/_twistd_unix.py", line 349, in _waitForStart
    untilConcludes(sys.__stderr__.write, msg)
  File "/home/evgeni/Devel/twisted/twisted/python/util.py", line 715, in untilConcludes
    return f(*a, **kw)
exceptions.TypeError: 'unicode' does not have the buffer interface
```

But, anyways, `nativeString` DTRT already as it will always do ASCII.

## 

WRT `future_bytes_repr` and using `%r`: I agree that this _should_ be done, but that will be out of scope for the initial fix, IMHO.

adiroiban:
> exceptions.TypeError: 'unicode' does not have the buffer interface

hmmm.... so for this test in python2 we have Unicode  but in Python3 we have bytes ? 

do you get the same behaviour outside of the test?

it looks like the mocking done in this test is wrong as I am expecting that on python2 `os.read()` will return a file object which returns str and not Unicode

---

WTR future_bytes_repr we can leave for later, but if this helps fix this issue than I think that doing this change is not out of scope ... as `%r` should be the right way of serializing bytes :)

markrwilliams:
> > exceptions.TypeError: 'unicode' does not have the buffer interface
> > hmmm.... so for this test in python2 we have Unicode but in Python3 we have bytes ?
> 
> do you get the same behaviour outside of the test?
> 
> it looks like the mocking done in this test is wrong as I am expecting that on python2 os.read() will return a file object which returns str and not Unicode

The mocking's categorically correct but more strict than the actual implementation.  The issue isn't with any call to `os.read`, but rather the call to `sys.__stderr__.write`.

The traceback goes through [`untilConcludes(sys.__stderr__.write, msg)`](https://github.com/evgeni/twisted/blob/d56805956fbf5ed32fd348c8394186c7c10a0526/twisted/scripts/_twistd_unix.py#L349).  `msg` has become a `unicode` object because `datastr` is `unicode` per the `.decode('ascii')` call, and on Python 2, `<str> % (<unicode>,)` -> `<unicode>`.

Normally `sys.__stderr__` is just a regular Python 2 `file` object, so `.write` will implicitly encode `unicode` with the system default encoding (`ascii` if all is Right.)  But in the [failing test](https://github.com/evgeni/twisted/blob/d56805956fbf5ed32fd348c8394186c7c10a0526/twisted/test/test_twistd.py#L1763-L1783), [it's patched with a `NativeStringIO`](https://github.com/evgeni/twisted/blob/d56805956fbf5ed32fd348c8394186c7c10a0526/twisted/test/test_twistd.py#L1772), which on Python 2 will be a `BytesIO` object.

Now `BytesIO` objects attempt to grab the backing buffer of the byte string being written, so [writing a unicode string to one will fail with just the error we see here.](https://eval.in/619085)

I think the test should be this strict.  I also think that using `nativeString` the way @evgeni has in d568059 is good for this PR.  I expect _generally_ that whatever's written to stderr will be a traceback into Python source that can be decoded as `ascii`.  That means merging this will make `twistd` _mostly_ work on Python 3.

But @adiroiban is also right that whatever appears in `stderr` doesn't necessarily have to be ASCII.  I've created a [new ticket (8731)](https://twistedmatrix.com/trac/ticket/8731) for the `repr` utility.  Once that's complete I'll be happy to change this code to use it.

But for now I'd like to merge this PR.

adiroiban:
@markrwilliams if you are happy with this branch, feel free to update the Trac ticket and merge it.

non-ascii stderr output is already crashing on trunk so this is not a reason to block this branch. 

I was not doing a review, but just trying to help in moving this forward.

---

while the tests should be as strict as possible, the mocks should also be as close as possible to the production code as otherwise we are just fooling ourselves by running tests using code which is never used in production

Thanks!

markrwilliams:
@evgeni thanks for making Twisted better!

PR #463, opened 2016-08-05
Fix building of C extensions, especially visible on Windows

rodrigc:
https://twistedmatrix.com/trac/ticket/8724


appveyor.yml
```diff
@@ -25,33 +25,63 @@ cache:
...
   # Only use pre-installed Python versions, which Appveyor may upgrade to
   # a later point release.
   # See: http://www.appveyor.com/docs/installed-software#python
   matrix:
-    - PYTHON: C:\Python27-x64
+    - PYTHON_HOME: C:\\Python27-x64
+      PYTHON_VERSION: "2.7"
+      PYTHON_ARCH: "64"
       TOXENV: py27-coverage-windows,codecov-publish
+    #- PYTHON_HOME: "C:\\PYTHON35-x64"
```
glyph, line 43: It looks like these commented-out lines should probably be added, but… in a separate PR?

appveyor.yml
```diff
@@ -25,33 +25,63 @@ cache:
...
+    #  PYTHON_VERSION: "3.5"
+    #  PYTHON_ARCH: "64"
+    #  TOXENV: 'py35-tests-windows'
 
 
 install:
+  # Install OpenSSH through chocolatey
```
glyph, line 50: This is for conch's integration tests, right?

bin/admin/run_with_env.cmd
```diff
@@ -0,0 +1,88 @@
...
+::
+:: Note: this script needs to be run with the /E:ON and /V:ON flags for the
+:: cmd interpreter, at least for (SDK v7.0)
+::
+:: More details at:
+:: https://github.com/cython/cython/wiki/64BitCythonExtensionsOnWindows
+:: http://stackoverflow.com/a/13751649/163740
+::
+:: Author: Olivier Grisel
+:: License: CC0 1.0 Universal: http://creativecommons.org/publicdomain/zero/1.0/
```
glyph, line 20: I think that this isn't a problem since it's upward-compatible with MIT, but let me check with a lawyer; I know "public domain" has been controversial in the past.

bin/admin/run_with_env.cmd
```diff
@@ -0,0 +1,88 @@
...
+:: MS Windows SDK for Windows 7 and .NET Framework 3.5 (SDK v7.0)
+::
+:: 32 bit builds, and 64-bit builds for 3.5 and beyond, do not require specific
+:: environment configurations.
+::
+:: Note: this script needs to be run with the /E:ON and /V:ON flags for the
+:: cmd interpreter, at least for (SDK v7.0)
+::
+:: More details at:
+:: https://github.com/cython/cython/wiki/64BitCythonExtensionsOnWindows
```
glyph, line 16: This link is broken, and I would very much like to understand where this file came from and why it needs to exist ;)

appveyor.yml
rodrigc, line 50: Yes for conch.  @hawkowl added it here: https://github.com/twisted/twisted/commit/dfeaf6cad87778193a7e37e120ba92caa18ee5f5  I just moved it around.

appveyor.yml
rodrigc, line 43: I wanted to leave those lines in there, but commented out, so
that I can work on a branch with those lines uncommented and gradually get a working Windows Python 3 appveyor build, and also to allow other folks to help the Windows Python 3 effort by uncommenting those lines in their private branches.

Right now I have been chasing all the appveyor changes in a branch in a separate repository and it has been painful.

glyph:
I'm a little concerned with the giant batch file, but modulo the copyright concerns I think this can land once the build is done.

We should really switch away from the `--text` builder on appveyor though - I can't see what tests _are_ being run, only which ones aren't.

bin/admin/run_with_env.cmd
glyph, line 20: And lo, a lawyer did descend, a crackle of lightning heralding his coming; as the clouds parted, a thunderous voice could be heard to say:

> Adding cc0 licensed stuff to a project is never a problem.

appveyor.yml
rodrigc, line 50: Actually, I think anything that uses pyOpenSSL needs this.  There are a few places in the tree outside conch which use pyOpenSSL.

bin/admin/run_with_env.cmd
rodrigc, line 20: I didn't go to law school, but that sounds good to me!

rodrigc:
Reducing the size of the test output was done because
Travis has a [4MB](https://github.com/travis-ci/travis-ci/issues/1382) size restriction on the amount of stuff that can be written to stdout/stderr during a build job.
Not sure if appveyor has a similar restriction, but if it does not, we should display more stuff for sure.

bin/admin/run_with_env.cmd
rodrigc, line 16: According to the version history ( https://github.com/cython/cython/wiki/CythonExtensionsOnWindows/_history ) this page got renamed to https://github.com/cython/cython/wiki/CythonExtensionsOnWindows

appveyor.yml
Lukasa, line 50: Why would PyOpenSSL need OpenSSH?

glyph, line 50: If it's because OpenSSH includes a libssl: pyOpenSSL should be grabbing Cryptography's Windows wheel; we should NOT be building our own cryptography binaries in CI.

hawkowl, line 50: This is for conch tests that run a server and call `ssh`....

hawkowl, line 50: (as in it calls the actual command line utility `ssh`, not any libssl or anything)

PR #464, opened 2016-08-05
[#8726] Port twisted.protocols.memcache to Python 3

charmander:
https://twistedmatrix.com/trac/ticket/8726

Includes some questionable `.decode("ascii")`s. Is them raising an exception something to worry about?

### DIFF
--- a/twisted/protocols/memcache.py
+++ b/twisted/protocols/memcache.py
@@ -25,12 +25,15 @@ See U{http://code.sixapart.com/svn/memcached/trunk/server/doc/protocol.txt} for
 more information about the protocol.
 """
 
+from __future__ import absolute_import, division
+
 from collections import deque
 
 from twisted.protocols.basic import LineReceiver
 from twisted.protocols.policies import TimeoutMixin
 from twisted.internet.defer import Deferred, fail, TimeoutError
 from twisted.python import log
+from twisted.python.compat import iteritems
 
 
 
@@ -69,7 +72,7 @@ class Command(object):
     @type _deferred: L{Deferred}
 
     @ivar command: name of the command sent to the server.
-    @type command: C{str}
+    @type command: C{bytes}
     """
 
     def __init__(self, command, **kwargs):
@@ -77,7 +80,7 @@ class Command(object):
         Create a command.
 
         @param command: the name of the command.
-        @type command: C{str}
+        @type command: C{bytes}
 
         @param kwargs: this values will be stored as attributes of the object
             for future use
@@ -189,7 +192,7 @@ class MemCacheProtocol(LineReceiver, TimeoutMixin):
         self._getBuffer.append(data)
         self._bufferLength += len(data)
         if self._bufferLength >= self._lenExpected + 2:
-            data = "".join(self._getBuffer)
+            data = b"".join(self._getBuffer)
             buf = data[:self._lenExpected]
             rem = data[self._lenExpected + 2:]
             val = buf
@@ -225,19 +228,18 @@ class MemCacheProtocol(LineReceiver, TimeoutMixin):
         This the end token to a get or a stat operation.
         """
         cmd = self._current.popleft()
-        if cmd.command == "get":
+        if cmd.command == b"get":
             if cmd.multiple:
-                values = dict([(key, val[::2]) for key, val in
-                               cmd.values.iteritems()])
+                values = {key: val[::2] for key, val in iteritems(cmd.values)}
                 cmd.success(values)
             else:
                 cmd.success((cmd.flags, cmd.value))
-        elif cmd.command == "gets":
+        elif cmd.command == b"gets":
             if cmd.multiple:
                 cmd.success(cmd.values)
             else:
                 cmd.success((cmd.flags, cmd.cas, cmd.value))
-        elif cmd.command == "stats":
+        elif cmd.command == b"stats":
             cmd.success(cmd.values)
 
 
@@ -253,9 +255,9 @@ class MemCacheProtocol(LineReceiver, TimeoutMixin):
         Prepare the reading a value after a get.
         """
         cmd = self._current[0]
-        if cmd.command == "get":
+        if cmd.command == b"get":
             key, flags, length = line.split()
-            cas = ""
+            cas = b""
         else:
             key, flags, length, cas = line.split()
         self._lenExpected = int(length)
@@ -279,7 +281,7 @@ class MemCacheProtocol(LineReceiver, TimeoutMixin):
         Reception of one stat line.
         """
         cmd = self._current[0]
-        key, val = line.split(" ", 1)
+        key, val = line.split(b" ", 1)
         cmd.values[key] = val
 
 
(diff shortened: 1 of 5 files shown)


twisted/protocols/memcache.py
```diff
@@ -622,7 +627,7 @@ def getMultiple(self, keys, withIdentifier=False):
         issuing C{checkAndSet} update later, using the corresponding method.
 
         @param keys: The keys to retrieve.
-        @type keys: C{list} of C{str}
```
rodrigc, line 625: Since you are taking a pass through this file can you change C{int}, C{str}, C{list}, C{bytes} to L{int}, L{str}, L{list}, L{bytes}?

According to the coding standard, basic Python types need to be wrapped in L{} so that they link back to the Python documentation:

https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

twisted/protocols/memcache.py
```diff
@@ -25,12 +25,15 @@ def doSomething(proto):
...
 
+from __future__ import absolute_import, division
+
 from collections import deque
 
 from twisted.protocols.basic import LineReceiver
 from twisted.protocols.policies import TimeoutMixin
 from twisted.internet.defer import Deferred, fail, TimeoutError
 from twisted.python import log
+from twisted.python.compat import iteritems
```
rodrigc: Can you import nativeString and networkString from twisted.python.compat?
In this file, for all changes where you did  foo.encode("ascii"), replace with networkString(foo),
and for all cases of bar.decode("ascii"), replace with nativeString(bar).

This will keep things compatible on Python 2 and Python 3

twisted/protocols/memcache.py
```diff
@@ -343,19 +347,19 @@ def lineReceived(self, line):
         Receive line commands from the server.
         """
         self.resetTimeout()
-        token = line.split(" ", 1)[0]
+        token = line.split(b" ", 1)[0]
         # First manage standard commands without space
-        cmd = getattr(self, "cmd_%s" % (token,), None)
+        cmd = getattr(self, u"cmd_" + token.decode("ascii"), None)
```
rodrigc: this is better as cmd = getattr(self, "cmd_" + nativeString(token), None)
No need to convert to unicode here, since this is a behavior change on Python 2.

rodrigc:
Thanks for the updates.  This is almost good to go in.
If I look at this coverage report, on line 242:
https://codecov.io/gh/twisted/twisted/compare/3e488bca70159bd8d7f5a62dc55e0ab9879f229b...5288541dce2e97b5b323ccdc0267bbe52050aaa5?src=pr#747769737465642F70726F746F636F6C732F6D656D63616368652E7079-242

It looks like the existing tests don't cover the **b'stats'** value.  Can you extend the existing tests by an additional test that tests for this value?

charmander:
@rodrigc 554fdf86df6a4ce2060c6309183d30a4c7a77b24 addressed that, but it looks like Codecov didn’t pick it up for some reason. (There are only two checks listed instead of the earlier four.) Should I merge trunk to try to trigger it?

twisted/protocols/memcache.py
```diff
@@ -299,20 +305,22 @@ def cmd_ERROR(self):
         cmd.fail(NoSuchCommand())
 
 
-    def cmd_CLIENT_ERROR(self, errText):
+    def cmd_CLIENT_ERROR(self, errBytes):
```
hawkowl: I believe this breaks our backwards compatibility policy, because it's renaming parts of our public API :(

It does make it clearer, though, so I believe that maybe something like:

```
def somefunc(self, errBytes=None, errText=None):
    if errText:
        raise DeprecationWarning("Passing 'errText' as a kwarg to somefunc is deprecated since Twisted 16.4.0")
    if errText and errBytes is None:
        errBytes = errText
    elif errText and errBytes:
        raise ValueError("Please specify only only one argument (errBytes).")
```

or something...

cc @rodrigc , @glyph

charmander: I didn’t notice it was public – whoops! In that case, the bytes still have to represent text, so maybe it would be best to just keep the name as-is.

adiroiban:
Changes looks good. Thanks!

Please include twisted/test/test_memcache.py to dist3.py so that we will run the test on python3

decode('ascii') on random input can be a problem as if we receive non 7bit ascii bytes decode will break... is better to just user repr() 

---

codecov.io might be slow to report the coverage

rodrigc:
@charmander Your build is failing on Python 3.3 and Python 3.4, like here: https://travis-ci.org/twisted/twisted/jobs/150782318#L2477

Python 3.3 and 3.4 did not support string formatting of bytes.  That only came into Python 3.5:
https://www.python.org/dev/peps/pep-0461/

Can you change your code so that it works on Python 3.3 and 3.4?  Twisted needs to work on those versions also.

charmander:
Ah, I took that into account for `twisted.protocol.memcache` but missed it in the tests. Sorry!

twisted/protocols/memcache.py
```diff
@@ -411,12 +419,12 @@ def _incrdecr(self, cmd, key, val):
...
             return fail(RuntimeError("not connected"))
-        if not isinstance(key, str):
+        if not isinstance(key, bytes):
             return fail(ClientError(
-                "Invalid type for key: %s, expecting a string" % (type(key),)))
+                "Invalid type for key: %s, expecting bytes" % (type(key),)))
         if len(key) > self.MAX_KEY_LENGTH:
             return fail(ClientError("Key too long"))
-        fullcmd = "%s %s %d" % (cmd, key, int(val))
+        fullcmd = b" ".join([cmd, key, networkString(str(int(val)))])
```
hawkowl: `twisted.python.compat.intToBytes` would be nicer here.

charmander: It would. Thanks!

hawkowl:
Thanks for this @charmander .

re: `encode('ascii')` (and by extension `networkString`, which does that internally), that is kind of the default assumption of Python 2/3 if you pass it Unicode, but I think random 8-bit trash is actually allowed: https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L495

I am not sure what this really means for this branch, I think we ought to allow binary trash in keys, since it used to work (and this branch might cause working Python 2 apps to stop working). So, I'll have a look through and point out some places where we might need to fix it.

twisted/protocols/memcache.py
```diff
@@ -303,7 +309,8 @@ def cmd_CLIENT_ERROR(self, errText):
         """
         An invalid input as been sent.
         """
-        log.err("Invalid input: %s" % (errText,))
+        errText = nativeString(errText)
```
hawkowl: This should `repr()` instead -- the err will be `b''` prefixed on Python 3, but that's fine, as it means binary garbage coming in won't cause this to fail.

twisted/protocols/memcache.py
```diff
@@ -312,7 +319,8 @@ def cmd_SERVER_ERROR(self, errText):
         """
         An error has happened server-side.
         """
-        log.err("Server error: %s" % (errText,))
+        errText = nativeString(errText)
```
hawkowl: Likewise with the repr comment.

twisted/test/test_memcache.py
```diff
@@ -5,6 +5,8 @@
 Test the memcache client protocol.
 """
```
hawkowl, line 7: This needs a `__future__` import, like the implementation function.

twisted/test/test_memcache.py
```diff
@@ -278,11 +281,11 @@ def test_invalidGetResponse(self):
         If the value returned doesn't match the expected key of the current
         C{get} command, an error is raised in L{MemCacheProtocol.dataReceived}.
         """
-        self.proto.get("foo")
+        self.proto.get(b"foo")
         s = "spamegg"
         self.assertRaises(
             RuntimeError, self.proto.dataReceived,
-            "VALUE bar 0 %s\r\n%s\r\nEND\r\n" % (len(s), s))
+            networkString("VALUE bar 0 %d\r\n%s\r\nEND\r\n" % (len(s), s)))
```
hawkowl: I would much prefer this and the next test to use `intToBytes` vs `networkString`.

hawkowl: Actually, other tests hardcode the length. I'd be fine with that too, then it's consistent.

charmander:
This shouldn’t attempt to decode keys anywhere… unless they’re included raw in some error messages. That doesn’t seem to be the case judging by a quick search of memcached, but I don’t know about past versions.

hawkowl:
So I've read through and am confident that no encoding/decoding is done on keys, so that shouldn't be a worry. I raised one or two places where encoding problems may occur,  depending on what the server sends (although it shouldn't send back binary trash, but I'd prefer to be cautious here)...

@charmander , thanks again! After those little things, I think we are good to land!

hawkowl:
I'm gonna merge this. Thanks again @charmander !

PR #465, opened 2016-08-05
ResumeProducing after state has changed.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8720


twisted/test/proto_helpers.py
```diff
@@ -142,6 +142,13 @@ class StringTransport:
     @ivar io: A L{io.BytesIO} which holds the data which has been written to
         this transport since the last call to L{clear}.  Use L{value} instead
         of accessing this directly.
+
+    @ivar lenient: A L{bool} that controls whether L{StringTransport} will
```
adiroiban: I know that @glyph has suggested this flag, but I think that we can do better by creating 2 string transports like StateEnforcingStringTransport and RealWorldStringTransport ... note that the names should be improved, but I hope that you get the idea.

In the class docstring we can also include all the comment from the Trac ticket with information about when to use one class and when to use another class.

I find the current docstring confusing as it does not explain why on earth you would like to run your tests with a mock which is not modeled after the real deal :)

adiroiban:
hm... I can not do inline comments for empty files.

The ticket is a defect, but the topfile associated with this branch is a misc.

Either the ticket needs to be updated or this needs to have a .bugfix explaining what problem was fixed 

---

In the Trac ticket there are discussion of all kind of corner cases which are limiting these changes.
Do we have automated tests for each of the described cases? For example, data produced on resume or resume will call _bodyDataFinished ?

Many thanks!

twisted/test/proto_helpers.py
glyph: For what it's worth: I'm OK with the flag but I'm more or less neutral on having multiple different classes.  -1 for using inheritance though ;)

glyph: @adiroiban - would you find it less confusing if the flag were called `extraStrict` and defaulted to `True`?

adiroiban: My biggest problem is with the lack documentation for this flag and the StringTransport.

As a normal dev, which has not read the ticket discussions, I will read this code and will wonder why do we not want to be `extraStrict` :)

So we can use the flag and a single class, but would help to also update the class docstring to include all the implementation details which were revealed in the Trac ticket and why the default implementation is more or less strict than the real world transports.

We already have StringTransportWithDisconnection which is using inheritance ... and by not using a flag it has the same signature as the real deal :)

I don't want to delay merging this code so I am OK with any fix.

I just wanted to leave my comments about how I would think that a 3rd party developer (not involved in this ticket) might perceive these changes

glyph:
@Lukasa Sorry :-( Can you address the conflicts here?

twisted/test/proto_helpers.py
```diff
@@ -154,13 +161,14 @@ class StringTransport:
...
 
-    def __init__(self, hostAddress=None, peerAddress=None):
+    def __init__(self, hostAddress=None, peerAddress=None, lenient=False):
         self.clear()
         if hostAddress is not None:
             self.hostAddr = hostAddress
         if peerAddress is not None:
             self.peerAddr = peerAddress
         self.connected = True
+        self.lenient = lenient
```
glyph: This attribute should be private, as per https://glyph.twistedmatrix.com/2015/05/separate-your-fakes-and-your-inspectors.html - we are publicly exposing some things that aren't part of "real" transports here, but we shouldn't be adding more errors of that variety :).

twisted/test/proto_helpers.py
glyph: OK.  It might be perhaps useful to include a sentence like "This enforces `resumeProducing` not be called after the connection is lost by default in order to ensure that any code which does so is not blindly expecting the `resumeProducing` call to have an impact, even on a closed connection; however, it may be safely set to `False` if you know that this is on purpose, and your code won't hang expecting further traffic or anything like that."  OK maybe that's a little long winded but you get the idea.

src/twisted/test/proto_helpers.py
```diff
@@ -143,11 +143,17 @@ class StringTransport:
         this transport since the last call to L{clear}.  Use L{value} instead
         of accessing this directly.
 
-    @ivar lenient: A L{bool} that controls whether L{StringTransport} will
-        throw errors if production is resumed after the connection has been
-        lost. By default L{StringTransport} enforces that C{resumeProducing}
-        cannot be called after the connection is lost, but real transports
-        don't do that. Defaults to C{False}.
+    @ivar lenient: By default L{StringTransport} enforces that
```
adiroiban: don't forget to also update the name here to `_lenient`

Lukasa:
So the only thing busting this build is @hawkowl's 3.3 problem.

glyph:
It doesn't seem like just @hawkowl's problem :)

glyph:
We really need to develop some discipline about pinning test dependencies (cough, https://github.com/rackerlabs/mimic/, cough) so that we don't have PRs getting stuck like this :-/

glyph:
@mithrandi indicates that the yak stack ends at https://github.com/pypa/setuptools/issues/776

glyph:
@Lukasa Would you mind pinning the setuptools version for the 3.3 tox env, to see if that would help?

glyph:
I am happy with @Lukasa's changes, and @dstufft has eyeballed mine just to make sure they're not too bad (there is a race condition when someone releases a new version of cryptography in the middle of a build, but hopefully before we care about this the bug will be fixed in setuptools and we will be doing dependency pinning like mimic does).

glyph:
ALL. 30. SUCCESSFUL. CHECKS.

PR #467, opened 2016-08-06
[#8622] Clean setup.py code and improve coverage.

adiroiban:
See: https://twistedmatrix.com/trac/ticket/8622


rodrigc:
This patch will break the twisted-depgraph utility mentioned here: https://twistedmatrix.com/trac/wiki/Plan/Python3

There are some OK cleanups in this patch, but overall this patch seems like a lot
of churn that doesn't accomplish anything.

adiroiban:
Thanks @rodrigc for the feedback. Much appreciated! I have reverted the dist3 changes.

The goal of this branch is to set things right for the code in setup.py and dist.py .

A comment is a poor way of warning about a private API... so t.python._setup should follow the Twisted convention for private API

dist.py was created to host the code from setup.py so that we can test it, but we ended still having code in setup.py  and missing coverage.

rodrigc:
I see very little value to this patch, and it seems like a lot of unnecessary churn.

Better coverage of setup.py can be achieved by running setup.py under the coverage command
in tox.ini

twisted/python/_setup.py
```diff
@@ -135,133 +157,96 @@ def __init__(self, *args, **kwargs):
...
+    ]
 
-        class my_build_ext(build_ext_twisted):
-            conditionalExtensions = extensions
-        kw.setdefault('cmdclass', {})['build_ext'] = my_build_ext
-    return kw
 
 
-def getVersion(base):
+def get_setup_args():
```
hawkowl: I would like this to be `def getSetupArgs(extensions=_EXTENSIONS):` to fit the coding standard, and mean you can pass `_EXTENSIONS` in explicitly without monkeypatching.

adiroiban: True. Thanks! Will update.

twisted/python/test/test_setup.py
```diff
@@ -7,36 +7,39 @@
...
+        Will return the arguments with a custom build_ext which knows how to
+        check whether they should be built.
         """
         good_ext = ConditionalExtension("whatever", ["whatever.c"],
                                         condition=lambda b: True)
         bad_ext = ConditionalExtension("whatever", ["whatever.c"],
                                         condition=lambda b: False)
-        args = get_setup_args(conditionalExtensions=[good_ext, bad_ext])
+        self.patch(_setup, '_EXTENSIONS', [good_ext, bad_ext])
+        args = get_setup_args()
```
hawkowl: Per https://github.com/twisted/twisted/pull/467#discussion_r74916242, this would not have a `patch` call...

twisted/python/test/test_setup.py
```diff
@@ -49,11 +52,13 @@ def test_conditionalExtensions(self):
...
     def test_win32Definition(self):
         """
-        When building on Windows NT, the WIN32 macro will be defined as 1.
+        When building on Windows NT, the WIN32 macro will be defined as 1 on
+        the extensions.
         """
         ext = ConditionalExtension("whatever", ["whatever.c"],
                                    define_macros=[("whatever", 2)])
-        args = get_setup_args(conditionalExtensions=[ext])
+        self.patch(_setup, '_EXTENSIONS', [ext])
```
hawkowl: Per https://github.com/twisted/twisted/pull/467#discussion_r74916242, this would not have a `patch` call...

PR #468, opened 2016-08-07
8701 acabhishek942 keyfingerprint

acabhishek942:
see http://twistedmatrix.com/trac/ticket/8701


twisted/conch/ssh/keys.py
```diff
@@ -782,18 +783,23 @@ def public(self):
         return Key(self._keyObject.public_key())
 
 
-    def fingerprint(self):
+    def fingerprint(self, hash_function=sha256):
```
adiroiban: the signature should be 
`def fingerprint(self, hashFunction=md5):
or 
`def fingerprint(self, digest=md5):`

we should not change the behaviour of exiting code using `fingerprint()` and hash_function does not comply with Twisted coding standard

rodrigc: I disagree with @adiroiban here.  The standard response to a Twisted review is to not change
behavior of existing code and I normally agree with it.  However, in this case I think someone who is more familiar with security
should look at this.
Changing the default behavior of the fingerprint function to sha256 is a reasonable
change as long as it is documented and in the release notes.

adiroiban: I am OK with changing the default behaviour as Key.fingerprint() is not really used inside Twisted (only in some scripts)... but I think that it should be done with a deprecation warning.

Feel free to communicate over the mailinglist and see if an exception is granted.

---

in terms of md5 collisions I am not aware of exploiting the md5 sum for impersonating private SSH keys in a useful way

http://natmchugh.blogspot.com.es/2015/09/md5-collisions-in-ssh-keys.html

http://crypto.stackexchange.com/questions/8758/how-hard-is-it-to-generate-a-partial-rsa-fingerprint-collision

---

I agree that we should move await from md5  but since this is not a real security issue we might want to do the standard deprecation dance

---

the conch.ssh.transport.SSHClientTransport  is not using this code so if we want to fix the security in conch I think that we should also address this code... maybe by passing a twisted.consh.ssh.keys.Key instance to verifyHostKey instead of the raw key + md5 fingerprint

right now we have this code

```
    def _ssh_KEXDH_REPLY(self, packet):
        """
        Called to handle a reply to a non-group key exchange message
        (KEXDH_INIT).
      """
        pubKey, packet = getNS(packet)
        f, packet = getMP(packet)
        signature, packet = getNS(packet)
        fingerprint = b':'.join([binascii.hexlify(ch) for ch in
                                 iterbytes(md5(pubKey).digest())])
        d = self.verifyHostKey(pubKey, fingerprint)
```

Lukasa: So I'd like to tag in @reaperhulk or @alex to weigh in on the security risks more formally than me, but here is my limited understanding of the risk here.

The risk only comes in one case: if the fingerprint is used for visual verification by a _user_ that the server is the one it is expecting. In this case, an MD5 collision can in principle be used to spoof a server's identity. This is a good reason to change, but it's not clear that the attack surface is huge, especially given that in my experience almost no users actually do this validation anyway. @adiroiban's suggestion that this method is unused within conch itself strongly suggests that there is minimal risk of attack here.

With that said, moving to SHA256 is a really good idea. However, I don't think it's an idea with security ramifications so profound that we need to skip the standard Twisted deprecation process. The security experts I just tagged may disagree with me though.

twisted/conch/ssh/keys.py
```diff
@@ -782,18 +783,23 @@ def public(self):
...
-            octets printed as hexadecimal with lowercase letters and separated
-            by colons.
+            The fingerprint of a public key consists of the output of the
+            message-digest algorithm for the given hash function(sha256 or MD5).
+            The input to the algorithm is the public key data as specified by [RFC4253].
+            The output of sha256 (default)[RFC4634] algorithm is presented to the
+            user in the form of base64 encoded sha256 hashes.
+            The output of the MD5[RFC1321] algorithm is presented to the user as
+            a sequence of 16 octets printed as hexadecimal with lowercase letters
+            and separated by colons.
```
Lukasa: You can't change this text like this. The paragraph above notes that this is a quote from RFC 4716, which you have now edited.

You can either remove the quote entirely, or you can add new text below this quote, but you can't change it.

twisted/conch/ssh/keys.py
```diff
@@ -802,9 +808,13 @@ def fingerprint(self):
...
-        return nativeString(
-            b':'.join([binascii.hexlify(x)
-                       for x in iterbytes(md5(self.blob()).digest())]))
+        fingerprint_digest=hash_function(self.blob()).digest()
+        if hash_function==sha256:
+            return nativeString(base64.b64encode(fingerprint_digest))
+        else:
+            return nativeString(
+                b':'.join([binascii.hexlify(x)
+                           for x in iterbytes(fingerprint_digest)]))
```
Lukasa: I have two notes here. Firstly, we're missing coverage of the old MD5 branch: we'll need to add a test that covers it and confirms the behaviour.

Secondly, I think this code reveals a bigger problem with this change, which is that `hash_function` is literally a Python callable. While that seems like a neat solution, it means that this function can really quite easily misbehave. For example, if I provide `hashlib.sha512` to the `hash_function`, we will fall into the bottom branch of this code and present it like an MD5 hash, when what we should do is refuse to process.

It would be better, I suspect, to use an enum here. Twisted has support for [constants with values](https://twistedmatrix.com/documents/current/core/howto/constants.html#constants-with-values) in an enumerated form. That would give us the opportunity to do this:

``` python
import hashlib
from twisted.python.constants import ValueConstant, Values

class FingerprintHashFunctions(Values):
    """
    Constants representing the supported hash functions for generating
    public key fingerprints.
    """
    MD5 = ValueConstant(hashlib.md5)
    SHA256 = ValueConstant(hashlib.sha256)
```

The method can then optionally accept this enum and fallback to using `FingerprintHasheFunctions.MD5` internally until the deprecation period is over. That way we can ensure that users can't accidentally misuse this function as easily, while also enabling extensibility (users can subclass this class and add extra values to support other hash functions if they want).

That also allows you to separate the act of building the digest from the act of rendering it. For all cases you build the digest identically, then you can just have an if block that says `if hashFunction is FingerprintHashFunctions.MD5` to control the rendering.

Lukasa: And in fact, @adiroiban has noted in the ticket itself that we may want to be able to control the representation as well. That would mean we'd want another enum (probably using `twisted.python.constants.Names`, not `Values`) that controls the representation.

twisted/conch/ssh/keys.py
reaperhulk: In OpenSSH the known_hosts model uses the complete public key and not the fingerprint. As you surmised, in general the fingerprint would be used as a shorthand to allow users to verify it manually (although it's possible somebody, somewhere is inappropriately using a fingerprint for programmatic comparison). I agree that the standard deprecation process should be fine.

acabhishek942:
@Lukasa @reaperhulk @rodrigc Can anyone please help me with the standard depreciation process?

rodrigc:
@acabhishek942  The document is here: http://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#how-to-deprecate-apis

I find it a bit difficult to follow, but you can ask followup questions if you need help.

twisted/conch/test/test_ckeygen.py
```diff
@@ -67,9 +67,13 @@ def test_printFingerprint(self):
         filename = self.mktemp()
         FilePath(filename).setContent(publicRSA_openssh)
         printFingerprint({'filename': filename})
+        printFingerprint({'filename': filename, 'format': 'sha256-base64'})
```
adiroiban: this needs to be in a new test case with a dedicated dosctring

you should not high existing tests, by adding more code to it... doing this to the extreme, we would have a single huge test for this whole code :)

twisted/conch/scripts/ckeygen.py
```diff
@@ -203,6 +205,8 @@ def _saveKey(key, options):
     @type options: L{dict}
     """
     keyTypeName = key.type().lower()
+    if not options.has_key('format'):
+        options['format'] = 'md5-hex'
```
adiroiban: shorter version: options.setdefault('format',  'md5-hex')

but here were are just printing the value and I don't think that the output format is public API, so we can use sha-base64 as the default format.

twisted/conch/scripts/ckeygen.py
```diff
@@ -239,7 +243,7 @@ def _saveKey(key, options):
     print('Your identification has been saved in %s' % (options['filename'],))
     print('Your public key has been saved in %s.pub' % (options['filename'],))
     print('The key fingerprint is:')
```
adiroiban: Maybe

```
print('The key fingerprint in %s is:' % (options['format']))
```

twisted/conch/ssh/keys.py
```diff
@@ -802,9 +817,21 @@ def fingerprint(self):
...
+            fingerprint = hashFunction[format](
+                self.blob()).digest()
+        except KeyError:
+            raise BadFingerPrintFormat(
+                "unsupported fingerprint format: %s" % format)
+
+        if format == 'sha256-base64':
+            return nativeString(base64.b64encode(fingerprint))
+        else:
+            return nativeString(
```
adiroiban: this is wrong... what would happen if I ask for 'md5-bogus'  of 'sha256-bogus' ? ... it looks like it will still use hex ... and I think that is wrong as instead we should have an error

twisted/conch/scripts/ckeygen.py
```diff
@@ -113,6 +113,8 @@ def generateDSAkey(options):
 
 
 def printFingerprint(options):
+    if not options.has_key('format'):
```
adiroiban: we have 

```
optFlags = [['fingerprint', 'l', 'Show fingerprint of key file.'],
```

so fingerprint is a flag.. it will never has a value.

you should make it an `optParameters` with a default value

twisted/conch/scripts/ckeygen.py
```diff
@@ -113,8 +113,7 @@ def generateDSAkey(options):
 
 
 def printFingerprint(options):
-    if not options.has_key('format'):
-        options['format'] = 'md5-hex'
+    options.setdefault('format', 'md5-hex')
```
adiroiban: this should no longer be needed as the option parsing code should take care of inserting the default value

twisted/conch/ssh/keys.py
```diff
@@ -819,19 +819,16 @@ def fingerprint(self, format='md5-hex'):
...
+            return nativeString(base64.b64encode(
+                hashFunction[format](self.blob()).digest()))
+        elif format == 'md5-hex':
             return nativeString(
                 b':'.join([binascii.hexlify(x)
-                           for x in iterbytes(fingerprint)]))
+                for x in iterbytes(hashFunction[format](self.blob()).digest())]))
+        else:
+            raise BadFingerPrintFormat(
+                "unsupported fingerprint format: %s" % format)
```
adiroiban: use something like

```
'Unsupported fingerprint format: "%s".' % (format,)
```

to force formatting even when we have tuples

```
>>> ra = (1, 2)
>>> 'bla %s' % ra
TypeError: not all arguments converted during string formatting
>>> 'bla %s' % (ra,)
'bla (1, 2)'
>>> 
```

twisted/conch/test/test_ckeygen.py
```diff
@@ -67,13 +67,35 @@ def test_printFingerprint(self):
...
+    def test_printFingerprintBadFingerPrintFormat(self):
+        """
+        L{printFigerprint} raises C{keys.BadFingerprintFormat} for formats
+        other than C{md5-hex} and C{sha256-base64}.
+        """
+        filename = self.mktemp()
+        FilePath(filename).setContent(publicRSA_openssh)
         self.assertRaises(BadFingerPrintFormat, printFingerprint,
             {'filename': filename, 'format':'sha-base64'})
+        self.assertRaises(BadFingerPrintFormat, printFingerprint,
```
adiroiban: a single assertion should be enough but would help to also check the message of the exception to make sure the right message was formatted

```
exception = self.assertRaises(BadFingerPrintFormat, printFingerprint, {'filename': filename, 'format':'sha-base64'})
self.assertEqual('bla', exception.message)
```

twisted/conch/test/test_keys.py
```diff
@@ -564,6 +564,38 @@ def test_notEqual(self):
         self.assertTrue(rsa1 != None)
 
 
+    def test_fingerprint(self):
```
adiroiban: this is for the default format... so this should be named `test_fingerprintDefault` and the docstring should explain the md5 is used when no format is speciffied.

we also need a test to check that when md5-hex format is explicitly call, all works fine

md5hex... so I think that it should be `test_fingerprint_md5_hex()`

twisted/conch/test/test_keys.py
```diff
@@ -564,6 +564,38 @@ def test_notEqual(self):
...
+        self.assertEqual(keys.Key(self.rsaObj).fingerprint('sha256-base64'),
+            'ryaugIFT0B8ItuszldMEU7q14rG/wj9HkRosMeBWkts=')
+        self.assertEqual(keys.Key(self.dsaObj).fingerprint('sha256-base64'),
+            'Wz5o2YbKyxOEcJn1au/UaALSVruUzfz0vaLI1xiIGyY=')
+
+
+    def test_fingerprintBadFormat(self):
+        """
+        A C{BadFingerPrintFormat} error is raised when format other than
+        C{md5-hex} or C{sha256-base64} is supplied.
```
adiroiban: here we can be more generic ... like `unsupported formats are requested`

the idea is that later we might want to add `sha1-hex` for example ,and then we might forget to update this docsting and we will end up with a docstring which lies :)

twisted/conch/scripts/ckeygen.py
```diff
@@ -33,10 +33,10 @@ class GeneralOptions(usage.Options):
                      ['type', 't', None, 'Specify type of key to create.'],
                      ['comment', 'C', None, 'Provide new comment.'],
                      ['newpass', 'N', None, 'Provide new passphrase.'],
-                     ['pass', 'P', None, 'Provide old passphrase.']]
+                     ['pass', 'P', None, 'Provide old passphrase.'],
+                     ['format', 'l', 'md5-hex', 'Fingerprint format of key file.']]
```
adiroiban: what was wrong with `fingerprint` as the name of this parameter?

btw, this is already changing a UX, so I think that it safe to use sha2 based format

twisted/conch/test/test_keys.py
```diff
@@ -564,6 +564,49 @@ def test_notEqual(self):
...
+        self.assertEqual(keys.Key(self.dsaObj).fingerprint('sha256-base64'),
+            'Wz5o2YbKyxOEcJn1au/UaALSVruUzfz0vaLI1xiIGyY=')
+
+
+    def test_fingerprintBadFormat(self):
+        """
+        A C{BadFingerPrintFormat} error is raised when unsupported
+        formats are requested.
+        """
+        self.assertRaisesRegexp(keys.BadFingerPrintFormat,
```
adiroiban: I think that we should use assertRaisesRegex as  assertRaisesRegexp was removed in py3.... but do we need regex here?

why not do an exact match on the exception's attributes?

twisted/conch/test/test_ckeygen.py
```diff
@@ -81,12 +106,12 @@ def test_saveKey(self):
         base.makedirs()
         filename = base.child('id_rsa').path
         key = Key.fromString(privateRSA_openssh)
-        _saveKey(key, {'filename': filename, 'pass': 'passphrase'})
+        _saveKey(key, {'filename': filename, 'pass': 'passphrase', 'format': 'md5-hex'})
```
adiroiban: please try to wrap lines at 79 chars. thanks!

also, maybe this whole dict can be given a name... like `options` so that we can call it like

```
 _saveKey(key, options)
```

and have an idea what that dict represents

acabhishek942: I think not naming the dictionary is alright because that has been followed all through this module.
Example : https://github.com/twisted/twisted/blob/trunk/twisted/conch/test/test_ckeygen.py#L84,
https://github.com/twisted/twisted/blob/trunk/twisted/conch/test/test_ckeygen.py#L221,
https://github.com/twisted/twisted/blob/trunk/twisted/conch/test/test_ckeygen.py#L158

twisted/conch/ssh/keys.py
```diff
@@ -782,18 +790,24 @@ def public(self):
         return Key(self._keyObject.public_key())
 
 
-    def fingerprint(self):
+    def fingerprint(self, format='md5-hex'):
```
Lukasa: [As before](https://github.com/twisted/twisted/pull/468#discussion_r73852166), I feel like this works better as an enumerated type, rather than using ad-hoc string formatting.

Now that you've swapped to declaring both the hash function and the serialization method, however, we can stop using values constants and just use [named constants](https://twistedmatrix.com/documents/current/core/howto/constants.html#constant-names):

``` python
import hashlib
from twisted.python.constants import NamedConstant, Names

class FingerprintFormats(Names):
    """
    Constants representing the supported forms of key fingerprints.
    """
    MD5_HEX = NamedConstant()
    SHA256_BASE64 = NamedConstant()
```

Any reason not to do this?

twisted/conch/ssh/keys.py
```diff
@@ -802,9 +816,18 @@ def fingerprint(self):
...
+        if format == 'sha256-base64':
+            return nativeString(base64.b64encode(
+                hashFunction[format](self.blob()).digest()))
+        elif format == 'md5-hex':
+            return nativeString(
+                b':'.join([binascii.hexlify(x)
+                for x in iterbytes(hashFunction[format](self.blob()).digest())]))
+        else:
+            raise BadFingerPrintFormat(
+                'Unsupported fingerprint format: %s' % (format,))
```
Lukasa: So here we have `if` tests on the entire string, as well as a `hashFunction` dict. The problem with that is that the `hashFunction` dict lookup serves no purpose: we have already had to check the string value in the `if` test, so _then_ doing a dict lookup is just giving us a hardcoded answer that we already know.

In this case, I think `hashFunction` can go.

twisted/conch/scripts/ckeygen.py
```diff
@@ -33,10 +33,10 @@ class GeneralOptions(usage.Options):
                      ['type', 't', None, 'Specify type of key to create.'],
                      ['comment', 'C', None, 'Provide new comment.'],
                      ['newpass', 'N', None, 'Provide new passphrase.'],
-                     ['pass', 'P', None, 'Provide old passphrase.']]
+                     ['pass', 'P', None, 'Provide old passphrase.'],
+                     ['fingerprint', 'l', 'sha256-base64', 'Fingerprint format of key file.']]
 
-    optFlags = [['fingerprint', 'l', 'Show fingerprint of key file.'],
-                ['changepass', 'p', 'Change passphrase of private key file.'],
+    optFlags = [['changepass', 'p', 'Change passphrase of private key file.'],
```
Lukasa: @glyph @hawkowl: Can I get a ruling on this? Does our compatibility policy apply to CLI tools like this one?

twisted/conch/ssh/keys.py
```diff
@@ -782,18 +801,24 @@ def public(self):
...
 
-    def fingerprint(self):
+    def fingerprint(self, format=FingerprintFormats.MD5_HEX):
         """
-        Get the user presentation of the fingerprint of this L{Key}.  As
-        described by U{RFC 4716 section
-        4<http://tools.ietf.org/html/rfc4716#section-4>}::
+        The fingerprint of a public key consists of the output of the
+        message-digest algorithm in the specified format.
+        Supported formats inclide C{md5-hex} and C{sha256-base64}
```
Lukasa: These need to be updated to be `L{}` to the correct things. Also, `inclide` should be `include`.

twisted/conch/scripts/ckeygen.py
glyph: The letter? No.  but hopefully we can stay in the spirit.  It seems like this meets it either way though; you're just adding stuff?  You're always allowed to add stuff.

Lukasa: Nah, it looks like fingerprint moved from being a flag to an option.

twisted/conch/scripts/ckeygen.py
```diff
@@ -33,7 +33,8 @@ class GeneralOptions(usage.Options):
                      ['type', 't', None, 'Specify type of key to create.'],
                      ['comment', 'C', None, 'Provide new comment.'],
                      ['newpass', 'N', None, 'Provide new passphrase.'],
-                     ['pass', 'P', None, 'Provide old passphrase.']]
+                     ['pass', 'P', None, 'Provide old passphrase.'],
+                     ['format', 'o', 'sha256-base64', 'Fingerprint format of key file.']]
```
acabhishek942: @glyph can you please review. and get a ruling on the compatibility policy.

glyph: I could have sworn I did this already, but, here it is again: 

This is adding stuff. You're always allowed to add stuff.  Is there some attribute or function being deleted here that I'm not seeing?

adiroiban: The latest version is just adding stuff.

so now we have

```
COMMAND --fingerprint --format=SOMETHING
```

I think that in the previous proposed version it was suggested to convert the `fingerprint` flag to an parameter

```
COMMAND --fingerprint=FORMAT
```

in trunk we now have fingerprint as a flag

```
COMMAND --fingerprint
```

glyph: Cool.  Better to do this as an addition in case anyone is using it in scripts or something :)

rodrigc:
Merged in https://github.com/twisted/twisted/pull/518

PR #470, opened 2016-08-07
_bytesRepr returns reprs for bytes that begin with b on Python 2 and 3.

markrwilliams:
See https://twistedmatrix.com/trac/ticket/8731 and https://github.com/twisted/twisted/pull/461#issuecomment-238083180

### DIFF
--- a/twisted/test/test_compat.py
+++ b/twisted/test/test_compat.py
@@ -887,6 +887,7 @@ class RawInputTests(unittest.TestCase):
         self.assertEqual(stdout.data, "Prompt")
 
 
+
 class FutureBytesReprTests(unittest.TestCase):
     """
     Tests for L{twisted.python.compat._bytesRepr}.
@@ -900,6 +901,7 @@ class FutureBytesReprTests(unittest.TestCase):
         """
         exc = self.assertRaises(TypeError, _bytesRepr, ["not bytes"])
         self.assertEquals(str(exc), "Expected bytes not ['not bytes']")
+
 
     def test_bytesReprPrefix(self):
         """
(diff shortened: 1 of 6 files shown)


rodrigc:
This is useful.  However, instead of creating a branch in your https://github.com/markrwilliams/twisted repository, can you create the 8731-bytes-repr-with-b in the https://github.com/twisted/twisted repository and file the pull request from that?

The buildbots only run on branches in the Twisted repository, so that will make it easier to look at all the build results.

twisted/test/test_compat.py
```diff
@@ -885,3 +885,25 @@ def write(self, data):
...
+
+    def test_bytesReprNotBytes(self):
+        """
+        L{twisted.python.compat._bytesRepr} raises a
+        L{TypeError} when called any object that is not an instance of
+        L{bytes}.
+        """
+        exc = self.assertRaises(TypeError, _bytesRepr, ["not bytes"])
+        self.assertEquals(str(exc), "Expected bytes not ['not bytes']")
+
```
rodrigc, line 904: You should run twistedchecker on this file.  It will complain that you need two newlines here, I think.

markrwilliams:
@rodrigc thanks for the feedback!  I'm closing this in favor of a PR from a branch on https://github.com/twisted/twisted

rodrigc:
@markrwilliams Thanks!  Keep that in mind also, when people submit PR's, instead of merging their PR directly from GitHub, it is better to take their PR and create a branch in https://github.com/twisted/twisted and make sure that the buildbots pass for that branch.

Not sure if you know this trick, but I do something like:

```
git clone https://github.com/twisted/twisted twisted_test
cd twisted_test
git fetch origin pull/470/head:8731-bytes-repr-with-b
git checkout 8731-bytes-repr-with-b
git log
git push origin 8731-bytes-repr-with-b
```

You get the idea.

PR #472, opened 2016-08-07
Fix twisted.test.test_stateful on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8718


acabhishek942:
looks good to me

PR #473, opened 2016-08-07
[8730] Send per-job status for Travis.

adiroiban:
Scope
=====

This updates the travis-ci build system to send separate commit status for each Travis-CI job.

In this way, when travis fails, you can go directly to the job.

RIght now, you have to go to the build, check the list of jobs and then click on the ones which are red.

See: https://twistedmatrix.com/trac/ticket/8730

Changes
=======

Add travis before after-success and after-failure hooks to update the commit status.

How to test
=========

See the commit status for this PR.

Leave your feedback if you find this an improvement, no improvement or an annoyance.

The current version has conflicts, but is not reviewed since 2016 so at this point there is not much use to keep it up to date with trunk

Thanks


adiroiban:
Closing as @glyph  was -1 on this See https://twistedmatrix.com/trac/ticket/8730#comment:7

I still think that we should do it as we already report independent jobs for buildbot, and now for circle-ci..

This gives you a direct link to the failed Travis-CI job.

Now, you need to click on the Travis build, look in the list for builds which failed and then click again :)

PR #476, opened 2016-08-08
8731 bytes repr with b

markrwilliams:
See: https://twistedmatrix.com/trac/ticket/8731

A version of #470 from Twisted's own repo so the buildbots will run.


twisted/scripts/_twistd_unix.py
```diff
@@ -198,6 +200,38 @@ def preApplication(self):
         self.oldstderr = sys.stderr
 
 
+    def _formatException(self, exception):
```
markrwilliams: This isn't strictly related to the problem described in the ticket, but without it twistd won't work correctly on Python 3.

rodrigc:
pyflakes is complaining about an unused import: https://travis-ci.org/twisted/twisted/jobs/150511323#L260

rodrigc:
I think this is looking OK.  Please update https://twistedmatrix.com/trac/ticket/8731 with a link to this PR,
and also in the **Branch** field in that ticket, please put the name of the branch.

twisted/python/compat.py
```diff
@@ -765,6 +765,29 @@ def _coercedUnicode(s):
...
+    @param bytestring: The string to repr.
+    @type bytestring: L{bytes}
+
+    @raise TypeError: The input is not L{bytes}.
+
+    @return: The repr with a leading 'b'.
+    @rtype: L{bytes}
+    """
+    if not isinstance(bytestring, bytes):
+        raise TypeError("Expected bytes not %r" % (bytestring,))
```
evgeni, line 782: shouldn't this be `"%s" % type(bytestring)`?

adiroiban, line 782: I guess that the intention was to get both the type and content from the repr() part.

for printing errors like this I usually truncate the input data at 50 chars ... so that if you use it for 16kb you will not get an excessive error message

twisted/topfiles/8731.bugfix
```diff
@@ -0,0 +1 @@
+Add helper to create bytes reprs that always begin with b.  twistd on UNIX platforms can report non-ASCII errors on Python 2 and all errors on Python 3 encountered during daemonization.
```
adiroiban: `Add helper to create bytes reprs that always begin with b.` ...  which helper? 
Please note that the release notes are provided out of the context of this commit .
Also, this is about a private API and the release notes should not contain private stuff

twisted/scripts/_twistd_unix.py
```diff
@@ -198,6 +199,38 @@ def preApplication(self):
...
+        """
+        Format the C{exception} in preparation for writing to the
+        status pipe.  This does the right thing on Python 2 if the
+        exception's message is Unicode, and in all cases limits the
+        length of the message, prior to encoding, to 100 codepoints.
+
+        This means the return message may be longer than 100 bytes.
```
adiroiban: I think that 100 byte is just a way to prevent excessive output..

the docstring can be a bit vague and inform that messages are truncated at about 100 characters :)

twisted/scripts/_twistd_unix.py
```diff
@@ -198,6 +199,38 @@ def preApplication(self):
...
+        # codec and the backslashreplace error handler.
+        exceptionLine = traceback.format_exception_only(exception.__class__,
+                                                        exception)[-1]
+        # remove exception name
+        exceptionMessage = exceptionLine.split(': ', 1)[-1]
+        # remove the trailing newline and limit the total length to
+        # the passed string to 100
+        truncatedMessage = exceptionMessage.strip()[:98]
+        formattedMessage = '1 %s' % truncatedMessage
+        # On Python 3, ensure the message
```
adiroiban: I think that this sentence it unfinished :)

twisted/scripts/_twistd_unix.py
adiroiban: minor comment
I think that this should have a less generic name :) so that people will not be tempted to use it for formatting any exception :)

also, if this fixed a problem in Python3 I think that this should refect in the releas notes and the Trac ticket description / title

twisted/scripts/_twistd_unix.py
adiroiban: Also, we don't have any tests for the exact limit of 100bytes.. so this is not part of the contract .

twisted/scripts/_twistd_unix.py
```diff
@@ -198,6 +199,38 @@ def preApplication(self):
...
+        @return: The formatted message, suitable for writing to the
+            status pipe.
+        @rtype: L{bytes}
+        """
+        # On Python 2 this will encode Unicode messages with the ascii
+        # codec and the backslashreplace error handler.
+        exceptionLine = traceback.format_exception_only(exception.__class__,
+                                                        exception)[-1]
+        # remove exception name
+        exceptionMessage = exceptionLine.split(': ', 1)[-1]
```
adiroiban: this looks fragile as it depends on the custom formatting of traceback... is there any problem if we also leave the exception name?

twisted/scripts/_twistd_unix.py
```diff
@@ -198,6 +199,38 @@ def preApplication(self):
...
+        # remove the trailing newline and limit the total length to
+        # the passed string to 100
+        truncatedMessage = exceptionMessage.strip()[:98]
+        formattedMessage = '1 %s' % truncatedMessage
+        # On Python 3, ensure the message
+        if _PY3:
```
adiroiban, line 227: I understand the struggle here, but the targeted code looks complicated ... maybe we can just live with different output on python2 and python3 and have the `if _PY3` branch moved to the testing code

twisted/scripts/_twistd_unix.py
adiroiban: hm...we have a test... so this is a problem :) maybe the test can be re-written to check for size less than 150 to emphasize that the 100bytes is not a hard limit.

twisted/test/test_twistd.py
```diff
@@ -1802,7 +1857,7 @@ def startService(self):
         self.assertEqual(
             self.mockos.actions,
             [('chdir', '.'), ('umask', 0o077), ('fork', True), 'setsid',
-             ('fork', True), ('write', -2, '1 ' + 'x' * 98),
+             ('fork', True), ('write', -2, b'1 ' + b'x' * 98),
```
adiroiban: instead of `b'x'` maybe you can add `u'\N{SNOWMAN}'.encode('utf-8')[:-1]` to get some non-ascii bytes and see what it happens if we receive partial UTF-8 characters

twisted/scripts/_twistd_unix.py
markrwilliams: I'd prefer to leave the name in.  I didn't because I was worried about gratuitously changing the error message, since it's a kind of public API.  But I've already changed it a lot by using `_bytesRepr`, so I'm less worried about it now.  And I also think seeing what would be the last line of the traceback would be very helpful.

Should I ask on the mailing list if anybody has strong opinions about the format of this message?

twisted/scripts/_twistd_unix.py
markrwilliams, line 227: I can't avoid it :(   The return value of `_formatException` gets written to a pipe, so it has to be bytes.  The truncated comment should read: "On Python 3, ensure the message is encoded as bytes in the same way as it is on Python 2".  So while you're right that part of the goal here is to produce consistent output, the bigger part is encoding a native Python 3 string to bytes, which is unavoidable.

twisted/test/test_twistd.py
markrwilliams: I like the sound of this a lot, but what would it test?  The code should never receive partial utf-8 sequences.  Here are the cases in which Unicode can be written to the pipe:
1. It's Python 2 and the exception raised in the child has a `unicode` message.  This can happen in lots of ways (`<some bytes> % (<some unicode>,)` -> `<unicode>`!)   In this case `traceback.format_exception_only` encodes the object into 'ascii' with 'backslashreplace', so we might see partial backslash representations (e.g. `"u'\N{BULLET}'"` -> '\u20`) if the code point occurs around the 100 byte mark.
2. It's Python 3.  In this case the exception message is truncated to 100 code points before encoding, so we might have _more_ than 100 bytes, but the backslash representations will be complete.

In other words, only in the Python 2 case will truncation occur after encoding, but it will be on backslash representations of code points, not their UTF-8 encoding.

Since the read side of the pipe, which this test exercises, is kind of a private API, or at least can't exist without the write side I just described, I don't think it needs to be tested against partial UTF-8 sequences.  I do think I'll test it against partial backslash representations, though, as that truncation can definitely occur.

Does that sound reasonable?

twisted/python/compat.py
markrwilliams, line 782: I prefer something like this too, though I usually use `repr`/reprlib`for it.  Should`_bytesRepr`do truncation?  If so, I'd prefer it be controlled by an argument that defaults to`False`, and that it use a`repr[lib].Repr` instance that's also retrieved from an argument, e.g.:

`def _bytesRepr(bytestring, truncate=False, reprInstance=Repr()):`

Does that seem OK?

twisted/scripts/_twistd_unix.py
markrwilliams: Agree on the name!  I updated the Trac ticket, too.

twisted/scripts/_twistd_unix.py
markrwilliams: I think the right thing to do is to change [the test below](https://github.com/twisted/twisted/pull/476#discussion-diff-73857170R1860) to verify that a truncated message (e.g., `b'Exception: blah \u20'`) works.  I also think the writer should truncate the message _after_ encoding.  That way we _know_ only 100 bytes will be written in the pipe.  I want to say it's impossible for 100 code points to end up being longer than `PIPE_BUF`, but I'd rather rewrite the code to be sure that's the case.  Otherwise there's a risk the writer will block, which would be terrible!

PR #479, opened 2016-08-08
Fix process related tests on Windows Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8735

These fixes are cherry-picked from @hawkowl 's branch:

https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/test/test_process.py
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/test/process_stdinreader.py
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/internet/_win32stdio.py
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/internet/test/process_cli.py
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/internet/_dumbwin32proc.py
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/python/compat.py


twisted/python/compat.py
```diff
@@ -756,6 +759,25 @@ def _coercedUnicode(s):
...
 
 
+def _maybeMBCS(s):
+    """
+    Convert the string C{s} to a L{unicode} string, if required.
+
+    @param s: The string to convert.
+    @type s: L{bytes} or L{unicode}
+
+    @return: The string, decoded using MBCS if needed.
```
glyph, line 769: I feel like this _really_ needs a [`@raise`](https://epydoc.sourceforge.net/epytext.html) explaining that MBCS is definitely _not_ a sure-fire encoding solution.

glyph:
I like what I see here but it concerns me that `test_manyProcesses` failed in _4_ places after this build run as opposed to the usual 0-1; hoping that the continuous integration gods are with us for the remainder here and it was just bad luck of an existing flaky test.

glyph:
@rodrigc - `test_manyProcesses` is definitely known to be buggy / race condition prone, and I don't want to make it this branch's responsibility to clean it up, but can you think of any reason that this might make processes linger longer than they used to?

rodrigc:
Well in the next run, I see that the many processes test passed? https://buildbot.twistedmatrix.com/boxes-supported?8735-rodrigc-process-py3=&num_builds=10

I am not sure about test_manyProcesses , there was some discussion about it in http://twistedmatrix.com/trac/ticket/3404 but I don't understand the issues.  I know that creation of processes is different between POSIX and Windows, so it looks like this test is pushing on this behavior.

twisted/python/compat.py
rodrigc, line 769: What do you want the text in the @raise to say?

glyph, line 769: Just that it raises `UnicodeDecodeError` if passed a bytestring that can't be decoded using the local encoding, and also that the local MBCS encoding may well be incapable of representing reasonable strings because it's tied to the charset.  If in fact I am understanding MBCS properly - every time I touch Windows I have to remind myself how all this legacy stuff works :-\

PR #480, opened 2016-08-09
Add missing twisted.words entries to dist3.py so that Python 3 tests run

rodrigc:
https://twistedmatrix.com/trac/ticket/8738


alex:
👍  lgtm

PR #485, opened 2016-08-11
Run separate Appveyor tests for select and iocp reactor

rodrigc:
https://twistedmatrix.com/trac/ticket/8677

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -55,9 +55,6 @@ environment:
 
 matrix:
   allow_failures:
-    # Some tests are still failing on Python 3.
-    # These tests will be fixed in separate tickets.
-    - PYTHON_VERSION: "3.5"
     # Some IOCP tests are failing on Appveyor:
     # https://twistedmatrix.com/trac/ticket/8679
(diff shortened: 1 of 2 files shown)


rodrigc:
This is a replacement for https://github.com/twisted/twisted/pull/412

rodrigc:
@opalmer can you review this?  You can see the appveyor results from this change here: https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1005

opalmer:
Looks good 👍 .

What would you think about adding build artifacts so the built wheels, coverage, etc can be downloaded after the build is complete?  Does not have to be in this PR of course...

rodrigc:
@opalmer In terms of coverage, this build is running coverage and uploading the coverage results to codecov.io.  In the past, I don't believe that the Twisted project made build artifacts available from the buildbots, so this would be a new thing for the Twisted project.  I think it is a reasonable idea though.

I think the current direction is to see if appveyor can be integrated with the release process,
so that for each release, the wheels for Windows would be generated from appveyor.

PR #487, opened 2016-08-11
[#8087] Add yield from

hawkowl:
https://twistedmatrix.com/trac/ticket/8087


twisted/topfiles/8087.feature
```diff
@@ -0,0 +1 @@
+twisted.internet.asyncioreactor has been added, a Twisted reactor that shares the asyncio eventloop, available on Python 3.4+. It can be used by passing --reactor=asyncio to twistd and similar utilities.
```
markrwilliams: woops, is this the right topfile comment?

twisted/topfiles/8087.feature
```diff
@@ -0,0 +1 @@
+'yield from' can now be used on Deferreds, when wrapped with twisted.internet.defer.ensureDeferred.
```
Lukasa: This text, while right, is a bit confusing. You don't have to wrap _every_ `Deferred` with `ensureDeferred`, just the "top level" one, right?

PR #489, opened 2016-08-12
[#8748] future-deferred adapters

hawkowl:
https://twistedmatrix.com/trac/ticket/8748


njsmith:
I'm not sure of the solution, but:
```
<njs> glyph: re: https://github.com/twisted/twisted/compare/trunk...8748-future-adapt, I'm not sure futureToDeferred is correct. Future's APIs are... kinda odd. Like, Future.exception() returns an exception, except when it raises it.
<glyph> njs: please comment inline :)
<njs> where?
<njs> that's not a PR, so not sure if comments there will go anywhere useful
<glyph> njs: I'm about to put it up for review actually
<runciter> hm
<runciter> njs: so it can raise an exception when the Future's in an invalid state: https://github.com/python/cpython/blob/d4d48743ac20854de104e08dd66972471684f676/Lib/asyncio/futures.py#L247-L263 
<njs> runciter: or if the future was cancelled
<runciter> looks like cancelled is worrisome
<njs> cancelled is what jumped out at me but in general the API seems subtle enough that I'm not confident that I stared at it long enough to make any recommendation :-)
<runciter> njs: hm, i guess that's because it returns None when there's no exception set?
<runciter> oh jeez
<runciter> njs: good catch :)
<runciter> i think i have a reproducing case for this
<runciter> https://gist.github.com/markrwilliams/b7a3c5b914d1f207e2250c421760d51c
<runciter> hm
<runciter> there we go
<runciter> njs: so _adapt will fail if the future's been canceled
<runciter> i'm pretty that never needs to call future.exception - it can just use Failure's traceback nabbing magic?
<runciter> njs: thanks!
<njs> runciter: yeah I think on balance I'd go with try: d.callback(future.result()) except: d.errback()
<runciter> njs: 👍
<njs> the limitation of this approach is that there's an ambiguity between errors from the future, and errors from future.result / d.callback
<njs> but those shouldn't happen?
<runciter> i *think* the only one that should happen is CanelledError
<njs> maybe try: result = future.result() except: d.errback() else: d.callback(result) would be better, in case the deferred was already called
<runciter> ah, got it
<njs> but yeah, InvalidStateError is documented as possible but should never happen, so I think it's safe
<runciter> well, if it happens, we want to fail loudly outside of user code i think
<njs> then you need something more complicated :-)
<njs> this is a weirdly awkward API! most of asyncio's core seems better-thought-through than this
<njs> I guess another issue is that this is all totally unsafe if KeyboardInterrupt arrives in the middle, but... maybe that's more like a fundamental architectural limitation :-)
```

src/twisted/test/test_defer.py
```diff
@@ -2964,3 +2964,78 @@ def test_deprecatedTimeout(self):
...
+
+        loop = new_event_loop()
+        loop.call_soon_threadsafe(d.errback, theFailure)
+        loop.run_until_complete(defer.deferredToFuture(loop, d))
+
+        self.assertEqual(results, [])
+        self.assertEqual(errors, [theFailure])
+
+
+    def test_futureToDeferredFutureCancelled(self):
```
moshez: Drive by: More tests of future to deferred needed.

src/twisted/internet/defer.py
```diff
@@ -753,6 +753,91 @@ def send(self, value=None):
...
+        """
+        Adapt an L{asyncio.Future} to a L{Deferred}.
+
+        @note: This creates a L{Deferred} from a L{asyncio.Future}, I{not} from
+            a C{coroutine}; in other words, you will need to call
+            L{asyncio.async}, L{asyncio.ensure_future},
+            L{asyncio.AbstractEventLoop.create_task} or create an
+            L{asyncio.Task} yourself to get from a C{coroutine} to a
+            L{asyncio.Future} if what you have is an awaitable coroutine and
+            not a L{asyncio.Future}.  (The length of this list of techniques is
```
wsanchez, line 808: newline after period

src/twisted/internet/defer.py
```diff
@@ -753,6 +753,91 @@ def send(self, value=None):
...
+        @param future: The Future to adapt.
+        @type future: L{asyncio.Future}
+
+        @return: A Deferred which will fire when the Future fires.
+        @rtype: L{Deferred}
+        """
+        def adapt(result):
+            try:
+                extracted = result.result()
+            except:
```
wsanchez, line 822: Does this need to be a bare except? I'm guessing that catching KeyboardInterrupt here isn't appropriate.

src/twisted/python/compat.py
```diff
@@ -40,6 +40,16 @@
...
+if sys.version_info >= (3, 4, 0):
+    _PY34PLUS = True
+else:
+    _PY34PLUS = False
+
+if sys.version_info >= (3, 5, 0):
+    _PY35PLUS = True
+else:
+    _PY35PLUS = False
+
```
wsanchez, line 52: Any reason why these aren't written are one-liners? As in:
```python
_PY34PLUS = sys.version_info >= (3, 4, 0)
_PY35PLUS = sys.version_info >= (3, 5, 0)
```

Not a big deal, but this just looks verbose to me.

src/twisted/test/test_defer.py
```diff
@@ -11,8 +11,17 @@
...
 import re
 
+from twisted.python.compat import _PY3, _PY34PLUS
+
+if _PY34PLUS:
+    from asyncio import new_event_loop, Future, CancelledError
+    asyncSkip = None
+else:
+    asyncSkip = "asyncio not available before python 3.4"
+
```
wsanchez, line 21: It would be more robust here, perhaps, to catch ImportError.  As it is, I'm assuming you checked about which versions include the asyncio module, but maybe a backport of it comes along…

Anyway, seems better to test for the feature directly if you can, rather than testing for interpreter versions, and here you should be able to test for the feature.

wsanchez (commented):
Mostly optional notes, but I think the bare except looks too fishy.

src/twisted/internet/defer.py
glyph: wait… why?

src/twisted/python/compat.py
glyph: Actually now that you mention it, maybe this should just be written inline everywhere, rather than even declaring these (ahem, undocumented) constants?  It's a short enough idiom...

src/twisted/test/test_defer.py
glyph: I've gone back and forth over this throughout the years, so I'm sympathetic to that point of view, but my current thinking goes like this:

We know that `asyncio` is part of the stdlib on 3.4+.  It _must_ be importable there or there's some kind of major problem with the python installation.  In the case where we expect it to be available because the interpreter says it should be, we go ahead and import it, and fail hard if something is so messed up that it's not importable.  In the case where we _don't_ think it should be available, there's no point in trying.

For _application_ code there's a benefit to being lenient and trying to function in the face of some ambiguity, but for _tests_ this still strikes me as the better way of dealing with conditional imports.

I'm open to being convinced otherwise though.

src/twisted/internet/defer.py
glyph: This is the same as catching an exception in `Deferred._runCallbacks`.  We don't just eat the exception, we construct a `Failure` and propagate it along the returned `Deferred`'s errback chain.  Calling out every bare except is still a good code review habit, so thanks for doing that :) but in this case, yes, I believe a bare except is the right call.

glyph:
@wsanchez - was that "address and merge" or "address and resubmit"?

glyph:
@wsanchez Ping?

src/twisted/internet/defer.py
wsanchez: Because newline after each sentence?
I thought that was a thing, but I can't find it in the [Twisted Coding Standard](http://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html), so never mind, I guess.  Maybe that was just for the narrative docs.

src/twisted/internet/defer.py
wsanchez: OK, I see that now.  Carry on.

src/twisted/python/compat.py
wsanchez: I don't feel strongly about that either way, but the one-liners seem nicer, is all.  :-)

src/twisted/test/test_defer.py
wsanchez: Hrm, that's fair.  I tend to think that it's better (and the intentions often clearer; though in this case the strings are pretty clear, so it's a thin argument here) to test for the think you need directly (import THING), and not indirectly (if python version is X then assume THING).

wsanchez:
Sorry Glyph, I fell into a hole for a while.

The only thing I thought *required* a fix was the bare except, but I missed that it was getting repackaged into a Failure, so this can merge as-is unless you wanna clean up `_PY35PLUS` and friends.

glyph:
> Sorry Glyph, I fell into a hole for a while.

No worries, it happens to us all the time.  Thanks for updating this; hopefully I can get to do something with it tomorrow.

glyph:
Since `_PY35PLUS` and friends are private implementation details, I'll happily clean those up later :).  Will merge as-is.

PR #490, opened 2016-08-12
Fix test_tcp on Windows Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8749

No joke, I ran the same test from the same checkout using Python 2.7
and Python 3.5 under Windows 10, and I got different errno values.

@hawkowl , please review this.  This is similar to the fix that you did here:
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/test/test_tcp.py

but I coded it differently, so you can review it.

### DIFF
--- a/twisted/internet/test/test_socket.py
+++ b/twisted/internet/test/test_socket.py
@@ -96,6 +96,7 @@ class AdoptStreamPortErrorsTestsBuilder(ReactorBuilder):
         reactor = self.buildReactor()
 
         port = socket.socket()
+        port.bind(("127.0.0.1", 0))
         port.listen(1)
         self.addCleanup(port.close)
 
@@ -119,6 +120,7 @@ class AdoptStreamPortErrorsTestsBuilder(ReactorBuilder):
         portSocket = socket.socket()
         self.addCleanup(portSocket.close)
 
+        portSocket.bind(("127.0.0.1", 0))
         portSocket.listen(1)
         portSocket.setblocking(False)
 
@@ -236,6 +238,7 @@ class AdoptDatagramPortErrorsTestsBuilder(ReactorBuilder):
         portSocket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
         self.addCleanup(portSocket.close)
 
+        portSocket.bind(("127.0.0.1", 0))
         portSocket.setblocking(False)
 
         # The file descriptor is duplicated by adoptDatagramPort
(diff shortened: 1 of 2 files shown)


twisted/test/test_tcp.py
```diff
@@ -1138,11 +1140,18 @@ def getHandleErrorCode(self):
         Return the errno expected to result from writing to a closed
         platform socket handle.
         """
-        # These platforms have been seen to give EBADF:
-        #
-        #  Linux 2.4.26, Linux 2.6.15, OS X 10.4, FreeBSD 5.4
-        #  Windows 2000 SP 4, Windows XP SP 2
-        return errno.EBADF
+        if platform.isWindows():
```
markrwilliams: I think this could be simplified to:

``` python
# Windows and Python 3: WSAENOTSOCK
# Windows and Python 2: EBADF
# Linux, FreeBSD, Mac OS X: return EBADF
if platform.isWindows() and _PY3:
    return errno.WSAENOTSOCK
return errno.EBADF
```

Note that all the listed Linux, OS X, and FreeBSD versions are ancient and now irrelevant.

twisted/internet/test/test_socket.py
```diff
@@ -79,7 +81,10 @@ def test_invalidDescriptor(self):
         exc = self.assertRaises(
             socket.error,
             reactor.adoptStreamPort, fileno, socket.AF_INET, ServerFactory())
-        self.assertEqual(exc.args[0], errno.EBADF)
+        if platform.isWindows() and _PY3:
+            self.assertEqual(exc.args[0], errno.WSAENOTSOCK)
```
markrwilliams, line 85: Using separate `self.assertEquals` might seem redundant, but I like it because the traceback will point at back to this line and the symbolic name for this constant.

twisted/internet/test/test_socket.py
```diff
@@ -91,6 +96,7 @@ def test_invalidAddressFamily(self):
         reactor = self.buildReactor()
 
         port = socket.socket()
+        port.bind(("127.0.0.1", 0))
```
markrwilliams, line 99: It's necessary to `bind` before `listen`ing on Windows? If you call `listen(2)` without calling `bind` on a Un*x-like platform will have it bind to all interfaces, with a port chosen by the kernel (equivalent to `bind('0.0.0.0', 0)`)

rodrigc, line 99: Yes it is necessary to bind() before calling listen() on Windows Python 3.  Windows Python 3 throws an OSError exception with WSA error 10022 (WSAEINVAL) if you don't bind().

twisted/test/test_tcp.py
rodrigc: OK, I made the change.

PR #492, opened 2016-08-13
https://tm.tl/4362 

glyph:


glyph:
WTF, github. This branch does not have 2130 files in it.  I think I'm going to have to close and open again?

PR #493, opened 2016-08-13
8754 kqueue error handling

glyph:
Use the correct exception types for the python 3 exception hierarchy.

### DIFF
--- a/twisted/internet/asyncioreactor.py
+++ b/twisted/internet/asyncioreactor.py
@@ -171,18 +171,18 @@ class AsyncioSelectorReactor(PosixReactorBase):
                                               self._readOrWrite, writer,
                                               False)
             self._writers[writer] = fd
-        except IOError as e:
+        except PermissionError:
             self._unregisterFDInAsyncio(fd)
-            if e.errno == errno.EPERM:
-                # epoll(7) doesn't support certain file descriptors,
-                # e.g. filesystem files, so for those we just poll
-                # continuously:
-                self._continuousPolling.addWriter(writer)
-            else:
-                raise
-        except BrokenPipeError as e:
+            # epoll(7) doesn't support certain file descriptors,
+            # e.g. filesystem files, so for those we just poll
+            # continuously:
+            self._continuousPolling.addWriter(writer)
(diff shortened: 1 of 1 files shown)


twisted/internet/asyncioreactor.py
```diff
@@ -171,18 +171,18 @@ def addWriter(self, writer):
...
-                self._continuousPolling.addWriter(writer)
-            else:
-                raise
-        except OSError as e:
+            # epoll(7) doesn't support certain file descriptors,
+            # e.g. filesystem files, so for those we just poll
+            # continuously:
+            self._continuousPolling.addWriter(writer)
+        except BrokenPipeError:
             # The kqueuereactor will raise this if there is a broken pipe
```
markrwilliams, line 184: i am a little worried about the code duplicated here and in the blank `except` statement, but only enough to mention it.  i think it's ok as it creates a place for this comment.  not a merge blocker.

glyph, line 184: I don't like it either, but there's no good way (that I'm aware of) for Python to express "for exceptions A, B, C, do this common logic, but for C only, do this extra stuff" without isinstance checks.  Would you prefer an isinstance check?

markrwilliams, line 184: no, i really wouldn't :(  after a few minutes of thinking i can't see a better way to do this.  maybe it'd be good to extend the comment to mention that it's ok to discard `BrokenPipeError` because all it means is that the pipe got closed, and that EPIPE is to writable pipes as an empty read is to readable pipes? or add another comment, in the `except:` block, about how if that code runs it's really an error and not some vagary of asyncio/python 3/posix?

glyph, line 184: I dunno, that feels like a pretty common pattern to me; a bunch of 'except's with specific errors, then a general one which is just the "okay I give up" clean-up logic.

markrwilliams, line 184: yeah, i think i'm just still upset that `OSError` _is_ `IOError`, and a verbose comment would make me less upset :)  

once the builds are green, i say merge!

PR #494, opened 2016-08-14
Fix test_stdlib.py on Windows with Python 3.5

rodrigc:
https://twistedmatrix.com/trac/ticket/8755

This fix is taken from @hawkowl 's branch: 
https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/logger/test/test_stdlib.py


acabhishek942:
looks good to me

PR #495, opened 2016-08-14
Fix test_deprecate.py in Windows with Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8756

This fix is taken from @hawkowl 's branch: https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/python/test/test_deprecate.py


itamarst:
LGTM.

PR #496, opened 2016-08-14
Fix test_compat on Windows with Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8757

This fix is taken from @hawkowl 's branch: https://github.com/twisted/twisted/blame/moar-windows-8025-8/twisted/test/test_compat.py


acabhishek942:
looks good to me

PR #502, opened 2016-08-16
Clarify docstring for twisted.internet.interfaces.IProtocol.dataReceived()

rodrigc:
http://twistedmatrix.com/trac/ticket/8763

### DIFF
--- a/src/twisted/internet/interfaces.py
+++ b/src/twisted/internet/interfaces.py
@@ -1807,11 +1807,13 @@ class IProtocol(Interface):
         callback will be made upon the receipt of each complete protocol
         message.
 
-        @param data: a string of indeterminate length.  Please keep in mind
-            that you will probably need to buffer some data, as partial
-            (or multiple) protocol messages may be received!  I recommend
-            that unit tests for protocols call through to this method with
-            differing chunk sizes, down to one byte at a time.
+        Please keep in mind that you will probably need to buffer some data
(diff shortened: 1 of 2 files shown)


src/twisted/internet/interfaces.py
```diff
@@ -1807,11 +1807,12 @@ def dataReceived(data):
...
 
-        @param data: a string of indeterminate length.  Please keep in mind
-            that you will probably need to buffer some data, as partial
-            (or multiple) protocol messages may be received!  I recommend
-            that unit tests for protocols call through to this method with
-            differing chunk sizes, down to one byte at a time.
+        Please keep in mind that you will probably need to buffer some data
+        as partial (or multiple) protocol messages may be received!  I
+        recommend that unit tests for protocols call through to this method
+        with differing chunk sizes, down to one byte at a time.
```
derwolfe, line 1813: Just out of curiosity - why remove the param declaration entirely?

rodrigc, line 1813: Because it is not a string, it is bytes.  The fact that it is bytes is covered by the @type field.  Once you have that, then the only thing to put in the @param field would be something like
**@param data: the data that is received** which is redundant.  The rest of the long description can go in the docstring outside of any @param tag

itamarst:
I'd probably re-add the "indeterminate length" bit. And replace "I recommend" with "we recommend". But seems fine, merge with that if you want.

PR #503, opened 2016-08-18
[#8764] Make Twisted versioned by year.month.patch

hawkowl:
https://twistedmatrix.com/trac/ticket/8764

### DIFF
--- a/src/twisted/topfiles/8764.feature
+++ b/src/twisted/topfiles/8764.feature
@@ -1,0 +1,1 @@
+Twisted is now versioned based on a year.month.patch basis.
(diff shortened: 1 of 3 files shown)


rodrigc:
pyflakes failed on this patch: https://travis-ci.org/twisted/twisted/jobs/153146285#L285

glyph:
Poke. Conflicts.

src/twisted/topfiles/8764.feature
```diff
@@ -0,0 +1 @@
+Twisted is now versioned based on a year.month.patch basis.
```
thijstriemstra, line 1: Can you elaborate a little bit more on this? Ie. what was it before? What's the improvement?

glyph: Previously it was year.release-per-year.patch.

glyph: The major improvement is that you can now set version constraints, because you can predict exactly when the next incompatible version will be.

hawkowl:
Supersceded by #684

PR #507, opened 2016-08-19
8771 ssh userauth keyboard interactive

einarfd:
This will fix a problem with keyboard interactive userauth on Python3 and is a fix for https://twistedmatrix.com/trac/ticket/8771


rodrigc:
Merged in https://github.com/twisted/twisted/pull/510

PR #509, opened 2016-08-19
[#8770] Port twisted.mail.smtp to Python 3

hawkowl:
https://twistedmatrix.com/trac/ticket/8770


rodrigc:
This is a big patch, but due to the extensive use of str in the twisted.mail module where bytes should have been used, this cannot be avoided.  I made a similar attempt to port twisted.mail to Python 3 and made many of the same changes.

👍 please add a topfile and merge this.

I have further changes to the twisted.mail module for Python 3, so getting this in the tree will facilitate this, otherwise the patch is huge.

glyph:
This looks like it got a passing review, but was never merged.

bump @hawkowl

exarkun:
It doesn't look like this was ever actually submitted for review.  The trac ticket says:

    It's ported and the tests pass, but it still needs more tests,
    as there's some quite big gaps in it.

eadamsatx:
There are 8 merge conflicts between this branch and trunk at this point.  I'd really like to use smtp on Python 3.  If I clean up the branch and add testing, is this approach still valid?

hawkowl:
Yes and no, I needed to break this up into several patches, some of which
might not have landed. I'll look soon, since I want to get this part done
and dusted.

redapple:
@rodrigc , @hawkowl ,
in case you haven't seen my comment the other day, I found some issues with current trunk `src/twisted/mail/smtp.py`: see https://github.com/scrapy/scrapy/issues/2375#issuecomment-299226510

PR #511, opened 2016-08-20
[#8773] pep-479 madness

hawkowl:
https://twistedmatrix.com/trac/ticket/8773


glyph:
Compatibility concern: since we've done a release where `Deferred` was iterable on 2.7, rolling it back is potentially incompatible.  This might be of interest to folks maintaining their own 2/3 cross-compatibility.

PR #514, opened 2016-08-21
Fix test_outputSignal test on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8583


rodrigc:
Closing after feedback

PR #519, opened 2016-08-23
8779 buildpy test dir and skip

markrwilliams:
See https://twistedmatrix.com/trac/ticket/8779

### DIFF
--- a/src/twisted/python/test/test_setup.py
+++ b/src/twisted/python/test/test_setup.py
@@ -295,6 +295,9 @@ class BuildPy3Tests(TestCase):
     Tests for L{BuildPy3}.
     """
 
+    if not _PY3:
+        skip = "BuildPy3 setuptools command used with Python 3 only."
+
     def test_find_package_modules(self):
         """
         Will filter the found modules including only the modules listed in
--- a/src/twisted/python/test/test_setup.py
+++ b/src/twisted/python/test/test_setup.py
@@ -10,10 +10,10 @@ import os
 
 
 from setuptools.dist import Distribution
-
+import twisted
 from twisted.trial.unittest import TestCase
 
-from twisted.python import _setup, dist3
+from twisted.python import _setup, dist3, filepath
 from twisted.python.compat import _PY3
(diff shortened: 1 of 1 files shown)


rodrigc:
👍 merge it

src/twisted/python/test/test_setup.py
```diff
@@ -326,26 +329,20 @@ def test_find_package_modules(self):
...
             )
-        # We are in _trial_temp which can be at PWD/_trial_temp when executed
-        # outside of tox or some tox temporary folder.
-        # When running under tox we use TOX_INI_DIR to find where the source
-        # code is located.
-        basePath = os.environ.get('PWD', '../')
-        # Overwrite the path if we are running with tox.
-        basePath = os.environ.get('TOX_INI_DIR', basePath)
-        packageDir = os.path.join(basePath, 'src', 'twisted', 'test')
+        twistedPackageDir = filepath.FilePath(twisted.__file__).parent()
```
glyph, line 332: This is definitely no worse than what was here before (and should not block this landing) but the correct, zipimport-respecting way to implement this idiom is:

``` python
from twisted.python.modules import getModule
packageDir = getModule("twisted.test").filePath.parent()
```

`ZipPath` does emulate the `.path` attribute in a way which ought to be amenable to the way zipimport works, so the rest of the test would still be correct.

markrwilliams: I thought we set `zip_safe=False` in setup.py?

glyph: We do, because people kept screwing this up, and we never got to the point where it all worked :).

glyph: Getting all the way to `zip_safe=True` is not a priority, but we also shouldn't needlessly make the situation worse; I'm just letting you know so you're aware for the future.  (This is, after all, an improvement already, since you're moving from `os.path` to `FilePath`!)

glyph:
Just waiting for appveyor now…

PR #520, opened 2016-08-24
Add assertion helpers to test log events

eeshangarg:
Hello everyone!

Sorry for taking so long to work on this (life happened!). This is a rewritten patch for [7934](https://twistedmatrix.com/trac/ticket/7934), as per @glyph's feedback in [this comment](https://twistedmatrix.com/trac/ticket/7934#comment:6). A few things I would like to say:
- I have deliberately chosen not to write any .rst documentation for these helpers, mainly because after [7983](https://twistedmatrix.com/trac/ticket/7983) (more on this below), a lot of new functionality had to be implemented and some of the existing functionality re-implemented. I don't know whether the ideas I had are worth documenting beforehand at this point and I would like some feedback and be a bit surer about the future of this patch before moving any further.
-  The changes in [7987](https://twistedmatrix.com/trac/ticket/7897) have been refactored to use these helpers as per feedback. Doing this gives us an idea of whether these helpers make testing log events more elegant or not. If not (a scenario I am well-prepared for), then there's no use to this code.
- After [7983](https://twistedmatrix.com/trac/ticket/7983), using `_loggerFor` meant that since the `Logger` that emits a certain event is created on the fly (i.e. it is not accessible as an attribute of any object) every time something needs to be logged, the `Logger` object in question is not accessible for testing. Therefore, these events needed to be tested by passing `list.append` as an observer to `GlobalLogPublisher`, see [here](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/test/test_protocol.py#L416). So I had to implement a different set of context managers and assertion helpers to test these "global" log events by passing an observer to `GlobalLogPublisher`. One of the most salient differences from the other general helpers (`assertLogEvent`, `assertLogEvents`, etc.) is that `log_logger` is not tested in the "global" helpers (since its not accessible anywhere but the event being tested against itself). On a different note, it turned out that (I'm not exactly sure why!) when `_loggerFor` is used, the emitted events are modified (by `GlobalLogPublisher` maybe?) to ensure compatibility with the old logging system, so I refactored some of [this existing code](https://github.com/twisted/twisted/blob/trunk/src/twisted/logger/_legacy.py#L46) to make sure that the relevant helpers (`assertGlobalLogEvent` and `assertGlobalLogEvents`) modified the arguments passed to them to ensure compatibility with the old-style events before making the final assertion.

Apart from all of the above, I discovered that using `_loggerFor` has a caveat (I don't know if this is just my code, @hawkowl, I would really appreciate your feedback on this). `_loggerFor` fails when logging failures. Some rough git grepping shows that `_loggerFor` has not been used to log failures anywhere within Twisted so far, so this could be something that warrants further discussion.  For instance,

``` python
class SomeObject(object):
    log = Logger()
    def someMethod(self):
        try:
            1/0
        except:
            self.log.failure("Math is hard!")
```

The above code snippet works as expected, the exception is caught and the failure is logged and can be tested using `assertLogEvent` or `assertLogEvent`, as you can see [here](https://github.com/twisted/twisted/compare/trunk...eeshangarg:7934-eeshangarg-logging-assertion-helpers?expand=1#diff-abd0faf23a054e752747fc89471425f9R310).

However, in the event that `_loggerFor` is used:

``` python
class SomeObject(object):
    def someMethod(self):
        try:
            1/0
        except:
            _loggerFor(self).failure("Math is hard!")
```

In this case, the exception isn't caught at all but is still raised (or _"reraised"_ maybe?). No event is emitted or logged, therefore, no event can be tested. The test fails at `1/0`. As you'll notice, I have left this unaddressed in my PR, mainly because I couldn't figure out a away around it and partly because I had already implemented so many things that I hadn't anticipated when I submitted the last patch to this ticket approximately 13 months ago. I didn't want to make any more changes without further review. Not to mention that, if this is a bug, it warrants a distinct ticket. :-)

Once we are finished with this ticket, I would love to help out with porting Twisted to the new logging system and writing tests for all events that are logged within the Twisted code base. Also, I would be up for working on writing documentation for these helpers and any new features I implement along the way. :-)

The rest, I leave it to you, and your fine sense of judgement! Looking forward to hearing everyone's feedback! Have a great day!

Thanks & regards,
Eeshan Garg

### DIFF
--- a/src/twisted/internet/test/test_protocol.py
+++ b/src/twisted/internet/test/test_protocol.py
@@ -420,14 +420,13 @@ class FactoryTests(TestCase):
         the L{repr} of the L{Factory} instance that is being stopped.
         """
         with capturingLogEventsFromGlobalLogPublisher() as events:
-            class MyFactory(Factory):
-                numPorts = 1
-            f = MyFactory()
+            f = Factory()
+            f.numPorts = 1
 
             f.doStop()
 
             assertGlobalLogEvent(
-                self, events[0], LogLevel.info, fullyQualifiedName(MyFactory),
+                self, events[0], LogLevel.info, fullyQualifiedName(Factory),
                 log_source=f, log_format='Stopping factory {factory!r}',
                 factory=f
             )
--- a/src/twisted/internet/test/test_protocol.py
+++ b/src/twisted/internet/test/test_protocol.py
@@ -19,7 +19,10 @@ from twisted.internet.protocol import (
     ConsumerToProtocolAdapter)
 from twisted.trial.unittest import TestCase
 from twisted.test.proto_helpers import MemoryReactorClock, StringTransport
-from twisted.logger import LogLevel, globalLogPublisher
+from twisted.logger import LogLevel
+from twisted.logger.test._assertionhelpers import (
+    assertGlobalLogEvent, capturingLogEventsFromGlobalLogPublisher)
+from twisted.python.reflect import fullyQualifiedName
 
 
 
@@ -399,18 +402,16 @@ class FactoryTests(TestCase):
         L{Factory.doStart} logs that it is starting a factory, followed by
         the L{repr} of the L{Factory} instance that is being started.
         """
-        events = []
-        globalLogPublisher.addObserver(events.append)
-        self.addCleanup(
-            lambda: globalLogPublisher.removeObserver(events.append))
+        with capturingLogEventsFromGlobalLogPublisher() as events:
+            f = Factory()
 
-        f = Factory()
-        f.doStart()
+            f.doStart()
 
-        self.assertIs(events[0]['factory'], f)
-        self.assertEqual(events[0]['log_level'], LogLevel.info)
-        self.assertEqual(events[0]['log_format'],
-                         'Starting factory {factory!r}')
+            assertGlobalLogEvent(
+                self, events[0], LogLevel.info, fullyQualifiedName(Factory),
+                log_source=f, log_format='Starting factory {factory!r}',
+                factory=f
+            )
 
 
     def test_doStopLoggingStatement(self):
@@ -418,21 +419,18 @@ class FactoryTests(TestCase):
         L{Factory.doStop} logs that it is stopping a factory, followed by
         the L{repr} of the L{Factory} instance that is being stopped.
         """
-        events = []
-        globalLogPublisher.addObserver(events.append)
-        self.addCleanup(
-            lambda: globalLogPublisher.removeObserver(events.append))
+        with capturingLogEventsFromGlobalLogPublisher() as events:
+            class MyFactory(Factory):
+                numPorts = 1
+            f = MyFactory()
 
-        class MyFactory(Factory):
-            numPorts = 1
+            f.doStop()
 
-        f = MyFactory()
-        f.doStop()
-
-        self.assertIs(events[0]['factory'], f)
-        self.assertEqual(events[0]['log_level'], LogLevel.info)
-        self.assertEqual(events[0]['log_format'],
-                         'Stopping factory {factory!r}')
+            assertGlobalLogEvent(
+                self, events[0], LogLevel.info, fullyQualifiedName(MyFactory),
+                log_source=f, log_format='Stopping factory {factory!r}',
+                factory=f
+            )
 
 
 
--- a/src/twisted/logger/test/_assertionhelpers.py
+++ b/src/twisted/logger/test/_assertionhelpers.py
@@ -1,0 +1,244 @@
+# -*- test-case-name: twisted.logger.test.test_assertionhelpers -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+General assertion helpers for testing log events.
+"""
+
+from contextlib import contextmanager
+
+from twisted.logger import globalLogPublisher
+from twisted.logger._legacy import _ensureCompatWithLegacyObservers
+
+
+
+@contextmanager
+def capturingLogEventsFrom(logger):
+    """
+    Context manager to accumulate log events emitted by a L{Logger}
+    instance by passing list.append as an observer.
+
+    @param logger: L{Logger} object that the event was emitted to.
+    @type logger: L{Logger}
+    """
+    events = []
+    logger.observer = events.append
+    if logger.source is not None:
+        logger.source._log = logger
+
+    yield events
+
+
+
+@contextmanager
+def capturingLogEventsFromGlobalLogPublisher():
+    """
+    Context manager to accumulate log events emitted by a L{Logger}
+    instance that is not directly accessible for testing, such as
+    when L{twisted.logger._loggerFor} is used to create a L{Logger}
+    on the fly. This is to avoid creating L{Logger} instance attributes in
+    the Twisted codebase itself to avoid namespace conflicts for when end
+    users subclass. In this case we don't assert log_logger in the event
+    dict and ignore it and accumulate the events in a list by passing
+    list.append as an observer to L{twisted.logger.globalLogPublisher}.
+    In the event that the L{Logger} is an instance attribute,
+    L{capturingLogEventsFrom} should be used in conjunction with
+    L{assertGlobalLogEvent} and L{assertGlobalLogEvents} for optimum test
+    coverage. Note that this is preferrably only for use in Twisted itself
+    to facilitate default logging.
+    """
+    events = []
+    globalLogPublisher.addObserver(events.append)
+
+    yield events
+
+    globalLogPublisher.removeObserver(events.append)
+
+
+
+def assertGlobalLogEvent(testCase, event, log_level, log_namespace,
+                         log_source=None, log_format=None, **kwargs):
+    """
+    Assertion helper for testing a single log event that has been
+    emitted using L{twisted.logger._loggerFor}, in which case
+    the L{Logger} that the event was emitted to may not be available
+    for testing, hence no argument for log_logger. This helper should
+    preferably be used in conjunction with the context manager
+    L{capturingLogEventsFromGlobalLogPublisher} for optimum test
+    coverage for events that can only be tested by passing an observer
+    to a L{globalLogPublisher}.
+
+    @param testCase: The L{TestCase} object that wants to test some
+        log events using L{assertLogEvent}.
+    @type testCase: L{TestCase}
+
+    @param event: The emitted event, which is a L{dict}.
+    @type event: L{dict}
+
+    @param log_level: A L{LogLevel}.
+    @type log_level: L{LogLevel}
+
+    @param log_namespace: The namespace associated with the L{Logger}.
+    @type log_namespace: L{str} (native string)
+
+    @param log_source: The source object that emitted the event. This
+        will be L{None} if the L{Logger} is not accessed as an
+        attribute of an instance or class.
+    @type log_source: L{object} or L{None}
+
+    @param log_format: The format string provided for use by observers
+        that wish to render the event as text. The format string uses
+        new-style PEP 3101 formatting and is rendered using the log
+        event (which is a L{dict}). This may be L{None}, if no format
+        string was provided. The format string is optional, except when
+        a failure is being logged using L{Logger.failure}, in which
+        case a format string describing the failure must be provided.
+    @type log_format: L{str} or L{None}
+
+    @param kwargs: Additional key/value pairs that we expect to see
+        in the event being tested. This L{dict} may contain key/value
+        pairs needed to effectively render the format string.
+    @type kwargs: L{dict}
+    """
+    expectedEvent = {
+        'log_level': log_level, 'log_source': log_source,
+        'log_format': log_format, 'log_namespace': log_namespace
+    }
+    expectedEvent.update(kwargs)
+    expectedEvent.update(log_time=event['log_time'])
+    if 'log_failure' in event:
+        expectedEvent.update(log_failure=event['log_failure'])
+
+    # Since we can't test log_logger, treat it like log_time
+    expectedEvent.update(log_logger=event['log_logger'])
+
+    # Compatibility with the old logging system, for more on this,
+    # see twisted.logger._legacy.LegacyLogObserverWrapper and
+    # twisted.logger._legacy.ensureCompatWithLegacyObservers
+    expectedEvent = _ensureCompatWithLegacyObservers(expectedEvent)
+    if event['log_format'] is not None:
+        expectedEvent.update(log_legacy=event['log_legacy'])
+
+    testCase.maxDiff = None
+
+    testCase.assertEqual(event, expectedEvent)
+
+
+
+def assertLogEvent(testCase, event, log_logger, log_level, log_namespace,
+                   log_source=None, log_format=None, **kwargs):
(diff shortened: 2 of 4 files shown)


markrwilliams:
@eeshangarg Hi!  Can you resubmit this PR with the [Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) selected?  That way we can work together on this.

eeshangarg:
Hello!

@markrwilliams: Yes, I can do that. But I am still having trouble getting tox to work on my system. I'm assuming you were the person who answered my questions on IRC about the problems I am running into?

Anyway, I installed tox 2.3.1 in a virtualenv and I still get this error http://pastebin.com/A3f6nRiA, whenever I run `tox -e py27-tests`. It would mean a lot to me if you could please help me figure out what I am doing wrong here. After that, I will resubmit the PR so the build passes and select the "allow maintainers to edit" option.

Looking forward to working on this with you!

Regards,
Eeshan

markrwilliams:
@eeshangarg Can you run these commands in your Twisted git clone and share the output:

```
ls build/py27-tests/lib/python2.7/site-packages/twisted/trial/
```

And

```
cat build/py27-tests/lib/python2.7/site-packages/twisted/_version.py
```

eeshangarg:
Hi @markrwilliams!

Output of the commands you asked me to run: http://pastebin.com/H6ivULHv

Thanks for all the help! I really appreciate it! :-)

Regards,
Eeshan

markrwilliams:
@eeshangarg Thanks for contributing to Twisted!  Could you run this command and share its output?

`/home/eeshangarg/twisted/build/py27-tests/bin/python  -c 'import twisted.trial; print(twisted.trial.__file__)'`

(Note that it should all be on one line)

eeshangarg:
Hi @markrwilliams!

Here's the output: http://pastebin.com/kDzH76dP

Also, about contributing to Twisted, I plan on keep going after this. :-)

Regards,
Eeshan

eeshangarg:
Hi @markrwilliams!

The build for this PR seems to be failing for Python 3 due to this exception https://travis-ci.org/twisted/twisted/jobs/154622677#L2858. But when I run the Python 3 tests on my branch locally, I can't reproduce the failure that the build exhibits. Instead, one of my tests fails, but it has nothing to do with not being able to import one of my modules. The test that fails on my system is failing due to differences between what `log_namespace` is on Python 2 and Python 3, which I am working to fix right now! I would really appreciate it if you could please give me some pointers as to why the build is failing with an exception that I can't replicate locally!

Thanks for all the help!

Regards,
Eeshan Garg

markrwilliams:
The version of Twisted you're developing against requires that individual modules be whitelisted in `twisted.python.dist3` for installation under Python 3.  Note that [recently](https://github.com/twisted/twisted/commit/b920626cdd07a15f9c589c2d7bf4afaee494a485) the name changed to `_dist3` and its contents from a whitelist to a _blacklist_, so this won't be necessary when you merge trunk.  Have I mentioned that you should merge in trunk? :)

Since `tox` only runs the code installed into the test environment, all of its Python 3 test environments will fail to find this module.

The solution is to add the following entries to the `testModules` list in `twisted.python.dist3`:

`"twisted.logger.test._assertionhelpers"`
`"twisted.logger.test.test_assertionhelpers"`

eeshangarg:
Hi @markrwilliams!

The tests pass on Python 3.4 and Python 2.7 now. I only got a single failing test on Python 3. It had to do with how `twisted.logger._logger._namespaceFromCallingContext` figures out the namespace for a class that is not at the module level on Python 3, for instance when a class is defined inside a test function for test purposes, as I did in `test_doStopLoggingStatement`. `twisted.python.reflect.fullyQualifiedName` is only good for getting the names of module-level objects. But on Python 3, `_namespaceFromCallingContext` treats non-module-level classes differently and actually specifies the exact scope of where the class was defined. So on Python 3, `_namespaceFromCallingContext`,  will return something like "twisted.internet.test.test_protocol.FactoryTests.test_doStopLoggingStatement.<locals>.MyFactory". On Python 2, it will just return "twisted.internet.protocol.Factory". So I just made sure that the tests passed on both versions by not making MyFactory a separate class definition inside the test itself, but by setting the attribute I needed to set directly on an instance of Factory.

The build should pass now, hopefully! I think I kinda messed up with git too. I had to rebase and push -f and I have a feeling that wasn't a good thing. I would appreciate some pointers on how to fix it, if it needs to be fixed.

Please let me know what you think!

Thanks and regards,
Eeshan

src/twisted/internet/test/test_protocol.py
```diff
@@ -19,7 +19,10 @@
     ConsumerToProtocolAdapter)
 from twisted.trial.unittest import TestCase
 from twisted.test.proto_helpers import MemoryReactorClock, StringTransport
-from twisted.logger import LogLevel, globalLogPublisher
+from twisted.logger import LogLevel
+from twisted.logger.test._assertionhelpers import (
```
markrwilliams, line 23: For posterity: the assertions are private as an answer to [point 3 from a previous review](https://twistedmatrix.com/trac/ticket/7934#comment:6), making this a "misc" change. They'll remain so until we've clearer sense of how the API works after making Twisted's own tests use them

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+    Context manager to accumulate log events emitted by a L{Logger}
+    instance by passing list.append as an observer.
+
+    @param logger: L{Logger} object that the event was emitted to.
+    @type logger: L{Logger}
+    """
+    events = []
+    logger.observer = events.append
+    if logger.source is not None:
+        logger.source._log = logger
```
markrwilliams, line 28: Is this so tests can confirm that a given event came from this logger?  It doesn't seem to be used anywhere in this patch.

Unfortunately this won't work in all cases.  Consider a class with a class-level logger that's also defined `__slots__`:

``` python
from twisted.logger.test._assertionhelpers import capturingLogEventsFrom


class HasSlots(object):
    __slots__ = ['attribute']
    log = Logger()

    def __init__(self, attribute):
        self.attribute = attribute

    def logAttribute(self):
        self.log.info("attribute: {attribute!}", attribute=self.attribute)


instance = HasSlots("value")
with capturingLogEventsFrom(instance.log):
    instance.logAttribute()
```

It looks like you could wrap the logger's `emit` [so that it always sets `log_trace=[]`](https://github.com/twisted/twisted/blob/e2f83ffc6e16358ec19d8c22a1e9f9f1e7f6031a/src/twisted/logger/_logger.py#L129), but as mentioned above this is complicated by the binding behavior of `Logger`.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+def capturingLogEventsFrom(logger):
+    """
+    Context manager to accumulate log events emitted by a L{Logger}
+    instance by passing list.append as an observer.
+
+    @param logger: L{Logger} object that the event was emitted to.
+    @type logger: L{Logger}
+    """
+    events = []
+    logger.observer = events.append
```
markrwilliams, line 26: `Logger` implements [the descriptor protocol](https://docs.python.org/2.7/howto/descriptor.html) so that each access to either enclosing class or instance results in [a unique logger instance](https://github.com/twisted/twisted/blob/trunk/src/twisted/logger/_logger.py#L88-L92).  That means this assignment won't have any effect on subsequently created loggers.  You can verify this in a Python shell:

```
>>> from twisted.logger.test.test_assertionhelpers import ThingThatLogs
>>> ThingThatLogs._log is ThingThatLogs._log
False
>>> thing = ThingThatLogs()
>>> thing._log is thing._log
False
```

The next if statement appears to work around this by reassigning the attribute on the thing to which the `Logger`'s bound.  Something like this:

```
>>> thing._log = thing._log
>>> thing._log is thing._log
True
```

There are several problems with this approach.

1) It only works if the instance (`source`) has named its `Logger` `_log`.  Consider the following:

``` python
from twisted.logger.test._assertionhelpers import capturingLogEventsFrom


class SomeClass(object):
    log = Logger()

    def logSomething(self):
        self.log.info("a message")


instance = SomeClass()
with capturingLogEventsFrom(instance.log) as events:
    instance.logSomething()

print(events)  # []
```

In fact, it will fail loudly if the class has defined slots:

``` python
class HasSlots(object):
    __slots__ = ['attribute']
    log = Logger()

    def __init__(self, attribute):
        self.attribute = attribute

    def logAttribute(self):
        self.log.info("attribute: {attribute!}", attribute=self.attribute)


instance = HasSlots("value")
with capturingLogEventsFrom(instance.log) as events:  # AttributeError: 'HasSlots' object has no attribute '_log'
    instance.logAttribute()
```
1. The instance has been permanently modified.  Even after the context manager's block completes, the observer will still be the `append` method of that particular `events` list!  That particular issue can be solved by saving and restoring the original observer, like so:

``` python
@contextmanager
def capturingLogEventsFrom(logger):
    ...
    events = []
    originalObserver = logger.observer
    logger.observer = events.append
    if logger.source is not None:
        logger.source._log = logger
    try:
        yield events
    finally:
        logger.observer = originalObserver
```

This is still problematic, though, because now `logger.source._log` is a single static instance instead of one that possesses `Logger`'s binding behavior 

1) is worse than 2) and has a simpler solution: change the signature of `capturingLogEventsFrom` to accept the instance and name of logger on that instance.  Like so:

``` python
def capturingLogEventsFrom(something, loggerName):
    logger = getattr(something, loggerName)
    ...
```

But there's a solution to 2) here as well, though it's a dirty hack.  We can circumvent the descriptor protocol by accessing the `Logger` via the instance, or its class', `__dict__`:

``` python
@contextmanager
def capturingLogEventsFrom(something, loggerName):
    logger = something.__dict__.get(
        loggerName,
        something.__class__.__dict__[loggerName])
    events = []
    logger.observer = events.append
    if logger.source is not None:
        logger.source._log = logger

    yield events
```

I'd prefer if we modified `Logger` to track is originating instance so that we could set the observer there.  I'll let other reviewers comment on this.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+    In the event that the L{Logger} is an instance attribute,
+    L{capturingLogEventsFrom} should be used in conjunction with
+    L{assertGlobalLogEvent} and L{assertGlobalLogEvents} for optimum test
+    coverage. Note that this is preferrably only for use in Twisted itself
+    to facilitate default logging.
+    """
+    events = []
+    globalLogPublisher.addObserver(events.append)
+
+    yield events
```
markrwilliams, line 54: This should be:

``` python
try:
    yield events
finally:
    globalLogPublisher.removeObserver(events.append)
```

That way if an exception occurs in the context manager's block, the observer will still be removed.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+    expectedEvent = {
+        'log_level': log_level, 'log_source': log_source,
+        'log_format': log_format, 'log_namespace': log_namespace
+    }
+    expectedEvent.update(kwargs)
+    expectedEvent.update(log_time=event['log_time'])
+    if 'log_failure' in event:
+        expectedEvent.update(log_failure=event['log_failure'])
+
+    # Since we can't test log_logger, treat it like log_time
```
markrwilliams, line 113: Is it true that you can't test it because of the binding behavior described above?  If so it'd be good to have a comment describing that.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+        case a format string describing the failure must be provided.
+    @type log_format: L{str} or L{None}
+
+    @param kwargs: Additional key/value pairs that we expect to see
+        in the event being tested. This L{dict} may contain key/value
+        pairs needed to effectively render the format string.
+    @type kwargs: L{dict}
+   """
+    expectedEvent = {
+        'log_logger': log_logger, 'log_level': log_level,
```
markrwilliams, line 170: If `capturingLogEventsFrom`'s API changes according to my API suggestions, this can no longer make an assertion about the identity of `log_logger`, because a new instance will be created on each attribute access.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+    if event['log_format'] is not None:
+        expectedEvent.update(log_legacy=event['log_legacy'])
+
+    testCase.maxDiff = None
+
+    testCase.assertEqual(event, expectedEvent)
+
+
+
+def assertLogEvent(testCase, event, log_logger, log_level, log_namespace,
```
markrwilliams, line 129: This is a very long signature!  But I'm not sure if it can be any shorter.

If `capturingLogEventsFrom'`s API changes according to my suggestion above, you can at least remove the `log_logger` argument.

src/twisted/logger/test/_assertionhelpers.py
```diff
@@ -0,0 +1,244 @@
...
+    @type log_namespace: L{str} (native string)
+
+    @param log_source: The source object that emitted the event. This
+        will be L{None} if the L{Logger} is not accessed as an
+        attribute of an instance or class.
+    @type log_source: L{object} or L{None}
+
+    @param log_format: The format string provided for use by observers
+        that wish to render the event as text. The format string uses
+        new-style PEP 3101 formatting and is rendered using the log
```
markrwilliams, line 157: Nice call out to the "Advanced String Formatting" PEP!

markrwilliams (commented):
The tests and documentation are excellent!  There are some issues with the implementation to sort out.  That's not surprising, though, because of the trickery involved in `Logger`'s implementation.

Please also note that there's [one branch of an if statement that needs coverage](https://codecov.io/gh/twisted/twisted/compare/e2f83ffc6e16358ec19d8c22a1e9f9f1e7f6031a...e0e1b304882b8067e28816ad0f75b29e879a4b46#7372632F747769737465642F6C6F676765722F746573742F5F617373657274696F6E68656C706572732E7079-110)

Thanks for all the thoughtful, hard work you've put into this.  I look forward to merging this after we sort out the implementation!

eeshangarg:
Hi @markrwilliams! Thanks for such a detailed review! I have exams this week, so I'll probably be working on this next week. My goal is to get this merged and work on porting Twisted to the new logging system and take on bigger projects as a Twisted contributor in general! Talk to you soon! :-)

Regards,
Eeshan

glyph:
Hey @eeshangarg !  I just wanted to use this today, and was sad to see that this stalled :). Any chance you'll have time to work on it soon?

eeshangarg:
@glyph: Hey! I know, I am sad I didn't get to finish this. Currently juggling a bunch of things in life and have a lot on my plate, so now might not be a good time. But I can definitely try and come back to this in a couple of weeks! However, if you find someone interested in taking over this before that, feel free to un-assign me! Thanks! :)

glyph:

This pull request hasn't been updated in quite some time, and has merge conflicts, so I'm going to close it to reduce the clutter in the PR queue.

This is not a rejection, just a clean-up; please do not hesitate to re-open a new PR with updated code when you have time.

PR #521, opened 2016-08-24
Fix AcceptEx() test on Windows

rodrigc:
https://twistedmatrix.com/trac/ticket/8780

According to the Windows API documentation for AcceptEx(): https://msdn.microsoft.com/en-us/library/windows/desktop/ms737524(v=vs.85).aspx

we need to use like this:

```
int iResult = 0;

iResult =  setsockopt( sAcceptSocket, SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, 
    (char *)&sListenSocket, sizeof(sListenSocket) );
```

We are not supposed to pass sAcceptSocket as the last parameter to setsockopt() in this case.


rodrigc:
@opalmer please review this winsock related item

opalmer:
👍 , LGTM.

PR #523, opened 2016-08-25
Add support in twisted.conch.ssh for hmac-sha2-384

acabhishek942:
see http://twistedmatrix.com/trac/ticket/8784

### DIFF
--- a/src/twisted/conch/ssh/transport.py
+++ b/src/twisted/conch/ssh/transport.py
@@ -17,7 +17,7 @@ import hmac
 import struct
 import zlib
 
-from hashlib import md5, sha1, sha256, sha512
+from hashlib import md5, sha1, sha256, sha384, sha512
 
 from cryptography.exceptions import UnsupportedAlgorithm
 from cryptography.hazmat.backends import default_backend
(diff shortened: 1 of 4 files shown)


src/twisted/conch/topfiles/8784.feature
```diff
@@ -0,0 +1 @@
+Add support in twisted.conch.ssh for hmac-sha2-384
```
adiroiban, line 1: this should be a release notes message and not a commit message.

It can be phrased as

```
twisted.conch.ssh now supports hmac-sha2-384
```

src/twisted/conch/ssh/transport.py
```diff
@@ -464,6 +465,7 @@ class SSHTransportBase(protocol.Protocol):
     supportedMACs = [
         b'hmac-sha2-512',
         b'hmac-sha2-256',
+        b'hmac-sha2-384',
```
adiroiban, line 468: I think that this should be just after hmac-sha2-512 as hmac-sha2-384 should be preferred over hmac-sha2-256 ...

see the comment about having this list based on preference

rodrigc:
Moved to https://github.com/twisted/twisted/pull/630

PR #530, opened 2016-08-28
Make Perspective Broker's NewStyleTests use IOPump

markrwilliams:
`twisted.spread.test.test_pb.NewStyleTests uses an actual TCP transport for all its tests. Instead it should use an`IOPump`like, say,`BrokerTests`.

https://twistedmatrix.com/trac/ticket/8793

This needs to be merged before I can fix https://twistedmatrix.com/trac/ticket/8792


rodrigc:
This test is failing on Python 3.3 due to garbage collection issues, similar to 
the ones I mentioned here:
https://github.com/twisted/twisted/pull/436#issuecomment-240164817

You will need to fix the garbage collection issue for this test, similar to how glyph fixed
it for https://github.com/twisted/twisted/pull/436 .

When the project drops Python 3.3 support, this will not be a problem, but until then, you need to take care of it.

markrwilliams:
@rodrigc yeah, it's because i switched over to `PBClientFactory` and i need to call the right clean up methods.  here comes the push!

markrwilliams:
@rodrigc that pep link (https://www.python.org/dev/peps/pep-0442/) was super helpful, btw!!

rodrigc:
👍 Merge it.

markrwilliams:
@rodrigc thanks for the quick turn around!

markrwilliams:
the signature for `connectServerAndClient` was `test, serverFactory, clientFactory`, but it returned `clientFactory, serverFactory, pump`.  i switch the two factories around to make it more consistent.  i'll merge when it goes green.

thanks again for the review, @rodrigc !

PR #531, opened 2016-08-29
Change 'str' to 'bytes' in deprecation message

rodrigc:
https://twistedmatrix.com/trac/ticket/8799

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -893,7 +893,7 @@ class ReactorBase(object):
                     return None
                 warnings.warn(
                     "Argument strings and environment keys/values passed to "
-                    "reactor.spawnProcess should be str, not unicode.",
+                    "reactor.spawnProcess should be bytes, not unicode.",
                     category=DeprecationWarning,
                     stacklevel=4)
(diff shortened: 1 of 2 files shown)


src/twisted/internet/base.py
```diff
@@ -893,7 +893,7 @@ def argChecker(arg):
                     return None
                 warnings.warn(
                     "Argument strings and environment keys/values passed to "
-                    "reactor.spawnProcess should be str, not unicode.",
+                    "reactor.spawnProcess should be bytes, not unicode.",
```
markrwilliams, line 896: would it be better to change "unicode" to "str" here?

rodrigc: No, because 'str' means different things on Python 2 and Python 3.

derwolfe:
LGTM. Thanks :)

src/twisted/internet/base.py
markrwilliams: if we go with `unicode` instead of native `string/unicode` generally, then it's fine here, too, and anyway this is obviously an improvement.

PR #532, opened 2016-08-31
Make Perspective Broker's NewCredTests use IOPump 

markrwilliams:
http://twistedmatrix.com/trac/ticket/8792


src/twisted/spread/test/test_pb.py
```diff
@@ -170,6 +169,52 @@ def disconnectClientFactory():
 
 
 
+@implementer(interfaces.IConnector)
+class ReconnectingConnector(object):
```
glyph: Please file a ticket for putting this into `proto_helpers`.  Given that this is a fake of a core Twisted interface, it shouldn't be squirreled away inside PB's tests.

src/twisted/spread/test/test_pb.py
```diff
@@ -170,6 +169,52 @@ def disconnectClientFactory():
...
+    @param reconnectCallback: the callback to call when C{connect} is
+        called
+    @type reconnectCallback: A callable
+
+    @param *args: Position arguments to pass C{reconnectCallback}
+    @param **kwargs: Keyword arguments to pass to C{reconnectCallback}
+    """
+
+    def __init__(self, reconnectCallback, *args, **kwargs):
+        self.reconnectCallback = reconnectCallback
```
glyph: As per https://glyph.twistedmatrix.com/2015/05/separate-your-fakes-and-your-inspectors.html we should not expose any public attributes on the test versions of stuff that are not present on the "real" versions of stuff.  These should all be `_`-prefixed.

src/twisted/spread/test/test_pb.py
glyph: Actually, there's already https://github.com/twisted/twisted/blob/c02c19441f6bd9d2431ac019ba36ac6abe33cfea/src/twisted/test/proto_helpers.py#L343.  Would it be too much trouble to refactor this to use that existing fake?

src/twisted/spread/test/test_pb.py
```diff
@@ -170,6 +169,52 @@ def disconnectClientFactory():
 
 
 
+@implementer(interfaces.IConnector)
+class ReconnectingConnector(object):
+    """
+    An L{IConnector} implementation that calls the specified callback
+    when its C{connect} method is called
+
+    @param reconnectCallback: the callback to call when C{connect} is
```
glyph: These should be `@ivar`; "param" is invalid markup for a class.

src/twisted/spread/test/test_pb.py
```diff
@@ -170,6 +169,52 @@ def disconnectClientFactory():
...
+        called
+    @type reconnectCallback: A callable
+
+    @param *args: Position arguments to pass C{reconnectCallback}
```
glyph: As above, including `*args` is invalid epytext and this should be just `@param args`.

Why isn't pydoctor catching this, I wonder?  Hmm.

src/twisted/spread/test/test_pb.py
```diff
@@ -170,6 +169,52 @@ def disconnectClientFactory():
...
+    @param **kwargs: Keyword arguments to pass to C{reconnectCallback}
```
glyph: As above, `kwargs`, no stars.

src/twisted/spread/test/test_pb.py
```diff
@@ -1434,19 +1511,34 @@ def test_loginLogout(self):
...
+        self.establishClientAndServer()
+        self.pump.flush()
+        # The perspective passed to cbLogin has gone out of scope.
+        # Ensure its __del__ runs...
+        gc.collect()
+        # ...and send its decref message to the server
+        self.pump.flush()
+        # Now allow the client to disconnect.
+        loginCompleted.callback(None)
         return d
```
glyph, line 1570: Synchronous tests should no longer return `Deferred`s.  This is what tells Twisted to run the main loop.

src/twisted/spread/test/test_pb.py
```diff
@@ -1472,23 +1564,31 @@ def logout(self):
...
+        self.establishClientAndServer()
+
+        # complete authentication
+        self.pump.flush()
+        # _PortalAuthChallenger and our Avatar should be dead by now;
+        # force a collection to trigger their __del__s
+        gc.collect()
+        # push their decref messages through
+        self.pump.flush()
         return d
```
glyph, line 1620: As above.

src/twisted/spread/test/test_pb.py
```diff
@@ -1518,10 +1617,10 @@ def cbAvatarIds(x):
...
             self.assertEqual(second, b'baz')
         d.addCallback(cbAvatarIds)
-        d.addCallback(self._disconnect, factory)
 
-        connector = reactor.connectTCP('127.0.0.1', self.portno, factory)
-        self.addCleanup(connector.disconnect)
+        self.establishClientAndServer()
+        self.pump.flush()
+
         return d
```
glyph, line 1652: As above.

src/twisted/spread/test/test_pb.py
```diff
@@ -1561,19 +1659,16 @@ def test_anonymousLogin(self):
...
 
         d.addCallback(self.assertEqual, 123)
 
-        d.addCallback(self._disconnect, factory)
-
-        connector = reactor.connectTCP("127.0.0.1", self.portno, factory)
-        self.addCleanup(connector.disconnect)
+        self.establishClientAndServer()
+        self.pump.flush()
         return d
```
glyph, line 1700: "

src/twisted/spread/test/test_pb.py
```diff
@@ -1584,18 +1679,17 @@ def test_anonymousLoginNotPermitted(self):
...
             self.assertEqual(len(errors), 1)
-            return self._disconnect(None, factory)
         d.addCallback(cleanup)
 
-        connector = reactor.connectTCP('127.0.0.1', self.portno, factory)
-        self.addCleanup(connector.disconnect)
+        self.establishClientAndServer()
+        self.pump.flush()
+
         return d
```
glyph, line 1721: "

src/twisted/spread/test/test_pb.py
```diff
@@ -1607,19 +1701,17 @@ def test_anonymousLoginWithMultipleCheckers(self):
...
 
         d.addCallback(self.assertEqual, 123)
 
-        d.addCallback(self._disconnect, factory)
+        self.establishClientAndServer()
+        self.pump.flush()
 
-        connector = reactor.connectTCP('127.0.0.1', self.portno, factory)
-        self.addCleanup(connector.disconnect)
         return d
```
glyph, line 1743: "

src/twisted/spread/test/test_pb.py
```diff
@@ -1641,10 +1732,9 @@ def cbLogin(perspective):
...
```
glyph, line 1766: "

src/twisted/spread/test/test_pb.py
```diff
@@ -1669,10 +1758,9 @@ def cbView(viewpoint):
 
         d.addCallback(self.assertTrue)
 
-        d.addCallback(self._disconnect, factory)
+        self.establishClientAndServer()
+        self.pump.flush()
 
-        connector = reactor.connectTCP("127.0.0.1", self.portno, factory)
-        self.addCleanup(connector.disconnect)
         return d
```
glyph, line 1792: "

src/twisted/spread/test/test_pb.py
markrwilliams: Switching over to `SynchronousTestCase` exposes a bug in PB's error handling.  The [gory details](http://twistedmatrix.com/trac/ticket/8792#comment:5) are in Trac, but it suffices to say that fixing PB's error handling deserves [its own ticket](http://twistedmatrix.com/trac/ticket/8803).

In the meantime, is it OK to continue to use an asynchronous test case that returns `Deferred`s?  Another alternative would be to write a local helper that acts like `t.t_async.TestCase.assertFailure` and use it instead of `failureResultOf`, but it seems like a bad idea to heap a temporary work around into this PR.

**Addendum**: Using `t.t._async.TestCase` with an `IOPump` _is_ silly, but it does result in a version of these tests that pass under PyPy.  So it has some merit!

glyph:
Hopefully the build gods will be kinder to us with this trunk update, and I will land forthwith.

glyph:
I'm going to use admin powers to override 😢  because both build failures are intermittent issues with @Lukasa's Microsoft™ HomePage™ Server™ (i.e. with fetching the URL https://python-hyper.org/h2/en/stable/objects.inv).  We should _really_ fix that issue, and check in a fallback `objects.inv`.

markrwilliams:
@glyph thanks!  and [oh look there's a new ticket](https://twistedmatrix.com/trac/ticket/8823) now that this one's closed, proving that the number of open tickets, like energy, is always conserved.

PR #533, opened 2016-08-31
Add EC support to t.conch.ssh.keys

acabhishek942:
see http://twistedmatrix.com/trac/ticket/8798


src/twisted/conch/ssh/keys.py
```diff
@@ -198,6 +240,16 @@ def _fromString_BLOB(cls, blob):
...
             )
+        elif keyType in [b'ecdsa-sha2-' + curve for curve in list(_curveTable.keys())]:
+            x, y, rest = common.getMP(rest, 2)
+            newKey =  cls(
+                ec.EllipticCurvePublicNumbers(
+                    x, y,
+                    _curveTable[keyType.split(b'-')[2]]
+                    ).public_key(default_backend())
+                )
+            newKey.ecKeyName = keyType
```
glyph: You're adding a new attribute here, `ecKeyName`, but it's not documented.  Better would be if it's not public!  This is not information that anyone should be using externally, as far as I can tell.

src/twisted/conch/ssh/keys.py
```diff
@@ -863,21 +987,42 @@ def type(self):
...
             return 'DSA'
+        elif isinstance(
+                self._keyObject, (ec.EllipticCurvePublicKey, ec.EllipticCurvePrivateKey)):
+            return 'EC'
         else:
             raise RuntimeError(
                 'unknown type of object: %r' % (self._keyObject,))
 
 
+    def getECKeyName(self):
```
glyph: I don't understand the purpose of this method; I think it should probably be removed.  Nothing external calls it - and nothing should.  The purpose of `Key` is to present a uniform interface for the operations supported by all key types.

src/twisted/conch/ssh/keys.py
```diff
@@ -863,21 +987,42 @@ def type(self):
...
-        or b'ssh-dss'.
+        defined in RFC 4253, Section 6.6. Currently this can only be b'ssh-rsa',
+        b'ssh-dss' and ecdsa-sha2-*.
 
         @return: The key type format.
         @rtype: L{bytes}
         """
-        return {'RSA': b'ssh-rsa', 'DSA': b'ssh-dss'}[self.type()]
+        if self.type() == 'EC':
+            return self.getECKeyName()
```
glyph: The logic for `getECKeyName` should probably be inline here.

src/twisted/conch/ssh/keys.py
```diff
@@ -932,6 +1079,22 @@ def data(self):
                 "p": numbers.public_numbers.parameter_numbers.p,
                 "q": numbers.public_numbers.parameter_numbers.q,
             }
+        elif isinstance(self._keyObject, ec.EllipticCurvePublicKey):
+            numbers = self._keyObject.public_numbers()
+            return{
+                "x": numbers.x,
+                "y": numbers.y,
+                "curve": self.getECKeyName(),
```
glyph: This could just call `self.sshType()`.

src/twisted/conch/ssh/keys.py
```diff
@@ -932,6 +1079,22 @@ def data(self):
...
+            }
+        elif isinstance(self._keyObject, ec.EllipticCurvePrivateKey):
+            numbers = self._keyObject.private_numbers()
+            return {
+                "x": numbers.public_numbers.x,
+                "y": numbers.public_numbers.y,
+                "private_value": numbers.private_value,
+                "curve": self.getECKeyName(),
```
glyph: Same here.

src/twisted/conch/ssh/keys.py
```diff
@@ -1198,7 +1379,18 @@ def sign(self, data):
             ret = common.NS(int_to_bytes(r, 20) + int_to_bytes(s, 20))
 
         else:
-            raise BadKeyError("unknown key type %s" % (self.type(),))
+            # Hash size depends on key size
```
glyph, line 1353: This should probably be `elif keyType == 'EC'` for clarity, even if that previous `else` block was just dead code :)

src/twisted/conch/ssh/keys.py
```diff
@@ -1241,8 +1434,20 @@ def verify(self, signature, data):
                 k = k.public_key()
             verifier = k.verifier(
                 signature, hashes.SHA1())
+
         else:
-            raise BadKeyError("unknown key type %s" % (self.type(),))
+            k = self._keyObject
```
glyph, line 1410: Same here, regarding `elif keyType == 'EC'`.

src/twisted/conch/ssh/keys.py
```diff
@@ -240,6 +301,12 @@ def _fromString_PRIVATE_BLOB(cls, blob):
         elif keyType == b'ssh-dss':
             p, q, g, y, x, rest = common.getMP(rest, 5)
             return cls._fromDSAComponents(y=y, g=g, p=p, q=q, x=x)
+        elif keyType in [b'ecdsa-sha2-' + curve for curve in list(_curveTable.keys())]:
+            x, y, private_value, rest = common.getMP(rest, 3)
+            newKey = cls._fromECComponents(x=x, y=y, curve=keyType,
+                private_value=private_value)
+            newKey.ecKeyName = keyType
```
glyph: Might it be a good idea to include `ecKeyName` in the arguments to `_fromECComponents`?  It seems like the construction is incomplete without it.

src/twisted/conch/ssh/keys.py
```diff
@@ -225,6 +277,15 @@ def _fromString_PRIVATE_BLOB(cls, blob):
             integer y
             integer x
 
+        EC keys::
+            string 'ecdsa-sha2-[identifier]'
+            integer x
+            integer y
+            integer private_value
+
+            identifier is the standard NIST curbe name.
```
glyph: "curve", presumably?

src/twisted/conch/ssh/keys.py
```diff
@@ -298,7 +373,7 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
             * a passphrase is not provided for an encrypted key
         """
         lines = data.strip().split(b'\n')
-        kind = lines[0][11:14]
+        kind = lines[0][11:14].replace(b' ', b'')
```
glyph: Just curious here - why is stripping spaces necessary where it wasn't before?

src/twisted/conch/ssh/keys.py
```diff
@@ -390,6 +465,17 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
                     )
                 ).private_key(backend=default_backend())
             )
+        elif kind == b'EC':
+            newKey = cls(load_pem_private_key(data, passphrase, default_backend()))
+
+            curve = _oidTable[str(decodedKey[2]).encode('utf-8')]
+            # Reverse look up the nist curve name to be referenced later
+            for k,v in _curveTable.items():
```
glyph: Just a coding-standard nitpick: there should be a space after the comma in this line.  Sorry the automatic checker didn't pick that up.

src/twisted/conch/ssh/keys.py
```diff
@@ -390,6 +465,17 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
...
+                if isinstance(curve, v.__class__):
+                    newKey.ecKeyName = b'ecdsa-sha2-' + k
```
glyph: Definitely feeling like this variable ought to be required for the construction of an EC key :-).  Having this set several lines away from `newKey` feels like a bug magnet.

glyph: From `newKey` being constructed, that is.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +710,39 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
         return cls(keyObject)
 
 
+    @classmethod
+    def _fromECComponents(cls, x, y, curve, private_value=None):
+        """
+        Build a key from EC components.
+
+        @type x: L{int}
```
glyph: Just a minor style thing - we usually put `@param` (or `@ivar`, etc) before `@type`.  It's not a required part of the coding standard but it's probably best to do it just for consistency.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +710,39 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
         return cls(keyObject)
 
 
+    @classmethod
+    def _fromECComponents(cls, x, y, curve, private_value=None):
```
glyph: Twisted coding standard says that this variable should be called `privateValue`, not `private_value`.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +710,39 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
...
+        """
+
+        publicNumbers = ec.EllipticCurvePublicNumbers(
+            x=x, y=y, curve=_curveTable[curve.split(b'-')[2]])
+        if private_value is None:
+            # We have public components.
+            keyObject = publicNumbers.public_key(default_backend())
+        else:
+            privateNumbers = ec.EllipticCurvePrivateNumbers(
+                private_value= private_value, public_numbers= publicNumbers)
```
glyph: There should be no spaces after the equal signs here.  I could swear that `twistedchecker` should report this.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +710,39 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
...
+            keyObject = privateNumbers.private_key(default_backend())
+
+        keyObject.ecKeyName = curve
```
glyph: Another `ecKeyName` that should have been passed to a constructor someplace :).

glyph: Wait a second - if this is set by `_fromECComponents`, is this actually necessary to set elsewhere?

src/twisted/conch/ssh/keys.py
```diff
@@ -955,6 +1118,13 @@ def blob(self):
             integer g
             integer y
 
+        EC keys::
+            string 'ecdsa-sha2-[identifier]'
+            integer x
+            integr y
```
glyph: "integer"?

src/twisted/conch/test/test_keys.py
```diff
@@ -188,6 +206,32 @@ def setUp(self):
     def tearDown(self):
         os.unlink(self.keyFile)
 
+    def test_size(self):
+        """
+        Test that size method returns the correct size.
```
glyph: The twisted standard for test docstrings is derived from https://jml.io/test-docstrings/.  In short: please don't use phrases like "test that" or "verify that", or the adjective "correct".  Just say `The .size() method returns the size, in bits, of the Key object.` or something like that.

src/twisted/conch/test/test_keys.py
```diff
@@ -188,6 +206,32 @@ def setUp(self):
...
+        self.assertEqual(keys.Key(self.rsaObj).size(), 768)
+        self.assertEqual(keys.Key(self.dsaObj).size(), 1024)
+        self.assertEqual(keys.Key(self.ecObj).size(), 256)
+        self.assertEqual(keys.Key(self.ecObj384).size(), 384)
+        self.assertEqual(keys.Key(self.ecObj521).size(), 521)
+
+
+    def test_getECKeyName(self):
+        """
+        Test that getECKeyName works properly.
```
glyph: See test docstring feedback above.

src/twisted/conch/test/test_keys.py
```diff
@@ -564,6 +627,14 @@ def test_notEqual(self):
         self.assertTrue(rsa1 != None)
 
 
+    def test_dataError(self):
+        """
+        Error test for data method
```
glyph: See test docstring feedback above.

src/twisted/conch/test/test_keys.py
```diff
@@ -682,6 +758,25 @@ def test_fromBlobDSA(self):
         self.assertTrue(dsaKey.isPublic())
         self.assertEqual(dsaPublicData, dsaKey.data())
 
+    def test_fromBlobECDSA(self):
+        """
+        A public ECDSA key is correctly generated from a public key blob.
```
glyph: Here, since you're not saying "test that", it's a bit more subtle; "`Key.fromString` loads elliptic curve keys from blobs", perhaps?

src/twisted/conch/ssh/keys.py
acabhishek942: This is required for EC keys,  `-----BEGIN EC PRIVATE KEY-----[11:14]` gives `EC`(extra space). It wasn't required before because `RSA` and `DSA` are of length 3

src/twisted/conch/ssh/keys.py
```diff
@@ -863,21 +942,27 @@ def type(self):
...
+        defined in RFC 4253, Section 6.6. Currently this can only be b'ssh-rsa',
+        b'ssh-dss' and ecdsa-sha2-*.
 
         @return: The key type format.
         @rtype: L{bytes}
         """
-        return {'RSA': b'ssh-rsa', 'DSA': b'ssh-dss'}[self.type()]
+
+        return {'RSA': b'ssh-rsa', 'DSA': b'ssh-dss',
+            'EC': b'ecdsa-sha2-nistp' + str(self.size()).encode('utf-8')}[self.type()]
```
the0id: I wouldn't recommend this.  It makes things less complex, but unnecessarily limits the types of keys supported to the bare minimum.

src/twisted/conch/ssh/keys.py
```diff
@@ -18,7 +18,10 @@
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
-from cryptography.hazmat.primitives.asymmetric import dsa, rsa, padding
+from cryptography.hazmat.primitives.asymmetric import dsa, rsa, padding, ec
+from cryptography.hazmat.primitives.serialization import (load_pem_private_key,
+                                            load_ssh_public_key)
```
Lukasa: What's up with the indentation here? This seems to be indented a long way, but not in line with anything else.

src/twisted/conch/ssh/keys.py
```diff
@@ -198,6 +241,14 @@ def _fromString_BLOB(cls, blob):
                     )
                 ).public_key(default_backend())
             )
+        elif keyType in [b'ecdsa-sha2-' + curve for curve in list(_curveTable.keys())]:
+            x, y, rest = common.getMP(rest, 2)
+            return cls(
+                ec.EllipticCurvePublicNumbers(
+                    x, y,
+                    _curveTable[keyType.split(b'-')[2]]
```
Lukasa: Can this get hoisted out? Lines are a consumable resource, as are variable names, and it'd be nice to have this on a separate line for debugging/understandability purposes.

src/twisted/conch/ssh/keys.py
```diff
@@ -258,6 +322,9 @@ def _fromString_PUBLIC_OPENSSH(cls, data):
         @rtype: L{twisted.conch.ssh.keys.Key}
         @raises BadKeyError: if the blob type is unknown.
         """
+
+        if data.startswith(b'ecdsa-sha2'):
+            return cls(load_ssh_public_key(data, default_backend()))
```
Lukasa, line 328: I think we need a comment here explaining why this `if` block is present. I have no idea why we can't fall into the standard `fromString_BLOB` path, but I'm assuming there's a good reason. Let's put that in a comment.

src/twisted/conch/ssh/keys.py
Lukasa: I'm not convinced by this line of argument. I'd rather see us adjust the logic than to make a somewhat arbitrary change than this. In this case, should we not strip 11 characters from the front and 17 from the back? e.g. change this to:

``` python
lines[0][11:-17]
```

That will handle the possibility that sometime down the road we get something like `-----BEGIN LUKASA'S BEST CRYPTO ALGORITHM PRIVATE KEY-----`.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +696,37 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
...
+        Build a key from EC components.
+
+        @param x: The affine x component of the public point used for verifying.
+        @type x: L{int}
+
+        @param y: The affine y component of the public point used for verifying.
+        @type y: L{int}
+
+        @param curve: NIST name of elliptic curve.
+        @type curve: L{str}
```
Lukasa: This is not of type `str`. Lower down in the code you call `curve.split(b'-')`. Definitionally, then, this is of type `bytes`.

src/twisted/conch/ssh/keys.py
```diff
@@ -624,6 +696,37 @@ def _fromDSAComponents(cls, y, p, q, g, x=None):
...
+
+        @param privateValue: The private value.
+        @type privateValue: L{int}
+        """
+
+        publicNumbers = ec.EllipticCurvePublicNumbers(
+            x=x, y=y, curve=_curveTable[curve.split(b'-')[2]])
+        if privateValue is None:
+            # We have public components.
+            keyObject = publicNumbers.public_key(default_backend())
```
Lukasa, line 722: Can we write a test to hit this LoC please?

src/twisted/conch/ssh/keys.py
```diff
@@ -853,7 +957,7 @@ def fingerprint(self, format=FingerprintFormats.MD5_HEX):
     def type(self):
         """
         Return the type of the object we wrap.  Currently this can only be
-        'RSA' or 'DSA'.
+        'RSA', 'DSA' and 'EC'.
```
Lukasa: (Nit) Should be `or`, not `and`.

src/twisted/conch/ssh/keys.py
```diff
@@ -863,21 +967,30 @@ def type(self):
...
 
 
+
     def sshType(self):
         """
         Get the type of the object we wrap as defined in the SSH protocol,
-        defined in RFC 4253, Section 6.6. Currently this can only be b'ssh-rsa'
-        or b'ssh-dss'.
+        defined in RFC 4253, Section 6.6. Currently this can only be b'ssh-rsa',
+        b'ssh-dss' and ecdsa-sha2-[identifier].
```
Lukasa: (nit) should be `b'ecda-sha2-[identifier]'`.

src/twisted/conch/ssh/keys.py
```diff
@@ -863,21 +967,30 @@ def type(self):
...
+        identifier is the standard NIST curve name
 
         @return: The key type format.
         @rtype: L{bytes}
         """
-        return {'RSA': b'ssh-rsa', 'DSA': b'ssh-dss'}[self.type()]
+
+        return (b'ecdsa-sha2-' +
+        _secToNist[self._keyObject.curve.name.encode('utf-8')] if self.type() == 'EC'
+        else {'RSA': b'ssh-rsa', 'DSA': b'ssh-dss'}[self.type()])
```
Lukasa: There's no prize for getting this into one gigantic conditional. Let's split this out into an `if` statement: it'll be substantially easier to read.

src/twisted/conch/ssh/keys.py
```diff
@@ -932,6 +1047,22 @@ def data(self):
...
+                "curve": self.sshType(),
+            }
+        elif isinstance(self._keyObject, ec.EllipticCurvePrivateKey):
+            numbers = self._keyObject.private_numbers()
+            return {
+                "x": numbers.public_numbers.x,
+                "y": numbers.public_numbers.y,
+                "privateValue": numbers.private_value,
+                "curve": self.sshType(),
+
```
Lukasa: Remove this blank line please.

src/twisted/conch/ssh/keys.py
```diff
@@ -967,7 +1105,8 @@ def blob(self):
                     common.MP(data['q']) + common.MP(data['g']) +
                     common.MP(data['y']))
         else:
```
Lukasa, line 1108: For unknown key types this used to raise an exception. I think it should probably still do that. Let's validate that we have the key type in question in our hands.

src/twisted/conch/ssh/keys.py
```diff
@@ -1006,7 +1153,8 @@ def privateBlob(self):
                     common.MP(data['q']) + common.MP(data['g']) +
                     common.MP(data['y']) + common.MP(data['x']))
         else:
```
Lukasa, line 1156: As above, this used to error on unknown key types, now it doesn't (or rather it errors weirdly). Can we re-establish that old behaviour?

src/twisted/conch/ssh/keys.py
```diff
@@ -1197,8 +1346,19 @@ def sign(self, data):
...
+            if keySize <= 256:
+                hashSize = hashes.SHA256()
+            elif keySize <= 384:
+                hashSize = hashes.SHA384()
+            else:
+                hashSize = hashes.SHA512()
+            signer = self._keyObject.signer(ec.ECDSA(hashSize))
+            signer.update(data)
+            ret = common.NS(signer.finalize())
+
```
Lukasa, line 1364: Again, we've lost our `BadKeyError` fallback. Can we replace it?

src/twisted/conch/ssh/keys.py
```diff
@@ -1241,8 +1402,20 @@ def verify(self, signature, data):
...
+                k=k.public_key()
+            keySize = self.size()
+            if keySize <= 256:
+                hashSize = hashes.SHA256()
+            elif keySize <= 384:
+                hashSize = hashes.SHA384()
+            else:
+                hashSize = hashes.SHA512()
+            verifier = k.verifier(
+                common.getNS(signature)[0], ec.ECDSA(hashSize))
```
Lukasa, line 1421: Again, we've lost our `BadKeyError` fallback: can we re-add it?

src/twisted/conch/ssh/keys.py
```diff
@@ -1241,8 +1402,20 @@ def verify(self, signature, data):
                 k = k.public_key()
             verifier = k.verifier(
                 signature, hashes.SHA1())
-        else:
-            raise BadKeyError("unknown key type %s" % (self.type(),))
+
+        elif keyType == 'EC':
+            k = self._keyObject
+            if not self.isPublic():
+                k=k.public_key()
```
Lukasa: Can we please hit this line of code?

src/twisted/conch/ssh/keys.py
```diff
@@ -1241,8 +1402,20 @@ def verify(self, signature, data):
...
+            keySize = self.size()
+            if keySize <= 256:
+                hashSize = hashes.SHA256()
+            elif keySize <= 384:
+                hashSize = hashes.SHA384()
+            else:
+                hashSize = hashes.SHA512()
```
Lukasa, line 1419: I'm a little perplexed by this block of code. Care to put a comment above it explaining why we choose the hash size this way?

Lukasa (changes requested):
This is a good start @acabhishek942! I have a few issues and concerns about the state of the patch, but they're all entirely workable: they relate mostly to the fallbacks for unexpected and unknown key types and wanting a little extra commenting and justification for choices that are not immediately transparent to me.

Note that, while I can eventually give a positive review, I cannot really justify merging this myself because I fundamentally don't understand SSH or conch that well. That means that I'm a reviewer of limited use: I can see some things that are wrong, but just because I don't see anything wrong doesn't mean that there isn't something that needs addressing. I'm happy to do some reviews, but eventually you'll need someone who understands conch better than me to do a final review. :smile:

src/twisted/conch/ssh/keys.py
```diff
@@ -322,7 +317,7 @@ def _fromString_PUBLIC_OPENSSH(cls, data):
         @rtype: L{twisted.conch.ssh.keys.Key}
         @raises BadKeyError: if the blob type is unknown.
         """
-
+        #decoding and splitting messes up ecdsa keys.
```
Lukasa: Ok, so the reasoning for why seems to be that `decodebytes` is actually doing a base64 decode, which is not what is expected for ECDSA keys. It'd be nice if we could say that more clearly.

src/twisted/conch/ssh/keys.py
```diff
@@ -1241,8 +1396,20 @@ def verify(self, signature, data):
                 k = k.public_key()
             verifier = k.verifier(
                 signature, hashes.SHA1())
-        else:
-            raise BadKeyError("unknown key type %s" % (self.type(),))
+
+        elif keyType == 'EC':
```
Lukasa: Ok, so now I see that we can't fall out the bottom of this because `type()` raises exceptions for unknown types. In that case, let's not bother with the `elif` here: there's no point in having the conditional because there can't be anything else. You should use `else` and `assert` the value instead. That applies to all of these cases.

Lukasa: Note it'll also solve your branch coverage issue.

Lukasa: To be clear, @glyph hates `assert` so he may want to do something else.

Lukasa (dismissed):
Ok, I've clarified my understanding about key types. I have left a further suggestion.

src/twisted/conch/ssh/keys.py
acabhishek942: I added elif with @glyph's suggestion here. https://github.com/twisted/twisted/pull/533#discussion-diff-78707127.

src/twisted/conch/ssh/keys.py
```diff
@@ -955,9 +1081,16 @@ def blob(self):
...
+            string 'ecdsa-sha2-[identifier]'
+            integer x
+            integer y
+
+            identifier is the standard NIST curve name
+
         @rtype: L{bytes}
         """
-        type = self.type()
+        type = self.type() #takes care of bad key type.
```
glyph: I realize that you added these comments as a result of @Lukasa's feedback, but this isn't a great way to address that concern.  We don't put comments on every function call everywhere indicating what exceptions might be raised :).  Instead, documenting this on `type` itself, and replying to @Lukasa's comments in-line, would be better.

glyph: Sorry, I forgot to lead with the actionable feedback there: please delete this comment.

src/twisted/conch/ssh/keys.py
```diff
@@ -993,8 +1127,16 @@ def privateBlob(self):
...
+        EC keys::
+            string 'ecdsa-sha2-[identifier]'
+            integer x
+            integer y
+            integer privateValue
+
+            identifier is the NIST standard curve name.
         """
-        type = self.type()
+        type = self.type() #takes care of bad key type.
```
glyph: Please delete this comment.

src/twisted/conch/ssh/keys.py
```diff
@@ -1222,7 +1377,8 @@ def verify(self, signature, data):
             signatureType, signature = common.getNS(signature)
         if signatureType != self.sshType():
             return False
-        if self.type() == 'RSA':
+        keyType = self.type() #takes care of bad key type
```
glyph: Please delete this comment.

src/twisted/conch/ssh/keys.py
glyph: There really ought to be spaces around the `=` sign, as well :). I don't know why txchecker is not complaining here...

src/twisted/conch/ssh/keys.py
glyph: My understanding is that the `BadKeyError()` was never actually raised before, since `type()` _always_ raised a `RuntimeError`?

src/twisted/conch/test/test_keys.py
```diff
@@ -167,6 +167,24 @@ def setUp(self):
             g=keydata.DSAData['g'],
             x=keydata.DSAData['x'],
             )._keyObject
+        self.ecObj = keys.Key._fromECComponents(
+            x=keydata.ECDatanistp256['x'],
+            y=keydata.ECDatanistp256['y'],
+            privateValue=keydata.ECDatanistp256['privateValue'],
+            curve=keydata.ECDatanistp256['curve']
+            )._keyObject
```
glyph: Minor nitpick: we usually un-indent on this line.

src/twisted/conch/test/test_keys.py
```diff
@@ -564,6 +612,14 @@ def test_notEqual(self):
         self.assertTrue(rsa1 != None)
 
 
+    def test_dataError(self):
+        """
+        The data() method raises RuntimeError for bad keys.
```
glyph: Minor nitpick: we usually use epytext syntax to do references in test names, so rather than the ambiguous "the data() method", we'd say, `L{keys.Key.data} raises RuntimeError for bad keys`.

src/twisted/conch/ssh/keys.py
glyph: This comment still hasn't been addressed; codecov shows it as uncovered: https://codecov.io/gh/twisted/twisted/compare/620b84494a0d59309d27c1228bc3886771051a32...b78e520b5f7e196b4a45ba649042918744b56d48#7372632F747769737465642F636F6E63682F7373682F6B6579732E7079-1404

src/twisted/conch/ssh/keys.py
glyph: What should _really_ be happening here - although this refactoring should happen in a follow-up branch, we need not block on it - is that this code should be doing `self._keyImpl.verify(...)`, and then we should have 3 different classes for `_keyImpl`, which wrap each of the different key types.  A class with this many if/elif trees is a code smell.

In the meanwhile, to compromise between coverage and quality, a [`# pragma: no branch` comment on the last `elif`](http://coverage.readthedocs.io/en/coverage-4.2/branch.html#structurally-partial-branches) ought to resolve the conflict between coverage and clarity.  Please file a ticket to refactor the guts of the if/elif trees out into their own separate classes though, and link to it from a subsequent comment, so we don't lose track of this refactoring.

glyph (commented):
This does still need its code coverage issues addressed, and `type` should be documented to `@raise RuntimeError` rather than the comments.  Almost there though :).  Thanks again for your contributions!

src/twisted/conch/ssh/keys.py
acabhishek942: Sorry I missed this earlier. I addressed this comment and also added tests in `signAndVerify` using private RSA and DSA keys, even though it is not related to this ticket but helps with the project coverage.

src/twisted/conch/ssh/keys.py
acabhishek942: the ticket is here https://twistedmatrix.com/trac/ticket/8854

src/twisted/conch/ssh/keys.py
acabhishek942: Yes, that's why I removed that part because that was dead code never executed.

src/twisted/conch/ssh/keys.py
```diff
@@ -3,7 +3,7 @@
 # See LICENSE for details.
 
 """
-Handling of RSA and DSA keys.
+Handling of RSA, DSA and EC keys.
```
hawkowl: We use the Oxford comma (so this should be "RSA, DSA, and EC keys").

src/twisted/conch/ssh/keys.py
```diff
@@ -854,31 +954,41 @@ def fingerprint(self, format=FingerprintFormats.MD5_HEX):
     def type(self):
         """
         Return the type of the object we wrap.  Currently this can only be
-        'RSA' or 'DSA'.
+        'RSA', 'DSA' or 'EC'.
```
hawkowl: Likewise, with the Oxford comma.

src/twisted/conch/ssh/keys.py
```diff
@@ -795,7 +894,8 @@ def isPublic(self):
         @return: C{True} if this is a public key.
         """
         return isinstance(
```
hawkowl, line 900: Could we get test coverage for this function?

src/twisted/conch/ssh/keys.py
```diff
@@ -1198,8 +1343,19 @@ def sign(self, data):
             # Make sure they are padded out to 160 bits (20 bytes each)
             ret = common.NS(int_to_bytes(r, 20) + int_to_bytes(s, 20))
 
-        else:
-            raise BadKeyError("unknown key type %s" % (self.type(),))
+        elif keyType == 'EC': # pragma: no branch
```
hawkowl, line 1352: Could you remove the pragma statement? We don't add coverage exceptions to Twisted code.

src/twisted/conch/test/test_keys.py
```diff
@@ -167,6 +167,24 @@ def setUp(self):
...
+            y=keydata.ECDatanistp256['y'],
+            privateValue=keydata.ECDatanistp256['privateValue'],
+            curve=keydata.ECDatanistp256['curve']
+            )._keyObject
+        self.ecObj384 = keys.Key._fromECComponents(
+            x=keydata.ECDatanistp384['x'],
+            y=keydata.ECDatanistp384['y'],
+            privateValue=keydata.ECDatanistp384['privateValue'],
+            curve=keydata.ECDatanistp384['curve']
+            )._keyObject
```
hawkowl: Seconding the nitpick here, the )s should be one back.

src/twisted/conch/topfiles/8798.feature
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh.keys supports EC keys
```
hawkowl: We support ECDSA keys now, right? We should probably mention that specifically.

hawkowl (approved):
Looks good, a few nitpicks. Fix them up as you see fit, and cc me when it's ready.

src/twisted/conch/ssh/keys.py
acabhishek942: @hawkowl the reason for pragma: no branch is here

acabhishek942:
@hawkowl Can you please review this?

PR #534, opened 2016-08-31
Port twisted.words.protocols.irc to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/6320


src/twisted/words/test/test_irc.py
```diff
@@ -9,13 +9,27 @@
...
 from twisted.trial.unittest import TestCase
 from twisted.words.protocols import irc
 from twisted.words.protocols.irc import IRCClient, attributes as A
-from twisted.internet import protocol, task
-from twisted.test.proto_helpers import StringTransport, StringIOWithoutClosing
-from twisted.python.filepath import FilePath
+
+
+
+def assertEqualBufferValue(bufferValue, val):
```
glyph: As per https://twisted.readthedocs.io/en/twisted-16.4.1/core/development/policy/coding-standard.html#docstrings - "all methods, functions, classes, and modules must have docstrings".  So this should have a docstring :).  The docstring should describe its behavior, the type and purpose of `bufferVal` and `val` parameters.

src/twisted/words/test/test_irc.py
```diff
@@ -9,13 +9,27 @@
...
+    # A buffer is always bytes, but sometimes
+    # we need to compare it to a utf-8 unicode string
+    if isinstance(val, unicode):
+        bufferValue = bufferValue.decode("utf-8")
+
+    if isinstance(bufferValue, list):
+        if isinstance(val[0], unicode):
+            bufferValue = [buf.decode("utf8") for buf in bufferValue]
+    assert bufferValue == val, "%s != %s" % (bufferValue, val)
```
glyph: Assertions are generally frowned upon in Twisted, because [they aren't always run; `python -OO` disables them](https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement).  This should be using `assertEqual`; which means this function should take a `TestCase` object as well.

src/twisted/words/test/test_irc.py
```diff
@@ -1634,7 +1644,13 @@ def setUp(self):
 
 
     def check(self, s):
-        self.assertEqual(self.f.getvalue(), s)
+        bufferValue = self.f.getvalue()
+        if isinstance(s, unicode):
+            # Convert bufferValue from bytes
```
glyph: Try to avoid using comments; instead, put this information into the docstring.  (i.e.: add a docstring explaining this method.)

src/twisted/words/test/test_irc.py
```diff
@@ -9,13 +9,27 @@
 import operator
 import time
 
+from twisted.internet import protocol, task
```
glyph, line 12: There are lots and lots of string constants in this module and it _might_ be helpful to have them be a consistent type between 2/3; a `from __future__ import unicode_literals` might catch potential issues.

src/twisted/words/test/test_irc.py
```diff
@@ -792,14 +806,14 @@ def test_support_CHANMODES(self):
         self._testFeatureDefault('CHANMODES', [('CHANMODES', 'b,,lk,ha,ha')])
 
         self.assertEqual(
-            self._parseFeature('CHANMODES', ''),
+            self._parseFeature('CHANMODES', ',,,'),
```
glyph, line 816: Why is this change relevant to the 2->3 port?  These values really ought to be python-version independent...

glyph (commented):
Thanks for doing this, @rodrigc.  It actually looks pretty good, except for the coverage issue.  Thankfully, the codecov report looks accurate this time: https://codecov.io/gh/twisted/twisted/compare/9fd621c0d425661bf23332cd1692ed256aa2e0b9...7eba1e677c22e36cf37141a98453cbf4b036335b#7372632F747769737465642F776F7264732F696D2F697263737570706F72742E7079-40 the issues are almost entirely in `ircsupport.py`.  One way forward would be to simply push those changes into a separate PR first, since aside from minor docstring issues, everything else here is ready to land.  ircsupport should be a pretty separate module, as well.

Can you address the docstring and coverage issues somehow and resubmit?  Thanks!

src/twisted/words/test/test_irc.py
rodrigc: This is needed.  Without this change, the test will fail on Python 3 like this:

```
[FAIL]
  File "/Users/crodrigues/twisted13/src/twisted/words/test/test_irc.py", line 821, in test_support_CHANMODES
    'noParam': ''})
  File "/Users/crodrigues/twisted13/src/twisted/trial/_synctest.py", line 425, in assertEqual
  File "/usr/local/Cellar/python3/HEAD-42a10253d066_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 820, in assertEqual
  File "/usr/local/Cellar/python3/HEAD-42a10253d066_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 1111, in assertDictEqual
twisted.trial.unittest.FailTest: {'addressModes': ''} != {'param': '', 'noParam': '', 'addressModes': '', 'setParam': ''}
- {'addressModes': ''}
+ {'addressModes': '', 'noParam': '', 'param': '', 'setParam': ''}

twisted.words.test.test_irc.ServerSupportedFeatureTests.test_support_CHANMODES
```

The underlying _parseChanModesParam() function is using a combination of map, lambda, and dict which is triggering this difference in behavior between 2 and 3.

src/twisted/words/protocols/irc.py
```diff
@@ -2620,9 +2622,17 @@ def connectionMade(self):
...
     def dataReceived(self, data):
-        basic.LineReceiver.dataReceived(self, data.replace('\r', ''))
+        data = data.replace('\r', '')
+        if isinstance(data, unicode):
+            data = data.encode("utf-8")
+        basic.LineReceiver.dataReceived(self, data)
+
 
     def lineReceived(self, line):
+        if isinstance("", unicode) and isinstance(line, bytes):
```
markrwilliams: `if isinstance("", unicode)` is a much better check than `if _PY3`

src/twisted/words/protocols/irc.py
```diff
@@ -2620,9 +2622,17 @@ def connectionMade(self):
...
+            # decode bytes from transport to unicode
+            line = line.decode("utf-8")
```
markrwilliams, line 2637: See previous comment on encoding.

src/twisted/words/protocols/irc.py
```diff
@@ -2620,9 +2622,17 @@ def connectionMade(self):
             self.register(self.nickname)
 
     def dataReceived(self, data):
-        basic.LineReceiver.dataReceived(self, data.replace('\r', ''))
+        data = data.replace('\r', '')
+        if isinstance(data, unicode):
+            data = data.encode("utf-8")
```
markrwilliams, line 2629: IRC is a binary protocol without a defined encoding.  That means on Pythons > 2 this can produce incorrect behavior, because messages can be in any encoding, or even multiple encodings.  Users of this API wouldn't be able to do much about that.

The protocol should not be implemented in terms of native strings any more than HTTP should be parsed in terms of native strings.  In fact, it'd be better to do this with HTTP, which _does_ have a defined encoding ([ISO 8859-1 for RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2) and [ASCII for RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2.4)), but Twisted does not.  It shouldn't do so here.

If this _does_ do encoding and decoding, all the line length checks (e.g., [`_safeMaximumLineLength`](https://github.com/twisted/twisted/pull/534/files#diff-71c9155319411063da926db195c663d5R1684)) have to run after encoding, not before.  Otherwise a line of multi-byte UTF-8 code points will appear to be of an acceptable length but will actually be too long on the wire.

src/twisted/words/protocols/irc.py
```diff
@@ -2620,9 +2622,17 @@ def connectionMade(self):
             self.register(self.nickname)
 
     def dataReceived(self, data):
-        basic.LineReceiver.dataReceived(self, data.replace('\r', ''))
+        data = data.replace('\r', '')
```
markrwilliams: I don't think this would work with a real TCP transport under Python 3, because the transport would call `dataReceived` with `bytes`, while `\r` and `''` are both native strings.

src/twisted/words/protocols/irc.py
```diff
@@ -2898,7 +2908,7 @@ class DccChat(basic.LineReceiver, styles.Ephemeral):
     delimiter = CR + NL
     client = None
     remoteParty = None
-    buffer = ""
+    buffer = b""
```
markrwilliams, line 2914: If this is `bytes`, then `DccChat.lineReceived` should decode the line before sending it to `self.client.privmsg`, or `IRCClient` should do all of its work in `bytes` and rely on a higher level API to handle decoding.

markrwilliams (commented):
I really appreciate the work done here!

I strongly disagree with the assumption that `IRCClient` should work in terms of native strings, decoding bytes as UTF-8 via a hard-coded value.

IRC is a binary protocol and [lacks a defined encoding](https://tools.ietf.org/html/rfc2812#section-2.2). IRCv3 _may_ allow Unicode in nicknames and tags, and it looks like at least [some of the previous implementation's encoding](https://github.com/twisted/twisted/commit/733a1dd36cf426b009d076a8ee98114282b4bd27) was in the name of IRCv3 support.  But IRC 2.7 and earlier do not explicitly support any particular encoding.  (I mean to imply that IRCv3 support should live outside `IRCClient`).

I think `IRCClient` should deal only in byte strings, and require that users or a high-layer API (`DecodesIRCClient`?) decode and encode data.

src/twisted/words/test/test_irc.py
rodrigc: @glyph I have addressed the coverage issues for ircsupport by adding more tests

rodrigc: @glyph Where do you want `from __future__ import unicode_literals` to be added?
I ran an experiment and added it to:

```
twisted/words/im/ircsupport.py
twisted/words/protocols/irc.py
twisted/words/test/test_irc.py
twisted/words/test/test_irc_service.py
twisted/words/test/test_ircsupport.py
```

and some of the tests started breaking on Python 2, without breaking on Python 3.

src/twisted/words/test/test_irc.py
rodrigc: Fixed.

src/twisted/words/test/test_irc.py
rodrigc: Fixed

src/twisted/words/test/test_irc.py
rodrigc: Fixed

src/twisted/words/protocols/irc.py
rodrigc: Fixed.

rodrigc:
In addition to the unit tests, I have been using this program to create an IRC server:

```
import sys

from twisted.cred import checkers, portal
from twisted.internet.endpoints import TCP4ServerEndpoint
from twisted.words import service


ROOM = 'room'

if __name__ == '__main__':
    log.startLogging(sys.stdout)

    # Initialize the Cred authentication system used by the IRC server.
    realm = service.InMemoryWordsRealm('testrealm')
    realm.addGroup(service.Group(ROOM))
    user_db = checkers.InMemoryUsernamePasswordDatabaseDontUse()
    user_db.addUser("user1", "bar") 
    user_db.addUser("user2", "bar") 
    user_db.addUser("user3", "bar") 
    user_db.addUser("user4", "bar") 
    p = portal.Portal(realm, [user_db])
    p.registerChecker(checkers.AllowAnonymousAccess())

    # IRC server factory.
    ircfactory = service.IRCFactory(realm, p)

    # Connect a server to the TCP port 6667 endpoint and start listening.
    endpoint = TCP4ServerEndpoint(reactor, 6667)
    endpoint.listen(ircfactory)
    reactor.run()
```

I have run this server under Python 2 and Python 3.

I have then connected a few IRC clients  (ircII and Limechat) to this server and have
performed these operations:

/JOIN
/DCC CHAT
/DCC SEND
/WHO
/TOPIC

and they work.

src/twisted/words/protocols/irc.py
```diff
@@ -405,6 +405,8 @@ def dataReceived(self, data):
         says CRLF.  (Also, the flexibility of LineReceiver to turn "line mode"
         on and off was not required.)
         """
+        if isinstance(data, bytes):
+            data = data.decode("utf-8")
```
nanonyme, line 409: Naively decoding input as UTF-8 sounds like a bad idea without ability to set user-defined (probably per-channel) fallback encoding. Normal IRC clients tend to work so that they first try to decode as UTF-8 and if that fails, fallback to configurable legacy 8-bit encoding

glyph: This is an important issue, but we could add encoding configurability in a follow-up PR.  Threading this through and providing the APIs to set it for each channel on all the various layers could be a separate issue.

rodrigc:
@glyph I just wanted to let you know that I did some additional testing with ircLogBot.py from http://twistedmatrix.com/documents/current/words/examples/

I did the test first with Python 2, then repeated the same test with Python 3.

I was able to connect the ircLogBot.py to `irc.freenode.net`, and was able to
connect the bot to a channel, and have it receive notifications.

Circlepuller:
I'm _really_ looking forward to this addition, as I don't believe there is really any decent IRC frameworks for Python 3.

rodrigc:
Moved to:
https://github.com/twisted/twisted/pull/561

PR #535, opened 2016-09-03
Fix insults terminal corruption

markrwilliams:
http://twistedmatrix.com/trac/ticket/8803

I went a little crazy with the tests and did find some Python-3 related problems.

### DIFF
--- a/src/twisted/conch/test/test_insults.py
+++ b/src/twisted/conch/test/test_insults.py
@@ -1,3 +1,4 @@
+
 # -*- test-case-name: twisted.conch.test.test_insults -*-
 # Copyright (c) Twisted Matrix Laboratories.
 # See LICENSE for details.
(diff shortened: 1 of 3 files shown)


src/twisted/conch/test/test_insults.py
```diff
@@ -7,10 +8,14 @@
 from twisted.test.proto_helpers import StringTransport
 
 from twisted.conch.insults.insults import ServerProtocol, ClientProtocol
-from twisted.conch.insults.insults import CS_UK, CS_US, CS_DRAWING, CS_ALTERNATE, CS_ALTERNATE_SPECIAL
+from twisted.conch.insults.insults import CS_UK, CS_US, CS_DRAWING, CS_ALTERNATE, CS_ALTERNATE_SPECIAL, BLINK, UNDERLINE
```
hawkowl: too long line here

src/twisted/conch/test/test_insults.py
```diff
@@ -1,3 +1,4 @@
+
```
hawkowl: spurious newline here?

src/twisted/conch/topfiles/8803.bugfix
```diff
@@ -0,0 +1 @@
+A Conch server no longer corrupts a client's display when attempting to set the cursor position, and insults' ServerProtocol's ECMA-48 terminal manipulation works on Python 3.
```
hawkowl: We should have a fqdn here, eg "twisted.conch servers using twisted.conch insults no longer corrupts..."

src/twisted/conch/test/test_insults.py
glyph: Looks like a separation between import sections (twisted, stdlib) to me.

glyph:
This looks _great_ to me; just restarting a couple of builds to see if we can get that checkmark to turn green…

PR #536, opened 2016-09-06
Port twisted.python.textattributes to Python 3, and add tests for it.

markrwilliams:
https://twistedmatrix.com/trac/ticket/8805


markrwilliams:
Closing while I work on a deprecation strategy for unicode strings in the formatting framework.

src/twisted/python/test/test_textattributes.py
```diff
@@ -5,23 +5,829 @@
...
+
+    def test_unicodeDeprecation(self):
+        """
+        L{_Attribute.__geitem__} emits a deprecation warning when
+        given a L{unicode}/L{str} object instead of a L{bytes} object.
+        """
+        self.attribute[u"unicode"]
+        message = ("Calling _Attribute.__getitem__ with a unicode/str"
+                   " object instead of a bytes object is deprecated"
+                   " since Twisted 16.4.1")
```
adiroiban: this should be 16.4.0. ... api changes should not be done on bugfix releases

src/twisted/conch/insults/helper.py
```diff
@@ -107,12 +107,11 @@ class TerminalBuffer(protocol.Protocol):
     """
     An in-memory terminal emulator.
     """
-    for keyID in (b'UP_ARROW', b'DOWN_ARROW', b'RIGHT_ARROW', b'LEFT_ARROW',
-                  b'HOME', b'INSERT', b'DELETE', b'END', b'PGUP', b'PGDN',
-                  b'F1', b'F2', b'F3', b'F4', b'F5', b'F6', b'F7', b'F8', b'F9',
-                  b'F10', b'F11', b'F12'):
-        execBytes = keyID + b" = object()"
-        execStr = execBytes.decode("ascii")
+    for keyID in ('UP_ARROW', 'DOWN_ARROW', 'RIGHT_ARROW', 'LEFT_ARROW',
```
adiroiban: why not have them as bytes? ... why use native string?

src/twisted/python/test/test_textattributes.py
markrwilliams: Ah, good point -- Thanks!

src/twisted/conch/insults/helper.py
markrwilliams: They're interpolated into a string that's then `exec`d (that is they're used to generate Python code).  To make that work, the final string had to be decoded as `ascii`.  No need to do that final decode if they're all native strings already.

src/twisted/python/_textattributes.py
```diff
@@ -49,7 +53,14 @@ def __repr__(self):
 
 
     def __getitem__(self, item):
-        assert isinstance(item, (list, tuple, _Attribute, str))
+        if isinstance(item, unicode):
+            warnings.warn("Calling _Attribute.__getitem__ with a unicode/str"
+                          " object instead of a bytes object is deprecated"
+                          " since Twisted 16.5.0",
```
moshez: .5.0 is released, should this string be updated?

src/twisted/python/test/test_textattributes.py
```diff
@@ -5,23 +5,829 @@
...
+    @type serializeCalls: a L{list} of L{tuple}s, each of which are
+        the arguments for one call to C{serialize}
+    """
+
+    def __init__(self):
+        self.serializeCalls = []
+
+
+
+class FakeAttribute(_Attribute):
```
moshez, line 41: Why note test a real _Attribute, and give it a write() that appends to a list or something?

src/twisted/python/test/test_textattributes.py
```diff
@@ -5,23 +5,829 @@
...
+        self.copyReturns = "<attrs>"
+
+        self.withAttributeCalls = []
+        self.withAttributeReturns = "<_withAttribute>"
+
+
+
+class FakeFormattingState(object):
+    """
+    A L{DefaultFormattingState} subclass that records all calls to its
```
moshez, line 88: It's not a subclass of DefaultFormattingState :smile:

moshez (changes requested):
Thanks for working on this! Python 3 support is awesome.

The only change that really has to be made is to figure out why the tests pass even though FakeFormattingState is not a subclass of DefaultFormattingState as far as I can see. All other comments are advisory.

glyph:
I may have lost the plot with `conch.insults` here - is this something that should just be merged after responding to @moshez's feedback?

cc @markrwilliams

src/twisted/python/_textattributes.py
twm:
```suggestion
                          " since Twisted NEXT",
```

wiml: Changed this to use the automatic "NEXT" string per @twm's suggestion

PR #537, opened 2016-09-07
[#8807] Add current dir to sys.path in twistd

sxamit:
https://twistedmatrix.com/trac/ticket/8807#ticket

Add main method in twistd.py that adds current dir to sys.path and calls run. Change build script to use new method


sxamit:
Workaround for the issue - export PYTHONPATH to env - adds the current dir to sys.path

hawkowl:
Please see https://twistedmatrix.com/trac/ticket/8807#comment:1.

sxamit:
@hawkowl Are there guidelines/best practice on how to install and run my code with twistd?

https://twistedmatrix.com/trac/ticket/8807#comment:2

PR #538, opened 2016-09-08
[#8808] Split out imap4 interfaces

hawkowl:
https://twistedmatrix.com/trac/ticket/8808


rodrigc:
This is just moving code around, so OK to merge.

PR #541, opened 2016-09-18
Port imap4-utf-7 codec implementation to Python 3 (take 2)

berdario:
After 4 years, I see that now Twisted has some Python3 compatibility (I'm actually using it... but upgrading is a pain, since support is still unstable) and some of the changes I did have been applied already.

This made it simple to recreate the changes to make IMAP4UTF7TestCase pass, without changing too much code

https://twistedmatrix.com/trac/ticket/6289

I tested my changes on both Python3.5 and Python2.7 by running

```
py.test test_imap.py -k UTF7
```


rodrigc:
This change fails many of the existing tests: 
https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1183

rodrigc (changes requested):
This change fails many of the existing tests: 
https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1183

berdario:
Yes, these failures are only in the CI environment. Running the tests with `py.test` works just fine

Running with `tox -e py27-tests twisted.mail.test.test_imap` instead doesn't start your test suite

https://gist.github.com/berdario/956f24e9f11968f11dee568588fa54c6

rodrigc:
It's OK to use py.test for your own testing, but these tests need to pass with trial.

berdario:
@rodrigc 
https://gist.github.com/101515aeee93cc7102539e844bbe6b46

All of the tests pass locally even with `trial`

rodrigc:
Well, the tests need to pass under automation as well, so your submission needs more work.

berdario:
Sure, but I'd expect that you or one of the other core developers should have some insight in how to reproduce locally failures in CI

Ideally, the [documentation](http://twistedmatrix.com/trac/wiki/TwistedDevelopment) would be up-to-date (information on how to run a single test by itself is missing in the "Tools for development" section... which is the reason why I didn't use `trial` at first)

markrwilliams:
@berdario The [documentation for trial](http://twistedmatrix.com/documents/current/core/howto/trial.html) describes the ways you can run tests at [the bottom of this section](http://twistedmatrix.com/documents/current/core/howto/trial.html#creating-an-api-and-writing-tests).  The short answer is to use a Fully Qualified Path Name (FQPN) to specify the test you want.  For instance, if you want to run just [testUnfeaturefulMessage](https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/test/test_imap.py#L4670), you'd do:

```
trial twisted.mail.test.test_imap.CopyWorkerTests.testUnfeaturefulMessage
```

It happens that this is exactly the reported name when a test fails, so you can just copy and paste each of these out.

Also, may I ask that you make sure you check "Allow edits from maintainers" on your PRs?  That way we can update your branch for you when it's gotten out of sync with trunk.

berdario:
@markrwilliams 
Sure, I think that the `Allow edits from maintainers` checkbox is already selected

In the meanwhile, I reproduced the failures, and I'm trying to fix them

src/twisted/mail/imap4.py
```diff
@@ -5493,33 +5495,35 @@ def encoder(s, errors=None):
...
                 del _in[:]
-            r.append('&-')
+            r += b'&-'
         else:
             _in.append(c)
     if _in:
-        r.extend(['&', modified_base64(''.join(_in)), '-'])
-    return (''.join(r), len(s))
+        r.extend(b'&' + modified_base64(''.join(_in)) + b'-')
+    return (bytes(r), len(s))
```
rodrigc, line 5910: bytes(r)  won't work in Python 3 because no encoding is specified.

src/twisted/mail/imap4.py
```diff
@@ -5493,33 +5495,35 @@ def encoder(s, errors=None):
...
+
 
 def decoder(s, errors=None):
     """
-    Decode the given C{str} using the IMAP4 specific variation of UTF-7.
+    Decode the given C{bytes} using the IMAP4 specific variation of UTF-7.
 
-    @type s: C{str}
+    @type s: C{bytes}
```
rodrigc: Since you are cleaning things up, any basic Python type should use L{}
for the tag, not C{}.  So that means:

C{bytes} -> L{bytes}
C{str} -> L{str}
C{tuple} -> L{tuple}
C{list} -> L{list}

This is mentioned in: https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

src/twisted/mail/imap4.py
```diff
@@ -5530,21 +5534,22 @@ def decoder(s, errors=None):
...
             decode = []
         elif decode:
             decode.append(c)
         else:
-            r.append(c)
+            r.append(c.decode())
     if decode:
-        r.append(modified_unbase64(''.join(decode[1:])))
+        r.append(modified_unbase64(b''.join(decode[1:])))
     return (''.join(r), len(s))
```
rodrigc: For return (''.join(r)
is this supposed to be dealing with bytes or unicode?
Since you are cleaning this up, you should clarify with:

return (u''.join(r)

or

return (b''.join(r)

as appropriate.

rodrigc (changes requested):

src/twisted/mail/imap4.py
berdario: It works just fine in Python3 since the `r` in `bytes(r)` is a `bytearray`

src/twisted/mail/imap4.py
berdario: ok, now I merged with trunk, so the `C{bytes}` are not there anymore

src/twisted/mail/imap4.py
berdario: Sure, I already did a [similar change](https://github.com/berdario/twisted/commit/e00024bd680f0ac44ce710acd5fc5c65c7e67f27) in the past, but I was not sure if it would just add noise

berdario:
Thanks for the prompt and quick code review

I think that now all the tests should pass

rodrigc:
@berdario: Thanks.  Good job!

PR #542, opened 2016-09-18
Move intTypes out of web/http.py and into compat.py

berdario:
https://twistedmatrix.com/trac/ticket/8829#ticket


rodrigc (changes requested):
This is not worth pulling into compat.py.

There are several places in the Twisted code which do:

```
from twisted.python.compat import long

lif isinstance(obj, (int, long)):
```

This is good enough.

berdario:
Ok, so I guess we should close this PR

PR #544, opened 2016-09-18
Fix some of the imap tests for Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8830


hawkowl:
i think you meant https://twistedmatrix.com/trac/ticket/8830#ticket not 8330 -- might need to go close 8830 too :)

PR #545, opened 2016-09-19
[#8817] Replace t.p.versions with incremental

hawkowl:
https://twistedmatrix.com/trac/ticket/8817


src/twisted/python/_setup.py
```diff
@@ -348,3 +351,163 @@ def _checkCPython(sys=sys, platform=platform):
 
 
 _isCPython = _checkCPython()
+
+notPortedModules = [
```
glyph, line 360: Do we really need to do this huge move in this branch?  It would be nice if we could avoid it on account of giant conflicts with every in-process py3 porting branch.

glyph (commented):
A few issues, all minor;
1. The thing with the move from `dist3`.  If we really need to do that in order to make this work, could we do it in a separate PR, that does nothing else, that lands first?
2. This is just a thought, not anything actionable, but it sure would be nice if we had some way to validate that `incremental.Version` is actually API-compatible.  Maybe we should make some tools for that.  Right now it sorta looks to me like it is, but mostly I'm going to have to take your word for it (I'm not revieiwng _all_ of incremental as if it were all one change).
3. This could really use a ''much'' better ticket description.  Why are we doing this?  What is the point, what is the advantage, what is the problem?  Pretend that someone who cares a lot about Twisted but doesn't know about any of our previous discussions looks at the ticket.  What's going on here?

Please address 1 and 3.  If you can address the first point by just reversing the dist3 move and avoiding any conflict headaches then you can just go ahead and land after that; if we need to deal with that with a new PR then please submit for re-review afterwards.

src/twisted/_version.py
```diff
@@ -1,11 +1,7 @@
...
+# This file is auto-generated! Do not edit!
+# Use `python -m incremental.update Twisted` to change this file.
 
-# This is an auto-generated file. Do not edit it.
+from incremental import Version
 
-"""
-Provides Twisted version information.
-"""
-
```
thijstriemstra, line 9: Why remove the module docstring?

glyph: For some reason when I was looking at this I must have thought that was removing this docstring from `twisted.python.version` (which is just a compatibility alias).  _This_ docstring should probably stick around.

hawkowl:
@glyph I removed the test suite, would adding it back (with the fixes for incremental changes, the only one being canonicalising `pre->rc`) satisfy review point 2?

hawkowl:
@glyph also the dist3 move is required because importing several modules when you are exec()ing a file turns into chaos. If I exec() a file which doesn't import anything else Twisted, it doesn't crash and burn upon trying to import incremental, because setup_requires hasn't even been parsed yet, so incremental doesn't yet exist. It also means that we don't import bits of Twisted in the installer (importing `twisted.python._dist3` imports t.p.compat, t.p.deprecate, t.p.versions).

I'll do the dist3 move in another ticket. It shouldn't cause any big conflicts, I think.

hawkowl:
Dist3 move is in https://github.com/twisted/twisted/pull/555...

hawkowl:
Okay, should be good for rereview. I reinstated the old tests to ensure compat, modifying them slightly to reflect Incremental's behaviour (pre -> rc but with backwards compat shims, to keep up with setuptools).

docs/core/development/policy/release-process.rst
```diff
@@ -281,9 +282,10 @@ Announce
   - Twitter, if you feel like it
   - ``#twisted`` topic on IRC (you'll need ops)
 
-5. Merge the release branch into trunk, closing the release ticket at the same time.
-6. Close the release milestone (which should have no tickets in it).
-7. Open a milestone for the next release.
+5. Run ``python -m incremental Twisted --dev`` to add a `dev0` postfix.
```
markrwilliams, line 285: Should there be a commit after this to store the new `dev0` version, and if so, should the merge in step 6 happen first?

src/twisted/python/_setup.py
```diff
@@ -36,12 +36,14 @@
 from setuptools import Extension, find_packages
 from setuptools.command.build_py import build_py
 
-from twisted import copyright
-from twisted.python.compat import _PY3
+# Try not to export anything else in Twisted unless we really have to.
```
markrwilliams: It'd be nice to see this expanded into general advice about modifying the file under the constraints of `exec` and moved into the docstring.

tox.ini
```diff
@@ -95,7 +94,7 @@ commands =
     python --version
 
     ; Install PyDoctor here so it DOESNT overwrite Twisted
-    py27-{alldeps}: pip install --no-deps epydoc pydoctor
+    py27-{alldeps,tests}: pip install --no-deps epydoc pydoctor
```
markrwilliams, line 97: Tell us more about why this got a `tests`.

docs/core/development/policy/release-process.rst
```diff
@@ -107,9 +107,9 @@ How to do a pre-release
 
 1. Check ​buildbot to make sure all supported platforms are green (wait for pending builds if necessary).
 2. If a previously supported platform does not currently have a buildbot, move from supported platforms to "expected to work" in ``INSTALL.rst``.
-3. In your Git-SVN-enabled Git repo, fetch and check out the new release branch.
-4. Run ``./bin/admin/change-versions --prerelease``
-5. Commit the changes made by change-versions
+3. In your Git repo, fetch and check out the new release branch.
+4. Run ``python -m incremental.update Twisted --newversion $RELEASErc1``
```
markrwilliams, line 111: The change from "prerelease" to "release candidate" isn't a big deal because only `pip` (and `dpkg` and `pkg` and...) care, right?

markrwilliams (commented):
Minor comments, most of which ask for clarification.  The most important one is confirming the release management documentation.

docs/core/development/policy/release-process.rst
hawkowl: Yeah, not a big deal. New setuptools force-normalise 'pre' into 'rc' anyway.

tox.ini
hawkowl: It got a tests because it had a tests before; it's now a 'alldeps' alias, but an earlier merge broke it :)

docs/core/development/policy/release-process.rst
hawkowl: No, I think the dev0 should happen before the merge (but after the tag). Will add that, anyway.

src/twisted/python/_setup.py
```diff
@@ -350,10 +355,14 @@ def _checkCPython(sys=sys, platform=platform):
 notPortedModules = [
     "twisted.conch.client.connect",
     "twisted.conch.client.direct",
+    "twisted.conch.tap",
     "twisted.conch.test.test_cftp",
     "twisted.conch.test.test_ckeygen",
     "twisted.conch.test.test_conch",
+    "twisted.conch.test.test_helper",
```
markrwilliams: this actually got ported [recently](https://github.com/twisted/twisted/pull/554)

src/twisted/python/_setup.py
```diff
@@ -5,6 +5,10 @@
 """
 Setuptools convenience functionality.
 
+This file must not import anything from Twisted, as it is loaded by C{exec} in
+C{setup.py}. If you need compatibility functions for this code,, duplicate them
```
markrwilliams: Double comma!

markrwilliams:
@hawkowl This looks ready!  I'll merge when the builds are green :)

PR #546, opened 2016-09-19
Run twist as a module

LilyFirefly:
See https://twistedmatrix.com/trac/ticket/8657.


src/twisted/__main__.py
```diff
@@ -0,0 +1,7 @@
+import sys
+from pkg_resources import load_entry_point
```
Lukasa, line 11: Can we have the standard module preamble here? That is:

``` python
# Copyright (c) Twisted Matrix Laboratories.
# See LICENSE for details.

"""
Some docstring
"""

from __future__ import division, absolute_import
```

Lukasa (changes requested):
Looks pretty good! One small change from me.

src/twisted/__main__.py
LilyFirefly: Adding the docstring here has caused the test to fail:

``` diff
       --log-level=   Set default log level. (options: "debug", "info", "warn",
                      "error", "critical"; default: "info")
       --version      Print version and exit.
-
-Make the twisted module executable with the default behaviour of running twist.
 Commands:
     conch            A Conch SSH service.
     dns              A domain name server.
```

The easiest fix is to omit the docstring again, but I'm not sure if that's best.

src/twisted/__main__.py
```diff
@@ -0,0 +1,17 @@
...
+Make the twisted module executable with the default behaviour of
+running twist.
+"""
+
+from __future__ import division, absolute_import
+
+import sys
+from pkg_resources import load_entry_point
+
+if __name__ == '__main__':
```
LilyFirefly, line 13: If I understand `__main__.py`, this will always be true in correct usage and therefore this isn't needed. Removing this would avoid needing to test the empty else branch for 100% coverage.

src/twisted/__main__.py
Lukasa: Hrm, OK, can you please submit a patch that removes the docstring and replaces it with a comment block, and include in that comment block a note about why there is no docstring.

src/twisted/__main__.py
Lukasa: I don't think there is any protection against importing `__main__`: at least, I am capable of doing it for tools like `pip`. So I think we still want the if branch. 

What we should probably do is test that no output is produced when importing the `__main__` module directly, in this case by delaying the import until a test case and then replicating [this test logic](https://github.com/twisted/twisted/blob/trunk/src/twisted/trial/test/test_warning.py#L161-L165) which mocks out stdout.

Lukasa: Alternatively you could mock out `load_entry_point` briefly, which is probably more direct.

src/twisted/test/test_main.py
```diff
@@ -0,0 +1,47 @@
...
+            self.assertEqual(output, message)
+        return d.addCallback(processEnded)
+
+    def test_twisted_import(self):
+        """Importing twisted.__main__ does not execute twist."""
+        output = StringIO()
+        monkey = self.patch(sys, 'stdout', output)
+
+        import twisted.__main__
+        assert twisted.__main__  # Appease pyflakes
```
Lukasa: Let's do `self.assertTrue()` at least.

Lukasa (changes requested):
One very small note, and then I think I'll be happy.

glyph:
Ping?

LilyFirefly:
@Lukasa I think I fixed your last concern. Can you confirm?

Lukasa (approved):
Agreed, I am happy.

Lukasa:
However, the branch is out of date, so I'm going to do a catch-up merge to force a new build.

Lukasa:
Also @Ian-Foote, can you please resubmit [the trac ticket](https://twistedmatrix.com/trac/ticket/8657) for review so that I can re-review it? (Add keyword `review`, assign to the empty entry at the top of the drop-down.)

Julian:
Why `python -m twisted` and not `python -m twist`? Not only is the latter shorter (sigh), but I'm sure when `twistd` was created someone probably also assumed it was going to be the be-all end-all of twisted invocations :P?

Lukasa:
I think the short answer for why not `python -m twist` is that to do that twisted would also have to install a module called `twist`.

Julian:
OK. I guess maybe to make myself feel better, twisted could just provide the same-ish API over `python -m twisted` even if we decided on somehow better semantics at some point. (Though tbh installing a `twist` module doesn't sound too bad to me).

Anyways, retracted (thanks for the answer, that does make sense).

Lukasa:
Yeah, I think `python -m twisted` will always run twisted in its _preferred_ mode, whatever that is. =)

Lukasa:
Thanks @Ian-Foote! :sparkles:

PR #547, opened 2016-09-20
4362 getaddrinfo

glyph:
https://twistedmatrix.com/trac/ticket/4362

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -17,9 +17,13 @@ from heapq import heappush, heappop, heapify
 
 import traceback
 
+from twisted.python.util import FancyEqMixin
+from twisted.python.compat import set
+from twisted.python.components import registerAdapter
 from twisted.internet.interfaces import IReactorCore, IReactorTime, IReactorThreads
 from twisted.internet.interfaces import IResolverSimple, IReactorPluggableResolver
 from twisted.internet.interfaces import IConnector, IDelayedCall
+from twisted.internet.interfaces import INameResolver
 from twisted.internet import fdesc, main, error, abstract, defer, threads
 from twisted.python import log, failure, reflect
 from twisted.python.compat import unicode, iteritems
@@ -209,6 +213,51 @@ class DelayedCall:
         return "".join(L)
 
 
+@implementer(INameResolver)
+class _ResolverComplexifier(object):
+    """
+    L{_ResolverComplexifier} adapts an L{IResolverSimple} provider to
+    L{INameResolver}.
+    """
+    def __init__(self, resolver):
+        """
+        @param resolver: A resolver which will be adapted to the new
+            L{INameResolver} interface.
+        @type resolver: A provider of L{IResolverSimple}.
+        """
+        self._resolver = resolver
+
+
+    def getAddressInformation(self, name, service, *args):
+        # In this case of a complexified IResolverSimple,
+        # getHostByName and getAddressInformation should - for
+        # consistency - both return the same single IP address. (IPv4
+        # or IPv6 depending on the behaviour of gethostbyname)
+
+        # getAddressInformation will return that IP address as part of
+        # an AddressInformation instance whose type, protocol and
+        # canonicalName are fixed.
+
+        # XXX: I don't fully understand the motivation behind
+        # _ResolverComplexifier or why lookup the address with
+        # getHostByName and then pass the address to getaddrinfo?
+        # -rwall
+        d = self._resolver.getHostByName(name)
+        def cbResolved(address):
+            family = socket.getaddrinfo(address, 0)[0][0]
+            return [
+                AddressInformation(
+                    family,
+                    socket.SOCK_STREAM,
+                    socket.IPPROTO_TCP,
+                    "",
+                    (address, service))]
+        d.addCallback(cbResolved)
+        return d
+
+registerAdapter(_ResolverComplexifier, IResolverSimple, INameResolver)
+
+
 
 @implementer(IResolverSimple)
 class ThreadedResolver(object):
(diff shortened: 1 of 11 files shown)


glyph:
continuing on from #492 since github has trouble with merges & PRs…

glyph:
I think I'm going to try to start off with a self-review here in order to get my head around the status that this is in and make a checklist of things to do before putting it back into review.

src/twisted/internet/base.py
```diff
@@ -209,6 +213,51 @@ def __str__(self):
...
+        # In this case of a complexified IResolverSimple,
+        # getHostByName and getAddressInformation should - for
+        # consistency - both return the same single IP address. (IPv4
+        # or IPv6 depending on the behaviour of gethostbyname)
+
+        # getAddressInformation will return that IP address as part of
+        # an AddressInformation instance whose type, protocol and
+        # canonicalName are fixed.
+
+        # XXX: I don't fully understand the motivation behind
```
glyph, line 241: Obviously this should not be a `XXX` comment here.

src/twisted/internet/base.py
```diff
@@ -278,6 +327,110 @@ def getHostByName(self, name, timeout = (1, 3, 11, 45)):
...
+        userDeferred, cancelCall = self._runningQueries[lookupDeferred]
+        del self._runningQueries[lookupDeferred]
+        userDeferred.errback(self._fail(query, "timeout error"))
+
+
+    def _checkTimeout(self, result, query, lookupDeferred):
+        try:
+            userDeferred, cancelCall = self._runningQueries[lookupDeferred]
+        except KeyError:
+            pass
```
glyph, line 385: Checking coverage locally, I can see that this `except` branch is un-covered.

glyph: What happened to codecov?!?

src/twisted/internet/base.py
```diff
@@ -506,12 +659,31 @@ def installWaker(self):
         raise NotImplementedError(
             reflect.qual(self.__class__) + " did not implement installWaker")
 
+
     def installResolver(self, resolver):
-        assert IResolverSimple.providedBy(resolver)
```
glyph, line 510: This could probably use a comment explaining how this object is being type-checked now.

src/twisted/internet/base.py
```diff
@@ -506,12 +659,31 @@ def installWaker(self):
...
+        """
+        if self._resolver is None:
+            return BlockingResolver()
+        return self._resolver
+
+
+    @resolver.setter
+    def resolver(self, newResolver):
+        """
+        Setter for resolver property.
```
glyph, line 682: Docstring should explain the behavior.

src/twisted/internet/base.py
```diff
@@ -567,7 +739,15 @@ def resolve(self, name, timeout = (1, 3, 11, 45)):
             return defer.succeed('0.0.0.0')
         if abstract.isIPAddress(name):
             return defer.succeed(name)
-        return self.resolver.getHostByName(name, timeout)
+        d = self.resolver.getAddressInformation(name, 0)
+        def cbGotInfo(addresses):
+            for info in addresses:
+                if info.family == socket.AF_INET:
+                    return info.address[0]
+            # XXX Test me
```
glyph, line 747: This code is in fact covered by tests so I'm not sure what this comment is referring to; it should be removed.

src/twisted/internet/interfaces.py
```diff
@@ -55,6 +55,70 @@ def getDestination():
 
 
 
+class INameResolver(Interface):
+    """
+    XXX Write me.
```
glyph, line 60: This docstring needs to be written.

src/twisted/internet/interfaces.py
```diff
@@ -55,6 +55,70 @@ def getDestination():
...
+            which case the L{Deferred} cancellation timeout will be the sum of
+            all the C{int}s.
+        @type: timeout: C{int} or an iterable of C{int}
+
+        @raise ValueError: If one of the specified flags is not supported by
+            the implementation.  All flags are optional.
+
+        @return: A L{Deferred} which will fire when the resolution completes.
+            If resolution is successful, the result will be a list of objects
+            with the following attributes:
```
glyph, line 104: This "with the following attributes" section should really be a reference to a type, rather than an ad-hoc description of attributes.

src/twisted/internet/base.py
```diff
@@ -278,6 +327,110 @@ def getHostByName(self, name, timeout = (1, 3, 11, 45)):
...
+            cancelCall.cancel()
+
+            if isinstance(result, failure.Failure):
+                userDeferred.errback(self._fail(query, result.getErrorMessage()))
+            else:
+                userDeferred.callback([AddressInformation(*r) for r in result])
+
+
+
+class AddressInformation(FancyEqMixin, object):
```
glyph, line 397: If this class is intended to be public, it should be in a module with less of a troubled history than `base`.  If not, it should be properly private.

src/twisted/internet/base.py
```diff
@@ -278,6 +327,110 @@ def getHostByName(self, name, timeout = (1, 3, 11, 45)):
...
+    """
+    A container for the results of
+    L{INameResolver.getAddressInformation}.
+    """
+    compareAttributes = ('family', 'type', 'protocol', 'canonicalName', 'address')
```
glyph, line 402: Euuuuggghhh.  _seriously_ consider adding a dependency on `attrs` for this rather than making more classes use `FancyEqMixin`.

src/twisted/internet/test/test_base.py
```diff
@@ -53,6 +59,206 @@ def _stop(self):
...
+
+        def fakeGetAddrInfo(*args):
+            raise IOError("ENOBUFS (this is a funny joke)")
+
+        self.patch(socket, 'getaddrinfo', fakeGetAddrInfo)
+
+        failedWith = []
+        resolver = ThreadedNameResolver(reactor)
+        d = resolver.getAddressInformation(*query, timeout=timeout)
+#        import pdb;pdb.set_trace()
```
glyph, line 213: Oops

src/twisted/internet/test/test_tcp.py
```diff
@@ -115,6 +115,28 @@ def connect(client, destination):
...
+    """
+    def _freePort(self, interface='127.0.0.1'):
+        probe = socket.socket()
+        try:
+            probe.bind((interface, 0))
+            return probe.getsockname()
+        finally:
+            probe.close()
+
+    def test_clientConnectionFailedStopsReactor(self):
```
glyph, line 130: I fail to see what this has to do with the change in this branch.

src/twisted/internet/test/test_tcp.py
```diff
@@ -115,6 +115,28 @@ def connect(client, destination):
 
 
 
+class TCPClientTestsBuilder(ReactorBuilder):
+    """
+    Builder defining tests relating to L{IReactorTCP.connectTCP}.
+    """
+    def _freePort(self, interface='127.0.0.1'):
+        probe = socket.socket()
```
glyph, line 123: This is a _bad_ way of probing for a free port; it's non-deterministic and race-condition-prone .

glyph: There are other ways of probing for a free port that are already implemented as part of reactor builder tests, using port 0 with `listenTCP`.  We should use that technique instead.

glyph:
ugh. and one more thing; we already have types representing addresses.  it might be a good idea to use `IPv4Address` and `IPv6Address` for the address component of the result.

glyph:
As I'm going through this and trying to de-duplicate information between the GAI results and `IPv4Address` / `IPv6Address`, it is increasingly clear that many of these flags don't make any sense for Twisted's native interfaces.

glyph:
The interface that we care about here is `HostnameEndpoint`, and what `HostnameEndpoint` wants to do is to transform the results of `getaddrinfo` into endpoints.  So here's what `HostnameEndpoint` does with each field:
- `family`: used as a value to determine whether the endpoint should be IPv4 or IPv6
- `socktype`: totally ignored, since we only care about `SOCK_STREAM` and we say so up front
- `proto`: totally ignored, since it conveys information equivalent to `socktype` - `IPPROTO_TCP` = `SOCK_STREAM`, `IPPROTO_UDP` = `SOCK_DGRAM`.
- `canonname`: totally ignored; we don't set `AI_CANONNAME` (and I can't figure out what )
- `sockaddr`; only the first and second (host and port) elements of the tuple are honored.  Arguably, scope ID and flow label really ought to be honored as well, for IPv6.

On the one hand, it seems like straightforwardly exposing the getaddrinfo struct in the same way that Python does avoids potentially eliminating any information that might be useful in weird edge cases. At the same time, the whole point of defining an _interface_ here is that we need to be able to define what this means in terms of DNS queries so that `twisted.names` can reconfigure the reactor to take on this functionality.

glyph:
Another aspect which may not be necessary is `flags`; going through the list of flags, applications wouldn't be able to pass most of these blindly through `HostnameEndpoint`.  There are also tons of non-portable flags.

The more I play with this the more I don't want to be mirroring the platform interfaces.

PR #548, opened 2016-09-22
asynchronously resolve a name to multiple address objects

glyph:
https://twistedmatrix.com/trac/ticket/4362


glyph:
I am bumping into this issue https://github.com/twisted/twistedchecker/issues/44 while writing the docstrings for this.

src/twisted/internet/_resolver.py
```diff
@@ -0,0 +1,263 @@
...
+
+"""
+IPv6-aware hostname resolution.
+
+@see: L{IHostnameResolver}
+"""
+
+from __future__ import division, absolute_import
+
+__metaclass__ = type
```
hawkowl, line 13: why do we need this

src/twisted/internet/interfaces.py
```diff
@@ -84,6 +84,111 @@ def getHostByName(name, timeout = (1, 3, 11, 45)):
...
+            transport; either C{'TCP'} for stream-oriented transports or
+            C{'UDP'} for datagram-oriented; see
+            L{twisted.internet.address.IPv6Address.type} and
+            L{twisted.internet.address.IPv4Address.type}.
+        @type transportSemantics: native L{str}
+
+        @return: The resolution in progress.
+        @rtype: L{IResolutionReceiver}
+        """
+
```
hawkowl, line 188: this seems like too many newlines

src/twisted/internet/test/test_resolver.py
```diff
@@ -0,0 +1,559 @@
...
+# See LICENSE for details.
+
+"""
+Tests for implementations of L{IHostnameResolver} and their interactions with
+reactor implementations.
+"""
+
+from __future__ import division, absolute_import
+
+__metaclass__ = type
```
hawkowl, line 11: again, what's this for?

hawkowl (approved):
pls look at these things and merge when you are happy, seems good to me

src/twisted/internet/_resolver.py
glyph: To avoid accidentally declaring an old-style class.

src/twisted/internet/test/test_resolver.py
glyph: Same as above. It avoids accidentally declaring an old-style class in py2.

hawkowl:
Why not add it to the list in t.test.test_nooldstyle or whatever?

glyph:
> Why not add it to the list in t.test.test_nooldstyle or whatever?

It's kind of a separate issue.  Although that would also be a good thing.

This list in `test_nooldstyle` has no docstring, though. It should have an `@ivar` so I know exactly what to do with it :)

PR #549, opened 2016-09-23
[#8281] Remove old-style classes in twisted.internet with the oldStyle decorator

hawkowl:
http://twistedmatrix.com/trac/ticket/8281


src/twisted/test/test_nooldstyle.py
```diff
@@ -145,7 +150,13 @@ def test_newStyleClassesOnly(self):
                     oldStyleClasses.append(fullyQualifiedName(val))
 
         if oldStyleClasses:
+
             self.todo = "Not all classes are made new-style yet. See #8243."
+
+            for x in forbiddenModules:
```
hawkowl, line 156: Just a note -- this is red because there are no forbidden modules that have old-style classes in them (this should always be red)

src/twisted/test/test_nooldstyle.py
```diff
@@ -24,6 +24,11 @@
     _skip = "Not running with TWISTED_NEWSTYLE=1"
 
 
+forbiddenModules = [
```
glyph, line 27: This should probably be called `forbiddenPackages` if its contents are only going to be packages.

src/twisted/test/test_nooldstyle.py
```diff
@@ -24,6 +24,11 @@
...
+    "twisted.internet",
```
glyph, line 28: On second thought, maybe it _should_ be called `forbiddenModules`, since it would be best to do this as narrowly as possible for its first iteration.

src/twisted/test/test_postfix.py
```diff
@@ -31,7 +31,7 @@ def testData(self):
             for q in quoted:
                 self.assertEqual(postfix.unquote(q), raw)
 
-class PostfixTCPMapServerTestCase:
+class PostfixTCPMapServerTestCase(object):
```
glyph, line 34: You could do the changes to private classes to `object` in a separate PR first.

src/twisted/pair/tuntap.py
```diff
@@ -251,7 +251,7 @@ def __init__(self, interface, proto, maxPacketSize=8192, reactor=None,
 
 
     def __repr__(self):
-        args = (self.protocol.__class__,)
+        args = (fullyQualifiedName(self.protocol.__class__),)
```
glyph, line 254: This isn't a bad idea, but how does it impact `twisted.internet`?

glyph (approved):
Thanks @hawkowl !  I'm happy to see us moving towards automated elimination of old-style classes.

It occurs to me that it might be interesting to factor this out into a library, or maybe even push it upstream into `six`, since it would be nice if other packages could push off of their py2 dependencies this way.

I can't really complain that this is too big, so if you can get it to merge cleanly after resolving conflicts, please feel free to land it.

src/twisted/pair/tuntap.py
hawkowl: Look at the inheritance tree :)

PR #550, opened 2016-09-25
https://twistedmatrix.com/trac/ticket/8839#ticket

glyph:


.travis/twistedchecker-trunk-diff.sh
```diff
@@ -25,7 +25,7 @@ if [ ! -d "$target" ]; then
 fi
 
 # Make sure we have trunk on the local repo.
-git fetch origin trunk:refs/remotes/origin/trunk
+git fetch origin +refs/heads/trunk:refs/remotes/origin/trunk
```
markrwilliams, line 28: I'm going to attempt to provide a gloss to this:

`git fetch origin +refs/heads/trunk:refs/remotes/origin/trunk`

...instructs `git` to fetch `refs/heads/trunk` from the remote named `origin` and store it locally in `refs/remotes/origin/trunk`.  It appears that `trunk` in the source field of the refspec should be expanded to `refs/heads/trunk`.  Maybe in this case it's not?

`+` means "update the local reference even if it's not a fast forward."  That seems a little worrisome to me, but it's the default behavior for `fetch` as configured by `git remote add`.  You can confirm this by cloning a repository (since `git clone` implies `git remote add`) and checking `.git/config`:

```
[remote "origin"]
        <snip>
        fetch = +refs/heads/*:refs/remotes/origin/*
```

This _seems_ OK, because my reading of the Git Book indicates that it's not different from what `git fetch origin trunk:refs/remotes/origin/trunk` should be anyway.   So there should be no regression.

Sources:
https://git-scm.com/book/en/v2/Git-Internals-The-Refspec
https://www.atlassian.com/git/tutorials/refs-and-the-reflog/special-refs

markrwilliams (approved):
I can't reproduce the issue described in the ticket locally.  I also can't figure out how the change results in different behavior after reviewing the Git documentation.  But it certainly doesn't do anything bad in my testing, and it's at worst harmless if my understanding of things is right.

OK to merge.  It'd be great to have somebody who understands `git` weigh in.

PR #551, opened 2016-09-25
[#8838] cfreactor on py3

hawkowl:
https://twistedmatrix.com/trac/ticket/8838


src/twisted/internet/test/test_process.py
```diff
@@ -862,6 +862,8 @@ class PTYProcessTestsBuilder(ProcessTestsBuilderBase):
     if platform.isWindows():
         skip = "PTYs are not supported on Windows."
     elif platform.isMacOSX():
+        skip = "PTYs are flaky from a Darwin bug. See #8840."
```
glyph, line 865: This might be a bit overly broad.  PTYs should totally be _supported_ on OS X (and they do work, to the extent that the platform works), just there are a few tests which are inadvertently exercising edge-cases around `close()`.

src/twisted/internet/cfreactor.py
```diff
@@ -167,36 +167,41 @@ def _socketCallback(self, cfSocket, callbackType,
...
+                        if rw[_WRITE]:
+                            why = readWriteDescriptor.doWrite()
+            except:
+                why = sys.exc_info()[1]
+                log.err()
+            if why:
+                self._disconnectSelectable(readWriteDescriptor, why, isRead)
+
+        readWriteDescriptor = self._fdmap[fd][2]
+        log.callWithLogger(readWriteDescriptor, _drdw)
```
glyph, line 203: Let me see if I can infer the purpose of this change: previously it was calling `why = log.callWithLogger(...)` and expecting `why` to get set to something truthy in the case where there was a problem, so we could notice that problem, and `_disconnectSelectable`.  Now we're allowing the exception to bubble up, always catching it, and setting `why =` properly, so in the case where there's an exception we _actually_ disconnect.

OK, that makes sense.

glyph: Setting `why` "properly" meaning that `callWithLogger` _just returns `None`_ in the case where there's an exception (and it logs it), which is not helpful to this code.

src/twisted/internet/cfreactor.py
```diff
@@ -167,36 +167,41 @@ def _socketCallback(self, cfSocket, callbackType,
...
-                # actually attempt a doRead/doWrite first.  -glyph
-                if isRead:
-                    if rw[_READ]:
-                        why = log.callWithLogger(
-                            readWriteDescriptor, readWriteDescriptor.doRead)
+        def _drdw():
+
+            why = None
+            isRead = False
+            src, skt, readWriteDescriptor, rw = self._fdmap[fd]
```
glyph: I don't like the fact that `readWriteDescriptor` is bound twice, once inside and once outside.  It's never _re_-bound, so could you just move this tuple-unpacking assignment outside of `_drdw`?

src/twisted/internet/cfreactor.py
```diff
@@ -167,36 +167,41 @@ def _socketCallback(self, cfSocket, callbackType,
...
+                    else:
+                        if rw[_WRITE]:
+                            why = readWriteDescriptor.doWrite()
+            except:
+                why = sys.exc_info()[1]
+                log.err()
+            if why:
+                self._disconnectSelectable(readWriteDescriptor, why, isRead)
+
+        readWriteDescriptor = self._fdmap[fd][2]
```
glyph: (see comment above about re-binding)

src/twisted/internet/udp.py
```diff
@@ -365,6 +365,7 @@ def connectionLost(self, reason=None):
         """
         log.msg('(UDP Port %s Closed)' % self._realPortNumber)
         self._realPortNumber = None
+        self.maxThroughput = -1
```
glyph, line 368: WOOP WOOP NEW PUBLIC ATTRIBUTE, WOOP WOOP

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
         finally:
             results = None
             running.pop()
+
+
+    def _waitUntilAllDisconnected(self, reactor, protocols):
```
glyph: This is just _horrible_!  Also: genius.

However, I think explaining why it's horrible might be a good idea here.  Link to jml's post https://jml.io/pages/how-to-disconnect-in-twisted-really.html about how to do this the "right" way, explain that it's using timed polling, that this is a workaround for legacy test cases, etc.

Also, there's no reason to put this on `TestCase`, especially given that it's not called from here.  `self` isn't even used in the body of the function.

This should maybe go in `proto_helpers`

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
...
+        """
+        Take a list of disconnecting protocols, callback a Deferred when
```
glyph: nope, not protocols, transports

glyph: Oops I mean the reverse

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
...
+        they're all done.
+
+        @param: The reactor to use.
+        @param: A list of L{ITransport} or similars.
```
glyph: not _transports_, _protocols_

glyph: also don't say "or similar"; either it's `IProtocol` or it's not

glyph: also maybe you were a little tired?  `@param ...` what?  which param?

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
...
             results = None
             running.pop()
+
+
+    def _waitUntilAllDisconnected(self, reactor, protocols):
+        """
+        Take a list of disconnecting protocols, callback a Deferred when
+        they're all done.
+
+        @param: The reactor to use.
```
glyph: which param?

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
...
+        d = defer.Deferred()
+        lc = None
+
+        def _check():
+            if not True in [x.transport.connected for x in protocols]:
+                d.callback(None)
+
+        lc = task.LoopingCall(_check)
+        lc.clock = reactor
+        lc.start(0)
```
glyph: `0` strikes me as unduly harsh here.  in the case where this doesn't work immediately I don't see any need to burn 100% CPU in the meanwhile.  start has a flag, `now=True`, which will run it once immediately, and that will take care of the case where everything's already ready.

src/twisted/trial/_asynctest.py
```diff
@@ -403,3 +403,26 @@ def stop():
...
+
+        d.addCallback(lambda _: lc.stop())
```
glyph: Just an implementation hint here: `lc.start` returns a `Deferred` that fires when it's stopped.  So you can just do `return lc.start(...)` and then in `check` do `lc.stop` which serves the dual purpose of stopping the looping call and firing the `Deferred`; bonus, you don't need to instantiate or manage your own `Deferred`.

glyph: Managing the Deferred is important, too; when it gets timed out, the canceler needs to be sure to _also_ stop the `LoopingCall`.

glyph: As it happens, `LoopingCall`'s `Deferred` _doesn't_ do this, but that's its bug, not yours :)

glyph: specifically this bug https://twistedmatrix.com/trac/ticket/6656

src/twisted/web/test/test_webclient.py
```diff
@@ -348,7 +348,11 @@ def tearDown(self):
             proto.transport.loseConnection()
         if connections:
             msg("Some left-over connections; this test is probably buggy.")
-        return self.port.stopListening()
+        self.port.stopListening()
```
glyph: You need to honor this `Deferred` as well; `stopListening` is explicitly allowed to return a `Deferred` (see here https://twistedmatrix.com/documents/16.4.1/api/twisted.internet.interfaces.IListeningPort.html#stopListening)

tox.ini
```diff
@@ -8,7 +8,9 @@
 ; * withcov - run the tests wrapped using the coverage.
 ; * nocov - run the tests directly, without using the coverage wrapper.
 ; * wheel - build the wheel distribution
-; * posix - prepare the tests to be run in a Linux/Unix/OSX environment.
+; * posix - prepare the tests to be run in a Linux/Unix/macOS environment.
```
glyph, line 11: okay this explosion of environments is getting kind of ridicuous.  If we can't use platform specifiers, I'm starting to think we should include a `Makefile` or _something_ that does the dance you need to figure out what tox environments are appropriate to run locally.  I have a shell script that does this right now:

``` bash
#!/bin/bash
cd ~/Projects/Twisted;
coverage erase;
detox $(tox -l | grep withcov-posix |
            while read x;
            do
                echo " -e ${x}";
            done
    ) -e apidocs \
      -e narrativedocs \
      -e topfile \
      -e manifest-checker \
      -e txchecker-travis;
coverage combine;
coverage html;
```

This is not a blocker for this branch but it's something we should do soon.

tox.ini
```diff
@@ -51,10 +53,17 @@ deps =
...
 
-    py27-{alldeps,tests,coverage}-posix: pysqlite
+    py27-{alldeps,tests,coverage}-{posix,macos}: pysqlite
     py27-{alldeps,tests,coverage}: soappy
 
-    alldeps-windows: pypiwin32
+    windows: pypiwin32
+
+    {alldeps,tests}-macos: pyobjc-core
+    {alldeps,tests}-macos: pyobjc-framework-cfnetwork
```
glyph: Also not really related to this branch, but it reminds me, we should be similarly discerning in `macosx_platform`, and not depend on all of `pyobjc`.

glyph: Also, canonical capitalization here is `"CFNetwork"`, as per https://pypi.python.org/pypi/pyobjc-framework-CFNetwork

src/twisted/internet/cfreactor.py
glyph: `callWithLogger` is a bad scene though, for real, we need a new/good logging system replacement for this idiom.

glyph (commented):
I wish I could just approve this; these improvements to reliability are pretty awesome!  The implementation shows clear signs of sleep deprivation in places, though, so I think it's probably best if we do one more round of review.

src/twisted/internet/udp.py
hawkowl: Existing public attribute!

src/twisted/web/test/test_webclient.py
hawkowl: I pass self.port to the following function, which should achieve the same effect?

src/twisted/internet/test/test_process.py
hawkowl: I had something that shouldn't be related to close crash, though... maybe i need to stress test them individually. Does this happen more on faster PCs than slower? I tried reproducing and it wouldn't...

src/twisted/internet/cfreactor.py
hawkowl: What is the idiom? I don't think the idiom exists in the new logging world; loggers are responsible for their own contexts.

glyph, line 203: That's why we need one. If it existed, we'd just need to use it.

src/twisted/internet/test/test_process.py
glyph: To be clear, we can make re-enabling the PTY tests that _do_ work a separate ticket.  We don't need it here.  Please just file it so that we don't forget.

tox.ini
hawkowl: Ticket filed on https://twistedmatrix.com/trac/ticket/8848#ticket

src/twisted/internet/test/test_process.py
hawkowl: Should #8840 suffice?

src/twisted/internet/udp.py
markrwilliams: It's [even documented](https://github.com/twisted/twisted/blob/5c1777777f85aaf799c2269f9f5dddcf0d59803e/src/twisted/internet/udp.py#L66-L67).

src/twisted/test/proto_helpers.py
```diff
@@ -834,3 +834,33 @@ def connectUNIX(self, address, factory, timeout=30, checkPID=0):
...
+    @param reactor: The reactor to use.
+    @type reactor: L{IReactor}
+
+    @param protocols: The protocols to wait for disconnecting.
+    @type protocols: A L{list} of L{IProtocol}s.
+    """
+    lc = None
+
+    def _check():
+        if not True in [x.transport.connected for x in protocols]:
```
markrwilliams, line 861: I think this reads better:

``` python
if not any(p.transport.connected for p in protocols):
```

src/twisted/test/proto_helpers.py
```diff
@@ -834,3 +834,33 @@ def connectUNIX(self, address, factory, timeout=30, checkPID=0):
         Fake L{IReactorUNIX.connectUNIX}, that raises L{_connectException}.
         """
         raise self._connectException
+
+
+
+def waitUntilAllDisconnected(reactor, protocols):
```
markrwilliams, line 840: Could this be a private API?

markrwilliams (commented):
Minor stuff.  Soon to be merged!

src/twisted/test/proto_helpers.py
hawkowl: It is private API, it is in `test.proto_helpers` :)

src/twisted/test/proto_helpers.py
hawkowl: wow i did not know about `any()`. will fix.

hawkowl:
Builds are green, I'll merge (since @markrwilliams went to bed :) )

PR #555, opened 2016-10-06
[#8852] Move _dist3 into _setup

hawkowl:
https://tm.tl/#8852


markrwilliams:
The diff is big but the change is simple and [it needs to happen](https://github.com/twisted/twisted/pull/545#issuecomment-249102734).  I'm going to merge this.  Thanks!

PR #557, opened 2016-10-08
Port ckeygen to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8855


rodrigc:
@einarfd can you review this?  This is one more piece of conch ported to Python 3

src/twisted/conch/ssh/keys.py
```diff
@@ -144,6 +144,10 @@ def fromString(cls, data, type=None, passphrase=None):
         @rtype: L{Key}
         @return: The loaded key.
         """
+        if isinstance(data, unicode):
+            data = data.encode("ascii")
+        if isinstance(passphrase, unicode):
+             passphrase = passphrase.encode("ascii")
```
einarfd: Wouldn't utf8 be a better choice here? Wouldn't we expect people to use non ascii characters in their passphrase?

src/twisted/conch/ssh/keys.py
```diff
@@ -1024,10 +1028,12 @@ def toString(self, type, extra=None):
             is not part of the key itself.  For public OpenSSH keys, this is
             a comment.  For private OpenSSH keys, this is a passphrase to
             encrypt with.
-        @type extra: L{bytes} or L{None}
+        @type extra: L{bytes} or L{unicode} or L{None}
 
         @rtype: L{bytes}
         """
+        if isinstance(extra, unicode):
+            extra = extra.encode("ascii")
```
einarfd: Since this can be a passphrase, using utf8 seems like a good idea here as well.

src/twisted/conch/ssh/keys.py
rodrigc: @einarfd OK, I changed the patch to utf-8

PR #560, opened 2016-10-12
https://twistedmatrix.com/trac/ticket/8858

glyph:


src/twisted/logger/test/test_json.py
```diff
@@ -280,6 +280,23 @@ def test_observeWritesEmptyRecordSeparator(self):
...
+        A L{FileLogObserver} created by L{jsonFileLogObserver} writes failures
+        serialized as JSON text to a file when it observes events.
+        """
+        io = StringIO()
+        logger = Logger(observer=jsonFileLogObserver(io))
+        try:
+            1 / 0
+        except:
+            logger.failure("failed as expected")
+        print("AN_VALUE", io.getvalue())
```
wsanchez: remove print()

src/twisted/logger/test/test_json.py
```diff
@@ -280,6 +280,23 @@ def test_observeWritesEmptyRecordSeparator(self):
...
+        reader = StringIO(io.getvalue())
+        events = list(eventsFromJSONLogFile(reader))
+        self.assertEqual(len(events), 1)
+
```
wsanchez, line 313: This probably should check that the de-serialized event contains a `log_failure` key?

wsanchez: Also wondering whether we need to ensure that the failure's traceback is being captured in the JSON.

Julian: (Agree on both of those)

glyph:
The Windows build failure is logged here: https://twistedmatrix.com/trac/ticket/8870

src/twisted/logger/test/test_json.py
```diff
@@ -280,6 +280,40 @@ def test_observeWritesEmptyRecordSeparator(self):
...
+        except:
+            logger.failure("failed as expected")
+        reader = StringIO(io.getvalue())
+        events = list(eventsFromJSONLogFile(reader))
+        self.assertEqual(len(events), 1)
+        self.assertEqual(len(before), 1)
+        [event] = events
+        [beforeEvent] = before
+        self.assertIn("log_failure", event)
+        f = event["log_failure"]
```
markrwilliams: I'd prefer longer variable names for the Failures here.

src/twisted/logger/test/test_json.py
```diff
@@ -280,6 +280,40 @@ def test_observeWritesEmptyRecordSeparator(self):
...
+        b = beforeEvent["log_failure"]
+        self.assertIsInstance(f, Failure)
+        deserializedTraceback = f.getTracebackObject()
+        previousTraceback = b.getTracebackObject()
+        def checkOneTraceback(tb):
+            self.assertEqual(tb.tb_frame.f_code.co_filename.rstrip("co"),
```
markrwilliams: An additional check that confirms the `f_code.co_name` is the same as the test method's would be nice.  I guess you didn't include that, though, because it would make the code depend on the method's name?

src/twisted/topfiles/8858.bugfix
```diff
@@ -0,0 +1 @@
+twisted.logger.jsonFileLogObserver no longer emits non-JSON data into its file in the case of tracebacks; additionally, twisted.logger.formatEventAsClassicLogText now includes traceback information for the log event it is passing.
```
markrwilliams: twisted.logger.jsonFileLogObserver no longer emits non-JSON ~~data into its file in the case of~~ **tracebacks into its file**; additionally, twisted.logger.formatEventAsClassicLogText now includes traceback information for the log event it ~~is passing~~ **formats**.

markrwilliams (commented):
Minor, non-merge-blocking comments.

src/twisted/logger/test/test_json.py
glyph: Yeah. If one detail is accurate, that's close enough; this is not a test for every detail of traceback generation or serialization.

src/twisted/topfiles/8858.bugfix
glyph: Yes, better grammar.

src/twisted/logger/test/test_json.py
glyph: Sure.

PR #561, opened 2016-10-12
Port twisted.words.protocols.irc to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/6320

Moved from https://github.com/twisted/twisted/pull/534


src/twisted/words/protocols/irc.py
```diff
@@ -88,7 +88,7 @@ def parsemsg(s):
     trailing = []
     if not s:
         raise IRCBadMessage("Empty line.")
-    if s[0] == ':':
+    if s[0:1] == ':':
```
acabhishek942, line 92: why is this change necessary?

src/twisted/words/test/test_irc.py
```diff
@@ -20,6 +20,19 @@
 
 
 
+def assertEqualBufferValue(bufferValue, val):
+    # A buffer is always bytes, but sometimes
+    # we need to compare it to a utf-8 unicode string
```
acabhishek942: wouldn't it be better to add this as documentation in quotes than comments?

src/twisted/words/test/test_irc.py
```diff
@@ -1635,7 +1648,13 @@ def setUp(self):
...
 
     def check(self, s):
-        self.assertEqual(self.f.getvalue(), s)
+        bufferValue = self.f.getvalue()
+        if isinstance(s, unicode):
+            # Convert bufferValue from bytes
+            # to unicode, so we can compare it to
+            # a unicode string.
+            bufferValue = bufferValue.decode("utf-8")
+        self.assertEqual(bufferValue, s)
```
acabhishek942: why not use `assertEqualBufferValue` here also

src/twisted/words/test/test_irc.py
```diff
@@ -1602,9 +1615,9 @@ def _createHeartbeat():
         self.assertEqual(self.client.hostname, 'foo')
 
         # Pump the clock enough to trigger one LoopingCall.
-        self.assertEqual(self.transport.value(), '')
+        assertEqualBufferValue(self.transport.value(), '')
```
acabhishek942: this way looks a little out of place. I think it would be better to have a helper function to only convert to utf-8 unicode string and use `assertEqual`  with the utf-8 string

src/twisted/words/protocols/irc.py
```diff
@@ -2621,9 +2625,17 @@ def connectionMade(self):
...
-        basic.LineReceiver.dataReceived(self, data.replace('\r', ''))
+        data = data.replace('\r', '')
+        if isinstance(data, unicode):
+            data = data.encode("utf-8")
+        basic.LineReceiver.dataReceived(self, data)
+
 
     def lineReceived(self, line):
+        if bytes != str and isinstance(line, bytes):
+            # decode bytes from transport to unicode
```
acabhishek942, line 2636: My understanding is `bytes != str` is used to distinguish python2 and python3. If that is the case it could be done using `_PY3` for understanding purpose.

src/twisted/words/protocols/irc.py
rodrigc: It's not 100% necessary, but it still works.
I was playing around with byte strings in different parts of the code earlier,
and found that:

```
foo = b"blah"
foo[0]   # Returns b'b' on Python 2, and 102 on Python 3 
foo[0:1] # Returns b'b' on Python 2 and Python 3
```

For native strings:

```
bar = "bar"
bar[0]  #returns "b" on Python 2 and Python 3
bar[0:1] # returns "b" on Python 2 and Python 3
```

Kamekameha: This is never true though in Python 3 if `s` is actually an instance of `bytes`, as `b':'` is not really equal to `':'`. So either prefix the `':'` with the appropriate "b" or use the rather preferable `startswith` method (with its argument b-prefixed as well, of course) as you won't have to slice the `s` bytes. In fact, all string literals at least in this function wrt. `s` should be prefixed with "b" to actually be Python 2/3 compatible. That is of course, if `s` is certainly an instance of `bytes`, but that I don't really know.

On that note, I'd actually use `b' :' in s` instead of `s.find(b' :') != -1`, but that's just me nitpicking. :)

mjeveritt:
@rodrigc where did this get to in the end? I'm seeing some strange output from buildbot between py3.4 and py3.5 with identical buildbot and twisted versions .. so I'm guessing python is processing strings/encoding slightly differently between minor versions...
An artefact from https://github.com/buildbot/buildbot/issues/3694
See also https://twistedmatrix.com/trac/ticket/6320

PR #562, opened 2016-10-15
Automatically use resetDelay in ReconnectingClientFactory

ghost:


glyph:
Hi @ghost!

It looks like this never had a ticket in Trac, so nobody ever looked at it.  Sorry about that!  Also, `ReconnectingClientFactory` is old and destined for deprecation; the new hotness these days is `ClientService` is the thing you probably want these days.

PR #563, opened 2016-10-15
[#8861] Deferred.send

hawkowl:
https://twistedmatrix.com/trac/ticket/8861

### DIFF
--- a/src/twisted/internet/test/_awaittests.py.3only
+++ b/src/twisted/internet/test/_awaittests.py.3only
@@ -91,8 +91,8 @@ class AwaitTests(TestCase):
 
     def test_twoDeep(self):
         """
-        An exception in a coroutine wrapped with L{ensureDeferred} will cause
-        the returned L{Deferred} to fire with a failure.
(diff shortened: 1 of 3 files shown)


src/twisted/internet/test/_awaittests.py.3only
```diff
@@ -100,7 +100,7 @@ class AwaitTests(TestCase):
         async def runone():
```
markrwilliams, line 100: The docstring for `test_twoDeep` is wrong.  Maybe it could read something like this:

```
"""
A coroutine wrapped with L{ensureDeferred} that awaits a
L{Deferred} suspends its execution until the inner L{Deferred}
fires.
"""
```

markrwilliams (commented):
One comment on the change test's doc string.  Thanks for the quick bug fix!

src/twisted/internet/test/_awaittests.py.3only
hawkowl: That works, done.

hawkowl:
Spurious failures on the travis/windows builder. Will merge.

PR #566, opened 2016-10-17
8864 safe str unicode

markrwilliams:
http://twistedmatrix.com/trac/ticket/8864

Note this changes the API for `log.LogPublisher.msg` under Python 2.

### DIFF
--- a/src/twisted/python/reflect.py
+++ b/src/twisted/python/reflect.py
@@ -436,6 +436,13 @@ def safe_str(o):
             return o.decode('utf-8')
         except:
             pass
+    if not _PY3:
+        # On Python 2, attempt to encode a unicode representation
+        # first.
+        try:
+            return unicode(o).encode('ascii', 'backslashreplace')
+        except:
+            pass
     try:
         return str(o)
     except:
--- a/src/twisted/test/test_log.py
+++ b/src/twisted/test/test_log.py
@@ -390,8 +390,8 @@ class LogPublisherTests(LogPublisherTestCaseMixin,
 
     def test_singleUnicode(self):
         """
-        L{log.LogPublisher.msg} encodes Unicode as ``ascii`` with the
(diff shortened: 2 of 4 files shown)


src/twisted/python/reflect.py
```diff
@@ -436,6 +436,13 @@ def safe_str(o):
             return o.decode('utf-8')
         except:
             pass
+    if not _PY3:
+        # On Python 2, attempt to encode a unicode representation
+        # first.
+        try:
+            return unicode(o).encode('ascii', 'backslashreplace')
+        except:
```
Julian, line 444: Exception? Although it seems the other except blocks here are already setting this precedent.

src/twisted/test/test_log.py
```diff
@@ -390,20 +390,19 @@ def testMultipleString(self):
 
     def test_singleUnicode(self):
         """
-        L{log.LogPublisher.msg} does not accept non-ASCII Unicode on Python 2,
-        logging an error instead.
+        L{log.LogPublisher.msg} encodes Unicode as ``ascii`` with the
+        ``backslashreplace`` on Python 2.
```
Julian: I like the wording below better ("with `backslashreplace` error handling)

src/twisted/python/reflect.py
markrwilliams: I guess it's important to catch `KeyboardInterrupt`?  I did leave the bare `except:` for consistency's sake.

src/twisted/test/test_log.py
markrwilliams: Me too!  Should now be fixed

src/twisted/python/reflect.py
Julian: Fair enough (sounds dubious, but hey, ain't your dubious)

PR #567, opened 2016-10-20
[#8863] Make t.p.versions.Version case insensitive

hawkowl:
https://twistedmatrix.com/trac/ticket/8863

### DIFF
--- a/src/twisted/python/_setup.py
+++ b/src/twisted/python/_setup.py
@@ -223,12 +223,12 @@ def getSetupArgs(extensions=_EXTENSIONS):
         requirements = ["zope.interface >= 3.6.0"]
 
     requirements.append("constantly >= 15.1")
-    requirements.append("incremental >= 16.10.0")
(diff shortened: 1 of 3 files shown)


src/twisted/python/_setup.py
```diff
@@ -223,12 +223,12 @@ class my_build_ext(build_ext_twisted):
         requirements = ["zope.interface >= 3.6.0"]
 
     requirements.append("constantly >= 15.1")
-    requirements.append("incremental >= 16.10.0")
+    requirements.append("incremental >= 16.10.1")
```
markrwilliams, line 227: The builds are green, so this is certainly available on PyPI.

markrwilliams (commented):
Considering that the [implementation in incremental is clear, well-tested, documented, and released](https://github.com/hawkowl/incremental/pull/24#pullrequestreview-5343953), this is 100% ready to merge.

PR #569, opened 2016-10-23
8828 scripts/ckeygen can generate ecdsa keys

acabhishek942:
see http://twistedmatrix.com/trac/ticket/8828


acabhishek942:
@hawkowl @glyph  Can you please review this?

src/twisted/conch/ssh/keys.py
```diff
@@ -382,7 +382,7 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
             except ValueError:
                 raise BadKeyError('invalid DEK-info %r' % (lines[2],))
 
-            if cipher == b'AES-128-CBC':
+            if cipher in (b'AES-128-CBC', b'AES-256-CBC'):
```
Lukasa, line 385: Why has `AES-256-CBC` been added in this PR? It seems unrelated.

src/twisted/conch/ssh/keys.py
```diff
@@ -416,6 +416,11 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
             b64Data = b''.join(lines[1:-1])
             keyData = decodebytes(b64Data)
 
+        if kind == b'EC':
+            #ECDSA keys don't need base64 decoding which is required
```
Lukasa: Nit: space after `#`.

src/twisted/conch/ssh/keys.py
```diff
@@ -1202,11 +1205,33 @@ def _toString_OPENSSH(self, extra):
...
+                    ) + extra).strip()
+
             b64Data = encodebytes(self.blob()).replace(b'\n', b'')
             if not extra:
                 extra = b''
             return (self.sshType() + b' ' + b64Data + b' ' + extra).strip()
         else:
+
+            if self.type() == 'EC':
+                #EC keys has complex ASN.1 structure hence we do this this way.
```
Lukasa: Nit: space after `#`.

src/twisted/conch/ssh/keys.py
```diff
@@ -1202,11 +1205,33 @@ def _toString_OPENSSH(self, extra):
         """
         data = self.data()
         if self.isPublic():
+            if self.type() == 'EC':
+                if not extra:
+                    extra = b''
```
Lukasa, line 1210: Is there any test for this? I wasn't able to find any.

Lukasa: In particular, I note that the RSA/DSA keys append `extra` with a space between the key ad the `extra`, but the code here does not. That's a cause for concern, especially when the behaviour of this code appears to be untested.

src/twisted/conch/ssh/keys.py
```diff
@@ -1202,11 +1205,33 @@ def _toString_OPENSSH(self, extra):
...
             b64Data = encodebytes(self.blob()).replace(b'\n', b'')
             if not extra:
                 extra = b''
             return (self.sshType() + b' ' + b64Data + b' ' + extra).strip()
         else:
+
+            if self.type() == 'EC':
+                #EC keys has complex ASN.1 structure hence we do this this way.
+                if not extra:
+                    #unencrypted private key
```
Lukasa: Nit: space after `#`.

Lukasa (changes requested):
So this seems like a good start. I lack some of the relevant SSH knowledge to be sure how right this is, but I was able to find some stuff worth addressing anyway!

src/twisted/conch/ssh/keys.py
acabhishek942: `AES-256-CBC` is required here because `serialization.BestAvailableEncryption` uses `AES-256-CBC` now. without this code raises a BadKeyError with unknown cipher

acabhishek942:
@Lukasa I added tests covering the key generation methods. Please take a look

Lukasa:
So this is much better! Unfortunately, the `run()` method and all its content is still uncovered. Can you write a test to cover that?

acabhishek942:
I added tests for `run()` method but haven't tested them locally as tests kept skipping for some reason even though I have `cryptography` and `pyasn1` installed. Hope the build passes here

Lukasa:
Hrm, these failures all look...wrong. Do you want to rebase onto the current master just to see if any of them get cleared up?

acabhishek942:
@Lukasa I rebased onto current trunk and still there are failures. Did I break these tests?

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -63,6 +64,46 @@ def setUp(self):
         self.patch(sys, 'stdout', self.stdout)
 
 
+
+    def test_runecdsa(self):
+        sys.argv = ['ckeygen', '-t', 'ecdsa', '-f', 'test_ecdsa', '--no-passphrase']
+        run()
```
Lukasa: So I suspect these tests are causing your breakage. This is changing the entire runtime environment of the test, which is a bit alarming. It would almost certainly be much better to execute ckeygen in a subprocess rather than do this.

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -63,6 +65,46 @@ def setUp(self):
...
+        self.assertTrue(pubKey.isPublic())
+
+
+    # def test_runBadKeytype(self):
+    #     filename = self.mktemp()
+    #     with self.assertRaises(SystemExit) as em:
+    #         subprocess.call(['ckeygen', '-t', 'foo', '-f', filename])
+    #     self.assertEqual(
+    #         'Key type was foo, must be one of: rsa, dsa, ecdsa',
+    #         em.exception.args[0])
```
acabhishek942: @Lukasa travis build is green but I still don't have full coverage because of this commented out test which I cannot figure out. Can you please take a look?

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -44,7 +44,7 @@ class GeneralOptions(usage.Options):
                 ['showpub', 'y', 'Read private key file and print public key.']]
 
     compData = usage.Completions(
-        optActions={"type": usage.CompleteList(["rsa", "dsa"])})
+        optActions={"type": usage.CompleteList(["rsa", "dsa", "ecdsa"])})
```
moshez: There's three different places the list "rsa", "dsa", "ecdsa" appears. This seems to be RY 😄 

Would it be better to have one place that does it?

Somewhere at the top of the file:

```
_supported_types = dict() 
def _support(tpname):
       def decorate(func):
              _supported_types[tpname] = func
              return func
    return tpname
```

and then:

```
@_support('ecdsa')
def generateECDSAkey(options):
      pass # code here
```

Here you would have

```
optActions = {'type': usage.CompleteList(list(_supported_types.keys())) }
```

below you would have

```
if options['type'] in _supported_types:
    _supported_types[option['type']](options)
else:
    sys.exit('Key type was %s, must be one of %s',
                  options['type'], ', '.join(_supported_list.keys()))
```

Especially in the wake of people starting to worry about post-quantum cryptography, I think it's important to make it more straight forward to add more types (all three listed here, as they are based on abelian groups, are vulnerable to Shor's algorithm, so we _know_ that nothing here is quantum-resistant).

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -220,7 +245,7 @@ def _saveKey(key, options):
     @param options:
     @type options: L{dict}
     """
-    keyTypeName = key.type().lower()
+    keyTypeName = 'ecdsa' if key.type() == 'EC' else key.type().lower()
```
moshez: This seems like it would accumulate more special cases with more key types. Should we give up and just use an explicit dictionary?

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -63,6 +65,46 @@ def setUp(self):
...
+        subprocess.call(['ckeygen', '-t', 'rsa', '-f', filename, '--no-passphrase'])
+        privKey = Key.fromFile(filename)
+        pubKey = Key.fromFile(filename + '.pub')
+        self.assertEqual(privKey.type(), 'RSA')
+        self.assertTrue(pubKey.isPublic())
+
+
+    # def test_runBadKeytype(self):
+    #     filename = self.mktemp()
+    #     with self.assertRaises(SystemExit) as em:
```
moshez: This seems unlikely to work. Possibly change subprocess.call to subprocess.check_call, and use subprocess.CalledProcessError? subprocess.call will usually not raise exceptions (unless it has problems actually starting the command.)

moshez (changes requested):
Thanks for working on this! I am a big fan of Elliptic Curves, and I think supporting it is important.

The only change which is really needed is to fix the test. Hopefully the comment I made is useful. Other than that, address the style comments if you think they're useful.

acabhishek942:
thanks @moshez for the review. I have tested ckeygen script with different parameters and it works, hopefully the builds will be green.

acabhishek942:
the builds are finally green. @Lukasa @moshez can you please review this

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -23,13 +23,21 @@
 
 
 
+_supportedKeytypes = dict()
```
Lukasa: Nit: `supportedKeyTypes` would be more consistent with the usage elsewhere.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -96,12 +105,13 @@ def handleError():
     raise
 
 
-
+@_support('rsa')
```
Lukasa: I don't love this method name. Might be better to call it `_keyGenerator('rsa')`?

Lukasa: While I'm here I should note that the decorator being added to these methods changes their function signature. The decorator should probably use [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps) to fix that up.

src/twisted/conch/scripts/ckeygen.py
Lukasa: I'm with @moshez here: let's have an explicit dictionary registry that handles this mapping.

Lukasa (changes requested):
Ok this is really much closer to making me happy. Thanks so much for improving the test coverage so much: you're also blazing a trail to improve the test coverage of the rest of ckeygen, which is really great.

I have a few notes inline that I'd like to see addressed before this can be merged.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -23,13 +23,24 @@
 
 
 
+supportedKeyTypes = dict()
+def _keyGenerator(keyType):
+    def assignkeygenerator(keygenerator):
+        supportedKeyTypes[keyType] = keygenerator
+        def wrapper(*args, **kwargs):
```
Lukasa, line 31: Any reason we're still not using `functools.wraps`?

Lukasa (changes requested):
Still a note about `functools.wraps`.

Lukasa (approved):
So I think I'm happy with these changes. I'm ready for them to be merged if we can get a clean test run. For that reason, I've merged your commits into a branch on this repository just so we get to see the buildbots as well.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -23,13 +24,25 @@
...
 
 
+supportedKeyTypes = dict()
+def _keyGenerator(keyType):
+    def assignkeygenerator(keygenerator):
+        supportedKeyTypes[keyType] = keygenerator
+        @wraps(keygenerator)
+        def wrapper(*args, **kwargs):
+            keygenerator(*args, **kwargs)
+        return wrapper
```
Lukasa, line 34: Sorry, this isn't quite right and that's why you're short on coverage. `@wraps` should go on `assignKeyGenerator`, not `wrapper`, and you shouldn't need `wrapper` at all.

Lukasa (changes requested):
There are some issues with the wrapper I'm afraid.

src/twisted/conch/scripts/ckeygen.py
acabhishek942: decorators with arguments works a little differently http://stackoverflow.com/questions/5929107/python-decorators-with-parameters, but I am quite why those lines are not covered I will try to figure this out tomorrow.

acabhishek942: I did a little digging here with the following snippets

```
supportedKeyTypes = dict()
def _keyGenerator(keyType):
    def decorator(method):
        supportedKeyTypes[keyType] = method
        @wraps(method)
        def wrapper(*args, **kwargs):
            print("foo")
            return method(*args, **kwargs)
        return wrapper
    return decorator
```

```
@_keyGenerator('rsa')
def generateRSAkey(options):
    # key generation code here
```

after executing these two code blocks `wrapper` method is not executed, and is executed only after an explicit call to `generateRSAkey` as following.

```
In [3]: generateRSAkey({'bits': '2048'})
Out[3]: foo

explicit call to keygenerator functions is currently missing in ckeygen hence the coverage issue, please guide how to proceed here
```

Lukasa: Yup, so we will need to hit those I'm afraid. =(

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -74,7 +74,12 @@ def run():
     if options['type']:
         if options['type'].lower() in supportedKeyTypes:
             print('Generating public/private %s key pair.' % (options['type']))
-            supportedKeyTypes[options['type'].lower()](options)
+            if options['type'].lower() == 'rsa':
+                generateRSAkey(options)
+            elif options['type'].lower() == 'dsa':
+                generateDSAkey(options)
+            else:
+                generateECDSAkey(options)
```
Lukasa: Ok, so while it's nice that this improves the coverage, I'm a bit confused now. I think your decorator is wildly over-thinking things.

I think you can achieve the coverage you want by adding the decorator _wrapper_ to the dictionary, rather than adding the raw function itself. That would let you keep using the dictionary here, but also fix your coverage problem.

Lukasa (changes requested):
Let's try to get our full coverage without removing the value the decorator adds.

acabhishek942:
this is still not hitting full coverage

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -30,7 +30,7 @@ def assignkeygenerator(keygenerator):
         supportedKeyTypes[keyType] = keygenerator
```
Lukasa: Sorry, what I meant is that this line should be moved underneath the declaration of `wrapper`, and then should be changed to `supportedKeyTypes[keyType] = wrapper`.

Lukasa (changes requested):

acabhishek942:
finally codecov agreed with this. I hope this patch is acceptable now

Lukasa (approved):
I am now happy with this. Thanks so much for all the work @acabhishek942! :sparkles: :cake: :sparkles:

PR #572, opened 2016-10-25
Add a PIDFile class that can be used as a context

glyph:
https://twistedmatrix.com/trac/ticket/8659


src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        @return: The contained process ID, or C{None} if the PID file does not
+            exist.
+        @rtype: L{int}
+
+        @raise EnvironmentError: If this PID file cannot be read.
+        @raise ValueError: If this PID file's content is invalid.
+        """
+        pidString = b""
+        try:
+            for pidString in self.filePath.open():
```
markrwilliams: This will retain a file descriptor for an indeterminate amount of time under PyPy.  Best to use a context manager!

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        Read the process ID stored in this PID file.
+
+        @return: The contained process ID, or C{None} if the PID file does not
+            exist.
+        @rtype: L{int}
+
+        @raise EnvironmentError: If this PID file cannot be read.
+        @raise ValueError: If this PID file's content is invalid.
+        """
+        pidString = b""
```
markrwilliams, line 133: Is this the empty string so that an empty file raises the same error as a file with a non-integral PID?  Would it be better to raise a more informative error message in the case of an empty file?  I think I'd appreciate the extra help debugging such a weird occurrence.

glyph:
Needs a topfile so the builders will be happy…

glyph:
Also, looks like the Windows build is failing?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+    PID file.
+
+    Manages a file that remembers a process ID.
+    """
+
+    _log = Logger()
+
+
+    @staticmethod
+    def format(pid=None):
```
moshez: Why does this need to allow None in?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        else:
+            return b""
+
+
+    def __init__(self, filePath):
+        """
+        @param filePath: The path to the PID file on disk.
+        @type filePath: L{IFilePath}
+        """
+        self.filePath = IFilePath(filePath)
```
moshez: If filePath is already supposed to implement IFilePath, why are we doing adaptation here?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+                break
+        except OSError as e:
+            if e.errno == errno.ENOENT:  # No such file
+                return None
+            raise
+
+        return int(pidString)
+
+
+    def write(self, pid=None):
```
moshez: Can this be separated into

```
def writeMyPID(self):
```

and

```
def write(self, pid):
       # Does not allow None here
```

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        @type filePath: L{IFilePath}
+        """
+        self.filePath = IFilePath(filePath)
+
+
+    def read(self):
+        """
+        Read the process ID stored in this PID file.
+
+        @return: The contained process ID, or C{None} if the PID file does not
```
moshez: This can be a dangerous API -- users must explicitly test for None, and are likely to not do this in test scenarios. Can we have this API instead raise a custom (or standard) exception like ValueError to indicate "no PID"?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        pid = self.read()
+        if pid is None:
+            return False
+
+        try:
+            kill(pid, 0)
+        except OSError as e:
+            if e.errno == errno.ESRCH:  # No such process
+                self._log.info("Removing stale PID file: {log_source}")
+                self.remove()
```
moshez: Potentially weird side-effect: a boolean function modifying global (filesystem) state. I realize an invalid PID is equivalent to an empty file in some respect, but this feels like it can cause hard-to-debug problems.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+
+class NonePIDFile(PIDFile):
+    """
+    PID file implementation that does nothing.
+
+    This is meant to be used as a "active None" object in place of a PID file
+    when no PID file is desired.
+    """
+
+    def __init__(self):
```
moshez, line 208: This is redundant.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        Exit a context using this PIDFile.
+
+        Removes the PID file.
+        """
+        self.remove()
+
+
+
+class NonePIDFile(PIDFile):
```
moshez: Why the inheritance? Should we have an explicit interface for a PIDFile?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,202 @@
...
+    def __enter__(self):
+        return self
+
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        pass
+
+
+
+nonePIDFile = NonePIDFile()
```
moshez, line 241: Should this class be private, or do we expect allowing people to construct new nonefiles?

src/twisted/application/runner/_runner.py
```diff
@@ -46,48 +42,41 @@ def __init__(self, options):
...
+        This requires that L{RunnerOptions.pidFile} also be specified;
         exit with L{ExitStatus.EX_USAGE} if kill is requested with no PID file.
         """
-        pidFilePath = self.options.get(RunnerOptions.pidFilePath)
+        pidFile = self.options.get(RunnerOptions.pidFile)
 
         if self.options.get(RunnerOptions.kill, False):
-            if pidFilePath is None:
-                exit(ExitStatus.EX_USAGE, "No PID file specified")
+            if pidFile is None:
```
moshez, line 71: This will never be None, right? It will be the NonePIDFile?

src/twisted/application/runner/_runner.py
```diff
@@ -201,9 +169,9 @@ class RunnerOptions(Names):
         Corresponding value: L{IReactorCore}.
     @type reactor: L{NamedConstant}
 
-    @cvar pidFilePath: The path to the PID file.
-        Corresponding value: L{IFilePath}.
-    @type pidFilePath: L{NamedConstant}
+    @cvar pidFile: The path to the PID file.
```
moshez: It's not "the path" anymore, right?

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        L{PIDFile.isRunning} returns true for a process that does exist.
+        """
+        pidFile = PIDFile(DummyFilePath())
+        pidFile.write(1337)
+
+        def kill(pid, signal):
+            return  # Don't actually kill anything
+
+        self.patch(_pidfile, "kill", kill)
```
moshez, line 168: I don't know how I feel about this, but I'm not sure I have a better alternative.

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -0,0 +1,291 @@
...
+
+        @note: In this case, we try killing C{init}, which is process #1 on
+        POSIX systems, so this test is not portable.  C{init} should always be
+        running and should not be killable by non-root users.
+        """
+        if SYSTEM_NAME != "posix":
+            raise SkipTest("This test assumes POSIX")
+
+        pidFile = PIDFile(DummyFilePath())
+        pidFile.write(1)  # PID 1 is init on POSIX systems
```
moshez, line 242: Running the tests in a typical docker container would not expose a problem with EPERM. I suggest using the same patching strategy used above.

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -0,0 +1,291 @@
...
+
+        def useContext():
+            with pidFile:
+                pass
+
+        self.assertRaises(AlreadyRunningError, useContext)
+
+
+
+class DummyFilePath(FilePath):
```
moshez, line 403: My usual attitude is that using real files is not a big problem since filesystems are usually reliable and you can have files that are local to the test environment and more-or-less guaranteeed-uniguely name. Not sure if this stub is worth the extra complexity?

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -0,0 +1,291 @@
...
+    Stub for L{twisted.python.filepath.FilePath} which returns a stream
+    containing the given data when opened.
+    """
+
+    def __init__(self, content=None):
+        self.setContent(content)
+
+
+    def open(self, mode="r"):
+        if not self._exits:
```
moshez: Is this supposed to be "exists"? (Ditto for below)

src/twisted/application/runner/test/test_runner.py
```diff
@@ -367,6 +375,58 @@ def exited(options):
 
 
 
+class DummyRunner(Runner):
```
moshez, line 378: Do we need inheritance here?

src/twisted/application/runner/test/test_runner.py
```diff
@@ -367,6 +375,58 @@ def exited(options):
...
+    def startReactor(self):
+        self.calledMethods.append("startReactor")
+
+
+    def reactorExited(self):
+        self.calledMethods.append("reactorExited")
+
+
+
+class DummyPIDFile(NonePIDFile):
```
moshez, line 407: Ditto -- do we need inheritance?

moshez (approved):
Thanks a lot for doing this -- I am a big fan of this refactor. This is much clearer than it used to be.

I made a bunch of comments, but I don't think any single one is a show-stopper. Please address those you think that have merit and either merge directly or put up for review if you think the changes are significant.

src/twisted/application/runner/_pidfile.py
wsanchez: Yes… that was because… `write()` takes a `None` pid and probably at one point that made sense but there's no longer a need for that here.  Removing.

wsanchez: 6098f0c83bb1f1b8153fe04f6ed2ab1f8c7c882d

src/twisted/application/runner/_pidfile.py
wsanchez: I guess I don't know the right way to right "object adaptable to IFilePath" in the docstring without sounding too verbose, or what the best practice for that is.

src/twisted/application/runner/_pidfile.py
wsanchez: Agreed.  I think I like custom, so we're very explicit.
- 5619da41be4a5f21a67c76dcdac25409e8561446  - Adds InvalidPIDFileError custom exception for the cases where the PID file has invalid contents.
- 71781a7991f38faca4d26e6564ecc6c73ebd0067 - Adds NoPIDFound custom exception for when there is no PID (non-existing PID file).

src/twisted/application/runner/_pidfile.py
wsanchez: 062493182c003280c20b016ca343bf6ee9a3028c - Adds writeRunningPID()

src/twisted/application/runner/_pidfile.py
wsanchez: I was conflicted about this myself; I prefer for changed to the underlying file to be explicit (ie. limited to `write()`, `remove()` and the like).

The reason I did this here anyway is that the PID file does not reference a valid PID right now, and there's a risk that another PID may spin up with that PID later, which is potentially an even nastier problem, so I felt like being proactive about it.

Another option here is to raise an exception and force the caller to deal with this.  I didn't go for that because I suspect that this would mean that all callers should basically add more code to catch this exception and… remove the file; otherwise, this case would raise an uncaught exception.

Thoughts?

src/twisted/application/runner/_pidfile.py
glyph: The evolving convention here, I think, is … don't do adaptation "inside" an API like this.  Require that the caller pass an `IFilePath` themselves, and if they want to do adaptation on their end, they can do it.

Adaptation is convenient, but it introduces coupling to the global state of the adapter registry, which in turn potentially has implications for threading, and generally we don't have great tools for telling us why an adaptation did or didn't work.  Generally the places where I find adaptation still makes sense are big heterogenous datastructures where a "walker" does adaptation over fragments of it (like twisted.web.template); if it's just an individual parameter, I almost never do adaptation directly like this any more.

src/twisted/application/runner/_pidfile.py
wsanchez: OK so:

4351561fb484d3f5919711e29336361128f1406a - Adds StalePIDFileError and raises that if the PID file is stale.  Caller handles it.  If we need to re-encapsulte that, options include a wrapper function, or a kwarg asking for this explicitly.

src/twisted/application/runner/_pidfile.py
wsanchez: I did that for isinstance to work, but an interface would be more cool.

src/twisted/application/runner/_pidfile.py
wsanchez: No, because superclass takes an argument…

src/twisted/application/runner/_pidfile.py
wsanchez: Everything here is currently private…

src/twisted/application/runner/_pidfile.py
wsanchez: 498f517795d5635152447f0bc768a94c97124625 removes adaptation

src/twisted/application/runner/_runner.py
wsanchez: Presently it could be None, but the next ticket will rework options handling, so let me punt on this issue until then.

src/twisted/application/runner/_runner.py
wsanchez: Correct.  8a772253d32ec2db6f19f211fbbb26eeca061229

src/twisted/application/runner/test/test_pidfile.py
wsanchez: I feel OK about this.  :-)

src/twisted/application/runner/test/test_pidfile.py
wsanchez: That would make this identical to `test_isRunningNotAllowed`.

src/twisted/application/runner/test/test_pidfile.py
wsanchez: Yes: c19e9d847afd2a8337bccd0e3cdf7b5c2d846e74

src/twisted/application/runner/test/test_runner.py
wsanchez: No, it's just expedient.

src/twisted/application/runner/test/test_runner.py
wsanchez: No, it's just expedient.

src/twisted/application/runner/_pidfile.py
wsanchez: IPIDFile added in d36c4695c6fdd246087afad3351f9577c8ae670f, inheritance removed.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,267 @@
...
+        @param filePath: The path to the PID file on disk.
+        @type filePath: L{IFilePath}
+        """
+        self.filePath = filePath
+
+
+    def read(self):
+        pidString = b""
+        try:
+            for pidString in self.filePath.open():
```
markrwilliams: This will leak a file descriptor/handle under PyPy and produce a `ResourceWarning` under Python 3.  Wrapping the `open` call with context manager would be best though it'll make the indentation even deeper.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,267 @@
...
+                break
+        except OSError as e:
+            if e.errno == errno.ENOENT:  # No such file
+                raise NoPIDFound("PID file does not exist")
+            raise
+
+        try:
+            return int(pidString)
+        except ValueError:
+            raise InvalidPIDFileError("non-integer PID value in PID file")
```
markrwilliams: It'd be nice to see the repr of the invalid value in the exception message.

src/twisted/application/runner/_pidfile.py
markrwilliams: Doesn't this no-argument initializer mean that `NonePIDFile` doesn't implement the `IPIDFile` interface?  If so, is this OK because we prefer to verify that an object provides an interface to verifying that its class implements it?

If this should take an argument to ensure that the class implements `NonePIDFile`, is a default value of `None` acceptable and a note in the doc string that the argument is ignored?

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -0,0 +1,416 @@
...
+    """
+    Tests for L{NonePIDFile}.
+    """
+
+    def test_interface(self):
+        """
+        L{NonePIDFile} conforms to L{IPIDFile}.
+        """
+        pidFile = NonePIDFile()
+        verifyObject(IPIDFile, pidFile)
```
markrwilliams, line 338: I guess this answers the implements vs provides verificatio :)

src/twisted/application/runner/test/test_pidfile.py
markrwilliams: More generally this will fail if the tests run as `root` (which is, of course, generally nuts.)  You could add a skip for this test that checks if it's running on a POSIX system and the process' effective user ID is 0 to avoid spurious failures in Docker containers.

markrwilliams (commented):
Resource leak, more explanation in a exception message, some interface questions, and a suggestion for testing in Docker containers.

src/twisted/application/runner/_pidfile.py
wsanchez: Fixed: bdd6fac18c7c6abc150bcc6ffcc1cdd4eff3b068

src/twisted/application/runner/_pidfile.py
wsanchez: Fixed: a486e26bb287daef819be3f864fbe7c27d502b19
Also added a test, and noticed one of the read tests wasn't right, so fixed that as well.

src/twisted/application/runner/test/test_pidfile.py
wsanchez: Yeah, the interface doesn't define the initializer; just the behavior of the instantiated object.

wsanchez: geteuid test skippage added in 6fe8c9d28a10480191461b8e622db5cd7e2316b8

wsanchez:
Reverted…

PR #573, opened 2016-10-26
https://tm.tl/8870

glyph:


glyph:
I believe @markrwilliams wanted to be tagged here. (But if anyone else wants to review, that would be great.)

PR #577, opened 2016-10-30
[#8878] Already failed Deferred being awaited

hawkowl:
https://twistedmatrix.com/trac/ticket/8878


moshez (approved):
Thanks for doing this, it's nice to have better support of the async syntax.

Left a comment about a possible clarity improvement. Please address iff you think it's a good idea. Either way, please merge 😄

PR #565, opened 2016-10-15
[#8862] Use extras in tox.ini

derwolfe:
https://twistedmatrix.com/trac/ticket/8862#ticket

PR #578, opened 2016-10-31
8862 - tox extras

derwolfe:
https://twistedmatrix.com/trac/ticket/8862#ticket

Builds on #565 and #568.

(I would have reopened #565, but I github's UI didn't seem to allow it.)

@rodrigc I updated this to try to use the new tox feature.


rodrigc:
@derwolfe I can confirm that this works, but it uses features from the latest 2.4 version of tox.
When I ran this with tox 2.3.1, the tests still ran but not all the dependencies installed.
With tox 2.4.1, the tests ran properly and all the dependencies installed.

rodrigc:
See: https://github.com/twisted/twisted/pull/568

PR #579, opened 2016-10-31
Port twisted.names.authority.BindAuthority to Py3

hynek:
Underway I had to fix a few bugs/annoyances.  Most notably:

- Remove a (debug!?) print statement when adding records.
- Reflect explicit $ORIGIN in instance.
- Add discouraging docstrings to public helper methods that should be private.
  These should be deprecated ASAP.
- Remove a bunch of commented out debug print statements.

I wrote a bunch of tests (the class was completely untested so far), but want to see Travis/coverage before writing more.


glyph:
There are a couple of twistedchecker failures - can you fix those? Thanks!

hynek:
Look at all those green build jobs!

adiroiban:
Thanks for the patch. The release notes only talks about porting to python3.

Since in this branch you have also fixed a few bugs, shouldn't we advertise this in the release notes?

src/twisted/names/authority.py
```diff
@@ -14,14 +14,24 @@
...
 from twisted.python import failure
-from twisted.python.compat import execfile
+from twisted.python.compat import execfile, nativeString, _PY3
+from twisted.python.filepath import FilePath
 
 
-def getSerial(filename = '/tmp/twisted-names.serial'):
-    """Return a monotonically increasing (across program runs) integer.
+
+def getSerial(filename='/tmp/twisted-names.serial'):
```
adiroiban, line 22: since this is targeting python3, is filename a native string, bytes or unicode?

with the move to python3 I think that it would help to be explicit if something is text or data.

src/twisted/names/authority.py
```diff
@@ -34,28 +44,14 @@ def getSerial(filename = '/tmp/twisted-names.serial'):
...
-#            self._meth = super(LookupCacherMixin, self)._lookup
-#
-#        if self._cache.has_key((name, cls, type)):
-#            return self._cache[(name, cls, type)]
-#        else:
-#            r = self._meth(name, cls, type, timeout)
-#            self._cache[(name, cls, type)] = r
-#            return r
-    serial = serial + ('%02d' % (ID,))
+        serialFile.write('%s %d' % (serial, id))
```
adiroiban: can we use FIlePath here? I think that it should take care of handling the path encoding issues.

src/twisted/names/authority.py
```diff
@@ -161,25 +156,32 @@ def _lookup(self, name, cls, type, timeout = None):
                 else:
                     ttl = default_ttl
 
-                if record.TYPE == dns.NS and name.lower() != self.soa[0].lower():
+                if record.TYPE == dns.NS and \
```
adiroiban: I think that you should use parenthesis to break long lines .... but twistechecker should have complain about this

src/twisted/names/authority.py
hynek: ISTM that for paths a filename is OK to be a native string.  Are there any precedences?

I’d also like to point out, that I just touched this function to make twistedchecker shut up.  Starting to fix it up seems out of scope?  NB I’m not asking because I’m lazy, I’m asking because I’m not sure.

hynek:
I thought about the release notes thing but wasn’t sure how to phrase it (especially because the bugs are rather subtle).

src/twisted/names/authority.py
```diff
@@ -34,28 +44,14 @@ def getSerial(filename = '/tmp/twisted-names.serial'):
         os.umask(o)
 
     with open(filename, 'r') as serialFile:
-        lastSerial, ID = serialFile.readline().split()
+        lastSerial, id = serialFile.readline().split()
```
glyph: `id` is a builtin, which is why this was previously capitalized :).

Perhaps `zoneID`?

src/twisted/names/authority.py
```diff
@@ -34,28 +44,14 @@ def getSerial(filename = '/tmp/twisted-names.serial'):
...
-#        if self._cache.has_key((name, cls, type)):
-#            return self._cache[(name, cls, type)]
-#        else:
-#            r = self._meth(name, cls, type, timeout)
-#            self._cache[(name, cls, type)] = r
-#            return r
-    serial = serial + ('%02d' % (ID,))
+        serialFile.write('%s %d' % (serial, id))
+
+    serial = serial + ('%02d' % (id,))
```
glyph: Nitpick: `.format` instead of `%` because it's \~fancier\~?

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
 
-        with open(filename, 'rb') as f:
-            lines = f.readlines()
+        Load records from C{filename}.
+
+        @param filename: file to read from
+        @type filename: L{bytes}
+        """
+        fp = FilePath(filename)
+        # XXX - not the best way to set an origin. It can be set using $ORIGIN
```
glyph: Per policy, XXX and FIXME comments should be links to tickets.

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
         lines = self.stripComments(lines)
         lines = self.collapseContinuations(lines)
         self.parseLines(lines)
 
 
     def stripComments(self, lines):
-        return [
-            a.find(';') == -1 and a or a[:a.find(';')] for a in [
+        """
+        This is a private API; do not use in application code.
```
glyph: Can you make this method private and then add a deprecated public alias?  Deprecation should be communicated structurally in a way that automated tooling can deal with, not with ad-hoc doc comments.

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
                 b.strip() for b in lines
             ]
-        ]
+        )
 
 
     def collapseContinuations(self, lines):
-        L = []
+        """
+        This is a private API; do not use in application code.
```
glyph: Same as above comment.

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
-        return filter(None, L)
+                    l[-1] += b' ' + line
+        return filter(None, (line.split() for line in l))
 
 
     def parseLines(self, lines):
-        TTL = 60 * 60 * 3
-        ORIGIN = self.origin
+        """
+        This is a private API; do not use in application code.
```
glyph: See above about private APIs.

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
+
+        # If the origin changed, reflect that within the instance.
+        self.origin = origin
 
 
     def addRecord(self, owner, ttl, type, domain, cls, rdata):
-        if not domain.endswith('.'):
-            domain = domain + '.' + owner
+        """
+        This is a private API; do not use in application code.
```
glyph: OK hang on a second. Are these APIs really private?  It feels like maybe somebody is using them.  Maybe we can just delete the deprecations in this branch and do that separately?

src/twisted/names/test/test_names.py
```diff
@@ -1038,3 +1040,161 @@ def test_lookupAddress(self):
...
+                    ip,
+                    604800,
+                ),
+                rr.payload,
+            )
+
+
+    def test_aaaaRecords(self):
+        """
+        AAAA records are loaded correctly.
```
glyph: As per https://jml.io/pages/test-docstrings.html, "correctly" is a non-word :).

src/twisted/names/test/test_names.py
```diff
@@ -1038,3 +1040,161 @@ def test_lookupAddress(self):
...
+                u"2001:db8:10::1",
+                604800,
+            ),
+            rr.payload,
+        )
+
+
+    def test_mxRecords(self):
+        """
+        MX records are loaded correctly.
```
glyph: https://jml.io/pages/test-docstrings.html again

src/twisted/names/test/test_names.py
```diff
@@ -1038,3 +1040,161 @@ def test_lookupAddress(self):
...
+            dns.Record_CNAME(
+                name="example.com", ttl=604800,
+            ),
+            rr.payload,
+        )
+
+
+    def test_invalidRecordClass(self):
+        """
+        Raises NotImplementedError on invalid records.
```
glyph: This sentence should have a subject.

glyph (commented):
Thanks for doing this, @hynek . This is generally a great branch and I like most of the cleanups.  However, if you could address the few nitpicks it would be great.

If you can push these changes to a branch on Twisted itself, then the all the buildbots should run and I _think_ the coverage issues will be addressed.  If not, please pacify codecov by adding tests for any uncovered lines.

src/twisted/names/authority.py
hynek: twistedchecker complained about ID

src/twisted/names/authority.py
hynek: How am I supposed to document them to make twistedchecker shut up?

src/twisted/names/authority.py
glyph: Technically (naming-standard-wise) it is correct to complain; it should just also be complaining about `id`.  `zoneID` ought to be OK with it though.

src/twisted/names/authority.py
glyph: Fixing it is definitely out of scope. And we should probably figure out how to make twistedchecker not complain about stuff that is merely adjacent to your diff rather than part of it...

src/twisted/names/authority.py
glyph: That seems like an unrelated question, but... txchecker-travis is passing :).

You already have a docstring here, so if you just delete the line `This is a private API; do not use in application code.` it should still be fine with it.

Am I misunderstanding the question?

src/twisted/names/authority.py
```diff
@@ -270,141 +289,253 @@ def setupConfigNamespace(self):
...
+        @param type: record type
+        @type type: L{str}
+
+        @param cls: record class
+        @type cls: L{str}
+
+        @param rdata: record data
+        @type rdata: L{list} of L{bytes}
+
+        @return: nothing
```
glyph: FYI: there's no `return` statement in the body of the function, so twistedchecker and pydoctor ought not to complain if you just omit `@return` entirely.  And anywhere twistedchecker *lets* us omit stuff we probably should :).  (If I'm wrong about this please correct me).

src/twisted/names/authority.py
hynek: but it’s ~slower~? :)

wsanchez:
Coverage is failing.

At least one case is directly relevant to your changes: https://codecov.io/gh/twisted/twisted/compare/f520d35bfd552f0da91b248171d25789170110ae...1dd9c449504b07652af650382fc46b566fcd8546#7372632F747769737465642F6E616D65732F617574686F726974792E7079-86

Others are not new but triggering due to reformatting of code:
https://codecov.io/gh/twisted/twisted/compare/f520d35bfd552f0da91b248171d25789170110ae...1dd9c449504b07652af650382fc46b566fcd8546#7372632F747769737465642F6E616D65732F617574686F726974792E7079-86

Since recently policy says no overriding of these is allowed, you probably need to un-reformat that code or test it.

wsanchez:
Actually I mis-read the "directly relevant" example; you removed an adjacent line, so again, cleanup is causing coverage to complain.

glyph:
It looks like @hynek did in fact address all the review feedback, and the coverage issue is just a re-formatted `NotImplementedError` line.  Formally speaking this is something that should be "in a different PR" but given that absolute coverage increased, I'm not going to be a pain in the butt about it :).

Since this has been a while I'm going to update the branch and push it to a twisted branch for full CI, but assuming CI is reasonably happy after that, I'll merge.

glyph:
Thanks to https://www.traviscistatus.com/incidents/3pfp91tltsbr I really didn't want to wait the additional 2 hours for another trunk merge, so: merged.

glyph:
(Happily the status just passed though, so no exception necessary!)

PR #580, opened 2016-10-31
Port twisted.protocols.sip to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8669


src/twisted/protocols/sip.py
```diff
@@ -109,9 +110,13 @@
 
 
 def dashCapitalize(s):
-    ''' Capitalize a string, making sure to treat - as a word separator '''
+    '''
+    Capitalize a string, making sure to treat - as a word separator
```
moshez: This is hard to read. Maybe "to treat '-' as a word separator"?

src/twisted/protocols/sip.py
```diff
@@ -543,59 +583,75 @@ class MessagesParser(basic.LineReceiver):
...
             self.reset()
         else:
-            # we have enough data and message wasn't finished? something is wrong
+            # We have enough data and message wasn't finished? something is wrong
             raise RuntimeError("this should never happen")
 
+
     def dataReceived(self, data):
         try:
+            if isinstance(data, unicode):
```
moshez, line 632: Why is this needed? Can we fix the calling site instead?

src/twisted/protocols/sip.py
```diff
@@ -543,59 +583,75 @@ class MessagesParser(basic.LineReceiver):
...
         raise NotImplementedError
 
+
     def lineLengthExceeded(self, line):
         self.invalidMessage()
 
+
     def lineReceived(self, line):
+        if _PY3 and isinstance(line, bytes):
+            line = line.decode("utf-8")
```
moshez, line 653: This seems like there should be an abstraction "force_bytes" which will do it?

src/twisted/protocols/sip.py
```diff
@@ -754,76 +828,104 @@ def sendMessage(self, destURL, message):
             raise RuntimeError("only UDP currently supported")
         if self.debug:
             log.msg("Sending %r to %r" % (message.toString(), destURL))
-        self.transport.write(message.toString(), (destURL.host, destURL.port or self.PORT))
+        data = message.toString()
+        if isinstance(data, unicode):
+            data = data.encode("utf-8")
```
moshez, line 833: Is it OK to assume message.toString() will always return a unicode string that we can encode into utf-8? The doc string doesn't specify what assumptions it makes on message

moshez (approved):
Some small comments. Feel free to address those you think are good and push.

src/twisted/protocols/sip.py
rodrigc: Clients outside of Twisted can be written with things like dataReceived("someString")
so that means in Py2 it is bytes, but in Py3 it is unicode.  So we need this conversion.

PR #581, opened 2016-11-01
Port dirdbm to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8888


src/twisted/persisted/dirdbm.py
```diff
@@ -60,120 +65,134 @@ def __init__(self, name):
...
-    
+        Decode a filename to get the key.
+        """
+        return base64.decodestring(k.replace(b'_', b'\n').replace(b"-", b"/"))
+
+
     def _readFile(self, path):
-        """Read in the contents of a file.
-        
+        """
```
moshez, line 95: These are redundant with getContent/setContent. Since they are internal, they can just be killed, right?

src/twisted/persisted/dirdbm.py
```diff
@@ -60,120 +65,134 @@ def __init__(self, name):
...
         @return: The value associated with C{k}
         @raise KeyError: Raised when there is no such key
         """
-        assert type(k) == str, "DirDBM key must be a string"
-        path = os.path.join(self.dname, self._encode(k))
+        assert type(k) == bytes, "DirDBM key must be bytes"
+        path = self._dnamePath.child(self._encode(k))
         try:
             return self._readFile(path)
         except:
```
moshez: Should probably be changed to except (OSError, IOError):

src/twisted/persisted/dirdbm.py
```diff
@@ -60,120 +65,134 @@ def __init__(self, name):
...
+        @type k: bytes
         @param k: key to lookup
-        
+
         @return: The value associated with C{k}
         @raise KeyError: Raised when there is no such key
         """
-        assert type(k) == str, "DirDBM key must be a string"
-        path = os.path.join(self.dname, self._encode(k))
+        assert type(k) == bytes, "DirDBM key must be bytes"
```
moshez: Do we want to also change this from assert to raising an explicit TypeError? (Ditto for next one)

src/twisted/persisted/dirdbm.py
```diff
@@ -235,18 +258,20 @@ def get(self, key, default = None):
...
 
+
     def __contains__(self, key):
         """
         C{key in dirdbm}
 
         @type key: str
         @param key: The key to test
-                
+
```
moshez: Should this just call has_key? This is literally copy/paste (and we had to fix the same issues twice)

src/twisted/test/test_dirdbm.py
```diff
@@ -85,21 +91,22 @@ def testDbm(self):
...
+                len(d2.items()) == 0, ".clear() failed")
         shutil.rmtree(copyPath)
 
-        # delete items
+        # Delete items
         for k, v in self.items:
             del d[k]
             self.assertNotIn(k, d, "key is still in database, even though we deleted it")
-        assert len(d.keys()) == 0, "database has keys"
+        assert len(list(d.keys())) == 0, "database has keys"
```
moshez: Docstring says that .keys() returns a list. Why is this converted into list here? This seems like a backwards-incompatible change. Can we just do the list() in the original function?

moshez (approved):
Mostly stylistic comments but I think the .keys() thing should be addressed. Since it is not a regression (it would work the same on Py 2, and this has not worked at all on Py 3), I don't see it as a show-stopper.

moshez:
Oh, I forgot -- thanks for porting to Python 3! So happy to see this journey continuing.

src/twisted/persisted/dirdbm.py
rodrigc: OK

src/twisted/test/test_dirdbm.py
rodrigc: OK

src/twisted/persisted/dirdbm.py
rodrigc: OK

src/twisted/persisted/dirdbm.py
rodrigc: _readFile() and _writeFile() call _open().  By default _open == open, but apparently it can be overridden.

src/twisted/persisted/dirdbm.py
rodrigc: OK

rodrigc:
@moshez can I merge this?  I have addressed most of your comments.

src/twisted/persisted/dirdbm.py
```diff
@@ -30,27 +30,32 @@
...
     flat files. It can only use strings as keys or values.
     """
-    
+
     def __init__(self, name):
         """
         @type name: str
         @param name: Base path to use for the directory storage.
         """
         self.dname = os.path.abspath(name)
```
wsanchez, line 55: I'm kinda surprised that FilePath lacks an `abspath()` equivalent…

src/twisted/persisted/dirdbm.py
```diff
@@ -60,158 +65,180 @@ def __init__(self, name):
...
         """
-        assert type(k) == str, "DirDBM key must be a string"
-        path = os.path.join(self.dname, self._encode(k))
+        if not type(k) == bytes:
+            raise TypeError("DirDBM key must be bytes")
+        path = self._dnamePath.child(self._encode(k))
         try:
             return self._readFile(path)
-        except:
+        except (OSError, IOError):
```
wsanchez: `EnvironmentError` covers both of these; all three are the same class on Py3.

src/twisted/persisted/dirdbm.py
```diff
@@ -60,158 +65,180 @@ def __init__(self, name):
...
-        
+
         @raise KeyError: Raised when there is no such key
         """
-        assert type(k) == str, "DirDBM key must be a string"
+        if not type(k) == bytes:
+            raise TypeError("DirDBM key must be bytes")
         k = self._encode(k)
-        try:    os.remove(os.path.join(self.dname, k))
+        try:    self._dnamePath.child(k).remove()
```
wsanchez: Code standard wants a newline here.

src/twisted/persisted/dirdbm.py
wsanchez: This is a bit orthogonal to porting to Py3, anyway…

src/twisted/persisted/dirdbm.py
```diff
@@ -220,99 +247,102 @@ def setdefault(self, key, value):
...
+
     def __contains__(self, key):
         """
-        C{key in dirdbm}
-
-        @type key: str
-        @param key: The key to test
-                
-        @return: A true value if C{self.has_key(key)}, a false value otherwise.
+        L{DirDBM.has_key}
```
wsanchez: `@see: L{DirDBM.has_key}`

src/twisted/test/test_dirdbm.py
```diff
@@ -108,59 +129,73 @@ def testModificationTime(self):
         # write(). I consider this a kernel bug, but it is beyond the scope
         # of this test. Thus we keep the range of acceptability to 3 seconds time.
         # -warner
-        self.dbm["k"] = "v"
-        self.assertTrue(abs(time.time() - self.dbm.getModificationTime("k")) <= 3)
+        self.dbm[b"k"] = b"v"
+        self.assertTrue(abs(time.time() - self.dbm.getModificationTime(b"k")) <= 3)
+        self.assertRaises(KeyError, self.dbm.getModificationTime, b"nokey")
```
wsanchez, line 134: Perhaps better as a separate test.  (for the KeyError)

src/twisted/persisted/dirdbm.py
```diff
@@ -220,99 +247,102 @@ def setdefault(self, key, value):
...
+
         @rtype: C{DirDBM}
         @return: The dirdbm this dirdbm was copied to.
         """
-        path = os.path.abspath(path)
-        assert path != self.dname
-        
-        d = self.__class__(path)
+        path = FilePath(path)
+        assert path != self._dnamePath
```
wsanchez, line 300: Not a blocker because this is existing behavior, but this isn't documented as not allowed, and it seems like a no-op may be more appropriate than raising `AssertionError`.

src/twisted/persisted/dirdbm.py
```diff
@@ -220,99 +247,102 @@ def setdefault(self, key, value):
...
+
         @return: Last modification date (seconds since epoch) of entry C{key}
         @raise KeyError: Raised when there is no such key
         """
-        assert type(key) == str, "DirDBM key must be a string"
-        path = os.path.join(self.dname, self._encode(key))
-        if os.path.isfile(path):
-            return os.path.getmtime(path)
+        if not type(key) == bytes:
+            raise TypeError("DirDBM key must be bytes")
```
wsanchez, line 331: Does `AssertionError` ➜ `TypeError` fit with the compatibility policy?
IIRC not really, though I strongly agree that this is an improvement.

wsanchez:
reviewed.  The only potentially blocking concern here is that I think the `AssertionError` ➜ `TypeError` isn't allowed by the compatibility policy.

PR #583, opened 2016-11-02
Ignore unexpected 1XX status codes.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8885

Some notes on this review:

1. I suspect we should log out a message when we drop a 1XX status code like this, but I'm reluctant to contribute to the continued use of the old logging system in twisted.web. Thoughts?
2. I wondered whether we should continue to surface 101 responses as per their old logic. We might accidentally regress something if users are expecting to be delivered those responses. Any thoughts there?

### DIFF
--- a/src/twisted/web/_newclient.py
+++ b/src/twisted/web/_newclient.py
@@ -461,6 +461,9 @@ class HTTPClientParser(HTTPParser):
             # going to do. We reset the parser here, but we leave
             # _everReceivedData in its True state because we have, in fact,
             # received data.
+            log.msg(
+                "Ignoring unexpected {} response".format(self.response.code)
+            )
             self.connectionMade()
             del self.response
             return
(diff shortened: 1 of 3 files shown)


wsanchez:
Reviewed.

Not a blocker, but I agree that logging something is appropriate since we're not generating a response here and it won't even show up in the access log… which may be somewhat unexpected for someone trying to understand why some request is being ignored.

I don't think that logging with the system in use by that module should be discouraged.  When `twisted.web` gets ported to the new logging system, your added log line will be part of that, and there's no logging compatibility contract in Twisted.

Lukasa:
Hello gigantic green tick, so nice to see you!

jchampio:
> I wondered whether we should continue to surface 101 responses as per their old logic. We might accidentally regress something if users are expecting to be delivered those responses. Any thoughts there?

This did in fact break mod_websocket's test bed; I was using Agent to verify that a 101 Upgrade was correctly handled by the server. Any recommendations for replacement client logic?

jchampio:
(forgot to mention @Lukasa)

Lukasa:
Yeah, good question. 

*Probably* we should consider adding a proper API for this. Maybe an interface for an object that can have 1XX responses delivered to it? Alternatively, proper support for 101 might be the best approach: 101 responses probably *should* be surfaced to the user.

PR #586, opened 2016-11-05
Actually return host and peer information from H2Stream.

Lukasa:
See https://twistedmatrix.com/trac/ticket/8893


hawkowl:
yaaaaay

PR #588, opened 2016-11-10
[#8899] Automat for client service

tomprince:
https://twistedmatrix.com/trac/ticket/8899

This changes the behavior in one subtle ways, and adds a dependency on [automat](https://pypi.org/project/Automat).

- `ClientService` will no longer start reconnecting until after it has disconnected, when it is restarted.


tomprince:
I'm not entirely happy with all the docstrings for `_ClientMachine`. I'd be happy for suggestions to make them more coherent.

glyph:
Hooray, twistedchecker is complaining again :).  Please fix the travis failures.

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
     """
 
+    _machine = MethodicalMachine()
+
+    def __init__(self, endpoint, factory, retryPolicy, clock, log):
+        """
+        @see L{ClientService.__init__}
+
+        @param log: The logger for the L{ClientService} instance this
+            state machine is assoicated to.
```
glyph: "associated"? :)

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
+    @see: L{ClientService}
     """
 
+    _machine = MethodicalMachine()
+
+    def __init__(self, endpoint, factory, retryPolicy, clock, log):
+        """
+        @see L{ClientService.__init__}
+
+        @param log: The logger for the L{ClientService} instance this
```
glyph: These lines are wrapped way too aggressively.  70 characters?

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
+            state machine is assoicated to.
+        @type log: L{Logger}
+        """
+        self._endpoint = endpoint
```
glyph, line 559: This is crying out for `attrs` :(

src/twisted/application/test/test_internet.py
```diff
@@ -829,3 +847,14 @@ def test_whenConnectedStopService(self):
         self.failureResultOf(a, CancelledError)
         self.failureResultOf(b, CancelledError)
 
+
+    def test_retryCancelled(self):
+        """
+        When L{ClientService.stopService} is called while waiting between connection attempts,
```
glyph: Not enough space on the earlier docstrings, way too much on this one (95 characters)?

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
+    @_machine.input()
+    def _connectionMade(self, protocol):
+        """
+        A connection has been made.
+
+        @param protocol: The protocol of the connection.
+        @type protocol: L{IProtocol}
+        """
+
+    @_machine.output()
```
glyph, line 700: I've found it helpful to group all the "outputs" in a section by themselves, since they are actual methods, as opposed to the input/state declarations, which are empty.  It's easier to read if they're separated a little bit, I think.

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
+    @_machine.state()
+    def _connecting(self):
+        """
+        The service has started connecting.
+        """
+
+    @_machine.state()
+    def _waiting(self):
+        """
+        The service is waiting for the reconnection period
```
glyph, line 588: Worth noting something about the state of the pending timed call that might need cleaning up?

src/twisted/application/internet.py
```diff
@@ -523,13 +524,375 @@ def policy(attempt):
...
+                        outputs=[_awaitingConnection],
+                        collector=firstResult)
+    _restarting.upon(whenConnected, enter=_restarting,
+                     outputs=[_awaitingConnection],
+                     collector=firstResult)
+    _stopped.upon(whenConnected, enter=_stopped,
+                  outputs=[_noConnection],
+                  collector=firstResult)
+
+    del firstResult
```
glyph: Maybe just define this as `_firstResult` outside the scope of the class so you don't need to `del` it?

src/twisted/application/internet.py
```diff
@@ -567,19 +929,10 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
...
-        self._timeoutForAttempt = retryPolicy
-        self._clock = clock
-        self._stopRetry = _noop
-        self._lostDeferred = succeed(None)
-        self._connectionInProgress = succeed(None)
-        self._loseConnection = _noop
-
-        self._currentConnection = None
-        self._awaitingConnected = []
+        self._machine = _ClientMachine(
```
glyph, line 946: hooray private variables and implementation details we can change

glyph (commented):
I actually thought the `_ClientService` docstrings weren't bad.  They _should_ be terse; the state machine should be largely self-describing.  All my feedback is basically aesthetic; I'd prefer it if you wrapped the docstrings at the standard (79 chars) width so that someone re-wrapping these won't cause diff noise.

glyph:
Since we can do this now:
sm

src/twisted/application/internet.py
tomprince: That was one of the orderings which I considered, but the ordering was deliberate. All the states are first, and then then there is each input, followed by all the new outputs triggered by that particular input.

I don't have strong feelings either way.

src/twisted/application/internet.py
tomprince: Perhaps, but I'm not sure why you singled out this particular machine state? All of the states have some associated data that needs handling at transitions?

src/twisted/application/internet.py
glyph: Neither do I; I'm happy to leave it as it is.

src/twisted/application/internet.py
```diff
@@ -522,14 +523,388 @@ def policy(attempt):
...
+
+    # State Transitions
+
+    _init.upon(start, enter=_connecting,
+               outputs=[_connect])
+    _init.upon(stop, enter=_init,
+               outputs=[])
+
+    _connecting.upon(start, enter=_connecting, outputs=[])
+    _connecting.upon(stop, enter=_disconnecting,
```
glyph, line 839: I feel like a lot of these `upon` invocations are bloating the line-count by wrapping lines too early.  It's a matter of taste though, so if you prefer this formatting you can leave it.

glyph (approved):
I'm still happy with it :).  Assuming the tests pass with the most recent trunk merge, land at your discretion.

hawkowl:
Pushed to twisted as 8899-automat-clientservice

hawkowl:
Twisted bb builds here: https://buildbot.twistedmatrix.com/boxes-supported?num_builds=2&&branch=8899-automat-clientservice

PR #589, opened 2016-11-11
Fix conch and cftp under Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8791


src/twisted/conch/scripts/cftp.py
```diff
@@ -181,40 +190,42 @@ def _dispatchCommand(self, line):
         if f is not None:
             return defer.maybeDeferred(f, rest)
         else:
-            self._ebCommand(failure.Failure(NotImplementedError(
-                "No command called `%s'" % command)))
+            errMsg = "No command called `%s'" % (command)
+            self._ebCommand(failure.Failure(NotImplementedError(errMsg)))
```
acabhishek942, line 194: why is this change necessary?

src/twisted/conch/test/test_cftp.py
```diff
@@ -994,24 +1041,34 @@ def testCdPwd(self):
         remote directory.
         """
         # XXX - not actually a unit test, see docstring.
-        homeDir = os.path.join(os.getcwd(), self.testDir)
+        homeDir = self.testDir
         d = self.runScript('pwd', 'lpwd', 'cd testDirectory', 'cd ..', 'pwd')
-        d.addCallback(lambda xs: xs[:3] + xs[4:])
+
+        def cmdOutput(output):
```
acabhishek942, line 1047: would be better to add docstring for this method

src/twisted/conch/test/test_cftp.py
```diff
@@ -1304,6 +1371,7 @@ def _cleanup(res):
 
         return d
 
+
     def testBatchFile(self):
         """Test whether batch file function of cftp ('cftp -b batchfile').
```
acabhishek942: this docstring also needs to be cleaned up

src/twisted/conch/test/test_cftp.py
```diff
@@ -1312,31 +1380,35 @@ def testBatchFile(self):
...
+            self.assertEqual(res[3:-2], [b'testDirectory', b'testRemoveFile',
+                                             b'testRenameFile', b'testfile1'])
 
         d = self._getBatchOutput(cmds)
         d.addCallback(_cbCheckResult)
         return d
 
+
     def testError(self):
         """Test that an error in the batch file stops running the batch.
```
acabhishek942: same here

src/twisted/conch/test/test_cftp.py
```diff
@@ -1312,31 +1380,35 @@ def testBatchFile(self):
...
-            self.assertNotIn(self.testDir, res)
+            self.assertNotIn(self.testDir.asBytesMode().path, res)
 
         d = self._getBatchOutput(cmds)
         d.addCallback(_cbCheckResult)
         return d
 
+
     def testIgnoredError(self):
         """Test that a minus sign '-' at the front of a line ignores
```
acabhishek942: same here

src/twisted/conch/test/test_conch.py
```diff
@@ -490,9 +498,9 @@ def test_clientRekey(self):
         process = ConchTestOpenSSHProcess()
         d = self.execute("", process, '-o RekeyLimit=2K')
         def finished(result):
-            self.assertEqual(
-                result,
-                '\n'.join(['line #%02d' % (i,) for i in range(60)]) + '\n')
+            expectedResult = '\n'.join(['line #%02d' % (i,) for i in range(60)]) + '\n'
+            expectedResult = expectedResult.encode("utf-8")
+            self.assertEqual(result, expectedResult)
```
acabhishek942, line 503: I feel there is a redundant line here it could be changed to `self.assertEqual(result, expectedResult.encode('utf-8'))` and remove this line `expectedResult = expectedResult.encode("utf-8")`

acabhishek942 (changes requested):
I have left some inline comments, please address them as you wish, my one other concern is the coverage issue which is bad mainly because of the unicode check performed at several places. I feel this could be improved by writing a helper function in twisted.python.compat or in some other module(I am not exactly sure which module should have this helper method)

src/twisted/conch/scripts/cftp.py
rodrigc: I separated the variables out to facilitate debugging as I was modifying the code.

PR #590, opened 2016-11-11
Add extractCurrentResult() and observeCurrentResult() to synchronously interact with a Deferred's result.

dstufft:
Ticket: https://twistedmatrix.com/trac/ticket/8900

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -171,15 +171,14 @@ def maybeDeferred(f, *args, **kw):
 
 
 
-def extractCurrentResult(deferred, raises=True, passthrough=False):
+def extractCurrentResult(deferred, raises=True):
     """
     Extract the current result from a L{Deferred} that has already fired all of
     its callbacks.
 
     Take a deferred that has already fired all of its callbacks and extract and
     return the result of the L{Deferred}. This process of extracting the result
-    will, by default, consume the result - error or not - unless the
-    L{passthrough} parameter is set to L{True}. If the L{Deferred} has an
+    will consume the result, error or otherwise. If the L{Deferred} has an
     error then the process of extracting the result will, by default, re-raise
     the original exception unless the L{raises} parameter is set to L{False}.
 
@@ -201,10 +200,6 @@ def extractCurrentResult(deferred, raises=True, passthrough=False):
     @param raises: Whether or not to raise the original exception (L{True}) or
         to return the failure object (L{False}).
 
-    @type passthrough: bool
-    @param passthrough: Whether or not to pass through the original value
-        (L{True}) or to consume any results, error or otherwise (L{False}).
-
     @raise NoCurrentResult: If C{deferred} has pending work to be done
         and has no current result to extract.
 
@@ -212,12 +207,6 @@ def extractCurrentResult(deferred, raises=True, passthrough=False):
     """
     def done(result):
         done.result = result
-
-        # We return the result here so that if anything else continued to use
-        # this Deferred after it's been unboxed then the act of unboxing
-        # has not broken the flow of values through the callback chain.
-        if passthrough:
-            return result
 
     # We need a place to store the result, and a way to determine the
     # difference between no result and any other possible object. So we'll
@@ -242,6 +231,81 @@ def extractCurrentResult(deferred, raises=True, passthrough=False):
     if done.result is nothing:
         raise NoCurrentResult(
             "Cannot extract the current result of a Deferred that has not "
+            "currently called all of its callbacks."
+        )
+    else:
+        if raises and isinstance(done.result, failure.Failure):
+            done.result.raiseException()
+        else:
+            return done.result
+
+
+
+def observeCurrentResult(deferred, raises=True):
+    """
+    Observes the current result from a L{Deferred} that has already fired all
+    of its callbacks.
+
+    Take a deferred that has already fired all of its callbacks and observe and
+    return the result of the L{Deferred}. This process of extracting the result
+    will *NOT* consume the result. If the L{Deferred} has an error then the
+    process of observing the result will, by default, re-raise the
+    original exception unless the L{raises} parameter is set to L{False}.
+
+    The primary purpose of this function is to allow code to be written that is
+    internally asynchronous but which can be consumed without an event loop in
+    a synchronous fashion by using L{observeCurrentResult} at the boundaries.
+    This enables libraries to be written that can function both with Twisted
+    and with synchronous code instead of requiring an asynchronous and a
+    synchronous implementation for all functionality.
+
+    @attention: This function CANNOT take a Deferred with pending asynchronous
+        work to be done and make it synchronous. It can ONLY extract results
+        from a L{Deferred} which has already fired its callbacks.
+
+    @type deferred: Deferred
+    @param deferred: The Deferred from which results will be extracted..
+
+    @type raises: bool
+    @param raises: Whether or not to raise the original exception (L{True}) or
+        to return the failure object (L{False}).
+
+    @raise NoCurrentResult: If C{deferred} has pending work to be done
+        and has no current result to extract.
+
+    @return: The result of the L{Deferred}.
+    """
+    def done(result):
+        done.result = result
+
+        # We return the result here so that if anything else continued to use
+        # this Deferred after it's been observed then the act of observing has
+        # not broken the flow of values through the callback chain.
+        return result
+
+    # We need a place to store the result, and a way to determine the
+    # difference between no result and any other possible object. So we'll
+    # create a sigil object and stash it locally and default our result to
+    # that.
+    # Note: This *has* to happen before adding the done callback to the
+    #       Deferred or else the callback can fire prior to setting our
+    #       default, thus causing the default to override the value from the
+    #       callback.
+    nothing = done.result = object()
+
+    # Now that the callback is fully setup and prepared, register it with the
+    # deferred. This *should* run immediately since only Deferreds that do not
+    # have any pending work should be passed into this function.
+    deferred.addBoth(done)
+
+    # If we've gotten to this point and we haven't had our collect function
+    # called, then we must be waiting on something in the Deferred chain. If so
+    # we'll error out here because there's nothing we can do to advance the
+    # state of the Deferred. If we've got a result then we'll go ahead and
+    # return it or raise the given exception.
+    if done.result is nothing:
+        raise NoCurrentResult(
+            "Cannot observe the current result of a Deferred that has not "
             "currently called all of its callbacks."
         )
     else:
@@ -1917,8 +1981,8 @@ class DeferredFilesystemLock(lockfile.FilesystemLock):
 __all__ = ["Deferred", "DeferredList", "succeed", "fail", "FAILURE", "SUCCESS",
            "AlreadyCalledError", "TimeoutError", "NoCurrentResult",
            "gatherResults", "maybeDeferred", "ensureDeferred",
-           "extractCurrentResult", "waitForDeferred", "deferredGenerator",
-           "inlineCallbacks", "returnValue",
+           "extractCurrentResult", "observeCurrentResult", "waitForDeferred",
+           "deferredGenerator", "inlineCallbacks", "returnValue",
            "DeferredLock", "DeferredSemaphore", "DeferredQueue",
            "DeferredFilesystemLock", "AlreadyTryingToLockError",
           ]
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -53,6 +53,17 @@ class TimeoutError(Exception):
 
 
 
+class NoCurrentResult(BaseException):
+    """
+    This error is raised when attempting to call L{extractCurrentResult} on a
+    L{Deferred} that has not yet called all of its callbacks.
+
+    This error inherits from L{BaseException} rather than L{Exception} to
+    reduce the likelihood of inadvertingly catching this exception as part of
+    an exception handler for other errors.
+    """
+
+
 def logError(err):
     """
     Log and return failure.
@@ -157,6 +168,87 @@ def maybeDeferred(f, *args, **kw):
         return fail(result)
     else:
         return succeed(result)
+
+
+
+def extractCurrentResult(deferred, raises=True, passthrough=False):
+    """
+    Extract the current result from a L{Deferred} that has already fired all of
+    its callbacks.
+
+    Take a deferred that has already fired all of its callbacks and extract and
+    return the result of the L{Deferred}. This process of extracting the result
+    will, by default, consume the result - error or not - unless the
+    L{passthrough} parameter is set to L{True}. If the L{Deferred} has an
+    error then the process of extracting the result will, by default, re-raise
+    the original exception unless the L{raises} parameter is set to L{False}.
+
+    The primary purpose of this function is to allow code to be written that is
+    internally asynchronous but which can be consumed without an event loop in
+    a synchronous fashion by using L{extractCurrentResult} at the boundaries.
+    This enables libraries to be written that can function both with Twisted
+    and with synchronous code instead of requiring an asynchronous and a
+    synchronous implementation for all functionality.
+
+    @attention: This function CANNOT take a Deferred with pending asynchronous
+        work to be done and make it synchronous. It can ONLY extract results
+        from a L{Deferred} which has already fired its callbacks.
+
+    @type deferred: Deferred
+    @param deferred: The Deferred from which results will be extracted..
+
+    @type raises: bool
+    @param raises: Whether or not to raise the original exception (L{True}) or
+        to return the failure object (L{False}).
+
+    @type passthrough: bool
+    @param passthrough: Whether or not to pass through the original value
+        (L{True}) or to consume any results, error or otherwise (L{False}).
+
+    @raise NoCurrentResult: If C{deferred} has pending work to be done
+        and has no current result to extract.
+
+    @return: The result of the L{Deferred}.
+    """
+    def done(result):
+        done.result = result
+
+        # We return the result here so that if anything else continued to use
+        # this Deferred after it's been unboxed then the act of unboxing
+        # has not broken the flow of values through the callback chain.
+        if passthrough:
+            return result
+
+    # We need a place to store the result, and a way to determine the
+    # difference between no result and any other possible object. So we'll
+    # create a sigil object and stash it locally and default our result to
+    # that.
+    # Note: This *has* to happen before adding the done callback to the
+    #       Deferred or else the callback can fire prior to setting our
+    #       default, thus causing the default to override the value from the
+    #       callback.
+    nothing = done.result = object()
+
+    # Now that the callback is fully setup and prepared, register it with the
+    # deferred. This *should* run immediately since only Deferreds that do not
+    # have any pending work should be passed into this function.
+    deferred.addBoth(done)
+
+    # If we've gotten to this point and we haven't had our collect function
+    # called, then we must be waiting on something in the Deferred chain. If so
+    # we'll error out here because there's nothing we can do to advance the
+    # state of the Deferred. If we've got a result then we'll go ahead and
+    # return it or raise the given exception.
+    if done.result is nothing:
+        raise NoCurrentResult(
+            "Cannot extract the current result of a Deferred that has not "
+            "currently called all of its callbacks."
+        )
+    else:
+        if raises and isinstance(done.result, failure.Failure):
+            done.result.raiseException()
+        else:
+            return done.result
 
 
 
@@ -1823,10 +1915,10 @@ class DeferredFilesystemLock(lockfile.FilesystemLock):
 
 
 __all__ = ["Deferred", "DeferredList", "succeed", "fail", "FAILURE", "SUCCESS",
-           "AlreadyCalledError", "TimeoutError", "gatherResults",
-           "maybeDeferred", "ensureDeferred",
-           "waitForDeferred", "deferredGenerator", "inlineCallbacks",
-           "returnValue",
+           "AlreadyCalledError", "TimeoutError", "NoCurrentResult",
+           "gatherResults", "maybeDeferred", "ensureDeferred",
+           "extractCurrentResult", "waitForDeferred", "deferredGenerator",
+           "inlineCallbacks", "returnValue",
            "DeferredLock", "DeferredSemaphore", "DeferredQueue",
            "DeferredFilesystemLock", "AlreadyTryingToLockError",
           ]
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -671,7 +671,7 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
 
     def test_extractCurrentResultReturnsFailureWithNoRaises(self):
         """
-        L{deferextractCurrentResult} returns the failure instance if
+        L{defer.extractCurrentResult} returns the failure instance if
         raises=False.
         """
         d = defer.Deferred()
@@ -708,28 +708,6 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         d = defer.Deferred()
         d.callback(None)
         self.assertEqual(inner(d), "success")
-
-
-    def test_extractCurrentResultPreservesCallbackChain(self):
-        """
-        L{defer.extractCurrentResult} chains the value that will be passed into
-        the next function of the callback chain if passthrough=True.
-        """
-        result1 = object()
-        result2 = object()
-
-        d = defer.Deferred()
-        d.callback(result1)
-
-        self.assertIs(defer.extractCurrentResult(d, passthrough=True), result1)
-
-        def cb(r):
-            self.assertIs(r, result1)
-            return result2
-
-        d.addCallback(cb)
-
-        self.assertIs(defer.extractCurrentResult(d), result2)
 
 
     def test_extractCurrentResultRaisesErrorIfPending(self):
@@ -786,6 +764,136 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
 
         self.assertIs(defer.extractCurrentResult(d, raises=False), result)
         self.assertIs(defer.extractCurrentResult(d), None)
+
+
+    def test_observeCurrentResult(self):
+        """
+        L{defer.observeCurrentResult} extracts the results from a Deferred that
+        has already had its callback function called.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.observeCurrentResult(d), result)
+        # Observing the result does not change the result, so calling it twice
+        # will return the same value.
+        self.assertIs(defer.observeCurrentResult(d), result)
+
+
+    def test_observeCurrentResultRaisesFailureException(self):
+        """
+        L{defer.observeCurrentResult} raises the original exception if the
+        result of the deferred is a failure.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        self.assertRaises(ValueError, defer.observeCurrentResult, d)
+        # Observing the result does not change the result, so calling it twice
+        # will raise the same value.
+        self.assertRaises(ValueError, defer.observeCurrentResult, d)
+
+        # Squash the error to avoid logging an unhandled error.
+        d.addErrback(lambda r: None)
+
+
+    def test_observeCurrentResultReturnsFailureWithNoRaises(self):
+        """
+        L{defer.observeCurrentResult} returns the failure instance if
+        raises=False.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        result = defer.observeCurrentResult(d, raises=False)
+
+        self.assertIsInstance(result, failure.Failure)
+        self.assertIsInstance(result.value, ValueError)
+
+        # Squash the error to avoid logging an unhandled error.
+        d.addErrback(lambda r: None)
+
+
+    def test_observeCurrentResultWithoutResultsNotCaughtByException(self):
+        """
+        L{defer.observeCurrentResult} raises a L{defer.NoCurrentResult}
+        exception which is not caught when catching L{Exception}.
+        """
+        def inner(d):
+            try:
+                defer.observeCurrentResult(d)
+            except Exception:
+                return "error"
+            return "success"
+
+        d = defer.Deferred()
+        self.assertRaises(defer.NoCurrentResult, inner, d)
+
+        # In order to get 100% patch coverage we need to exercise the "error"
+        # and "success" paths. This is not actually testing the functionality
+        # of observeCurrentResult but is testing the functionality of the test
+        # function `inner`.
+        d = defer.Deferred()
+        d.errback(Exception())
+        self.assertEqual(inner(d), "error")
+        # Squash the error to avoid logging an unhandled error.
+        d.addErrback(lambda r: None)
+        d = defer.Deferred()
+        d.callback(None)
+        self.assertEqual(inner(d), "success")
+
+
+    def test_observeCurrentResultPreservesCallbackChain(self):
+        """
+        L{defer.observeCurrentResult} chains the value that will be passed into
+        the next function of the callback chain.
+        """
+        result1 = object()
+        result2 = object()
+
+        d = defer.Deferred()
+        d.callback(result1)
+
+        self.assertIs(defer.observeCurrentResult(d), result1)
+
+        def cb(r):
+            self.assertIs(r, result1)
+            return result2
+
+        d.addCallback(cb)
+
+        self.assertIs(defer.observeCurrentResult(d), result2)
+
+
+    def test_observeCurrentResultRaisesErrorIfPending(self):
+        """
+        L{defer.observeCurrentResult} raises an error if there currently any
+        pending work to be done on the L{Deferred}.
+        """
+        d = defer.Deferred()
+
+        self.assertRaises(defer.NoCurrentResult, defer.observeCurrentResult, d)
+
+
+    def test_observeCurrentResultRaisesErrorIfPendingChainedDeferred(self):
+        """
+        L{defer.observeCurrentResult} raises an error if there is currently any
+        pending work to be done on the L{Deferred} if it is via a chained
+        L{Deferred}.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.observeCurrentResult(d), result)
+
+        d2 = defer.Deferred()
+        d.addCallback(lambda _: d2)
+
+        self.assertRaises(defer.NoCurrentResult, defer.observeCurrentResult, d)
 
 
     def test_maybeDeferredSync(self):
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -645,6 +645,149 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         self.assertEqual(callbackResult[1], "Callback Result")
 
 
+    def test_extractCurrentResult(self):
+        """
+        L{defer.extractCurrentResult} extracts the results from a Deferred that
+        has already had its callback function called.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.extractCurrentResult(d), result)
+
+
+    def test_extractCurrentResultRaisesFailureException(self):
+        """
+        L{defer.extractCurrentResult} raises the original exception if the
+        result of the deferred is a failure.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        self.assertRaises(ValueError, defer.extractCurrentResult, d)
+
+
+    def test_extractCurrentResultReturnsFailureWithNoRaises(self):
+        """
+        L{deferextractCurrentResult} returns the failure instance if
+        raises=False.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        result = defer.extractCurrentResult(d, raises=False)
+
+        self.assertIsInstance(result, failure.Failure)
+        self.assertIsInstance(result.value, ValueError)
+
+
+    def test_extractCurrentResultWithoutResultsNotCaughtByException(self):
+        """
+        L{defer.extractCurrentResult} raises a L{defer.NoCurrentResult}
+        exception which is not caught when catching L{Exception}.
+        """
+        def inner(d):
+            try:
+                defer.extractCurrentResult(d)
+            except Exception:
+                return "error"
+            return "success"
+
+        d = defer.Deferred()
+        self.assertRaises(defer.NoCurrentResult, inner, d)
+
+        # In order to get 100% patch coverage we need to excerise the "error"
+        # and "success" paths. This is not actually testing the functionality
+        # of extractCurrentResult but is testing the functionality of the test
+        # function `inner`.
+        d = defer.Deferred()
+        d.errback(Exception())
+        self.assertEqual(inner(d), "error")
+        d = defer.Deferred()
+        d.callback(None)
+        self.assertEqual(inner(d), "success")
+
+
+    def test_extractCurrentResultPreservesCallbackChain(self):
+        """
+        L{defer.extractCurrentResult} chains the value that will be passed into
+        the next function of the callback chain if passthrough=True.
+        """
+        result1 = object()
+        result2 = object()
+
+        d = defer.Deferred()
+        d.callback(result1)
+
+        self.assertIs(defer.extractCurrentResult(d, passthrough=True), result1)
+
+        def cb(r):
+            self.assertIs(r, result1)
+            return result2
+
+        d.addCallback(cb)
+
+        self.assertIs(defer.extractCurrentResult(d), result2)
+
+
+    def test_extractCurrentResultRaisesErrorIfPending(self):
+        """
+        L{defer.extractCurrentResult} raises an error if there currently any
+        pending work to be done on the L{Deferred}.
+        """
+        d = defer.Deferred()
+
+        self.assertRaises(defer.NoCurrentResult, defer.extractCurrentResult, d)
+
+
+    def test_extractCurrentResultRaisesErrorIfPendingChainedDeferred(self):
+        """
+        L{defer.extractCurrentResult} raises an error if there is currently any
+        pending work to be done on the L{Deferred} if it is via a chained
+        L{Deferred}.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.extractCurrentResult(d), result)
+
+        d2 = defer.Deferred()
+        d.addCallback(lambda _: d2)
+
+        self.assertRaises(defer.NoCurrentResult, defer.extractCurrentResult, d)
+
+
+    def test_extractCurrentResultDoesntPassthroughResults(self):
+        """
+        L{defer.extractCurrentResult} does not pass through results.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.extractCurrentResult(d), result)
+        self.assertIs(defer.extractCurrentResult(d), None)
+
+
+    def test_extractCurrentResultDoesntPassthroughErrorResults(self):
+        """
+        L{defer.extractCurrentResult} does not pass through errors.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        # Reach inside the deferred and pull out the result.
+        result = d.result
+
+        self.assertIs(defer.extractCurrentResult(d, raises=False), result)
+        self.assertIs(defer.extractCurrentResult(d), None)
+
+
     def test_maybeDeferredSync(self):
         """
         L{defer.maybeDeferred} should retrieve the result of a synchronous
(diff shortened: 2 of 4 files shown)


src/twisted/internet/defer.py
```diff
@@ -52,6 +52,12 @@ class TimeoutError(Exception):
     """
 
 
+class InvalidStateError(Exception):
```
dstufft: This should maybe derive from `BaseException` instead so that a:

``` python
try:
    undefer(d)
except Exception:
    pass

```

Does not catch an invalid `undefer()` call the same as an exception that came from the callbacks internal to it.

rodrigc:
(1)  Patch needs a topfile news fragment, as specified here: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

(2)  Patch needs to modify the tests to increase coverage.  If you install the [Codecov browser plugin](https://github.com/codecov/browser-extension)
       and look at https://github.com/twisted/twisted/pull/590/files , this patch only has 16% code coverage

dstufft:
@rodrigc Yea I know, I was hoping to get some feedback on the idea and implementation before I did that :)

src/twisted/internet/defer.py
dstufft: Another thought about why this should come from `BaseException`:

There's not really much recovering from this. At best maybe you can run it through the event loop or synchronously drive it and then pass it through again... but if you can do that why didn't you do it in the first place?

src/twisted/internet/defer.py
```diff
@@ -159,6 +165,65 @@ def maybeDeferred(f, *args, **kw):
         return succeed(result)
 
 
+def unbox(deferred, raises=True, passthrough=True):
```
dstufft: I'm not sure about the `passthrough` argument here.

Essentially, the problem boils down to if you don't have `passthrough=False` then when the `Deferred` gets garbage collected if it was in a failure state then it prints a spurious traceback to stdout. This sees to be a problem that other utilities like `gatherResults` has had and they have a `consumeErrors` parameter which defaults to `False` but is documented that you should almost always use `consumeErrors=True`.

So I guess the question is, what behavior with results makes sense?
- Always pass through results, success or failure and if someone wants something different then they need to add their own callback that throws away the result.
- Always pass through successful results but consume any failure results under the assumption that if you've unboxed the `Deferred` then you've handled the failure (and if not, you can add your own callback that re-inserts it).
- Never pass through any result and "break" the callback chain if someone continues to re-use the `Deferred` instance after you've unboxed it. If you want that to continue working then you'll need to add your own callback that re-inserts the result.
- Add a `passthrough` parameter that controls in both the success and the failure case whether or not it should be passed through or not.
  - If this is the answer, should we pass through results (success and failure) by default or not?
- Add a `consumeErrors` parameter that controls whether or not `unbox()` will consume errors.
  - Should this default to `False` as the other utility functions do (though I think they do for backwards compatibility) or should this default to `True` to more likely do the right thing?
  - Should we always return success results then or should we always consume them? Does it make sense to add a `consumeSuccesses` parameter for success results? If so what should it default to?

src/twisted/internet/defer.py
```diff
@@ -52,6 +52,12 @@ class TimeoutError(Exception):
     """
 
 
+class InvalidStateError(BaseException):
```
Julian: The name here sounds a bit too general / risky to have it start being used by other things if all it currently means is e.g. `UnboxingWithoutResult`

src/twisted/internet/defer.py
```diff
@@ -159,6 +165,65 @@ def maybeDeferred(f, *args, **kw):
         return succeed(result)
 
 
+def unbox(deferred, raises=True, passthrough=True):
+    """
+    Unbox a L{Deferred} that has already been fully processed.
+
+    Take an already processed (i.e. no longer waiting on any callbacks to be
+    fired) and unbox it. If the result is a L{Failure}, then it will re-raise
```
Julian: Missing a word here (deferred), and I think it'd be a bit nicer to flip the order of the english sentence so that you talk about the happy normal case first, and then "it reraises" is your otherwise at the end.

src/twisted/internet/defer.py
```diff
@@ -159,6 +165,65 @@ def maybeDeferred(f, *args, **kw):
         return succeed(result)
 
 
+def unbox(deferred, raises=True, passthrough=True):
+    """
+    Unbox a L{Deferred} that has already been fully processed.
```
Julian: Do other places use the lingo "fully processed", I'm not experienced enough to know, but I think common vernacular if that's what you mean is "that has already fired its result" or the like.

src/twisted/internet/defer.py
```diff
@@ -159,6 +165,65 @@ def maybeDeferred(f, *args, **kw):
...
+
+    Take an already processed (i.e. no longer waiting on any callbacks to be
+    fired) and unbox it. If the result is a L{Failure}, then it will re-raise
+    the original exception, otherwise it will return the result of the
+    L{Deferred}.
+
+    B{This function CANNOT take a Deferred with pending asynchronous work to be
+    done and make it synchronous. It can ONLY unbox a Deferred which already
+    has successfully completed.}
+
```
Julian, line 265: I think even if you don't borrow the existing terminology that at least linking to `successResultOf` might be a nice crosslink.

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,121 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
         self.assertEqual(callbackResult[1], "Callback Result")
 
 
+    def test_unbox(self):
+        """
+        L{defer.unbox} should unbox a Deferred that has already had it's
```
Julian: its

src/twisted/internet/defer.py
dstufft: After thinking about this more, I _think_ the right answer is one of:
- Always pass through successful responses and always consume errors.
- Always pass through successful responses and add a `consumeErrors` parameter.
  - Defaulting to `True` to do the right thing by default.
  - Defaulting to `False` to match `gatherResults` and friends.
- Always consume everything.

The reasoning for this is that there's very little downside to passing through a successful result. At worst it will just hold a reference to it until the `Deferred` itself gets garbage collected. So it just always passes that through.

However, if you don't consume errors it prints spurious error results to stdout so there is an incentive to consume those and not pass them through. Adding a `consumeErrors` parameter aligns with other utilities and allows control on this.

The downside is errors and success ends up being treated differently although that could possibly be solved by consuming everything and treating `unbox()` as something you should generally only call on a `Deferred` that is no longer going to be kept around or used.

dstufft: My intuition here is that:

``` python
def unbox(deferred, raises=True, consumeErrors=True):
    ...
```

Is the "right" way forward, but I'm not as versed in what is normal in Twisted land.

Julian: So I think the behavior I expected before reading this (even though I'm no deferred expert) is to always raise, and always passthrough.

Someone who wanted `raise=False` can try/except and someone who wanted specifically the failure and not the exception sounds like they'll be unhappy but I think that's the "fault" of the interaction between failures and exceptions, not of your code here -- i.e., if that was a common use case, I'd expect to see an inverted failure object which is an exception that carries around the original failure.

As for passthrough which it sounds like is more important to address -- I'd have expected always passthrough for a bunch of reasons, some dogmatic and some practical

0) I'm discarding the asymmetric options immediately because I'd find it surprising to treat failures differently than successes personally
1) no side effects always being better than side effects
2) the typical use case for me is like having a webapp which passes deferreds all around, and then I have some way of synchronizing those, but other internals there like "log deferred errors" still might be `addErrback`ing the deferred I get my hands on, and I'd still want those to work

src/twisted/internet/defer.py
```diff
@@ -159,6 +165,65 @@ def maybeDeferred(f, *args, **kw):
...
+    # return it or raise the given exception.
+    if not results:
+        raise InvalidStateError(
+            "Cannot call unbox with a Deferred that has pending work to be "
+            "done."
+        )
+    else:
+        [result] = results
+
+        if raises and isinstance(result, failure.Failure):
```
Julian: I think you'll run into trouble here if you have a function that callbacks with a failure (but maybe that will already fail in a place you don't control).

But like, doing:

`result, failure = [], []`

and then `addCallback` / `addErrback` would let you distinguish "properly".

Also I think you need tests both for this specifically and for the `BaseException` vs. `Exception` distinction (unless I missed an existing one for the latter).

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,121 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.unbox(d), result)
+
+
+    def test_unboxRaisesFailureException(self):
+        """
+        L{defer.unbox} should raise the original exception if the result of the
```
Julian: s/should raise/raises (and same with below)

I think generally AIUI we try to follow https://jml.io/pages/test-docstrings.html

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,121 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+        self.assertRaises(ValueError, defer.unbox, d)
+
+        # Without this the test fails because of an unhandled error.
+        d.addErrback(lambda _: None)
+
+    def test_unboxReturnsFailureWithNoRaises(self):
+        """
+        L{defer.unbox} should return the failure instance if raises=False.
+        """
+        d = defer.Deferred()
```
Julian, line 677: Might solve your initial motivation to add `passthrough` if you `def cbDeferred(self): d = Deferred(); d.callback(1234); return d` and `def ebDeferred(self, exception): d = Deferred(); d.addErrback(lambda _: None); d.errback(exception); return d` and did your `lambda _: None` there in the latter so you didn't have to keep repeating it.

src/twisted/topfiles/8900.feature
```diff
@@ -0,0 +1,2 @@
+Added twisted.internet.defer.unbox function to enable the unboxing of a
+"resolved" Deferred.
```
Julian: Seems like yet a 3rd language here on what we call "a deferred that's already fired".

Julian (changes requested):

src/twisted/internet/defer.py
dstufft: The reason I added `raise=False` was when I went to implement `successResultOf`, `failureResultOf`, and `assertNoResult` in terms it was hard to replicate the behavior of those functions without it:
- `successResultOf` uses the failure's traceback printing mechanisms which seem to be subtly different than Python's built in ones. If we're happy with the output on failure changing slightly then this one isn't a problem.
- `failureResultOf` returns the failure, which is impossible to do if `unbox()` doesn't provide that information to it.
- `failureResultOf` when there are expected exception types has the same problem as `sucessResultOf` with the failure message relying on details of failure's traceback formatting.
- `assertNoResult` wants to silence failures but _not_ successes, so the current `passthrough` parameter doesn't help it here, but a `consumeErrors` would.

In my own personal use cases I'm hoping to use the `unbox()` function in functions that can either be driven synchronously or asynchronously depending on what "fetcher" is passed into them (either sync or async). I want to just always return `Deferred` from these functions but I need a way to hide the fact that `Deferreds` are involved at all for end users who want to call the code synchronously. The expected way for this to look is something like (simplified):

``` python
import requests
import treq

def sync_fetcher(url):
    resp = requests.get(url)
    resp.raise_for_status()
    return resp.content

def async_fetcher(url):
    d = treq.get(url)
    d.addCallback(treq.content)
    return d


def api(thing, fetcher):
    d = Deferred()
    d.addCallback(fetcher)
    d.callback("https://example.com/api/{}".format(thing))
    return d


# When being used from async code
d = api("my-thing", fetcher=async_fetcher)
d.addCallback(print)

# When being used from sync code
print(unbox(api("my-thing", fetcher=sync_fetcher)))

```

The goal is to essentially treat the `unbox()` function sort of like string encoding/decoding in that it turns the code into "normal" sync like code that hides the fact `Deferreds` are in play. However, if errors don't get consumed then the above code is going to print a spurious error to stdout when the `Deferred` gets garbage collected. The way to work around this would be:

``` python
d = api("my-thing", fetcher=sync_fetcher)
try:
    print(unbox(d))
finally:
    d.addErrBack(lambda _: None)
```

Which fails at my goal of being able to tell people with "normal" sync looking code to just wrap the `Deferred` returning functions with an `unbox()` and forget about it.

src/twisted/internet/defer.py
dstufft: Ok, I'll switch it to `UnboxingWithoutResult`.

src/twisted/internet/defer.py
dstufft: I wasn't completely happy with the "has already fired its result" because of the behavior of:

``` python

d = Deferred()
d2 = Deferred()

d.addCallback(lambda _: d2)
d.callback(None)
```

Has the `Deferred` `d` "fired" its result? It's had `callback()` called on it so it seems to me like it had, but it's been chained to another `Deferred` that has not yet fired its result. Ultimately I don't really care much what the terminology is, I was just worried about that being confusing.

Julian:
Just putting this here for full context, some stream of consciousness discussion from IRC:

```
 12:34:37        tos9| dstufft: OK now maybe back to IRC would be helpful just for back and forth :/
 12:34:50        tos9| dstufft: Your use case looks similar to mine, but also a bit odd to me
 12:34:59  dstufft| what's the difference between d.addCallbacks(cb, eb) and d.addCallback(cb); d.addErrback(eb)
 12:35:01  dstufft| is there one?
 12:35:10        tos9| dstufft: The latter will catch even errors in the callback function
 12:35:42        tos9| dstufft: You want to hide users from deferreds, but still tell them they need to use a black box function to unbox them?
 12:36:25        tos9| One they especially have to write `from deferreds import unbox` or whatever to use?
 12:38:16     dstufft| tos9: well by "hide" I don't mean "make it so they never ever have to know that Deferreds are here" but more like "give a simple function they can use 
                       to avoid having to learn about Deferred or munge with them". This is so I don't have to add a foo() and a sync_foo() method to my API
 12:38:34        tos9| OK
 12:38:57        tos9| dstufft: Wouldn't it be even better though if your synchronous API had a way to inject things into your async API that handled the unboxing?
 12:39:04        tos9| and just plain not return a deferred in that case?
 12:39:55        tos9| like right now you're injecting a thing which you "know" makes all your deferred synchronous, but then still bubbling them up to people
 12:40:34        tos9| that kind of messes with your return types though. bleh. Sorry, my brain is still catching up.
 12:41:13        tos9| dstufft: (I have pretty much exactly your broad use case by the way, I have two clients, one using treq and one using requests)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 12:42:42     dstufft| tos9: Yea, this is for a thing in pip (well in packaging, but consumed by pip), and in pip I might just start threading Deferred throughout the code 
                       base even in the sync case (Which is all pip is going to be for now, full out sync, no event loop) but I want this library to be re-usable for people 
                       too, but I'm not like, super worried about telling sync users if they don't want to deal with the Deferred then unbox them
 12:43:52  dstufft| but if they have to mess around adding callbacks or errbacks to get a "normal" workflow where unbox() raises an error and doesn't spew errors on stdout 
                    even if they've been raised, then I'll need to either add my own unbox function that does the callbacks/errbacks for them, or add sync_foo
 12:45:07          * | tos9 nods
 12:45:29        tos9| dstufft: The thing that's bugging me is I'm still trying to decide whether your use case is exactly mine or not
 12:45:47        tos9| Because right now I'm basically treating my synchronization as plucking (peeking at) a deferred in its normal flow
 12:46:02        tos9| And the only thing that seems fairly "obvious" to me is that unbox should take exactly one of those two strategies
 12:46:15        tos9| either it's "peek at a deferred synchronously" or it's "end a deferred synchronously"
 12:46:42        tos9| but the two of us are taking opposite strategies :), so I'm trying to decide if you're right and I should change mine.
 12:46:46  dstufft| Yea, that's how I originally wrote unbox because it seemed like a simple enough thing to just return the result from the callback so that the deferred 
                    wasn't "ended" 
 12:47:01  dstufft| but ending is how I expected it to be used
 12:47:07  dstufft| in the common case
 12:47:10  dstufft| I could be wrong though!
 12:47:51  dstufft| it wasn't until I discovered that simply passing the failure through on errback did the "spew to stdout" thing :/
 12:49:12        tos9| dstufft: I think I have to ponder more to form an opinion that I'm confident -- or maybe someone else will weigh in
 12:49:23        tos9| maybe I'll try to rereview later tonight if that's OK and I have some more clarity?
 12:49:30     dstufft| tos9: that's fine :)
 12:49:36  dstufft| I'm fixing up the other things now
 12:50:08        tos9| k cool. Gonna also paste this whole thing into the ticket even though it's stream of consciousness just so that the full context is there
```

src/twisted/internet/defer.py
dstufft: I've added a test for the `BaseException` versus `Exception` distinction. However to the best I can tell `Deferred` itself treats `Deferred().errback()` the exact same as `Deferred().callback()` other than constructing the original `Failure` instance that will be passed through the chain.

IOW, it's impossible to have a `Deferred` that returns a `Failure`.

dstufft:
Github hid the original comment, but I think the main open questions are around raising the exception versus returning the failure and how to handle passthrough. You can see the original comments and follow up comments at https://github.com/twisted/twisted/pull/590#pullrequestreview-8317487.

exarkun:
This is a very frequently requested feature.  It's been turned down many times in the past.  It'd be nice to see some discussion here of why it makes sense to add it now when it never did before.

dstufft:
@exarkun Do you have any link to relevant discussions in the past? I tried searching the bug tracker but couldn't find any tickets open or otherwise but I don't think there is an obvious terminology for what this is called to search by. Since I don't have any context about why it was turned down previously I can't make an argument about why it makes sense now when it never did before.

I can tell you what my use case is though (and I did earlier, though I'm happy to reiterate), and @Julian had a similar use case. I also see some similar looking code in https://github.com/twisted/tubes/blob/master/tubes/undefer.py though not exactly the same since there is a `yield skip` in the middle of what would be the `unbox()` function here. I don't know enough about tubes to know if you could move that earlier or not. I've also been able to replace some other parts of Twisted using the `unbox()` function, so the idiom has been used at least once or twice inside of Twisted itself.

exarkun:
I don't have any links, sorry.  I'm sure there is some discussion in the twisted-python archives but it's pretty hard to find anything there.

The gist of the argument against this functionality is that in a normal program, the only way you know a Deferred has a result for you is if you are a callback on that Deferred and you have just been called.  In that case, you don't need this "unbox" function because the result is passed to you as an argument.

Any other time you might call such an "unbox" API, you have no idea if the Deferred has a result for you or not.  The only way you could use this API with any chance of success is to poll "unbox" and handle the "UnboxingWithoutResult" you get repeatedly.

@radix wrote https://github.com/radix/synchronous-deferred to solve the kind of problem issue 8900 described.  Apparently from his experience with that he decided any effect-like library is a better approach to the problem.

Twisted employs this give-me-the-result pattern in trial to help write tests where you actually _can_ know when there's a result (or if you're wrong, an exception is great, then your test can fail).  I don't recall any other places it's used, I'd be interested to look at them if you point them out.

The `yield skip` in the tubes code is important where it is.  That function is effectively just translating between two different callback APIs.  It's not reaching in to a Deferred for its result at a particular point: it's waiting for a result by adding a callback.

dstufft:
@exarkun For my use case, I'm using `Deferred` but there's no event loop or anything it's waiting on. The callbacks are just doing blocking IO inside of them. The reason for this is so that people can swap out the one place that IO is done at runtime with something that returns a `Deferred` itself and uses Twisted's or asyncio's event loop instead of just directly doing blocking I/O and have the thing just work.

In my use case I can pretty clearly say that either the `Deferred` has a result for you _always_ at that point in time or you should not be using `unbox()` because you're being driven via an event loop.

exarkun:
If they write code that uses `unbox` then they cannot swap out the implementation for one that uses an event loop.  If they do so, their `unbox`-calling code will suddenly start failing.

dstufft:
@exarkun Right, but my library will work both ways. They can either use it synchronously with `unbox` or asynchronously without by passing in a a sync or async implementation. They can even slowly start pulling the `unbox` further up the call stack to progressively convert more of the program to be `Deferred` supporting before adding an event loop.

exarkun:
It doesn't seem convincing that, because the API can be used with some Deferreds from one library when used in one mode, the API should be added to Twisted so it is available for use with all Deferreds.

Maybe you should put this API in your library?  Then it's close at hand for the audience to which it is relevant.

dstufft:
@exarkun If I have to do something like that I might, though if this keeps coming up and there are at least two people in the last day or two alone that have need for it, it seems like having multiple copies of this function in different libraries and applications is less than ideal.

Julian:
I'll admit I haven't yet wrapped my head entirely around @dstufft's use case, mostly because I'm so deep in my own (which only happened to touch on this since I did implement it locally as a one off) but --

@exarkun it does happen to be that my own use case _is_ in testing, but not in a place where I have access to a test case per se (I have a testing helper, and a lack of desire to write code that looks like `Helper(test_case=self)` just so I can access functionality I want "globally"). I.e., I literally want to call `successResultOf`, but I have no test case, and if in fact my assumption is wrong (that I somehow end up with a deferred that hasn't fired) I do want the exception. That was why I felt it was compelling enough to review this, was that to me it doesn't feel very much like creating something new, it's what's there, just without asking me to have a test case around.

I do obviously believe your skepticism though and again I haven't properly understood @dstufft's use case myself yet -- does it make you feel any less skeptical if this stayed in `twisted.trial`?

exarkun:
As a test helper, I'm believe it's useful (at least, I don't know of a better approach).  I'm pretty sure I'm even the one who committed `successResultOf`, `failureResultOf`, and `assertNoResult` to trial?  I reviewed the code, at least.  I don't have any problem with making these APIs more easily used in test code.  In general, it seems like at this point the pyunit inheritance-based approach to providing testing tools should have been proven suboptimal and we've only failed to move on to something better due to inertia.

Julian:
@exarkun that was a leading question because I knew you were the one :P

And yep full agreement on the unittest thing.

@dstufft how dirty does it make you feel to say what we're actually doing is improving the testing helpers, and then if you're really convinced that your API is how it should be, that you just use those in real code illicitly :D? Sorry for the runaround, it's what happens when you have to deal with someone [me] who's not really an expert in twisted lore (_cough_).

dstufft:
@Julian Well, if it's not included in what @hawkowl splits out into a separate deferred library then I can't use it in the main consumers of my library anyways (pip and setuptools) so I'll probably just add it to my library or skip the idea of using `Deferred` as my base primitive there all together.

Julian:
That sounds like a separate question to me -- I don't know what @hawkowl's plans are, but I'd assume / hope that `successResultOf` and friends would be part of a split-off-`Deferred`'s test helpers too.

Julian:
(FWIW personally I think the right next step rather than yanking your chain and asking for work that may or may not benefit you is for someone to actually understand your use case specifically and then make a call on whether it sounds like this is the right thing to add for it, but I still haven't done so, sorry :/)

dstufft:
This started with me going "Does anyone have any good patterns for making an API that can be used synchronously _and_ asynchronously?". Then @hawkowl suggested just using `Deferred` for both cases and pointed me at this snippet of code:

``` python
@attr.s
class URLResponse(object):
    url = attr.ib()
    response = attr.ib()
    headers = attr.ib()

def do_something_with_content(resp):
    print("got url", resp.url)
    return resp

@inlineCallbacks
def get_async(url):
    import treq
    resp = yield treq.get(url)
    content = yield resp.content()
    return URLResponse(url=url, headers=resp.headers, content=content)

def get_sync(url):
    import requests
    resp = requests.get(url)
    return URLResponse(url=url, headers=resp.headers, content=resp.content)

def _api(url, fetcher):

    d = Deferred()
    d.addCallback(lambda _: fetcher(url))
    d.addCallback(do_something_with_content)
    d.callback(None)
    return d

def async_api(url):
    return _api(url, get_async)

def sync_api(url):
    return _api(url, get_sync)
```

Which is great, it does basically everything I need, except that it forces every consumer of the library to write their software using the `Deferred` idiom which tends to be a bit infectious since (currently) every thing you do has to be registered as a callback of the `Deferred`. That makes it hard or impossible to use this hypothetical API in code that isn't already using `Deferred` everywhere unless you have an `unbox()` like function available to you from somewhere (either writing it yourself as a one off, or pulling it in from somewhere else).

I'm _trying_ to make something that both sync and async users can consume and I'm even willing to add a slight annoyance to sync users to make it happen. Currently though without something like this available (either in my own library or in Twisted or whatever) it feels like the cost of using `Deferred` is much higher than it's worth to make things that async users can easily consume since my main use case is sync.

I'm not sure how else to really describe my use case. To me it feels fairly obvious but I am probably too close to the problem.

src/twisted/internet/defer.py
```diff
@@ -160,6 +166,65 @@ def maybeDeferred(f, *args, **kw):
...
+    @param raises: Whether or not to raise the original exception (L{True}) or
+        to return the failure object (L{False}).
+
+    @type passthrough: bool
+    @param passthrough: Whether or not to pass through the original value
+        (L{True}) or to consume any results, error or otherwise (L{False}).
+
+    @raise UnboxingWithoutResult: If C{deferred} has pending work to be done
+        and has no result to unbox.
+
```
thijstriemstra, line 275: an @since marker would be useful here.

src/twisted/internet/defer.py
```diff
@@ -160,6 +166,65 @@ def maybeDeferred(f, *args, **kw):
...
 
+def unbox(deferred, raises=True, passthrough=True):
+    """
+    Unbox a L{Deferred} that has already fired its result.
+
+    Take a deferred that has already fired its result and unbox it, returning
+    the result of the L{Deferred}. If the L{Deferred} is in an error state then
+    it will re-raise the original exception.
+
+    B{This function CANNOT take a Deferred with pending asynchronous work to be
```
glyph: This would be better as an `@note`, I think.

src/twisted/internet/defer.py
```diff
@@ -160,6 +166,65 @@ def maybeDeferred(f, *args, **kw):
 
 
 
+def unbox(deferred, raises=True, passthrough=True):
```
glyph: I think that choosing a good name here is really, really important.

"unbox" is a bit jargon-y, and is only halfway accurate in this case.  Maybe something like `extractImmediateResult()`?

The docstring should also explain the one-and-only valid use-case for this function.  As @exarkun noted, lots of people want this function, except what they _actually_ want is "spin the reactor and block until a result is available", which is something we cannot do.

src/twisted/internet/defer.py
```diff
@@ -160,6 +166,65 @@ def maybeDeferred(f, *args, **kw):
...
+    """
+    Unbox a L{Deferred} that has already fired its result.
+
+    Take a deferred that has already fired its result and unbox it, returning
```
glyph: This docstring should also note what happens to the callback chain.  The result is *consumed*, whether it's a `Failure` or a success result.

src/twisted/internet/defer.py
```diff
@@ -160,6 +166,65 @@ def maybeDeferred(f, *args, **kw):
...
+    @raise UnboxingWithoutResult: If C{deferred} has pending work to be done
+        and has no result to unbox.
+
+    @return: The result of the L{Deferred}.
+    """
+    results = []
+
+    @deferred.addBoth
+    def done(result):
+        results[:] = [result]
```
glyph: This is _purely_ subjective, so please do with this feedback as you please, but I find the slicing/assigning idiom for extracting nonlocal variables to be somewhat hard to read, because you have to realize that the list isn't really a list, it's a holder.

Instead, I've started doing this:

```python
def done(result):
    done.result = result
nothing = object()
done.result = nothing
deferred.addBoth(done)
if done.result is not nothing:
    doSomethingWith(done.result)
else:
    "..."
```

glyph (changes requested):
Normally I wouldn't block _exclusively_ on an issue like choosing a good name, but this is an _enormous_ foot-gun that needs to be thoroughly explained, documented, and appropriately cautioned against.

It would also be good to make sure that it's possible to refactor `successResultOf` / `failureResultOf` to use this functionality, since they're doing basically the same thing.

src/twisted/internet/defer.py
dstufft: I can do that, I didn't really have an opinion so I just stole the technique you used in https://github.com/twisted/tubes/blob/master/tubes/undefer.py :)

src/twisted/internet/defer.py
dstufft: ``extractImmediateResult()`` is kind of long, what about just ``extractResult``? Or if that doesn't convey the meaning well enough, what about ``extractCurrentResult()`` which I think is clearer that it's what the current state is, not some future state and not "block and get a result".

glyph:
Responding to @exarkun : 

> It doesn't seem convincing that, because the API can be used with some Deferreds from one library when used in one mode, the API should be added to Twisted so it is available for use with all Deferreds.
> 
> Maybe you should put this API in your library? Then it's close at hand for the audience to which it is relevant.

There have been a long series of discussions on various bugtrackers, on IRC, and in person at PyCon about this functionality, and various representatives of Twisted (myself and @hawkowl) have been present and suggesting that it would make sense for it to appear in Twisted itself - but we didn't record that very well when this ticket was filed.

So, the case is as follows:

Deferred, as a data structure, is better-suited than Future to a semi-synchronous API, because a Deferred (unlike a Future) has no coupling to an event loop.

There are a number of libraries and applications (requests, and pip via requests) that are beginning a slow transition towards a fully asynchronous API, since with the addition of `asyncio`, event-driven has gone from the unofficial best concurrency model of Python to the only official, blessed concurrency model.  In the meanwhile, though - especially in the case of requests - there are about a trillion lines of synchronous Python code that use it which need to be supported for the forseeable future.

Therefore, such libraries need an _internally_ asynchronous data structure that they can use for passing results around through their various layers, while still exposing the same synchronous API on top.  Deferreds would make a great choice for this layer.

Using Futures for this is a problem because of their aforementioned coupling to an event loop.  Using `await` syntax is a non-starter because of the very long tail of Python 2 which still needs to be supported.  Using Effect presents other problems - it's only lightly maintained right now (we just passed the one-year mark since any changes were made).  It's also designed to facilitate a much larger change to a codebase than simply "make it async" - things like pip and requests which have an expectation of "just blocking" would need to internally refactor to use Deferreds as their return types in order to use Deferreds, but in the case of Effects they'd separately need to implement their own custom sync performers _and_ loop-specific async performers, as well as have a performer harness to integrate with event loops.  Deferreds allow for abstracting over more of this.

Further, continuing the trend of splitting out micro-libraries which other projects are interested in, we intend to follow the example of `constantly` and `incremental` and split out `twisted.internet.defer` as `deferred` on PyPI so that these projects can use it directly.

The reason it makes sense to include this functionality in Twisted is that once said library is available, we want this (i.e. "depend on `deferred`") to be the _preferred_ way of evolving a library towards async I/O support (and thus also possibly `asyncio` support).  Having to install a separate dependency just for this one function, and/or re-implement it in every project, would be a real sub-optimal compromise.

The reason @dstufft is doing this now, rather than waiting past that point, is to develop a proof-of-concept of such a split-out.

I hope that this explains the motivation to any onlookers!

src/twisted/internet/defer.py
dstufft: What exactly happens to the callback chain is currently sort of up for discussion? For my use case I want the result consumed and the deferred is finished (and if we don't at least consume the failure cases we get an ugly output when the Deferred gets garbage collected). However @Julian had a use case where he just wanted to "peek" into the deferred and not munge the results of it at all. 

In addition when I was implementing ``successResultOf`` and ``failureResultOf`` in terms of this (which is already done in this PR btw) I needed a variety of behaviors so I ended up making this function take two boolean arguments to control what happens to the result and what happens if it's an error.

Obviously given pretty much any behavior (except ``raises=True`` being the only behavior) you can implement the other behaviors as a wrapper around this, however that then loses some of the benefit of providing this function in the first place I think.

I'm not sure though, for me I think ``raises=True, passthrough=False`` because that makes "incorrect" values for ``raises`` / ``passthrough`` more obvious and I think it matches the most common use case for this function best (wanting to turn ``Deferred`` using code into a sync function).

dstufft:
Just a note since @glyph called it out explicitly:

This PR already has reimplemented ``successResultOf`` and ``failureResultOf`` using the ``extractCurrentResult`` function.

I've also renamed ``unbox()`` to ``extractCurrentResult()``. I believe that this should satisfy @glyph's concerns re: the naming, and I think it better describes what it's doing than ``extractImmediateResult()``, which could be interpreted to mean "immediately finish running the ``Deferred`` and then extract the result".

hawkowl:
FWIW, I have seen 'current' result used elsewhere when referring to a Deferred.

exarkun:
@glyph It would probably be good to capture some of that thought process somewhere easier to find.  What are the chances future maintainers of this API are going to come read the full PR discussion?

dstufft:
@exarkun Where would that get captured at? I did add a little bit to the doc string but I'm not sure if there is a better place?

exarkun:
@dstufft The docstring seems like it covers the right stuff now.  Part of glyph's point was that there's a bigger picture here and that's guiding certain decisions.  It  seems like that big picture should be capture somewhere (else, it's difficult for people who weren't in the referenced conversations to be guided by it).  This feels more like some kind of "vision" doc than anything else.  Maybe it's just a blog post on labs, maybe it's a wiki page.

dstufft:
@exarkun Ok! I'll leave that to @glyph or someone else to write the blog post/wiki/vision documentation then :)

glyph:
@exarkun 

> It would probably be good to capture some of that thought process somewhere easier to find. What are the chances future maintainers of this API are going to come read the full PR discussion?

Agreed. I wrote it here first because I don't know where the right place is to get this in front of maintainers. I started thinking about that but figured I'd better just get my thoughts down here given the opportunity, they can be copied and pasted elsewhere without much trouble :)

In the future, stuff like this really needs to go on the mailing list as it is discussed. (Which is totally my bad.)  Especially if we have an in-person discussion that might affect the project, someone needs to write it up and give everyone a heads up.  Generally we need to use the ML more, especially as cross-project discussions fracture across multiple different IRC channels.  In the future I'll try to fire off a message when each discussion occurs.

I think it might be a little premature to put it somewhere official-looking like labs.tm or the docs, but a wiki page might be appropriate at this point.

glyph:
As I was writing that, I realized that a simple link might do the trick.  I posted to the mailing list [here](https://twistedmatrix.com/pipermail/twisted-python/2016-December/030905.html) about this thread.

termim:
There is a typo in the docstring - look for "originak"

glyph:
Another thought about this PR:

With this function avoiding the foot-gun nature of extracting a synchronous result as much as possible, the ill-considered `.result` attribute of `Deferred` itself really ought to be deprecated.

dstufft:
@glyph presumably that would be another PR?

glyph:
Oh, yes, absolutely. Just noting it.

src/twisted/internet/defer.py
```diff
@@ -53,6 +53,13 @@ class TimeoutError(Exception):
 
 
 
+class NoCurrentResult(BaseException):
+    """
+    This error is raised when attempting to call extractCurrentResulton a
```
exarkun: Could be `L{extractCurrentResult}`.  Also, missing space before the "on" here.

src/twisted/internet/defer.py
```diff
@@ -53,6 +53,13 @@ class TimeoutError(Exception):
...
+    L{Deferred} that has not yet fired all of its callbacks.
```
exarkun: Typically, we say the Deferred fires or is fired.  The callbacks are just called, I think.  Unless the lingo has moved on while I wasn't paying attention.

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
 
 
+def extractCurrentResult(deferred, raises=True, passthrough=False):
+    """
+    Extract the current result from a L{Deferred} that has already fired all of
+    its callbacks.
+
+    Take a deferred that has already fired all of its callbacks and extract and
+    return the result of the L{Deferred}. This process of extracting the result
+    will, by default, consume the result, error or not unless the
```
exarkun: I think ` ... consume the result - error or not - unless the ...`.  Certainly not just a single comma before the "error or not", though.  The aside is just "error or not" and it should be delimited on both ends by some punctuation.  This could still be commas but there's already a lot of commas here so I think dashes would be less confusing.

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
+    L{passthrough} parameter is set to L{True}. If the L{Deferred} is in an
```
exarkun: "has an error result"?  "is in an error" seems weird.

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
+    error then the process of extracting the result will, by default, re-raise
+    the original exception unless the L{raises} parameter is set to L{False}.
+
+    The primary purpose of this function is to allow code to be written that is
+    internally asynchronous but which can be consumed without an event loop in
+    an synchronous fashion by using L{extractCurrentResult} at the boundaries.
```
exarkun: "a synchronous ..."

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
+    @type deferred: Deferred
+    @param deferred: The Deferred from which results will be extracted..
+
+    @type raises: bool
+    @param raises: Whether or not to raise the original exception (L{True}) or
+        to return the failure object (L{False}).
+
+    @type passthrough: bool
+    @param passthrough: Whether or not to pass through the original value
+        (L{True}) or to consume any results, error or otherwise (L{False}).
```
exarkun: I'd prefer different functions for the behavior granted by these different flags.  For example, `extractCurrentResult` seems like it should always have the `passthrough=False` behavior.  That's a reasonable interpretation of "extract".  The `passthrough=True` behavior could be exposed separately as (for example) `observeCurrentResult`.  "Observe" is reasonably interpreted as not modifying the target object (except to quantum physicists of course :wink:).


Providing _four_ different APIs for all combos of the two flags is probably ridiculous and unhelpful and I don't have any suggestions for the names of things along the `raises` dimension anyway, so (unless you have some ideas) just fixing this for `passthrough` would be okay with me.

While I have no doubt this would produce a better API, this is clearly a bikeshed.  I leave the decision to you.

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
+    @raise NoCurrentResult: If C{deferred} has pending work to be done
+        and has no current result to extract.
+
+    @return: The result of the L{Deferred}.
+    """
+    def done(result):
+        done.result = result
+
+        # We return the result here so that if anything else continued to use
+        # this Deferred after it's been unboxed then the act of unboxing
```
exarkun: might be clearer not to talk about unboxing in the comments anymore

src/twisted/internet/defer.py
```diff
@@ -160,6 +167,87 @@ def maybeDeferred(f, *args, **kw):
...
+    deferred.addBoth(done)
+
+    # If we've gotten to this point and we haven't had our collect function
+    # called, then we must be waiting on something in the Deferred chain. If so
+    # we'll error out here because there's nothing we can do to advance the
+    # state of the Deferred. If we've got a result then we'll go ahead and
+    # return it or raise the given exception.
+    if done.result is nothing:
+        raise NoCurrentResult(
+            "Cannot call extractCurrentResult with a Deferred that has not "
```
exarkun: Clearly one *can* **call** it :wink:

src/twisted/internet/defer.py
```diff
@@ -53,6 +53,13 @@ class TimeoutError(Exception):
 
 
 
+class NoCurrentResult(BaseException):
```
exarkun, line 56: I'd like to see some maintainer-oriented documentation about the decision to make this a `BaseException` instead of an `Exception`.

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,150 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+                defer.extractCurrentResult(d)
+            except Exception:
+                return "error"
+            return "success"
+
+        d = defer.Deferred()
+        self.assertRaises(defer.NoCurrentResult, inner, d)
+
+        # In order to get 100% patch coverage we need to excerise the pass
+        # statement. This is a bit silly because we're basically testing our
```
exarkun: This talk about a "pass statement" which is confusing.  I think it must mean the `return "error"` statement which maybe used to just be `pass`?

I don't know that calling this silly is helping anyone (except maybe you, to blow off steam after having to write some code you thought was pointless to satisfy a policy you disagree with).  Just explaining why you run through the code twice here seems sufficient.

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,150 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+        """
+        L{defer.extractCurrentResult} does not pass through results.
+        """
+        result = object()
+
+        d = defer.Deferred()
+        d.callback(result)
+
+        self.assertIs(defer.extractCurrentResult(d), result)
+        self.assertIsNot(defer.extractCurrentResult(d), result)
```
exarkun: Assert that the result becomes `None` perhaps?  Asserting it's not the result leaves the field wide open for what the behavior actually is.  I don't see any reason to let the test suite leave the intent this wide.

src/twisted/test/test_defer.py
```diff
@@ -645,6 +645,150 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+        L{defer.extractCurrentResult} does not pass through errors.
+        """
+        d = defer.Deferred()
+        d.errback(ValueError())
+
+        # Reach inside the deferred and pull out the result.
+        result = d.result
+
+        self.assertIs(defer.extractCurrentResult(d, raises=False), result)
+        self.assertIsNot(defer.extractCurrentResult(d), result)
```
exarkun: Ditto comment for the non-error case.

src/twisted/topfiles/8900.feature
```diff
@@ -0,0 +1,2 @@
+Added twisted.internet.defer.extractCurrentResult function to enable extracting
```
exarkun: The style guide for news fragments suggests "The new function twisted.internet.defer.extractCurrentResult ..."

src/twisted/trial/_synctest.py
```diff
@@ -679,20 +680,24 @@ def successResultOf(self, deferred):
...
-        deferred.addBoth(result.append)
-        if not result:
+        try:
+            result = defer.extractCurrentResult(deferred, raises=False)
+        except defer.NoCurrentResult:
             self.fail(
-                "Success result expected on %r, found no result instead" % (
-                    deferred,))
-        elif isinstance(result[0], failure.Failure):
+                "Success result expected {0!r}, found no result "
```
exarkun: This dropped the "on" which is important to the meaning.

src/twisted/trial/_synctest.py
```diff
@@ -679,20 +680,24 @@ def successResultOf(self, deferred):
...
+                "instead".format(deferred)
+            )
+
+        if isinstance(result, failure.Failure):
             self.fail(
-                "Success result expected on %r, "
-                "found failure result instead:\n%s" % (
-                    deferred, result[0].getTraceback()))
+                "Success result expected on {0!r}, "
+                "found failure result instead:\n{1}".format(
```
exarkun: Any particular reason to use `{0}` and `{1}` instead of just `{}`?

exarkun (changes requested):
Thanks!  I've been meaning to get to this for a while now, sorry it took so long.  I'll try to be quicker on the next iteration.

src/twisted/internet/defer.py
dstufft: By documentation do you just mean a comment, or do you explicitly want documentation that would show up on the website?

src/twisted/trial/_synctest.py
dstufft: Habit for Python 2.6 compatibility.

src/twisted/internet/defer.py
exarkun: I think it'd be good if users could find it without diving into the source code.  That could still just be part of a docstring (and maybe `extractCurrentResult`\ 's docstring if that makes more sense) since those go into the generate api docs served on the website.

If it doesn't easily fit into the API docs, though, I'd be satisfied with a comment.  At least then the info is recorded and someone might be able to improve the doc situation later on.

dstufft: I've added documentation to speak to this.

dstufft:
Just as a note, I've taken the action that @exarkun asked for and I've removed the ``passthrough=True/False`` parameter to ``extractCurrentResult``.

There are now two functions, ``extractCurrentResult`` and ``observeCurrentResult`` which have pretty much the same functionality, except ``extractCurrentResult`` will set the current result of the ``Deferred`` to ``None`` and ``observeCurrentResult`` will preserve the current result. They both still take a ``raises`` parameter.

exarkun:
Thanks.  See #690.

PR #591, opened 2016-11-14
Stop reading from TCP connection if the client isn't reading from it.

Lukasa:
This patch addresses *literally exactly* the problem in https://twistedmatrix.com/trac/ticket/8868. This does not fully address this style of attack, only literally this instance.


src/twisted/internet/producer_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Stop streaming data to the consumer.
+        """
+        if self._finished:
+            return
+        self._finished = True
+        self._coopTask.stop()
+
+
+    # IPushProducer implementation:
```
hawkowl: Twistedchecker thinks this comment is also a blank line :(

src/twisted/internet/producer_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+    def pauseProducing(self):
```
hawkowl: Docstrings here (`@see IProducer.whatever` will suffice)

src/twisted/internet/test/test_tls.py
```diff
@@ -340,16 +340,13 @@ class AbortSSLConnectionTests(ReactorBuilder, AbortConnectionMixin, ContextGener
 
     def buildReactor(self):
         reactor = ReactorBuilder.buildReactor(self)
-        try:
-            from twisted.protocols import tls
-        except ImportError:
-            return reactor
+        from twisted.internet import producer_helpers
 
         # Patch twisted.protocols.tls to use this reactor, until we get
```
hawkowl, line 345: Is this comment still relevant?

src/twisted/test/proto_helpers.py
```diff
@@ -837,6 +838,46 @@ def connectUNIX(self, address, factory, timeout=30, checkPID=0):
...
+    A pull producer which writes 10 times only.
+    """
+
+    counter = 0
+    stopped = False
+
+    def __init__(self, consumer):
+        self.consumer = consumer
+        self.result = Deferred()
+
```
hawkowl, line 852: Extra newline required here

src/twisted/internet/producer_helpers.py
```diff
@@ -0,0 +1,113 @@
+# -*- test-case-name: twisted.test.test_producer_helpers -*-
```
hawkowl: Why is this a public module when it has no public items in it?

src/twisted/internet/producer_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+# Copyright (c) Twisted Matrix Laboratories.
```
hawkowl: This module also needs an `__all__`

hawkowl (dismissed):
Please address the comments + the twistedchecker errors, then merge.

src/twisted/internet/test/test_tls.py
Lukasa: I think so: 5206 was never fixed, but I think the breadcrumb is still useful.

Lukasa:
Ok so txchecker genuinely wants it both ways. There is no way to have txchecker accept a comment before a method in Twisted: if there is one empty line then the comment it is too few blank lines, if there are two empty lines then there are too many blank lines.

I view this as a txchecker bug.

PR #592, opened 2016-11-14
Time out HTTP clients after 60 seconds of inactivity.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/3746


glyph:
A+. Will merge once updated branch passes tests.

PR #594, opened 2016-11-17
Fix for the KeyError exception

kgkdev13:
This pull request fixes the KeyError exception like this:

```
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/var/srv/websocket/modules/managers/queuemanager.py", line 164, in queueProcessor
    client.dropConnection()
  File "/usr/local/lib/python2.7/dist-packages/autobahn-0.16.0-py2.7.egg/autobahn/websocket/protocol.py", line 861, in dropConnection
    self._closeConnection(abort)
  File "/usr/local/lib/python2.7/dist-packages/autobahn-0.16.0-py2.7.egg/autobahn/twisted/websocket.py", line 139, in _closeConnection
  File "/usr/local/lib/python2.7/dist-packages/Twisted-15.4.0-py2.7-linux-x86_64.egg/twisted/internet/_newtls.py", line 217, in loseConnection
    FileDescriptor.loseConnection(self)
  File "/usr/local/lib/python2.7/dist-packages/Twisted-15.4.0-py2.7-linux-x86_64.egg/twisted/internet/abstract.py", line 407, in loseConnection
    self.stopReading()
  File "/usr/local/lib/python2.7/dist-packages/Twisted-15.4.0-py2.7-linux-x86_64.egg/twisted/internet/abstract.py", line 421, in stopReading
    self.reactor.removeReader(self)
  File "/usr/local/lib/python2.7/dist-packages/Twisted-15.4.0-py2.7-linux-x86_64.egg/twisted/internet/epollreactor.py", line 333, in removeReader
    EPOLLIN, EPOLLOUT)
  File "/usr/local/lib/python2.7/dist-packages/Twisted-15.4.0-py2.7-linux-x86_64.egg/twisted/internet/epollreactor.py", line 319, in _remove
    del selectables[fd]
KeyError: 306
```


glyph:
Thanks for contributing to Twisted!  Did you open a ticket in Trac, as specified in https://github.com/twisted/twisted/blob/trunk/CONTRIBUTING ?

kgkdev13:
The trac is here https://twistedmatrix.com/trac/ticket/8907

kgkdev13:
I can't guess why codecov is not passing. What do I need to do?

acabhishek942:
@kgkdev13 you need to write a test where the condition `if fd in selectables:` evaluates to False to go from partial coverage to full coverage. If you install codecov chrome extension(assuming you are on chrome) you can see partial coverage of that condition. Here is the extension https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf

kgkdev13:
Completed tests. Check pls

src/twisted/internet/test/test_fdset.py
```diff
@@ -115,6 +115,19 @@ def fail():
...
 
+    def test_removeReader_selectables_nokey(self):
+        """
+        L{reactor.removeReader()} accepts an L{IReadDescriptor} provider
+        previously passed to C{reactor.addReader()}. Checks if it doesn't
+        raise KeyError exception
+        """
+        reactor, fd, server = self._simpleSetup()
+        reactor.addReader(fd)
+        reactor._selectables = {}
```
exarkun, line 126: Can you provoke this failure in a more realistic way?  It doesn't seem likely that the affected application goes around erasing the reactor's internal state.  A realistic test would provide more confidence that this really fixes the problem.  Also, it would embody a more complete understanding of the cause of the failure which will assist future maintainers of the reactor and test suite.

glyph: I also feel like I should note if an application _does_ go around erasing the reactor's internal state, that is specifically unsupported :)

kgkdev13: If I could know about a real problem.. I just see that after some days of work my WebSocket Server raises the exception above and it stops accepting new connections. So I need to restart it.
I looked at traceback and thought that it would be better to fix in the sources of twisted.
Do you have any ideas about why such error happens?

exarkun: Can you reproduce the problem?  I don't think most people encounter this when using epollreactor.  That suggests the bug isn't that epollreactor just randomly and arbitrarily forgets about file descriptors sometimes - but that something about your application/environment/usage provokes this bug.  If we could understand what it is that does that, it might give us some clues about what's wrong - and thus suggest a better test.

DefaultUser: I have the same problem and it seems to be related to `twisted.internet.threads.deferToThread()` - at least for me. I created a .rpy script that triggers this exception after several connection attempts (see https://gist.github.com/DefaultUser/f8bd6b1cb60fc260fd397113878bebeb). It takes a while till it actually happens, so you need to spam F5 for a while

exarkun: Thanks for the contribution @DefaultUser !  Having a reproducable example of this problem makes it much easier to understand what's wrong.

In the case of the linked rpy script, the problem looks like it's in the application code.  It uses `deferToThread` to schedule a function to run in a non-reactor thread.  Then, that function calls a Twisted API directly (`request.write` and `request.finish`).  This isn't supported and broken behavior as a result is just what I'd expect.  Instead, the Twisted APIs need to be invoked by calling *back* into the reactor thread.  Something like:

    def _render_thread(self, request):
        time.sleep(1)
        def do_it():
            request.write("<html><body>something from a thread</body></html>")
            request.finish()
        reactor.callInThread(do_it)

(Not that this is an example of _good_ multithreaded coded - it's almost certainly an example of _bad_ multithreaded code, in fact.  But this is the smallest possible change to the example code to make it use the Twisted APIs correctly.)

Again, thanks for sharing this example.  It helps to know what kind of behavior is possible from this misuse of the Twisted APIs (this might help other people diagnose such misuses).  In fact, perhaps the original reporter will report that this is analogous to his or her circumstances and then we'll finally understand the original problem report.  I hope so.

schlenk: I see this a lot with ldaptor running under Crochet, so might be a similar problem with crossing thread boundaries there.

exarkun: A minimal reproducing example would go a long way towards helping make progress on this.

twm:
Based on [@exarkun's review above](https://github.com/twisted/twisted/pull/594#pullrequestreview-9365272) and the threading details at the top of the traceback, specifically:

```pytb
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
...
```

I believe that the cause of this issue is use of reactor APIs from the wrong thread. I'm going to close this PR and the associated ticket. However, if I'm wrong about this, please feel free to reopen.

PR #595, opened 2016-11-17
Propagate reactor.callLater to the HTTPChannel.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/8904


src/twisted/web/http.py
```diff
@@ -2296,6 +2300,17 @@ def timeOut(self, value):
         self._channel.timeOut = value
 
 
+    @property
+    def callLater(self):
```
markrwilliams, line 2673: This should have a docstring.

src/twisted/web/http.py
```diff
@@ -2296,6 +2300,17 @@ def timeOut(self, value):
...
+        return self._channel.callLater
+
+
+    @callLater.setter
+    def callLater(self, value):
```
markrwilliams, line 2682: This should also have a docstring.

src/twisted/web/http.py
markrwilliams: (Also thank goodness that `proxyForInterface` generates a class that inherits from `object`!)

src/twisted/web/http.py
```diff
@@ -2296,6 +2300,17 @@ def timeOut(self, value):
         self._channel.timeOut = value
 
 
+    @property
+    def callLater(self):
+        return self._channel.callLater
```
markrwilliams: For the sake of clarity:  `HTTPChannel` inherits from `TimeoutMixin`, which defines a `callLater` method that just imports the reactor and calls _its_ `callLater`.  So patching this patches _that_.

markrwilliams (commented):
Looks good!  Just missing some docstrings.

src/twisted/web/http.py
Lukasa: Does this need a docstring? I've been documenting all of these as ivars because that's how they appear in use.

markrwilliams: Both [`pydoctor`](https://twistedmatrix.com/documents/16.5.0/api/twisted.internet.task.Cooperator.html#running) and [`twistedchecker`](https://github.com/twisted/twistedchecker/blob/25350dfd35a403198accf0d423e5cba9ac2ddec8/twistedchecker/checkers/docstring.py#L146-L177) (note that there's an exemption for setters, but not getters) expect `property` methods to have docstrings.  It does seem odd that they're not documented as `ivar`s, since the fact that they're descriptors is an implementation detail.

I'm guessing this is because the coding standard requires all methods to have docstrings, and these are fundamentally methods.

glyph (approved):
This seems like a bug which makes testing harder. Please proceed!

PR #596, opened 2016-11-18
https://twistedmatrix.com/trac/ticket/8906#ticket

glyph:
here we go


moshez (approved):
Manually checked https://travis-ci.org/twisted/twisted/jobs/176914894
Looks good.

PR #597, opened 2016-11-18
Pass the factory and site to the H2Stream constructor.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/8909


src/twisted/web/test/test_http2.py
```diff
@@ -1425,6 +1425,45 @@ def test_resetAfterBody(self):
...
+
+        requestBytes = factory.clientConnectionPreface()
+        requestBytes += buildRequestBytes(self.getRequestHeaders, [], factory)
+        connection.makeConnection(transport)
+        connection.dataReceived(requestBytes)
+
+        # We need to wait for the stream cleanup callback to drain the
+        # response.
+        cleanupCallback = connection._streamCleanupCallbacks[1]
+        return defer.gatherResults([d, cleanupCallback])
```
adiroiban, line 1370: is this async code?

can we just use `self.successResultOf(cleanupCallback)`

Maybe this can also be written as

```
connection = self.getConnection(requestFactory=SuperRequest)

```

and then  

```
def getConnection(self, requestFactory):
    connection = H2Connection()
    connection.requestFactory = requestFactory

    def drain_response():
        cleanupCallback = connection._streamCleanupCallbacks[1]
        self.successsResultOf(cleanupCallback)

    self.addCleanup(drain_response)

    return connection
```

src/twisted/web/test/test_http2.py
```diff
@@ -1425,6 +1425,45 @@ def test_resetAfterBody(self):
...
+        def validateFactoryAndSite(args):
+            site, factory = args
+            self.assertIs(site, connection.site)
+            self.assertIs(factory, connection.factory)
+        d.addCallback(validateFactoryAndSite)
+
+        requestBytes = factory.clientConnectionPreface()
+        requestBytes += buildRequestBytes(self.getRequestHeaders, [], factory)
+        connection.makeConnection(transport)
+        connection.dataReceived(requestBytes)
```
adiroiban: can we have this extracted into something like

```
def connectAndReceive(self, connection):
    factory = FrameFactory()
    transport = StringTransport()
    requestBytes = factory.clientConnectionPreface()
    requestBytes += buildRequestBytes(self.getRequestHeaders, [], factory)
    connection.makeConnection(transport)
    connection.dataReceived(requestBytes)
```

I had to read the code more than once to see what exactly is `factory` what is `httpFactory` and what is `site, factory = args`

with this extraction, the pieces which are not relevant to the test should be moved away and should make the test easier to read

src/twisted/web/test/test_http2.py
Lukasa: Yeah, this is async code. The H2 implementation needs to handle priority, which means that it using a `LoopingCall` to send data. This is an annoying bit of indirection that I don't like much, but it does work. =) Regardless, it means that we do need the `defer.gatherResults` here to make sure we don't leave any looping calls running.

src/twisted/web/test/test_http2.py
Lukasa: I'm open to doing that.

src/twisted/web/test/test_http2.py
adiroiban: maybe in this case the looping call code should allow injecting a reactor so that we can control the loop iterations... but this is a minor comment and I am ok to merge it as it is

Lukasa: If you look at the `H2Connection` constructor you'll see that it does, in fact, allow injecting a reactor.

adiroiban: OK. Sorry for the noise. Since you are not injecting the reactor in this test, I was assuming that the current implementation does not support its... as otherwise why would you want to spin the whole reactor, when you can get full control over the execution of this code ? :)

Lukasa: =) Yeah, so we could avoid it, but I'm not sure what the best solution for that is right now. For the moment spinning the whole reactor is fine, but we may want to consider whether we should refactor some of these tests to inject custom reactors when we don't need the full-sized one.

PR #598, opened 2016-11-19
Explicitly ignore extension modules.

tomprince:
twistedchecker blows up on extension modules. This explicitly ignores the ones in twisted's tree. See https://github.com/twisted/twistedchecker/issues/118 for the upstream bug.


tomprince:
https://twistedmatrix.com/trac/ticket/8910#ticket

PR #600, opened 2016-11-20
Port pyhtmlizer to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8913


src/twisted/python/htmlizer.py
```diff
@@ -6,23 +6,33 @@
 HTML rendering of Python source.
 """
 
-import tokenize, cgi, keyword
+from twisted.python.compat import _PY3
+
+import tokenize, keyword
+if _PY3:
```
glyph: `_PY3` conditionals should be avoided as much as possible in application code.  The first line of defense is checking `twisted.python.compat` for existing stuff, and in this case `twisted.python.compat.escape` is already `html.escape` on py3 and `cgi.escape` on py2.

src/twisted/python/htmlizer.py
```diff
@@ -6,23 +6,33 @@
...
     currentCol, currentLine = 0, 1
     lastIdentifier = parameters = 0
+    encoding = "utf-8"
 
     def __init__(self, writer):
         self.writer = writer
 
+
     def printtoken(self, type, token, sCoordinates, eCoordinates, line):
+        if _PY3 and type == tokenize.ENCODING:
```
glyph: The second line of defense is module-level imports and `ImportError`.  This should instead be something like:

```python
try:
    from tokenize import ENCODING
except ImportError:
    ENCODING = 'ENCODING_NOT_DEFINED_HERE'

```
This way, if a dot-release of py2 ever starts defining encoding (which it might do, given that `-*- coding -*-` comments still do something on py2) it will silently start working, rather than forcibly breaking on py2 even if it might work.

src/twisted/python/htmlizer.py
```diff
@@ -60,35 +71,58 @@ def __init__(self, writer):
         reflect.accumulateClassList(self.__class__, "noSpan", noSpan)
         self.noSpan = noSpan
 
+
     def write(self, token, type=None):
-        token = cgi.escape(token)
+        if _PY3 and isinstance(token, bytes):
```
glyph: Is there any reason this can't be an unconditional `.decode("utf-8")` for bytes, even on py2?

src/twisted/python/htmlizer.py
```diff
@@ -60,35 +71,58 @@ def __init__(self, writer):
...
+            token = token.decode("utf-8")
+        token = escape(token)
+        if _PY3:
+            token = token.encode("utf-8")
```
glyph: Same here.

src/twisted/python/htmlizer.py
```diff
@@ -60,35 +71,58 @@ def __init__(self, writer):
...
 
     Results in much smaller HTML output.
     """
     noSpan = ["endmarker", "indent", "dedent", "op", "newline", "nl"]
 
+
+
 def filter(inp, out, writer=HTMLWriter):
-    out.write('<pre>')
+    if _PY3:
```
glyph: This conditional import should go in `twisted.python.compat`.

src/twisted/python/htmlizer.py
```diff
@@ -60,35 +71,58 @@ def __init__(self, writer):
...
-    out.write('</pre>\n')
+    out.write(b'</pre>\n')
+
+
 
 def main():
     import sys
-    with open(sys.argv[1]) as f:
-        filter(f, sys.stdout)
+    if _PY3:
```
glyph: The idiom here ought to be `getattr(sys.stdout, "buffer", sys.stdout)`, since not all instances of `sys.stdout` will have a buffer, even on py3.

glyph (approved):
I have lots of general nitpicking about the use of `_PY3`, but they're all just general rules you should follow, and not really important enough to block this from going ahead.  Address the comments if you like, but feel free to merge even without changes.

PR #603, opened 2016-11-21
[#6800] CertificateOptions TLS protocols

hawkowl:
https://tm.tl/#6800


src/twisted/internet/_sslverify.py
```diff
@@ -1521,10 +1594,43 @@ def __init__(self,
...
         if method is None:
-            # If no method is specified set things up so that TLSv1.0 and newer
-            # will be supported.
             self.method = SSL.SSLv23_METHOD
-            self._options |= SSL.OP_NO_SSLv3
+
+            if tlsProtocols is None:
+                # If no method or protocols are specified set things up so that
+                # TLSv1.0 and newer will be supported.
+                self._options |= SSL.OP_NO_SSLv3
```
Lukasa: Is it worth setting `OP_NO_SSLv2` here as well, just to be on the safe side?

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,65 @@
...
+
+        if versions.index(fromVersion) > versions.index(toVersion):
+            raise ValueError(
+                "fromVersion must be an earlier version than toVersion")
+
+        if fromVersion is TLSVersion.Highest:
+            raise ValueError(
+                ("fromVersion cannot be 'Highest', to use just the highest "
+                 "version, pass it directly to CertificateOptions, not in a "
+                 "range"))
```
Lukasa: Is there any reason to enforce this constraint here?

src/twisted/internet/_sslverify.py
```diff
@@ -1521,10 +1594,43 @@ def __init__(self,
...
+
+                if tlsProtocols._toVersion is not TLSVersion.Highest:
+                    for x in versions[versions.index(
+                            tlsProtocols._toVersion):]:
+                        if (x is not tlsProtocols._toVersion and
+                            x is not TLSVersion.Highest):
+                            excludedVersions.append(x)
+
+                for version in excludedVersions:
+                    self._options |= _tlsDisableFlags[version]
```
Lukasa: It may be worth factoring this whole code block out to a helper function.

Lukasa (dismissed):
I have a few minor notes, but generally speaking this patch is in good shape.

src/twisted/internet/_sslverify.py
hawkowl: it's in the base `self._options` above

src/twisted/internet/_sslverify.py
hawkowl: yes and that it makes my brain hurt

docs/core/howto/ssl.rst
```diff
@@ -1,8 +1,3 @@
-
-:LastChangedDate: $LastChangedDate$
-:LastChangedRevision: $LastChangedRevision$
-:LastChangedBy: $LastChangedBy$
-
```
exarkun, line 5: What's with this change?

docs/core/howto/ssl.rst
```diff
@@ -204,10 +199,40 @@ For example,
     options = CertificateOptions(..., dhParameters=dhParams)
 
 Another part of the TLS protocol which ``CertificateOptions`` can control is the version of the TLS or SSL protocol used.
-This is often called the context's "method".
+There are two ways to configure it, either by configuring a list of protocols, or by the legacy way of setting the internal context's "method".
```
exarkun: "There are two ways to configure it: either by ... or by ..." (no commas).  Or "There are two ways to configure it:: <rst bulleted list>".

docs/core/howto/ssl.rst
```diff
@@ -204,10 +199,40 @@ For example,
...
+By default, Twisted will configure it to use TLSv1.0 or later and disable the insecure SSLv3 protocol, but this manual control over protocols can be helpful if you need to support legacy SSLv3 systems, or you wish to restrict it down to just the strongest of the TLS versions.
```
exarkun: Commas in this sentence can go.

docs/core/howto/ssl.rst
```diff
@@ -204,10 +199,40 @@ For example,
...
+
+.. code-block:: python
+
+    from twisted.internet.ssl import (CertificateOptions, TLSVersion,
+                                      TLSVersionRange)
+    options = CertificateOptions(
+        ...,
+        tlsProtocols=TLSVersionRange(TLSVersion.TLSv1_0,
+                                     TLSVersion.TLSv1_2))
+
```
exarkun, line 213: Wow, what a bizarre API.  I don't see what this extra class brings.  The obvious API would be `tlsProtocols={TLSv1_0, TLSv1_1, TLSv1_2}` with a helper API for generating a whole range - `tlsProtocols=tlsVersionRange(TLSv1_0, TLSv1_2)` (just returning a set).

If the docs explained the `Highest` feature, this would look a little bit less weird.  I'd say, at a minimum, the docs must explain that feature and encourage its use.  In fact, they should start with an example that shows what most people should do most of the time and then dive into obscure cases like the two examples given here.

But I'd still probably opt for a simpler set-based API.  Twisted can expose a `GOOD_TLS_VERSIONS` constant that does the same job as `Highest` does now - without the weird `TLSVersionRange` thing.  Failing that, `TLSVersionRange` should implement some formal interface instead of being a simple concrete type with some hard-coded logic.  This at least brings some functionality with the extra complexity (allowing implementation of dynamic protocol selection logic) which might justify it.

docs/core/howto/ssl.rst
```diff
@@ -204,10 +199,40 @@ For example,
...
+    from twisted.internet.ssl import (CertificateOptions, TLSVersion,
+                                      TLSVersionRange)
+    options = CertificateOptions(
+        ...,
+        tlsProtocols=TLSVersionRange(TLSVersion.SSLv3,
+                                     TLSVersion.TLSv1_2))
+
+
+Newer and future OpenSSL versions may completely remove the ability to negotiate the insecure SSLv3 protocol, in which case, this will not allow you to re-enable it.
```
exarkun: Commas here should go.

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,109 @@
 from twisted.python import log
 from twisted.python._oldstyle import _oldStyle
 
+# If we don't have TLS v1.3 yet, we can't disable it -- this is just so when it
+# makes it into OpenSSL, connections end up going to v1.3
+if not hasattr(SSL, "OP_NO_TLSv1_3"):
```
exarkun: Don't ever use `hasattr`.  It masks exceptions.  Use `getattr(o, n, default) is not default`.

In this case, you can just do `_OP_NO_TLSv1_3 = getattr(SSL, "OP_NO_TLSv1_3", 0x00)` though.

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,109 @@
...
+        @param fromVersion: The minimal version to support.
+        @type fromVersion: L{TLSVersion} constant
+
+        @param toVersion: The maximal version to support (where Highest is the
+            newest at all times)
+        @type toVersion: L{TLSVersion} constant
+        """
+        versions = list(TLSVersion.iterconstants())
+
+        if versions.index(fromVersion) > versions.index(toVersion):
```
exarkun: NamedConstants compare for order.  You can just do `fromVersion > toVersion` to express this idea.

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,109 @@
...
+                # Otherwise fall back to 1.2.
+                tlsProtocols = TLSVersionRange(TLSVersion.TLSv1_2,
+                                               TLSVersion.Highest)
+        else:
+            tlsProtocols = TLSVersionRange(tlsProtocols,
+                                           tlsProtocols)
+
+    excludedVersions = []
+
+    for x in versions[0:versions.index(tlsProtocols._fromVersion)]:
```
exarkun: `0` is the default.

src/twisted/internet/_sslverify.py
```diff
@@ -1490,6 +1598,11 @@ def __init__(self,
             earlier in the list are preferred over those later in the list.
         @type acceptableProtocols: L{list} of L{bytes}
 
+        @param tlsProtocols: The TLS protocols this peer wishes to speak.
+            Either a single protocol from L{TLSVersion} or a L{TLSVersionRange}
+            with a lowest and highest L{TLSVersion}.
+        @type tlsProtocols: L{TLSVersion} constant or L{TLSVersionRange}
```
exarkun: Eww?  Don't do this, please.  If people want to be able to specify a single version, replace `TLSVersionRange` with something that doesn't make doing that so complicated that you're tempted to have a parameter that is either a scalar or a vector.

src/twisted/internet/_sslverify.py
```diff
@@ -1400,8 +1507,9 @@ def __init__(self,
 
         @param method: The SSL protocol to use, one of SSLv23_METHOD,
             SSLv2_METHOD, SSLv3_METHOD, TLSv1_METHOD (or any other method
-            constants provided by pyOpenSSL).  By default, a setting will be
-            used which allows TLSv1.0, TLSv1.1, and TLSv1.2.
+            constants provided by pyOpenSSL). By default, a setting will be
+            used which allows TLSv1.0, TLSv1.1, and TLSv1.2. Can not be used
+            with C{tlsProtocols}.
```
exarkun: Maybe say that `tlsProtocols` is preferred, too.

src/twisted/topfiles/6800.feature
```diff
@@ -0,0 +1 @@
+twisted.internet.ssl.CertificateOptions now takes a tlsProtocols parameter, which allows you to explicitly pick the versions of TLS or SSL you wish to negotiate. For example, you can now specify TLSv1.2 and TLSv1.0 but not TLSv1.1.
```
exarkun: I don't actually see how to use the API to implement the example given here.  What did I miss?

exarkun (commented):
Thanks.  Some comments inline.

src/twisted/topfiles/6800.feature
hawkowl: Ugh okay I need to update that. (see follow up comments)

docs/core/howto/ssl.rst
hawkowl: Originally, I used a set based API, but @Lukasa discouraged that -- one of the reasons that swayed me was that it binds us to OpenSSL (Windows' SChannel and possibly others accept a range in their API, vs particular protocols). I am honestly not sure what's better.

An alternate API would be possibly a `minimumTLSVersion` and a `maximumTLSVersion` with defaults of TLSv1.0 and a 'highest' canary respectively, and replace the concrete type and maybe make the API better? Then you can set just `minimumTLSVersion` if you want like, 1.2+, and not accidentally lock yourself out of future versions...

hawkowl: (the min/max version being on `CertificateOptions`, that is)

src/twisted/internet/_sslverify.py
hawkowl: Ah yeah, that's much cleaner.

docs/core/howto/ssl.rst
Lukasa: To expand on this a bit: there is essentially zero utility in being able to exclude a TLS version in the middle of your supported range. For example, if you want to support TLSv1 and TLSv1.2, there is no good reason to exclude TLSv1.1: clients that support TLSv1.2 will still use it, and clients that don't support TLSv1.2 will still have the option of choosing TLSv1.1 if they want to (they won't for various historical reasons, but that option exists).

In that case, using a set-based approach is just a really substantial increase in redundancy as well as an opportunity to do stupid stuff. On *top* of that, it means that if you have an application that you wrote in 2016 that is still being used in 2024 when a hypothetical TLSv1.4 is used, the set of "included only" TLS versions would *exclude* the use of TLSv1.4. That's a really silly limitation, so any implementation must either be a blacklist of TLS versions or must include some way to encode the notion of "X version *or better*".

I'm not wedded to the range, and alternative implementations that meet those goals would be fine by me, but I think the set is a bad choice for those reasons.

exarkun: > On top of that, it means that if you have an application that you wrote in 2016 that is still being used in 2024 when a hypothetical TLSv1.4 is used, the set of "included only" TLS versions would exclude the use of TLSv1.4. 

Yes.  This is what the `GOOD_TLS_VERSIONS` constant I invented for Twisted to expose is for.  It will include TLSv1.4 in 2024 and applications that use it for `tlsProtocols` will inherit that change without any code changes.

Lukasa: Sure, but we don't expect people to use this API unless they don't like the default. I assume `GOOD_TLS_VERSIONS` would *be* the default (if not, why not?), so I'm worried about those who have a use-case that requires this API but who want an unbounded upper end on their TLS versions (e.g. spec-complaint HTTP/2 implementations).

exarkun: > there is essentially zero utility in being able to exclude a TLS version in the middle of your supported range.

I think the point of this feature is to allow users to express whatever version choice they want.  Saying "TLSv1.0 to newest TLS, but not TLSv1.1" might be an extreme edge case but the point of the new API is to support all the edge cases.  The _current_ API already does a reasonable job letting you deal with the common case of "newest support TLS version".  This change is about providing a mechanism for doing better than that in the edge cases.

Lukasa: > Saying "TLSv1.0 to newest TLS, but not TLSv1.1" might be an extreme edge case but the point of the new API is to support all the edge cases.

Is it? If it is, then you're right, this API doesn't support that. But I wasn't aware that that was the design goal.

Now that I am aware, let me say that I think that's a *bad* design goal. I don't see any reason to allow the user to say "I want only SSLv2 and TLSv1.1, please". I am completely unable to come up with a scenario where that request is necessary, but "I want everything between SSLv2 and TLSv1.1 inclusive" is not a suitable replacement.

If it were *really*, *really* necessary, is there any reason that simply constructing two separate endpoints using the same protocol factory wouldn't suffice? I am entirely at a loss for the motivation for supporting all possible combinations of TLS versions.

If the only acceptable outcome for this feature is "allow all combinations of TLS versions" then I'm actually -1 on the feature until someone convinces me that it's necessary.

exarkun: > Is it? If it is, then you're right, this API doesn't support that. But I wasn't aware that that was the design goal.

I'm going to say it is.  I think I get to do that because I filed #6800 ;)  (That's not to say I think you can't object to that goal.)

>  I don't see any reason to allow the user to say "I want only SSLv2 and TLSv1.1, please".

This is a straw-man, I think.  SSLv2 is almost entirely unused now.  And it's pretty radically different from any of the TLS versions, anyway.  I think the "TLSv1.0 or TLSv1.2 but not TLSv1.1" (or, after time passes, any other "recent X, recent Z, but not recent Y between them") could be useful.  Perhaps TLSvZ has the best known security properties but is not yet widely supported, TLSvX has pretty good security properties, and an attack is discovered against TLSvY.  That seems like a legitimate reason to want this feature, to me.

> If the only acceptable outcome for this feature is "allow all combinations of TLS versions" then I'm actually -1 on the feature until someone convinces me that it's necessary.

I'm open to the idea that not _all_ combinations need to be supported but I don't see a simple API that allows the combinations that seem to make sense without also allowing everything else.

Lukasa: > Perhaps TLSvZ has the best known security properties but is not yet widely supported, TLSvX has pretty good security properties, and an attack is discovered against TLSvY.

The reason I've discounted such a use-case is that not only has it so-far never happened, but the odds of an attack against TLSvY not *also* affecting TLSvX seem really unlikely to me. I have to admit that that's a matter of "gut feel" rather than anything else, so I'm willing for someone whose gut is more trustworthy to just tell me I'm wrong here.

docs/core/howto/ssl.rst
glyph: My understanding is that these turds are unnecessary artifacts of the Lore migration that provide no utility, so when editing docs we remove them.  Irrelevant to this specific change, of course.

docs/core/howto/ssl.rst
glyph: I'm confused about this feedback.  What's wrong with the commas?  One of them, at least, "By default, Twisted", seems to serve the purpose of clarifying the default behavior.  "By default Twisted" might mean "With the default (version of) Twisted"; this is of course made clear later in the sentence but requiring the reader to move forward in order to understand the context just makes the sentence harder to read.

Similarly, it is common practice to separate independent clauses with a comma and a conjunction.  See rule 3b here: http://www.grammarbook.com/punctuation/commas.asp

Rule 3c on that page implies implies that the comma _is_ necessary in this case, since the independent clause after "but" has its own subject.

If I were going to have feedback on this sentence it would be "too many independent clauses in one sentence" :).

exarkun: Sorry, I mis-clicked apparently.  This feedback was meant for the previous line.

exarkun: Ah no, that feedback is somewhere else.  Sorry, I don't know.  I'll say I'm out of practice with github reviews.

glyph: Oh. On the previous line one of the commas looks fine to me, but the other should be a colon :).

docs/core/howto/ssl.rst
mithrandi: FWIW there is a server my application connects to that supports TLS 1.0 and TLS 1.2 but apparently not TLS 1.1 (the remote end stops responding partway through the handshake when trying TLS 1.1; it's not my server so it's hard to diagnose further). I don't really need this feature because hardcoding TLS 1.2 (or TLS 1.2-and-higher) is good enough, but I thought I'd mention it.

glyph: It sounds to me like we should explore the potential use-cases a bit here.

## 0. Don't Mess With Anything

Worth repeating that this is what _almost everyone_ should do; the customization API here is for edge cases.

## 1. Decrease minimum TLS version to be lower than that supported by Twisted by default in order to support older clients.

If you run a non-security-critical service and you want to consciously opt in to a bad, old version of TLS that isn't _so_ bad it's disabled yet, but has widespread deployment in your expected client base, 

## 2. Increase minimum TLS version to be higher than that required by Twisted by default in order to be more secure when you control both ends of the wire.

Twisted makes certain assumptions about your client base; if your clients are known, limiting to the most recent / best version of TLS is better than accepting potentially old/bad versions.

## 3. Limit maximum TLS version for security testing purposes (to emulate older clients).

If one is writing a testing tool like https://www.ssllabs.com/ssltest/, one needs a way to clamp the maximum TLS version that we will negotiate, mainly in order to ensure that our peer isn't doing something dumb.  This is _extremely esoteric_ and ideally would be hard to do by accident.

## 4. Skip an intermediate TLS version in order to hot-fix a future vulnerability

This is the one that @Lukasa's gut is telling us doesn't exist.

Keep in mind that this may be an _implementation_ vulnerability and not just a spec vulnerability though, which seems more plausible to me.  Especially if we were to (say) rapidly cycle from TLS 1.3 1.4 1.5, and 1.4 doesn't see a lot of deployment, it does make sense to me that there might be a "surprise" vulnerability in OpenSSL in the implementation of 1.4.

However, even if this case were to exist, the "right" fix is to get security updates for your TLS library or for Twisted, not to hack around it in your app.  So this strikes me as interesting mainly for completeness rather than "real" usage; the utility in doing it now would be to make it easy to do a fast release of Twisted with the relevant version skipped in the future, rather than exposing it as a do-it-yourself security fix for end users.

glyph: Enumerating all the use-cases like that suggests a totally different API, given how its behavior should change as code ages.

For example, let's say you're using the API for reason 2; you want to be *more* secure by the default.  If we have an API that expresses that specifically, and you were to do `CertificateOptions(increaseMinimumVersionTo=TLS_v1_3)`, in a few years when TLS 1.3 has been phased out in favor of something new and we've decided to EOL it in Twisted, that option would be a no-op.  However, if the API were a simple enumeration, and you did simply `CertificateOptions(tlsProtocols=set([TLSVersion.TLSv1_1, TLSVersion.TLSv1_2, TLSVersion.TLSv1_3]))` then when TLS 1.5 comes out, if you haven't been maintaining this software, you end up with a _less_ secure configuration, which is the opposite of what you wanted.

glyph: @mithrandi - Thanks for the data-point.  Real-world data is priceless when discussing issues like this.

Lukasa: @glyph I'd suggest the increaseMinimum/decreaseMaximum style API works fine for me (in the sense that it's basically isomorphic to the range API, with fewer types).

@mithrandi So here is the thing: the effect on a *client* of saying "TLS1 and 1.2, not 1.1" is *extremely* minimal. Clients don't send a list of supported versions: they just send the highest they support. In OpenSSL the only effect will be to remove any cipher suites that exist in TLS1.1 and not either of the other protocols. I don't think I'm worried about that use case. 😉

glyph: @Lukasa So if I understand you properly, you're saying that my use-case number 4 is actually effectively _impossible_ to support on clients, since the protocol doesn't have an affordance for that?  So a set-based API in the client case is actually unsupportable?

markrwilliams: I second @Lukasa in that a TLS vulnerability in a newer version that isn't in an older one is unlikely, and will add that I've had great difficulty convincing implementations (specifically Bouncy Castle) to honor a higher-than-default lower bound.  I would like an API that makes the upper and lower bounds explicit.  I'm in favor of a min/max-only API.

glyph: I'm inclined to say that we should just give `CertificateOptions` two arguments; `requireAtLeastVersion=` and, say, `artificiallyLowerProtocolMaximum=`, to mirror their functional purpose.  If we just use symmetric `minimumVersion`/`maximumVersion` argument names, it makes it sound like the range is just arbitrary, but in fact opinionated lower levels and implementation restrictions might decide to ignore our instructions anyway since these have specific security implications; the minimum and maximum can then have different documentation that explains the situation in which each might be ignored.

Lukasa: > So if I understand you properly, you're saying that my use-case number 4 is actually effectively impossible to support on clients, since the protocol doesn't have an affordance for that?

Not *impossible* per se, just that the effect it has is not what you'd expect. Suppose you configure a client to support TLSv1 and TLSv1.2 only. The dialog is like this:

1. Client sends Client Hello, and the version field in there is set to the highest supported version (TLSv1.2).
2. The server sends back a Server Hello, selecting the highest version both it and the client support. The server only supports up to TLSv1.1, so it sends back TLSv1.1.

Now, the effect of "skip intermediate versions" is that this handshake will fail, because the server chose TLSv1.1 which the client doesn't want to support. This is *in spite* of the fact that both the client and server may have a supported common protocol in TLSv1! The client will just say "welp didn't support that protocol version, raise an error". That's a pretty stupid failure mode when in fact the two endpoints *can* communicate.

As to your proposed argument names, that's fine by me.

src/twisted/internet/_sslverify.py
```diff
@@ -1525,11 +1604,27 @@ def __init__(self,
...
-            # will be supported.
             self.method = SSL.SSLv23_METHOD
-            self._options |= SSL.OP_NO_SSLv3
+
+            if minimumTLSVersion is None:
+                minimumTLSVersion = TLSVersion.TLSv1_0
+
+            if maximumTLSVersion and minimumTLSVersion > maximumTLSVersion:
+                raise ValueError(
+                    "minimumTLSVersion needs to be before maximumTLSVersion")
```
Lukasa: This reads a bit weirdly to me: maybe "maximumTLSVersion must exceed minimumTLSVersion"?

Lukasa: Also this check can probably be brought into `_getExcludedTLSProtocols`.

src/twisted/internet/_sslverify.py
```diff
@@ -1525,11 +1604,27 @@ def __init__(self,
         self._mode = SSL.MODE_RELEASE_BUFFERS
 
         if method is None:
-            # If no method is specified set things up so that TLSv1.0 and newer
-            # will be supported.
             self.method = SSL.SSLv23_METHOD
-            self._options |= SSL.OP_NO_SSLv3
+
+            if minimumTLSVersion is None:
+                minimumTLSVersion = TLSVersion.TLSv1_0
```
Lukasa: This value should probably take account of `maximumTLSVersion`: if you have set `maximumTLSVersion=SSLv3` (because you hate yourself) then this default value will fail, even though we have enough information to DTRT.

src/twisted/test/test_sslverify.py
```diff
@@ -840,6 +840,214 @@ def test_singleUseKeys(self):
...
+        with self.assertRaises(TypeError) as e:
+            sslverify.OpenSSLCertificateOptions(
+                privateKey=self.sKey,
+                certificate=self.sCert,
+                method=SSL.SSLv23_METHOD,
+                maximumTLSVersion=sslverify.TLSVersion.TLSv1_2,
+            )
+
+        # Best error message
+        self.assertEqual(e.exception.args, ("nope",))
```
Lukasa, line 952: Really?

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,65 @@
...
+
+    @return: The versions we want to disable.
+    @rtype: L{list} of L{TLSVersion} constants.
+    """
+    versions = list(TLSVersion.iterconstants())
+
+    excludedVersions = []
+
+    for x in versions[0:versions.index(oldest)]:
+        excludedVersions.append(x)
```
Lukasa: This is just a list comprehension:

```python
excludedVersions = [version for version in versions[0:versions.index(0)]]`
```

src/twisted/internet/_sslverify.py
```diff
@@ -21,6 +22,65 @@
...
+
+    if newest:
+        for x in versions[versions.index(newest):]:
+            if x is not newest:
+                excludedVersions.append(x)
```
Lukasa: This can also be written as a comprehension, but it matters less.

Lukasa (dismissed):
So I'm in general happy with this implementation, though the API discussion is ongoing.

src/twisted/test/test_sslverify.py
hawkowl: yes that's the actual exception argument

src/twisted/internet/_sslverify.py
hawkowl: done

src/twisted/internet/_sslverify.py
hawkowl: done

src/twisted/internet/_sslverify.py
hawkowl: done

dpnova:
Looks great, thanks @hawkowl 

Also sorry I never got to this :/ Really appreciated !

glyph:
(not a review, but: note pyflakes & txchecker fails)

glyph:
I still really feel like `minimumTLSVersion` and `maximumTLSVersion` are insufficiently expressive.

As I mentioned [here](https://github.com/twisted/twisted/pull/603#discussion_r89170067) there are really 4 use-cases here and the two arguments implemented here do not satisfy the use-case 2 that I proposed.

Let me expand on it slightly, because this is an actual thing that has happened, in the wild, more than once:

Let's say we're maintaining a Twisted-using application, Blub.  Blub knows that its typical client base is more modern than Twisted's relatively conservative guess, so it sets a more aggressive minimumTLSVersion of 1.2.

Years pass.  Blub, being relatively feature-complete, is reduced to a skeleton crew of maintainers, but still has a large (and relatively happy) user community.  However, since years have passed, there are now game-over exploits in TLS versions up to 1.5, which is now Twisted's minimum.

Again: this has really happened, more than once, back when Twisted's minimum version was SSL2 and the "aggressive" version was `SSLv3_METHOD`.  That, of course, is even _worse_ because it prohibits upgrading, but even being allowed to negotiate SSLv3 would be a disaster nowadays.

What I want is a `increaseMinimumVersion=TLSVersion` argument that _increases_ the lower bound, and when used against future Twisted versions will simply be ignored (perhaps emitting a warning), and a `decreaseMinimumVersion=TLSVersion` which _decreases_ the lower bound, allowing for negotiation of old versions, in explicit awareness of their potential insecurity.  The former could be passed by anyone that wants to just bump their security a bit, without worrying that it'll be a time bomb for future versions of their software; the latter is explicitly a dangerous security decrease that the implementor would be cautioned to keep an eye on.

I know that @Lukasa and @hawkowl have explicitly rejected this argument, but I'd like someone with more explicit crypto-smarts, like @alex, @lvh, or @reaperhulk to also tell me I'm dumb.

glyph:
(Also I want some kind of molly-guard around `maximumTLSVersion` because woo buddy is it a bad idea to set that except in _very_ specific circumstances...)

hawkowl:
All builds are green. I have a `noLessThanTLSVersion` and a `atLeastTLSVersion`, as well as a `reduceSecurityToTLSVersion`. Review time!

src/twisted/topfiles/6800.feature
```diff
@@ -0,0 +1 @@
+twisted.internet.ssl.CertificateOptions now takes minimumTLSVersion and maximumTLSVersion parameters, allowing you to bracket your TLS support or easily set a minimum with unbounded maximum, unlike the now-deprecated method passing API.
```
glyph: This is no longer accurate :)

src/twisted/internet/_sslverify.py
```diff
@@ -1481,11 +1580,46 @@ def __init__(self,
...
+            excludedVersions = _getExcludedTLSProtocols(
+                noLowerThanTLSVersion, reduceSecurityToTLSVersion)
+
+            for version in excludedVersions:
+                self._options |= _tlsDisableFlags[version]
         else:
+            warnings.warn(
+                ("Passing method to twisted.internet.ssl.CertificateOptions "
+                 "was deprecated in Twisted NEXT. Please use a combination "
+                   "of noLowerThanTLSVersion, atLeastTLSVersion, and "
```
glyph: This indentation is slightly weird.

src/twisted/internet/_sslverify.py
```diff
@@ -1481,11 +1580,46 @@ def __init__(self,
...
+                   "reduceSecurityToTLSVersion instead."),
```
glyph: I think we might want to caution users that they just should stop passing the method entirely and that Twisted will do a good thing by default.  As much as possible, users should accept the defaults :)

src/twisted/internet/_sslverify.py
```diff
@@ -1440,12 +1515,32 @@ def __init__(self,
...
+            value. You probably want to use this over C{noLowerThanTLSVersion}.
+        @type atLeastTLSVersion: L{TLSVersion} constant
+
+        @param noLowerThanTLSVersion: The minimum TLS version to use. If not
+            specified, it is a generally considered safe default (TLSv1.0). If
+            you want to raise your minimum TLS version to above that of this
+            default, use C{atLeastTLSVersion}.
+        @type noLowerThanTLSVersion: L{TLSVersion} constant
+
+        @param reduceSecurityToTLSVersion: The maximum TLS version to use. If
```
glyph: This might warrant some emphasis or some ALL CAPS or something - don't touch unless you know what you're doing, "reduce security" is not a joke, etc.

src/twisted/internet/_sslverify.py
```diff
@@ -1456,8 +1551,12 @@ def __init__(self,
             ciphers for the current platform.
 
         @raise TypeError: if C{trustRoot} is passed in combination with
-            C{caCert}, C{verify}, or C{requireCertificate}.  Please prefer
+            C{caCert}, C{verify}, or C{requireCertificate}. Please prefer
```
glyph: [2 spaces after full stop](http://www.heracliteanriver.com/?p=324), please.  (This was correct before.)

glyph: https://github.com/glyph/python-docstring-mode will format this correctly automatically, btw.

glyph (commented):
OK. For the most part, very happy with this.  But, upon further consideration of the names "`noLowerThanTLSVersion`" and "`atLeastTLSVersion`" I think they sound too close to synonyms (and, literally, they are).

As I was running through some options I realized that what they are trying to express is that one is _increasing_ the minimum allowable TLS version and one is _decreasing_ it.  So, for symmetry, I think

- `decreaseMinimumTLSVersionTo=`
- `increaseMinimumTLSVersionTo=`
- `decreaseMaximumTLSVersionTo=`

That gets rid of the word "security" but it have "decrease" and "maximum" so I feel like it gets the point across reasonably well.

This is just a search-and-replace though, and I'm otherwise happy with it.  So please edit (if you have a different idea for the names, we can chat) and land.

PR #606, opened 2016-11-26
[#8920] Remove deprecated stuff in twisted.mail.tap

hawkowl:
https://tm.tl/#8920


glyph (approved):
This looks great - deleting dummy crypto keys, removing 10x as much code as is added (and the tiny cleanup you sneaked in is tiny, but I will note that it is the cause of the conflicts...).  Please just resolve the conflicts and merge.

PR #607, opened 2016-11-26
Fix 8919 -- Procmon Python 3.5 bytes/string issues

moshez:
Author: moshez

Reviewer: wsanchez

Fixes: #8919 

On Python 3, procmon now handles process output without exceptions


moshez:
Still a work in progress.

wsanchez (approved):

PR #609, opened 2016-11-26
[#8925] Remove twisted.protocols.telnet

hawkowl:
https://tm.tl/#8925


derwolfe (approved):
1. This was deprecated as of version 2.1 which satisfies the deprecation policy's [timeline](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#incompatible-changes)
2. This implementation of `telnet` constitutes a maintenance burden as it duplicates the `telnet` implementation at [t.c.telnet](https://github.com/twisted/twisted/blob/160ecab6505e6634a8ea5c33fa209f90ffd5e52d/src/twisted/conch/telnet.py).

This looks ready to land to me.

Thanks :)

PR #611, opened 2016-11-29
[#8922] make HostnameEndpoint use new async multi-resolver from reactor

glyph:
Author: glyph

Reviewer: <names>

Fixes: #8922

twisted.internet.endpoints.HostnameEndpoint now uses the passed reactor's
implementation of twisted.internet.interfaces.IReactorPluggableResolver to
resolve hostnames rather than its own deferToThread/getaddrinfo wrapper; this
makes its hostname resolution pluggable via a public API.


glyph:
Test failures are because `inet_pton` doesn't exist on Windows, and our implementation in `twisted.python` apparently behaves oddly

glyph:
More detail: the issue is that _our_ `inet_pton` considers the empty string to be a valid IPv6 address, which diverges from pretty much every platform's behavior.

This causes the call to `abstract.isIPv6Address` in `tcp.Port.__init__` to return `True`, which in turn causes it to think its address family is IPv6 which makes it think it needs to call `getaddrinfo` to get a valid scope ID for its address.  This fails because `""` is not actually a valid IPv6 address.

glyph:
Although [supposedly `inet_pton` exists on 3.4+](https://docs.python.org/3.5/library/socket.html#socket.inet_pton).

glyph:
And indeed, on 3.4+ only one test in `twisted.internet.abstract` is failing.

glyph:
All the relevant builds have actually succeeded, but due to [this incident](https://www.traviscistatus.com/incidents/nq0sf0srvx8s) we are stuck in a rather long queue for the OS X builds to finish.

src/twisted/internet/_idna.py
```diff
@@ -0,0 +1,54 @@
...
+    not available; C{service_identity} is somewhat stricter about this.
+
+    @param text: A domain name, hopefully.
+    @type text: L{unicode}
+
+    @return: The domain name's IDNA representation, encoded as bytes.
+    @rtype: L{bytes}
+    """
+    try:
+        import idna
```
dstufft, line 26: Does it make sense to instead do:

```python
try:
    from idna import encode as idna_encode, decode as idna_decode
except ImportError:
    def idna_encode(str):
        return str.encode("idna")
    def idna_decode(octets):
        return octects.decode("idna")
```

At the module level, and then just do:

```python
def _idnaBytes(text):
    return idna_encode(text)

def _idnaText(octects):
    return idna_decode(octects)
```

That way the import conditional stuff happens once at module scope instead of each time these functions are called? Particularly since IIRC missing imports are not cached like successful imports are.

src/twisted/internet/endpoints.py
```diff
@@ -694,26 +717,43 @@ def connect(self, protocolFactory):
...
         wf = protocolFactory
-        d = self._nameResolution(self._host, self._port)
+        d = Deferred()
+        if self._badHostname:
+            ude = ValueError("invalid hostname: {}".format(self._hostText))
+            d.errback(ude)
+            return d
+        addresses = []
+        @provider(IResolutionReceiver)
+        class MyReceiver(object):
```
dstufft: ``MyReceiver`` is a pretty crummy name, maybe something like ``EndpointReceiver``?

dstufft (commented):
I didn't give this the most indepth review ever, but on the surface it looks sane to me. I made a few comments not as a "these are wrong" but as a food for thought.

src/twisted/internet/_idna.py
glyph: That's a good point - but I'd rather do it in a separate PR, since these lines are moved unmodified from another module in this one.

src/twisted/internet/abstract.py
```diff
@@ -481,30 +481,49 @@ def fileno(self):
...
+    Determine whether the given string represents an IP address of the given
+    family; by default, an IPv4 address.
 
     @type addr: C{str}
     @param addr: A string which may or may not be the decimal dotted
-    representation of an IPv4 address.
+        representation of an IPv4 address.
+
+    @param family: The address family to test for.  (This parameter has only
+        been available since Twisted 16.7; previosuly L{isIPAddress} could only
```
derwolfe: previosuly/previously

src/twisted/internet/abstract.py
```diff
@@ -481,30 +481,49 @@ def fileno(self):
...
+        test for IPv4 addresses.)
+    @type family: C{bool}
```
derwolfe: It seems like `bool` isn't the correct type for the address family parameter. Would `C{int}` or something more descriptive like what is found in [twisted/tcp.py](https://github.com/twisted/twisted/blob/e38cc25a67747899c6984d6ebaa8d3d134799415/src/twisted/internet/tcp.py#L329-L332) be more appropriate?

derwolfe (commented):
Sorry - I didn't have time for a full review, but I'll come back later (assuming this is still open:) )

src/twisted/internet/endpoints.py
glyph: Yep, will fix.

src/twisted/internet/abstract.py
glyph: Thanks for the catch. Will fix.

src/twisted/internet/abstract.py
glyph: `L{int}` would indeed be correct. Will fix.

src/twisted/internet/endpoints.py
```diff
@@ -694,26 +717,43 @@ def connect(self, protocolFactory):
...
+            @staticmethod
+            def resolutionBegan(resolutionInProgress):
+                pass
+            @staticmethod
+            def addressResolved(address):
+                addresses.append(address)
+            @staticmethod
+            def resolutionComplete():
+                d.callback(addresses)
```
hawkowl, line 777: I am not sure why TwistedChecker isn't complaining about the lack of newlines here :(

src/twisted/internet/endpoints.py
```diff
@@ -694,26 +717,43 @@ def connect(self, protocolFactory):
...
+                    yield TCP6ClientEndpoint(
+                        self._reactor, eachAddress.host, eachAddress.port,
+                        self._timeout, self._bindAddress
+                    )
+                if isinstance(eachAddress, IPv4Address):
+                    yield TCP4ClientEndpoint(
+                        self._reactor, eachAddress.host, eachAddress.port,
+                        self._timeout, self._bindAddress
+                    )
                     # Yields an endpoint for every address returned by GAI
```
hawkowl: this comment seems... misaligned now

src/twisted/internet/endpoints.py
```diff
@@ -694,26 +717,43 @@ def connect(self, protocolFactory):
         connection which is established first.
         """
         wf = protocolFactory
-        d = self._nameResolution(self._host, self._port)
+        d = Deferred()
+        if self._badHostname:
+            ude = ValueError("invalid hostname: {}".format(self._hostText))
```
hawkowl: this does not seem like a good variable name (can it just be done inline?)

src/twisted/conch/test/test_forwarding.py
```diff
@@ -65,17 +40,13 @@ def test_channelOpenHostnameRequests(self):
         is resolved using HostnameEndpoint's resolver.
         """
         sut = forwarding.SSHConnectForwardingChannel(
-            hostport=('fwd.example.org', 1234))
+            hostport=(b'fwd.example.org', 1234))
```
hawkowl: Is this making it satisfy the current interface of `SSHConnectForwardingChannel`? Or will it not accept unicode now (on py3, if it's even ported)?

src/twisted/internet/endpoints.py
```diff
@@ -679,7 +677,32 @@ def __init__(self, reactor, host, port, timeout=30, bindAddress=None,
...
+            host = normalize('NFC', host)
+            self._hostText = host
+            if isIPAddress(host) or isIPv6Address(host):
+                self._hostBytes = host.encode("ascii")
+            else:
+                try:
+                    self._hostBytes = _idnaBytes(host)
+                except UnicodeError:
+                    self._badHostname = True
+                    self._hostText = (host.encode("ascii", "backslashreplace")
```
hawkowl: I have no idea what this code does. Comments maybe?

src/twisted/internet/endpoints.py
glyph: This is correct behavior on TwistedChecker's part.  3 lines between top-level suites, 2 lines between methods on top-level classes.  There are no required newlines between nested callbacks - and inner classes are just a kind of nested callback.  The VWS requirements are to make it easy to scan a file and see where the globally-visible methods are; within a function newlines would be confusing because they'd imply top-level stuff was being defined.

src/twisted/conch/test/test_forwarding.py
```diff
@@ -67,15 +42,11 @@ def test_channelOpenHostnameRequests(self):
...
             hostport=('fwd.example.org', 1234))
         # Patch channel and resolver to not touch the network.
-        sut._reactor = MemoryReactorClock()
-        self.patchHostnameEndpointResolver(
-            request=('fwd.example.org', 1234),
-            response=(AF_INET6 ,'::1'),
-            )
-
+        memoryReactor = MemoryReactorClock()
+        sut._reactor = deterministicResolvingReactor(memoryReactor, ['::1'])
```
exarkun, line 46: This test used to verify that the connect attempt was made to fwd.example.org.  It doesn't seem to verify that any longer.  I hacked the implementation to break it and the test still passes.

It seems like `deterministicResolvingReactor` actually encourages this kind of testing mistake.  It would be  better if it took something like `{"fwd.example.org": (AF_INET, "::1")}`.

glyph: I didn't realize I'd actually pooched the assertion, but I did realize this was sub-optimal.  See https://github.com/twisted/twisted/commit/f19151e59bb9b64306ce023d85f5621cdb4ad473 - once that's landed, I will adjust some of these tests.

glyph: To avoid losing track of this in the meanwhile, I've [recorded it in a ticket](https://twistedmatrix.com/trac/ticket/8939)

exarkun: Cool, thanks.

PR #612, opened 2016-11-30
Support OpenSSH keys with Windows line endings.

clokep:
OpenSSH keys with Windows (CRLF) were supported in Twisted <= 16.5.0, but broke in 16.6.0 (in #533, specifically).

See ticket https://twistedmatrix.com/trac/ticket/8928 for more information.

I've added a test case and fixed this for the one case I ran into, I'm unsure if there are other places this could happen! (Also, I haven't contributed to Twisted before and briefly read over [the contributing guide](https://twistedmatrix.com/trac/wiki/ContributingToTwistedLabs), but please let me know if there are other things I should do!)


markrwilliams:
@clokep Thanks for the PR!  The code considers lines to end with '\n' and not '\r\n' has been around for a while.  openssh (via openssl `crypto`, it appears) certainly parses private keys with lines that end in '\r\n' so this should get merged.

The patch and its test look great!  This needs a [topfile](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles) in `src/twisted/conch/topfiles`.  Please add one, then add the "review" keyword to the Trac ticket.  This will put it into our [review queue](https://twisted.reviews).

clokep:
Yeah it's been around a while, but f585c83fd6ddca9ea2d2683ad8b644347243d2e3 changed line 371 to use negative indexing, which then gets an off-by-one error with the leftover `\r` character. This was implicitly working before since it was indexed from the beginning of the line, see:

```diff
 lines = data.strip().split(b'\n')
-kind = lines[0][11:14].replace(b' ', b'')
+kind = lines[0][11:-17]
```

I've added a topfile and will request review. Thanks!

src/twisted/conch/topfiles/8928.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh.keys.Key.fromString now supports OpenSSL private keys with Windows line endings (\r\n) queries again (broken since 16.6.0).
```
markrwilliams: I suggest removing "queries".

markrwilliams (commented):
Thanks for the topfile :) One small change to the wording there, then I'll merge when the builds are green.

src/twisted/conch/topfiles/8928.bugfix
clokep: Oops. That was left-over from the example I copied from. Embarrassing. I've pushed a change.

markrwilliams:
@clokep You're too fast!  I'll merge when the builds go green.

clokep:
@markrwilliams Looks like the builds completed, one is in the "allowed failures" (so I'll ignore that one).

The other one is pylint failure on a long line in the tests. Is there a code style for how to reformat that while keeping it legible?

markrwilliams:
@clokep Twisted's old enough that its [coding style guide](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html) predates PEP 8.  You can always check your code by installing `twistedchecker` via `pip install twistedchecker` (Python 2 only for now) and running it:

```
twistedchecker /path/to/twisted/module.py
```

Note that this reports *all* violations in `module.py`, many of which will likely not be your fault.

The build failed because [a line in `t.c.test.test_keys`](https://github.com/twisted/twisted/pull/612/files#diff-bb38fcf5834bea31d05fc56f4c33bca5R384) is [85 characters long and the limit is 79](https://travis-ci.org/twisted/twisted/jobs/180410789#L269).  Wrapping that line at 79 characters should make the build pass.

One way to wrap that line is to rewrite the assert to look like this:

```python
        self.assertEqual(
            keys.Key.fromString(privateDSAData),
            keys.Key.fromString(privateDSAData.replace(b'\n', b'\r\n')),
        )
```

glyph:
Merged trunk and pushed to a twisted/ branch to update the buildbots.

PR #613, opened 2016-12-02
Revert "Add a PIDFile class that can be used as a context"

glyph:
Reverts twisted/twisted#572

Reopens #8659

This was merged with open twistedchecker errors and windows test failures, and caused trunk builds to start failing, as shown here:

https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1851/job/wjox40730busejof

Given that that trunk build actually failed in the middle of the tests added by this branch I am suspicious that it's relevant.  More importantly though the last commit on this pull request's branch history is a big red X.


glyph:
Given that the windows test failure is the issue, I'll use the handy fact that the PR causes appveyor to run before I hit the "merge" button. (Although given the massive backlog of mac builds, this will have to go to trunk with the big red "override" merge button.)

glyph:
It will be another 8 hours before build results are available, so I might have to go early with this one.  Not only has this caused the trunk / merge build to fail on Windows, it fails in such a way that [it takes the full build timeout time (1 hour per build)](https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1851) before it finally dies.

Given that we're already suffering from [the Travis OS X outage](https://www.traviscistatus.com/incidents/nq0sf0srvx8s), I think letting this go for longer while more Appveyor builds pile up behind it would be worse than just going ahead with the revert right now.  So: going ahead with the revert.

glyph:
The commit above shows a failing status because the branch didn't exist any more by the time that it got run.  However, the merge did cause trunk to go from red back to green, as per  https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1857

PR #615, opened 2016-12-02
Fix parsing of EC keys

rodrigc:
https://twistedmatrix.com/trac/ticket/8931


glyph:
Oops, looks like a pretty platform-independent test failure :)

rodrigc:
Needs more work

PR #617, opened 2016-12-05
[#7945] External platform support package

hawkowl:
https://tm.tl/#7945


src/twisted/internet/iocpreactor/iocpsupport.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+A shim to the underlying IOCP support code.
+
+See U{https://github.com/twisted/twisted-platform-support}.
+"""
+
+from twisted_platform_support._iocp import (
```
adiroiban: why are we importing private stuff from an external library?

if those attributes in twisted_platform_support are supposed to be used outside of twisted_platform_support, I think that they should be public.

If the whole twisted_platform_support is private to twisted, then I think that we should keep it in the main twisted repo and make it easy to create new versions and run tests.

with an external library, we would either need to duplicate a lot of tests... or release minor version each need we need to test some change

glyph:
@adiroiban commented on a diff which is now outdated, but the changes don't appear to have addressed the comment:

> why are we importing private stuff from an external library?
> 
> if those attributes in twisted_platform_support are supposed to be used outside of twisted_platform_support, I think that they should be public.
> 
> If the whole twisted_platform_support is private to twisted, then I think that we should keep it in the main twisted repo and make it easy to create new versions and run tests.
> 
> with an external library, we would either need to duplicate a lot of tests... or release minor version each need we need to test some change

Splitting stuff out _purely_ on the basis of being in C (or pyrex) is going to make certain kinds of maintenance tasks really annoying.  For example: this moves the IOCP support C code to a different package, but the IOCP python code itself remains in Twisted.

This creates practical problems like, for example, `pip install git+https://github.com/twisted/twisted@<some revision>`.  In order for a 3rd party project to try out current trunk, they'll need to find a paired set of revisions from the two repositories.  This will, similarly, create CI problems; we need to do a _release_ of the platform components before even submitting a PR to Twisted to update the code that depends on those components; such a release will then break the ability to `pip install` previous versions of Twisted if it's an incompatible change.

I'm at a bit of a loss for how to resolve this.  I wonder if @dstufft could shed any light; are there packaging tricks we could try?  Can a setuptools extra contain modules or packages itself, for example?  If we put two distributions in the same repository, can they be made to show up for a single `pip install` command?

dstufft:
@glyph No I don't believe so. If this is split out it should probably just be a fully fledged package with it's own public API and such.

glyph:
Thanks for the feedback, @dstufft .  I think on IRC we were coming to the consensus that this needs to be done slightly differently; in particular, "the IOCP reactor" may want to be its own distribution.

PR #620, opened 2016-12-05
Add a PIDFile class that can be used as a context, attempt #2.2

wsanchez:
Reverts twisted/twisted#613
…which is to say:  re-opens https://github.com/twisted/twisted/pull/572
…which is to say: https://twistedmatrix.com/trac/ticket/8659


glyph:
I just re-ran the OS X builders since the failures looked spurious and unrelated to me.  The travis failure is real, although it's only twistedchecker, so this just needs some style fixups.

wsanchez:
Right now I'm trying to isolate which tests are causing Appveyor to take forever.

I think the remaining `twistedchecker` complaints are bogus:
 * One complaint is use of a parameter `exc_type`, which is simply matching the stdlib signature.
 * There are some "missing docstring" complaints for methods in `PIDFile` that are documented by the `IPIDFile` interface.  I thought the interface docs are supposed to be sufficient?

glyph:
> I think the remaining twistedchecker complaints are bogus:

Nevertheless, the build is still failing, and it's a required status, so you can either submit a patch to twistedchecker or have temporary workarounds.

> One complaint is use of a parameter exc_type, which is simply matching the stdlib signature.

Since the runtime always passes these positionally you can just call it `excType`.  I normally end-run around this particular problem with `@contextlib.contextmanager` which tends to make the implementations shorter as well, but unfortunately I don't think there's a way to do that and make the type itself a context manager.

> There are some "missing docstring" complaints for methods in PIDFile that are documented by the IPIDFile interface. I thought the interface docs are supposed to be sufficient?

This is indeed a bug in twistedchecker: https://github.com/twisted/twistedchecker/issues/44.  Hopefully we can get it fixed at some point soon.  My workaround thus far has just been docstrings that all say `@see: L{IWhatever}`.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+        @return: The contained process ID.
+        @rtype: L{int}
+
+        @raise NoPIDFound: If this PID file does not exist.
+        @raise EnvironmentError: If this PID file cannot be read.
+        @raise ValueError: If this PID file's content is invalid.
+        """
+
+
+    def write(pid):
```
glyph: This interface seems a bit wide to me.  What's the use-case for writing some _other_ process's running PID?  Surely if you needed to do that for some reason you could construct a concrete `ForeignProcessPID` with the other process's PID, or parameterize the concrete `os.getpid` or somseuch.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+
+from twisted.logger import Logger
+
+
+
+class IPIDFile(Interface):
+    """
+    PID file.
+
+    Manages a file that remembers a process ID.
```
glyph, line 20: I kinda feel like this line by itself is sufficiently descriptive and `"PID file"` above is unnecessary.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+class PIDFile(object):
+    """
+    Concrete implementation of L{IPIDFile} based on C{IFilePath}.
+    """
+
+    _log = Logger()
+
+
+    @staticmethod
+    def format(pid):
```
glyph: We should avoid exporting any public API surface that we don't need; in this case, `PIDFile`'s job should just concretely implementing `IPIDFile` and should avoid exporting any other public methods.  This `format` implementation, for example, is tightly coupled to the `read` implementation, and substituting it, invoking it, or overriding it would not actually work without dealing with that dependency.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+        @rtype: L{bool}
+
+        @raise EnvironmentError: If this PID file cannot be read.
+        @raise InvalidPIDFileError: If this PID file's content is invalid.
+        @raise StalePIDFileError: If this PID file's content refers to a PID
+            for which there is no corresponding running process.
+        """
+
+
+    def __enter__():
```
glyph, line 68: Something to consider: if `__enter__` and `__exit__` are part of `IPIDFile`, then every implementor must provide their own implementation.  But the `__enter__` / `__exit__` implementation on `PIDFile` actually works for all implementations of `IPIDFile`.  Perhaps it should just be a top-level `@contextmanager`-decorated function that takes any `IPIDFile` provider so you can do `with usingPIDFile(MyPIDFile(...)):`?  This would eliminate a bunch of redundant method definitions.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+        except OSError as e:
+            if e.errno == errno.ENOENT:  # No such file
+                raise NoPIDFound("PID file does not exist")
+            raise
+
+        try:
+            return int(pidString)
+        except ValueError:
+            raise InvalidPIDFileError(
+                "non-integer PID value in PID file: {!r}".format(pidString)
```
glyph, line 138: Including the path name of the PID file in question would be super useful for diagnostics.

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,270 @@
...
+            )
+
+
+    def write(self, pid):
+        self.filePath.setContent(self.format(pid=pid))
```
glyph: This is a race-condition-prone implementation mechanism.

If two `twist` processes start up simultaneously, one should win the race, and one should lose; a pidfile is a sort of lock file.  With this strategy, it's look-before-you-leap with all the attendant problems that might cause.

The only really _right_ way to create a "file" with "contents" atomically is to use a symlink, which is what `twisted.python.lockfile.FilesystemLock` does.  However, if we want to preserve the existing contract of a pidfile being a literal file with bytes in it and not a symlink, we can use `FilePath.create` instead.

My concern with this is that I don't want a public `.write` method floating around with these _semantics_; I'm OK with the bug existing as long as a follow-up PR can fix it without disturbing the new public interface.

glyph: (`FilePath.create` uses `O_EXCL` and `O_CREAT` making it atomic.)

wsanchez:
I get that we have a code standard and we should comply with it, but these are both cases where the solution is to make the code jankier just to work around the checker tool's bugs.

In the past it was OK to ignore those warnings; though I heard you say at PyCon that doing so required as checker ticket, which seems fair (and which I didn't do for the `exc_type` case, because it seems unlikely to ever get fixed, but my bad.

Requiring compliance without exceptions for bugs seems like a step backwards.  I'll do it, but it's lame.

glyph (commented):
Thanks for doing this.  Not sure if this should technically be in review or not (oh no, my internet points!) but I had some thoughts anyway.

Most of these are non-blocking observations.  The major thing is the interface concern with atomicity; as it's specified right now, it presents a public interface which I don't think can be _made_ properly atomic.  I realize that these are all in private files though, so perhaps that concern is not blocking either!

However, given that, a radically simpler implementation might be possible: the only "public" API that the actual service runner touches is a context manager, not `writeRunningPID` or `remove` or anything like that; the only thing that seems to be calling these methods outside of tests is the `__enter__` method calling things on `self`.  So perhaps the interface should just be "it's a context manager" for now, no abstract `IPIDFile`, and then the pluggable implementation can just be a `FilePath`-based thing, or a no-op context manager.

glyph:
> I get that we have a code standard and we should comply with it, but these are both cases where the solution is to make the code jankier just to work around the checker tool's bugs.

The problem is that nobody (myself included!) is actually fixing the checker tool's bugs.  Leaving the checker tool in this state creates an unpleasant environment for newer contributors, who are even less well-versed in the nuances of what's OK to merge.  The checker was disabled for a while to allow for discretion in this sort of thing, and _thousands_ of new errors (most of them legit violations of the coding standard) got merged in.  This, in turn, makes it harder to work on Twisted because you always need to have the latest origin/trunk and run the actual (very slow) whole-codebase tox job instead of letting your editor run `twistedchecker` directly on whole files that (used to be) warnings-clean.  So things are a bit stricter now.

> In the past it was OK to ignore those warnings; though I heard you say at PyCon that doing so required as checker ticket, which seems fair (and which I didn't do for the exc_type case, because it seems unlikely to ever get fixed, but my bad.

The idea with that policy was to encourage folks to be aware of the bugs and gently encourage them to fix stuff without blocking.  But that idea was unfortunately a failure, so I am hoping this will motivate us to clean up some of these edge cases that are pain points instead of letting them linger indefinitely.

> Requiring compliance without exceptions for bugs seems like a step backwards. I'll do it, but it's lame.

One potential bugfix would be to add support for something like `# noqa:twistedchecker:<bug url>` like flake8 has https://flake8.pycqa.org/en/latest/user/ignoring-errors.html?highlight=noqa#in-line-ignoring-errors so that the workarounds can be simple turds that can be cleaned up instead of having to distort the shape of the code itself to work around tool bugs.  But, we'd still need someone to actually go do the work of adding that to twistedchecker :).

wsanchez:
Regarding the pseudo-review, this isn't in review because it was reviewed in #572 short of the CI issues.  That said, I'll read and address what I can now.  The point is this is to work towards a publicly exportable API, so that's all relevant, but maybe some would be better in separate tickets (haven't looked in any detail yet).

Will put it back to review afterwards, since it looks like I'll need to make some non-trivial though perhaps minor changes for Windows.

glyph:
> The point is this is to work towards a publicly exportable API, so that's all relevant, but maybe some would be better in separate tickets (haven't looked in any detail yet).

About 90% of the way through I realized that it probably didn't need to be blocking since this API doesn't leak into any public namespace I can see, but I did figure it would still be useful to inform the design.  So yeah, address as you can.

src/twisted/application/runner/_pidfile.py
wsanchez: 75043e2fb7e63c527a0233dd91c81242d7279481

src/twisted/application/runner/_pidfile.py
wsanchez: The idea here was that a PID file stores a PID, but isn't necessarily about storing a PID for the currently running process, though `writeRunningPID` is provided as a convenience for the most common expected use case.

But given that the context manager is tied to the current process, that seems thin logic in hindsight…

a0a4c9f1b18f26b3e176a743ca458c60c88c9398

src/twisted/application/runner/_pidfile.py
wsanchez: Right, if this is public, it should have a `parse()` buddy function.  Will make private.

20cef8e606bfb39365b87a5015f77debd639da30

src/twisted/application/runner/_pidfile.py
wsanchez: `IPIDFile` really only exists because of `NonePIDFile`  I think using `PIDFile` as a context manager shouldn't require wrapping it in `usingPIDFile`; that seems clunky.

Maybe a class decorator?

glyph: A class decorator a-la tubes's `@tube` would be fine.

(Although I am increasingly feeling like it should just be 2 separate functions, no classes at all; a no-op contextmanager and a `recordingPidInFile()` context manager.)

wsanchez:
OK, green!  Marking for review.

.travis/twistedchecker-trunk-diff.sh
```diff
@@ -8,37 +8,43 @@
 # * SCRIPT_NAME twisted/words/
 # * SCRIPT_NAME twisted.words
 
-target=$1
+set -e
```
thijstriemstra: what are these new options, can you use the verbose name of the option, or add a comment?

src/twisted/application/runner/_pidfile.py
```diff
@@ -0,0 +1,303 @@
+# -*- test-case-name: twisted.application.runner.test.test_pidfile -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+PID file.
+"""
```
thijstriemstra, line 7: I'm always for an `@since` marker, even when it's a private module.

tox.ini
```diff
@@ -78,6 +78,8 @@ setenv =
    # Help tests know where the base directory is.
    TOX_INI_DIR = {toxinidir}
 
+   txchecker-travis-required: TWISTEDCHECKER_SKIP_WARNINGS = W9208
```
thijstriemstra, line 82: a comment here describing this line would be appropriate.

.travis/twistedchecker-trunk-diff.sh
wsanchez: -e means exit with an error if any command exits with an error that isn't checked by the script.
-u means exit with an error if a variable that is not set is referenced.

These are good practices in Bourne shell scripts, and do not have verbose names, at least none that are portable.

thijstriemstra: it saves us a google search or 2 if that would be in a comment near that code.. I think anything non-python should be even more clearly documented.

src/twisted/application/runner/_pidfile.py
wsanchez: Hrm, I've never used that…

It doesn't seem appropriate for private API if the point is to let developers know how far back they can go and use the API.  `twisted.application.runner` is still undergoing a fair bit of incompatible churn (this is part of that churn), so I think we should hold off until the API is stable enough to make public.

.travis/twistedchecker-trunk-diff.sh
wsanchez: Comment added: 88533bd6c060634f761702cbf6057119e4ca0262

tox.ini
wsanchez: Comment added: 88533bd6c060634f761702cbf6057119e4ca0262

wsanchez:
@thijstriemstra:  Thanks for the feedback.

Was that a full review?  If so, please confirm and remove the review keyword from https://twistedmatrix.com/trac/ticket/8659 so you get credit for it.

thijstriemstra:
No it wasn't a full review @wsanchez, sorry ;) But it's looks good to me so far.

glyph:
Looks like the coverage issue is just with the am-I-root test in-line in a unit test: https://codecov.io/gh/twisted/twisted/compare/63dc0387ae49134f68305873a822a76a3367d897...1842680d2762fa018e97c6dc409e22e9ca703576#7372632F747769737465642F6170706C69636174696F6E2F72756E6E65722F746573742F746573745F70696466696C652E7079-283

The right way to address this - which is better style anyway, IMHO - is to just do the am-i-root test outside of the test case, in an expression: `test_something.skip = (geteuid() == 0 and "nope, im root")` or similar.

wsanchez:
@glyph:

Hrm.  I can't say I agree about that being a better style.  Raising `SkipTest` keeps the logic in the method and your suggestion means adding logic at class definition time, which seems a bit on the janky side.

In this case it's worse, because `geteuid` isn't importable on all platforms, so I would have to define a function that does this, and again the conditional in question won't get exercised, so that should fail coverage also.

Finally, the `geteuid() == 0 and "nope, im root"` trick is just an attempt to put the offending missing code on the same line as covered code, which hides the lack of coverage, which would be more explicitly and clearly accomplished with the more sinful looking but actually less problematic `#noqa` trick, in which you actually fess up to what you are doing.

I could just remove the check for root; it won't fail if you are root; it's just a bogus-but-passing test in that configuration, so I thought it better to call it out.

glyph:
> Hrm. I can't say I agree about that being a better style. Raising SkipTest keeps the logic in the method and your suggestion means adding logic at class definition time, which seems a bit on the janky side.

For _almost_ all cases, I agree; over the years I've been eliminating one case after another of things I thought might be a good idea at module-import/class-definition time, and this is one of the _very_ few that remain.

However, the reason it remains is a good one, I think: we should not be supporting an infinite number of configurations; we should have a few different 'skip' variables, consolidate them all in one place, and then set 'skip' to them.  For example, I've been trying to consolidate the test logic's "OpenSSL isn't installed" into one giant global variable for a while, and "I'm not root" should be a similarly central test, rather than each test one-offing its own way to check.

> In this case it's worse, because geteuid isn't importable on all platforms, so I would have to define a function that does this, and again the conditional in question won't get exercised, so that should fail coverage also.

Ultimately the problem here is that we don't have a coverage-submitting buildbot that runs as root. IIRC the tests fail if you run them as root already (although I can't seem to find the ticket).

As such, maybe a more practical suggestion is to literally delete this conditional, and allow it to fail when run as root, adding the conditional back in when we support "running as root" as a platform.

> Finally, the geteuid() == 0 and "nope, im root" trick is just an attempt to put the offending missing code on the same line as covered code, which hides the lack of coverage, which would be more explicitly and clearly accomplished with the more sinful looking but actually less problematic #noqa trick, in which you actually fess up to what you are doing.

Mea culpa on that one - it's just trading expression coverage for line coverage :).  But, our line in the sand _is_ drawn at line coverage and not expression coverage right now.

> I could just remove the check for root; it won't fail if you are root; it's just a bogus-but-passing test in that configuration, so I thought it better to call it out.

Oops. Didn't even see this line when I wrote the reply above.  Yeah, I think removing the check is the right thing for now, although "support running as root" is probably a pretty good idea so we can write tests for "drop privileges" and such.

wsanchez:
> However, the reason it remains is a good one, I think: we should not be supporting an infinite number of configurations; we should have a few different 'skip' variables, consolidate them all in one place, and then set 'skip' to them. For example, I've been trying to consolidate the test logic's "OpenSSL isn't installed" into one giant global variable for a while, and "I'm not root" should be a similarly central test, rather than each test one-offing its own way to check.

I agree that some test tooling like ways to check "I am root" are a good idea, though I'm not sure that's needed yet.  I totally don't agree that setting `.skip` is any better than raising `SkipTest`, though.

> As such, maybe a more practical suggestion is to literally delete this conditional, and allow it to fail when run as root, adding the conditional back in when we support "running as root" as a platform.

I'm not convinced; in this case, the condition is a guard for the possibility that we might support running as root in the future.  Tests that will break if you are running as root are fine; this is here to prevent a false positive should that come up later.  That doesn't really fix anything except that if you run all tests as root you might notice you don't have coverage for different reasons.

It's not news that I think that conforming to strict rules even in the face of bugs and incompleteness of the tools we use to check them over actual real correctness is a poor policy position, but I'll accept defeat again.  Hopefully, when someone deicides to add a root builder, someone will comb through all of the tests and notice false positives like this one, but I think a conditional like this if unlikely to *ever* get re-added if we decide to support running as root.

It does make me want to note that we should have a big warning somewhere that running Twisted software as root isn't tested/supported, if that's the position, which includes the "launch as root and drop privileges" case.  Which also means that all `twistd` options that require starting as root are actually unsupported.  But that's another episode.

> But, our line in the sand is drawn at line coverage and not expression coverage right now.

Our line in the sand is more accurately "whatever is implemented by `coverage`".  I wouldn't be surprised if it colored a line with an embedded and unexercised conditional as yellow, but I'll least that as a test for someone else.  In any case, I wouldn't encourage one to exploit a weakness in `coverage` as a way to pass our requirements; I'd encourage the opposite.

glyph:
> I totally don't agree that setting .skip is any better than raising SkipTest, though.

Another reason that it's better is that it's effectively declarative (albeit computed) rather than imperative.  This has implications for stuff like test parallelization and result reporting; if the runner can't see into the test implementation to figure out whether it's actually going to do anything or not, it's going to treat it as an equivalent unit of work.

wsanchez:
> Another reason that it's better is that it's effectively declarative (albeit computed) rather than imperative. This has implications for stuff like test parallelization and result reporting; if the runner can't see into the test implementation to figure out whether it's actually going to do anything or not, it's going to treat it as an equivalent unit of work.

Meh.  Tests aren't equivalent units of work; the only way to know cost (which is what I assume you are referring to) is to remember what they cost in the past or something.

Result reporting can certainly figure out that a skipped test is a skipped test at reporting time, regardless of how it was marked as skipped.

glyph:
> I'm not convinced; in this case, the condition is a guard for the possibility that we might support running as root in the future. Tests that will break if you are running as root are fine; this is here to prevent a false positive should that come up later.

In the _general_ case, "test things that might be interesting in a future testing configuration" is something we definitely shouldn't be writing tests for.

I'll concede that "running as root" is a thing that people do, that certain features require, so the lack of a testing configuration is definitely a problem with our testing infrastructure and not the system under test here.

> That doesn't really fix anything except that if you run all tests as root you might notice you don't have coverage for different reasons.

Our coverage threshold is already based on _combined_ coverage, so unless we remove all "not as root" builders we shouldn't have that issue.

> It's not news that I think that conforming to strict rules even in the face of bugs and incompleteness of the tools we use to check them over actual real correctness is a poor policy position, but I'll accept defeat again.

We should probably have that discussion somewhere isolated from one specific example.  It is amazing to me that almost every single thing you do ends up running into this wall head-first one way or another - other folks at least manage to avoid it _sometimes_ :).

> Hopefully, when someone deicides to add a root builder, someone will comb through all of the tests and notice false positives like this one, but I think a conditional like this if unlikely to ever get re-added if we decide to support running as root.

Wait - why is it unlikely to get re-added?  Wouldn't it _need_ to get re-added?

> It does make me want to note that we should have a big warning somewhere that running Twisted software as root isn't tested/supported, if that's the position, which includes the "launch as root and drop privileges" case. Which also means that all twistd options that require starting as root are actually unsupported. But that's another episode.

To be clear about the policy position: the policy position is that "supported" == "we have a builder that tests in that environment".  Hence the "supported" in the URL https://buildbot.twistedmatrix.com/boxes-supported

We've never had a supported builder that runs as root.  Don't get me wrong, we should - but we have lots of commonly-run but unsupported configurations.  (We don't even support FreeBSD right now, despite the fact that lots of people use it; we don't support PyPy yet, despite the fact that we actually run it in production.)  If we warned loudly about all of these the README would be nothing but pages and pages of disclaimers.

glyph:
> Meh. Tests aren't equivalent units of work; the only way to know cost (which is what I assume you are referring to) is to remember what they cost in the past or something.

I'm not talking about implementing perfectly fair scheduling between workers - I'm saying that if you can tell in advance that a test should be skipped, you don't need to do deal with the overhead of sending messages to _any_ worker; you can just skip it during test discovery.

wsanchez:
> Wait - why is it unlikely to get re-added? Wouldn't it need to get re-added?

The test is incorrect when run as root, but it is incorrect in a manner that will result in a misleading success result.  So I suspect nobody will notice this, ever, so it won't get fixed.

It's not a huge deal; skipping the test doesn't start flagging bad code, but I think it's useful to know that a test doesn't apply versus it happens to work.  Certainly that would be more correct.

> I'm not talking about implementing perfectly fair scheduling between workers - I'm saying that if you can tell in advance that a test should be skipped, you don't need to do deal with the overhead of sending messages to any worker; you can just skip it during test discovery.

OK.  Meh.  :-)

> To be clear about the policy position: the policy position is that "supported" == "we have a builder that tests in that environment". 

Yes, and that's not the policy I have a problem with.  What I dislike is being forced to work around a policy and tooling combination even when it means inferior code, and I think this is an example of that.

>  It is amazing to me that almost every single thing you do ends up running into this wall head-first one way or another - other folks at least manage to avoid it sometimes :).

It's a superpower.

wsanchez:
OK, merging

PR #621, opened 2016-12-08
[#8937] Remove deprecated parts of FilePath

hawkowl:
https://tm.tl/#8937

PR #623, opened 2016-12-09
Revert "[#8937] Remove deprecated parts of FilePath"

glyph:
Reverts twisted/twisted#621

As @exarkun [put it](https://twistedmatrix.com/trac/ticket/8937#comment:6):

> Regressions for at least two other projects. Beyond just being a technical violation of the compatibility policy. Probably should revert?

Yeah, we should.


glyph:
Twistedchecker, of course, fails because the removed code wasn't up to snuff…

PR #624, opened 2016-12-11
[#6712] Use HostnameEndpoint with Agent for HTTPS

glyph:
https://twistedmatrix.com/trac/ticket/6712


glyph:
This is probably going to fail the nodeps builders, since `wrapClientTLS` depends pretty intimately on SSL actually existing for SSL tests.  I'll fix the computation of the skips soon, but it's otherwise in good shape for a review.

itamarst:
In general looks good (I skimmed the tests a bit more than earlier code). However, since the goal was to make Agent work with IPv6, perhaps a test that actually does that should be added?

glyph:
Thanks @itamarst !

Since you didn't specify a "next action" I'm going to interpret this as "add a test that verifies IPv6 connectivity for Agent, then land".

The reason I didn't _begin_ with such a test is that, as it turns out, there's no new code here specifically for IPv6, so there was no opportunity to write a test that failed.  Previously there was no way to mock out name resolution, and once name resolution is mocked out, the handling of an IPv6 string vs. an IPv4 string is entirely internal to `HostnameEndpoint`.

That said, those are just reasons that a unit test was not required in TDD discipline; a little integration test to make sure nothing blows up once an `IPv6Address` gets in the mix would not be amiss here.  I'll add and merge.

itamarst:
(It did seem like TDD would be tricky, yeah, since code changed so much.)

PR #625, opened 2016-12-12
[#8941] spawnProcess should handle Unicode better

hawkowl:
https://tm.tl/#8941


src/twisted/internet/interfaces.py
```diff
@@ -1142,82 +1142,83 @@ def spawnProcess(processProtocol, executable, args=(), env={}, path=None,
...
-                          keys with empty values in an effort to unset
-                          environment variables I{won't} unset them.
+            sequence of strings.  The first string should be the executable's
+            name.
+        @type args: L{list} with L{bytes} or L{unicode} items.
+
+        @type env: a L{dict} mapping L{bytes}/L{unicode} keys to
+            L{bytes}/L{unicode} items, or L{None}.
+        @param env: the environment variables to pass to the child process.
+            The resulting behavior varies between platforms.  If::
```
glyph: This `::` is incorrect; that will make the stuff below a code block, which I don't think is what you want.  Let me guess - you're working around https://github.com/glyph/python-docstring-mode/issues/1 ?

hawkowl: hmm yeah. if i get rid of the :: after running python-docstring-mode, it'll be fine?

src/twisted/internet/_dumbwin32proc.py
```diff
@@ -28,19 +29,39 @@
...
 from twisted.python.win32 import quoteArguments
+from twisted.python.util import _replaceIf
 
 from twisted.internet import error
 
 from twisted.internet import _pollingfile
 from twisted.internet._baseprocess import BaseProcess
 
+
+@_replaceIf(_PY3, lambda x: os.fsdecode(x))
```
wsanchez: Shouldn't `os.fsdecode` work directly, rather than wrapping it in a lambda?

src/twisted/internet/interfaces.py
glyph: Yep, just delete one `:`.  Although one reason that bug hasn't been fixed is that nested lists are often over-complex for API docs, and should be made flatter :)

src/twisted/internet/_dumbwin32proc.py
hawkowl: It's not available on 2.x, so it has to be wrapped like this.

glyph: It could just be `getattr(os, 'fsdecode', None)`?

src/twisted/topfiles/8941.feature
```diff
@@ -0,0 +1 @@
+twisted.internet.reactor.spawnProcess now does not raise a deprecation warning on Unicode arguments. It will encode Unicode arguments down to bytes using the filesystem encoding on UNIX and Python 2 on Windows, and pass Unicode through unchanged on Python 3 on Windows.
```
glyph: The right language here would be "emit", not "raise".  "raise" should generally be reserved for exceptions.

src/twisted/internet/_dumbwin32proc.py
```diff
@@ -178,13 +199,13 @@ def __init__(self, reactor, protocol, command, args, environment, path):
 
         if _PY3:
             # Make sure all the arguments are Unicode.
-            args = [_maybeMBCS(x) for x in args]
+            args = [_fsdecode(x) for x in args]
```
glyph, line 202: This is not a new change in this PR, so not feedback for this PR, but all this transcoding stuff should happen _before_ any of the syscalls; if we're going to bail out due to an encoding error we should do it _fast_, before we do a bunch of potentially destructive work that I'm not entirely sure will even be cleaned up properly...

src/twisted/internet/interfaces.py
```diff
@@ -1142,82 +1142,83 @@ def spawnProcess(processProtocol, executable, args=(), env={}, path=None,
...
+
+            If it is the string 'w', similar setup to the previous case will
+            occur, with the pipe being readable by the child instead of
+            writeable.  The parent process can write to that file descriptor
+            using L{IProcessTransport.writeToChild}.  This is useful for the
+            child's stdin.
+
+            If childFDs is not passed, the default behaviour is to use a
+            mapping that opens the usual stdin/stdout/stderr pipes.
+        @type childFDs: L{dict} of L{int} to L{int} or L{str}
```
glyph, line 1224: given that this is a new type, is it `native L{str}`?

glyph:
This looks like an improvement. Please fix the NEWS file, and land.  The other feedback you may want to take under consideration for future maintenance.

One other bit of feedback here is that I think _documenting_ the significance of `unicode` vs. `bytes` in `IReactorProcess` would be very useful, or at least pointing at the relevant section of the Python documentation.

PR #626, opened 2016-12-12
Port test_distreporter to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8943


glyph:
Please note that every build on this branch is currently causing appveyor to time out, which pushes out the build queue by 4 hours each time.  If you can, please ensure that the next commit pushed here addresses the Windows failures, so it doesn't push things out unduly further.

rodrigc:
@adiroiban can you kill this build on appveyor: https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-1979

src/twisted/trial/_dist/test/test_workertrial.py
```diff
@@ -102,13 +102,16 @@ def test_forwardCommand(self):
...
-        client.callRemote(workercommands.Run, testCase="doesntexist")
+        client.callRemote(workercommands.Run, testCase=b"doesntexist")
         self.readStream = clientTransport.io
         self.readStream.seek(0, 0)
         main(self.fdopen)
         self.assertIn(
             "No module named 'doesntexist'", self.writeStream.getvalue())
 
+    if _PY3:
+        test_forwardCommand.skip = "Does not work on Python 3 (https://tm.tl/8944)"
```
mesozoic, line 113: I don't have a lot of context for how critical this test is. Should we consider folding the follow-up issue into this branch? If not, can you weigh in on the impact of possibly shipping a version of Twisted that doesn't test this?

src/twisted/trial/_dist/workercommands.py
```diff
@@ -15,14 +15,14 @@ class Run(Command):
     """
     Run a test.
     """
-    arguments = [('testCase', String())]
-    response = [('success', Boolean())]
+    arguments = [(b'testCase', String())]
```
mesozoic, line 18: If we switched this to use `t.p.amp.Unicode` instead, would we be able to avoid the Unicode Sandwich in `t.t._dist.workerreporter` while retaining compatibility with Python 2.x?

mesozoic:
* It looks like Buildbot already has a clean run against the current SHA. 👊
* I get kind of anxious about the [Unicode Sandwich](http://nedbatchelder.com/text/unipain/unipain.html#35) in `twisted.trial._dist`, but there might not be a simpler option.
* In comments above, I asked for clarification on one newly disabled test and wonder if there's a way to avoid calling encode/decode, but those are not showstoppers in my view.

Pending responses to above, good to merge IMO.

src/twisted/trial/_dist/workercommands.py
rodrigc: Using t.p.amp.Unicode didn't seem to work when I tried it.

src/twisted/trial/_dist/test/test_workertrial.py
rodrigc: This test doesn't seem too critical.  Omitting this test, but bringing more of the trial code into Python 3 allows more coverage.  Since trial is exercised by running all of the Twisted tests, that is a good indication that it is OK.

src/twisted/trial/_dist/workercommands.py
markrwilliams: It should be only Unicode on Python 3 since `TestCase`, `Failure`, etc. define these in terms of native strings.

PR #627, opened 2016-12-14
Don't raise error if reason can't be parsed from status line.

bwright1558:


rodrigc:
Make sure you follow all the steps listed here for submitting a patch to Twisted:

tomprince:
Hey @bwright1558,

Thanks for your contribution. As @rodrigc indicated in his link, we require a ticket on the issue tracker [here](http://twistedmatrix.com/trac/) for any changes.

However, I don't think this is something we want to merge, as the HTTP specification [says](https://tools.ietf.org/html/rfc7230#section-3.1.2) that the reason is required (although it may be empty). 
It isn't clear to me why we should be more lenient that the specification in this regard, I'm going to close this PR, but feel free to open an issue (and re-open this PR) with the motivation for this change.

(Also, all changes require tests, if you do decide to re-open this)

PR #628, opened 2016-12-15
[#8945] Remove oldstyle classes from twisted.python

hawkowl:
https://tm.tl/#8945


tomprince (approved):
This looks good. There were a few twistedchecker errors. I fixed the ones about missing docstrings, but there is still one about build_ext_twisted. I think that falls under https://github.com/twisted/twistedchecker/issues/76 as we are following the distutils naming here. In any case, this is an existing error, so I suggest merging, assuming everything else is passing.

PR #574, opened 2016-10-27
[#8892] Made it more clear where the existing `connection` parameter comes from.

cwaldbieser:
https://twistedmatrix.com/trac/ticket/8892

It is not really clear from the documentation where the `connection` parameter for `twisted.conch.endpoints.SSHCommandClientEndpoint.existingConnection()` is supposed to come from.  The example program _does_ use `proto.transport.conn`, but calling this out in the documentation makes it more obvious.

PR #631, opened 2016-12-16
[#8892] Made it more clear where the existing `connection` parameter comes from.

rodrigc:
https://twistedmatrix.com/trac/ticket/8892

Moved from https://github.com/twisted/twisted/pull/574


rodrigc:
@cwaldbieser thanks for the contribution!

PR #632, opened 2016-12-17
[#8950] Skip on network error in API building tests

hawkowl:
https://twistedmatrix.com/trac/ticket/8950


src/twisted/python/test/test_release.py
```diff
@@ -356,12 +357,38 @@ def test_replaceInFile(self):
...
+    network failures -- mamely, APIBuilder being unable to get the "object
+    inventory" of other projects.
+
+    @param func: The function to decorate.
+
+    @return: A decorated function which won't fail if the object inventory
+        fetching fails.
+    """
+    @functools.wraps(func)
+    def _(*a, **kw):
```
wsanchez: Can this be `wrapper(*args, **kwargs)`?
`_` is OK for a symbol that's ignored, but this is actually used again later, so ew.

src/twisted/python/test/test_release.py
```diff
@@ -356,12 +357,38 @@ def test_replaceInFile(self):
...
+        try:
+            func(*a, **kw)
+        except FailTest as e:
+            if "Failed to get object inventory" in e.args[0]:
```
wsanchez: Suggest `e.args[0].startswith("Failed to get object inventory from ")`

src/twisted/python/test/test_release.py
```diff
@@ -356,12 +357,38 @@ def test_replaceInFile(self):
...
+                raise SkipTest(
+                    ("This test is prone to intermittent network errors. "
+                     "See ticket 8753. Exception was: %r") % (e,))
```
wsanchez: % formatting is gross.

wsanchez (approved):
Looks good.

Comments in the review are not blockers.

wsanchez:
codecov is failing, looks like `doNotFailOnNetworkError` needs to be tested for the case where it catches an exception that isn't the one it's trying to ignore.

PR #634, opened 2016-12-20

rodrigc:


src/twisted/python/_release.py
```diff
@@ -39,14 +40,50 @@
...
+    Execute a vector of arguments.
 
     This is a wrapper around L{subprocess.check_output}, so it takes
     the same arguments as L{subprocess.Popen} with one difference: all
     arguments after the vector must be keyword arguments.
     """
     kwargs['stderr'] = STDOUT
-    return check_output(args, **kwargs)
+    output = check_output(args, **kwargs)
+    if not isinstance(output, str):
```
glyph, line 57: The care that is being taken with the output type here should be reflected in the docstring.  What type does `runCommand` return?

src/twisted/python/_release.py
```diff
@@ -158,7 +195,7 @@ def exportTo(fromDir, exportDir):
         """
         runCommand(["git", "-C", fromDir.path,
                     "checkout-index", "--all", "--force",
-                    # prefix has to end up with a "/" so that files get copied
+                    # Prefix has to end up with a "/" so that files get copied
```
glyph: I'm assuming that this unrelated change is to quiet down twistedchecker :).  If you want to make a code-identifier the subject of a sentence, you can fake out twistedchecker by using punctuation at the start of the sentence.  Often this means you can use `` `backticks` `` around the identifier, but in this case, the more precise / correct fix is to just say `--prefix has to end up with ...`, which twistedchecker (by design) won't complain about.

src/twisted/python/_release.py
```diff
@@ -211,6 +248,7 @@ def __repr__(self):
 
     def getVersion(self):
         """
+
```
glyph: This seems wrong to me - docstrings shouldn't have to start with a blank line. Is twistedchecker complaining about it?  If so perhaps we should file a bug; "`@return` is the only thing in a docstring" seems like an OK thing that pydoctor could be corrected to understand (making the `@return` the summary, etc).

src/twisted/python/_release.py
```diff
@@ -482,13 +532,19 @@ def _writeMisc(self, fileObj, header, tickets):
         if not tickets:
             return
 
-        fileObj.write(header + '\n' + '-' * len(header) + '\n')
+        line = header + '\n' + '-' * len(header) + '\n'
+        if not isinstance(line, bytes):
```
glyph: `line` is a variable that you just synthesized; you should know its type.  Rather than `isinstance`-checking it after creating it (a scenario which I believe can only occur on python2 where implicit encoding is allowed), you should use something like [`_matchingString`](https://twistedmatrix.com/documents/16.6.0/api/twisted.python.compat.html#_matchingString) for the `header`, and explicit `b` prefixes.

src/twisted/python/_release.py
```diff
@@ -39,14 +40,50 @@
...
     kwargs['stderr'] = STDOUT
-    return check_output(args, **kwargs)
+    output = check_output(args, **kwargs)
+    if not isinstance(output, str):
+        output = output.decode("utf-8")
+    return output
+
+
+
+def setStrContent(fileObj, content, encoding="utf-8"):
```
glyph: Naming suggestion: I like using the word `lenient` in functions like this, because it explicitly makes it clear that it's accepting of multiple types.

src/twisted/python/test/test_release.py
```diff
@@ -379,43 +390,45 @@ def test_build(self):
...
         inputPath = FilePath(self.mktemp()).child(packageName)
         inputPath.makedirs()
-        inputPath.child("__init__.py").setContent(
+        inputPathContent = (
             "def foo():\n"
             "    '%s'\n"
             "def _bar():\n"
             "    '%s'" % (docstring, privateDocstring))
-
+        setStrContent(inputPath.child("__init__.py"), inputPathContent)
```
glyph: This binds another name in a very long function, which means one more thing for a reader to scan up and down to see if it's used.

My preference would be to do something like

```python
        setStrContent(
              inputPath.child("__init__.py"),
              "def foo():\n"		              "def foo():\n"
              "    '%s'\n"		              "    '%s'\n"
              "def _bar():\n"		              "def _bar():\n"
              "    '%s'" % (docstring, privateDocstring)
        )
```
to keep it an expression and avoid the subsequently unused `inputPathContent` name.

src/twisted/python/test/test_release.py
```diff
@@ -734,8 +755,11 @@ def test_writeHeader(self):
         """
         output = StringIO()
         self.builder._writeHeader(output, "Super Awesometastic 32.16")
+        content = output.getvalue()
+        if not isinstance(content, str):
```
glyph: `isinstance` checks in tests are the worst, since they morph the tests to test different things on different versions of Python, which is easy to screw up in a way where you're testing the wrong thing.  I think it's right here, but I'm not even confident in my ability to always catch that as a reviewer.  You're always using a `BytesIO` here, so you know that it's going to be `bytes`; perhaps just a `b` prefix on the test value wold do what you want?

glyph (commented):
Thanks for porting this!

The major issue is the failing tests everywhere, and that obviously needs to be fixed.  But in the interests of keeping you unblocked I made sure to do a full review as well.

If the tests were passing, this would all be optional feedback, since the changes are entirely in non-public code and therefore interface compatibility is not a concern, but the failures do necessitate a re-review.

So please address the test failures, and then take any of the comments here into consideration as you see fit, and then resubmit for review.

Thanks again!  Looking forward to deleting the last entry from that list of modules :).

src/twisted/python/_release.py
rodrigc: runCommand returns the return value of subprocess.check_output(): https://docs.python.org/3/library/subprocess.html#subprocess.check_output

glyph: Well, yes, but what I mean is:

1. this should be documented in its docstring as an `@rtype`
2. its type, specifically, should be documented.  is it supposed to be `L{bytes}` or `L{unicode}`?

src/twisted/python/_release.py
```diff
@@ -39,14 +40,50 @@
...
+    """
+    if not isinstance(content, bytes):
+        content = content.encode(encoding)
+    fileObj.setContent(content)
+
+
+
+def getStrContent(fileObj, encoding="utf-8"):
+    """
+    Get the content from a L{FilePath} and return it as L{str}.
```
glyph, line 81: Since this is designed specifically to return it as a `str` as opposed to `bytes` or `unicode`, it should use the disambiguating prefix "native".  Older code that is confused about text/bytes often documents types as `str` and that's sometimes vague when it comes to how other cross-version code should expect to consume it.

glyph:
The coverage issues look like... some of the _tests_ aren't even running?

rodrigc:
Superseded by https://github.com/twisted/twisted/pull/713

PR #635, opened 2016-12-20
Force packages that are already new-style clean to remain that way.

tomprince:
https://twistedmatrix.com/trac/ticket/8952#ticket


rodrigc (approved):
This looks OK

PR #637, opened 2016-12-20
Make Fancy{Str,Eq}Mixin conditionally new-style.

tomprince:
This is half the old-style classes in `twisted.python.util`. The other two aren't meant to be subclassed and don't have full coverage; while these are subclassed all over the place.

No ticket, since spambayes doesn't like me.


rodrigc (approved):
This looks OK.  Can you create a ticket for it?
I went in the spam filter and deleted some false positives associated with your ID.

tomprince:
Closing in favor of #628.

PR #638, opened 2016-12-20
Port test_pbfailure to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8809

### DIFF
--- a/src/twisted/spread/pb.py
+++ b/src/twisted/spread/pb.py
@@ -455,9 +455,11 @@ class CopiedFailure(RemoteCopy, failure.Failure):
     def printTraceback(self, file=None, elideFrameworkCode=0, detail='default'):
         if file is None:
             file = log.logfile
+        failureType = self.type
+        if not isinstance(failureType, str):
+            failureType = failureType.decode("utf-8")
         file.write("Traceback from remote host -- ")
-        file.write(self.traceback)
-        file.write(self.type + ": " + self.value)
+        file.write(failureType + ": " + self.value)
         file.write('\n')
 
 
--- a/src/twisted/spread/pb.py
+++ b/src/twisted/spread/pb.py
@@ -299,7 +299,8 @@ class RemoteReference(Serializable, styles.Ephemeral):
         """
         Register a callback to be called if our broker gets disconnected.
 
-        This callback will be called with one argument, this instance.
+        @param callback: a callable which will be called with one
+                         argument, this instance.
         """
         assert callable(callback)
         self.disconnectCallbacks.append(callback)
@@ -310,6 +311,8 @@ class RemoteReference(Serializable, styles.Ephemeral):
     def dontNotifyOnDisconnect(self, callback):
         """
         Remove a callback that was registered with notifyOnDisconnect.
+
+        @param callback: a callable
         """
         self.disconnectCallbacks.remove(callback)
         if not self.disconnectCallbacks:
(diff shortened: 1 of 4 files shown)


src/twisted/spread/test/test_pbfailure.py
```diff
@@ -390,7 +405,10 @@ def generatorFunc():
         self.assertRaises(StopIteration, copy.throwExceptionIntoGenerator, gen)
         self.assertEqual(len(exception), 1)
         exc = exception[0]
-        self.assertEqual(exc.remoteType, "exceptions.AttributeError")
+        if _PY3:
```
glyph: As I keep saying, `if _PY3` is an anti-pattern and we should do it in as few places as possible.  It's just a recipe for coverage problems in the future.  We definitely shouldn't be doing it in test bodies, since it is an instance of [the "conditional test logic" code smell](http://xunitpatterns.com/Conditional%20Test%20Logic.html).

A better way to express this here would be `expectedExceptionType = qual(AttributeError)`.

(That said, this is not bad enough to block on, just please keep it in mind, since every `if _PY3` is technical debt we're eventually going to need to get rid of.)

glyph (approved):
One minor note on `if _PY3`; fix if you have time to do so, but mainly it's a note for the future. Please land.

src/twisted/spread/pb.py
```diff
@@ -455,9 +491,11 @@ class CopiedFailure(RemoteCopy, failure.Failure):
     def printTraceback(self, file=None, elideFrameworkCode=0, detail='default'):
         if file is None:
             file = log.logfile
+        failureType = self.type
+        if not isinstance(failureType, str):
+            failureType = failureType.decode("utf-8")
```
exarkun, line 496: This is interesting.  It looks like it would only work on Python 2 unintentionally because `failureType` probably usually only contains ASCII code points?

That is, on Python 2, if failureType is not `str` then most likely it is `unicode` and `unicode_str.decode("utf-8")` is equivalent to `unicode_str.encode("ascii").decode("utf-8")` which fails if `unicode_str` contains code points which ASCII can't encode.

src/twisted/spread/pb.py
```diff
@@ -455,9 +491,11 @@ class CopiedFailure(RemoteCopy, failure.Failure):
...
         file.write("Traceback from remote host -- ")
-        file.write(self.traceback)
-        file.write(self.type + ": " + self.value)
+        file.write(failureType + ": " + self.value)
```
exarkun, line 498: Having just decoded `failureType` to unicode, wouldn't it make most sense to concatenate it with unicode rather than a native string?

src/twisted/spread/pb.py
```diff
@@ -892,16 +1020,29 @@ def _sendMessage(self, prefix, perspective, objectID, message, args, kw):
...
+        @param findObjMethod: A callable which takes C{objectID} as argument.
+        @param requestID: The requiest ID.
+        @param objectID: The object ID.
+        @param message: The message.
+        @param answerRequired:
+        @param netArgs: Arguments.
+        @param netKw: Keyword arguments.
         """
         if not isinstance(message, str):
             message = message.decode('utf8')
```
exarkun, line 1048: Huh, and here's that decode pattern again.  Confusing.

src/twisted/spread/pb.py
```diff
@@ -1425,13 +1633,24 @@ def remote_respond(self, response, mind):
         return d
 
 
-    # IUsernameHashedPassword:
     def checkPassword(self, password):
+        """
+        L{IUsernameHashedPassword}
+
+        @param password: The password.
+        @return: L{_PortalAuthChallenger.checkMD5Password}
```
exarkun, line 1641: Pedantically, it doesn't return that.  It just returns a similar result given its similar input.

PR #642, opened 2016-12-22
In KEXINIT packet, put EC algorithms first in list of host key algorithms

rodrigc:
http://twistedmatrix.com/trac/ticket/8958

### DIFF
--- a/src/twisted/conch/client/default.py
+++ b/src/twisted/conch/client/default.py
@@ -130,6 +130,29 @@ def isInKnownHosts(host, pubKey, options):
 
 
 
+def getHostKeyAlgorithms(host, options):
+    """
+    Look in known_hosts for a key corresponding to C{host}.
+    This can be used to change the order of supported key types
+    in the KEXINIT packet.
+
+    @param host: the host to check in known_hosts
+    @param options: L{twisted.conch.client.options.ConchOptions}
+    @return: L{list} of L{str} representing key types or L{None}.
+    """
+    knownHosts = KnownHostsFile.fromPath(FilePath(
+        options['known-hosts']
+        or os.path.expanduser(_KNOWN_HOSTS)
+        ))
+    keyTypes = []
+    for entry in knownHosts.iterentries():
+        if entry.matchesHost(host):
+            if entry.keyType not in keyTypes:
+                keyTypes.append(entry.keyType)
+    return keyTypes or None
+
+
+
 class SSHUserAuthClient(userauth.SSHUserAuthClient):
 
     def __init__(self, user, options, *args):
--- a/src/twisted/conch/client/default.py
+++ b/src/twisted/conch/client/default.py
@@ -136,8 +136,10 @@ def getHostKeyAlgorithms(host, options):
     This can be used to change the order of supported key types
     in the KEXINIT packet.
 
+    @type host: L{str}
     @param host: the host to check in known_hosts
-    @param options: L{twisted.conch.client.options.ConchOptions}
+    @type options: L{twisted.conch.client.options.ConchOptions}
+    @param options: options passed to client
     @return: L{list} of L{str} representing key types or L{None}.
     """
     knownHosts = KnownHostsFile.fromPath(FilePath(
--- a/src/twisted/conch/client/knownhosts.py
+++ b/src/twisted/conch/client/knownhosts.py
@@ -431,7 +431,7 @@ class KnownHostsFile(object):
             does not match the given key.
         """
         for lineidx, entry in enumerate(self.iterentries(), -len(self._added)):
-            if entry.matchesHost(hostname):
+            if entry.matchesHost(hostname) and entry.keyType == key.sshType():
                 if entry.matchesKey(key):
                     return True
                 else:
--- a/src/twisted/conch/test/test_knownhosts.py
+++ b/src/twisted/conch/test/test_knownhosts.py
@@ -961,6 +961,26 @@ class KnownHostsDatabaseTests(TestCase):
             ui.userWarnings)
 
 
+    def test_getHostKeyAlgorithms(self):
+        """
+        For a given host, get the host key algorithms for that
+        host in the known_hosts file.
+        """
+        hostsFile = self.loadSampleHostsFile()
+        hostsFile.addHostKey(
+            b"www.twistedmatrix.com", Key.fromString(otherSampleKey))
+        hostsFile.addHostKey(
+            b"www.twistedmatrix.com", Key.fromString(ecdsaSampleKey))
+        hostsFile.save()
+        options = {}
+        options['known-hosts'] = hostsFile.savePath.path
+        algorithms = default.getHostKeyAlgorithms(
+                         b"www.twistedmatrix.com", options)
+        expectedAlgorithms = [b'ssh-rsa', b'ecdsa-sha2-nistp256']
+        self.assertEqual(algorithms, expectedAlgorithms)
+
+
+
 class FakeFile(object):
     """
     A fake file-like object that acts enough like a file for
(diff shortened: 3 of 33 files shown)


rodrigc:
@the0id please review this

rodrigc:
@acabhishek942 @the0id please review this

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,29 @@ def isInKnownHosts(host, pubKey, options):
...
 
 
+def getHostKeyAlgorithms(host, options):
+    """
+    Look in known_hosts for a key corresponding to C{host}.
+    This can be used to change the order of supported key types
+    in the KEXINIT packet.
+
+    @param host: the host to check in known_hosts
+    @param options: L{twisted.conch.client.options.ConchOptions}
```
acabhishek942: please include the type of `host` and `options` here

acabhishek942 (approved):
Please address my inline comment if you think thats important, otherwise this change looks ready to merge

glyph:
@acabhishek942 In the future (at least until we are fully moved to github) please take care to remove the `review` keyword on the Trac ticket to indicate this officially counts as a review.  Thanks for taking the time to do a code review though!

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,31 @@ def isInKnownHosts(host, pubKey, options):
 
 
 
+def getHostKeyAlgorithms(host, options):
+    """
+    Look in known_hosts for a key corresponding to C{host}.
+    This can be used to change the order of supported key types
+    in the KEXINIT packet.
+
+    @type host: L{str}
```
glyph, line 139: While this isn't _wrong_ exactly, convention is usually to put `@param` first, rather than `@type`.

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,31 @@ def isInKnownHosts(host, pubKey, options):
 
 
 
+def getHostKeyAlgorithms(host, options):
```
glyph, line 133: Was it really necessary to expose a new _public_ API to address this?

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,31 @@ def isInKnownHosts(host, pubKey, options):
...
+    """
+    Look in known_hosts for a key corresponding to C{host}.
+    This can be used to change the order of supported key types
+    in the KEXINIT packet.
+
+    @type host: L{str}
+    @param host: the host to check in known_hosts
+    @type options: L{twisted.conch.client.options.ConchOptions}
```
glyph, line 141: Why does this take a `ConchOptions` rather than a `KnownHostsFile`?  It seems that this is duplicating the logic involved in determining the proper location of the `known_hosts` file.

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,31 @@ def isInKnownHosts(host, pubKey, options):
...
+    @param options: options passed to client
+    @return: L{list} of L{str} representing key types or L{None}.
```
glyph, line 143: Should this really be native `str` or is it more properly `unicode` or `bytes`?

src/twisted/conch/client/default.py
```diff
@@ -130,6 +130,31 @@ def isInKnownHosts(host, pubKey, options):
...
+    """
+    knownHosts = KnownHostsFile.fromPath(FilePath(
+        options['known-hosts']
+        or os.path.expanduser(_KNOWN_HOSTS)
+        ))
+    keyTypes = []
+    for entry in knownHosts.iterentries():
```
glyph, line 150: It seems like this should be a `set` rather than a `list`, since order isn't really significant.  This would also be easier to spell / implement, as simply `return set([entry.keyType for entry in knownHosts.iterentries() if entry.matchesHost(host)])`

src/twisted/conch/client/knownhosts.py
```diff
@@ -431,7 +431,7 @@ def hasHostKey(self, hostname, key):
             does not match the given key.
         """
         for lineidx, entry in enumerate(self.iterentries(), -len(self._added)):
-            if entry.matchesHost(hostname):
+            if entry.matchesHost(hostname) and entry.keyType == key.sshType():
```
glyph, line 434: This feels like a separate fix that should have been addressed in a different ticket?

src/twisted/conch/test/test_knownhosts.py
```diff
@@ -53,9 +53,14 @@
     b'343Hd2QHiIE0KPZJEgCynKeWoKz8v6eTSK8n4rBnaqWdp8MnGZK1WGy05MguXbyCDuTC8AmJXQ'
     b'==')
 
+ecdsaSampleEncodedKey = (
```
glyph, line 56: Just for future reference, we should try to generate keys for tests rather than having a bunch of hard-coded ones hanging around.  This is the sort of thing automated security scanners pick up and complain about.  There's a case to be made that they're technically wrong for complaining, but they're also the sort of thing that people blindly copy/paste from tests or examples into production systems, so in practice I think they're right.

This is just repeating a pretty well-established pattern in this file, so I wouldn't blame you for doing it here, but it's something to keep in mind for the future.

glyph (commented):
I know this has already been reviewed, and I don't want to propose a revert, but I think we may not be properly propagating our attention to details in reviews to new reviewers - the number of "+1" reviews on relatively complex changes and new public APIs concerns me a little.  Just wanted to leave my own take on the review here as an example :).

src/twisted/conch/client/default.py
rodrigc: Maybe this could have been left as an internal API.  Maybe there could be other clients outside of the conch client that could guess at a preferred host key type based on what is in known_hosts

src/twisted/conch/client/default.py
rodrigc: There seem to be other places in Twisted that put @type before @param such as https://github.com/twisted/twisted/blob/trunk/src/twisted/words/xish/domish.py#L158 so I tried following that, but I'll keep that in mind next time around.

src/twisted/conch/client/default.py
rodrigc: After going through things and testing it out, str seems to be the most workable.

src/twisted/conch/client/knownhosts.py
rodrigc: I had it in a separate ticket, but it was easier to combine in one, because the fixes are interrelated, having to do with deducing the host key algorithm from the known_hosts file,
and using that to specify the order of algorithms in the client's kexinit packet.

src/twisted/conch/client/default.py
rodrigc: During my testing, I found that  in the OpenSSH command-line client that the order that entries are traversed in known_hosts seems to matter.  I tried to keep the behavior of the conch client the same as OpenSSH in this area, so kept it as a list.

src/twisted/conch/client/default.py
glyph: Thanks for the reference to the counterexample; I wasn't aware that was there. I'll have to go through and fix it at some point :)

src/twisted/conch/client/default.py
glyph: if that is the case, perhaps it makes sense to move it to a method on `KnownHostsFile`?  Perhaps good to do this before the next release?

src/twisted/conch/client/default.py
glyph: Aah, interesting. I guess that's why it talks about "order" in the docstring? That raises the question though: how can this be "used" to do that?

PR #643, opened 2016-12-22
[#8960] Officially deprecate `getPage`.

tomprince:
https://twistedmatrix.com/trac/ticket/8960#ticket

### DIFF
--- a/docs/web/examples/index.rst
+++ b/docs/web/examples/index.rst
@@ -9,8 +9,9 @@ Examples
 twisted.web.client
 ------------------
 
-- :download:`getpage.py` - use ``twisted.web.client.getPage`` to download a web page.
-- :download:`dlpage.py` - add callbacks to ``twisted.web.client.downloadPage`` to display errors that occur when downloading a web page
+- :download:`httpclient.py` - use ``twisted.web.client.Agent`` to download a web page.
+- (deprecated) :download:`getpage.py` - use ``twisted.web.client.getPage`` to download a web page.
+- (deprecated) :download:`dlpage.py` - add callbacks to ``twisted.web.client.downloadPage`` to display errors that occur when downloading a web page
 
 
 XML-RPC
--- a/src/twisted/web/client.py
+++ b/src/twisted/web/client.py
@@ -29,7 +29,7 @@ from zope.interface import implementer
 from twisted.python import log
 from twisted.python.compat import _PY3, networkString
 from twisted.python.compat import nativeString, intToBytes, unicode, itervalues
-from twisted.python.deprecate import deprecatedModuleAttribute
+from twisted.python.deprecate import deprecatedModuleAttribute, deprecated
 from twisted.python.failure import Failure
 from incremental import Version
 
@@ -737,6 +737,32 @@ def _makeGetterFactory(url, factoryFactory, contextFactory=None,
     return factory
 
 
+
+def _deprecateGetPageClasses():
+    """
+    Mark the protocols and factories associated with L{getPage} and
+    L{downloadPage} as deprecated.
+    """
+    for klass in [
+        HTTPPageGetter, HTTPPageDownloader,
+        HTTPClientFactory, HTTPDownloader
+    ]:
+        deprecatedModuleAttribute(
+            Version("Twisted", 16, 7, 0),
+            getDeprecationWarningString(
+                klass,
+                Version("Twisted", 16, 7, 0),
+                replacement="https://pypi.org/project/treq/ or twisted.web.client.Agent")
+            .split("; ")[1],
+            klass.__module__,
+            klass.__name__)
+
+_deprecateGetPageClasses()
+
+
+
+@deprecated(Version("Twisted", 16, 7, 0),
+            "https://pypi.org/project/treq/ or twisted.web.client.Agent")
 def getPage(url, contextFactory=None, *args, **kwargs):
     """
     Download a web page as a string.
@@ -753,6 +779,9 @@ def getPage(url, contextFactory=None, *args, **kwargs):
         *args, **kwargs).deferred
 
 
+
+@deprecated(Version("Twisted", 16, 7, 0),
+            "https://pypi.org/project/treq/ or twisted.web.client.Agent")
 def downloadPage(url, file, contextFactory=None, *args, **kwargs):
     """
     Download a web page to a file.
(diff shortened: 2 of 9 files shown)


docs/web/examples/index.rst
```diff
@@ -9,8 +9,9 @@ Examples
 twisted.web.client
 ------------------
 
-- :download:`getpage.py` - use ``twisted.web.client.getPage`` to download a web page.
-- :download:`dlpage.py` - add callbacks to ``twisted.web.client.downloadPage`` to display errors that occur when downloading a web page
+- :download:`httpclient.py` - use ``twisted.web.client.Agent`` to download a web page.
+- (deprecated) :download:`getpage.py` - use ``twisted.web.client.getPage`` to download a web page.
+- (deprecated) :download:`dlpage.py` - add callbacks to ``twisted.web.client.downloadPage`` to display errors that occur when downloading a web page
```
derwolfe, line 14: Should these remain in the documentation as deprecated options? I looked through the [compatibility policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html) but didn't see a mention of how to handle deprecated code samples. My opinion would be to remove them outright from the documentation to make it so that they can no longer be discovered (but I might be in the minority on that one).

src/twisted/web/client.py
```diff
@@ -753,6 +779,9 @@ def getPage(url, contextFactory=None, *args, **kwargs):
         *args, **kwargs).deferred
 
 
+
+@deprecated(Version("Twisted", 16, 7, 0),
+            "https://pypi.org/project/treq/ or twisted.web.client.Agent")
```
derwolfe: (optional) Since the deprecation message is being used multiple times in this file, what would you think of pulling  into a private module variable?

src/twisted/web/test/test_cgi.py
```diff
@@ -119,7 +120,9 @@ def testCGI(self):
         cgiFilename = self.writeCGI(DUMMY_CGI)
 
         portnum = self.startServer(cgiFilename)
-        d = client.getPage("http://localhost:%d/cgi" % portnum)
+        d = client.Agent(reactor).request(
+            'GET', "http://localhost:%d/cgi" % (portnum,))
```
derwolfe: (optional) it might be more consistent to use `'` or `"` instead of mixing the two. But, since the linter doesn't care, I probably shouldn't either :)

src/twisted/web/test/test_cgi.py
```diff
@@ -240,7 +251,9 @@ def testReadEmptyInput(self):
             cgiFile.write(READINPUT_CGI)
 
         portnum = self.startServer(cgiFilename)
-        d = client.getPage("http://localhost:%d/cgi" % portnum)
+        agent = client.Agent(reactor)
+        d = agent.request("GET", "http://localhost:%d/cgi" % (portnum,))
```
derwolfe: Should this `GET` be specified as bytes?

src/twisted/web/test/test_distrib.py
```diff
@@ -87,8 +88,10 @@ def testDistrib(self):
         r2.putChild("here", self.sub)
         f2 = MySite(r2)
         self.port2 = reactor.listenTCP(0, f2)
-        d = client.getPage("http://127.0.0.1:%d/here/there" % \
-                           self.port2.getHost().port)
+        agent = client.Agent(reactor)
+        d = agent.request("GET", "http://127.0.0.1:%d/here/there" % \
```
derwolfe: Should this `GET` be specified as bytes?

src/twisted/web/test/test_distrib.py
```diff
@@ -128,14 +131,15 @@ def _requestTest(self, child, **kwargs):
...
             requesting the resource.
 
         @return: A L{Deferred} which fires with the result of the request.
         """
         mainPort, mainAddr = self._setupDistribServer(child)
-        return client.getPage("http://%s:%s/child" % (
-            mainAddr.host, mainAddr.port), **kwargs)
+        agent = client.Agent(reactor)
+        url = "http://%s:%s/child" % (mainAddr.host, mainAddr.port)
+        return agent.request('GET', url, **kwargs).addCallback(client.readBody)
```
derwolfe: Should this `GET` be specified as bytes?

src/twisted/web/test/test_webclient.py
```diff
@@ -1381,3 +1395,101 @@ def test_hostBracketIPv6AddressLiteral(self):
         self.assertEqual(uri.host, b"::1")
         self.assertEqual(uri.netloc, b"[::1]:80")
         self.assertEqual(uri.toBytes(), b'http://[::1]:80/index.html')
+
+
+
+class DeprecationTests(unittest.TestCase):
+    """
+    Tests that l{getPage} and friends are deprecated.
```
derwolfe: 1. `L{...}`
2. It seems like this should use the namespace `L{client.getPage}` as below.

derwolfe (changes requested):
Thanks @tomprince! Most of the review feedback is related to tiny nits. Please address and resubmit for review :)

derwolfe:
@tomprince  sorry - by resubmit for review I meant to say address and feel free to merge. Thanks!

docs/web/examples/index.rst
tomprince: The (possibly unwritten) policy is that documentation sticks around as long as the code does. People inheriting a code base that is using the deprecated functionality, and are assiduously updating and cleaning up `DeprecationWarnings` may need to look at the documentation to see what the deprecated things are doing, to figure out how to replace them.

hawkowl: I'll note that when I deprecated `method`, I also removed the documentation for it from our narrative docs. I guess it depends on how bad it is if someone uses it. (I personally am on the side of deleting it, and having the documentation for the One Way. Previous versions of the docs still exist on our website forever.).

derwolfe (approved):

docs/web/examples/index.rst
tomprince: I'm all in favor of putting loud warnings on the narrative documentation of things being deprecated.

exarkun: If you delete it, folks maintaining software written against Twisted X.Y can't easily find documentation for the code they're maintaining anymore.  Please leave the docs until the actual implementation goes away.  Certainly, make it clear that they're docs for a deprecated API until it is deleted.

PR #644, opened 2016-12-23
[#8966] Adding curve25519 and ed25519

the0id:
This adds curve25519 and ed25519 to Twisted.  Cryptography doesn't support 25519 yet, so pyNaCL is required.


the0id:
I'm trying to create a ticket for this at twistedmatrix.com, but Trac doesn't seem to be working.

the0id:
I'm getting a txchecker error: Invalid name "privateED25519_openssh" for type constant (should match (([A-Z_][A-Z0-9_]*)|(__.*__)|([a-z_][a-zA-Z0-9]+)|([A-Z][a-zA-Z0-9]+))$)

I may be missing something, but when I run the below locally I get a match.
`re.search('(([A-Z_][A-Z0-9_]*)|(__.*__)|([a-z_][a-zA-Z0-9]+)|([A-Z][a-zA-Z0-9]+))$', "privateED25519_openssh")` 

If I use re.match, it does not return a hit, but neither do any of the other constant names in the file.

src/twisted/conch/ssh/transport.py
```diff
@@ -1275,31 +1279,44 @@ def _ssh_KEX_ECDH_INIT(self, packet):
...
-                        curve, pktPub).public_key(default_backend())
-
-        # We need to convert to hex,
-        # so we can convert to an int
-        # so we can make it a multiple precision int.
-        sharedSecret = MP(
-                       int(
-                        binascii.hexlify(
-                          ecPriv.exchange(ec.ECDH(), theirECPub)), 16))
+        if self.kexAlg.find(b'25519') != -1:
```
glyph: Just a nitpick, but, this is perhaps more idiomatically expressed as `if b'25519' in self.kexAlg`

src/twisted/python/_setup.py
```diff
@@ -96,6 +96,7 @@
         'pyasn1',
         'cryptography >= 0.9.1',
         'appdirs >= 1.4.0',
+        'pynacl >= 1.0.1',
```
glyph: Additional cryptographic dependencies are often met with high levels of scrutiny.  If possible, I'd really like to see this implemented in a way which makes pynacl an _optional_ dependency.

This doesn't necessarily mean that the entry in `setup.py` needs to go away, but it strikes me that "cryptography yes, pynacl no" might end up being a configuration someone cares about.

src/twisted/conch/ssh/transport.py
```diff
@@ -1275,31 +1279,44 @@ def _ssh_KEX_ECDH_INIT(self, packet):
...
+            # Turn the public key into a public key object
+            theirECPub = nacl.public.PublicKey(pktPub)
+
+            # Always generate new private and public keys
+            ecPriv = nacl.public.PrivateKey.generate()
+            ecPub = ecPriv.public_key
+            encPub = ecPub._public_key
```
glyph, line 1296: This looks like it's reaching into the private implementation details of a dependency.  Is it?  If so - why does it need to?

src/twisted/conch/ssh/transport.py
```diff
@@ -19,6 +19,10 @@
 
 from hashlib import md5, sha1, sha256, sha384, sha512
 
+import nacl.public
```
glyph: See comment in `setup.py` about being an optional dependency.

src/twisted/conch/ssh/transport.py
```diff
@@ -1275,31 +1279,44 @@ def _ssh_KEX_ECDH_INIT(self, packet):
...
+
+            # Always generate new private and public keys
+            ecPriv = nacl.public.PrivateKey.generate()
+            ecPub = ecPriv.public_key
+            encPub = ecPub._public_key
+            sharedSecret = MP(
+                int(binascii.hexlify(
+                  crypto_scalarmult(ecPriv._private_key,
+                                  theirECPub._public_key)), 16))
+        else:
```
glyph, line 1301: One other nice thing about making this an optional dependency would be that we could potentially factor out this extremely long function into a couple of smaller cipher-management functions so as to avoid monolithic changes like having to indent this lengthy "else" block even though the lines aren't being touched.

src/twisted/conch/ssh/transport.py
```diff
@@ -1634,31 +1660,46 @@ def _ssh_KEX_ECDH_REPLY(self, packet):
...
+
+                encPub = self.ecPub._public_key
+            else:
+                # Take the provided public key and transform it into a format
+                # for the cryptography module
+                theirECPub = ec.EllipticCurvePublicNumbers.from_encoded_point(
+                                self.curve, pubKey).public_key(
+                                default_backend())
+
+                # We need to convert to hex,
```
glyph, line 1693: Given that this comment is repeated twice, and now the `MP(...` expression is repeated 4 times, some refactoring seems to be in order.

src/twisted/conch/test/keydata.py
```diff
@@ -84,6 +85,16 @@
   'curve': b'ecdsa-sha2-nistp521'
 }
 
+ED25519Data = {
```
glyph, line 88: I realize that this is just repeating an instance of an existing pattern, but could we replace this with something that generates a test key on the fly, rather than supplying static data?  Long term, we should be deleting every static private key in this module.

src/twisted/python/_setup.py
glyph: This is not mandatory feedback, feel free to push back on this, if you disagree.  I am not really sure about the political environment re: nacl as a crypto lib. Perhaps @reaperhulk or @alex could opine as well, if this is a legitimate concern.

src/twisted/conch/test/test_keys.py
```diff
@@ -1025,6 +1113,28 @@ def test_toOpenSSHECDSA(self):
...
+    def test_toOpenSSHED25519(self):
+        """
+        L{keys.Key.toString} serializes a ED25519 key in OpenSSH format.
+        """
+        key = keys.Key.fromString(keydata.privateED25519openssh)
+        self.assertEqual(key.public().toString('openssh', b'comment'),
+                         keydata.publicED25519openssh)
+
+        # We need to patch the comparison value because SecureRandom
+        # gets patched when run as a unit test.
```
glyph: How is this comparison value "patched"?  Also; can this refer more specifically to what does the `SecureRandom` patching, rather than just alluding to it?  Long-term, this is the sort of thing we should parameterize rather than patch.

glyph (commented):
Thanks for your continued work on Conch's cipher suites.

We should discuss the results of this review, but my general feeling here is that even if the dependency isn't optional, we are 4 mostly copy/pasted algorithms in to needing to refactor the way that conch deals with curves.  I think this one might be a bridge too far.  Do you see an easy way to modularize?  I'm OK with deferring to a subsequent ticket, but I just worry that this is becoming less and less maintainable.

Please address this feedback and resubmit.

src/twisted/python/_setup.py
reaperhulk: libsodium is definitely still the "less acceptable to upper management" crypto library, but its presence here is because OpenSSH already extensively uses crypto primitives from djb and OpenSSL doesn't (yet) support them. If you want to support these things in the near term you'll need to pull in pynacl or do it in pure Python (don't do that).

glyph: The question is really: should Twisted be concerned about environments where someone will say "uh oh, pynacl, you can't install that, denied"?  Or is this the kind of thing that will be configured at a policy level?  (After all you'd need to actually use these ciphers in your deployment to exercise this code.)

reaperhulk: Good question. I've never been a huge fan of optional deps, but a project like twisted probably does need to care about things like that. It doesn't seem unreasonable to state that a base install of twisted conch must support RSA and NIST ecdsa (via a non-optional cryptography dep), but that users who want access to ed25519 should install pynacl. ed25519 is likely to appear in cryptography soon-ish as well, albeit only accessible to users linking against the latest OpenSSL.

src/twisted/conch/ssh/transport.py
the0id: I remember if X in Y causing problems with python3, but that seems to be incorrect and I agree if X in Y is better.  Changed.

src/twisted/conch/ssh/transport.py
the0id: It is, and it's because of how pynacl decided to do their secret key generation.  They call their method "salsa" which follows the standard way of generating a curve25519 secret key but then throws an extra transform on it. Which means to get it to work with Twisted I have to call the private method.

src/twisted/conch/ssh/transport.py
the0id: I see what you're saying, and have had similar thoughts, but I don't know if I'm convinced it's needed yet.

DH, ECDH, and curve25519 all do something like this, but each of them do it in a slightly different way.  So if we were to refactor and abstract this behavior I think we'd just end up moving the same code to a different location.

src/twisted/conch/test/keydata.py
the0id: I don't have a strong opinion on whether or not this should be done, but I think it'd be best to do it as it's own ticket.

src/twisted/conch/test/test_keys.py
the0id: When trial is run, SecureRandom is patched by test_keys.py:217 to return '\xff' for each byte requested.

I wasn't sure how detailed to go into this explanation in the comment since it's only relevant when running trial.  I saw it that it was more important to let a normal user know that every time this function is called the output will be slightly different.

src/twisted/python/_setup.py
the0id: I have issues with how pynacl operates myself, but I found it to be the lesser evil.  There are other libraries that support *25519, but some only support curve25519, some only support ed25519, and most of them haven't been updated in at least a year or two.

How I see things is that pynacl is optional to the extent that cryptography and pyasn1 are optional.  They're not needed unless you want to use conch, and since *25519 is the default (and strongly encouraged) algorithm of OpenSSH it should be included when conch is.

src/twisted/conch/ssh/transport.py
glyph: I don't quite understand; wouldn't "throwing an extra transform" on it mean that the values are different?

src/twisted/python/_setup.py
glyph: @the0id I don't think you're grasping my concern here:

Let's say I work at a company that's using Twisted for SSH, but that has a heavy-weight dependency audit process for each new library that gets deployed.  Now, we're still adopting new dependencies and causing more trouble for people working at such shops; however, new _crypto_ dependencies are often given much tighter scrutiny.  By making this change unconditional, I have to upgrade Twisted and go through the pynacl review process at the exact same time - rather than installing new Twisted immediately and kicking off the review process for pynacl, which, once it gets approved, gives me a new capability.

src/twisted/conch/ssh/transport.py
the0id: Yup. If you use the standard way of computing the key, it will give you a different value than most other curve25519 implementations, OpenSSH included.

Why they do this I don't know, I'm sure they have their reasons, but it makes it incompatible with just about every other implementation unless you call it this way.

the0id: Yup.  If you use the standard method it will return a value that works for pynacl, but won't work for almost every other implementation, including OpenSSH.

Why they do this I have no idea, I'm sure they have their reasons, but this was the only way I could get it to work.

src/twisted/python/_setup.py
the0id: I see what you're saying now, and making curve25519 an optional dependency may not be a much work as I thought. So I'll look into it.

glyph: I should note; it's OK to make the `conch` extra hard depend on this, since for most users, that should be the right thing.  I'm mainly talking about leaving the code in such a state that an escape hatch like a separate `conch-no25519` extra would be possiblel.

alex:
I think you have a merge issue, this appears to be readding back 1024-bit DH and a few other curves we removed.

the0id:
Now that I've had some time to look at this again, making pynacl an option dependency was pretty easy.  With the exception that pyflakes doesn't like how I'm importing nacl.public into _kex.py but not using it.

I see why pyflakes is objecting to this, but I'm doing this to check to see if pynacl is installed before adding curve25519 as a supported kex.  I could create a nacl.public object to make pyflakes happy, but that doesn't seem like a good solution to me.

How should I go about this?

alex:
You can use `requireModule` for that: https://github.com/twisted/twisted/blob/trunk/src/twisted/conch/test/test_openssh_compat.py#L14

the0id:
@alex yeah as with our conversation on #749 I think those curves should be in Twisted, and since they were originally included in this request I'm adding them back in.

1024-DH I put back in as well because it was also originally in this merge, and while I know it's been deprecated, it's still supported and I  think it's worth supporting here as it's still widely used and I believe Twisted should be as compatible with as many applications as possible.

the0id:
@alex  Oh thanks for the link, I'll look into that. :+1:

alex:
This PR is for adding X25519 and Ed25519 support, disagreements about 1024-bit DH or other curves belong somewhere else, not here.

the0id:
@alex I guess you're right.  I'm reverting those changes and will open a new PR.

the0id:
I'm not sure if my last commit made things better or worse, but the root of the problem seems to be the same.

The tests are failing because nacl isn't present. Since those tests are skipped, codecov fails.  Is there a way to get TravisCI to use nacl without having it be required in _setup.py?  How should I do this?

glyph:
My preference would be for you to add it as a new required dependency of the `conch` extra, but then add a *new* extra, `conch_no_25519` or somesuch, which doesn't include pynacl.  You can then test it both ways by adding tox environments which install each extra.

the0id:
@glyph I see what you're saying now.  No problem, can do, as soon as I figure out how to do. :)

the0id:
I can't tell if I did this right.  The new tests ran but codecov didn't.

glyph:
Codecov has the occasional reliability issue :-\

the0id:
I got 99 commits, so lets merge this one.

src/twisted/conch/ssh/_kex.py
```diff
@@ -126,6 +127,25 @@ class _DHGroupExchangeSHA1(object):
 
 
 @implementer(_IFixedGroupKexAlgorithm)
+class _DHGroup1SHA1(object):
```
alex: This still needs to be removed.

src/twisted/conch/ssh/keys.py
```diff
@@ -3,7 +3,7 @@
 # See LICENSE for details.
 
 """
-Handling of RSA, DSA, and EC keys.
+Handling of RSA, DSA, EC, and *25519 keys.
```
alex: This should just say `Ed25519`, this module doesn't do anything with X25519

src/twisted/conch/ssh/keys.py
```diff
@@ -289,6 +304,9 @@ def _fromString_PRIVATE_BLOB(cls, blob):
         elif keyType == b'ssh-dss':
             p, q, g, y, x, rest = common.getMP(rest, 5)
             return cls._fromDSAComponents(y=y, g=g, p=p, q=q, x=x)
+        elif keyType == b'ssh-ed25519':
+            pr = nacl.signing.SigningKey(common.getNS(rest)[0])
```
alex: The variable name `pr` has me a bit confused; can we pick something more intuitive?

src/twisted/conch/ssh/keys.py
```diff
@@ -589,7 +617,10 @@ def _guessStringType(cls, data):
             return 'public_lsh'
         elif data.startswith(b'('):
             return 'private_lsh'
-        elif data.startswith(b'\x00\x00\x00\x07ssh-') or data.startswith(b'\x00\x00\x00\x13ecdsa-'):
+        elif data.startswith(b'\x00\x00\x00\x0bssh-'):  # ED25519
```
alex, line 624: The data will continue `ssh-ed25519` right? Might be easier to read if you include that in the string.

src/twisted/conch/ssh/keys.py
```diff
@@ -786,7 +841,8 @@ def __repr__(self):
                         o = o[:-1]
                     lines.append('\t' + o)
             lines[-1] = lines[-1] + '>'
-            return '\n'.join(lines)
+            reprstr = '\n'.join(lines)
+        return reprstr
```
alex, line 849: These changes appear to be unrelated, there's nothing wrong with just `return` when you have the value.

src/twisted/conch/ssh/keys.py
```diff
@@ -901,7 +958,10 @@ def public(self):
         @rtype: L{Key}
         @return: A public key.
         """
-        return Key(self._keyObject.public_key())
+        if self.type() is 'ED25519':
```
alex: This should not use `is`, it should use `==`

src/twisted/conch/ssh/keys.py
```diff
@@ -958,8 +1018,13 @@ def type(self):
...
             return 'DSA'
         elif isinstance(
-                self._keyObject, (ec.EllipticCurvePublicKey, ec.EllipticCurvePrivateKey)):
+                self._keyObject, (ec.EllipticCurvePublicKey,
+                                  ec.EllipticCurvePrivateKey)):
             return 'EC'
+        elif isinstance(
+                self._keyObject, (nacl.signing.VerifyKey,
+                                  nacl.signing.SigningKey)):
+            return 'ED25519'
```
alex, line 1031: Conventionally, it's typed `Ed25519`, lowercase `d`.

src/twisted/conch/ssh/keys.py
```diff
@@ -1189,69 +1271,115 @@ def _toString_OPENSSH(self, extra):
...
+                b64Data = encodebytes(self.blob()).replace(b'\n', b'')
+                if not extra:
+                    extra = b''
+                retstr = (self.sshType() + b' '
+                          + b64Data + b' ' + extra).strip()
         else:
-
-            if self.type() == 'EC':
+            if self.type() == 'ED25519':
+                # OpenSSH loosely follows the openssh-key-v1 specification
```
alex: This logic should be split out into its own method for readability.

src/twisted/conch/ssh/keys.py
```diff
@@ -1414,7 +1544,18 @@ def verify(self, signature, data):
...
+        if keyType == 'ED25519':
+            if self.isPublic():
+                key = self._keyObject
+            else:
+                key = self._keyObject.verify_key
+
+            try:
+                key.verify(data, common.getNS(signature)[0])
+                return True
+            except:
```
alex: Catch a specific exception, not a bare catch.

src/twisted/conch/ssh/transport.py
```diff
@@ -1578,8 +1602,17 @@ def ssh_KEXINIT(self, packet):
             # Connection was disconnected while doing base processing.
             # Maybe no common protocols were agreed.
             return
+        # Curve25519
+        if self.kexAlg.find(b'curve25519') >= 0:
```
alex: This can just be `b'curve25519' in self.kexAlg`

alex:
There's also a merge conflict.

rodrigc:
You need to take a merge from trunk and move your 8966.feature file to the newsfragments directory.  Things have moved around due to this: https://github.com/twisted/twisted/pull/763

src/twisted/conch/ssh/keys.py
the0id: It will, and I see what you're saying, but I think it's better how it is.  It's a little faster to process and it fits the convention of the rest of the checks. :)

src/twisted/conch/ssh/keys.py
the0id: Moving the return where it is now fits in better with the block. Each conditional creates a reprstr value, with a single return of that value at the end.

src/twisted/conch/ssh/keys.py
the0id: I think it's better to leave this all caps because the rest of the return strings are in all caps.

the0id:
@alex, @rodrigc thanks for the review.  I've made some of the changes requested, and gave my reasons for not changing the others.

@rodrigc thanks for the head's up about moving the file to newsfragments.  @markrwilliams helped me out with the same thing on another PR. :)

src/twisted/conch/ssh/keys.py
the0id: Good catch. :+1:

the0id:
The last two times appveyor has run two of the tests fail because openssh can't be installed.  

```
Build started
git clone -q https://github.com/twisted/twisted.git C:\projects\twisted
git fetch -q origin +refs/pull/644/merge:
git checkout -qf FETCH_HEAD
Restoring build cache
Cache 'C:\Users\appveyor\AppData\Local\pip\Cache' - Restored
Running Install scripts
choco install openssh -confirm
Chocolatey v0.10.7
Installing the following packages:
openssh
By installing you accept licenses for the packages.
Progress: Downloading openssh 0.0.18.0... 45%openssh not installed. An error occurred during installation:
 The operation has timed out.
openssh package files install completed. Performing other installation steps.
The install of openssh was NOT successful.
openssh not installed. An error occurred during installation:
 The operation has timed out.
Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures
 - openssh (exited 1) - openssh not installed. An error occurred during installation:
 The operation has timed out.
Command exited with code 1
```

rodrigc:
See:
https://twistedmatrix.com/pipermail/twisted-python/2017-August/031596.html

the0id:
@rodrigc Thanks, I missed that message.

markrwilliams:
@the0id You put a lot of effort into this PR - thanks for that!

However, given that I'm writing from a future where [`cryptography` supports curve25519](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519), I think this PR should be closed and a new one opened that uses `cryptography` instead of `pynacl`.  That would eliminate the need to call into a private API.

src/twisted/conch/ssh/keys.py
```diff
@@ -404,57 +427,66 @@ def _fromString_PRIVATE_OPENSSH(cls, data, passphrase):
...
-                ).private_key(backend=default_backend())
-            )
+        # ED25519 Keys just say "OPENSSH KEY".
+        # berDecoder doesn't seem to support ed25519 yet.
+        if kind == b'OPENSSH':
+            # OpenSSH keeps ed25519 keys in the openssh-key-v1 format.
+            # One file can have multiple keys.
+            # For now just grab the last private key.
+            privKey = common.getNS(
+                keyData[keyData.rindex(b'25519') + 5:], 2)[1][:32]
```
cjwatson, line 437: This isn't very accurate parsing; it should parse the documented private key protocol properly.

I've just opened #1053 to support the new private key format.  Assuming that lands, I think this PR should be reworked on top of that, since it will then just need to introduce a new key type rather than also introducing a new format.

cjwatson:
> However, given that I'm writing from a future where [`cryptography` supports curve25519](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519), I think this PR should be closed and a new one opened that uses `cryptography` instead of `pynacl`. That would eliminate the need to call into a private API.

So, um.  Current `cryptography` supports `x25519`, but it doesn't in fact support `ed25519`.  There's some [work in progress](https://github.com/pyca/cryptography/pull/4114) for that.  I tried rewriting this PR on top of that branch and #1053, but I quickly ran into `ed25519.Ed25519PublicKey.from_public_bytes` apparently not working properly, so I'm stalled on that for now until I get feedback from the author of that `cryptography` branch.

Even aside from that, though, this as-yet-unlanded `cryptography` support requires OpenSSL >= 1.1.1, and that's going to be quite a pain for me to deploy in the environments where I need to use this Conch Ed25519 support.  Given that it isn't in Ubuntu 18.04, realistically it means I'm likely to be stuck with complicated and maintenance-intensive backports until at least 2020; I'd really prefer to be running OpenSSL libraries that our security team is maintaining, so I don't want to have to maintain an OpenSSL backport if I can possibly avoid it, and nor do I want to use statically-linked wheels.

I'm certainly not arguing with `cryptography` being the right approach long-term.  But particularly given the likely OpenSSL deployment issues, could we perhaps revisit sticking with `pynacl` for now?  It wouldn't need to be permanent.

cjwatson:
> I'm certainly not arguing with `cryptography` being the right approach long-term. But particularly given the likely OpenSSL deployment issues, could we perhaps revisit sticking with `pynacl` for now? It wouldn't need to be permanent.

Refinement: we could bring up this support with `cryptography` once the support there has landed, but then also add support for using `pynacl` as an alternative; I think this would basically just be a matter of providing shim versions of `Ed25519PublicKey` and `Ed25519PrivateKey`.  That way, we'd have mostly modern code, and when the time comes that we can reasonably assume OpenSSL 1.1.1 everywhere it'll be easy to drop the `pynacl`-based shims.

glyph:
> Refinement: we could bring up this support with `cryptography` once the support there has landed, but then also add support for using `pynacl` as an alternative; I think this would basically just be a matter of providing shim versions of `Ed25519PublicKey` and `Ed25519PrivateKey`. That way, we'd have mostly modern code, and when the time comes that we can reasonably assume OpenSSL 1.1.1 everywhere it'll be easy to drop the `pynacl`-based shims.

This sounds great - are you planning to proceed with this strategy?

cjwatson:
> This sounds great - are you planning to proceed with this strategy?

Probably yes, but I think the best approach is going to be to land a "clean" version first using only `cryptography`, and then add the `pynacl` (or whatever) approach if it still seems to be necessary at that point.  This work has taken long enough that it's possible that any such compatibility work might be overtaken by events: for example, Ubuntu 18.04 has now been updated to OpenSSL 1.1.1, although not yet to 1.1.1b which is required by the Ed25519 support in `cryptography`.

When this was last active I was blocked on work landing in `cryptography`.  That's done now as of 2.6, so I've belatedly picked this up again.  [#9681](https://twistedmatrix.com/trac/ticket/9681) and [#9682](https://twistedmatrix.com/trac/ticket/9682) are the next steps; after that I'll push a PR for [#9683](https://twistedmatrix.com/trac/ticket/9683), and then I'll have probably two more PRs to add curve25519-sha256 key exchange support and Ed25519 key support respectively.  Of course I'll credit @the0id where appropriate in places where I borrowed from them.

glyph:
@cjwatson This branch is *relatively* recent so I'm reluctant to close the PR, but also it appears to be a mess of conflicts.  Is it still relevant, or has the useful code been extracted to other places?

cjwatson:
@glyph I am still working on this, but I don't think it will involve this PR directly, except as a reference, so closing it wouldn't cause a problem for me.  I've extracted the useful parts and will of course credit the submitter of this PR where appropriate as well.

My work has been going very slowly partly because the review cycle is very long (not whining, I understand how things go) and GitHub doesn't support managing chains of PRs in a sensible way, so I haven't been able to make my progress visible until each PR in the stack is merged.

The current sequence is as follows.  Each step depends on the one before it.  I have all of the patches for all of these prepared locally and ready to go, aside from minor tidying up like rebasing onto whatever trunk revision is current and adjusting commit messages.

 * [#9682](https://twistedmatrix.com/trac/ticket/9682) (PR #1172, awaiting review since the end of July)
 * [#9683](https://twistedmatrix.com/trac/ticket/9683)
 * Add `curve25519-sha256` key exchange support
 * Add support for Ed25519 SSH keys

glyph:
OK, the conflicts look pretty gnarly so yeah, let's close it :).  Thanks for sequencing the work this way.

PR #645, opened 2016-12-23
[#8962] generate test certificate with SANs

glyph:
https://tm.tl/8962


derwolfe (approved):
Thanks @glyph - I don't see anything that looks like it needs to change. Please feel free to merge at your convenience.

PR #646, opened 2016-12-24
[#7879] Fix documentation for HTTP client to, among other things, mention that Agent does HTTPS verification.

glyph:


docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-``Agent`` 's constructor takes an optional second argument, a context factory.
-Generally you don't want to create the context factory yourself: it's easy to get wrong.
-Instead, Twisted provides helpers for building correctly configured ones.
-For example, you can use :api:`twisted.internet.ssl.optionsForClientTLS <optionsForClientTLS>`, which takes care of all hostname certificate verification for you.
+Everything you've read so far applies whether the scheme of the request URI is *HTTP* or *HTTPS* .
 
+By default, since version 15.0.0 (if you're using an earlier version, please upgrade, as this is a critical security feature!), Twisted will automatically validate ``https`` URLs against your platform's trust store to the best of its ability.
+Be sure to ``pip install twisted[tls]`` in order to get all the dependencies necessary to do TLS properly, but beyond that you should not need to do much.
```
exarkun: This is a bit hand-wavey.  "Should not"?  When might this not hold?  "do much"?  Like what?

I think it would be better to just refer to installation docs here, if you think folks reading this document will need help getting a working installation of Twisted.

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+Customizing your HTTPS Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-``Agent`` 's constructor takes an optional second argument, a context factory.
-Generally you don't want to create the context factory yourself: it's easy to get wrong.
-Instead, Twisted provides helpers for building correctly configured ones.
-For example, you can use :api:`twisted.internet.ssl.optionsForClientTLS <optionsForClientTLS>`, which takes care of all hostname certificate verification for you.
+Everything you've read so far applies whether the scheme of the request URI is *HTTP* or *HTTPS* .
 
+By default, since version 15.0.0 (if you're using an earlier version, please upgrade, as this is a critical security feature!), Twisted will automatically validate ``https`` URLs against your platform's trust store to the best of its ability.
```
exarkun: I think the warning here will read better if it's pulled out of this parenthetical.  We probably have some kind of `.. warning` thing?  I'd try sticking it in such a thing at the top of this HTTPS section.

exarkun: "By default" and "automatically" feel somewhat redundant.  I'd say "Twisted will automatically validate ..." or "By default, Twisted will validate ..."

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+Be sure to ``pip install twisted[tls]`` in order to get all the dependencies necessary to do TLS properly, but beyond that you should not need to do much.
 
+However, for some uses, you may need to customize Agent's use of HTTPS; for example, to provide a client certificate, or to use a custom certificate authority for an internal network.
 
+Here, we're just going to show you how to inject the relevant TLS configuration into an Agent, so we'll use the simplest possible example, rather than a more useful but complex one.
 
+``Agent``'s constructor takes an optional second argument, which allows you to customize its behavior with respect to HTTPS.
```
exarkun, line 357: "... second argument which allows ..." (no comma)

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+The object passed here must implement the :api:`twisted.web.iweb.IPolicyForHTTPS` interface.
```
exarkun: not "implement".  "provide", yes?

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
 
-
-Here's an example which shows how to use ``Agent`` to request
-an *HTTPS* URL with no certificate verification.
-
-
-    
-
-
+So, here's our example.
```
exarkun: This line doesn't add a lot (and the example doesn't start for two more sentences :smile: ).

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+Using the very helpful ``badssl.com`` web API, we will construct a requst that fails to validate, because the certificate has the wrong hostname in it.
```
exarkun: "... fails to validate because the ..." (no comma)

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+The following Python program should produce a verification error, when run as ``python example.py https://wrong.host.badssl.com/``.
```
exarkun: "... verification error when run ..." (no comma)

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+        agent = Agent(reactor)
+        requested = agent.request(b"GET", sys.argv[1])
+        def gotResponse(response):
+            print(response.code)
+        def noResponse(failure):
+            failure.trap(ResponseFailed)
+            print(failure.value.reasons[0].getTraceback())
+        return requested.addCallbacks(gotResponse, noResponse)
 
+This is because the certificate returned by ``wrong.host.badssl.com`` is actually for ``badssl.com``, and the hostnames don't match.
```
exarkun: Might be worth reminding the reader (having just digested that Python program) what "This" is (eg "The verification error is caused by ...").

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
-        contextFactory = optionsForClientTLS(u"https://example.com/")
-        agent = Agent(reactor, contextFactory)
-        d = agent.request("GET", "https://example.com/")
-        d.addCallbacks(display, err)
-        d.addCallback(lambda ignored: reactor.stop())
-        reactor.run()
-    
-    if __name__ == "__main__":
-        main()
+    import sys
```
exarkun, line 365: Any automatic documentation testing facilities available yet?

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+        requested = agent.request(b"GET", sys.argv[1])
+        def gotResponse(response):
+            print(response.code)
+        def noResponse(failure):
+            failure.trap(ResponseFailed)
+            print(failure.value.reasons[0].getTraceback())
+        return requested.addCallbacks(gotResponse, noResponse)
 
+This is because the certificate returned by ``wrong.host.badssl.com`` is actually for ``badssl.com``, and the hostnames don't match.
+In our pretend scenario, we want to construct an ``Agent`` that validates HTTPS certificates normally, _except_ for this one host, where it ignores the hostname mismatch and supplies the correct hostname manually, to work around this problem.
```
exarkun: Probably works better as two sentences with the split right around "where it ..." (and with some minor rewording).

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+In order to do that, we will supply our own policy that creates a per - netloc–i.e. per host+port combination – TLS configuration, like so:
```
exarkun: Ouch on the `-`s here.  I sorta get what you're saying but figuring it out seems likely to derail most people from the content.  How about "per-netloc (per-host and port)"?  Or "for each netloc (i.e., each host+port combination), we will supply ..."?

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+    from twisted.web.client import Agent, ResponseFailed, BrowserLikePolicyForHTTPS
+
+    @implementer(IPolicyForHTTPS)
+    class OneHostnameWorkaroundPolicy(object):
+        def __init__(self):
+            self._normalPolicy = BrowserLikePolicyForHTTPS()
+        def creatorForNetloc(self, hostname, port):
+            if hostname == b"wrong.host.badssl.com":
+                return optionsForClientTLS(u"badssl.com")
+            else:
```
exarkun: You can chop a useless line off the example by deleting this else and just making the other case the fall-through.

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+    from twisted.web.iweb import IPolicyForHTTPS
+    from twisted.web.client import Agent, ResponseFailed, BrowserLikePolicyForHTTPS
+
+    @implementer(IPolicyForHTTPS)
+    class OneHostnameWorkaroundPolicy(object):
+        def __init__(self):
+            self._normalPolicy = BrowserLikePolicyForHTTPS()
+        def creatorForNetloc(self, hostname, port):
+            if hostname == b"wrong.host.badssl.com":
+                return optionsForClientTLS(u"badssl.com")
```
exarkun: Would `self._normalPolicy.creatorForNetloc(u"badssl.com", port)` work here?  Are you making a point by using `optionsForClientTLS` instead?

What about the reverse question about the other case?

ie, can both branches call the same function and just vary the arguments?  That might be clearer and you can remove one concept from the example of you make these consistent.

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+            else:
+                return self._normalPolicy.creatorForNetloc(hostname, port)
+
+    @react
```
exarkun, line 408: btw, don't think I didn't notice this.

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
-
-
-    
-
-
+So, here's our example.
+Using the very helpful ``badssl.com`` web API, we will construct a requst that fails to validate, because the certificate has the wrong hostname in it.
+The following Python program should produce a verification error, when run as ``python example.py https://wrong.host.badssl.com/``.
 
 .. code-block:: python
```
exarkun, line 363: Does this make this a Python 2 code block or is it unspecified whether this is Python 2 or Python 3 code?  Is it meant to be polyglot?

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+        def creatorForNetloc(self, hostname, port):
+            if hostname == b"wrong.host.badssl.com":
+                return optionsForClientTLS(u"badssl.com")
+            else:
+                return self._normalPolicy.creatorForNetloc(hostname, port)
+
+    @react
+    def main(reactor):
+        agent = Agent(reactor, OneHostnameWorkaroundPolicy())
+        requested = agent.request(b"GET", sys.argv[1])
```
exarkun: I asked about Python 2 vs Python 3 because I wonder what type `sys.argv[1]` is.

docs/web/howto/client.rst
```diff
@@ -331,74 +331,85 @@ This function returns a ``Deferred`` that fires with the body after the request
...
+            print(response.code)
+        def noResponse(failure):
+            failure.trap(ResponseFailed)
+            print(failure.value.reasons[0].getTraceback())
+        return requested.addCallbacks(gotResponse, noResponse)
+
+Now, invoking ``python example.py https://wrong.host.badssl.com/`` will happily give us a ``200`` status code; however, running it with ``https://expired.badssl.com/`` or ``https://self-signed.badssl.com/`` or any of the other error hostnames should still give an error.
+
+Using this TLS policy mechanism, you can customize Agent to use any feature of TLS that Twisted has support for, including the examples given above; client certificates, alternate trust roots, and so on.
+For a more detailed explanation of what options exist for client TLS configuration in Twisted, check out the documentation for the :api:`twisted.internet.ssl.optionsForClientTLS <optionsForClientTLS>` API and the :doc:`Using SSL in Twisted <../../core/howto/ssl>` chapter of this documentation.
```
exarkun, line 427: It might be nice to standardize on a "Further Reading" section in the docs which contains tips like this one (providing users with the benefit of navigational simplicity in accessing the information).

exarkun (approved):
Thanks.  Generally quite good.  A lot of nitpicking follows but it's not *all* grammar hijinx (not that I wish anyone to discount the significance of grammar).

Please address to your satisfaction and merge.

docs/web/howto/client.rst
glyph: I agree; there's no `..warning::` that I can see, so I think I'l use a `..note::` for this.

docs/web/howto/client.rst
glyph: Yeah I'll just delete the whole "but beyond that" bit.

glyph: There _are_ about 45 different ways that getting system trust roots can fail, unfortunately, but the way to address that is not to document all the problems repeatedly, but just to fix them.

docs/web/howto/client.rst
glyph: no :-(

derwolfe: Could you use a `literalinclude` and test the included code?

glyph: @derwolfe Yes, but there's a bunch of infrastructure that needs to be written to make that workable.  It's harder than it used to be now that we use `tox` because you have to discover where the docs are relative to the code, add them to `sys.path` temporarily, and so on.

docs/web/howto/client.rst
glyph: As far as I can tell, your comment here is just backwards.  "which allows you to customize its behavior" is (I believe) a nonrestrictive phrase, the closest analog of which I can find in this list of examples https://www.grammarly.com/blog/comma-before-which/ is "Rob tripped over his shoes, which he had left lying in the middle of the floor" and "Everyone loved Robin’s video, which she had filmed in her garage", both of which are listed as _requiring_ a comma (they don't even seem to think it's optional).  So I'm going to ignore it. 

In the future, if you're going to provide grammar rule feedback like this, can you provide some a citation the author can confirm their understanding of the rule you're trying to suggest adherence to?  I have to admit I'm not _confident_ that your understanding of the rules here is wrong and mine is right, but my searching isn't turning up counterexamples.  If it's a stylebook thing, having a link also simplifies future feedback for other authors, for consistency.

docs/web/howto/client.rst
glyph: Yes. Good catch.

docs/web/howto/client.rst
glyph: Point taken, deleted.

docs/web/howto/client.rst
glyph: Haha. Also, "requst"…

glyph: OK, so once more into the comma breach: http://www.writingwithclarity.com/2012/01/comma-uncertainty-when-to-use-it-before-because/

Here, I'd argue that the "essential" information is that the request fails to validate, and the "wrong hostname" is of interest, but peripheral.  So I'm leaving the comma in.

docs/web/howto/client.rst
glyph: (just to be clear, I originally typed "restrictive" but then corrected it to "nonrestrictive" above)

docs/web/howto/client.rst
glyph: OK! _here_ there should indeed be no comma, the following Python program will _not_ produce an error when run in other ways, so this clause is restrictive and the comma is an error :)

docs/web/howto/client.rst
glyph: Consulting the sphinx docs, apparently this specifically means "python 2": http://www.sphinx-doc.org/en/1.5.1/config.html#confval-highlight_language but my intention is that it be polyglot, because readers of the documentation might always be using either version.

docs/web/howto/client.rst
glyph: Yes, good catch. Ambiguous antecedents are a plague. Starting sentences with a pronoun is a bad habit. This is definitely a tic in my writing I strive to rid myself of... ;-)

docs/web/howto/client.rst
glyph: Agreed.

docs/web/howto/client.rst
glyph: Wow I don't know how the spaces and the dashes got into that peculiar arrangement.  It definitely reads like word salad.  Will fix.

glyph: I think actually a little bit of constructive inaccuracy would be good here. We don't need to talk about the port number, it's clearly supplied to the function if the user wants to do more with it, so "per-hostname" is probably a better way to put it.  (Maybe even not terribly inaccurate, given that we're not actually dispatching on the port number at all.)

docs/web/howto/client.rst
glyph: I think I was trying to make a point (illustrating that you can construct these objects yourself if you want to, pointing out the API that you might want to use is `optionsForClientTLS` to try to steer people away from building their own context factory) but you're right, fewer concepts is better.  The TLS doc already exists and those concepts are explained there.

docs/web/howto/client.rst
glyph: Good thought.

docs/web/howto/client.rst
glyph: I know you are not the biggest fan of decorator abuse :-) but in the case of `twisted.internet.task.react`, I actually think the argument-handling stuff is _by far_ worse than useless (if you want to do any error-handling at all, you need to take `*args` and then pass it to a different function) and using it this way just encourages the user to think of it as "here's how you say where the main function is".

I'd actually like to do a `react`++ which is designed to be used this way, and only runs the function when it's defined in `__main__`, still allowing it to be imported, and maybe do something else with the command-line args.  We should discuss that elsewhere though.

docs/web/howto/client.rst
glyph: Wince.

glyph: OK.  Thanks for pointing this out.  Lucky for us, implicitly, the type of `IAgent.request`'s second argument is actually "ASCII-only bytes", because IRIs officially aren't 8-bit clean.

glyph: … which further means that simply adding an `.encode("ascii")` to the end here manages to make it work with py2 & py3 simultaneously, thanks to `str`'s confusion on py2.

glyph: (and testing the script manually on both just to make sure seems to work)

docs/web/howto/client.rst
glyph: I might want to make that as a separate change.

glyph: Specifically, in this case, I really want to mention `optionsForClientTLS` first, because there's a lot of ways to screw that up, and putting them in a sentence rather than a list makes the order more significant.

glyph:
Hopefully all the buildbots are alive and happy now.

docs/web/howto/client.rst
derwolfe: That makes sense. Thanks 😄!

docs/web/howto/client.rst
exarkun: My reading of the statement is that it is a restrictive clause.  The context of the sentence makes the HTTPS customization the critical information being conveyed here.  Maybe that actually means the sentence should be something more like "You can customize the behavior of Agent with respect to HTTPS by using its optional second argument." and that's the edit I should have suggested rather than just the comma removal (since *technically* "Agent's constructor takes an optional second argument." makes perfect sense on its own).

wrt grammar citations, I've long thought we should officially adhere to a particular style guide.  Then folks would have something to look at to figure this stuff out in advance and there'd be no confusion about where the rules come from (and we really don't want to allow just any random citation as justification since there are conflicting rules).

(no action required, in any case, just wanted to follow-up)

glyph: > My reading of the statement is that it is a restrictive clause. The context of the sentence makes the HTTPS customization the critical information being conveyed here. Maybe that actually means the sentence should be something more like "You can customize the behavior of Agent with respect to HTTPS by using its optional second argument." and that's the edit I should have suggested rather than just the comma removal (since technically "Agent's constructor takes an optional second argument." makes perfect sense on its own).

My understanding of "restrictive" is that, while it is often colloquially described as "provides important information", what it actually means is that it provides _essential_ information, i.e. information without which the meaning would be _different_.  If you can provide the additional information in separate sentences, then the clause, while it might be important, is not restrictive.

To phrase this in terms of the things we're talking about here, in this case the clause would be restrictive:

>> Agent's constructor takes one argument which allows you to customize its behavior with respect to HTTPS.

but non-restrictive in this case:

>> Agent's constructor takes a second argument, which allows you to customize its behavior with respect to HTTPS.

since in the first case if you eliminate the "which" clause, the sentence could legitimately be interpreted as "Agent's constructor takes _only_ one argument", whereas in the second case you could write it as: 

>> Agent's constructor takes a second argument.  It allows you to customize its behavior with respect to HTTPS.

glyph: > wrt grammar citations, I've long thought we should officially adhere to a particular style guide. Then folks would have something to look at to figure this stuff out in advance and there'd be no confusion about where the rules come from (and we really don't want to allow just any random citation as justification since there are conflicting rules).

I definitely agree. The key thing that's stopped me from doing exactly that is that the two style guides I might want to adopt, http://www.apstylebook.com or http://www.chicagomanualofstyle.org/home.html, both require a site-license.

Is there a good freely-available style guide we could adopt?

glyph: > (no action required, in any case, just wanted to follow-up)

Yeah, definitely came across that way, but thanks for making this explicit. No action required for any of my responses either. :)

docs/web/howto/client.rst
meejah: +1, would use "react++" ;)

PR #649, opened 2016-12-29
[#8877] allow DNS resolution over IPv6 with twisted.names

reactormonk:
This commit lets twisted query ipv6 DNS servers by using a ipv6 socket by
default and converting ipv4 nameservers to ipv6.

https://tm.tl/8877

### DIFF
--- a/src/twisted/names/client.py
+++ b/src/twisted/names/client.py
@@ -30,6 +30,7 @@ from twisted.internet import error, defer, interfaces, protocol
 from twisted.python import log, failure
 from twisted.names import (
     dns, common, resolve, cache, root, hosts as hostsModule)
+from twisted.internet.abstract import isIPv6Address
 
 
 
@@ -211,7 +212,7 @@ class Resolver(common.ResolverBase):
             return self.dynServers[self.index - serverL]
 
 
-    def _connectedProtocol(self):
+    def _connectedProtocol(self, interface=''):
         """
         Return a new L{DNSDatagramProtocol} bound to a randomly selected port
         number.
@@ -219,7 +220,8 @@ class Resolver(common.ResolverBase):
         proto = dns.DNSDatagramProtocol(self, reactor=self._reactor)
         while True:
             try:
-                self._reactor.listenUDP(dns.randomSource(), proto)
+                self._reactor.listenUDP(dns.randomSource(), proto,
+                                        interface=interface)
             except error.CannotListenError:
                 pass
             else:
@@ -260,7 +262,10 @@ class Resolver(common.ResolverBase):
         @return: A L{Deferred} which will be called back with the result of the
             query.
         """
-        protocol = self._connectedProtocol()
+        if isIPv6Address(args[0][0]):
+            protocol = self._connectedProtocol(interface='::')
+        else:
+            protocol = self._connectedProtocol()
         d = protocol.query(*args)
         def cbQueried(result):
             protocol.transport.stopListening()
--- a/src/twisted/names/test/test_client.py
+++ b/src/twisted/names/test/test_client.py
@@ -24,7 +24,7 @@ from twisted.names.error import DNSQueryTimeoutError
 from twisted.names.common import ResolverBase
 
 from twisted.names.test.test_hosts import GoodTempPathMixin
-from twisted.names.test.test_util import MemoryReactor
+from twisted.names.test import test_util
 
 from twisted.test import proto_helpers
 
@@ -332,11 +332,11 @@ class ResolverTests(unittest.TestCase):
         """
         protocol = StubDNSDatagramProtocol()
 
-        servers = [object(), object()]
-        dynServers = [object(), object()]
+        servers = [("::1", 53), ("::2", 53)]
+        dynServers = [("::3", 53), ("::4", 53)]
         resolver = client.Resolver(servers=servers)
         resolver.dynServers = dynServers
-        resolver._connectedProtocol = lambda: protocol
+        resolver._connectedProtocol = lambda interface: protocol
 
         expectedResult = object()
         queryResult = resolver.queryUDP(None)
@@ -499,7 +499,7 @@ class ResolverTests(unittest.TestCase):
         L{client.Resolver._connectedProtocol} should pass that reactor
         to L{twisted.names.dns.DNSDatagramProtocol}.
         """
-        reactor = MemoryReactor()
+        reactor = test_util.MemoryReactor()
         resolver = client.Resolver(resolv=self.mktemp(), reactor=reactor)
         proto = resolver._connectedProtocol()
         self.assertIs(proto._reactor, reactor)
@@ -528,6 +528,20 @@ class ResolverTests(unittest.TestCase):
         self.assertEqual(len(set(protocols)), 2)
 
 
+    def test_ipv6Resolver(self):
+        """
+        If the resolver is ipv6, open a ipv6 port.
+        """
+
+        fake = test_util.MemoryReactor()
+        resolver = client.Resolver(servers=[('::1', 53)],
+                                   reactor=fake)
+        resolver.query(dns.Query(b'foo.example.com'))
+        [(proto, transport)] = fake.udpPorts.items()
+        interface = transport.getHost().host
+        self.assertEqual("::", interface)
+
+
     def test_disallowedPort(self):
         """
         If a port number is initially selected which cannot be bound, the
@@ -536,7 +550,7 @@ class ResolverTests(unittest.TestCase):
         ports = []
 
         class FakeReactor(object):
-            def listenUDP(self, port, *args):
+            def listenUDP(self, port, *args, **kwargs):
                 ports.append(port)
                 if len(ports) == 1:
                     raise CannotListenError(None, port, None)
--- a/src/twisted/topfiles/8877.bugfix
+++ b/src/twisted/topfiles/8877.bugfix
@@ -1,0 +1,1 @@
+twisted.names.client.Resolver can now resolve names with IPv6 DNS servers.


rodrigc:
Please fix style issues pointed out by twistedchecker:
https://travis-ci.org/twisted/twisted/jobs/187467971

reactormonk:
The PR causes a lot of other tests to fail (probably mocking-based etc.), so I wanted to know if this is a good way to fix this bug before diving into them.

I also have no idea what happens if you somehow lock your system out of ipv6 for whatever misguided reason.

rodrigc:
In https://travis-ci.org/twisted/twisted/jobs/187508712 , twisted.internet.test.test_resolver.HostnameResolutionTests.test_ipv6Resolver is failing because getHostByName()
cannot be found.

rodrigc:
You need a topfile news fragment, as specified here: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

src/twisted/names/client.py
```diff
@@ -187,7 +190,7 @@ def parseConfig(self, resolvConf):
                 self.search = L.split()[1:]
                 self.domain = None
         if not servers:
-            servers.append(('127.0.0.1', dns.PORT))
+            servers.append(('::1', dns.PORT))
```
rodrigc: It looks like you are changing things to append the IPv6 address for localhost instead of the IPv4 address for localhost.
This looks like a backwards incompatible change, which seems wrong.

src/twisted/names/client.py
```diff
@@ -174,7 +174,10 @@ def parseConfig(self, resolvConf):
         for L in resolvConf:
             L = L.strip()
             if L.startswith(b'nameserver'):
-                resolver = (nativeString(L.split()[1]), dns.PORT)
+                parsedString = L.split()[1]
+                if b'.' in parsedString:
+                    parsedString = b'::' + parsedString
```
rodrigc: So it looks like if you see a line like:

nameserver 128.0.0.1

you are changing the address to

::128.0.0.1

This seems wrong also, and would explain why some of the tests are failing.

rodrigc (changes requested):
For your patch, using an IPv6 socket to query an IPv6 DNS server is OK.
But the parts of your patch where you convert IPv4 nameservers to IPv6
is not OK, and will cause lots of things to break.

src/twisted/names/client.py
reactormonk: That should work as expected, you can use a ipv6 socket to connect to an ipv4 address this way.

rodrigc:
You are assuming that on all platforms it is possible to connect to an IPv4 address with
an IPv6 socket, and it will work.  I think that assumption is wrong, even though it may be legal according to the spec.

reactormonk:
The other idea would be to have two connections open, one for ipv4 and one for ipv6, but that gets a whole lot more complicated, from what I understand about how the twisted framework handles async, because you have to coordinate them somehow. Not sure if I could implement that correctly.

I could write a simpler solution that uses ipv4 if that's the first nameserver is ipv4, and assumes a correct ipv6 implementation on first nameserver being ipv6.

reactormonk:
According to https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses the correct format would be `::ffff:`, but that won't work on OpenBSD or old Windows machines.

reactormonk:
Ok, a different solution. Two different resolvers, one for ipv4 and one for ipv6. Just not sure how to write a test for it correctly.

glyph:
@reactormonk thanks for working on this!

src/twisted/names/client.py
```diff
@@ -439,6 +451,43 @@ def _cbLookupZone(self, result, connector):
 
 
 
+class IPv6Resolver(Resolver):
+    """
+    Selects only IPv6 resolv.conf entries.
+    """
+    def maybeAddNameserver(self, nameserver, servers):
+        resolver = (nativeString(nameserver), dns.PORT)
+        if b':' in nameserver:
```
kyrias: This is sort of an ugly way to check whether it's a v4 or v6 address, but it probably doesn't matter since I don't know of a single system that lets you specify an alternate port in `resolv.conf`.  The "nicest" way IMO would be to use the `ipaddress` package which is in the stdlib since 3.3, but requires an additional external dependency for py2, so I guess that's probably out.

glyph: [`isIPAddress`](https://twistedmatrix.com/documents/16.6.0/api/twisted.internet.abstract.html#isIPAddress) and [`isIPv6Address`](https://twistedmatrix.com/documents/16.6.0/api/twisted.internet.abstract.html#isIPv6Address) might be helpful here.

glyph: Adding new dependencies isn't beyond the pale, either…

kyrias: 3.3 added https://docs.python.org/3/library/ipaddress.html which allows you to parse an IP address, returning either a `IPv4Address` or `IPv6Address` object.  For 2.7-3.2 it has been backported and put in PyPI though: <https://pypi.python.org/pypi/ipaddress>, so might make sense to have those functions use that, to offload it from twisted.

glyph: For something this simple I think the functions already in Twisted should be fine.  Separately, it might be a good idea to replace their _implementations_ with the `ipaddress` module.

kyrias: Yeah that's what I meant, sorry for the confusion.

reactormonk:
What's currently tripping me up is https://github.com/twisted/twisted/blob/trunk/src/twisted/names/client.py#L190 - otherwise I could simply check if there's no configured servers and continue on to the next resolver. Why does that exist, and can I delete it?

Relevant commit: https://github.com/twisted/twisted/commit/44a16d7e2c

It works with it, but it would require a (localhost) roundtrip each time in case you don't have IPv6 nameservers. I don't think that's a very good tradeoff.

glyph:
@reactormonk - wow, that's... weird. I suspect that we may have been using this on twistedmatrix.com (which, for dogfooding purposes, runs a resolver itself) and someone thought that was useful to add "just in case".

As far as I know there's no OS that runs a recursive resolver over the real actual DNS port; I believe that this would be reasonable to just remove.

glyph:
(BTW, if you hit 'y' in github on a file, it will redirect you to the full revision, so that links persistently refer to the same code.  For example, your link would be https://github.com/twisted/twisted/blob/88a71947b41af21065f7fdfa56a3ffe22a7f5c57/src/twisted/names/client.py#L190 - which is helpful when viewing the conversation after those files have been modified.)

mithrandi:
This is the behaviour documented by glibc in resolv.conf(5) (emphasis mine):

Internet  address  of  a  name  server  that  the  resolver should query, either an IPv4 address (in dot notation), or an IPv6 address in colon (and possibly dot) notation as per RFC 2373.  Up to MAXNS (currently 3, see <resolv.h>) name servers may be listed, one per keyword.  If there are multiple servers, the resolver library queries them in the order listed.  **If no nameserver entries are present, the default is  to use the name server on  the local machine.**  (The algorithm used is to try a name server, and if the query times out, try the next, until out of name servers, then repeat trying all the name servers until a maximum number of retries are made.)

glyph:
Thanks @mithrandi .

OK, @reactormonk - so there's a reason for this behavior.  The next question is: why is it tripping you up?  It seems like the query should just time out as expected if no nameserver is running on localhost.

exarkun:
As discussed on IRC, it seems like a lot of these changes aren't necessary.  The problem arises from trying to send UDP traffic to an address with a different address family than the socket is bound to.  An easy fix for this, since each query (and therefore each write) uses its own UDP socket, is to teach `_connectedProtocol` to bind to an address family that matches the address family of the nameserver address about to be queried.

reactormonk:
Yes, I did some overengineering. I'll take another look in a few hours which should solve the problem in an easier manner, where above change isn't required.

reactormonk:
Where would I put integration tests? Or should I just mock-test the part of the code I changed.

src/twisted/names/test/test_client.py
```diff
@@ -528,6 +528,24 @@ def query(self, address, query, timeout=10, id=None):
         self.assertEqual(len(set(protocols)), 2)
 
 
+    def test_ipv6Resolver(self):
+        """
+        If the resolver is ipv6, open a ipv6 port.
+        """
+        counter = []
+
+        class FakeReactor(MemoryReactor):
```
markrwilliams: This particular tests has to subclass this particular `MemoryReactor` and call into its `listenUDP`, but `test_disallowedPort` does not.  That's because `test_disallowedPort` needs to make [this condition](https://github.com/twisted/twisted/blob/63f7b78854198fbc25e779b6e852b7a5dba2a6fd/src/twisted/names/dns.py#L2911) true to hit the [`except` block](https://github.com/twisted/twisted/blob/63f7b78854198fbc25e779b6e852b7a5dba2a6fd/src/twisted/names/dns.py#L2915-L2916).

_This_ test, however, exercises the _client_'s [`listenUDP`](https://github.com/twisted/twisted/pull/649/files#diff-7a7cb05a7b26a752daf0078ea416302fR215), so it needs to ensure the transport has been set on the `DNSDatagramProtocol`.  Doing that fails the condition linked to above, so that `startListening` is not called. `test_util.MemoryReactor.listenUDP` does that  by calling [`makeConnection` on its protocol](https://github.com/twisted/twisted/blob/63f7b78854198fbc25e779b6e852b7a5dba2a6fd/src/twisted/names/test/test_util.py#L134)

I have three opinions after writing this story:

1. I'd suggest importing `test_util` so that the superclass name would read `test_util.MemoryReactor`.  We have two `MemoryReactor`s running around and since it's critical that the one inside `twisted.names` be used here, it'd be nice to make its source explicit.
2. You might consider returning the result of `listenUDP` just to stay consistent with `test_util.MemoryReactor.listenUDP`'s interface.
3. Maybe a comment or docstring explaining why this particular `MemoryReactor` is critical to the test?

markrwilliams (commented):
A few comments on making the tests clearer.

reactormonk:
No idea why the codecov complains here.

rodrigc:
codecov is complaining because on this line https://codecov.io/gh/twisted/twisted/compare/63f7b78854198fbc25e779b6e852b7a5dba2a6fd...2b47531af81531c881a88e082cb0ad2f6a00dab6#7372632F747769737465642F6E616D65732F746573742F746573745F636C69656E742E7079-541

bot cases where the if statement can be true or false are not being tested.

reactormonk:
Eh yes, that's the intent. Should I add a random additional call to satisfy codecov?

markrwilliams:
What about:

```python
     def test_ipv6Resolver(self):
        """
        If the resolver is ipv6, open a ipv6 port.
        """
        interfaces = []

        class FakeReactor(MemoryReactor):
            def listenUDP(self, port, *args, **kwargs):
                interfaces.append(kwargs.get('interface'))
                super(FakeReactor, self).listenUDP(port, *args, **kwargs)

        resolver = client.Resolver(servers=[('::1', 53)],
                                   reactor=FakeReactor())
        resolver.query(dns.Query(b'foo.example.com'))
        self.assertEqual(interfaces, ['::'])
```

This is nice because if `listenUDP` is called too many times you'll see what it was called with.

reactormonk:
Nice idea. Also helps in case anyone wonders why the `FakeReactor` actually inherits.

src/twisted/names/client.py
```diff
@@ -211,15 +212,19 @@ def pickServer(self):
             return self.dynServers[self.index - serverL]
 
 
-    def _connectedProtocol(self):
+    def _connectedProtocol(self, ipv6=False):
```
exarkun: Optional: make it `interface` and then just pass it through to `listenUDP` as the interface argument - no conditional required here.

src/twisted/names/test/test_client.py
```diff
@@ -24,7 +24,7 @@
 from twisted.names.common import ResolverBase
 
 from twisted.names.test.test_hosts import GoodTempPathMixin
-from twisted.names.test.test_util import MemoryReactor
+from twisted.names.test import test_util
```
exarkun, line 27: Looks like this change isn't needed.  Only `MemoryReactor` is being used.

src/twisted/names/test/test_client.py
```diff
@@ -528,6 +528,25 @@ def query(self, address, query, timeout=10, id=None):
...
+        # DNSDatagramProtocol will be called.
+        class FakeReactor(test_util.MemoryReactor):
+            def listenUDP(self, port, *args, **kwargs):
+                interfaces.append(kwargs.get('interface'))
+                super(FakeReactor, self).listenUDP(port, *args, **kwargs)
+
+        resolver = client.Resolver(servers=[('::1', 53)],
+                                   reactor=FakeReactor())
+        resolver.query(dns.Query(b'foo.example.com'))
+        self.assertEqual(interfaces, ['::'])
```
exarkun: `MemoryReactor` will give you back the information you need without needing this subclass.  There should be one listening udp port after the query and you can get it from `fake.udpPorts`.  The value is a `MemoryDatagramTransport` with a `getHost` method that describes where it's listening.

So:

`[(proto, transport)]  = fake.udpPorts.items(); interface = transport.getHost().host; assertEqual("::", interface)` - more or less.  `MemoryDatagramTransport` doesn't actually _notice_ it's been told to listen on IPv6, of course, but I don't think this matters in this case (it would be a nice thing to fix when merging `test_util.MemoryReactor` into `proto_helpers.MemoryReactor`).

src/twisted/topfiles/8877.bugfix
```diff
@@ -0,0 +1 @@
+Twisted now can resolve names with IPv6 DNS servers
```
exarkun: The style guide suggests the subject of the news fragment should be a FQPN.  So I'd say "twisted.names.client.Resolver can now ...".  Also, the fragment should be a complete sentence - so it should end with a `.`.

exarkun (approved):
Thanks!  Essentially looks good.  A few minor comments, the most significant about the test and not the implementation.

Please address and, when CI is happy, merge.  Thanks.

src/twisted/names/test/test_client.py
markrwilliams: That's true, but there are *two* `MemoryReactor`s in Twisted.  This makes it clear which one the test uses.

wsanchez:
Branching repo for Buildbot: 8877-getHostByName-ipv6

glyph:
@exarkun I don't think @reactormonk has commit? :)

PR #650, opened 2016-12-29
[#8969] Fix for twisted.internet.sendmsg segfault on Linux Python 2

exvito:
Trac ticket at https://twistedmatrix.com/trac/ticket/8969

### DIFF
--- a/src/twisted/python/_sendmsg.c
+++ b/src/twisted/python/_sendmsg.c
@@ -291,7 +291,7 @@ static PyObject *sendmsg_sendmsg(PyObject *self, PyObject *args, PyObject *keywd
 
         /* Unpack the tuples into the control message. */
         struct cmsghdr *control_message = CMSG_FIRSTHDR(&message_header);
-        while ( (item = PyIter_Next(iterator)) ) {
+        while ( (item = PyIter_Next(iterator)) && control_message!=NULL ) {
             int type, level;
             Py_ssize_t data_len;
             size_t data_size;
--- a/src/twisted/python/_sendmsg.c
+++ b/src/twisted/python/_sendmsg.c
@@ -278,6 +278,14 @@ static PyObject *sendmsg_sendmsg(PyObject *self, PyObject *args, PyObject *keywd
                 PyErr_NoMemory();
                 goto finished;
             }
+            /* From Python 3.5.2 socketmodule.c:3891:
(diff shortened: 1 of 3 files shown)


rodrigc:
Does this fix:
https://twistedmatrix.com/trac/ticket/5649
?

exvito:
@rodrigc,

I can't reproduce the behaviour in https://twistedmatrix.com/trac/ticket/5649 here. With that code, from trunk, I get a segfault and not a SIGABRT. With this PR I get an exception with errno=ENOTSOCK, which seems to make sense, given STDIN is being used.

PS: Quickly tested on an Ubuntu 16.04 virtual machine.

exvito:
Dear reviewers,

I need guidance to find a way of ensuring 100% diff coverage.

Given that the test for "does not segfault" is calling and disregarding the result or acceptable exceptions like socket.error, in this case I wrapped the call in a try/except block that travis is not fully covering. The except block is only hit on Mac OS, confirmed at least 10.9 and 10.10.

Any pointers? Maybe some other strategy is needed?...
Thanks in advance.

rodrigc:
Moved to https://github.com/twisted/twisted/pull/651

src/twisted/python/_sendmsg.c
```diff
@@ -278,6 +278,14 @@ static PyObject *sendmsg_sendmsg(PyObject *self, PyObject *args, PyObject *keywd
...
                 goto finished;
             }
+            /* From Python 3.5.2 socketmodule.c:3891:
+               Need to zero out the buffer as a workaround for glibc's
+               CMSG_NXTHDR() implementation.  After getting the pointer to
+               the next header, it checks its (uninitialized) cmsg_len
+               member to see if the "message" fits in the buffer, and
+               returns NULL if it doesn't.  Zero-filling the buffer
+               ensures that this doesn't happen. */
+            memset(message_header.msg_control, 0, all_data_len);
```
exarkun, line 288: `PyMem_Calloc`?  Or  even `PyMem_RawCalloc`.

exvito: From what I can tell from the Python docs, those are not available on Python 2, for which _sendmsg.c is needed. Am I mistaken?

rodrigc: PyMem_Calloc() is new in Python 3.5, and apparently not in Python 2.

PR #651, opened 2016-12-29
[#8969] Fix for twisted.internet.sendmsg segfault on Linux Python 2

rodrigc:
https://twistedmatrix.com/trac/ticket/8969


exvito:
@rodrigc, any particular reason for moving #650 here?

Also, any recommendation on how to achieve 100% diff coverage like I commented in #650?

Thanks in advance.

rodrigc:
https://github.com/twisted/twisted/pull/650 was not running on all the buildbots, while https://github.com/twisted/twisted/pull/651 was running on all the buildbots, including the OS X buildbots, thus getting us more coverage.  Looks like codecov is getting 100% coverage, so I'm OK with this.

PR #145, opened 2016-06-04
addTimeout function on defer.Deferred

cyli:
Original ticket:  https://twistedmatrix.com/trac/ticket/5786

PR #653, opened 2017-01-01
Convert code to use twisted.internet.error.TimeoutError instead of twisted.internet.defer.TimeoutError

rodrigc:
http://twistedmatrix.com/trac/ticket/8970

### DIFF
--- a/src/twisted/internet/interfaces.py
+++ b/src/twisted/internet/interfaces.py
@@ -77,7 +77,7 @@ class IResolverSimple(Interface):
             preference to A records.  If there are multiple records of
             the type to be returned, one will be selected at random.
 
-        @raise twisted.internet.defer.TimeoutError: Raised
+        @raise twisted.internet.error.TimeoutError: Raised
             (asynchronously) if the name cannot be resolved within the
             specified timeout period.
         """
--- a/src/twisted/names/client.py
+++ b/src/twisted/names/client.py
@@ -281,7 +281,7 @@ class Resolver(common.ResolverBase):
         When the last timeout expires, the query is considered failed.
 
         @rtype: C{Deferred}
-        @raise C{twisted.internet.defer.TimeoutError}: When the query times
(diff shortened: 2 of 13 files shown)


exarkun (changes requested):
Doesn't this break any application code that tried to handle a `twisted.internet.defer.TimeoutError` from these APIs?  Unless I'm missing something major, that would seem to preclude merging this set of changes.

That's not to say there's no way to move code away from defer.TimeoutError to something else - but it needs  to be done in a way that gives applications time to migrate before being broken.

For example, introduce an exception which subclasses defer.TimeoutError but has a nicer name which we can all agree should be used in the long term (perhaps twisted.internet.error.TimeoutError :wink: ).  Then deprecate defer.TimeoutError and point at the subclass.  Then (after the requisite number of releases) remove defer.TimeoutError and (necessarily) make twisted.internet.error.TimeoutError no longer have it as a base class.

There may be problems with that specific scenario (the thing that immediately comes to mind: what if there's an API that currently can raise twisted.internet.error.TimeoutError _or_ defer.TimeoutError right now?) but perhaps something in that general direction will prove workable.

rodrigc:
I'm going to say that this is not worth doing.  In https://github.com/twisted/twisted/pull/145 , @cyli added new functionality for adding timeouts to deferreds.  This uses `TimeoutError`.
In previous reviews, @glyph mentioned [here](http://twistedmatrix.com/trac/ticket/8533#comment:12) that he doesn't want to add more interdependencies
from the `twisted.internet` module into the Deferred class.

Otherwise, what I would have done is subclass `twisted.internet.defer.TimeoutError` from `twisted.internet.errors.TimeoutError`, and deprecate `twisted.internet.defer.TimeoutError`.

It would have been nice if TimeoutError was removed from the Deferred class in https://github.com/twisted/twisted/commit/0d2674cb7567e33ae8bdfca67ab9e4847643696e ,
but that ship has sailed.

PR #654, opened 2017-01-01
Mark twisted.internet.defer.timeout as Deprecated

rodrigc:
https://twistedmatrix.com/trac/ticket/8971


src/twisted/test/test_defer.py
```diff
@@ -2943,3 +2943,24 @@ def test_willNotAllowNonDeferredOrCoroutine(self):
...
+    def test_deprecatedTimeout(self):
+        """
+        L{twisted.internet.defer.timeout} is deprecated.
+        """
+        deferred = defer.Deferred()
+        defer.timeout(deferred)
+        self.assertFailure(deferred, defer.TimeoutError)
+        warningsShown = self.flushWarnings([self.test_deprecatedTimeout])
+        self.assertEqual(len(warningsShown), 1)
+        self.assertIdentical(warningsShown[0]['category'], DeprecationWarning)
```
exarkun: I think `assertIs` is preferred.  That's the stdlib unittest spelling for this assertion.

src/twisted/internet/defer.py
```diff
@@ -160,6 +160,8 @@ def maybeDeferred(f, *args, **kw):
 
 
 
+@deprecated(Version('Twisted', 16, 8, 0),
```
exarkun: Presumably this will be a 17.x.y now but I'm not sure what the current system for picking such numbers is, nor how one predicts this in advance. I heard some folks talking about `Version("Twisted", "NEXT", 0, 0)`?  Maybe that was a joke of some sort though.

Probably at minimum bump this to some 17.x.y guess.

exarkun (approved):
Thanks, looks good, just minor comments.  Address and merge.

rodrigc:
@hawkowl what is the next version number of Twisted going to be, so I can make sure this patch has the right one?

PR #655, opened 2017-01-01
[#8860] Make flattenEvent handle "log_format" is None case

cdunklau:
https://twistedmatrix.com/trac/ticket/8860


wsanchez (approved):
OK.

PR #657, opened 2017-01-04
Avoid decoding default values in the unicode case.

exarkun:
Fixes https://twistedmatrix.com/trac/ticket/8974


exarkun:
@hawkowl @adiroiban

src/twisted/web/test/test_http_headers.py
```diff
@@ -371,6 +371,9 @@ def test_getRawHeadersDefaultValue(self):
         h = Headers()
         default = object()
         self.assertIdentical(h.getRawHeaders(u"test", default), default)
+        self.assertIdentical(h.getRawHeaders(u"test", None), None)
+        self.assertEqual(h.getRawHeaders(u"test", [None]), [None])
+        self.assertEqual(h.getRawHeaders(u"test", [u"\N{SNOWMAN}"]), [u"\N{SNOWMAN}"])
```
adiroiban: This is fine, but I think that these new last test should be done in a separate test function, with a docstring describing the reason why we have the tests in the first place.

I would argue that the docstring for `getRawHeaders` should be updated from 

```
@rtype: L{list} of strings, same type as C{name}
```

to

```
@rtype: L{list} of strings, same type as C{name} with the exception of the default value, which is returned as such.
```

exarkun: I think Twisted should have testtools-style `expectThat` APIs or should use Hypothesis to make it easier to make a lot of assertions without an explosive number of test methods. :)

I left these assertions in this existing test method because the prevailing convention in this module is to make multiple similar assertions in a single test method.

If there's a project policy to adhere to the overall convention rather than reflect local convention, I can update this new code to do so.

adiroiban: I am not aware of any project policy ... so for now is up to the branch author and the reviewer.
I made only a minor comment and I am OK to merge this as it is as long as you don't consider that this could improve the current code.

Project policies should be discussed over the mailing list, so feel free to merge this as it is.

Thanks!

PR #658, opened 2017-01-04
Script for converting a forked PR into a branch in the Twisted repo.

wsanchez:
https://twistedmatrix.com/trac/ticket/8975


admin/pr_as_branch
```diff
@@ -0,0 +1,56 @@
...
+TICKET_NUMBER="$1"; shift;
+  BRANCH_NAME="$1"; shift;
+
+
+# Do The Right Thing
+
+#repo="https://github.com/twisted/twisted.git";
+repo="git@github.com:twisted/twisted.git";
+
+clone="$(mktemp -d -t twisted)";
```
glyph: This mode of invocation is specific to BSD's `mktemp`, and won't work on Linux.

Can we just write this script in Python instead? ;)

Alternately, `mktemp -d -t twisted.XXXXXXXX`.

admin/pr_as_branch
```diff
@@ -0,0 +1,56 @@
...
+
+git clone --progress "${repo}" "${clone}";
```
glyph: This is a _super_ expensive way to do this; it will result in ~100M of network traffic on every invocation, which is, for example, more bandwidth than @hawkowl's totalitarian oppressors allow her in a given month.

Possible alternatives: 

- `git clone --depth 1 --progress "${repo}" "${clone}"` will result in an order-of-magnitude bandwidth reduction if you need to re-clone every time for some reason
- check for a `.git` directory in $PWD
- look at `"$(dirname $(dirname "$0"))/.git"` since if they have this script presumably they have a repository checkout already

admin/pr_as_branch
```diff
@@ -0,0 +1,56 @@
...
+
+# Usage
+
+usage() {
+  program=$(basename "$0");
+
+  if [ $# != 0 ]; then echo "$@"; echo ""; fi;
+
+  echo "${program}: usage:";
+  echo "    ${program} <pull_request_number> <trac_ticket_number> <branch_name>";
```
glyph: It would be helpful to note that the branch name is not actually the full branch name and the ticket number doesn't need to be repeated.

admin/pr_as_branch
glyph: It would also be beneficial to use an existing checkout if the user has one because some users may need to use the `https` github endpoint and reconfiguring this script by editing it is not a great way to go.

admin/pr_as_branch
```diff
@@ -0,0 +1,56 @@
...
+
+
+# Usage
+
+usage() {
+  program=$(basename "$0");
+
+  if [ $# != 0 ]; then echo "$@"; echo ""; fi;
+
+  echo "${program}: usage:";
```
glyph, line 14: A more thorough usage message explaining the purpose of this script would be helpful.  Just a sentence or two, maybe even just a link to https://twistedmatrix.com/pipermail/twisted-python/2016-December/030921.html

glyph (commented):
Given that it's an administrative shell script, please just address the review points listed, test locally, and land away.

glyph:
Oh, and, you know. Topfile.

admin/pr_as_branch
wsanchez: 6a0a8112272dd18cc185673ef66969212041a0de

admin/pr_as_branch
wsanchez: 0026ae33bd753d754bb2763d40c10feada5dae79

admin/pr_as_branch
wsanchez: Right, I forget Linux mktemp is annoying.

Python: uh, exercise left for the reader. 

cd6d82cd1475365af73798a2414350bb5ceb858d

admin/pr_as_branch
wsanchez: hawkowl should use the network in my office, eh?  Well, this will annoy me on my tiny Interweb straw in Canada, so OK.

`--depth 1` added in dcc8bdc5af0dfa387c1ff26a70029909f05261ca

wsanchez: Your suggested `dirname` nesting is missing a pair of quotes, believe it or not.

wsanchez: Extra optimizations for @hawkowl in 30453ea185e0191d9498ce3e4442c2e0b2f075aa

PR #662, opened 2017-01-06
Add cwd back to sys.path

jobevers:
Fixes https://twistedmatrix.com/trac/ticket/8978#ticket


rodrigc:
Looks OK, but next time follow the steps at https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

- need to add a topfile
- branch name should follow what is in the wiki link
- add *review* keyword in Trac ticket

Other than that looks good

PR #359, opened 2016-07-20
Fix test_iocp.py and test_tcp.py on Windows with Python 3

rodrigc:
See:
https://twistedmatrix.com/trac/ticket/8631

PR #663, opened 2017-01-06
Fix twisted --help-reactors on Windows

rodrigc:
https://twistedmatrix.com/trac/ticket/8745


exarkun:
Conditionals like this in a dropin file don't work very well.  For example, if the gtk2 dependencies aren't installed then the dropin cache is generated with no gtk2 reactor.  If you later install the gtk2 dependencies, this doesn't trigger the dropin.cache to be regenerated so you still have no gtk2 reactor available.

src/twisted/plugins/twisted_reactors.py
```diff
@@ -21,42 +21,78 @@
...
+    cf = Reactor(
+        'cf' , 'twisted.internet.cfreactor',
+        'CoreFoundation integration reactor.')
+
+    __all__.extend([
+         "cf"
+    ])
+
+
+if requireModule("twisted.internet.asyncioreactor") is not None:
```
exarkun: You could factor out the repeated `requireModule` calls into a helper that operates on `Reactor` instances, then filter a complete list of `Reactor` instances through it and only put the names of the ones that remain in `__all__`.  Then, the helper could also have a test.

tomprince:
I'm not sure if anything should be done about this, but it feels a bit weird to avoid creating `Reactor` instances, since those are designed to be creatable even if the corresponding module isn't importable.

@mithrandi also brought up the point (in relation to something else) that twisted won't regenerate the plugin cache unless this file changes, so if you install the dependencies of a reactor, the corresponding reactor plugin may not show up.

src/twisted/plugins/twisted_reactors.py
```diff
@@ -4,59 +4,70 @@
...
+    @param shortName: Short name of reactor
+    @param moduleName: The fully-qualified module name of the reactor
+    @param description: Description of reactor
+    @return: newly created reactor
+    @rtype: L{Reactor} or L{None}
+    """
+    newReactor = None
+    if requireModule(moduleName) is not None:
+        newReactor = Reactor(shortName, moduleName, description)
+        __all__.extend([shortName])
```
glyph: `list.extend([foo])` is exactly equivalent to `list.append(foo)` and it's a bit more confusing. Could you change this to be `append`?

src/twisted/plugins/twisted_reactors.py
```diff
@@ -4,59 +4,70 @@
...
+    return newReactor
+
+
+__all__ = []
```
glyph: Minor stylistic tweak: I think it's better style to declare a variable above any of its uses; since `_addReactorToAll` modifies `__all__` it would be nice to declare it above.

src/twisted/plugins/twisted_reactors.py
```diff
@@ -4,59 +4,70 @@
...
+    """
+
+    @param shortName: Short name of reactor
+    @param moduleName: The fully-qualified module name of the reactor
+    @param description: Description of reactor
+    @return: newly created reactor
+    @rtype: L{Reactor} or L{None}
+    """
+    newReactor = None
+    if requireModule(moduleName) is not None:
```
glyph, line 23: This strikes me as a recipe for silent failure.  Some reactors we expect to work in certain configurations; other reactors we expect to be absent.  I think that trying to dynamically figure this out without recording the error anywhere might be confusing.

src/twisted/plugins/twisted_reactors.py
```diff
@@ -4,59 +4,70 @@
 from __future__ import absolute_import, division
 
 from twisted.application.reactors import Reactor
-from twisted.python.compat import _PY3
+from twisted.python.reflect import requireModule
 
-default = Reactor(
+def _addReactorToAll(shortName, moduleName, description):
+    """
+
```
glyph, line 15: This function needs an explanation of its behavior.

src/twisted/plugins/twisted_reactors.py
rodrigc: With the existing implementation, if you do:

trial --help-reactors

The list that appears:

(1)   Displays reactors which aren't supported.  For example, if I do that on a Mac,
       I see:
          iocp	Win32 IO Completion Ports-based reactor.
          win32    Win32 WaitForMultipleObjects-based reactor.

     No way these Windows reactors can run on a Mac.

       I also see:
    gi  	GObject Introspection integration reactor.
    glib2	GLib2 event-loop integration reactor.
    gtk2	Gtk2 integration reactor.
    gtk3	Gtk3 integration reactor.
   wx  	wxPython integration reactor.

    but I don't have glib2 or gtk2 or gtk3 or wxPython installed.

   I also see:
     epoll	epoll(4)-based reactor.

   but that is Linux-specific and can't run on Mac.


(2)   Doesn't display reactors which are supported.  For example, if I do this on Windows with Python 3, I *don't* see:
           iocp	Win32 IO Completion Ports-based reactor.
           win32   Win32 WaitForMultipleObjects-based reactor.

But those reactors should be there, because they work on Python 3 now.

So the existing implementation lists reactors it shouldn't, and
doesn't list reactors that it should!

exarkun: Why do you say iocp and win32 reactors work on Python 3?

https://github.com/twisted/twisted/blob/a4423e916265db65ed8e6b8747000d34d4ba86e5/src/twisted/plugins/twisted_reactors.py#L32-L62

Neither reactor is in the `if _PY3:` suite.

rodrigc: See:
https://github.com/twisted/twisted/pull/359
https://github.com/twisted/twisted/pull/485

rodrigc: @glyph Where do you want to record the error if a reactor cannot be imported?
Do you want to use log.msg(), or store the error to a variable, or something else?

exarkun: @rodrigc I see.  This was confusing because you asserted that the reactors work on Windows but this PR is actually a necessary step in _making_ them work on Windows.

Why didn't you just move them into the _PY3 block?  That would be a simpler way to make this work.

rodrigc: Sure I can do that, but since I'm here, I might as well clean things up, so that
this file is doing more sensible stuff and not listing reactors that don't work at all on certain platforms due to hard-coded logic.

exarkun: Maybe.  The trouble is just that the intent of the PR was far from clear and this confused a lot of the discussion that resulted.

So, back to technical matters.  The point about dropin.cache not being regenerated means that this change will make `--help-reactors` wrong in a different way (I commented on the Twisted ticket about this a few days ago).

A different approach that won't suffer from this problem is to define all of the reactors in the dropin and then filter them down to those that actually work at some later point (probably in some part of the `--help-reactors` implementation).

rodrigc:
Moving to https://github.com/twisted/twisted/pull/675

src/twisted/plugins/twisted_reactors.py
glyph: @exarkun - moving them into the `_PY3` block is the wrong thing to do; that block is for reactors that _only_ work on PY3.  They should live outside all the blocks.

glyph: @rodrigc - I don't want to log an error for reactors that we _expect_ not to be available; certainly there will be some import errors.  I just want the expectation declared in some way that errors we _don't_ expect (asyncio not being importable on >py34 for example) result in a traceback, probably via twisted.logger.

PR #664, opened 2017-01-07
Skip builds on appveyor when only touching quotes.

tomprince:


exarkun:
Seems sensible ... where's the ticket?

PR #665, opened 2017-01-07
8967 test newtls avoid econnaborted

markrwilliams:
https://twistedmatrix.com/trac/ticket/8967


glyph:
Thanks for tackling this complex issue!  Much cleaner expression of the test's intent.  LGTM.  Will land.

PR #667, opened 2017-01-08
[#8980] respect port number when simple resolver installed

glyph:
https://twistedmatrix.com/trac/ticket/8980

### DIFF
--- a/src/twisted/internet/test/test_resolver.py
+++ b/src/twisted/internet/test/test_resolver.py
@@ -520,6 +520,23 @@ class LegacyCompatibilityTests(UnitTest, object):
         self.assertEqual(self.successResultOf(success), '192.168.3.4')
 
 
+    def test_portNumber(self):
+        """
+        L{SimpleResolverComplexifier} preserves the C{port} argument passed to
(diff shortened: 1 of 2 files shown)


src/twisted/internet/test/test_resolver.py
```diff
@@ -520,6 +520,23 @@ def test_simplifier(self):
...
 
 
+    def test_portNumber(self):
+        """
+        L{SimpleResolverComplexifier} preserves the C{port} argument passed to
+        C{resolveHostName} in its returned addresses.
+        """
+        simple = SillyResolverSimple()
+        complex = SimpleResolverComplexifier(simple)
+        receiver = ResultHolder(self)
```
markrwilliams, line 530: Unrelated to this PR, but why does `ResultHolder` accept the `TestCase`?  Nothing uses it.

markrwilliams (approved):
A question about related code.  Please merge.

src/twisted/internet/test/test_resolver.py
glyph: It's a test helper, so it's there in case we want to add some assertions or sanity checks.

PR #668, opened 2017-01-09
[#8983] Make twist respect the --reactor option

GaretJax:
Addresses https://twistedmatrix.com/trac/ticket/8983


src/twisted/application/twist/test/test_options.py
```diff
@@ -117,6 +117,20 @@ def test_installReactor(self):
...
+        L{TwistOptions.installReactor} installs the chosen reactor after the
+        command line options have been parsed.
+        """
+        self.patchInstallReactor()
+
+        options = TwistOptions()
+        options.subCommand = "test-subcommand"
+        options.parseOptions(["--reactor=fusion"])
+
+        self.assertEqual(set(self.installedReactors), set(["fusion"]))
```
exarkun, line 123: Great!

One thing about the way all of this code is being tested is that, instead of monkey-patching the options object, it could probably be using `twisted.internet.test.modulehelpers.NoReactor`.  This is a helper that basically monkey-patches the `twisted.internet.reactor` module.  I think this would isolate the tests a bit more from the implementation and go a step further in asserting the thing we actually want to assert (that the selected reactor is installed).  A possible downside is that I don't think you'll be able to install the "fusion" reactor this way (since no such thing exists at the moment :smile: ).

This is not a blocking comment for this PR.  It's more like a suggestion for a follow-up improvement of this test suite.

src/twisted/application/twist/_options.py
```diff
@@ -188,6 +187,8 @@ def subCommands(self):
 
 
     def postOptions(self):
+        self.installReactor()
+
```
exarkun: It looks like this is too late.  The unit test probably passes because it doesn't have the same condition as the real implementation: that if no reactor is installed, importing twisted.internet.reactor installs one and attempting to install a second one fails.  Instead, the unit test doesn't notice that the default reactor gets installed by an accidental import somewhere.

This is a good example of why it might be better to make the tests more realistic. :)

I had a hunch that `postOptions` might be too late so I ran `twist` manually and indeed:

```
$ twist --reactor poll web
  File "/home/exarkun/Environments/twisted-dev/bin/twist", line 11, in <module>
    load_entry_point('Twisted', 'console_scripts', 'twist')()
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/twist/_twist.py", line 132, in main
    options = cls.options(argv)
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/twist/_twist.py", line 38, in options
    options.parseOptions(argv[1:])
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/twist/_options.py", line 159, in parseOptions
    Options.parseOptions(self, options=options)
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/python/usage.py", line 277, in parseOptions
    self.postOptions()
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/twist/_options.py", line 190, in postOptions
    self.installReactor()
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/twist/_options.py", line 77, in installReactor
    self["reactor"] = installReactor(name)
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/reactors.py", line 82, in installReactor
    installer.install()
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/application/reactors.py", line 60, in install
    namedAny(self.moduleName).install()
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/internet/pollreactor.py", line 186, in install
    installReactor(p)
  File "/home/exarkun/Work/LeastAuthority/twisted/src/twisted/internet/main.py", line 32, in installReactor
    raise error.ReactorAlreadyInstalledError("reactor already installed")
twisted.internet.error.ReactorAlreadyInstalledError: reactor already installed
```

Perhaps immediately following the super-class call to `parseOptions` would be a better place.  Or, even earlier, inside `opt_reactor` itself (isn't that where twistd and trial do it?)

exarkun (dismissed):
Thanks.  Great start.  This needs a small further tweak to actually get it working but then it's probably great to merge (pending CI results).

src/twisted/application/twist/_options.py
GaretJax: Thanks for the review.
`parseOptions` calls `postOptions` as last thing it does, so that would be too late as well.
I like the `opt_reactor` approach and while digging into the Options for `twistd` I found this comment:
```
    def opt_reactor(self, shortName):
        # Actually actually actually install the reactor right at this very
        # moment, before any other code (for example, a sub-command plugin)
        # runs and accidentally imports and installs the default reactor.
```

Will propose an improved version.

src/twisted/application/twist/test/test_twist.py
```diff
@@ -45,12 +45,12 @@ def patchInstallReactor(self):
...
         self.installedReactors = {}
 
-        def installReactor(name):
+        def installReactor(_, name):
             reactor = MemoryReactor()
             self.installedReactors[name] = reactor
             return reactor
 
-        self.patch(_options, "installReactor", installReactor)
+        self.patch(TwistOptions, "installReactor", installReactor)
```
GaretJax, line 53: Not feeling particularly strong about this one, as we're patching the object being tested itself. I'm open to suggestions.

glyph:
Looks like @exarkun's comments were addressed.  I manually checked for the `twisted.internet.error.ReactorAlreadyInstalledError` and didn't see one, so I'll merge this when CI is happy.

glyph:
Since we already had a passing build, and all required checks had passed, I'm going to skip the extra CI run here.

src/twisted/application/twist/_options.py
```diff
@@ -59,7 +60,15 @@ def opt_reactor(self, name):
         The name of the reactor to use.
         (options: {options})
         """
-        self["reactorName"] = name
+        # Actually actually actually install the reactor right at this very
+        # moment, before any other code (for example, a sub-command plugin)
+        # runs and accidentally imports and installs the default reactor.
+        try:
+            self["reactor"] = self.installReactor(name)
```
wsanchez, line 67: This is not the right method to install the reactor in, because this command line will not work:

```
twist --reactor X --reactor Y
```

This should install Y, but here X will get installed first, then probably raise an error when it tries to install Y.

wsanchez: https://twistedmatrix.com/trac/ticket/9002

PR #669, opened 2017-01-10
Define the Request interface HTTPChannel depends on.

markrwilliams:
http://twistedmatrix.com/trac/ticket/8981


glyph:
Don't have time for a full review at the moment, but have you considered using [`proxyForInterface`](https://twistedmatrix.com/documents/16.6.0/api/twisted.python.components.html#proxyForInterface) in some of the existing tests to ensure that said interface is properly fully describing the interactions between `HTTPChannel` (and `H2Stream` as well, I think?) and `Request`?

markrwilliams:
@glyph Good idea!  I did [just what you suggested](https://github.com/twisted/twisted/pull/669/commits/87814a012ea116fdf578f6c6a97bdf2c1cef6dac) and found some test failures.  It looks like all the tracebacks bottom out in `test_http`.  I'll push another commit that sends the failed API calls to `.original` and see what happens next.

src/twisted/web/http.py
```diff
@@ -377,6 +379,75 @@ def parseContentRange(header):
 
 
 
+class _IDeprecatedHTTPChannelToRequestInterface(Interface):
+    """
+    The interface L{HTTPChannel} expects of L{Request}.
```
exarkun, line 384: Maybe "uses" instead of "expects".  The full interface of the `Request` type is wider - but `HTTPChannel` doesn't use the rest of it.

Also, a reference in the other direction might be handy.  That is, in the `HTTPChannel` documentation for its own `requestFactory` attribute - saying it must create providers of this interface.

src/twisted/web/test/test_http.py
```diff
@@ -249,7 +298,9 @@ def test_noPipeliningApi(self):
         """
         b = StringTransport()
         a = http.HTTPChannel()
-        a.requestFactory = DummyNewHTTPHandler
+        requestFactory = _IDeprecatedHTTPChannelToRequestInterfaceProxy.wrap(
+            DummyHTTPHandler)
```
exarkun: It would probably be nice to hoist these type definitions.  It looks like many are repeated at least a couple times - `DummyHTTPHandler` many more than that.

src/twisted/web/http.py
```diff
@@ -377,6 +379,75 @@ def parseContentRange(header):
...
+
+        @param path: The request's path.  Note: this is actually what
+            RFC7320 calls the URI.
+        @type path: L{bytes}
+
+        @param version: The request's HTTP version.
+        @type version: L{bytes}
+        """
+
+
```
exarkun, line 449: `HTTPChannel.requestDone` also requires that requests compare equal to themselves (as you noticed in your proxy, I think :smile:).  That could probably be avoided with an identity comparison in that method, replacing the equality comparison.

exarkun (approved):
Thanks.  Just some minor comments inline.  Address and merge when CI is green.  Thanks again.

src/twisted/web/http.py
markrwilliams: > Also, a reference in the other direction might be handy. That is, in the HTTPChannel documentation for its own requestFactory attribute - saying it must create providers of this interface.

so a subsequent PR will introduce a new streaming interface, making a liar out of that docstring.  but i'll cross that bridge when i come to it!

markrwilliams: hm, actually i'm loathe to put the reference from `HTTPChannel`'s docstring back to this interface, since it's supposed to be private.  if people subclass HTTPChannel i don't want them to feel like they should also construct a provider of this interface.

src/twisted/web/http.py
markrwilliams: this is tricky enough that i'm willing to admit i'm stumped.  switching `!=` to `is not` [here](https://github.com/twisted/twisted/blob/e920939343ac5f6795c4509e0fe25c33a2447531/src/twisted/web/http.py#L1997) and removing the `__eq__`/`__ne__` implementation from the proxy breaks the tests, because [`Request._cleanup` passes the `Request` instance to `HTTPChannel.requestDone`](https://github.com/twisted/twisted/blob/e920939343ac5f6795c4509e0fe25c33a2447531/src/twisted/web/http.py#L632).  in other words, we end up with:

```python
if actualRequestInstance is not proxyAroundActualRequestInstance: raise TypeError
```

that's why the `__eq__` implementation does `self.original == other`.

the Completely Correct Approach, i think, would be to add `__ne__` to the interface and then make `Request` implement it.  that seemed pretty bad so i didn't do it, but maybe it's a good idea after all?

exarkun: Seems like you need to do that, yea.

markrwilliams: OK, so the trouble with adding `__eq__`/`__ne__` to the interface and implementing `Request.__eq__`/`__ne__` is that the proxied class gets compared to a _wrapped_ instance.

Specifically:

```python
if request != self.requests[0]: raise TypeError
```

ends up being 

```python
if requests.__eq__(self.requests[0]): raise TypeError
```

the guts of this end up being

```python
def __eq__(self, other):
    return self is other
```

since the `__eq__` implementation gets proxied through to the underlying request object, that in turns ends up being

```
return actualRequestInstance is not proxyAroundActualRequestInstance
```

...which is of course `False`.  i'm going to claim this difficulty is exactly why i did what i did with the tests' proxy subclass.

i could add an identity function to the interface that returned `id(self)`, but that seems excessive.  is there an easy to do this i'm not seeing?

markrwilliams: For posterity: we discussed this on IRC and @exarkun suggested 9a5f758.

PR #670, opened 2017-01-10
Make Agent set `Content-Length: 0` for some empty-payload requests

exarkun:
Fixes https://twistedmatrix.com/trac/ticket/8984


src/twisted/web/_newclient.py
```diff
@@ -813,7 +813,14 @@ def writeTo(self, transport):
             else:
                 return self._writeToContentLength(transport)
         else:
-            self._writeHeaders(transport, None)
+            # If the method semantics anticipate a body, include a
```
markrwilliams, line 832: I'd like to see `_writeToContentLength` renamed `_writeToContentLengthWithBodyProducer` or something and this code lifted into a new method called `_writeToContentLengthWithoutBody`.  I think it's confusing to have a `_writeToContentLength` method that doesn't always write the content length.

markrwilliams (commented):
A comment on a possible refactoring.

It looks like the build's failing because h2's `objects.inv` couldn't be downloaded.  I guess [the fix didn't take](http://twistedmatrix.com/trac/ticket/8823)?

src/twisted/web/_newclient.py
exarkun: Thanks!  I think `_writeToContentLength` does always write the content length.  The fix is all in `writeTo` which sometimes called `_writeHeaders` directly with no `Content-Length` (or `Transfer-Encoding`).

I guess what's true now is that sometimes a `Content-Length` headers is written by something _other_ tyhan `_writeToContentLength` ... maybe that's what you mean.  And, yea, I can see how that might confuse someone at some point.

exarkun:
The failure I see on travis-ci is about a missing topfile, I don't see an objects.inv-related failure.

markrwilliams:
> The failure I see on travis-ci is about a missing topfile, I don't see an objects.inv-related failure.
View changes

Travis got through unscathed, but a Fedora builder did not:
https://buildbot.twistedmatrix.com/builders/fedora23-py2.7/builds/2336

exarkun:
Ah!  I see.  I looked at the objects.inv fix and left a comment on the Twisted ticket.  It certainly looks to me like there's still ample room for failures.

exarkun:
Any idea why the codecov report says this noticably worsens coverage?  It looks to me like all lines of the diff are covered.

src/twisted/web/_newclient.py
markrwilliams: > I guess what's true now is that sometimes a Content-Length headers is written by something other tyhan _writeToContentLength ... maybe that's what you mean. And, yea, I can see how that might confuse someone at some point.

Yep, that's it!  There are two things that bother me:

1. `_newclient` passes part of the HTTP wire protocol to private methods (`TEorCL`).  I can see why it does this, and the protocol format hasn't changed in 20 years, and it has nothing to do with this PR, so it's a lesser issue :)
2. `Content-Length` can be written outside of the method that says `_writeToContentLength`.  It's true that `_writeToContentLength` assumes a lot of things - most importantly that there's a body producer - so I thought maybe it could be given a name that reflected its actual behavior.

These are nits, though.

markrwilliams:
> Any idea why the codecov report says this noticably worsens coverage? It looks to me like all lines of the diff are covered.

No :(  Will try to look into it.

src/twisted/web/_newclient.py
```diff
@@ -797,6 +797,15 @@ def f(passthrough):
         return d
 
 
+    def _writeToEmptyBodyContentLength(self, transport):
+        """
+        Write this request to the given transport using content-length to frame
```
markrwilliams, line 808: `twistedchecker` complains that `transport` isn't documented, nor is the return value.

src/twisted/web/_newclient.py
```diff
@@ -807,14 +816,18 @@ def writeTo(self, transport):
             been completely written to the transport or with a L{Failure} if
             there is any problem generating the request bytes.
         """
-        if self.bodyProducer is not None:
-            if self.bodyProducer.length is UNKNOWN_LENGTH:
-                return self._writeToChunked(transport)
+        if self.bodyProducer is None:
+            # If the method semantics anticipate a body, include a
+            # Content-Length even if it is 0.
+            # https://tools.ietf.org/html/rfc7230#section-3.3.2
```
markrwilliams, line 834: :+1:

src/twisted/web/test/test_newclient.py
```diff
@@ -2040,6 +2040,42 @@ def test_sendRequestBodyWithLength(self):
         self.assertEqual(self.transport.value(), b"abc")
 
 
+    def _sendRequestEmptyBodyWithLength(self, method):
```
markrwilliams, line 2043: Got to document the `method` argument...

markrwilliams (approved):
This looks good!  I commented on the locations of the three `twistedchecker` complaints that prevented this from going green.

Please address and merge!

exarkun:
buildbot/osx10.10-py2.7 not yet reporting its results because it's been offline for almost 3 days.  Considering the distance this code has from anything even remotely OSX related and the other green build results (ignoring the spurious test_singleThread failure which I filed #8989 for), I'm going to merge.

redapple:
This change triggers a duplicate `Content-Length: 0` with Scrapy>=1.1 ([reported on StackOverflow](https://stackoverflow.com/questions/42966405/scrapy-post-header-has-two-content-length-fields)).

[Scrapy 1.1 patched this same bug on its end](https://github.com/scrapy/scrapy/pull/1800) by adding a `Content-Length: 0` when passing bodyproducer as `None`.
We can change that on Scrapy side to always pass a producer, with length 0 for empty bodies.
But maybe you also want to write `Content-Length: 0` only if it's not already in passed headers.

glyph:
Hi @redapple - thanks for the report!

Can you file this on twistedmatrix.com with the "release blocker: regression" ticket type?

redapple:
hi @glyph , done: https://twistedmatrix.com/trac/ticket/9097.

glyph:
Thanks! Hopefully someone will review it :).

PR #671, opened 2017-01-11
8985 - error on unicode

derwolfe:
See https://twistedmatrix.com/trac/ticket/8985


src/twisted/web/topfiles/8985.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web.Resouce.putChild raises a TypeError when the supplied path is not bytes (broken since python3 has been supported).
```
derwolfe, line 1: I'm not sure if it is even accurate (a) to call this a bugfix and (b) to state that this has been broken since python  3 support was added. The reasoning I chose to state it this way is because I'd expect a user would only have encountered this with the introduction of python3 support, unless the user had explicitly tried to pass unicode text in as a path on python2.

derwolfe:
Closing in favor of an alternative implementation.

PR #672, opened 2017-01-12
Add cwd back to sys.path for trial

rodrigc:
https://twistedmatrix.com/trac/ticket/8978#ticket


exarkun:
Note I closed the referenced Twisted ticket as a duplicate of one filed a few days earlier.  Also it's not clear this behavior _should_ be restored since various people have implied that the change was intentional.

rodrigc:
Well, I did the work to remove bin/trial from the system,
so the change of not adding cwd to sys.path for trial was unintentional on my part.

rodrigc:
The other thing that was done in the Twisted code tree to deal with the path of things
was to move everything to the src directory in : https://github.com/twisted/twisted/commit/59709f63ad3cb0e0aa3ec0e3ed46000fbb92f50c for http://twistedmatrix.com/trac/ticket/8214

That affects testing and working on the Twisted codebase itself.
This fix doesn't affect the work done for ticket 8214.
It also restores behavior introduced in https://github.com/twisted/twisted/commit/cb98a917e8176873a8237c37de8a2c7a178e6266
that apparently people who use trial to test their own code were depending on.

rodrigc:
Based on complaints I have seen on the mailing list, I think this behavior *should* be
restored for trial.  There are some people who use trial for testing their own code
(not Twisted itself), and the change in behavior that occurred is annoying.
I missed it when I converted trial to a console script, because all the Twisted unit tests passed.

I added a new unit test for this behavior, so it will be noticeable if this breaks in future.

Julian (approved):
lgtm, appreciate you hitting this given me being one of afformentioned-slightly-annoyed people :P

Julian:
Erm, could you address what @exarkun pointed out in some way though with ticket logistics? Gonna just consider this relevant to 8972 given that it's the open ticket, but maybe you should just move the topfile or something.

exarkun:
@glyph made a point on the mailing list, https://twistedmatrix.com/pipermail/twisted-python/2017-February/031145.html

Julian:
@rodrigc sounds like consensus was to WONTFIX then -- can you close?

hawkowl:
This is a won't fix, yes.

PR #673, opened 2017-01-13
[WIP] Add some hypothesis tests for `ClientService`.

tomprince:
This also fixes some bugs that were thus uncovered.

This depends on https://github.com/HypothesisWorks/hypothesis-python/pull/425

### DIFF
--- a/src/twisted/internet/task.py
+++ b/src/twisted/internet/task.py
@@ -808,7 +808,7 @@ class Clock:
         return self.calls
 
 
-    def advance(self, amount):
+    def advance(self, amount = None):
         """
         Move time on this clock forward by the given amount and run whatever
         pending calls should be run.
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -52,6 +52,7 @@ extras =
 
 ;; dependencies that are not specified as extras
 deps =
+    git+https://github.com/tomprince/hypothesis@invariants
```
tomprince, line 55: This shouldn't be merged.

src/twisted/python/_setup.py
```diff
@@ -77,6 +77,7 @@
     'twisted-dev-tools >= 0.0.2',
     'python-subunit',
     'sphinx >= 1.3.1',
+    'hypothesis >= 3.6.1'
```
tomprince, line 80: This will need to be bumped to take advantage of hopefully merged invariant support upstream.

src/twisted/internet/task.py
```diff
@@ -817,8 +817,14 @@ def advance(self, amount):
...
         """
-        self.rightNow += amount
         self._sortCalls()
+        if amount is not None:
+            self.rightNow += amount
+        else:
+            if self.calls:
+                self.rightNow = max(self.calls[0].getTime(), self.rightNow)
+            else:
+                return
```
tomprince, line 827: This should be in a separate PR, and probably exist as a separate function, rather than using a default argument.

tomprince:
I'm not going to get back to this, probably.

PR #674, opened 2017-01-13
Make clock.advance() run till next delayed call.

tomprince:
https://twistedmatrix.com/trac/ticket/8991

### DIFF
--- a/src/twisted/internet/task.py
+++ b/src/twisted/internet/task.py
@@ -808,7 +808,7 @@ class Clock:
         return self.calls
 
 
-    def advance(self, amount):
+    def advance(self, amount=None):
         """
         Move time on this clock forward by the given amount and run whatever
         pending calls should be run.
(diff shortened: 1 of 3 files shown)


src/twisted/test/test_task.py
```diff
@@ -90,6 +90,36 @@ def testAdvance(self):
...
+    def test_advanceDefaultNoCalls(self):
+        """
+        Advancing the clock without specifying an amount when there
+        are no pending calls won't advance the clock.
+        """
+        c = task.Clock()
+
+        c.advance()
+
+        self.assertEqual(c.seconds(), 0)
```
exarkun, line 120: This seems like it's an encouragement to silent failure.  For example, it mixes poorly with code like:

```
while my_condition:
    clock.advance()
```

I'd probably make advancing to the next event when there is no next event an error, instead.

src/twisted/internet/task.py
```diff
@@ -808,7 +808,7 @@ def getDelayedCalls(self):
         return self.calls
 
 
-    def advance(self, amount):
+    def advance(self, amount=None):
```
exarkun, line 811: -1 on overloading `advance` with this.  How about `advanceToNextEvent`?

tomprince:
I'm not going to get back to this, probably.

PR #675, opened 2017-01-13
Fix trial --help-reactors on Windows

rodrigc:
https://twistedmatrix.com/trac/ticket/8745

### DIFF
--- a/src/twisted/application/app.py
+++ b/src/twisted/application/app.py
@@ -20,7 +20,7 @@ from twisted.internet import defer
 from twisted.persisted import sob
 from twisted.python import runtime, log, usage, failure, util, logfile
 from twisted.python._oldstyle import _oldStyle
-from twisted.python.reflect import qual, namedAny
+from twisted.python.reflect import (qual, namedAny, requireModule)
 
 # Expose the new implementation of installReactor at the old location.
 from twisted.application.reactors import installReactor
@@ -501,8 +501,9 @@ class ReactorSelectionMixin:
         """
         rcts = sorted(self._getReactorTypes(), key=attrgetter('shortName'))
         for r in rcts:
-            self.messageOutput.write('    %-4s\t%s\n' %
-                                     (r.shortName, r.description))
+            if requireModule(r.moduleName) is not None:
+                self.messageOutput.write('    %-4s\t%s\n' %
+                                         (r.shortName, r.description))
         raise SystemExit(0)
 
 
--- a/src/twisted/application/app.py
+++ b/src/twisted/application/app.py
@@ -20,7 +20,7 @@ from twisted.internet import defer
 from twisted.persisted import sob
 from twisted.python import runtime, log, usage, failure, util, logfile
 from twisted.python._oldstyle import _oldStyle
-from twisted.python.reflect import (qual, namedAny, requireModule)
+from twisted.python.reflect import (qual, namedAny, namedModule)
 
 # Expose the new implementation of installReactor at the old location.
 from twisted.application.reactors import installReactor
@@ -500,10 +500,21 @@ class ReactorSelectionMixin:
         Display a list of possibly available reactor names.
         """
         rcts = sorted(self._getReactorTypes(), key=attrgetter('shortName'))
+        notWorkingReactors = ""
         for r in rcts:
-            if requireModule(r.moduleName) is not None:
+            try:
+                namedModule(r.moduleName)
                 self.messageOutput.write('    %-4s\t%s\n' %
                                          (r.shortName, r.description))
+            except ImportError as e:
+                notWorkingReactors += ('    !%-4s\t%s (%s)\n' %
+                                       (r.shortName, r.description, e.args[0]))
+
+        if notWorkingReactors:
+            self.messageOutput.write('\n')
+            self.messageOutput.write('    reactors not available '
+                                     'on this platform:\n\n')
+            self.messageOutput.write(notWorkingReactors)
         raise SystemExit(0)
 
 
--- a/src/twisted/plugins/twisted_reactors.py
+++ b/src/twisted/plugins/twisted_reactors.py
@@ -13,59 +13,59 @@ default = Reactor(
 __all__.append('default')
 
 select = Reactor(
-    'select', 'twisted.internet.selectreactor', 'select(2)-based reactor.')
+    'select', 'twisted.internet.selectreactor', 'select(2) based reactor.')
 __all__.append('select')
 
 poll = Reactor(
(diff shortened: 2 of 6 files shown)


src/twisted/application/app.py
```diff
@@ -501,8 +501,9 @@ def opt_help_reactors(self):
         """
         rcts = sorted(self._getReactorTypes(), key=attrgetter('shortName'))
         for r in rcts:
-            self.messageOutput.write('    %-4s\t%s\n' %
-                                     (r.shortName, r.description))
+            if requireModule(r.moduleName) is not None:
+                self.messageOutput.write('    %-4s\t%s\n' %
+                                         (r.shortName, r.description))
         raise SystemExit(0)
```
exarkun, line 518: Cool.  I like the result behavior a lot.  On trunk@HEAD on Linux I get this behavior:

```
$ twistd --help-reactors
    cf          CoreFoundation integration reactor.
    default     A reasonable default: poll(2) if available, otherwise select(2).
    epoll       epoll(4)-based reactor.
    gi          GObject Introspection integration reactor.
    glib2       GLib2 event-loop integration reactor.
    gtk2        Gtk2 integration reactor.
    gtk3        Gtk3 integration reactor.
    iocp        Win32 IO Completion Ports-based reactor.
    kqueue      kqueue(2)-based reactor.
    poll        poll(2)-based reactor.
    select      select(2)-based reactor.
    win32       Win32 WaitForMultipleObjects-based reactor.
    wx          wxPython integration reactor.
```

yet I can't use many of these reactors, eg

```
$ twistd --reactor wx -n web 
...
twistd: The specified reactor cannot be used, failed with error: No module named wxPython.wx.
```

So this improves the behavior on all platforms, not just Windows.

I think that what would be one step better, and perhaps this relates to what @glyph was asking for, is to have some explanation as to why each other reactor is unavailable.  However, I'm not sure I really like my only idea for presenting this, which would be something along the lines of:

```
$ trial --help-reactors
    default     A reasonable default: poll(2) if available, otherwise select(2).
    epoll       epoll(4)-based reactor.
    poll        poll(2)-based reactor.
    select      select(2)-based reactor.
    ...
   !wx           wxPython integration reactor. (No module named wxPython.wx.)
```

Note how this identifies that the reactor exists but is not usable and gives at least a minimal description of why it is unusable.  The `!` is a bit opaque and squashing the failure into just the exception message often loses more information than you want... But perhaps we can come up with something else that accomplishes the two things this example tries to accomplish.

I'm also not sure that such an improvement needs to block _this_ work.  It's hard to say whether this is a strict improvement over trunk or a bit of sideways movement.  I _do_ like it but that's essentially up to my subjective sense of aesthetics.

src/twisted/plugins/twisted_reactors.py
```diff
@@ -4,59 +4,68 @@
 from __future__ import absolute_import, division
 
 from twisted.application.reactors import Reactor
-from twisted.python.compat import _PY3
```
exarkun, line 7: I also like that this module no longer has any conditionals.

src/twisted/test/test_application.py
```diff
@@ -889,3 +891,54 @@ def install():
...
+        self.message = self.options.messageOutput.getvalue()
+
+
+    def test_asyncIOPy2(self):
+        """
+        --help-reactors should NOT display the asyncIO reactor on Python 2
+        """
+        self.assertNotIn(twisted_reactors.asyncio.description, self.message)
+    if _PY3:
+        test_asyncIOPy2.skip = "Not applicable on Python 3"
```
exarkun: I might instead write this and the following test as a single test with the `if _PY3: ...` conditional inside the test method, switching between either asserting in or not in depending on.  This is an optional change.

src/twisted/test/test_application.py
```diff
@@ -889,3 +896,54 @@ def install():
...
         self.assertIn(message, e.args[0])
         self.assertIn("help-reactors", e.args[0])
+
+
+
+class HelpReactorsTests(unittest.TestCase):
+    """
+    --help-reactors lists the available reactors
+    """
+    def setUp(self):
```
exarkun, line 906: The reactor list is fairly stable and the platforms on which reactors work almost never change so it's probably pretty reasonable to test this behavior against the real reactor plugins.

There's more than one unit under tests here, though.  First, there's the plugins themselves: epollreactor, iocpreactor, etc.  Then there's the dropin which defines a collection of plugins.  Finally, there's the filtering and reporting logic of `ReactorSelectionMixin`.  The ideal I would strive for, in general, is to separate these units out and test them independently (and then perhaps perform an integration test to ensure they fit together).  In this case, that would suggest parameterizing the `ReactorSelectionMixin` and providing a set of reactor plugins defined by the test suite to it.  The tests is then independent of platform or Python version or changes to the reactors Twisted actually provides (or even third-party reactor plugins which the tests will currently examine - allowing something pathelogical like a Linux port of the iocp reactor - which would break the last test in this TestCase if installed).

I'd consider this optional based on the caveats given at the top.  It might make a nice follow-up refactoring, though.

exarkun: Also, this TestCase is a poster-child for testtools-style `expectThat` assertions. :)

src/twisted/topfiles/8745.bugfix
```diff
@@ -0,0 +1 @@
+trial --help-reactors will now display iocp and win32er reactors with Python 3.
```
exarkun, line 1: Note `twistd` also has a `--help-reactors` option and this fixes that as well.

Also, did  you mean to have a bugfix and a feature news fragment for this change?  It seems like one or the other is enough to draw user attention to it... But if you think the iocp/win32er aspect of the change is significant enough to merit special attention, that's okay with me.

exarkun (approved):
Thanks.  I like this a lot.  Other comments follow but given green CI and changes to at least address the points I didn't mark as optional, I think this should be merged.

src/twisted/application/app.py
glyph: I second this sentiment:

- the error reporting here is definitely an improvement
- the `!` in the reactor list would be good (maybe a "*" and then at the bottom of "*: reactors not available on this platform, --reactor=whatever for details")
- it need not block this work.

src/twisted/plugins/twisted_reactors.py
glyph: YES! Every removal of the `_PY3` constant is a victory for humankind.

glyph:
Unfortunately it does look like the CI failures are relevant to the change being made in this branch.  (I'm working on getting the required OS X builder back online.)

PR #676, opened 2017-01-13
Make sure to unregister the producer when timing out a HTTP connection.

Lukasa:
Resolves [Twisted issue 8992](https://twistedmatrix.com/trac/ticket/8992).

### DIFF
--- a/src/twisted/web/topfiles/8992.bugfix
+++ b/src/twisted/web/topfiles/8992.bugfix
@@ -1,1 +1,1 @@
-twisted.web's HTTP server no longer leaks file descriptors when timing out inactive clients using TLS connections.
+twisted.web.http.HTTPChannel is less likely to leak file descriptors when timing out clients using HTTPS connections. In some cases it is still possible to leak a file descriptor when timing out HTTP clients: further patches will address this issue.
--- a/src/twisted/web/topfiles/8992.bugfix
+++ b/src/twisted/web/topfiles/8992.bugfix
@@ -1,0 +1,1 @@
+twisted.web's HTTP server no longer leaks file descriptors when timing out inactive clients using TLS connections.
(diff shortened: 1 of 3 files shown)


Lukasa:
There is *no way* this patch reduces coverage that much. No chance.

src/twisted/web/topfiles/8992.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web's HTTP server no longer leaks file descriptors when timing out inactive clients using TLS connections.
```
exarkun: Style guide says the subject should be a FQPN.  So perhaps "twisted.web.http.HTTPChannel now ...".  Also, given the comment about how the code is still using `loseConnection` and not `abortConnection`, maybe it's better to claim it tries to fix one case and not imply that connections never leak at all with the new version.

exarkun (approved):
Thanks.  Looks good.  Just a quibble with the news fragment.  Please address and merge.  Please also file a follow-up ticket for disconnecting non-cooperating clients (ie, using abortConnection).  Which maybe belongs in TimeoutMixin?  Or as part of an even more general "try to close gracefully but after a while just kill it" API.

exarkun:
Also, per development guidelines, please file a ticket for the unrelated failure at https://buildbot.twistedmatrix.com/builders/windows7-64-py2.7-coverage/builds/711/steps/default/logs/problems :/

Lukasa:
Thanks @exarkun. The follow-up ticket was in fact already opened by past-me as [8902](https://twistedmatrix.com/trac/ticket/8902). Tracking ticket for the build failure has just been opened as [8994](https://twistedmatrix.com/trac/ticket/8994). Topfile has been reworded.

I'll let the builders run again and then merge. Thanks for the review!

PR #677, opened 2017-01-14
Fixes for `ClientService`.

tomprince:
https://twistedmatrix.com/trac/ticket/8993

This is based on #673 without the tests. The tests add a new dependency and need some improvement.

I'm putting this up for review without tests, since this is a release blocker, and in case somebody wants to pick up and write some manual tests for this.

### DIFF
--- a/src/twisted/application/internet.py
+++ b/src/twisted/application/internet.py
@@ -830,8 +830,9 @@ class _ClientMachine(object):
 
     _init.upon(start, enter=_connecting,
                outputs=[_connect])
-    _init.upon(stop, enter=_init,
-               outputs=[])
+    _init.upon(stop, enter=_stopped,
+               outputs=[],
+               collector=lambda _: succeed(None))
 
     _connecting.upon(start, enter=_connecting, outputs=[])
     # Note that this synchonously triggers _connectionFailed in the
@@ -847,7 +848,10 @@ class _ClientMachine(object):
     _waiting.upon(start, enter=_waiting,
                   outputs=[])
     _waiting.upon(stop, enter=_stopped,
-                  outputs=[_waitForStop, _stopRetrying, _finishStopping],
+                  outputs=[_waitForStop,
+                           _cancelConnectWaiters,
+                           _stopRetrying,
+                           _finishStopping],
                   collector=_firstResult)
     _waiting.upon(_reconnect, enter=_connecting,
                   outputs=[_connect])
@@ -863,7 +867,8 @@ class _ClientMachine(object):
     _disconnecting.upon(start, enter=_restarting,
                         outputs=[_resetFailedAttempts])
     _disconnecting.upon(stop, enter=_disconnecting,
-                        outputs=[])
+                        outputs=[_waitForStop],
+                        collector=_firstResult)
     _disconnecting.upon(_clientDisconnected, enter=_stopped,
                         outputs=[_cancelConnectWaiters,
                                  _finishStopping,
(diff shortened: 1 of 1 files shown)


glyph (changes requested):
Thanks for splitting this out, Tom.  Just two things:

- Please put this on a Twisted branch for the full CI run, and
- Please add a unit test that exercises this state transition.

Regarding the second point, I think that any time Hypothesis finds a problem via generative testing, it's helpful to write a "normal" test to illustrate what's supposed to happen in a more readable way.

markrwilliams:
@glyph #685 should address both points :)

tomprince:
> Regarding the second point, I think that any time Hypothesis finds a problem via generative testing, it's helpful to write a "normal" test to illustrate what's supposed to happen in a more readable way.

I *vehemently* disagree with this. Most of the failures here are `stopService` doesn't always return a deferred. I want a test that says "`stopService` always returns a deferred", not 5 tests that say "in this particular case, `stopService` returns a deferred. For a few of reasons:
- as the need for this PR shows, a bunch of individual tests isn't enough to demonstrate that the code has the desired behavior.
- having 5 tests doesn't tell you any more than the one test does, while being more code to maintain, and more code to look at to try and figure out what is going on.
- even if the tests fix all the bugs discovered by hypothesis, they don't necesarily catch all the permutations that hypothesis is testing. For example https://github.com/twisted/twisted/pull/685/files#diff-69fd7afc697207aa5a8ddef3aeb69ce8R879 is to catch one change in this PR, but that line can be triggered in at least a couple of paths through the state machine, and future changes might only affect one path not the other.

markrwilliams:
I agree with @tomprince.  The non-Hypothesis tests are valuable as regression tests, but they do not give me confidence that the state machine is correct.  Studying the [state chart](https://cloud.githubusercontent.com/assets/704247/22052008/2dc63146-dcfc-11e6-9ce0-d32a8c7fd035.png) made me more confident.  But that studying would be better done by a program.

I think Hypothesis *did* study the state machine to the point that it proved @tomprince's invariant better than my ad-hoc tests.  I also hope Twisted uses Hypothesis more.  So @glyph -- are you worried about non-determinism?  In my (relatively limited) experience if you assert your code's properties precisely the non-determinism isn't an issue.

Having said that, I think the real advantage to Automat is that it makes your code statically verifiable in a way it wouldn't be otherwise.  Questions like "does `stopService` always return a `Deferred`" are answerable by exhaustive graph traversal.  I believe that in addition to sorting out how we'll use Hypothesis, we should also talk [elsewhere](https://github.com/glyph/automat/pulls) about ways we can it easier to assert properties about Automat machines.

tomprince:
I don't actually think that the tests in #685 are valuable in the long term. They are valuable *now* so we can get `ClientService` fixed, but I think once the hypothesis tests land, we should consider aggressively pruning the non-hypothesis tests that duplicate what is being tested by hypothesis.

glyph:
> I don't actually think that the tests in #685 are valuable in the long term. They are valuable now so we can get ClientService fixed, but I think once the hypothesis tests land, we should consider aggressively pruning the non-hypothesis tests that duplicate what is being tested by hypothesis.

OK. I'm willing to put down my earlier suggestion to me simply not understanding Hypothesis well enough yet.  I still would like to see a way of introducing some level of determinism explicitly, but I'm OK with that being something that happens in the context of Hypothesis.

(This was going to be more or less my suggestion before you said it - let's add these tests for now so we can see what's wrong in the existing idiom, and then have another branch that deletes a bunch of tests later, demonstrating Hypothesis's superiority for test suite maintainability...)

PiDelport:
I'd like to chip in a small note here that Hypothesis supports [providing explicit examples](https://hypothesis.readthedocs.io/en/latest/details.html#providing-explicit-examples), if you want to make sure that some examples are always tested, and not subject to non-determinism.

When providing explicit examples like this, the Hypothesis tests can guarantee the same deterministic coverage of existing unit tests in addition to the coverage of Hypothesis's example generation.

tomprince:
Although I don't think it currently supports that for its [stateful testing support](https://hypothesis.readthedocs.io/en/latest/stateful.html) which the tests we have for this code use.

DRMacIver:
It doesn't, but FWIW it's relatively easy to write a test that explicitly invokes the state machine: A failing state machine test just creates a state machine object, calls some methods and then calls teardown. You can do that manually just as well as automatically.

So this doesn't necessarily have to be either/or. You can have the Hypothesis tests and add a test when it finds new failures that builds on rather than having to rewrite the machinery you use for Hypothesis.

exarkun:
Here's an example of exercising a specific set of transitions for a Hypothesis stateful test:

https://github.com/LeastAuthority/leastauthority.com/blob/5c94279cd9a28e8f1116102f2ef01fec5ff5f43b/src/lae_automation/test/test_subscription_converger.py#L699-L719

PR #679, opened 2017-01-15
DelayedCall needs to call str(self), not bytes(self) to work on Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8995


src/twisted/internet/test/test_base.py
```diff
@@ -183,6 +183,7 @@ def setUp(self):
         Create two L{DelayedCall} instanced scheduled to run at different
```
wsanchez: If you don't mind changing `instanced` to `instances` here, that'd be nice.  :-)

src/twisted/internet/test/test_base.py
```diff
@@ -193,6 +194,7 @@ def test_str(self):
         C{str}, includes the unsigned id of the instance, as well as its state,
         the function to be called, and the function arguments.
         """
+        DelayedCall.debug = False
```
wsanchez: Shouldn't this also be a call to `self.patch()`?

I guess I've never altered an already-patched value in a test before, but if the `patch` call in `setUp` goes away for some reason, then this line will be altering global state.

src/twisted/internet/test/test_base.py
```diff
@@ -183,6 +183,7 @@ def setUp(self):
         Create two L{DelayedCall} instanced scheduled to run at different
         times.
         """
+        self.patch(DelayedCall, 'debug', True)
```
wsanchez: So… why change all tests to set `debug` on (except `test_str` below)?

The result is that now we are *not* testing everything with `debug=False`, which is the normal config.  It would make more sense to me to subclass this test suite and set `debug=True` in that subclass, so we are testing in both configs, if that's the approach.

But simpler still would be to add tests for `__str__` and `cancel` and `__init__` to test how they behave with `debug=True` and when `self._str` is set.  That seems more unit-test-y.

wsanchez (dismissed):

rodrigc:
@wsanchez Can I proceed with this?  I made the changes.

glyph (approved):
Looks like the feedback was addressed, please proceed.

PR #680, opened 2017-01-16
Eliminate inspect.getargspec() deprecation warnings on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8997


src/twisted/python/deprecate.py
```diff
@@ -653,7 +653,7 @@ def warnAboutFunction(offender, warningString):
 
 
 
-def _passed(argspec, positional, keyword):
+def _passed_argspec(argspec, positional, keyword):
```
exarkun: camelCase naming convention. :)

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
 
 
+def _passed_signature(signature, positional, keyword):
+    """
+    Take an I{inspect.Signature}, a tuple of positional arguments, and a dict of
+    keyword arguments, and return a mapping of arguments that were actually
+    passed to their passed values.
+
+    @param signature: The signature of the function to inspect.
+    @type signature: I{inspect.Signature}
```
exarkun: `L{inspect.Signature}` - sort of like **L**ink to this FQPN.

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
+    @type keyword: L{dict}
+
+    @return: A dictionary mapping argument names (those declared in C{argspec})
+        to values that were passed explicitly by the user.
+    @rtype: L{dict} mapping L{str} to L{object}
+    """
+    if isinstance(spec, inspect.ArgSpec):
+        # Python 2
+        return _passed_argspec(spec, positional, keyword)
+    elif isinstance(spec, inspect.Signature):
```
exarkun: Uncovered branch here.

Also, it feels a little wrong that the application code using this goes out of its way to generate two different types to use as an argument and then this code has to go out of its way to sort that out and dispatch to the right thing.  I'd consider throwing this out and just being sure to call the right function at the call site.

src/twisted/python/deprecate.py
```diff
@@ -714,10 +787,15 @@ def function(tweedledum=1, tweedledee=2):
     @rtype: 1-argument callable taking a callable and returning a callable.
     """
     def wrapper(wrappee):
-        argspec = inspect.getargspec(wrappee)
+        if hasattr(inspect, "getfullargspec"):
```
exarkun: As a first approximation, don't ever use `hasattr`.  It masks exceptions which can produce confusing behavior.

Generally `getattr(inspect, name, None)` is a good replacement.

In this case, since this pattern is repeated in the test module, perhaps you want to stick something into `twisted.python.compat` for "non-deprecated get-a-signature-like-function" and use that directly from each of these callers.

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
+    return result
+
+
+
+def _passed(spec, positional, keyword):
+    """
+    Passthru function for L{_passed_argspec} or L{_passed_signature}
+
+    @param spec: The argument specification for the function to inspect.
+    @type spec: I{inspect.ArgSpec} or I{inspect.Signature}
```
exarkun: `L{...}` again

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
+    kwargs = None
+    for (n, (name, param)) in enumerate(signature.parameters.items()):
+        if param.kind == inspect.Parameter.VAR_POSITIONAL:
+            # Varargs, for example: *args
+            result[name] = positional[n:]
+            n = len(positional)
+        elif param.kind == inspect.Parameter.VAR_KEYWORD:
+            # Variable keyword args, for example: **my_kwargs
+            kwargs = result[name] = {}
+        elif param.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD:
```
exarkun: Uncovered branch here - and there are several other parameter types supported by `Signature`.  This code needs some new tests exercising those new parameter types to be complete.

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
+            if n < len(positional):
+                result[name] = positional[n]
+    if len(positional) > (n + 1):
+        raise TypeError("Too many arguments.")
+    for name, value in keyword.items():
+        if name in signature.parameters.keys():
+            if name in result:
+                raise TypeError("Already passed.")
```
exarkun, line 749: This mirrors the behavior of the old code but it sure would be nice to know *what* was already passed.

src/twisted/python/deprecate.py
```diff
@@ -696,6 +696,79 @@ def _passed(argspec, positional, keyword):
...
+            result[name] = value
+        elif kwargs is not None:
+            kwargs[name] = value
+        else:
+            raise TypeError("no such param")
```
exarkun, line 754: And to know what parameter someone thought existed, here, but doesn't.

src/twisted/python/deprecate.py
```diff
@@ -714,10 +787,15 @@ def function(tweedledum=1, tweedledee=2):
...
+        else:
+            # Python 2
+            spec = inspect.getargspec(wrappee)
         @wraps(wrappee)
         def wrapped(*args, **kwargs):
-            arguments = _passed(argspec, args, kwargs)
+            arguments = _passed(spec, args, kwargs)
             for this, that in argumentPairs:
                 if this in arguments and that in arguments:
                     raise TypeError("nope")
```
exarkun, line 791: This is a fairly unfortunate exception (though it's not new in this PR so it would be better to fix it separately).

exarkun (changes requested):
Thanks!  A few comments.  Most significantly, I think there's some missing test coverage that needs to be added.

src/twisted/python/deprecate.py
rodrigc: OK

src/twisted/python/deprecate.py
rodrigc: OK

src/twisted/python/deprecate.py
rodrigc: OK

rodrigc:
Looks like there is a bug in pydoctor, because when I changed to L{inspect.ArgSpec},
the doc build gives an error: https://buildbot.twistedmatrix.com/builders/documentation/builds/2426/steps/api-documentation/logs/stdio

rodrigc:
@exarkun Please review the updated patch.  I cannot log into Trac via: https://twistedmatrix.com/trac/github/login

rodrigc:
@exarkun Can I proceed with this?  I increased the test coverage and added handling for more types of function parameters.

exarkun:
Hard to remember to re-review when it's not marked for review on trac.  Have you complained about login problems on the mailing list?

src/twisted/python/test/test_deprecate.py
```diff
@@ -1067,3 +1068,56 @@ def func(a=3, b=4):
...
+        dummyParameters = parameters.copy()
+        dummyParameters['c'] = FakeParameter("fake", "fake")
+        fakeSig = FakeSignature(dummyParameters)
+        self.assertRaises(TypeError, _passedSignature, fakeSig, (1, 2), {})
+    if not getattr(inspect, "signature", None):
+        test_invalidParameterType.skip = "inspect.signature() not available"
+
+
+if sys.version_info >= (3,):
+    _path = FilePath(__file__).parent().child("_deprecatetests.py.3only")
```
exarkun, line 1102: `twisted.python.compat._PY3`?

Also huh.  Can I read about this pattern somewhere?

src/twisted/python/test/test_deprecate.py
```diff
@@ -1067,3 +1068,56 @@ def func(a=3, b=4):
             return a + b
 
         self.assertRaises(TypeError, func, a=3, b=4)
+
+
+    def test_invalidParameterType(self):
+        """
+        Create a fake signature with an invalid parameter
+        type to test error handling.
```
exarkun: To what end?  What are the desired error behaviors being exercised here?  "_passedSignature raises TypeError when ..." is a good pattern.

src/twisted/python/test/_deprecatetests.py.3only
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
```
exarkun, line 3: Will this file actually get included in the distribution?  Having a Python source file without a ".py" extension sounds like it will trip over all kinds of weirdness.

Does the CI system exercise installation of this code somewhere?

src/twisted/python/test/_deprecatetests.py.3only
```diff
@@ -0,0 +1,58 @@
...
+    def test_passedKeywordOnly(self):
+        """
+        Keyword only arguments follow varargs.
+        They are specified in PEP 3102.
+        """
+        def func1(*a, b=True):
+            """
+            b is a keyword-only argument, with a default value.
+            """
+            pass
```
exarkun: The `pass` is unnecessary here.  The docstring serves the syntactic purpose of having _something_ in the body.

exarkun (changes requested):
Thanks.  Basically looks good now but I have some concerns about incredibly boring packaging details  related to the new non-.py source file.  Hopefully those concerns are groundless and with a bit more explanation this can land mostly as-is.

src/twisted/python/test/test_deprecate.py
rodrigc: The buildbots create the test environments by calling tox.  tox in turn creates a virtualenv, and installs Twisted into that virtualenv via setup.py, so the installer and this stuff are tested.
I copied this pattern from what @hawkowl did here: https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/test/test_coroutines.py#L38

rodrigc:
@exarkun regarding the packaging details, this is a pattern that @hawkowl started
https://github.com/twisted/twisted/blame/trunk/MANIFEST.in#L38

Since twistedchecker only works on Python 2, and running Python3/pylakes against the twisted codebase will fail in certain areas because the code is not fully ported to Py3,
during automation (Travis), pyflakes and twistedchecker are run in Python2 environments against the Twisted codebase.

So, this trick that @hawkowl introduced is the only way to get twistedchecker to pass on Python 2 with these types of Python 3 only language constructs.

Long-term, we need to port all of Twisted to Py3, and also port tools like twistedchecker
and pydoc to Py3, but that will take time. :)

src/twisted/python/test/_deprecatetests.py.3only
rodrigc: See my explanation here: https://github.com/twisted/twisted/pull/680#issuecomment-275959440

This CI systems exercises installation of this since we use tox, and
tox installs into a chroot via setup.py, which uses the MANIFEST.in change that @hawkowl made, so this does get installed in the distribution.

src/twisted/python/test/_deprecatetests.py.3only
rodrigc: OK

src/twisted/python/test/test_deprecate.py
rodrigc: So @hawkowl didn't use _PY3 when she originally added this for awaittests,
because for awaittests, she needed to Check for Python >= 3.5.
While not quite necessary, I just followed the same convention.

src/twisted/python/test/test_deprecate.py
rodrigc: I put this test in here so that codecov would hit this line: https://github.com/twisted/twisted/blob/fd5d5e005d7065cf516cd17e4565161272831e4d/src/twisted/python/deprecate.py#L741
Setting param.kind to some dummy value that is not in the enumeration of inspect.Parameter values was the only way I could figure out how to do it.

src/twisted/python/test/test_deprecate.py
```diff
@@ -961,7 +962,12 @@ def checkPassed(self, func, *args, **kw):
         @return: L{_passed}'s return value
         @rtype: L{dict}
         """
-        return _passed(inspect.getargspec(func), args, kw)
+        if getattr(inspect, "signature", None):
```
exarkun: (required)
The docstring above refers to `passed` which no longer exists (and .. didn't before, since it was `_passed`).

(optional)
I kinda also want to say something about the duplication between the logic here and the logic inside `_mutuallyExclusiveArguments`.  It seems like it might be better if there were just one implementation of that idea and these tests could exercise it.

src/twisted/python/test/test_deprecate.py
exarkun: Okay.  Put that information into the code so it's available to future maintainers. :)

src/twisted/python/test/_deprecatetests.py.3only
exarkun: (required)
This should be written up in the Python 3 porting guide, I think. Note that one thing learned during initial porting efforts was that splitting files into Python 2 / Python 3 caused a lot of problems (conflict related). Perhaps this pattern is only going to be used for new files, though, and so that experience is not relevant. Just mentioning it.

There is some code in Twisted that employs an alternate idiom, putting the source into a string and then exec()ing it when necessary. I think it would be good to employ one or the other of these - not both.

exarkun (approved):
Thanks.  A couple more points, nothing major.  Please address them and then merge.

src/twisted/python/test/_deprecatetests.py.3only
rodrigc: Right now, these arcane contortions are needed to make forward progress, and there is no other clean way.  As time goes on, I would like to see twistedchecker and pydoctor work on Py3, so that we can run them on Py3 as part of the automation on Travis.
That way, people can just focus on writing code, and not have to do
contortions to appease broken tools.

I've submitted a few patches to both those tools to get the ball rolling:

https://github.com/twisted/pydoctor/pulls/rodrigc
https://github.com/twisted/twistedchecker/pulls/rodrigc

exarkun: This seems orthogonal to the request to document the process, though?

PR #681, opened 2017-01-16
Add support for delayed connection abort after timeout to HTTPChannel and H2Connection

Lukasa:
See https://twistedmatrix.com/trac/ticket/8902


src/twisted/test/proto_helpers.py
```diff
@@ -122,6 +122,9 @@ class StringTransport:
     @ivar disconnecting: A C{bool} which is C{False} until L{loseConnection} is
         called, then C{True}.
 
+    @ivar aborted: A L{bool} which is C{False} until L{abortConnection} is
```
glyph: A thought: perhaps this should document which interface it's coming from?

src/twisted/test/proto_helpers.py
```diff
@@ -159,6 +162,7 @@ class StringTransport:
     """
 
     disconnecting = False
+    aborted = False
```
glyph: This expands the interface of `StringTransport` further in a bad direction - to export an apparently public attribute that isn't part of its interface.

There's already another public attribute you could use for this purpose - `disconnected` - https://github.com/twisted/twisted/blob/aeeffc7a94880f156d30526b73c89348b6948db0/src/twisted/internet/abstract.py#L201  This attribute isn't really documented or explained well either, but at least it sticks to existing crummy interfaces and doesn't drag in a new one just for testing [that we really shouldn't have](https://glyph.twistedmatrix.com/2015/05/separate-your-fakes-and-your-inspectors.html)

src/twisted/test/proto_helpers.py
```diff
@@ -222,8 +226,10 @@ def loseConnection(self):
 
     def abortConnection(self):
         """
-        Abort the connection. Same as C{loseConnection}.
+        Abort the connection. Same as C{loseConnection}, but also toggles the
+        C{aborted} instance variable to C{True}.
```
glyph, line 230: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/test/proto_helpers.py
```diff
@@ -222,8 +226,10 @@ def loseConnection(self):
...
         """
+        self.aborted = True
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # This is a terrible violation of the abstraction later of
+        # _genericHTTPChannelProtocol, but we need to do it because
+        # policies.TimeoutMixin doesn't accept a reactor on the object.
+        # See https://twistedmatrix.com/trac/ticket/8488
+        protocol._channel.callLater = clock.callLater
+        protocol.makeConnection(transport)
+        protocol.dataReceived(b'POST / HTTP/1.0\r\nContent-Length: 2\r\n\r\n')
+        self.assertFalse(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Force the initial timeout.
+        clock.advance(60)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Watch the transport get force-closed.
+        clock.advance(14)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+        clock.advance(1)
+        self.assertTrue(transport.disconnecting)
+        self.assertTrue(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # This is a terrible violation of the abstraction later of
+        # _genericHTTPChannelProtocol, but we need to do it because
+        # policies.TimeoutMixin doesn't accept a reactor on the object.
+        # See https://twistedmatrix.com/trac/ticket/8488
+        protocol._channel.callLater = clock.callLater
+        protocol.makeConnection(transport)
+        protocol.dataReceived(b'POST / HTTP/1.0\r\nContent-Length: 2\r\n\r\n')
+        self.assertFalse(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Force the initial timeout.
+        clock.advance(60)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Move forward nearly to the timeout, then fire connectionLost.
+        clock.advance(14)
+        protocol.connectionLost(None)
+
+        # Check that the transport isn't forcibly closed.
+        clock.advance(1)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # This is a terrible violation of the abstraction later of
+        # _genericHTTPChannelProtocol, but we need to do it because
+        # policies.TimeoutMixin doesn't accept a reactor on the object.
+        # See https://twistedmatrix.com/trac/ticket/8488
+        protocol._channel.callLater = clock.callLater
+        protocol.makeConnection(transport)
+        protocol.dataReceived(b'POST / HTTP/1.0\r\nContent-Length: 2\r\n\r\n')
+        self.assertFalse(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Force the initial timeout.
+        clock.advance(60)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+
+        # Move an absurdly long way just to prove the point.
+        clock.advance(2**32)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
+            errorCode=h2.errors.NO_ERROR,
+            lastStreamID=0
+        )
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
+
+        # Advance the clock to see that we abort the connection.
+        reactor.advance(14)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
+        reactor.advance(1)
+        self.assertTrue(transport.disconnecting)
+        self.assertTrue(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
+        self.assertFalse(transport.aborted)
+
+        # Advance the clock, but right before the end fire connectionLost.
+        reactor.advance(14)
+        conn.connectionLost(None)
+
+        # Check that the transport isn't forcibly closed.
+        reactor.advance(1)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
+        reactor.advance(100)
+
+        self.assertTimedOut(
+            transport.value(),
+            frameCount=2,
+            errorCode=h2.errors.NO_ERROR,
+            lastStreamID=0
+        )
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
+
+        # Advance the clock an arbitrarily long way, and confirm it never
+        # aborts.
+        reactor.advance(2**32)
+        self.assertTrue(transport.disconnecting)
+        self.assertFalse(transport.aborted)
```
glyph: see above about `aborted` (adding a comment here so it's easy to track)

src/twisted/web/_http2.py
```diff
@@ -103,11 +103,23 @@ class H2Connection(Protocol, TimeoutMixin):
     @ivar _sender: A handle to the data-sending loop, allowing it to be
         terminated if needed.
     @type _sender: L{twisted.internet.task.LoopingCall}
+
+    @ivar abortTimeout: The number of seconds to wait after we attempt to shut
+    the transport down cleanly to give up and forcibly terminate it. This
```
glyph: Not a blocker, but, to nitpick here, it would be nice if the indentation were consistent with the surrounding parameters (including the one you added :))

src/twisted/web/_http2.py
```diff
@@ -213,10 +225,33 @@ def timeoutConnection(self):
         self.conn.close_connection(error_code=error_code)
         self.transport.write(self.conn.data_to_send())
 
+        # Don't let the client hold this connection open too long.
+        if self.abortTimeout is not None:
+            # We use self.callLater because that's what TimeoutMixin does, even
+            # though we have a perfectly good reactor sitting around. See
+            # https://twistedmatrix.com/trac/ticket/8488.
```
glyph, line 232: Thanks for calling this out.

src/twisted/web/_http2.py
```diff
@@ -103,11 +103,23 @@ class H2Connection(Protocol, TimeoutMixin):
     @ivar _sender: A handle to the data-sending loop, allowing it to be
         terminated if needed.
     @type _sender: L{twisted.internet.task.LoopingCall}
+
+    @ivar abortTimeout: The number of seconds to wait after we attempt to shut
```
glyph, line 107: Does this really need to be a public attribute?

glyph: (Even if H2Connection itself isn't exposed - I'm not sure that it is - it would be nice to be consistent with the surrounding variables, if there's no need for it to be public.)

src/twisted/web/_http2.py
```diff
@@ -103,11 +103,23 @@ class H2Connection(Protocol, TimeoutMixin):
...
+    the transport down cleanly to give up and forcibly terminate it. This
+    is only used when we time a connection out, to prevent errors causing
+    the FD to get leaked. If this is L{None}, we will wait forever.
+    @type abortTimeout: L{int}
+
+    @ivar _abortingCall: The L{twisted.internet.base.DelayedCall} that will be
```
glyph: staaaaaaaate machiiiiiiiiiines

src/twisted/web/test/test_http.py
```diff
@@ -242,6 +242,107 @@ def test_requestBodyDefaultTimeout(self):
...
+        """
+        If a timed out transport ends up calling C{connectionLost}, it prevents
+        the force-closure of the transport.
+        """
+        clock = Clock()
+        transport = StringTransport()
+        factory = http.HTTPFactory()
+        protocol = factory.buildProtocol(None)
+
+        # This is a terrible violation of the abstraction later of
```
glyph: Could you perhaps put this into a function, so you don't have a copy/pasted comment lying around?  Something like `factory = parameterizeTimeoutMixin(HTTPFactory, reactor=clock)`?  Add parens to taste.

src/twisted/web/test/test_http2.py
```diff
@@ -2470,3 +2470,102 @@ def test_noTimeoutIfConnectionLost(self):
...
         reactor.advance(101)
         self.assertEqual(transport.value(), sentData)
+
+
+    def test_timeoutEventuallyForcesConnectionClosed(self):
+        """
+        When a L{H2Connection} has timed the connection out, and the transport
+        doesn't get torn down within 15 seconds, it gets forcibly closed.
+        """
+        frameFactory = FrameFactory()
```
glyph: These 3 tests share their first ~20 lines of code.  Any chance you could put this into a function to avoid the maintenance burden of updating all of them the next time someone rolls in here?

src/twisted/web/_http2.py
```diff
@@ -103,11 +103,23 @@ class H2Connection(Protocol, TimeoutMixin):
     @ivar _sender: A handle to the data-sending loop, allowing it to be
         terminated if needed.
     @type _sender: L{twisted.internet.task.LoopingCall}
+
+    @ivar abortTimeout: The number of seconds to wait after we attempt to shut
+    the transport down cleanly to give up and forcibly terminate it. This
+    is only used when we time a connection out, to prevent errors causing
+    the FD to get leaked. If this is L{None}, we will wait forever.
```
glyph: Is "wait forever" even really a valid use-case here?  Why allow it to be `None`?  An `int` can hold a very, very large value; surely even waiting 2**64 seconds could be functionally forever.

glyph: It might also be helpful here to note that this is the timeout _from_ the time that the other timeout has fired until `abortConnection`, not the amount of time from the start.

glyph (changes requested):
Thanks for fixing this bug, @Lukasa !

Begging your pardon though, I'm going to ask for a re-review on this one, mainly because I have concerns about test suite maintainability, and this leaks some new public interface stuff.  I'm reasonably happy with the implementation though so poke me when you have addressed the feedback and hopefully the next review can be a quick rubber stamp.

src/twisted/test/proto_helpers.py
Lukasa: Is disconnected suitable here? I'm not looking for whether the connection is actually gone, I'm looking explicitly for whether we called `abortConnection`.

src/twisted/web/_http2.py
Lukasa: `timeOut` is a public attribute, and is in fact set by the `Site` object, so I was being consistent with that. This does seem very much like the kind of thing users may want to change.

src/twisted/test/proto_helpers.py
Lukasa: The interface what is coming from, aborted? I don't think it's on any interface: I *ad hoc*'d it up.

src/twisted/web/_http2.py
Lukasa: I think the first sentence does say that: "The number of seconds to wait after we attempt to shut the transport down cleanly to give up and forcibly terminate it". Do you want this reworded?

src/twisted/test/proto_helpers.py
glyph: `disconnecting` is set synchronously; `disconnected` is not set synchronously by `loseConnection`, but it _is_ set by `abortConnection`, because `abortConnection` actually causes the connection to go away.

Inasmuch as these existing flags have clear semantics, `disconnected` means "there's no transport any more", which is more or less what you're asking `abortConnection` to accomplish.

Lukasa:
Uh...what the hell, AppVeyor?

glyph (approved):

glyph:
@Lukasa - AppVeyor looks happy to me

PR #683, opened 2017-01-17
Port FTP protocol to Python 3

leth:
Trac: https://twistedmatrix.com/trac/ticket/8865
I'd update the Trac ticket, but login seems to be broken!


src/twisted/test/test_ftp_options.py
```diff
@@ -19,15 +19,15 @@ class FTPOptionsTests(TestCase):
     Tests for the command line option parser used for C{twistd ftp}.
     """
 
-    usernamePassword = ('iamuser', 'thisispassword')
+    usernamePassword = (b'iamuser', b'thisispassword')
```
leth, line 22: I'm not certain whether these should really be bytes; if they weren't it'd need a change to `twisted.cred.checkers:FilePasswordDB`.

Any advice would be appreciated!

src/twisted/protocols/ftp.py
```diff
@@ -736,10 +739,20 @@ class FTP(basic.LineReceiver, policies.TimeoutMixin, object):
...
 
     def reply(self, key, *args):
         msg = RESPONSE[key] % args
-        self.sendLine(msg)
+        self.sendEncodedLine(msg)
+
+
+    def sendEncodedLine(self, line):
+        """
+        (Private) Encodes and sends a line
```
rodrigc, line 755: The convention in Twisted is that if a function is private, it's name should be
prefixed by an underscore.  However, sendEncodedLine( is not needed.  Instead, delete this function and modify sendLine()

src/twisted/protocols/ftp.py
```diff
@@ -2485,11 +2503,27 @@ def _cb_greeting(self, greeting):
     def sendLine(self, line):
         """
         (Private) Sends a line, unless line is None.
+
+        @param line: Line to send
+        @type line: bytes
         """
         if line is None:
             return
```
rodrigc, line 2518: Change sendLine so that right here it does:

```
if isinstance(line, unicode):
    line = line.encode(_encoding)
```

src/twisted/protocols/ftp.py
```diff
@@ -2485,11 +2503,27 @@ def _cb_greeting(self, greeting):
     def sendLine(self, line):
         """
         (Private) Sends a line, unless line is None.
+
+        @param line: Line to send
+        @type line: bytes
```
rodrigc: change this to:

```
@type line: L{bytes} or L{unicode}
```

src/twisted/protocols/ftp.py
```diff
@@ -2593,6 +2627,7 @@ def lineReceived(self, line):
         (Private) Parses the response messages from the FTP server.
         """
         # Add this line to the current response
+        line = line.decode(self._encoding)
```
rodrigc: This is OK for Python 3, but on Python 2 this is now making line a ```unicode``` on Python 2, when before it was a ```str```.  This might not be what we want.  Can you change this to:

```
if isinstance(line, bytes) and bytes != str:
    line = line.decode(self._encoding)
```

src/twisted/protocols/ftp.py
```diff
@@ -3155,12 +3190,14 @@ class FTPFileListProtocol(basic.LineReceiver):
...
     )
-    delimiter = '\n'
+    delimiter = b'\n'
+    _encoding = 'latin-1'
 
     def __init__(self):
         self.files = []
 
     def lineReceived(self, line):
+        line = line.decode(self._encoding)
```
rodrigc:
```
if isinstance(line, bytes) and bytes != str:
    line = line.decode(self._encoding)
```

rodrigc (changes requested):

rodrigc:
Can you modify one of the tests so that `sendLine()` is called with:

1. a `None`
2. a byte string: `b"whatever`
3. a native string `"whatever"`

That will increase the coverage count.

rodrigc:
As an additional check, can you create a Python 2.7 (must be 2.7 virtualenv) and run inside the virtualenv, while in the top-level directory of your twisted checkout:

```
python -m pip install tox
python -m tox -e pyflakes
python -m tox -e txchecker-travis-all
```

and fix any errors that you see.  I think right now, the patch is clean, but it is good to double check.

src/twisted/protocols/ftp.py
leth: IIRC `sendLine` is a method on the superclass; I didn't want to mess with that one, since it'd prevent sending unencoded lines. I'll take another look at this, thanks.

src/twisted/protocols/ftp.py
leth: Ah I hadn't realised that'd perform differently, good spot.
Would it be it be less confusing to just say `if PYTHON2:` or is the convention to test for features instead?

src/twisted/protocols/ftp.py
leth: I'm more than happy to change the name, I was following what the rest of the code in this file did.

rodrigc:
No, we want to avoid putting checks for PYTHON2

rodrigc:
It's safe to modify sendLine as I mentioned.  Look at twisted/words/protocols/irc.py for an example

leth:
Sorry, I've rebased it because the checker was flagging up fixes in areas I'd not changed.

src/twisted/test/test_ftp_options.py
rodrigc: Changing those to bytes is fine.

rodrigc:
Your latest changes look good.
Can you add two more tests to increase coverage?

(1)  Can you create a test that hits the `FTPOverflowProtocol. connectionMade()` function?
(2) For the FTP.lineReceived.processFailed() function, can you create a test which exercises
      the **Unexpected FTP error** branch of the if statement?

Other than that, this is close to being good to go.

leth:
That should do it!
Could you update Trac? The login link still does nothing when I click it...

rodrigc:
Your latest changes look good.  I downloaded your changes and did some further
testing using Python 3.6 on MacOS X.  When I ran:

I saw these test failures:

```
twisted.test.test_ftp.FTPServerPasvDataConnectionTests.test_NLSTNonASCIIBytes ... [FAILURE]
twisted.test.test_ftp.FTPServerPasvDataConnectionTests.test_NLSTUnicode ... [FAILURE]
twisted.test.test_ftp.FTPServerPortDataConnectionTests.test_NLSTNonASCIIBytes ... [FAILURE]
twisted.test.test_ftp.FTPServerPortDataConnectionTests.test_NLSTUnicode ... [FAILURE]
twisted.test.test_ftp.IsGlobbingExpressionTests.test_isGlobbingExpressionNoGlob ... [FAILURE]
```

Could you take a look at the encoding related failures?
For the globbing related failures, it looks like fnmatch() returns different things on Python 2 and Python 3.

rodrigc:
Don't lose hope, you've done a lot of good work and are close!!

leth:
Oh, why is travis/tox not testing python3.6?
I'll take a look at it tomorrow, but if it's a mac specific problem then I won't be able to test it.

rodrigc:
I saw the same problem on Linux with Python 3.6: https://buildbot.twistedmatrix.com/builders/ubuntu16.04-py3.6/builds/224

leth:
Looks like they were all globbing problems!

rodrigc:
Fixed in https://github.com/twisted/twisted/pull/687

Thanks!!

rodrigc:
Fixed in https://github.com/twisted/twisted/pull/687

Thanks!!

PR #685, opened 2017-01-18
Fixes for `ClientService`

markrwilliams:
https://twistedmatrix.com/trac/ticket/8993

This is #677 with tests and a minor change to the state machine.

### DIFF
--- a/src/twisted/application/internet.py
+++ b/src/twisted/application/internet.py
@@ -830,8 +830,9 @@ class _ClientMachine(object):
 
     _init.upon(start, enter=_connecting,
                outputs=[_connect])
-    _init.upon(stop, enter=_init,
-               outputs=[])
+    _init.upon(stop, enter=_stopped,
+               outputs=[],
+               collector=lambda _: succeed(None))
 
     _connecting.upon(start, enter=_connecting, outputs=[])
     # Note that this synchonously triggers _connectionFailed in the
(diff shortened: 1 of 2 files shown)


src/twisted/application/internet.py
```diff
@@ -831,8 +841,8 @@ def _unawait(self, value):
     _init.upon(start, enter=_connecting,
                outputs=[_connect])
     _init.upon(stop, enter=_stopped,
-               outputs=[],
-               collector=lambda _: succeed(None))
+               outputs=[_deferredSucceededWithNone],
```
tomprince, line 844: This change doesn't seem to add any clarity to what the code is doing, and add more code that needs to be reasoned about.

src/twisted/application/test/test_internet.py
```diff
@@ -860,3 +860,103 @@ def test_retryCancelled(self):
...
+                                          startService=False,
+                                          clock=clock)
+        d = service.stopService()
+        self.assertIsNone(self.successResultOf(d))
+
+
+    def test_whenConnectedErrbacksOnStopService(self):
+        """
+        L{ClientService.whenConnected} returns a L{Deferred} that
+        errbacks with L{CancelledError} if the connection fails before
```
tomprince: This could do with a better explanation. It is specifically when it is called between connection attempts or during the delay before reconnecting that this is testing.

src/twisted/application/internet.py
markrwilliams: I added it to confirm that all `stop`s have outputs.  I found it made the state chart clearer, too:

image

glyph (approved):
This looks good to me.

For the record, I agree with @tomprince that many of these tests should be deleted once the hypothesis tests which cover the same behavior are landed.

PR #686, opened 2017-01-18
Avoid incorrect streaming-producer resume in IOPump

exarkun:
https://twistedmatrix.com/trac/ticket/9003


src/twisted/test/test_iosim.py
```diff
@@ -39,3 +44,85 @@ def test_writeSequence(self):
         a.writeSequence([b"b", b"c", b"d"])
 
         self.assertEqual(b"".join(a.stream), b"abcd")
+
+
+
+@implementer(IPushProducer)
+class StrictPushProducer(object):
```
markrwilliams, line 51: This might be a good candidate for a state machine.

src/twisted/test/test_iosim.py
```diff
@@ -39,3 +44,85 @@ def test_writeSequence(self):
...
+    """
+    def _testStreamingProducer(self, mode):
+        """
+        Connect a couple protocol/transport pairs to an L{IOPump} and then pump
+        it.  Verify that a streaming producer registered with one of the
+        transports does not receive invalid L{IPushProducer} method calls and
+        ends in the right state.
+        """
+        serverProto = Protocol()
+        serverTransport = FakeTransport(serverProto, isServer=False)
```
markrwilliams: I think `t.t.iosim.FakeTransport.isServer` only affects its `startTLS` method, so `isServer=False` here doesn't actually matter.  It's a little surprising though.

src/twisted/test/test_iosim.py
```diff
@@ -39,3 +44,85 @@ def test_writeSequence(self):
...
+            )
+        self._state = u"running"
+
+
+
+class IOPumpTests(TestCase):
+    """
+    Tests for L{IOPump}.
+    """
+    def _testStreamingProducer(self, mode):
```
markrwilliams, line 267: I'm going to guess that twistedchecker yells about this being undocumented, unless we disabled it again.

markrwilliams (approved):
Some mostly idle thoughts.  Please add a topfile and address any twistedchecker errors, then merge.

_EDIT_ I suppose you need to add coverage for `StrictProducer`: https://codecov.io/gh/twisted/twisted/compare/06e19509e57a0156bb4d1c84b73d0eacada6384c...c15462c896fd8f5712a698d75bb8cacceab2e41e?src=pr

src/twisted/test/test_iosim.py
exarkun: Probably.  Automat won't let you inspect the current state of the machine though, which makes it a little annoying.  I'd basically have to continue to maintain the same manual state as the code maintains now, plus have the automat stuff.  Probably not a net win.

(@glyph)

markrwilliams: If you'd like to know the state just to make errors more informative, Automat now does that:

```python
from automat import MethodicalMachine
from twisted.internet.interfaces import IPushProducer


def _returnNone(outputs):
    list(outputs)


@implementer(IPushProducer)
class StrictPushProducer(object):
    _machine = MethodicalMachine()

    @_machine.state(initial=True)
    def running(self):
        pass

    @_machine.state()
    def stopped(self):
        pass

    @_machine.state()
    def paused(self):
        pass

    @_machine.input()
    def stopProducing(self):
        pass

    @_machine.input()
    def pauseProducing(self):
        pass

    @_machine.input()
    def resumeProducing(self):
        pass

    running.upon(pauseProducing, enter=paused, outputs=[],
                 collector=_returnNone)
    running.upon(stopProducing, enter=stopped, outputs=[],
                 collector=_returnNone)
    paused.upon(resumeProducing, enter=running, outputs=[],
                collector=_returnNone)
    paused.upon(stopProducing, enter=stopped, outputs=[],
                collector=_returnNone)


validTransitions = StrictPushProducer()
validTransitions.pauseProducing()
validTransitions.resumeProducing()
validTransitions.stopProducing()

invalidTransition = StrictPushProducer()
invalidTransition.resumeProducing()
```

`invalidTransition` will fail like so:
```
  File "/tmp/machine.py", line 54, in <module>
    invalidTransition.resumeProducing()
  File "/home/mrw/src/python/automat/automat/_methodical.py", line 97, in resumeProducing
    outputs = transitioner.transition(self)
  File "/home/mrw/src/python/automat/automat/_core.py", line 147, in transition
    inputSymbol)
  File "/home/mrw/src/python/automat/automat/_core.py", line 129, in outputForInput
    raise NoTransition(state=inState, symbol=inputSymbol)
automat._core.NoTransition: no transition for MethodicalInput(method=<function resumeProducing at 0x7fcb3ef65938>) in MethodicalState(method=<function running at 0x7fcb3ef65500>)
```

The reporting could be improved but I believe it provides the same information the current implementation of `StrictPushProducer` does.

exarkun: Yea, without the ability to customize this exception at all, I don't think it's appropriate.  This is a good exception to tell the state machine maintainer that they screwed up.  Otherwise, it's an implementation detail leaking out.

glyph:
Did I miss something in an outdated comment?  Why was this merged with failing a twistedchecker build?

src/twisted/test/test_iosim.py
glyph: I filed an issue: https://github.com/glyph/automat/issues/46

exarkun:
Sorry, I should have left a comment I guess?  The twistedchecker build was failing with a known twistedchecker bug: https://github.com/twisted/twistedchecker/issues/124

So I merged without fixing it, per https://twistedmatrix.com/trac/wiki/ReviewProcess#Reviewers:Howtoreviewachange ("There is one caveat ...")

Maybe `ReviewProcess` should direct the author to leave some kind of indication that this case has been encountered?

glyph:
The relevant phrase from ReviewProcess that was not followed is "and then link from the bug report to the twisted ticket where you saw the spurious failure".  Until this comment, https://github.com/twisted/twistedchecker/issues/124 didn't contain any reference to this PR or the attendant ticket.  The goal of this stipulation is to ensure that tool bugs that are encountered regularly are visible, and have lots of comments on them so we can see they are causing drag on development and motivate effort to fix them appropriately.

This could be clearer, however.  It should probably also say to explicitly call out recognition of the build failure on the ticket or PR for the merge, so that folks like yours truly don't come along and poke people unnecessarily :).

exarkun:
> so that folks like yours truly don't come along and poke people unnecessarily :).

I'm glad there's some supervision.  Sorry about the omission on my part.  I think I read the _beginning_ and the _end_ of the line from ReviewProcess 10 times but somehow managed to never read the middle until you pointed it out.

PR #690, opened 2017-01-23
(in-org) twisted.internet.defer.extractCurrentResult

exarkun:
See #590


src/twisted/internet/defer.py
```diff
@@ -160,6 +171,151 @@ def maybeDeferred(f, *args, **kw):
...
+    the original exception unless the L{raises} parameter is set to L{False}.
+
+    The primary purpose of this function is to allow code to be written that is
+    internally asynchronous but which can be consumed without an event loop in
+    a synchronous fashion by using L{extractCurrentResult} at the boundaries.
+    This enables libraries to be written that can function both with Twisted
+    and with synchronous code instead of requiring an asynchronous and a
+    synchronous implementation for all functionality.
+
+    @attention: This function CANNOT take a Deferred with pending asynchronous
```
exarkun, line 193: Apparently pydoctor doesn't like `@attention`:

```
found unknown field on 'twisted.internet.defer.extractCurrentResult': <Field 'attention' None [u'This function CAN...']>
found unknown field on 'twisted.internet.defer.observeCurrentResult': <Field 'attention' None [u'This function CAN...']>
```

Which is too bad, since epydoc itself seems okay with this, at least according to <http://epydoc.sourceforge.net/fields.html>.

I suppose the thing to do is leave this as-is and file a pydoctor bug to make this work.

src/twisted/internet/defer.py
```diff
@@ -53,6 +53,17 @@ class TimeoutError(Exception):
 
 
 
+class NoCurrentResult(BaseException):
+    """
+    This error is raised when attempting to call L{extractCurrentResult} on a
```
Julian, line 59: Looks like now this is raised by both functions. Is mentioning one better than either mentioning both or neither and just describing the situation (I have no opinion, so maybe it's fine as is)

src/twisted/internet/defer.py
```diff
@@ -53,6 +53,17 @@ class TimeoutError(Exception):
...
+    L{Deferred} that has not yet called all of its callbacks.
+
+    This error inherits from L{BaseException} rather than L{Exception} to
+    reduce the likelihood of inadvertingly catching this exception as part of
+    an exception handler for other errors.
+    """
```
exarkun, line 65: Okay so basically the pattern being anticipated is ...

```
try:
    blah = extractCurrentResult(d)
except Exception as e:
    # Handle the failure result of the Deferred
```

And the idea is to allow `NoCurrentResult` to pass through this because it's not the failure result of `d` but an expectation failure in this example code itself.

Won't the application code want to handle `NoCurrentResult` somehow though?  I'd expect the pattern you'd see much more commonly is ...

```
try:
    blah = extractCurrentResult(d)
except NoCurrentResult:
    # Do ... uh ... something ...
except Exception as e:
    # Handle the failure result of the Deferred
```

Okay, that's not a great example since I can't quite imagine what you'd do in the `NoCurrentResult` handler.  It seems like you'd want to do some kind of translation to a more application-relevant exception.  You don't want `NoCurrentResult` bubbling to the top, killing your program, and dumping a gnarly stack trace at the end user, do you?

Ideally, you really just want this to be a compiler error, in fact... but nevermind.

Also, it seems like `BaseException` suffers from a dilution effect over time.  The more of these exceptions there are (_especially_ if folks actually do let them escape like the first example above does), the more tempting it is to write the first example like ...

```
try:
    blah = extractCurrentResult(d)
except:
    # Handle the failure result of the Deferred
```

which is absolutely awful unless you want to write some code that doesn't get killed by a surprising `BaseException` somewhere. :)

So ... I don't know ... Some sample application code would probably go some way towards revealing whether the second example is legit or not.  If it is, I think `Exception` is an alright base class for this exception.

exarkun (commented):
Thanks.  Apart from a question about `NoCurrentResult` this looks pretty good.

Do you have some code that might shed some light on the question there?

src/twisted/internet/defer.py
dstufft: My assumption is that getting a ``NoCurrentResult`` is going to mean that you need to fix your code and isn't something that you're generally going to be catching. I modeled it somewhat after ``GeneratorExit`` in that it's a ``BaseException`` primarily because it's not expecting to get caught. That being said, it is arguably closer in spirit to something like the error that is thrown when you hit a recursion limit or a ``NameError`` or something along that line, which are all inherited from ``Exception``.

Ultimately though, it's really because in my mind, I couldn't think of a case when you would be expecting to catch the ``NoCurrentResult`` error, except that you might catch it by accident. I'm happy to go either way with it though.

glyph:
I'm a little confused about the status of this PR.  Why'd the other one get closed?

glyph:
(The circleci failure is just that `topfiles` is now called `newsfragments`.)

exarkun:
> I'm a little confused about the status of this PR. Why'd the other one get closed?

Note I didn't get any notification about this comment.  I merely happened upon it while trying to remind myself of the state of this work for other reasons.

The other one was closed because it was from a different org so didn't get the full CI run and either the tooling to generate CI results in that case didn't exist at the time or I forgot about them.

src/twisted/internet/defer.py
Julian: Perhaps code in a loop is a more reasonable example there?

I.e.

```
while True:
    do_dumb_sleep_poll-y_stuff_repeatedly()
    try:
        blah = extractCurrentResult(d)
    except NoCurrentResult:
        continue
    except Exception:
        handle_failure()
    else:
        handle_success()

File: src/twisted/internet/defer.py [hunk 4]
```diff
@@ -161,6 +172,151 @@ def maybeDeferred(f, *args, **kw):
...
+    This enables libraries to be written that can function both with Twisted
+    and with synchronous code instead of requiring an asynchronous and a
+    synchronous implementation for all functionality.
+
+    @attention: This function CANNOT take a Deferred with pending asynchronous
+        work to be done and make it synchronous. It can ONLY extract results
+        from a L{Deferred} which has already fired its callbacks.
+
+    @type deferred: Deferred
+    @param deferred: The Deferred from which results will be extracted..
```
Julian, line 198: Double period here.

src/twisted/internet/defer.py
```diff
@@ -161,6 +172,151 @@ def maybeDeferred(f, *args, **kw):
...
+
+    @type raises: bool
```
Julian, line 200: This boolean flag still makes me uneasy personally but if others are ok with it maybe at least calling it `reraise` reads a bit more straightforwardly? `extractCurrentResult(raises=True)` vs. `extractCurrentResult(reraise=True)`

src/twisted/internet/defer.py
```diff
@@ -161,6 +172,151 @@ def maybeDeferred(f, *args, **kw):
...
+
+    # We need a place to store the result, and a way to determine the
+    # difference between no result and any other possible object. So we'll
+    # create a sigil object and stash it locally and default our result to
+    # that.
+    # Note: This *has* to happen before adding the done callback to the
+    #       Deferred or else the callback can fire prior to setting our
+    #       default, thus causing the default to override the value from the
+    #       callback.
+    nothing = done.result = object()
```
Julian, line 220: Maybe having just one of these (and stashing it as e.g. `NoCurrentResult._SENTINEL` or whatever) is a bit shorter than creating it each time?

src/twisted/internet/defer.py
```diff
@@ -161,6 +172,151 @@ def maybeDeferred(f, *args, **kw):
...
+
+    # If we've gotten to this point and we haven't had our collect function
+    # called, then we must be waiting on something in the Deferred chain. If so
+    # we'll error out here because there's nothing we can do to advance the
+    # state of the Deferred. If we've got a result then we'll go ahead and
+    # return it or raise the given exception.
+    if done.result is nothing:
+        raise NoCurrentResult(
+            "Cannot observe the current result of a Deferred that has not "
+            "currently called all of its callbacks."
```
Julian, line 310: Maybe s/currently/yet reads a bit clearer?

src/twisted/test/test_defer.py
```diff
@@ -676,6 +676,257 @@ def test_cancelGatherResultsWithAllDeferredsCallback(self):
...
+            try:
+                defer.extractCurrentResult(d)
+            except Exception:
+                return "error"
+            return "success"
+
+        d = defer.Deferred()
+        self.assertRaises(defer.NoCurrentResult, inner, d)
+
+        # In order to get 100% patch coverage we need to excerise the "error"
```
Julian, line 732: Do we not use `pragma: no cover`? If we do we should just use that for this specific case instead of carrying useless test code IMO.

src/twisted/topfiles/8900.feature
```diff
@@ -0,0 +1,3 @@
+The new functions twisted.internet.defer.extractCurrentResult and
+twisted.internet.defer.observeCurrentResult enables extracting or observing the
```
Julian, line 2: s/enables/enable

src/twisted/internet/defer.py
exarkun: Doesn't that seem like just the kind of use of this API that is wrong and needs to be discouraged?  This isn't how you use Deferreds...

Re-reading this discussion it seems to me like the problem to be solved is "is this an exception from the Deferred I passed in or an exception from the implementation of extractCurrentResult" and I don't think making `NoCurrentResult` a `BaseException` instead of an `Exception` accomplishes this goal.

Instead, putting some more state onto `NoCurrentResult` would make it possible to differentiate.  Attach the Deferred to it (great way to create lots of extra garbage, I guess)?  I'm not sure what else you would do.  This is an example of a perennial problem with Python exceptions - you can't reliably tell where they came from.  This particular API probably isn't going to solve the problem.

Short of some earth-moving idea, I think it makes sense to just switch to subclassing `Exception` and move on.

src/twisted/test/test_defer.py
exarkun: If NoCurrentResult changes to be an Exception then this whole thing goes away!  Yay.  Also there's a typo in "excerise".

glyph:
@dstufft Any chance you have a moment to submit a superseding PR, now that being in-org is not required for most of our CI tooling? :)

glyph:
Is anyone still interested in this functionality? I feel like events have overtaken it in various ways (semi-sync libraries are no longer popular, @dstufft clearly isn't working on this any more, `defer.py` has changed a bunch so there are conflicts now, etc)

PR #691, opened 2017-01-24
successResultOf, failureResultOf, assertNoResult should accept coroutines

wsanchez:
https://twistedmatrix.com/trac/ticket/9006


wsanchez:
I don't know what to do about `ubuntu16.04-py2.7-newstyle-coverage` failing…

wsanchez:
OK fixed that, but now the pyflakes that runs in Py2.7 fails.

src/twisted/newsfragments/9006.feature
```diff
@@ -0,0 +1,2 @@
+twisted.trial.successResultOf, twisted.trial.failureResultOf, and
```
hawkowl, line 1: these aren't actually the fqpns for these functions?

wsanchez: The fqpns are in private modules; these are (I think) the public names.

wsanchez: Like I don't think I want to say `twisted.trial._synctest._Assertions.successResultOf` here.

glyph:
Those are… not the public names of those methods.

```
>>> from twisted.trial import successResultOf
  File "<input>", line 1, in <module>
    from twisted.trial import successResultOf
ImportError: cannot import name 'successResultOf'
```

I think you meant `twisted.trial.unittest.SynchronousTestCase.successResultOf` ?

PR #692, opened 2017-01-26
Do not open file with 'U' flag on Python 3 for universal newline mode.

rodrigc:
See:
https://docs.python.org/3/library/functions.html#open

### DIFF
--- a/src/twisted/python/compat.py
+++ b/src/twisted/python/compat.py
@@ -245,7 +245,11 @@ def execfile(filename, globals, locals=None):
     """
     if locals is None:
         locals = globals
-    with open(filename, "rbU") as fin:
+    if _PY3:
+        mode = "rb"
+    else:
+        mode = "rbU"
+    with open(filename, mode) as fin:
         source = fin.read()
     code = compile(source, filename, "exec")
     exec(code, globals, locals)
--- a/src/twisted/python/compat.py
+++ b/src/twisted/python/compat.py
@@ -245,11 +245,7 @@ def execfile(filename, globals, locals=None):
     """
     if locals is None:
         locals = globals
-    if _PY3:
-        mode = "rb"
-    else:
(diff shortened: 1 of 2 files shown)


src/twisted/python/compat.py
```diff
@@ -245,7 +245,11 @@ def execfile(filename, globals, locals=None):
     """
     if locals is None:
         locals = globals
-    with open(filename, "rbU") as fin:
+    if _PY3:
+        mode = "rb"
+    else:
+        mode = "rbU"
```
Lukasa: Hang on, what the hell is this?

[The Python 2.7 documentation says](https://docs.python.org/2/library/functions.html#open) (note my emphasis):

> In addition to the standard fopen() values mode may be `'U'` or `'rU'`. Python is usually built with universal newlines support; supplying `'U'` **opens the file as a text file**, but lines may be terminated by any of the following: the Unix end-of-line convention `'\n'`, the Macintosh convention `'\r'`, or the Windows convention `'\r\n'`.

How the hell are we passing both binary mode *and* universal newline mode? This simply makes no sense. Given that the Python 3 version does not provide universal newlines, we clearly don't need them, so why don't we avoid introducing a branch altogether and just change the mode to `rb` unconditionally.

Lukasa (changes requested):
It's not at all clear to me that this fix needs to be anything more than one character.

src/twisted/python/compat.py
rodrigc: @Lukasa fine by me, let's see what the buildbots have to say.

rodrigc:
@Lukasa buildbots worked for this.  api docs build is failing due to an unrelated issue: https://github.com/twisted/twisted/pull/694

Can I proceed with this?

Lukasa (approved):
Yup. Down with universal newlines!

PR #694, opened 2017-01-26
Catch errors from urlopen() during apidocs build

rodrigc:
http://twistedmatrix.com/trac/ticket/9013

### DIFF
--- a/src/twisted/python/_pydoctor.py
+++ b/src/twisted/python/_pydoctor.py
@@ -95,11 +95,12 @@ class TwistedSphinxInventory(SphinxInventory):
 
             # Check if URL exists.
             response = self._getURLAsHEAD(fullURL)
-            if response.code == 200:
-                return fullURL
-            else:
-                # Bad URL resolution.
-                return None
+            if response:
+                if response.code == 200:
+                    return fullURL
+                else:
+                    # Bad URL resolution.
+                    print("BAD URL resolution, code: ", response.code)
 
         return None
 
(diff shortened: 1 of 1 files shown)


exarkun:
We should really start keeping a local cache of the intersphinx material.  This is the 3rd or 4th "fix" for this problem I've seen.

rodrigc:
@exarkun @Lukasa can I proceed with this?  I know it is a bandage, but without it, Travis is marked as failing, and this blocks all merges to trunk.

src/twisted/python/_pydoctor.py
```diff
@@ -114,9 +115,13 @@ def _getURLAsHEAD(self, url):
...
 
         @return: The response for the HEAD method.
-        @rtype: urllib2 response
+        @rtype: urllib2 response or L{None}
         """
-        return urllib2.urlopen(HeadRequest(url))
+        try:
+            return urllib2.urlopen(HeadRequest(url))
+        except Exception as e:
+            print("Error opening {}: {}".format(url, e))
```
exarkun, line 123: https://github.com/twisted/twisted/pull/694/files#diff-aa8f1554609c110cd5bedee40ed3e19dR37 says there's a workflow for dealing with such things.  Can you do what that docstring says or fix the docstring if it says to do the wrong thing?

exarkun (approved):
Thanks.  Just one change required, I think, then go ahead and merge.

src/twisted/python/_pydoctor.py
rodrigc: The link you provided just takes me back to this diff.  Are you referring to this docstring: https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_pydoctor.py#L38

exarkun: Uh weird.  Yes, that docstring.

rodrigc: I'll open a separate ticket for that later today since the current ptoblem is specific to Activestate.
In order to report to them I need to create an account on their site to log into their forums.

PR #695, opened 2017-01-27
Get rid of RunnerOptions

wsanchez:
https://twistedmatrix.com/trac/ticket/9015


src/twisted/application/runner/_runner.py
```diff
@@ -31,19 +29,73 @@ class Runner(object):
...
+        @param reactorExited: Hook to call after the reactor exits.
+        @type reactorExited: callable that takes the keyword arguments
+            specified by C{reactorExitedArguments}
+
+        @param reactorExitedArguments: Keyword arguments to pass to
+            C{reactorExited} when it is called.
+        @type reactorExitedArguments: L{dict}
         """
-        self.options = options
+        self.reactor                = reactor
```
glyph, line 82: eeugggh it might be time for `attrs` :)

(Definitely not a blocking comment here, just observing that it would help here.)

glyph: More germane to the actual change here - do we want all these attributes to be public?  Some of them possibly should be (`reactor`?), but, clearly some are private.

src/twisted/application/runner/_runner.py
```diff
@@ -118,15 +164,14 @@ def startLogging(self):
     def startReactor(self):
         """
         Register C{self.whenRunning} with the reactor so that it is called once
-        the reactor is running and start the reactor.
-        If L{RunnerOptions.reactor} is specified in C{self.options}, use that
-        reactor; otherwise use the default reactor.
+        the reactor is running, then start the reactor.
```
glyph, line 170: This doesn't document an important side-effect; it _sets_ the reactor attribute.

src/twisted/application/runner/_runner.py
```diff
@@ -31,19 +29,73 @@ class Runner(object):
...
+        @type reactor: L{IReactorCore}
+
+        @param pidFile: The file to store the running process ID in.
+        @type pidFile: L{IPIDFile}
+
+        @param kill: Whether this runner should kill an existing running
+            instance of the application.
+        @type kill: L{bool}
+
+        @param defaultLogLevel: The default log level to start the logging
```
glyph, line 52: When does this default apply?  Is it just for legacy log events?

src/twisted/application/runner/_runner.py
```diff
@@ -31,19 +29,73 @@ class Runner(object):
...
+        defaultLogLevel=LogLevel.info,
+        logFile=stderr, fileLogObserverFactory=textFileLogObserver,
+        whenRunning=lambda **_: None, whenRunningArguments={},
+        reactorExited=lambda **_: None, reactorExitedArguments={},
+    ):
         """
-        @param options: Configuration options for this runner.
-        @type options: mapping of L{RunnerOptions} to values
+        @param reactor: The reactor to start and run the application in.
+        @type reactor: L{IReactorCore}
```
glyph, line 43: As implemented here, this `@type` is wrong; sometimes it's `None`, if it's given the default.

Personally I'd rather just eliminate the default and force the caller to pass a usable value.

src/twisted/application/runner/test/mockreactor.py
```diff
@@ -0,0 +1,40 @@
...
+"""
+Mock reactor.
+"""
+
+from twisted.internet import error
+from twisted.test.proto_helpers import MemoryReactor as BaseReactor
+
+
+
+class MockReactor(BaseReactor):
```
glyph, line 13: Could we maybe just put the `install` method on `MemoryReactor` itself, to avoid the proliferation of classes?  Or perhaps put this in `proto_helpers` itself?

If you don't feel that this is sufficiently mature to move to a central location, that's OK, but please file a refactoring ticket so that we don't lose track of this new reactor mock.

src/twisted/application/twist/_twist.py
```diff
@@ -84,39 +84,35 @@ def startService(reactor, service):
 
 
     @staticmethod
-    def runnerOptions(twistOptions):
+    def runnerArguments(twistOptions):
```
glyph, line 87: This seems like an odd placement and signature for this method.  We know that these are arguments for `Runner` - shouldn't it just be constructing one, rather than synthesizing keyword arguments which have no other use than to be passed to one?

It strikes me that perhaps this should be `Runner.fromTwistOptions` instead?

glyph (approved):
Generally this seems like a move in the right direction to me, so I think it should land.  Hooray for refactoring the internals of a private interface :).  But, some of these comments would be blocking for the larger goal of "make this public", so please address them in subsequent tickets.

Thanks for working on Twist-the-library, I'm sure this is going to improve our operational viability :).

src/twisted/application/runner/_runner.py
wsanchez: The "incorrect" part of that strikes me as a weak argument; there's a million places where we document a type where `None` is also an option, and I think that requiring that all such cases add an "or None" qualifier in text is super lame.

I also don't agree that forcing the caller to add some boilerplate to import and pass in the default reactor is an improvement here.  The point of this API to to "start the Twisted crap" up for you, not make you figure out how to do that yourself; if you don't have an opinion about which reactor is used, it's OK for this to choose one for you.

src/twisted/application/runner/_runner.py
wsanchez: It's the default for new events.  See `LogLevelFilterPredicate.defaultLogLevel`.

What we're setting up here is the filter predicate so that you have a way to start using the namespace and log level filtering mechanisms that are not currently set up by `twistd` so you have to set that up yourself, and I imagine few people do.  What's not provided yet is a way to override the log level for a namespace, but that'll come later.

However, `twist` right now has a `--log-level` argument which uses to set the global default.

src/twisted/application/runner/_runner.py
wsanchez: Yeah I thought of `attrs`, but one pile of changes at a time.

I'm not sure about the public/private argument, but will revisit.

src/twisted/application/runner/_runner.py
wsanchez: yes… yes.

wsanchez: To clarify, the side effect is that it installs the default reactor if none was specified and sets the attribute.

I'll add that to the docstring for now, but whether that's the right place to do this deserves some more consideration.

wsanchez: OK FINE GLYPH maybe the caller should have to pass in the reactor. FINE.

src/twisted/application/runner/test/mockreactor.py
wsanchez: A weird thing about reactors is the weird use of modules as global singletons.  A reactor doesn't really have an `install` method, so that's why I didn't put it on `MemoryReactor`.  The reactor *module* has an install method, until it is replaced by a reactor, then it doesn't any more.  Or something.  It's all pretty odd.

src/twisted/application/twist/_twist.py
wsanchez: This is an artifact of the old RunnerOptions hooey.  I agree it should go away, but I haven't started cleaning up `twist` yet.  After I get the underlying `runner` API fixed up, that's next.

src/twisted/application/runner/_runner.py
wsanchez: Addressed both in https://github.com/twisted/twisted/pull/708

src/twisted/application/runner/_runner.py
wsanchez: OK yeah, so… addressed in #708
:-)

src/twisted/application/runner/_runner.py
wsanchez: Removed this in #708

src/twisted/application/runner/test/mockreactor.py
wsanchez: Removed this in #708

src/twisted/application/twist/_twist.py
wsanchez: Removed this in #708

PR #696, opened 2017-01-28
Replace cgi.escape with twisted.python.compat.escape

rodrigc:
https://twistedmatrix.com/trac/ticket/9018


exarkun (approved):
Looks like a slam dunk.  Please merge.

PR #697, opened 2017-01-30
Do not start a new console when spawning a new process on Windows

rodrigc:
http://twistedmatrix.com/trac/ticket/5726

Moving patch submitted by @alecu to GitHub


rodrigc:
@alecu @opalmer please review this.

opalmer (approved):
Looks good. 

I initially thought `DETACHED_PROCESS` might be better suited here but that completely detaches the process from the original console which is probably not the behavior most people would expect. Eventually exposing the process creation flags is probably a better approach as that would allow someone more explicit control over the behavior around process creation. Of course that's a bigger change and separate from this PR...

src/twisted/test/test_process.py
```diff
@@ -2340,6 +2340,64 @@ def test_findShebang(self):
...
+class Win32CreateProcessFlagsTests(unittest.TestCase):
+    """
+    Check the flags passed to CreateProcess.
+    """
+
+    @defer.inlineCallbacks
+    def test_flags(self):
+        """
+        Verify that the flags passed to win32process.CreateProcess() prevent a
+        new console window from being created. See http://tm.tl/5726
```
glyph: I think that including the script inline (minimized as much as possible) would be better than linking to the closed ticket.

src/twisted/test/test_process.py
```diff
@@ -2340,6 +2340,64 @@ def test_findShebang(self):
...
+        """
+        from twisted.internet import _dumbwin32proc
+        flags = []
+        realCreateProcess = _dumbwin32proc.win32process.CreateProcess
+
+        def fakeCreateprocess(appName, commandLine, processAttributes,
+                              threadAttributes, bInheritHandles, creationFlags,
+                              newEnvironment, currentDirectory, startupinfo):
+            """
+            See the Windows API documentation for I{CreateProcess} for further details.
```
glyph, line 2381: twistedchecker shouldn't be checking a callback like this, and this documentation is all duplicative with the existing docs, so this docstring should probably be elided.

glyph (approved):
This looks great, modulo the minor tweaks I suggested.  Thanks for noticing and fixing!

rodrigc:
@alecu thanks for the submission.  Sorry it took 5 years to get this fix in.

PR #698, opened 2017-01-31
limit the globalLogBeginner buffer to dramatically less than 65k

pjenvey:
https://twistedmatrix.com/trac/ticket/9020

### DIFF
--- a/src/twisted/logger/_global.py
+++ b/src/twisted/logger/_global.py
@@ -23,6 +23,8 @@ from ._io import LoggingFile
 from ._file import FileLogObserver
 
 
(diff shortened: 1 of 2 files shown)


exarkun:
I didn't see that you'd created this until after I worked on <https://github.com/twisted/twisted/pull/699>

glyph:
#699 adds some test coverage and docs and so is generally a slightly better approach, so I'm closing this.  Thanks for your contribution nevertheless.

(In the future, please attach the 'review' keyword to the trac ticket so it doesn't fall through the cracks.  At some point we'll build a good Github process for dealing with this queue, but that's the most sure-fire way for the time being.)

PR #700, opened 2017-02-01
Remove deprecated RPC code

rodrigc:
https://twistedmatrix.com/trac/ticket/8464


glyph (approved):
L G T M

SO glad to be rid of this code.  Thanks for cleaning it up!

rodrigc:
Should I wait until 5/18/2017 to remove this, as per the comments for a similar review here:
https://github.com/twisted/twisted/pull/701#issuecomment-276685943

glyph:
Oh, crud, that's a good point. I had already forgotten that 16.x wasn't month-based and so 16.2 and 17.2 aren't a year apart.

So yes, this approval stands, but we should probably wait a few months to click the merge button.

rodrigc:
I'll set my alarm clock for 5/19/2017!

PR #702, opened 2017-02-02
DeferredSemaphore docstring documents constructor parameter

jisaacstone:
https://twistedmatrix.com/trac/ticket/9026#ticket

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1573,6 +1573,10 @@ class DeferredSemaphore(_ConcurrencyPrimitive):
     """
 
     def __init__(self, tokens):
+        """
+        @param tokens: initial value of L{tokens} and L{limit}
+        @type tokens: L{int}
+        """
         _ConcurrencyPrimitive.__init__(self)
         if tokens < 1:
             raise ValueError("DeferredSemaphore requires tokens >= 1")


src/twisted/internet/defer.py
```diff
@@ -1573,6 +1573,10 @@ class DeferredSemaphore(_ConcurrencyPrimitive):
     """
 
     def __init__(self, tokens):
+        """
+        @param tokens: initial value of C{tokens} and C{limit}
```
rodrigc: where is C{limit} set?

exarkun: Five lines below?  What do you mean?

markrwilliams: Maybe the confusion is that source code markup like `C{limit}` often refers to a callable's argument, but there's no `limit` argument.

markrwilliams: [Epytext can resolve `L{something}` in a method docstring to an instance variable `something`](http://epydoc.sourceforge.net/epytext.html) (See section 3.3).  While it's fine to use `C{...}` here, if you switch these to `L{tokens}` and `L{limit}`, pydoctor will generate links to those documented `@ivars`.  That would be pretty neat!

markrwilliams (commented):
This is a good documentation improvement!

I left a suggestion on how to associate the `tokens` parameter with the two instance variables it sets.  Please also add a [topfile](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles).  I think this can be a `misc` change.

Thanks for the contribution!!

src/twisted/internet/defer.py
jisaacstone: Did not know that, thanks!

jisaacstone:
Topfile in :)

markrwilliams:
@jisaacstone This looks good to me!  I've pushed this branch to twisted/twisted so the Buildbots will run it.  We'll see if that doesn't get codecov to change its mind about coverage.

PR #704, opened 2017-02-03
Fix reference to used IO class in flattenString

kapyshin:
[Ticket #9028](https://twistedmatrix.com/trac/ticket/9028)


exarkun:
Thanks!  Looks great.  Merging.

PR #705, opened 2017-02-05
Fix the URL.child docstring

exarkun:
https://twistedmatrix.com/trac/ticket/9021

### DIFF
--- a/src/twisted/python/_url.py
+++ b/src/twisted/python/_url.py
@@ -528,9 +528,10 @@ class URL(object):
 
         For example::
 
-            >>> (URL.fromText(u"http://localhost/a/b?x=y")
-                 .child(u"c", u"d").asText())
-            u'http://localhost/a/b/c?x=y'
+            >>> u = URL.fromText(u"http://localhost/a/b?x=y")
+            >>> v = u.child(u"c", u"d")
+            >>> print(v.asText())
+            u'http://localhost/a/b/c/d?x=y'
 
         @param segments: A path segment.
         @type segments: L{tuple} of L{unicode}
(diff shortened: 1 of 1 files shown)


src/twisted/python/_url.py
```diff
@@ -528,9 +528,10 @@ def child(self, *segments):
 
         For example::
 
-            >>> (URL.fromText(u"http://localhost/a/b?x=y")
-                 .child(u"c", u"d").asText())
-            u'http://localhost/a/b/c?x=y'
+            >>> u = URL.fromText(u"http://localhost/a/b?x=y")
+            >>> v = u.child(u"c", u"d")
+            >>> print(v.asText())
+            u'http://localhost/a/b/c/d?x=y'
```
Julian: Uh. 2 dumb questions.

1) epydoc doesn't have doctests then I take it?
2) Is the policy that we use Py2 reprs then? I assume the answer must be yes I suppose.

Regardless, lgtm.

Julian (approved):

src/twisted/python/_url.py
rodrigc: @Julian what do you mean by doctests?  There is a builder which generates documentation based on epydoc and reports errors: https://buildbot.twistedmatrix.com/builders/documentation/builds/2533

Julian: @rodrigc If this were sphinx (well, or just regular stdlib doctest but those are dark times) then this error would have (or could have) failed an executable doctest which would have shown that the output didn't match what was executed.

So I was just guessing (well, learning) that apparently then nothing actually executes the documentation examples?

exarkun: Generally speaking, Twisted doesn't use doctests.

exarkun: > Is the policy that we use Py2 reprs then? I assume the answer must be yes I suppose.

No idea.  This docstring is a bit weird.  Not many Twisted docstrings contain this kind of example.  Reading this, I notice that it's still wrong - print wouldn't show the repr().

PR #707, opened 2017-02-14
inlinecallbacks: pass deferred on to next callback

meilof:
fix problem where the callback added by inlineCallbacks would not return
its argument, so callbacks next in the chain would not work anymore


exarkun:
Thanks!  Please file a ticket at https://twistedmatrix.com/trac describing the problem, use-case, anything else you think might be relevant.  Then link to this PR and submit it for review.  Also, you'll need to add some unit tests for the change.

glyph:
Thanks for your interest in contributing to Twisted :).

As described on the trac ticket, however, this behavior is explicitly desired, so I am closing the pull request.

I hope you will make further contributions - looking forward to them :).

PR #708, opened 2017-02-14
More cleanup for twisted.application.runner

wsanchez:
http://twistedmatrix.com/trac/ticket/9038

 * Use `attrs` for `Runner` instance variables
 * Make all `Runner` class and instance variables private
 * Make `reactor` a required argument to `Runner`


glyph:
Using attrs is great but we should have a direct dependency in the `setup.py` for it.

src/twisted/application/runner/test/test_runner.py
```diff
@@ -68,11 +68,31 @@ def test_runInOrder(self):
         """
         L{Runner.run} calls the expected methods in order.
         """
-        runner = DummyRunner({})
+
+        calledMethods = []
+
+        self.patch(
```
glyph: Bad as inheritance is, global patching seems worse.  Why didn't you like `DummyRunner` as an approach to testing this?

src/twisted/application/runner/test/test_runner.py
```diff
@@ -68,11 +68,31 @@ def test_runInOrder(self):
         """
         L{Runner.run} calls the expected methods in order.
         """
-        runner = DummyRunner({})
+
+        calledMethods = []
```
glyph: Superficially, it seems to me that the @durin42 & @nathanielmanistaatgoogle orthodoxy here would be to say that there's a clear interface here separable from `Runner` which is a different layer: the side-effect-y methods called out here (`killIfRequested`, `startLogging` etc).  However, this test appears to be the _only_ place in the entire codebase where we separate those layers, so perhaps it would be unnecessary overhead to actually separate them properly?

@ - mentions in place to see if the thought leaders in question have a thought to lead in this case.

src/twisted/application/runner/test/test_runner.py
```diff
@@ -86,12 +106,23 @@ def test_runUsesPIDFile(self):
...
+
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            self.exited = True
+
+        self.patch(NonePIDFile, "__enter__", __enter__)
+        self.patch(NonePIDFile, "__exit__", __exit__)
 
-        self.assertFalse(pidFile.entered)
-        self.assertFalse(pidFile.exited)
+        pidFile = NonePIDFile()
```
glyph: Rather than abusing `NonePIDFile` here, at this point couldn't you just write a little in-line `@contextmanager` that set a flag, and pass that, skipping the `.patch` calls?

src/twisted/application/twist/test/test_twist.py
```diff
@@ -157,18 +138,27 @@ def test_run(self):
             Runner, "run", lambda self: None
         )
 
-        twistOptions = Twist.options(["twist", "web"])
-        runnerArguments = Twist.runnerArguments(twistOptions)
-        Twist.run(runnerArguments)
+        twistOptions = Twist.options([
```
glyph, line 141: I do have to say that seeing these little classmethod invocations gives me a good feeling about where this API is going; it seems quite expressive.

glyph (changes requested):
Thanks @wsanchez !

This does seem mostly like a good change, but every call to `.patch` feels to me like technical debt we should be accounting for in the future.  Some of the uses here seem trivial to eliminate.  I wouldn't say that we _need_ to eliminate all of them, but it would be good at least to figure out if there's a future direction that would reduce this or it's just a one-off corner case where we're wanting to test this.  (I suspect it won't be, since the long-term idea here is in fact to make it into a library.)

Given that, can you try to address these comments and then resubmit? Thanks!

src/twisted/application/runner/test/test_runner.py
wsanchez: Basically: @moshez made fun of me for subclassing in a prior review.

wsanchez: I'm having a hard time caring either way, but I don't want to switch this back because of the whim-du-jour of the reviewer, even if it's @glyph.

What's the problem with patching here that you want to avoid?  There's a lot of patching here that I can't simply replace with subclasses, so I don't get why this case is special, or are you asking me to rewrite everything that calls `self.patch`?

glyph: Ideally neither subclassing or `.patch` would be required, but subclassing is (in this case, anyway) I'd say tolerable.

`.patch` is the worst possible way to change things to make them testable, because they're process-global.  It should always be at the bottom of the mocking hierarchy.  It is a good "option of last resort" for garbage legacy code that is really not indirectable in any other way, and hence isn't something we should remove from _trial_, but we should avoid it in our own tests as much as possible.  Hopefully @moshez will concur on this general hierarchy: composition > inheritance > patching.

Moving on to a practical problems involved with this case: let's say that `trial` wants to adopt `Runner` as a way to run itself.  If there's any problem with the patching, the global instance that Trial is using for its _own_ task-running will suddenly have these methods replaced.  Now, in this *exact* case probably nothing happens, but in the rare case that some horrible bug causes a patch cleanup to fail, your process can no longer shut down, or finish starting up, because this test's patches are still in place.

src/twisted/application/runner/test/test_runner.py
durin42: My general feeling is normally that it's not worth doing this kind of detangling, especially if it's a single place for testing purposes (I'm typically opposed to the idea of "interface-for-testing" as a philosophy).

That is, I suspect (without having enough understanding of what I'm seeing in this PR) that it's not worth doing the cleanup. I'd love to look at the notion in more detail at PyCon though if you're interested - it could be an informative exploration.

src/twisted/application/runner/test/test_runner.py
durin42: (barging in here too since @glyph seems to have invited trouble from me) As a general guideline, I think I'd say subclassing is better then global patching every time. Every experience I've had with global patching has been fraught when the time comes to refactor in the future, moreso than with subclassing.

wsanchez: OK, resurrecting `DummyRunner`.

wsanchez:
OK I think I've addressed @glyph review, resubmitting.

glyph (approved):

glyph:
Looks like the review was addressed, as advertised.

PR #709, opened 2017-02-15
[#9040] use .tox directory

glyph:
https://twistedmatrix.com/trac/ticket/9040


rodrigc (changes requested):
I'm OK with this change, but some tests are failing on Windows such as: `twisted.test.test_plugin.DeveloperSetupTests.test_developmentPluginAvailability`

hawkowl (changes requested):
The IOCP builder doesn't like it, and some Travis builds are failing. plz fix

glyph:
Redoing this from scratch will be easier than dealing with the conflicts, so I'll do that if this ever starts bothering me again :)

PR #710, opened 2017-02-17
Add a tox environment for downstream testing

### LINKED ISSUE #9042: Add treq to Twisted CI
|-|-|
|Trac ID|trac#9042|
|Type|enhancement|
|Created|2017-02-16 13:19:41Z|
|Branch|https://github.com/twisted/twisted/tree/9042-ci-for-treq|

See [mailing list discussion](http://twistedmatrix.com/pipermail/twisted-python/2017-February/031174.html).

```
trac-id__9042 9042
type__enhancement enhancement
reporter__exarkun exarkun
priority__normal normal
milestone__None None
branch__9042_ci_for_treq 9042-ci-for-treq
branch_author__ 
status__new new
resolution__None None
component__release_management release management
keywords__None None
time__1487251181816127 1487251181816127
changetime__1492323056342274 1492323056342274
version__None None
owner__exarkun exarkun

```
</details>

exarkun:
https://twistedmatrix.com/trac/ticket/9042


rodrigc:
I would say instead of calling it `downstream`, just call it `treq`.  Each time you add a new downstream build, just call it what you are testing.

Be warned that Travis can get overloaded, so adding more builds there should be done with
caution.
Adding builds like this to the buildbot cluster and making them "required" for the overall build to succeed may be more manageable in terms of overall resources.

.travis.yml
```diff
@@ -48,6 +48,8 @@ matrix:
       env: TOXENV=txchecker-travis-required
     - python: 2.7
       env: TOXENV=txchecker-travis-all
+    - python: 2.7
```
hawkowl: I think we should not make Treq able to fail the build; it relies on external websites (httpbin?) for its integration testing. Of course, that hasn't been an issue on treq development, so... use your discretion.

hawkowl (approved):
I think this shouldn't gate the build right now, as it does fail. The external state concern of mine is secondary. I think, make it non-dependent on passing for now (with the regression fix moving it out of allowed-failures), and merge.

.travis.yml
glyph: Maybe we should get rid of that particular misfeature of treq's CI?

hawkowl: @glyph Maybe. I don't know how we'll do that in the short term, though :)

glyph:
Closing PR due to staleness and merge conflicts.

PR #711, opened 2017-02-23
[WIP] Tweak transfer decoder setup to allow easy extensiblity.

rmax:
This allows to customize transfer decoders.

One use case is to handle HTTP connection data loss gracefully without
dropping the response.


rmax:
This allows to fix https://github.com/scrapy/scrapy/issues/2586

rmax:
Closing as there is not enough need from our side.

PR #713, opened 2017-02-23

rodrigc:


acabhishek942:
look good to me. Please merge

src/twisted/python/_release.py
```diff
@@ -39,14 +40,66 @@
...
+    if not isinstance(output, str):
+        output = output.decode("utf-8")
+    return output
+
+
+
+class FilePathStrContent(FilePath):
+    """
+    Wrapper around L{FilePath} to make dealing with
+    L{bytes} and L{str}
```
exarkun, line 70: `str` isn't really a distinct type anymore.  Do you mean `unicode`?  In general, this class seems like a bad idea.  It encourages lack of clarity about the string type being operated on.

Put another way, this is really deferring the actual work of porting.

src/twisted/python/_release.py
```diff
@@ -39,14 +40,66 @@
...
+    @return: command output
+    @rtype: L{str}
     """
     kwargs['stderr'] = STDOUT
-    return check_output(args, **kwargs)
+    output = check_output(args, **kwargs)
+
+    # The check_output() function always returns bytes
+    # on Python 2 and 3, but we want this function to always return
+    # str.
```
exarkun, line 60: What?  Why?  That sounds awful.  `str` is the worst string type.

src/twisted/python/test/test_release.py
```diff
@@ -1211,8 +1221,10 @@ def createFakeSphinxProject(self):
         files.  This includes a single source file ('index.rst') and the
         smallest 'conf.py' file possible in order to find that source file.
         """
-        self.sourceDir.child("conf.py").setContent(self.confContent)
-        self.sourceDir.child("index.rst").setContent(self.indexContent)
+        confPy = FilePathStrContent(self.sourceDir.child("conf.py"))
+        indexRst = FilePathStrContent(self.sourceDir.child("index.rst"))
+        confPy.setContent(self.confContent)
+        indexRst.setContent(self.indexContent)
```
exarkun, line 1227: The coverage report says this whole function is un-executed.  Is that a codecov bug or does the CI system not exercise this code?

src/twisted/python/test/test_release.py
```diff
@@ -1400,7 +1412,7 @@ def test_remove(self):
         """
         reposDir = self.makeRepository(self.tmpDir)
         testFile = reposDir.child('some-file')
-        testFile.setContent("something")
```
exarkun, line 1403: I don't think a module can be considered ported as long as it uses `str` with `FilePath.child`.  Some Python APIs legitimately take `str` as an argument - but not this one.

exarkun (changes requested):

src/twisted/python/_release.py
rodrigc: The code in this module is expecting the output of this function to be
`str` or _native string_.  I found it easier to make this change here rather than change
all the places that use output from this

src/twisted/python/test/test_release.py
rodrigc: That's not clear from the docstring for `FilePath.child`

src/twisted/python/test/test_release.py
rodrigc: This test requires sphinx to be installed in order to run, otherwise the test is skipped.  It looks like the builders which upload coverage does not install sphinx.

src/twisted/python/_release.py
rodrigc: I disagree.  The intent of this helper class is to put the bytes logic in one place
just for this module when dealing with FilePath.  If you don't like that, then
the alternative is to go and change all the instances where setContent() / getContent()
is used in this module and do the conversions in those places.

I found it easier to use a helper class.

glyph: I think it's worth keeping in mind that this is
1. private API, not public
2. used only as a _tool_, not even indirectly exposed to applications.

I agree that this is not the best approach, in that it adds additional wrapper code bloat rather than dealing with the type issues at each call site, but it seems like it would be expedient to land this and address the review feedback as small follow-ups, as there's no compatibility burden associated with the sub-optimal solution here.

src/twisted/python/test/test_release.py
glyph: I think what @exarkun is saying is not that you can't call `FilePath.child` with a type called `str` on either Python version (indeed you can, it's quite intentional ambidextrous when it comes to text and bytes), but rather that `str` is an ambiguous 2/3 type, and any code passing _what it considers to be a `str`_ (i.e.: it doesn't know whether it's really talking about text or bytes, so it's using an ambiguous type) has not really been through the process of porting.

Generally, I agree with this sentiment, but I'm willing to be a bit more lax with `_release.py` because of the lack of compatibility contract; we can address each call site in a small clean-up PR after the module is ported.

exarkun: Absent further context, I would likely be convinced by the private interface argument here.  However, I have observed a number of regressions introduced by Python 3 porting work in recent Twisted releases (a much higher rate than I've ever observed from other kinds of development) and to me this signals a need to be more careful in further porting efforts.  It's _less_ work to address this properly now than to land something half-complete and then, should the need arise, deal with bug fixes and potentially issuing bug-fix releases.

It also seems to me that at this point the way to port `str`-using code should be quite well understood and there should be no real blockers to porting this code according to those methods.

glyph: The only type of development beating it is probably backpressure and connection control in HTTP ;-) (heyyy @Lukasa ).

I'm inclined to agree, as I can't disagree with your impressions here; there is a certain mind-numbing tedium to both writing and reviewing py3 porting code, and that does result in some mistakes.  Perhaps addressing this feedback will be a good opportunity for someone (not necessarily @rodrigc, any committer can address it) to get better at identifying the appropriate text or bytes type to use in various cases.

rodrigc:
Replaced by https://github.com/twisted/twisted/pull/782

PR #714, opened 2017-02-23
[#9049] Stop hashing random bytes for no particular reason

alex:
crypto-gibberish is bad.

https://twistedmatrix.com/trac/ticket/9049


glyph:
Thanks alex! Looks good to me. I think that this was originally a timestamp or something? In any case yes cryptogibberish is bad.

sha1 on your mind today for some reason? Just stopping by? 😏

alex:
It was SHA1 gibberish going back at least as far as 2005, I stopped scanning the history at that point.

alex:
The codecov results appear to be nonsense. Can someone do something so the buildbot runs and this can be merged?

glyph:
Thing done.

alex:
👍  thanks! we're all green here.

glyph:
Thanks for the fix, @alex.

alex:
Happy to!

PR #721, opened 2017-02-27
Remove explicit use of _PY3 in failure.py

rodrigc:
http://twistedmatrix.com/trac/ticket/9057


Lukasa (approved):
This seems reasonable to me. Please bring it back up to date with master and then, if the tests are green, go ahead and merge.

PR #722, opened 2017-02-27
Fix PB call with keyword arguments between Py2 client and Py3 server

rodrigc:
https://twistedmatrix.com/trac/ticket/9047


rodrigc:
@tardyp can you look at this PR?  I couldn't figure out how to write a unit test which
uses keyword arguments with bytes keys , and have the test work on Py2 and Py3.

src/twisted/spread/flavors.py
```diff
@@ -109,6 +109,11 @@ def remoteMessageReceived(self, broker, message, args, kw):
         """
         args = broker.unserialize(args)
         kw = broker.unserialize(kw)
+        # Need this to interoperate with Python 2 clients
+        # which may try to send use keywords where keys are of type
+        # bytes.
+        if [key for key in kw.keys() if isinstance(key, bytes)]:
+            kw = dict((k.decode('utf8'), v) for k, v in kw.items())
```
tardyp, line 116: I was hoping we can find out a better implementation.

This works, but is imho a bit ugly and inefficient.

I don't know enough spread to know where exactly is the better fix (maybe in unserialize or unjelly)

rodrigc:
@glyph can you review this?  This is the last major outstanding problem in Twisted that was discovered during porting of buildbot to Python 3.  It was reported by @tardyp

glyph:
@rodrigc I'll put it on my list, but right now my schedule is pretty full.

src/twisted/spread/flavors.py
glyph: Let's go with "correct and inefficient" for now, rather than "efficient and wrong" :).

glyph:
Looks like the build failures were openssl-1.1.0 related, so hopefully the next time I roll around to looking at this it'll be a green build :)

src/twisted/spread/flavors.py
```diff
@@ -109,6 +109,11 @@ def remoteMessageReceived(self, broker, message, args, kw):
         """
         args = broker.unserialize(args)
         kw = broker.unserialize(kw)
+        # Need this to interoperate with Python 2 clients
+        # which may try to send use keywords where keys are of type
+        # bytes.
+        if [key for key in kw.keys() if isinstance(key, bytes)]:
```
glyph, line 115: To be specific about @tardyp's criticism:

1. This examines _every_ key, where it only really needs to examine one.  This test could be much more efficient as `any(key for key in kw.keys() if isinstance(key, bytes))`.
2. This might also be more efficient as an in-place modification of `kw`; additionally, it might be more correct. If we have determined that a certain object is really supposed to be a kwargs map, then we know it ought to have native str types in it, and those should show up in other places if we can apply that heuristic more globally.  It would also avoid copying over the entire dictionary.
3. It does the conversion in the py2 case even though that's arguably _more_ wrong; although `**{u'...': ...}` _works_ in py2, it's only kinda by accident (and the runtime is converting back into `str` under the covers anyway).

It might also be worthwhile to do some up-front negotiation to discover if this is required; two 3-only clients could easily say to each other "hey, we're both py3, no need for that gross bytes/args check, everything should check out" earlier on in the connection.

glyph (approved):
I think that this is a step in the direction, but if it's going to land as-is, it really needs a bit more follow-up work.

@rodrigc - please land, but file a follow-up (and link to it from the comment below) to record the performance issues, as well as the fact that testing could be more thorough here.  In order to ensure the desired goal of the change (2/3 interop), there really ought to be a binary dump of a protocol message generated by 2 and 3, and ensure that processing each has the desired impact.  We should also really have some edge-case tests for what happens with non-ascii keyword argument names.

But, the performance regression here is hypothetical, I haven't looked at speed.twistedmatrix.com, and if it's serious we can always optimize later.

src/twisted/spread/flavors.py
rodrigc: I created a new ticket: https://twistedmatrix.com/trac/ticket/9105

glyph: Thanks!

PR #723, opened 2017-02-28
Allow for a missing status-line phrase.

exarkun:
https://twistedmatrix.com/trac/ticket/7673


Lukasa (approved):
Web servers are the *absolute* worst. I am greatly looking forward to the day where HTTP/1.1 is no more.

This patch looks great @exarkun: assuming the tests pass, you can go ahead and merge it.

PR #725, opened 2017-02-28
[#9064] Added Support for HTTP multiple same content-length headers

anio:
Added Support for HTTP multiple same content-length headers.

[RFC7230#section-3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2):
>    If a message is received that has multiple Content-Length header
>    fields with field-values consisting of the same decimal value, or a
>    single Content-Length header field with a field value containing a
>    list of identical decimal values (e.g., "Content-Length: 42, 42"),
>    indicating that duplicate Content-Length header fields have been
>    generated or combined by an upstream message processor, then the
>    recipient MUST either reject the message as invalid or replace the
>    duplicated field-values with a single valid Content-Length field
>    containing that decimal value prior to determining the message body
>    length or forwarding the message.

https://twistedmatrix.com/trac/ticket/9064

### DIFF
--- a/src/twisted/web/_newclient.py
+++ b/src/twisted/web/_newclient.py
@@ -507,7 +507,7 @@ class HTTPClientParser(HTTPParser):
                     b'content-length')
                 if contentLengthHeaders is None:
                     contentLength = None
-                elif len(contentLengthHeaders) == 1:
+                elif len(set(contentLengthHeaders)) == 1:
                     contentLength = int(contentLengthHeaders[0])
                     self.response.length = contentLength
                 else:


glyph:
Hi @anio ! Thanks very much for your contribution. In the future, please file a ticket on Trac, as described in [the contributing guide](https://github.com/twisted/twisted/blob/b7f704386d4b2f383a947f7afb7a96c84eccb07c/CONTRIBUTING#L22-L30), and link it to a PR on github.  (While this system will eventually be simplified, for right now it helps us keep track of incoming PRs, given that much of our development tooling predates Github).

glyph:
I've created a ticket for you and submitted it for review so hopefully a reviewer will come along and look at it.  Thanks!

Lukasa (changes requested):
Thanks for this! The patch looks reasonable to me, but it's really important that we add a test for this functionality. Do you mind doing that?

anio:
OK @glyph, thank you very much!

> @Lukasa: it's really important that we add a test for this functionality. Do you mind doing that?

Sorry @Lukasa,  i don't think!

hawkowl:
Superseded by https://github.com/twisted/twisted/pull/735

PR #726, opened 2017-03-01
Revert "8864 safe str unicode"

markrwilliams:
Reverts twisted/twisted#566

See:

http://twistedmatrix.com/trac/ticket/8864#comment:9


glyph:
Since the topfile shouldn't be necessary for reverts, I'm going to use administrator powers to merge this.

PR #729, opened 2017-03-05
Replace xrange with twisted.python.compat.range

rodrigc:
https://twistedmatrix.com/trac/ticket/9069


glyph (approved):
The remaining coverage issues that I can see are spurious reporting of context-adjacent code not having branch coverage rather than actually any uncovered lines in the diff itself; the few unrelated changes appear to be all fairly straightforward cleanups.  As such I'd say that this can land assuming the rest of CI is happy with it, and you concur with my assessment of the coverage problems.

PR #732, opened 2017-03-05
Check more files with pyflakes3

rodrigc:
http://twistedmatrix.com/trac/ticket/9072


hawkowl:
@rodrigc could we exclude compat.py and test_compat.py, as those will always have errors?

PR #733, opened 2017-03-05
Use twistedchecker 0.7.1 and run under Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9074

Using twistedchecker and pylint under Python 3 is becoming increasingly important,
now that Python 3 only constructs are starting to go into Twisted, such as *await*.


glyph (approved):
These changes look good to me; when CI has finished catching up, please land.

PR #84, opened 2016-05-29
Deprecate UsernameHashedPassword

mesozoic:
See https://twistedmatrix.com/trac/ticket/8368

PR #734, opened 2017-03-05
Deprecate UsernameHashedPassword

exarkun:
https://twistedmatrix.com/trac/ticket/8368

Update of https://github.com/twisted/twisted/pull/84


exarkun:
The brand new `getDeprecatedModuleAttribute` needs tests.

twm:
I have added tests of the `getDeprecatedModuleAttribute` method. Putting this back into review.

docs/core/development/policy/compatibility-policy.rst
```diff
@@ -532,10 +541,24 @@ Making calls to the deprecated code without raising these warnings can be done u
...
+        """
+        Tests for L{UsernameHashedPassword}.
+        """
+        def test_initialisation(self):
+            """
+            The initialisation of L{UsernameHashedPassword} will set C{username}
+            and C{hashed} on it.
+            """
+            UsernameHashedPassword = self.getDeprecatedModuleAttribute(
+                'twisted.cred.credentials', 'UsernameHashedPassword', Version('Twisted', 16, 3, 0))
```
glyph: Might want to update this version number to be a little more recent

src/twisted/cred/test/test_cred.py
```diff
@@ -23,6 +24,11 @@
 
 
 
+# The Twisted version in which UsernameHashedPassword is first deprecated.
+_uhpVersion = Version("Twisted", "NEXT", 0, 0)
```
glyph, line 28: Nice to factor this out!

src/twisted/cred/test/test_simpleauth.py
```diff
@@ -52,6 +53,10 @@ def test_interface(self):
 
 
 
+_uhpVersion = Version("Twisted", "NEXT", 0, 0)
```
glyph: Why is this duplicated here rather than imported?

src/twisted/newsfragments/8368.removal
```diff
@@ -0,0 +1 @@
+twisted.cred.credentials.UsernameHashedPassword is now deprecated.
```
glyph: Can we add a little more color on what was wrong with it / why it was deprecated? I don't want folks copy/pasting this into their own codebases, since it's not just moved or renamed, it's kind of backwards

glyph (approved):
Love it. So glad to see this finally get done.  Thanks to the whole team who made it happeN :)

twm:
@glyph Thanks for the review! I believe that I have addressed your feedback, to the extent that I can, so I will merge once the build passes.

PR #735, opened 2017-03-05
[#9064] Agent: Support duplicate Content-Length headers

emattiza:
Note: Current code-base is compliant. ValueError is raised at first sign of duplicates. However, it would be nice to gracefully reconcile poorly created responses instead of raising an error; but for specific cases. These cases are where the multiple headers are duplicate, yet have equivalent, singular field-values. Also, when a single header and it's field-value is a list of lengths, when list contents are equal, parser should reconcile for appropriate contentLength and enter body processing.

This is some unit-testing to help in creating this change on [https://github.com/twisted/twisted/pull/725](PR#725) and [https://twistedmatrix.com/trac/ticket/9064](Trac#9064). It looks like there would need to be a pretty large change to the header parser, and not just a one-line fix to allHeadersReceived unfortunately. On debugging the unit-tests, exceptions are raised for attempting to integer convert the zeroth list entry of the parsed content-length header field-value for the field-value list case. The parser does not in fact list convert the field-value for use in the set operation proposed originally. This should definitely not be pulled in yet, but the test framework is there now to test the requested support functionality.
Signed-off-by: emattiza <emattiza@gmail.com>


rodrigc:
The tests aren't passing, see https://travis-ci.org/twisted/twisted/builds/207944260

You should try running the tests before pushing to github, as per https://github.com/twisted/twisted/blob/trunk/README.rst#unit-tests

emattiza:
@rodrigc Sorry! I was aware they were failing, should have awaited the push on my end until i had the fix in place. No newsfile in the folder yet, ran tests w/ debug on my end for the HTTPClientParser suite to work through this.

hawkowl:
Running the tests on the builder fleet...

src/twisted/web/_newclient.py
```diff
@@ -279,7 +279,14 @@ def lineReceived(self, line):
                     header = b''.join(self._partialHeader)
                     name, value = header.split(b':', 1)
                     value = value.strip()
-                    self.headerReceived(name, value)
+                    # Handle comma-separated list as field-value list
```
hawkowl: I think this violates https://tools.ietf.org/html/rfc7230#section-3.2.6 (it doesn't check for quoting, for example). I also don't think we should be doing this here, since it could break things that do the , parsing later, for example.

src/twisted/web/_newclient.py
```diff
@@ -332,7 +339,10 @@ def headerReceived(self, name, value):
             headers = self.connHeaders
         else:
             headers = self.headers
-        headers.addRawHeader(name, value)
+        if type(value) is list:
```
hawkowl: Again, here. I don't think we should be doing this here.

src/twisted/web/_newclient.py
```diff
@@ -507,7 +517,7 @@ def allHeadersReceived(self):
                     b'content-length')
```
hawkowl: Instead, I think we should be parsing the content-length header here.

hawkowl (dismissed):
Hi @emattiza , thanks for this. The tests are good, but I think the implementation should be moved to target content-length more specifically, lest it break implementations that aren't prepared for the generic , splitting (the current implementation also would split a quoted header, which is against the spec). Move it to just target content-header, and then I think we're good to land. Thanks again.

src/twisted/web/_newclient.py
emattiza: implemented parser changes inside of headerReceived.

anio:
Any change after a year? Do you need any help?

anio:
It seems after 3 years, twisted is not supporting duplicate content-length (Part of [HTTP RFC](https://tools.ietf.org/html/rfc7230#section-3.3.2)) yet.

Congratulation dudes!

adiroiban:
Hi Ali. Thanks for your feedback.

Twisted is an open source project, that is as good as the contribution from the volunteers.

If you have the time and knowledge and would like to see this issue fix please send a pull request.

We would be happy to review and merge it.

------------

This is a PR that was started by Evan a long time ago and a new review was requested just now.

You can also send a review for the current PR.

twm:
I started to review this, but noticed we're allowing some invalid forms like:

    Content-Length: -1

And:

    Content-Length: +100

So I've pushed up changes that bring this into alignment with the [server implementation](https://github.com/twisted/twisted/commit/592217e951363d60e9cd99c5bbfd23d4615043ac). While those changes addressed a request smuggling issue, I don't think that this is such a big deal on the client side.

src/twisted/web/_newclient.py
```diff
@@ -201,6 +202,10 @@ class HTTPParser(LineReceiver):
     @ivar headers: All of the non-connection control message headers yet
         received.
 
+    @ivar connHeaders: All of the connection control message headers yet
+        received. See L{CONNECTION_CONTROL_HEADERS} and
+        L{isConnectionControlHeader}.
+
```
glyph, line 208: Love to see documentation for existing stuff show up

src/twisted/web/_newclient.py
```diff
@@ -644,6 +640,77 @@ def _ensureValidURI(uri):
     raise ValueError(f"Invalid URI {uri!r}")
 
 
+def _decint(data: bytes) -> int:
```
glyph, line 643: this looks… like a good idea… but uh, why is it in the *client* implementation? doesn't the server need to do this too?

glyph (approved):
This looks like a straightforward bug fix with great embedded maintainer documentation, but I am a little confused about why this logic is client-specific.  If you could add a note in the docs as to why, and whether this could be merged with something server-side, that would be great.

src/twisted/web/_newclient.py
twm: I previously fixed this on the server side (8ebfa8f6577) but the implementations could certainly be aligned.

I see that `twisted.web.http.HTTPClient` makes the same error though (this has me confused as I thought we deleted the old client with `getPage` and friends, but this one doesn't seem to even have been deprecated).

I filed #12158 and #12159 for this.

twm:
Thank you for the review @glyph! I'll go ahead and merge this as-is and address the tickets I filed as follow-up work.

PR #740, opened 2017-03-11
9032 HostnameEndpoint Memory Reactor

markrwilliams:
http://twistedmatrix.com/trac/ticket/9032

This aims to fix a release blocker.


markrwilliams:
There's an unrelated OpenSSL issue that's causing the Windows and macOS builders to fail.

**EDIT** #743 is an attempt at fixing those.

src/twisted/internet/endpoints.py
```diff
@@ -716,6 +809,30 @@ def __init__(self, reactor, host, port, timeout=30, bindAddress=None,
...
+        @param reactor: The reactor to check.
+
+        @return: A L{IHostnameResolver} provider.
+        """
+        if not IReactorPluggableNameResolver.providedBy(reactor):
+            warningString = deprecate.getDeprecationWarningString(
+                reactor.__class__,
+                Version("Twisted", "NEXT", 0, 0),
+                format=("Passing HostnameEndpoint %(fqpn)s"
+                        " was deprecated in %(version)s"),
```
tomprince: It isn't clear to me that this should be calling out passing a specific type as being deprecated. I think it would be better to simply say passing a reactor that doesn't implement `IReactorPluggableNameResolver` is deprecated. Particularly since the fix might be to make the reactor implement the new interface, rather than changing the reactor being passed.

src/twisted/web/test/test_agent.py
```diff
@@ -1425,7 +1425,7 @@ def test_deprecatedDuckPolicy(self):
         WebClientContextFactory} to do it.
         """
         def warnMe():
-            client.Agent(MemoryReactorClock(),
+            client.Agent(deterministicResolvingReactor(MemoryReactorClock()),
```
tomprince, line 1428: Can you file a ticket for either making this public, or extending MemoryReactorClock with this functionality?

src/twisted/topfiles/9032.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet.endpoints.HostnameEndpoint and twisted.web.client.Agent work again with reactors that do not provide IReactorPluggableNameResolver.  This undoes the changes that broke downstream users such as treq.testing.  Note that passing reactors that do not provide IReactorPluggableNameResolver to either is deprecated.
```
tomprince: I think this would be better split into two news files. One for the bugfix, and another one for the deprecation. Also, I don't think it makes sense to call out that this is undoing a previous change: the new behavior is actually different than the old and the point isn't to undo it, but just provide proper deprecation warnings for the change.

src/twisted/internet/endpoints.py
```diff
@@ -643,6 +649,89 @@ def _resolvedHostConnect(self, resolvedHost, protocolFactory):
...
 
+@implementer(IHostnameResolver)
+class _SimpleHostnameResolver(object):
+    """
+    An L{IHostnameResolver} provider that invokes a provided callable
+    to resolve hostnames.
+
+    @ivar _nameResolution: the callable L{resolveHostName} invokes to
+        resolve hostnames.
+    @type _nameResolution: A L{callable} with the same signature as
```
tomprince: There are some missing words here.

src/twisted/internet/endpoints.py
```diff
@@ -643,6 +649,89 @@ def _resolvedHostConnect(self, resolvedHost, protocolFactory):
...
+
+        @param transportSemantics: Ignored in this implementation.
+
+        @return: The resolution in progress.
+        @rtype: L{IResolutionReceiver}
+        """
+        resolutionReceiver.resolutionBegan(HostResolution(hostName))
+        d = self._nameResolution(hostName, portNumber)
+        d.addCallback(self._deliver, resolutionReceiver)
+        d.addErrback(self._ebLog, hostName)
```
tomprince: I don't think there is any value in having these callbacks out-of-line. There is a bunch of code like this in twisted, but I *think* that newer code tends to avoid it.

src/twisted/internet/endpoints.py
```diff
@@ -888,6 +1016,22 @@ def cancelRemainingPending(result):
         return d
 
 
+    def _nameResolution(self, host, port):
+        """
+        Resolve the hostname string into a tuple containing the host
+        address.
```
tomprince: It would probably be good to make it clear at least in the docstring and perhaps in the method name that this method is just used as a fallback when the reactor provided to the endpoint doesn't implement the new interface.

src/twisted/internet/endpoints.py
```diff
@@ -888,6 +1016,22 @@ def cancelRemainingPending(result):
...
+
+        @param host: A unicode hostname to resolve.
+
+        @param port: The port to include in the resolution.
+
+        @return: A L{Deferred} that fires with L{_getaddrinfo}'s
+            return value.
+        """
+        return self._deferToThread(self._getaddrinfo, host, port, 0,
+                socket.SOCK_STREAM)
```
tomprince, line 1021: I'm not sure that it makes sense to allow overriding `deferToThread` and `getaddrinfo` individually, instead of just overriding this function.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1687,6 +1688,381 @@ class WithResolver(proxyForInterface(
...
+        @param reactor: The L{MemoryReactor} under test.
+
+        @return: List of calls to L{IReactorTCP.connectTCP}
+        """
+        return reactor.tcpClients
+
+
+    def connectArgs(self):
+        """
+
```
tomprince, line 1906: There is an extra newline here.

tomprince (changes requested):
This looks like a good improvement. There are a couple of small changes, and I didn't have time to look at the tests.

src/twisted/internet/endpoints.py
markrwilliams: It does - I took this approach because I had pretensions of testing each of these methods.  But coverage is good, so I might as well switch back to closures.

markrwilliams: Also the docstrings looked nicer on instance methods than they did on closures.

src/twisted/internet/endpoints.py
markrwilliams: The log message should include the interface required, but I'd like it to also include the FQPN of the reactor to make debugging tests easier.

tomprince: Including the FQPN of the reactor seems fine. But I'd go with something like:

> Passing a reactor that doesn't implement IReactorPluggableNameResolver to HostnameEndpoint was deprecated <...>. (`%(fqpn)s` was passed, which doesn't currently implement <...>

that makes it clear that a possible fix is making the reactor implement the given interface.

src/twisted/internet/endpoints.py
tomprince: If they are closures, they might not need docstrings (certainly they could do with more abbreviated docstrings in that case).

src/twisted/web/test/test_agent.py
markrwilliams: Done!  https://twistedmatrix.com/trac/ticket/9103

src/twisted/internet/endpoints.py
markrwilliams: `deferToThread` has to be replaceable to make this function testable.  The original implementation also made `getaddrinfo` replaceable, apparently with the intent of verifying that `deferToThread` invokes it.

I was inclined to leave these two staticmethods in for two reasons:

1. My tests looked more like the existing ones, apparently verifying the same behaviors.
2.  A vague concern that some tests, inside or outside Twisted itself, relied on it.  

A cursory GitHub search indicated that 2. wasn't the case.  1. remains true and useful, I think.

src/twisted/internet/test/test_endpoints.py
markrwilliams: `twistedchecker` complains if there isn't a newline here, and it complains if the return type isn't documented.

src/twisted/internet/endpoints.py
markrwilliams: I guess it already does mention the interface required of the reactor.

src/twisted/internet/endpoints.py
markrwilliams: They don't - I found them useful while writing the patch, so I thought I'd leave them in.

src/twisted/internet/test/test_endpoints.py
tomprince: That seems like a bug in twistedchecker. Or an indication there should be some prose here.

markrwilliams: I think it's a bug in twistedchecker.  All of the other test classes that define `connectArgs` have the same docstring, minus the newline, presumably to quiet twistedchecker.

src/twisted/internet/endpoints.py
tomprince: Looking at the tests, everywhere that `._deferToThread` and `._getaddrinfo` are overridden could be replaced by

```
self._nameResolution = fakegetaddrinfo
```

(modulo fakegetaddrinfo taking some additional arguments that it ignores). That would mean that `_nameResolution` isn't covered by the tests but I don't think the coverage that it currently gets is meaningful, since in tests it is just calling fake function that aren't verified to be equivalent to the real thing anyway.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1687,6 +1688,412 @@ class WithResolver(proxyForInterface(
...
+            IPv4Address("TCP", ipv4Host, port),
+            IPv6Address("TCP", ipv6Host, port)
+        ])
+        self.assertEqual(self.resolutionCompleteCallCount, 1)
+
+
+
+class HostnameEndpointFallbackNameResolutionTests(unittest.TestCase):
+    """
+    Verify that L{HostnameEndpoint._fallbackNameResolution} defers a
```
glyph: "verify that" -> https://jml.io/pages/test-docstrings.html

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1687,6 +1688,412 @@ class WithResolver(proxyForInterface(
...
+    name resolution call to a thread.
+    """
+
+    def test_fallbackNameResolution(self):
+        """
+        A L{Deferred} is returned that fires with the resoution of the
```
glyph: Active voice not passive voice, "`_fallbackNameResolution` returns a `Deferred` that..."

src/twisted/internet/test/test_endpoints.py
glyph: Please file & link the bug in twistedchecker so we know about it, per policy :)

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1687,6 +1688,412 @@ class WithResolver(proxyForInterface(
...
+
+        self.assertEqual(host, expectedHost)
+        self.assertEqual(port, expectedPort)
+        self.assertEqual(timeout, expectedTimeout)
+        self.assertEqual(bindAddress, expectedBindAddress)
+
+
+    def test_endpointConnectFailure(self):
+        """
+        If L{HostnameEndpoint.connect} is invoked and there is no server
```
glyph: "when" is preferable to "if", because it stipulates the existence of a supported configuration.

src/twisted/web/test/test_agent.py
glyph: `twisted.testing` HERE WE COME

glyph (approved):
Some minor observations inline, but pending a clean CI run I think this is more than baked.  Please address and land.

src/twisted/internet/test/test_endpoints.py
markrwilliams: Looks like @rodrigc filed the bug: https://github.com/twisted/twistedchecker/issues/124

src/twisted/internet/endpoints.py
markrwilliams: @tomprince I've filed [a new ticket](https://twistedmatrix.com/trac/ticket/9147) that we can use to track this improvement.

markrwilliams:
The failing Codecov status appears to be down to partial coverage on [this line](https://codecov.io/gh/twisted/twisted/pull/740/diff#D2-705).  That could be covered by resolving an IPv6 address. 

We now [skip tests that need IPv6 on platforms that don't support it](http://twistedmatrix.com/trac/ticket/9144), but the way we determine that [needs to be generalized](https://twistedmatrix.com/trac/ticket/9146#ticket).  Rather than introduce another instance of this function's definition in this patch, I'll merge with the slightly reduced coverage and address it in the patch for [9146](https://twistedmatrix.com/trac/ticket/9146#ticket).

PR #742, opened 2017-03-11
[#5573] Support UNIX adoptStreamPort / adoptStreamConnection

exvito:
Trac ticket at https://twistedmatrix.com/trac/ticket/5573

Notes:
* Picked up 2 year old patch by others.
* Addressed 2 year old pending review.
* Added minor changes to make tests pass on Linux + Mac OS.
* I don't quite understand/like the `assert unixEnabled` lines, but left them there for consistency with existing code. Comments?

Side comment:
While going through this, I understood that some UNIX socket code is based on TCP socket code via inheritance. Refactoring such dependencies into common base classes and then have UNIX and TCP implementations (and who knows, others, like bluetooth, or something?) inherit from there would be a cleaner approach -- discussion outside of this PR's scope, of course.


exvito:
Done for the day:
* Appveyor is doing its thing slowwwwwly: there are failures there that will need to be evaluated and addressed.
* Codecov is improved with the last three commits. Not sure yet how to address the SkipTest in test_unix.py:740. Maybe it gets covered once the Windows tests work as they should.

exvito:
Not sure why codecov/patch is failing:
* Line 332 in `src/twisted/internet/unix.py` is not even in the diff.
* Line 437 in `src/twisted/internet/posixbase.py` should have been covered by AppVeyor.

I'll request this PR's review and hopefully someone can point out the culprit.

glyph:
> Not sure why codecov/patch is failing:

Since I have to keep explaining this, I put it on the wiki: https://twistedmatrix.com/trac/wiki/Coverage

exvito:
Thanks @glyph for the input and new wiki entry: it's valuable reference and I was aware of most of it. :-)

After two full reads I still don't get why codecov/patch is failing in this PR for the reasons I stated above: one of the lines isn't on the diff, why is it complaining? the other should have been covered by AppVeyor, but apprently wasn't. :-(

I hope the reviewer(s) can help not only pinpoint my mistake here but also guide the progress of this PR into merging.

Again thanks.

glyph:
> After two full reads I still don't get why codecov/patch is failing in this PR for the reasons I stated above: one of the lines isn't on the diff, why is it complaining? the other should have been covered by AppVeyor, but apprently wasn't. :-(

The only line I can see that it's complaining about is https://codecov.io/gh/twisted/twisted/compare/d8b5236f4df02900a005242362a25de5c6a935c4...972284bd67f6ea67123a644ef2878969c3bd8e2f/diff#D2-888 - what the yellow tag on that line number is telling me is that `reactor.running` is always True.  (The other line appears close enough to the diff to show up but I'm pretty sure it's fine).

Given that this branch is fairly well covered overall, and this may be a race condition that the test is correctly compensating for, I'll give it a pass.  However, it would be good to use some common utility functionality for this kind of `stop` call; this can't be the only test that needs to do `if reactor.running: reactor.stop()` :)

exvito:
@glyph,

Thanks for your review and merging.

Following your codecov link, I now see that one of the lines I saw as not being covered is now green... Maybe codecov was slow in its updates? (strange) Maybe the windows coverage was only updated when the buildbots ran and not with AppVeyor? Not really sure.

My next planned contribution will be simplifying the test in #652, addressing the TODO there, using this newly merged capability.

Again thanks.

PR #744, opened 2017-03-14
OpenSSL 1.1.0 compatibility fixes

exarkun:
http://twistedmatrix.com/trac/ticket/8898


markrwilliams:
@exarkun https://github.com/twisted/twisted/pull/743 :)
Not in review yet because I wanted to test the builders.  But here's what I learned:

1. Minimum RSA key size is 1024 bits: https://github.com/twisted/twisted/pull/743#issuecomment-286463789  - so server.pem doesn't need to be rerun
2. There's a change in the TLS state machine - or somewhere in there - that seems to make early terminated client connections fail with EOF now: https://github.com/twisted/twisted/pull/743/commits/9bf53b71cc5e0f24a0bb616f7455827fb46b29f3

I'm trying to bisect down to the OpenSSL commit that introduced 2 to verify that this is what's actually happening, and that trapping a new error isn't masking a real issue.

PR #745, opened 2017-03-16
Bump version to 17.1.1.dev0 in trunk

rodrigc:
https://twistedmatrix.com/trac/ticket/9083

### DIFF
--- a/src/twisted/_version.py
+++ b/src/twisted/_version.py
@@ -7,5 +7,5 @@ Provides Twisted version information.
 
 from incremental import Version
 
-__version__ = Version('Twisted', 17, 1, 0, dev=0)
+__version__ = Version('Twisted', 17, 1, 1, dev=0)
 __all__ = ["__version__"]


src/twisted/_version.py
```diff
@@ -7,5 +7,5 @@
 
 from incremental import Version
 
-__version__ = Version('Twisted', 17, 1, 0, dev=0)
+__version__ = Version('Twisted', 17, 1, 1, dev=0)
```
exarkun, line 10: The next trunk release will be 17.2.0, not 17.1.1.

exarkun (changes requested):

rodrigc:
@exarkun I'm not sure if the next version of Twisted will be 17.2.0, since the release process has changed: http://twistedmatrix.com/documents/current/core/development/policy/release-process.html#version-numbers
@hawkowl can clarify.

src/twisted/_version.py
rodrigc: I'm not sure if the next version of Twisted will be 17.2.0 or not, since the release process has changed: 

http://twistedmatrix.com/documents/current/core/development/policy/release-process.html#version-numbers

@hawkowl can clarify.

exarkun: Either way, it won't be 17.1.1.

glyph: The problem being solved here (and perhaps someone should put this in the ticket description explicitly) is that a development copy of Twisted trunk after 17.1.0 will not satisfy the pkg_resources version constraint `>=17.1.0`.  The point here is not that the next version will literally be `17.1.1`, but that the development version is `17.1.0+epsilon` so that it will compare as `>17.1.0`.

There are two ways to accomplish this in terms of a version specifier before we know exactly what the next release is going to be; `17.1.0.post1` or `17.1.(something greater than zero).dev`.  In discussing this with craig (which I thought was in a comment, but I guess maybe was on IRC?) we decided that `.dev` was more in the spirit of an "in development" version (since `.post` is more for "fixed release metadata, no code changes"), so `17.1.1` is where we ended up.

rodrigc: Discussion was held here: https://github.com/hawkowl/incremental/issues/29

I haven't been on IRC lately, so did not participate in any discussion there on the topic.

exarkun: Consider these version constraints:

  * `twisted==17.1.*`
  * `twisted>=17.1,<17.2`

One might expect any of these to be satisfied by any of the 17.1.* releases (17.1.0, 17.1.1, 17.1.2, etc).  One might expect it _not_ to include any changes added to trunk@HEAD after the 17.1.x release branch is created.

However, if trunk@HEAD declares itself as `17.1.1.dev0` then it falls within the declared range.  It would be surprising and unfortunate to end up with the wrong version in this case.  And since it is _so_ easily avoided I don't see why we would choose to do it this way.

The 17.1 release branch is the branch that is going to continue on to 17.1.1.  Marking _that_ branch as 17.1.1.dev0 seems unambiguously correct.

The trunk branch is the branch that is going to continue in to `17.(Y>1)` (or possibly `(X>17).Y` if there's some catastrophe).  The problem that it is difficult to predict what X and Y are going to be (aside from knowing the minimum possible values) doesn't seem like a justification to screw up things for the 17.1.x line.

So, please don't put this version into trunk.  Put it into the release branch.  Put 17.3.0.dev0 into trunk.  Or 17.4.0.dev0 if that seems more realistic.  And if it turns out to have the wrong Y component - oh well, that's kind of unavoidable with an unpredictable versioning scheme like the one used by Twisted.  Practice harder to predict the next release more accurately, I guess.  Or let go of the idea that `postN` is only for metadata changes.  But, again, none of that is reason to screw up the 17.1.x line.

(edited to fix the version constraint examples)

glyph: This concern seems spurious to me, but I confess it might be a failure of imagination on my part.

Version constraints are for installing the correct version of a package from a package repository which contains releases.  In what circumstance would you find yourself with a random git revision of Twisted trunk as an installable option?  Pip implicitly excludes prereleases (including any `.devN`, I believe) if they were to happen to be uploaded to a package index; hence the need for the `--pre` flag.  You'd have to explicitly install a working copy of Twisted into an environment in order to trip this condition.  If you did that, you'd also find that pip doesn't include a constraint solver and will happily violate your previously-specified `==` and `>=` constraints.

If we let go of the idea that `.post0` is only for metadata changes, both of your proposed version specifiers still match, so I don't see how that would help.

So I can't imagine a scenario where `17.1.1.dev0` creates any practical problem; `17.1.0.dev0` creates a very real (frequent) annoyance with setuptools entrypoints, detailed on the incremental ticket.

That said: we could just say that the official release version to go on trunk is always just "always increment the minor release number, unless the previous release is 12, in which case increment the major and reset the minor to 1", as this would address your concern about pinning below a particular release, and even if we get the release prediction wrong, it doesn't cause any version specifiers to compare inaccurately.

exarkun: > In what circumstance would you find yourself with a random git revision of Twisted trunk as an installable option?

It's more likely that it would be an _already installed_ version and pip would leave it there as "good enough".  This might happen in the course of trying to integrate Twisted fixes that were developed downstream and need to be pushed upstream (so I want to have a development version of Twisted in my environment to verify the fixes still work).  Once there's a workflow that involves installing a development version of Twisted, I'd really like the tools to protect me from user error.  I don't know about other people but I have a virtualenv working-set size of about twenty and sometimes I get confused about what's supposed to be in different environments (and sometimes I screw up and do something that _puts_ the wrong thing in some of them).

> If you did that, you'd also find that pip doesn't include a constraint solver and will happily violate your previously-specified == and >= constraints.

Yes.  Unfortunately, the safeguards I want don't _really_ exist since the kind of version checks that cause "problems" for entrypoints don't really get applied anywhere else - once a virtualenv has had the wrong version of things installed, oftentimes you're just screwed and it's up to you, the user, to notice and fix things.  However, such checks _could_ eventually be done.  And I could also invoke a manual check myself, for example as part of my test runner startup.

Regarding `--pre`, I don't think it's much relevant given a pre release could be installed explicitly.  However, I'll note that I _do_ often use `--pre` because I have _other_ requirements that I sometimes need a pre release of.  Since `--pre` isn't scoped, once I have to use it for one project I'm at risk for getting a pre release of anything.  So this is another reason to make sure any pre-releases that _might_ get discovered (but I'm not claiming these versions of Twisted will be discoverable in my particular workflow) at least compare correctly against other versions.

> That said: we could just say that the official release version to go on trunk is always just "always increment the minor release number, unless the previous release is 12, in which case increment the major and reset the minor to 1", as this would address your concern about pinning below a particular release, and even if we get the release prediction wrong, it doesn't cause any version specifiers to compare inaccurately.

I think that what's most surprising about push-back to my push-back (I don't know if I would have commented in the first place if I'd know how far this thread would eventually extend) is that the change you suggest here is just as trivial as the change proposed by this PR and it is objectively (it's tempting to say "obviously" as well but perhaps that's just due to my personal perspective) more correct.  It has the small shortcoming that you point out but this seems like a non-issue to me.

rodrigc:
@hawkowl can you comment on this review, since you maintain `incremental`, and also have written down the Twisted release policy regarding version numbers?

I notice you released Twisted 17.1 in February 2017, so that doesn't seem to follow the policy you have written down.

glyph:
The first prerelease was cut in January, which is what gave it the `.1` version number.

adiroiban:
I am +1 for 17.1.1.dev0  ... but the ticket and PR description is brief.

It does not describe how to reproduce the current problem and the steps required to check that this PR is fixing that problem.

So I assumed that in order to check that this is fixed I had to run these commands (adapted from  https://github.com/hawkowl/incremental/issues/29)

``` 
virtualenv 9083-twisted-dev/
cd 9083-twisted-dev/
. bin/activate
pip install "git+https://github.com/buildbot/buildbot#subdirectory=pkg"
pip install "git+https://github.com/buildbot/buildbot#subdirectory=master"
pip install --upgrade "git+https://github.com/twisted/twisted@9083-rodrigc-version"
buildbot create-master
```

but I ended up with this error

```
/home/adi/tmp/9083-twisted-dev ----------------------------------------- @xps
$ buildbot create-master
  File "/home/adi/tmp/9083-twisted-dev/bin/buildbot", line 11, in <module>
    load_entry_point('buildbot===0.9.7-33-g239f842', 'console_scripts', 'buildbot')()
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/scripts/runner.py", line 757, in run
    subcommandFunction = reflect.namedObject(subconfig.subcommandFunction)
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/twisted/python/reflect.py", line 166, in namedObject
    module = namedModule('.'.join(classSplit[:-1]))
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/twisted/python/reflect.py", line 152, in namedModule
    topLevel = __import__(name)
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/scripts/create_master.py", line 30, in <module>
    from buildbot.master import BuildMaster
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/master.py", line 39, in <module>
    from buildbot.buildbot_net_usage_data import sendBuildbotNetUsageData
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/buildbot_net_usage_data.py", line 37, in <module>
    from buildbot.process.buildstep import _BuildStepFactory
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/process/buildstep.py", line 262, in <module>
    util.ComparableMixin):
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/buildbot/process/buildstep.py", line 398, in BuildStep
    @deprecate.deprecated(versions.Version("buildbot", 0, 9, 0))
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/twisted/python/deprecate.py", line 293, in deprecationDecorator
    function, version, None, replacement)
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/twisted/python/deprecate.py", line 238, in getDeprecationWarningString
    _fullyQualifiedName(callableThing), version, format, replacement)
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/twisted/python/deprecate.py", line 198, in _getDeprecationWarningString
    'version': getVersionString(version)}
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/incremental/__init__.py", line 518, in getVersionString
    result = '%s %s' % (version.package, version.short())
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/incremental/__init__.py", line 221, in short
    gitver = self._getGitVersion()
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/incremental/__init__.py", line 422, in _getGitVersion
    return self._parseGitDir(os.path.join(upOneMore, '.git'))
  File "/home/adi/tmp/9083-twisted-dev/local/lib/python2.7/site-packages/incremental/__init__.py", line 389, in _parseGitDir
    with open(headFile, "r") as f:
IOError: [Errno 2] No such file or directory: '/home/adi/.git/HEAD'
```

adiroiban (changes requested):
See my last comment.

Any direction of how to test this is much appreciated

rodrigc:
@adiroiban repo steps:

```
python3 -m venv my_venv
source my_venv/bin/activate

pip install "git+https://github.com/buildbot/buildbot@v0.9.5#subdirectory=pkg"
pip install "git+https://github.com/buildbot/buildbot@v0.9.5#subdirectory=master"
pip install "git+https://github.com/buildbot/buildbot@v0.9.5#subdirectory=worker"
pip install --upgrade "git+https://github.com/twisted/twisted"
```

Now try:

```
buildbot create-master
```

and get:

```
pkg_resources.DistributionNotFound: The 'Twisted>=17.1.0' distribution was not found and is required by buildbot
```

now try:

```
pip install --upgrade "git+https://github.com/twisted/twisted@9083-rodrigc-version"
```

then

```
buildbot create-master
```

and it works.

adiroiban:
Using trunk I got the expected error

```
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/pkg_resources/__init__.py", line 829, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'Twisted>=17.1.0' distribution was not found and is required by buildbot
```
but if I am updating to your branch I still got the same error with incremental, as in the previous report

```
Installing collected packages: Twisted
  Found existing installation: Twisted 17.1.0.dev0
    Uninstalling Twisted-17.1.0.dev0:
      Successfully uninstalled Twisted-17.1.0.dev0
  Running setup.py install for Twisted ... done
Successfully installed Twisted-17.1.1.dev0
(9083-twisted-dev) 
/home/adi/tmp/9083-twisted-dev ----------------------------------------- @xps
$ buildbot create-master
  File "/home/adi/tmp/9083-twisted-dev/bin/buildbot", line 11, in <module>
    load_entry_point('buildbot==0.9.5', 'console_scripts', 'buildbot')()
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/scripts/runner.py", line 757, in run
    subcommandFunction = reflect.namedObject(subconfig.subcommandFunction)
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/twisted/python/reflect.py", line 166, in namedObject
    module = namedModule('.'.join(classSplit[:-1]))
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/twisted/python/reflect.py", line 152, in namedModule
    topLevel = __import__(name)
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/scripts/create_master.py", line 30, in <module>
    from buildbot.master import BuildMaster
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/master.py", line 39, in <module>
    from buildbot.buildbot_net_usage_data import sendBuildbotNetUsageData
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/buildbot_net_usage_data.py", line 36, in <module>
    from buildbot.process.buildstep import _BuildStepFactory
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/process/buildstep.py", line 261, in <module>
    util.ComparableMixin):
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/buildbot/process/buildstep.py", line 380, in BuildStep
    @deprecate.deprecated(versions.Version("buildbot", 0, 9, 0))
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/twisted/python/deprecate.py", line 293, in deprecationDecorator
    function, version, None, replacement)
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/twisted/python/deprecate.py", line 238, in getDeprecationWarningString
    _fullyQualifiedName(callableThing), version, format, replacement)
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/twisted/python/deprecate.py", line 198, in _getDeprecationWarningString
    'version': getVersionString(version)}
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/incremental/__init__.py", line 518, in getVersionString
    result = '%s %s' % (version.package, version.short())
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/incremental/__init__.py", line 221, in short
    gitver = self._getGitVersion()
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/incremental/__init__.py", line 422, in _getGitVersion
    return self._parseGitDir(os.path.join(upOneMore, '.git'))
  File "/home/adi/tmp/9083-twisted-dev/lib/python3.5/site-packages/incremental/__init__.py", line 389, in _parseGitDir
    with open(headFile, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/adi/.git/HEAD'
(9083-twisted-dev) 
```

I think that this is an error with `incremental` and not with your branch...but things are still broken and since exarkun is -1 on this , I don't think that we should hurry and merge it, but instead wait for incremental to be fixed.

I have reported the error https://github.com/hawkowl/incremental/issues/32

rodrigc:
We are now on 17.5.0

rodrigc:
After doing a Twisted release, @hawkowl is now bumping the version number in trunk to have *dev* in it:
https://github.com/twisted/twisted/commit/cebdcb9aae69ed9190e46359fb2415b3e2a13dce

So right now the released version is **17.9.0**, and the version in trunk is **17.9.1.dev0**.
That works for me.

src/twisted/_version.py
rodrigc: @exarkun I'm in agreement with everything you wrote in https://github.com/twisted/twisted/pull/745#discussion_r107141416

rodrigc: @glyph regarding your question:

>  In what circumstance would you find yourself with a random git revision of Twisted trunk as an installable option?

I've been telling a few people to install Twisted from my branch to test out some email fixes directly from GitHub via pip: https://github.com/buildbot/buildbot/issues/3633#issuecomment-330712358

So installing from a random git version is possible.  Therefore it is important to have
the correct version in the repository at all times, even beyond a release version.

rodrigc:
Long term, I think that if we do a release **17.x.0**, then right after the release, the version
should be bumped to **17.x.0.post0**.
This can't happen right now because the **incremental** package does not support **post**.
There is a patch to fix that here:

https://github.com/twisted/incremental/pull/37

src/twisted/_version.py
glyph: @rodrigc What I specifically meant by "an installable option" is, "if I `pip install twisted==X.Y.Z`, when will I get a git version as a result".  To get your version from a git repo, someone would need to do `pip install git+https://.../@...`; i.e. explicitly request installation from a particular artifact, rather than a version specifier.

But, you're right of course, that the version metadata should be as accurate as possible.  My kingdom for a bzr-style revision counter :-)

I had trouble re-understanding this thread, so I went back to [the relevant spec](https://www.python.org/dev/peps/pep-0440/) and wrote out the options I think we discussed so far:

After release X.Y.0, we have:

1. X.Y.0.post0, which is wrong because tooling may rightly consider it equivalent to X.Y.0
2. X.Y.0.dev0, which is wrong because it stipulates an *earlier* version than X.Y.0 according to the standardized sorting algorithm; this causes the aforementioned entrypoints issue, because you want to be able to use the dev version of Twisted with stuff that required at-least-the-previous-version.
3. X.(Y+1).0.dev0 which is wrong because X.(Y+1) may never exist
4. X.Y.1.dev0 which is wrong because X.Y.1 will happen off a different branch, and may not contain any of the changes present in the version of `trunk` it's applied to, etc.

In conclusion:

Upon reflection, I think I didn't really understand @exarkun's complaint, I thought he was suggesting a "Y" where he was actually suggesting "Y+1", and I now agree that option 3 is by far the best option we have.  So: sorry for my pushback, push-un-backed.

(I do still wish we'd just landed the .1 version at the time and had this discussion after the fact though :).)

rodrigc: Yes, Option 3 sounds fine to me.  It basically boils down to whatever @hawkowl chooses, since she's the one controlling the releases and the version numbering.

PR #747, opened 2017-03-19
[#9086] Trim whitespace in gzip encoder's accept-encoding check

hawkowl:
https://twistedmatrix.com/trac/ticket/9086

### DIFF
--- a/src/twisted/web/server.py
+++ b/src/twisted/web/server.py
@@ -17,6 +17,7 @@ from __future__ import division, absolute_import
 
 import copy
 import os
+import re
 try:
     from urllib import quote
 except ImportError:
@@ -568,7 +569,7 @@ class GzipEncoderFactory(object):
 
     @since: 12.3
     """
-
+    _gzipCheckRegex = re.compile(br'(:?^|[\s,])gzip(:?$|[\s,])')
     compressLevel = 9
 
     def encoderForRequest(self, request):
@@ -576,18 +577,17 @@ class GzipEncoderFactory(object):
         Check the headers if the client accepts gzip encoding, and encodes the
         request if so.
         """
-        acceptHeaders = request.requestHeaders.getRawHeaders(
-            'accept-encoding', [])
-        supported = ','.join(acceptHeaders).split(',')
-        if 'gzip' in supported:
+        acceptHeaders = b','.join(
+            request.requestHeaders.getRawHeaders(b'accept-encoding', []))
+        if self._gzipCheckRegex.search(acceptHeaders):
             encoding = request.responseHeaders.getRawHeaders(
-                'content-encoding')
+                b'content-encoding')
             if encoding:
-                encoding = '%s,gzip' % ','.join(encoding)
+                encoding = b','.join(encoding + [b'gzip'])
             else:
-                encoding = 'gzip'
+                encoding = b'gzip'
 
-            request.responseHeaders.setRawHeaders('content-encoding',
+            request.responseHeaders.setRawHeaders(b'content-encoding',
                                                   [encoding])
             return _GzipEncoder(self.compressLevel, request)
 
--- a/src/twisted/web/test/test_web.py
+++ b/src/twisted/web/test/test_web.py
@@ -911,6 +911,25 @@ class GzipEncoderTests(unittest.TestCase):
                           zlib.decompress(body, 16 + zlib.MAX_WBITS))
 
 
+    def test_whitespaceInAcceptEncoding(self):
+        """
+        If the client request passes a I{Accept-Encoding} header which mentions
+        gzip, with whitespace inbetween the encoding name and the commas,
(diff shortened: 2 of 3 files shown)


src/twisted/web/server.py
```diff
@@ -525,7 +525,8 @@ def encoderForRequest(self, request):
         """
         acceptHeaders = request.requestHeaders.getRawHeaders(
             'accept-encoding', [])
-        supported = ','.join(acceptHeaders).split(',')
+        supported = [encoding.strip()
+                     for encoding in ','.join(acceptHeaders).split(',')]
```
dreid: This join/split dance has to allocate roughly `sum((len(h) for h in acceptHeaders)) + len(acceptHeaders) - 1` and then just throws it away.

In practice this is limited to something like 8kb because the total twisted request size is 16k with a maximum of 500 headers, so it's not like a single request is likely to exhaust all your memory by sending the maximum number of `Accept-Encoding` headers. But it's still a lot of malloc and memcpy.

Some simple loops are probably better?

```python
for header in acceptHeaders:
    for encoding in header.split(','):
        if 'gzip' in encoding.strip():
           encoding = ...
           break
```

Of course, split and strip are also going to do a lot of copying, so maybe a regexp instead?

Something along the lines of 
```python
gzipRe = re.compile(r'[\s,]?gzip[\s,]?')
```

and 
```python
for header in acceptHeaders:
    if gzipRe.match('header'):
        encoding = ...
        break
```

glyph:
ubuntu16.04-py2.7-newstyle-coverage is failing. is this a bug in incremental?

hawkowl:
cc @glyph it seems to be fine now :) i think it was a git hiccup?

src/twisted/web/server.py
adiroiban: +1 for simple loops as they are easier to read :)

src/twisted/web/test/test_web.py
```diff
@@ -777,6 +777,25 @@ def test_encoding(self):
...
+    def test_whitespaceInAcceptEncoding(self):
+        """
+        If the client request passes a I{Accept-Encoding} header which mentions
+        gzip, with whitespace inbetween the encoding name and the commas,
+        L{server._GzipEncoder} automatically compresses the data.
+        """
+        request = server.Request(self.channel, False)
+        request.gotLength(0)
+        request.requestHeaders.setRawHeaders(b"Accept-Encoding",
+                                             [b"deflate, gzip"])
```
adiroiban, line 923: I find the current test code hard to read... why we use gzip and not deflate?

I know that we only support .gz and .bz2 ... but the test should not make this assumption,

We might later add support for deflate, and then this test might fail ... and then you might not know why

src/twisted/web/server.py
twm: I don't think that regexp is quite right:

```py
>>> import re
>>> gzipRe = re.compile(br'[\s,]?gzip[\s,]?')
>>> gzipRe.match(b'gzip')
<_sre.SRE_Match object; span=(0, 4), match=b'gzip'>
>>> gzipRe.match(b'x-gzip')
>>> gzipRe.match(b'gzipzzz')  # should not match
<_sre.SRE_Match object; span=(0, 4), match=b'gzip'>
>>> bool(gzipRe.match(b'deflate, gzip'))  # match() is not right
False
```

The regexp should use `^` and `$` to allow matches at the beginning and end of the string. `search()` should be used to allow non-prefix matches:

```py
>>> gzipRe2 = re.compile(br'(:?^|[\s,])gzip(:?$|[\s,])')
>>> bool(gzipRe2.search(b'gzip'))
True
>>> bool(gzipRe2.search(b'deflate, gzip'))
True
>>> bool(gzipRe2.search(b'deflate, gzip,gzip'))
True
>>> bool(gzipRe2.search(b'deflate, gzip,br'))
True
>>> bool(gzipRe2.search(b'deflate,gzip,br'))
True
>>> bool(gzipRe2.search(b'x-gzip,br'))
False
>>> bool(gzipRe2.search(b'deflate,br'))
False
```

I would keep everything as bytes so that we don't need to decode the header value.

src/twisted/web/test/test_web.py
twm: Since this is a test of `GzipEncoderFactory` I don't think that this is particularly a problem. (I'd also wager we won't ever add support for deflate, since it's basically supported by a subset of the clients that gzip is, and historically clients are buggy. Brotli&mdash;`br`&mdash;would be a more useful addition.)

twm (commented):
1. I agree with @dreid that it would be better to avoid join/split in header processing here. Whether to use a regexp or not is up to you. I think that we have already demonstrated the potential hazards there. :)
2. I'd also like to avoid decoding the header values&mdash;`bytes` all the way! This will make Twisted more robust against odd inputs.

Thanks!

PR #749, opened 2017-03-21
Remove small and obscure curves from conch

alex:


markrwilliams (approved):
Nothing to it!  I've pushed the branch to twisted/twisted to get the build bots running.  Once the builds are green I'll merge.

the0id:
I don't understand why this was done.  The curves worked and provided better security.  Why remove them?  I now have to use less secure connections than I did before.

alex:
These curves did not provide better security, many of them provided less than 128-bits of security *by design*. Further, these were all obscure alternate curves which have not been as rigorously reviewed, and OpenSSH did not support any of them.

the0id:
I agree that some of the curves did provide less security, but nistt571 is stronger than nistp521.

And they're all NIST curves, so nistt571 is as rigorously reviewed as nistp256.  Plus they're all recommenced for use by [RFC5656](https://tools.ietf.org/html/rfc5656#page-17).

I do see the argument of wanting to be in line with OpenSSH, but I think it's better to conform to an RFC, with more supported curves, than limiting what it can do to a single application. :)

PR #750, opened 2017-03-22
[#9090] Remove deprecated parameters on h2.H2Connection

alexwlchan:
Using explicit `client_side` and `header_encoding` parameters on
`H2Connection` was deprecated in h2 2.5.0, and is being removed
entirely in h2 3.0.  Switch to the new behaviour, which is using
an `H2Configuration` instance.

https://twistedmatrix.com/trac/ticket/9090#ticket

### DIFF
--- a/src/twisted/web/_http2.py
+++ b/src/twisted/web/_http2.py
@@ -25,6 +25,7 @@ from collections import deque
 from zope.interface import implementer
 
 import priority
+import h2.config
 import h2.connection
 import h2.errors
 import h2.events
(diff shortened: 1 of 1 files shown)


Lukasa:
Hey @alexwlchan! As a note for future Twisted contribution you should mark your Trac ticket as "awaiting review" by adding the "review" keyword and reassigning it to no-one (the empty space at the top of the list of names in the dropdown). However, I'll review this myself for now.

Lukasa (changes requested):
The code changes look fine, however you need to add a topfile which is used to modify the Twisted changelog. In this case, you'll want to put an empty file called `9090.misc` in `src/twisted/web/topfiles`. Mind doing that?

alexwlchan:
> In this case, you'll want to put an empty file called `9090.misc` in `src/twisted/web/topfiles`. Mind doing that?

Done.

Lukasa:
Ok, I'm going to push this to the Twisted buildbots so we get a full build.

Lukasa (approved):
Ok, this looks good to me. Thanks @alexwlchan! :sparkles:

PR #751, opened 2017-03-23
[#9092] Remove deprecated use of .settings and .errors in h2

alexwlchan:
https://twistedmatrix.com/trac/ticket/9092#ticket

### DIFF
--- a/src/twisted/python/_setup.py
+++ b/src/twisted/python/_setup.py
@@ -106,7 +106,7 @@ _EXTRA_OPTIONS = dict(
          'pyobjc-framework-CFNetwork',
          'pyobjc-framework-Cocoa'],
     windows=['pypiwin32'],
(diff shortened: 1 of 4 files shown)


Lukasa (approved):
Cool, I think this looks good. I'm going to get a full builder run.

Lukasa (changes requested):
Yup, that's a thing we missed: there is a test change you need here. The dependencies are tested. Grep for `h2 >= 2.5.0, < 3.0` and you'll find the places you need to change.

Lukasa:
The Windows test failure is not relevant to this PR, so I'm going to ignore it. :smile:

Lukasa:
Ok, I'm happy. Let's do it. Thanks @alexwlchan! :sparkles:

PR #752, opened 2017-03-23
[#9093] A few more h2.errors instances in twisted

alexwlchan:
/me gnashes teeth

Okay, I think this is the final bit. Over to @Lukasa for review. Related to #751.

https://twistedmatrix.com/trac/ticket/9093

### DIFF
--- a/src/twisted/web/test/test_http2.py
+++ b/src/twisted/web/test/test_http2.py
@@ -2060,7 +2060,7 @@ class H2FlowControlTests(unittest.TestCase, HTTP2TestHelpers):
         del frames[-1]
         frames.append(
             frameFactory.buildRstStreamFrame(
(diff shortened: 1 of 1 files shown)


Lukasa:
As ticket 9092 is fixed, you'll need to open a new one I'm afraid. :(

alexwlchan:
> As ticket 9092 is fixed, you'll need to open a new one I'm afraid.

Whaaaaaaaat. Such effort. Much hard. Anyway, done.

> Okay, I think this is the final bit.

This time I’ve grepped the entire codebase for "h2.errors" and "h2.settings", rather than just going on the Travis output. I don’t see anywhere else that needs to change.

Lukasa:
Heh, that seems like the best long-term approach. :grin:

Lukasa (approved):
LGTM. When the builders are green we'll be good to go.

PR #753, opened 2017-03-23
[#9094] pop3 server should omit DELE-ted messages from LIST and UIDL

bra-fsn:
https://twistedmatrix.com/trac/ticket/9094


glyph:
Hi @bra-fsn !

It looks like the test `twisted.mail.test.test_pop3.UtilityTests.testUIDListLineFormatter` is failing.

Could you update this PR so that it passes our CI checks?  We no longer have any special committer-only build infrastructure, and this doesn't look platform specific, so you can verify it using `tox` on your own machine.

When it's ready for a re-review, just reattach the 'review' keyword over in Trac and a reviewer will have another look.

Thanks for your contribution to Twisted!

glyph:
I'll leave the ticket open since it sounds like this issue might still be relevant, but given that this code is failing tests and hasn't been updated in years, I'm going to close the pull request to decrease clutter. Feel free to reopen if you would like to fix it up.

PR #754, opened 2017-03-24
[trac:5310] Make t.web.client.RequestGenerationFailed a public api.

adiroiban:
Scope
=====

See: https://twistedmatrix.com/trac/ticket/5310

Changes
=======

Changes are based on https://github.com/twisted/twisted/compare/trunk...agent-exceptions-5310

I went for a different formatting for long imports, as I find that sorting them in alphabetic order and having a single import per line result in cleaner diffs

How to test
========

here is an example of using the new public export 
```
from twisted.web.client import Agent, RequestGenerationFailed

agent = Agent(reactor)
class Whatever(object):
    """
    I should probably implement some interfaces or something.
    """

def eb_handle(failure):
    """
    The actual traceback is found in the failure list.
    """
    failure.trap(RequestGenerationFailed)
    failure.value.reasons[0].printTraceback()

def bb_done(it):
    reactor.stop()

d = agent.request("POST", "http://example.com/", bodyProducer=Whatever())
d.addErrback(eb_handle)
d.addBoth(bb_done)

reactor.run()
```


glyph:
I agree that the new style of re-organizing the imports is good, but doing it at the same time as the change makes it harder to review the code.  In the future, please do this sort of clean-up either before or after the main change.

adiroiban:
with the existing import style, it would have been harder anyway to review the changes.

I have done a review on the initial patch sent by @exarkun and I had to triple check how the imports were reorganized

this is why I have pushed this is here

it might be a bit harder to review, but I think that if we are not doing these changes are drive by fixed we will never have time to do them...

and even if we have time to do them in separate tickets, the tickets will touch multiple files and it will take a long time for the ticket to be reviewed... and then it will increase the probability of getting of conflicts ...and will make the whole process painful 

sorry for the trouble, but I think that the current diff is not worst that the original one

glyph:
Sorry, I did not mean that to be a review.  What I meant to say was: I would have had time to do a quick review of it yesterday if the import diff had been easy to read.  As it is it will just have to wait for someone with a little more time.  Which is fine :-).

PR #759, opened 2017-04-01
[9101] tcp: adding support for SO_REUSEPORT

wdauchy:
this makes it possible to listen to the same tcp port using different
processes under linux >= v3.9

### DIFF
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -451,7 +451,7 @@ class _TCPServerEndpoint(object):
     A TCP server endpoint interface
     """
 
-    def __init__(self, reactor, port, backlog, interface):
+    def __init__(self, reactor, port, backlog, interface, listenMultiple):
         """
         @param reactor: An L{IReactorTCP} provider.
 
@@ -463,11 +463,15 @@ class _TCPServerEndpoint(object):
 
         @param interface: The hostname to bind to
         @type interface: str
+
+        @param listenMultiple: allow multiple processes to listen to the same port
+        @type listenMultiple: bool
         """
         self._reactor = reactor
         self._port = port
         self._backlog = backlog
         self._interface = interface
+        self._listenMultiple = listenMultiple
 
 
     def listen(self, protocolFactory):
@@ -479,7 +483,8 @@ class _TCPServerEndpoint(object):
                              self._port,
                              protocolFactory,
                              backlog=self._backlog,
-                             interface=self._interface)
+                             interface=self._interface,
+                             listenMultiple=self._listenMultiple)
 
 
 
@@ -487,7 +492,7 @@ class TCP4ServerEndpoint(_TCPServerEndpoint):
     """
     Implements TCP server endpoint with an IPv4 configuration
     """
-    def __init__(self, reactor, port, backlog=50, interface=''):
+    def __init__(self, reactor, port, backlog=50, interface='', listenMultiple=False):
         """
         @param reactor: An L{IReactorTCP} provider.
 
@@ -499,8 +504,12 @@ class TCP4ServerEndpoint(_TCPServerEndpoint):
 
         @param interface: The hostname to bind to, defaults to '' (all)
         @type interface: str
+
+        @param listenMultiple: allow multiple processes to listen to the same port
+        @type listenMultiple: bool
         """
-        _TCPServerEndpoint.__init__(self, reactor, port, backlog, interface)
+        _TCPServerEndpoint.__init__(self, reactor, port, backlog, interface,
+                listenMultiple)
 
 
 
@@ -508,7 +517,8 @@ class TCP6ServerEndpoint(_TCPServerEndpoint):
     """
     Implements TCP server endpoint with an IPv6 configuration
     """
-    def __init__(self, reactor, port, backlog=50, interface='::'):
+    def __init__(self, reactor, port, backlog=50, interface='::',
+            listenMultiple=False):
         """
         @param reactor: An L{IReactorTCP} provider.
 
@@ -520,8 +530,12 @@ class TCP6ServerEndpoint(_TCPServerEndpoint):
 
         @param interface: The hostname to bind to, defaults to C{::} (all)
         @type interface: str
+
+        @param listenMultiple: allow multiple processes to listen to the same port
+        @type listenMultiple: bool
         """
-        _TCPServerEndpoint.__init__(self, reactor, port, backlog, interface)
+        _TCPServerEndpoint.__init__(self, reactor, port, backlog, interface,
+                listenMultiple)
 
 
 
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -910,7 +910,7 @@ class SSL4ServerEndpoint(object):
     """
 
     def __init__(self, reactor, port, sslContextFactory,
-                 backlog=50, interface=''):
+                 backlog=50, interface='', listenMultiple=False):
         """
         @param reactor: An L{IReactorSSL} provider.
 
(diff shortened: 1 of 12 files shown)


wdauchy:
I couldn't validate codecov tests since it requires linux >= 3.9 and not available on windows. However it does work on my local machine. In order to go through codecov, I would have to remove the test.

hawkowl:
@wdauchy Hi! Do you have a ticket on https://twistedmatrix.com/trac that we can reference it against? When we have a ticket number, we can sync it to our buildbot builders (of which some have a kernel version greater than 3.9) so we can run it against the full test suite.

Please file a ticket, or add it to the description, and add the 'review' tag to the trac ticket with a link to this PR, then we can move this along :)

hawkowl:
@wdauchy Oh, I found the ticket. http://twistedmatrix.com/trac/ticket/9101 :)

hawkowl:
Running tests on builders...

wdauchy:
@hawkowl thanks for the update; is there anything else I can do to help? From what I have seen, failing tests are either on windows or ubuntu 12.04 which has linux 3.2

wdauchy:
@hawkowl ping?

hawkowl:
@wdauchy Sorry, there was a problem on our build infra that prevented the tests from going green. Re-running...

wdauchy:
@hawkowl how does it sound to you? is there anything I can improve?

hawkowl:
@wdauchy I think it might be better implemented how we do the other TCP options: https://github.com/wdauchy/twisted/blob/66929c6d2843e23a6bc16deff0ae8d55e979ac2e/src/twisted/internet/tcp.py#L302 , is that possible? It would reduce the API changes a lot (which I think is a bit excessive). You needn't implement it in iocpreactor, as Windows doesn't support it, so you can just do a `twisted.python.runtime.platform.isLinux()` check for the tests.

wdauchy:
@hawkowl thanks for your review. I will remove the iocpreactor part as soon as possible; I indeed did not catch it was windows specific; that is understood now.
I don't think it is possible to implement it as a regular tcp option since it has to be set before binding; so it requires us to pass it as argument; do you see any other way to do it in order to avoid API changes?
for the test, I used "hasattr(socket, "SO_REUSEPORT"), which is supposed to avoid issues on both windows or mac. I can however replace it with platform.isLinux() if you prefer; I might also be wrong if those are defined on windows and mac.

hawkowl:
@wdauchy you're right about the hasattr way, because an isLinux check isn't sufficient on older Linuxes. Right now, SO_REUSEPORT is just a Linux specific addition, I don't think it'll ever get adopted by macOS or Windows.

Let me think about the API ramifications. @glyph , you have any ideas, maybe?

glyph:
`SO_REUSEADDR` on Windows actually behaves a lot like `SO_REUSEPORT` on Linux; however, it describes this behavior as "non-deterministic" and mostly seems to talk about it as a security hole.  See here:  See here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx

Someone with more time than me should do the research into how to get consistent behavior between platforms, and add an API that does that, rather than trying to make the exact flags (which mean different things anyway) apply on both platforms.

wdauchy:
@hawkowl @glyph so on windows we  may set SO_REUSEADDR in case of multiListen set and SO_EXCLUSIVEADDRUSE otherwise
I don't see how to get a real consistent behavior here, since the exact reuseport functionnality is simply not supported elsewhere.

For now, I put back the IOCP interfaces, since it was breaking the endpoints interface; I did not saw a clean way to avoid it.

src/twisted/internet/tcp.py
```diff
@@ -961,6 +964,8 @@ def createInternetSocket(self):
         s = base.BasePort.createInternetSocket(self)
         if platformType == "posix" and sys.platform != "cygwin":
             s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            if self.listenMultiple and hasattr(socket, "SO_REUSEPORT"):
```
moshez, line 967: This should probably fail if listenMultiple has been requested and we cannot do it, since it literally changes semantics (e.g., we can't run the program twice).

src/twisted/internet/endpoints.py
```diff
@@ -463,11 +463,15 @@ def __init__(self, reactor, port, backlog, interface):
 
         @param interface: The hostname to bind to
         @type interface: str
+
+        @param listenMultiple: allow multiple processes to listen to the same port
+        @type listenMultiple: bool
```
moshez, line 468: This feature is called:

* listenMultiple (in some places)
* multiListen (in others)
* SO_REUSEPORT (by the kernel level documentation that someone will read and wonder "does Twisted support it)

I suggest collapsing all of these into a single name "reuseport" (lose the SO), thus keeping close to the kernel terminology and not inventing our own.

src/twisted/internet/interfaces.py
```diff
@@ -728,7 +728,7 @@ def lookupZone(name, timeout=None):
 
 class IReactorTCP(Interface):
 
-    def listenTCP(port, factory, backlog=50, interface=''):
+    def listenTCP(port, factory, backlog=50, interface='', listenMultiple=False):
```
moshez, line 731: This is changing an interface to add a new optional parameter.

This is a backwards incompatible change -- since someone *external* to Twisted who has written their own reactor is now no longer complying with this interface.

If we want to do a backwards incompatible change, we should go by the appropriate process -- probably following the "exceptions" section in https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#procedure-for-exceptions-to-this-policy

An alternative is to add reuse port in a separate interface (e.g., :code:`IReactorTCPListenMulti`), which we would only add to the reactor if we detected SO_REUSEPORT in socket)

Also -- presumably we would like to be able to REUSEPORT on IPv6 as well?
(Either way, whether we do a backwards incompatible change or if we add a new interface,
it would be worthwhile to chunk those into one change.)

OTOH -- I'm not that sure we need the support in listenSSL -- since that functionality can easily be accomplished with the txsni plugin (and we probably want to get away from SSL having to lock-step with TCP)

moshez (changes requested):
Thanks for working on this! I've recently had the occasion to have to shim in reuseport support into Twisted for an example, and that was pretty painful -- so it would be wonderful to have it in the core.

I've listed a few things that would be better done in a different way.

moshez:
FYI -- there has been no response on this for about 6 months. I am going to continue working on this in #994 . Once it is green on buildbot, I will close this one with a link.

wdauchy:
@moshez sorry for the lack of reply.

wiml:
> Someone with more time than me should do the research into how to get consistent behavior between platforms, and add an API that does that, rather than trying to make the exact flags (which mean different things anyway) apply on both platforms.

FWIW, [this Stack Overflow reply](https://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t#answer-14388707) does an excellent job of covering different OS's various `REUSE` options, and how there are at least two distinct behaviors being controlled by `SO_REUSE`*foo* sockopts.

wbarnha:
Has this been abandoned? I would really like to get support for SO_REUSEPORT.

adiroiban:
The last update on this PR was in 2017/2018.
I would say this is not active.

glyph:
> Has this been abandoned? I would really like to get support for SO_REUSEPORT.

@wbarnha The original author definitely seems uninterested in addressing the feedback and conflicts, so I'm going to go ahead and close this PR. If you'd like to see this feature, open a new PR to address #9101 (maybe have a look at #994 too, to see if there's anything useful to be gleaned there?), and feel free to base it on this code.

wdauchy:
> @wbarnha The original author definitely seems uninterested in addressing the feedback and conflicts, so I'm going to go ahead and close this PR. If you'd like to see this feature, open a new PR to address #9101 (maybe have a look at #994 too, to see if there's anything useful to be gleaned there?), and feel free to base it on this code.

Feel free to follow-up indeed. I thought someone had done the work since then.

PR #761, opened 2017-04-06
update _cryptography_backports.py

ofek:


rodrigc (changes requested):
This change looks OK.
However, you need to follow the process at: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

* you need to create a ticket at twistedmatrix.com
* you need a newsfragment

These things are explained in the link I listed above.

ofek:
See #873

PR #763, opened 2017-04-10
[#8133] Port to towncrier

hawkowl:
https://twistedmatrix.com/trac/ticket/8133


src/twisted/web/newsfragments/.gitignore
```diff
@@ -0,0 +1 @@
+!.gitignore
```
glyph, line 1: What is the significance of all these `.gitignore` files?

hawkowl: They prevent git from removing the folder after the RM runs towncrier.

hawkowl: (this was done by the old `NEWS` files before this change)

glyph: It would seem that this would be more clearly expressed by just having a file called `.exists-for-towncrier` in each directory?

src/twisted/news/topfiles/NEWS
```diff
@@ -1,267 +0,0 @@
-Ticket numbers in this file can be looked up by visiting
```
glyph, line 1: Why are we deleting all the NEWS files?

NEWS
```diff
@@ -1,6 +1,9 @@
 Ticket numbers in this file can be looked up by visiting
 http://twistedmatrix.com/trac/ticket/<number>
 
+.. towncrier release notes start
```
glyph, line 4: Is this intended to be parsed as ReST? If so, would it make sense to rename this file `NEWS.rst`?

src/twisted/news/topfiles/NEWS
hawkowl: they deduplicate the main NEWS.rst and will no longer be updated

NEWS
hawkowl: It would make sense, yes!

hawkowl: done

src/twisted/web/newsfragments/.gitignore
hawkowl: this is the git idiom, but i can change it if you hate it that much

glyph: But it doesn't exist for git, it exists for towncrier.  (I don't even know what a `!` means in this file, and if I saw these lying around without this context, I'd delete them because I would have assumed some tool mistakenly created them.)

adiroiban: I think that a better fix would be to have a README file which will explain what developers should put into it and the basic rules... maybe just a link to the documentation

markrwilliams: `git` is terrible, but this is an idiom.  Even if it's a dumb idiom I don't want Twisted to have an exception, especially for something that's supposed to make contributing easier for newcomers.

The file stays in the picture.

markrwilliams (approved):
The stage is set - once the builds are green I'll merge.

rodrigc:
@hawkowl @markrwilliams This section: http://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles should be changed to mention *newsfragments* directories instead of *topfiles*.

markrwilliams:
@rodrigc [Done](https://twistedmatrix.com/trac/wiki/ReviewProcess?sfp_email=&sfph_mail=&action=diff&version=80&old_version=79&sfp_email=&sfph_mail=).  Thanks!

rodrigc:
@markrwilliams I think it is appropriate to mention the new requirement to populate the newsfragments directory.  This is a visible change that surprises people who have topfiles in their branch

PR #764, opened 2017-04-10
[#9111] Replace twisted.python.deprecate with eventually

hawkowl:
https://twistedmatrix.com/trac/ticket/9111#ticket


glyph:
Closing due to the massive number of merge conflicts; probably doing this again from scratch would be easier. Looks like the code review already suggested that the naming should be moving in a different direction anyway…

Thanks for trying to get this done nonetheless hawkie!

PR #748, opened 2017-03-21
9019 conch remove diffie hellman group1 sha1

markrwilliams:
https://twistedmatrix.com/trac/ticket/9019

PR #765, opened 2017-04-10
Re-adding in the RFC 5656 recommended curves and DH-Group1.  

the0id:
This is effectively rolling back #748 and #749.

I think removing these curves is a bad idea.  The curves worked, and in the case of nistt571, provided stronger encryption that is available now.  Plus the removed curves are the ones recommended by [RFC 5656](https://tools.ietf.org/html/rfc5656#page-17).  Re-enabling support these curves both increases security and usability.

I also think it's worth re-enabling DHGroup1.  I know this exchange has been deprecated by OpenSSH, and for good reason, but I don't think that's a good enough reason to remove it from Twisted.

DHGroup1 is still widely supported by many clients, including OpenSSH, and thus supporting it makes Twisted more compatible with those clients.  And there isn't a security risk in supporting DHGroup1.  It is now at the bottom of the list of supported key exchanges and so it will only ever be used if no other exchange is supported.  When this happens we either have the choice of allowing the user do what they need to do, or make them find another client because we don't like the exchange.

### DIFF
--- a/src/twisted/conch/ssh/_kex.py
+++ b/src/twisted/conch/ssh/_kex.py
@@ -132,7 +132,7 @@ class _DHGroup14SHA1(object):
     (2048-bit MODP Group). Defined in RFC 4253, 8.2.
     """
 
-    preference = 7
+    preference = 6
     hashProcessor = sha1
     # Diffie-Hellman primes from Oakley Group 14 (RFC 3526, 3).
     prime = long('32317006071311007300338913926423828248817941241140239112842'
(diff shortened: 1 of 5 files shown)


hawkowl:
group1-sha1 is disabled by default in modern (secure) OpenSSHs, and what a 8+ year old RFC recommends, security-wise, is not what we should be reintroducing. If this affects anyone, they should update their OpenSSH (or Twisted SSH server) to a supported-by-upstream version. Sorry.

glyph:
@the0id If there is some specific reason besides RFC compliance (which, while important, is always secondary to security, and I trust @alex's analysis here) to want to include some of these curves, please raise it on the mailing list so we can discuss them separately.  Trying to re-enable these all at once is definitely a losing proposition though :)

PR #768, opened 2017-04-18
Adding a public API to select kex and public key algorithms.

the0id:
This creates a public API that allows a developer to specify what key exchanges or public key algorithms to use.  Including those not part of the Twisted default algorithms.

As I mentioned in the [ticket](https://twistedmatrix.com/trac/ticket/9118) the unit tests aren't completed yet because I wanted to make sure this would be acceptable first.


the0id:
I don't understand why I'm not getting 100% coverage.  Codecov says that line 599 is not being covered, but when I step through in the debugger its covered by both test_addSupportedKeys tests in test_transport.py.

@markrwilliams maybe this is another one of your surprises? :)

glyph:
> I don't understand why I'm not getting 100% coverage. Codecov says that line 599 is not being covered, but when I step through in the debugger its covered by both test_addSupportedKeys tests in test_transport.py.

The problem is that this hasn't yet been pushed to a Twisted branch, so the buildbots haven't run on your latest commit.

the0id:
@glyph thanks for the reply, but I was asking about Codecov/patch saying I only have 99.31% coverage not the buildbot.

From what I can tell the line should be covered but codecov disagrees with me and for some reason it's word is better than mine. :)

glyph:
> I was asking about Codecov/patch saying I only have 99.31% coverage not the buildbot.

When you are stepping through that line, are you running the tests in a manner that they are run on travis?  Or is your environment more like one of the buildbots?

the0id:
> When you are stepping through that line, are you running the tests in a manner that they are run on travis? 

Yes but no.

It was a problem that I was running `trial twisted.conch.test.test_transport` to step through the lines because running the complete `trial twisted` causes problems with my debugger.  When `twisted.conch.test.test_transport` ran `ssh-dss` was in the `supportedPublicKeys` variable, and thus line 599 was covered, but not when `trial twisted` ran.

This was because I was neglecting how the memory was being allocated when using the API.  The behavior was as intended, but it wasn't doing it the right way.  The update should fix that.

But I seem to be having some problems with Travis and Appveyor.  They sometimes seem to be failing on random tests that aren't related to the code I'm changing.  Am I the only one this is happening to or is it a known issue?

glyph:
> I seem to be having some problems with Travis and Appveyor. They sometimes seem to be failing on random tests that aren't related to the code I'm changing. Am I the only one this is happening to or is it a known issue?

You seem to have gotten particularly unlucky, but:

1. Yes, it's a known issue.
2. We are working on it: It's much better now than it's been in the past, and fixing build reliability issues is (to the extent that it is even possible for us to set priorities on a volunteer-driven project) a high priority.  Please report any tracebacks you see that are truly unrelated.
3. It's presently tracked here: https://twistedmatrix.com/trac/ticket/8879 and there is an accompanying PR that we just re-build every so often to see if it fails.  (It needs its newsfile updated.)

the0id:
> You seem to have gotten particularly unlucky

Story of my life. :)

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+         @type publicKeys: L{dict}
+         @param publicKeys: A L{dict} with the name of the public key as the key
+            and a value that is a L{tuple} consisting of an instance of the key method,
+            a L{str} that is the corresponding sec name.
+        """
+        del self.supportedPublicKeys[:]
+        self.addSupportePublicKeys(publicKeys)
+
+
+    def addSupportePublicKeys(self, publicKeys, sort=True):
```
markrwilliams: Typo - addSupporte**d**PublicKeys

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+        @param kexes: A L{list} of kex name strings
+        """
+        # Clear out the currently supported key exchanges.
+        _kex._kexAlgorithms.clear()
+
+        # Set the new exchanges.
+        self.addSupportedKexes(kexes)
+
+
+    def addSupportedKexes(self, kexes):
```
markrwilliams, line 488: This is an instance method that modifies a module-level dictionary.  Adding a key exchange algorithm for one transport would affect all transports, and that would certainly surprise and concern me!

It appears that [key exchange negotiation uses the `supportedKeyExchanges` instance variable](https://github.com/twisted/twisted/blob/3d11cba9fe1f50fc55cb5622d4f47dfcc9c9f204/src/twisted/conch/ssh/transport.py#L867-L875), so it looks like it's possible to add a key exchange algorithm by changing this variable, which would in turn only allow it for a specific transport.  I might be missing some subtlety, though!

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+        @param publicKeys: A L{dict} with the name of the public key as the key
+            and a value that is a L{tuple} consisting of an instance of the key method,
+            a L{str} that is the corresponding sec name.
+        """
+        for k in publicKeys:
+            if k not in self.supportedPublicKeys:
+                self.supportedPublicKeys.append(k)
+
+            if k.startswith(b'ecdsa'):
+                curve = publicKeys[k]
```
markrwilliams, line 547: Is it possible not to modify this module-level state?

src/twisted/conch/test/test_transport.py
```diff
@@ -1285,6 +1288,122 @@ class ServerSSHTransportTests(ServerSSHTransportBaseCase, TransportTestCase):
...
+
+        del transport.SSHTransportBase.supportedPublicKeys[:]
+        for key in basekeys:
+            transport.SSHTransportBase.supportedPublicKeys += [key]
+
+        keys._curveTable = old_curve.copy()
+        keys._secToNist = old_sec.copy()
+
+
+    def test_setSuppotedKeys(self):
```
markrwilliams: Typo - test_setSuppo**r**tedKeys

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
     _blockedByKeyExchange = None
 
+
+    def setSupportedKexes(self, kexes):
+        """
+        Clear out the list of supported key exchanges
+        Then call addSupportedKexes to populate the list.
+
+        @type kexes: L{list}
+        @param kexes: A L{list} of kex name strings
```
markrwilliams: `kexes` is a list of `bytes`.

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+        # Set the new exchanges.
+        self.addSupportedKexes(kexes)
+
+
+    def addSupportedKexes(self, kexes):
+        """
+        Append the list of supported key exchanges.
+
+        @type kexes: L{list}
+        @param kexes: A L{list} of kex name strings
```
markrwilliams: `kexes` is a list of `bytes`.

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+
+
+    def addSupportePublicKeys(self, publicKeys, sort=True):
+        """
+        Populate the supported public key list.
+
+        @type publicKeys: L{dict}
+        @param publicKeys: A L{dict} with the name of the public key as the key
+            and a value that is a L{tuple} consisting of an instance of the key method,
+            a L{str} that is the corresponding sec name.
```
markrwilliams: a **L{bytes}** that is...

src/twisted/conch/ssh/transport.py
markrwilliams: The `sort` parameter isn't documented.

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+            if k not in self.supportedPublicKeys:
+                self.supportedPublicKeys.append(k)
+
+            if k.startswith(b'ecdsa'):
+                curve = publicKeys[k]
+                keys._curveTable[k] = curve[0]
+
+                keys._secToNist[curve[1]] = k
+
+        # Sort the keys so the strongest ones are preferred?
```
markrwilliams, line 552: Is the question mark here an indication that you're not sure if this is the right way to sort the supported public key types?

src/twisted/conch/ssh/transport.py
```diff
@@ -516,6 +515,99 @@ class SSHTransportBase(protocol.Protocol):
...
+        if sort:
+            # Reverse sort to bring the stronger keys up first.
+            self.supportedPublicKeys.sort(reverse=True)
+
+            # If 'ssh-rsa' or 'ssh-dss' are present, remove and add
+            # to make sure they're in the right order
+            if b'ssh-rsa' in self.supportedPublicKeys:
```
markrwilliams, line 559: I'm also unsure about the sort order!  This moves RSA and DSA key types to the end of the list, but what about the relative order of elliptic curve key types?

[`openssh` allows users to specify the complete list, to remove items from the default, or to _append_ items to it](https://man.openbsd.org/sshd_config#PubkeyAcceptedKeyTypes).  That seems like an acceptable API.  This could also be applied to changing the supported key exchange algorithms.

src/twisted/conch/test/test_ssh.py
```diff
@@ -843,7 +848,7 @@ class SSHFactoryTests(unittest.TestCase):
 
     def makeSSHFactory(self, primes=None):
         sshFactory = factory.SSHFactory()
-        gpk = lambda: {'ssh-rsa' : keys.Key(None)}
+        gpk = lambda: {b'ssh-rsa' : keys.Key(None)}
```
markrwilliams, line 857: Whoops!  Good catch!

src/twisted/conch/test/test_transport.py
```diff
@@ -1732,14 +1851,165 @@ def setUp(self):
...
```
markrwilliams: Typo - test_setSuppo**r**tedKeys.

markrwilliams (changes requested):
This is a welcome addition, and the API generally looks good!

Can you work on a way to set key exchange algorithms for a single transport and also remove the sorting logic for public key types?

src/twisted/conch/ssh/transport.py
the0id: The subtlety is how the list is ordered. :)

Before Twisted ordered the kexes based on their preference ([EC screwed that up a little](http://twistedmatrix.com/trac/ticket/9155)), and to keep doing that I recreate the kex algorithms dictionary and modified `_kex.getSupportedKeyExchanges()`.

But you're right, it shouldn't change things at a module level, so I've made changes to this part of the code so `kexAlgorithms` is now a SSH*Transport class variable.

src/twisted/conch/ssh/transport.py
the0id: It is possible, but _curveTable and _secToNist are just reference tables.  Just having a curve in there doesn't mean it's advertised as being supported during negotiations, and I feel that it could cause more issues making them instance variables.

src/twisted/conch/ssh/transport.py
the0id: Has anything I've ever done suggest I'm sure about anything? :)

This is to indicate that sorting is dependent on the sort parameter.

src/twisted/conch/ssh/transport.py
the0id: Neither am I (see above).

Sorting public keys is a problem because there's no way to designate which one is preferred like there is with key exchanges.  But like kexes they should be ordered so the most preferred ones are used.

The default curves can put the strongest key first by a reverse list sort, and then `ssh-rsa` and `ssh-dsa` can be appended on after.  This is a good first step, but there's room for improvement.  And if `sort` is False then the order will be whatever is passed into `addSupportedKexes()`

I think the link you provided would work if in the future Twisted supported adding keys via a cli argument. Plus the API provided allows for the addition of crypto backends to be included for ciphers that aren't supported by Cryptography.

glyph:
@the0id it looks like this needs a *lot* of conflict resolution 😢 .  Do you think you could poke at it again?

PR #769, opened 2017-04-20
9100-evilham-pop3-spaces-in-pass

evilham:
PR Fixing 9100: http://twistedmatrix.com/trac/ticket/9100


src/twisted/mail/pop3.py
```diff
@@ -36,7 +36,7 @@
 from twisted import cred
 
 ##
-## Authentication
+#  Authentication
```
exarkun: Probably should leave the comment style alone unless you want to make a case for  changing it throughout the module.

src/twisted/mail/pop3.py
```diff
@@ -855,7 +858,13 @@ def do_PASS(self, password):
 
         @type password: L{bytes}
         @param password: A password.
+
+        @type args: L{tuple} or L{None}
+        @param args: Other words composing the password.
```
exarkun: It can't actually be `None` because it's `*args`.  It can be any tuple, including an empty tuple.

src/twisted/mail/pop3.py
```diff
@@ -565,7 +565,8 @@ def state_COMMAND(self, line):
             return self.processCommand(*line.split(' '))
         except (ValueError, AttributeError, POP3Error, TypeError) as e:
             log.err()
-            self.failResponse('bad protocol or server: %s: %s' % (e.__class__.__name__, e))
+            self.failResponse(
+                'bad protocol or server: %s: %s' % (e.__class__.__name__, e))
```
exarkun: If you want to change these lines, I'd suggest this style:

    self.failResponse(
                'bad protocol or server: %s: %s' % (e.__class__.__name__, e)
    )

src/twisted/mail/pop3.py
```diff
@@ -855,7 +858,13 @@ def do_PASS(self, password):
 
         @type password: L{bytes}
         @param password: A password.
+
+        @type args: L{tuple} or L{None}
+        @param args: Other words composing the password.
         """
+        if args:
+            password += ' ' + ' '.join(args)
```
exarkun: Or `password = " ".join((password,) + args)`.

Regardless of that, `b" "` - not `' '`.  Byte strings need to be explicit now.

src/twisted/mail/test/test_pop3.py
```diff
@@ -1,3 +1,4 @@
+# -*- test-case-name: twisted.mail.pop3 -*-
```
exarkun: This is backwards.  `twisted.mail.test.test_pop3` is the test-case-name for twisted/mail/pop3.py.  Declaring this variable with this value here won't do anything.

src/twisted/mail/test/test_pop3.py
```diff
@@ -382,51 +427,56 @@ def testAuthListing(self):
...
 
+
     def _cbTestAuthListing(self, ignored, client):
         self.assertTrue(client.response[1].startswith('+OK'))
         self.assertEqual(sorted(client.response[2:5]),
                          ["AUTH1", "AUTHLAST", "SECONDAUTH"])
         self.assertEqual(client.response[5], ".")
 
+
     def testIllegalPASS(self):
```
exarkun: Please add a docstring for this modified test and rename it to follow the naming convention - `test_illegalPASS`.

src/twisted/mail/test/test_pop3.py
```diff
@@ -382,51 +427,56 @@ def testAuthListing(self):
...
-        return d.addCallback(self._cbTestEmptyPASS, client, dummy)
+        return self.runTest(
+            ['PASS ',
+             'QUIT'],
+            ['+OK <moshez>',
+             '-ERR USER required before PASS',
+             '+OK '])
+
+
+    def testWhiteSpaceInPASS(self):
```
exarkun: Please add a docstring for this new test method and rename it to follow the naming convention.  Also, please add more test coverage for the new functionality.  For example, multiple words separated by a space and multiple consecutive spaces.  Other kinds of whitespace would also be an interesting thing to test (I don't think that's broken but having the test coverage would be beneficial).

Also, a positive-path test where passwords with spaces in them lead to an authentication _success_ would be good.  Otherwise the test suite just shows it isn't a parse error - it doesn't show it actually works.

src/twisted/mail/test/test_pop3.py
```diff
@@ -1056,12 +1133,14 @@ def _flush(self):
...
 
+
+
 class POP3MiscTests(unittest.TestCase):
     """
     Miscellaneous tests more to do with module/package structure than
     anything to do with the Post Office Protocol.
     """
-    def test_all(self):
+    def testAll(self):
```
exarkun: This is a transformation in the wrong direction. :smile:   Please revert.

src/twisted/topfiles/9100.bugfix
```diff
@@ -0,0 +1 @@
+twisted.mail.pop3 now accepts passwords that contain spaces.
```
exarkun: Very nice.  I'd say `twisted.mail.pop3.POP3 ...` or `The POP3 server implemented by twisted.mail.pop3 ...` though to avoid client/server confusion

exarkun (changes requested):
Thanks!  This is quite a good contribution, particularly for your first.  I've left a few comments inline.  Sorry they're not all for things you could have discovered on your own by looking at developer docs.  I hope they're all clear and reasonable.  Please let me know if the case appears to be otherwise.

Apart from the inline comments, I'd also like to suggest that formatting changes (like inserting blank lines and re-wrapping lines to <80 columns) are better done as a separate patch - either before or after the more interesting change you want to submit.  It makes the review of both pieces easier.

Thanks again.

evilham:
Great! Thanks for the comments, will fix 'em later.
However:
some of the changes I had done, were a direct consequence of running `twistedchecker` and trying to kill some Warnings / Errors it was giving. Does that mean, we should file a bug (or maybe even a couple) against `twistedchecker`?
I mean it particularly because of the `POP3MiscTests.testAll` / `test_all` naming thingy :).

I clarify, I picked this bug as an easy one to get to know Twisted's ways of contributing, and maybe even improving the documentation for that ;).

src/twisted/mail/pop3.py
evilham: My bad, I copied the description from `do_AUTH` where that of `processCommand` would have been much better. Will change to:
```
@type args: L{tuple} of L{bytes}
@param args: Other words composing the password to be joined with a space.
```

src/twisted/mail/test/test_pop3.py
evilham: :-D I wasn't sure at all with this one, this needs to be better in the Wiki! What does it do when properly set up anyway?

src/twisted/mail/pop3.py
evilham: This brings me down a rabbit hole: doesn't that mean stuff like `return self.processCommand(*line.split(' '))` in line 565 should use `b' '` as well? On a quick check, I see this overall in `pop3.py`.
If that's the case, am I correct to assume we should open a new ticket and fix it there?

exarkun: Those are also good changes to make.  And filing a new ticket to handle those changes is the right idea, yea.  These particular strings should *probably* still change since they're new in this branch (but you *could* also leave them as long as you promise to fix the whole module after this branch is merged :smile: ).

src/twisted/mail/test/test_pop3.py
exarkun: It does a few things:

  * It defines a buffer-local variable when the file is opened in emacs.  The variable name is `test-case-name` and the value is `twisted.mail.pop3`.
  * This, in turn, advises emacs' *twisted-dev-mode* that when you hit F9 to run the test suite for this file it should pass `twisted.mail.pop3` to trial (this is why it doesn't make sense - `trial twisted.mail.pop3` won't run any tests).
  * It also advises the trial itself what tests to run (see the `--testmodule` option).
  * And this is also used by the BuildBot "quick" builder to run a subset of the test suite related to files being changed to try to get some results back more quickly than if you have to run the whole test suite.  However, I think the "quick" builder was removed a long time ago.

exarkun:
> Does that mean, we should file a bug (or maybe even a couple) against twistedchecker?
> I mean it particularly because of the POP3MiscTests.testAll / test_all naming thingy :).

Hmmm yes, sounds like it.  twistedchecker should definitely not be telling people to rename things like `test_all` to things like `testAll`.

evilham:
@exarkun: can we merge this branch and close this ticket?

I feel like I'm fixing other stuff due to the rookie mistake of messing up with coding standards on the same PR :).

Also, I think the pop3 module is simple enough for me to haunt some of its bugs listed [here](https://twistedmatrix.com/trac/query?status=assigned&status=new&status=reopened&component=mail&order=priority).

Incidentally, I just opened [Trac Issue 9120](https://twistedmatrix.com/trac/ticket/9120) referring to this module's Python3 support and coding standards.

I guess it's a good idea to bring up some of the issues we detected in the mailing list?

glyph:
The one remaining problem here is the minor gap in diff coverage; it looks like `tried_user` is always passed in https://codecov.io/gh/twisted/twisted/compare/d3a7f6a5f8f673e52398f630566a303edf6e82d3...05f23e3c0ff5cd0ef5ff84c404c1dd6b88603f5c/diff#D1-521 and this log.err call is never invoked: https://codecov.io/gh/twisted/twisted/compare/d3a7f6a5f8f673e52398f630566a303edf6e82d3...05f23e3c0ff5cd0ef5ff84c404c1dd6b88603f5c/diff#D2-792

Can you address these two minor issues, and merge trunk, so we can land?  Thanks!

evilham:
Hi @glyph,

As for the `tried_user` argument:
that should not be the case, on line `pop3.py:521` there is a test calling `self.run_PASS(b'testuser', b'fooz barz')`; which means it should branch into that line.
My local run of `trial --coverage` does not show an issue here.

I reverted the indentation change that was causing the coverage gap. As mentioned before, that was being caused by the rookie mistake of mixing coding standards fix and bug fix on the same PR.

I'd think this is acceptable, as the thing keeping me from working on [Trac Issue 9120](https://twistedmatrix.com/trac/ticket/9120) is precisely that this hasn't landed :-). 9120 should result in Python3 support in `t.m.pop3` and better tests and coverage for the module.

PS: Travis says 4190.6 fails because of topfile and there is a topfile, am I doing this wrong?

evilham:
My bad, I was also interpreting it as `tried_user` never being passed, turns out, the case that was not covered was it _being_ passed :).

Just added a test to account for a wrong user being passed and that should solve the partial hit.

glyph:
> PS: Travis says 4190.6 fails because of topfile and there is a topfile, am I doing this wrong?

The topfile has changed recently with our recent conversion from our newsfiles stuff to towncrier.  You just need to put it in a different directory now.

https://twistedmatrix.com/trac/ticket/8133

evilham:
Thank you, I had missed that and didn't think to re-check the docs.
Build is fine now in addition to yesterday's coverage-related changes.

src/twisted/mail/pop3.py
```diff
@@ -565,7 +565,9 @@ def state_COMMAND(self, line):
             return self.processCommand(*line.split(' '))
         except (ValueError, AttributeError, POP3Error, TypeError) as e:
             log.err()
-            self.failResponse('bad protocol or server: %s: %s' % (e.__class__.__name__, e))
+            self.failResponse(
+                        'bad protocol or server: %s: %s' % (e.__class__.__name__, e)
```
adiroiban, line 569: I am trying to understand what are the rules for this indentation... I see this double indentation in other parts of this PR.

adiroiban:
wow... GitHub with "Update this branch" allows me to trigger a merge in others repos.

I hope that with the updated twistedchecker, this PR will get all the errors related to the coding convention

rodrigc (changes requested):
@evilham Can you close this PR, and submit a new one based on latest trunk?

The code changed a lot in trunk after https://github.com/twisted/twisted/pull/878

adiroiban:
I assume that you can just update with trunk and fix the conflicts and push a new change to this same PR. no need to close this one and open a new one

rodrigc:
@adiroiban No, actually in this case I prefer a new PR, thanks.

exarkun:
> @adiroiban No, actually in this case I prefer a new PR, thanks.

Why?  Updating the branch behind a PR is standard practice.

rodrigc:
@exarkun Normally I agree but in this case, @Evilham made a lot of style changes which overlap with other changes made elsewhere, so to make a clean start, I'm requesting a new patch.

While it may be a "standard practice", I don't believe in one size fits all practices, and it isn't too hard to submit a new PR, which is why I asked.

evilham:
Great, will close now and open a new one with _only_ the code changes.
Note to self: update trac.

PR #770, opened 2017-04-20
9116: add failAfterFailures= to ClientService.whenConnected()

warner:
Adds a new argument "failAfterFailures=" to whenConnected(), allowing applications to detect when the service has failed to connect after some number of attempts.

refs ticket:9116 https://twistedmatrix.com/trac/ticket/9116


docs/core/howto/endpoints.rst
```diff
@@ -152,6 +152,26 @@ Keep in mind that you may need to wrap this up for your particular application,
 For example, that little snippet is slightly oversimplified: at the time ``connectedNow`` is run, the bot hasn't authenticated or joined the channel yet, so its message will be refused.
 A real-life IRC bot would need to have its own method for waiting until the connection is fully ready for chat before chatting.
 
+Reporting an Initial Failure
+----------------------------
+
+Often times, a failure of the very first connection attempt is special. It may indicate a problem that won't go away by just trying harder. The service may be configured with the wrong hostname, or the user may not have an internet connection at all (perhaps they forgot to turn on their wifi adapter).
```
glyph, line 158: In fact the failure of the first connection here is no longer special!  `failAfterFailures` just adds bounds to the behavior of `whenConnected`, which is semantically the same whether it's the first connection or not.

src/twisted/application/internet.py
```diff
@@ -761,31 +772,63 @@ def _cancelConnectWaiters(self):
...
 
     @_machine.input()
-    def whenConnected(self):
+    def whenConnected(self, failAfterFailures=None):
         """
         Retrieve the currently-connected L{Protocol}, or the next one to
         connect.
 
+        @param failAfterFailures: number of connection failures after which
+        the Deferred will deliver a Failure (None means the Deferred will
```
glyph, line 812: Convention is to indent here, as https://github.com/glyph/python-docstring-mode would.

src/twisted/application/internet.py
```diff
@@ -761,31 +772,63 @@ def _cancelConnectWaiters(self):
...
+        very first connection failure signal an error. Use 2 to allow one
+        failure but signal an error if the subsequent retry then fails.
+        @type failAfterFailures: L{int} or None
+
         @return: a Deferred that fires with a protocol produced by the factory
             passed to C{__init__}
-        @rtype: L{Deferred} firing with L{IProtocol} or failing with
-            L{CancelledError} the service is stopped.
+        @rtype: L{Deferred} that may:
+        * fire with L{IProtocol}
```
glyph, line 821: `*` should be `-` if you're trying to do bullets, I think.

glyph (approved):
This actually looks remarkably close to what I thought it should do :). So I think if you address the documentation concern, and ensure that the build is run on all builders, this can land.  Thanks!

warner:
Cool, I've made those changes and pushed to a branch here in the twisted/twisted repo.

Replacing this PR by #772, since I couldn't figure out how to switch this PR to a different branch.

PR #771, opened 2017-04-23
Retry Interaction for MySQL connection issues

jameshilliard:
This is an initial attempt at fixing [this issue](http://twistedmatrix.com/trac/ticket/4964).

This change is based off of this workaround which we are currently using in production:
```python
class ReconnectingConnectionPool(adbapi.ConnectionPool):
    """Reconnecting adbapi connection pool for MySQL.

    This class improves on the solution posted at
    http://www.gelens.org/2008/09/12/reinitializing-twisted-connectionpool/
    by checking exceptions by error code and only disconnecting the current
    connection instead of all of them.

    Also see:
    http://twistedmatrix.com/pipermail/twisted-python/2009-July/020007.html

    (2006: MySQL server has gone away
     2013: Lost connection to MySQL server
     1213: Deadlock found when trying to get lock)
    """
    def _runInteraction(self, interaction, *args, **kw):
        try:
            return adbapi.ConnectionPool._runInteraction(self, interaction, *args, **kw)
        except MySQLdb.OperationalError, e:
            if e[0] not in (2006, 2013, 1213):
                raise
            log.msg("%s got error %s, retrying operation" % (self.__class__.__name__, e))
            conn = self.connections.get(self.threadID())
            self.disconnect(conn)
            # try the interaction again
            return adbapi.ConnectionPool._runInteraction(self, interaction, *args, **kw)
```

### DIFF
--- a/src/twisted/enterprise/adbapi.py
+++ b/src/twisted/enterprise/adbapi.py
@@ -463,18 +463,44 @@ class ConnectionPool:
     def _runInteraction(self, interaction, *args, **kw):
         conn = self.connectionFactory(self)
         trans = self.transactionFactory(self, conn)
-        try:
-            result = interaction(trans, *args, **kw)
-            trans.close()
-            conn.commit()
-            return result
-        except:
-            excType, excValue, excTraceback = sys.exc_info()
+        if self.reconnect and self.dbapiName == 'MySQLdb':
+            from MySQLdb import OperationalError
             try:
-                conn.rollback()
+                result = interaction(trans, *args, **kw)
+                trans.close()
+                conn.commit()
+                return result
+            except OperationalError as e:
+                # 2006: MySQL server has gone away
+                # 2013: Lost connection to MySQL server
+                # 1213: Deadlock found when trying to get lock
+                if e[0] not in (2006, 2013, 1213):
+                    raise
+                log.msg("%s got error %s, retrying operation" % (self.__class__.__name__, e))
+                conn = self.connections.get(self.threadID())
+                self.disconnect(conn)
+                # try the interaction again
+                return self._runInteraction(interaction, *args, **kw)
             except:
-                log.err(None, "Rollback failed")
-            compat.reraise(excValue, excTraceback)
+                excType, excValue, excTraceback = sys.exc_info()
+                try:
+                    conn.rollback()
+                except:
+                    log.err(None, "Rollback failed")
+                compat.reraise(excValue, excTraceback)
+        else:
+            try:
+                result = interaction(trans, *args, **kw)
+                trans.close()
+                conn.commit()
+                return result
+            except:
+                excType, excValue, excTraceback = sys.exc_info()
+                try:
+                    conn.rollback()
+                except:
+                    log.err(None, "Rollback failed")
+                compat.reraise(excValue, excTraceback)
 
 
     def _runQuery(self, trans, *args, **kw):


glyph:
Hi @jameshilliard !  Thanks for submitting this fix.

Is this in a state where you'd like a project member to look at it and review it? If so, please attach the `review` keyword to the corresponding ticket in Trac, and link to this issue.  As you can see, we have many PRs open in various states of completeness, and https://twisted.reviews/ is how we keep track of which ones need attention from core contributors.  (To answer the inevitable follow-up question, yes, one day this will probably be tracked in Github somehow, but for now, that's where it is…)

jameshilliard:
@glyph Yes it should be ready for review, I tested it by modifying the timeout values in the mysql config to reproduce the issue, I couldn't think of a good way to do an automated test for that however. I've tagged the [issue](https://twistedmatrix.com/trac/ticket/4964) with review and linked to this PR.

glyph:
Thanks! An automated test will be a must to get this landed, but perhaps the reviewer can suggest a path forward.

ldanielburr:
Hi @jameshilliard and thanks for tackling this issue.  I've run into it myself and solved by implementing a MySQLTransaction class, implementing an execute() method on that class, and handling the errors there.  This class gets assigned to the transactionFactory attribute of the ConnectionPool, which I think is a bit tidier from a composability standpoint, as opposed to having a conditional that sniffs for the "MySQLdb" driver, and which must be executed every time.

Please note that I'm just providing some feedback on the solution, and if you don't think it is useful, no worries :)

jameshilliard:
@ldanielburr Mind sharing that code?

ldanielburr:
@jameshilliard here's the gist: https://gist.github.com/ldanielburr/2787a24191c21f3b450c9f80287814fc

jameshilliard:
@ldanielburr Issue I see there seems to be the same one I have with optional imports, you are importing `pymysql.err.OperationalError` in order to catch the error so you would still have to have a conditional to sniff the mysql driver so that you don't import it when using non-mysql drivers. AFAICT there's no way to catch a specific database driver error without actually importing the error first.

ldanielburr:
@jameshilliard The problem can occur regardless of the driver being used, so ultimately you need to handle pymysql, MySQLdb, MySQL Connector/Python, etc.  What I was trying to illustrate is that you could handle all of the driver-specific details once, at configuration/service-startup time, rather than via a conditional check in _runInteraction, repeatedly :)

jameshilliard:
@ldanielburr I'm not actually sure this issue exists with postgresql(we use postgresql in a backend process with adbapi for some other projects and I haven't seen a similar error come up there). So you think it's better to do the conditional driver check and import at configuration/service-startup time to load a driver specific transactionFactory? One advantage to doing it in `_runInteraction ` is that I get more granular control and can reset the specific connection thread that had a problem rather than doing a full disconnect and reconnect.

ldanielburr:
@jameshilliard sorry, I'm confused at to where postgresql entered the picture.  I agree that postgresql servers behave differently,

Regarding the driver check: I'm saying that, given that there are multiple python packages out there supporting MySQL (pymysql, MySQL Connector/Python, and MySQLdb), and that each package will have the same problem involving stale/closed connections to the MySQL database server, it might make sense to detect the installed package once, when the connection pool is created, rather than check the driver name each time _runInteraction is called.

Also, can you please clarify what you mean when you say "reset the specific connection thread that had a problem rather than doing a full disconnect and reconnect"?  By doing it in the Transaction subclass, you are accomplishing the same thing, because the Transaction class is being executed in that thread.  Am I misunderstanding something?

Lastly, sorry if this is turning into a long discussion; if you'd prefer to just go ahead with your solution, just say the word and I'll withdraw my suggestion :)

jameshilliard:
@ldanielburr Oh I misread pymysql as psycopg2 for some reason, nvm.

I wasn't sure doing it in the Transaction subclass was operating on the individual thread.

I'm not a fan of my solution either in regards to sniffing the driver, I'd like to come up with a way to generalize the driver detection and exception import so it works with all the mysql drivers automatically.

ldanielburr:
@jameshilliard Hmm, thinking about it some more, how about this? https://gist.github.com/ldanielburr/1cf4466904ff3c6b1919134ab7922b67

It is just a sketch, but it does normalize errors across all the major MySQL database driver packages I can currently find.

jameshilliard:
@ldanielburr Yeah, that's better, although still relies on having an accurate list of mysql drivers, I wonder if it's possible to test for the presence of OperationalError itself in the driver.

ldanielburr:
@jameshilliard I would be leery of testing for the presence of OperationalError, because other database drivers also define an exception class with that name, e.g., psycopg2.OperationalError, so the mere existence of such a class doesn't mean we're dealing with MySQL.

I don't have any good ideas as to how to better detect that the user wants to run MySQL; there is no such introspection capability in DBAPI.  If the powers that be in Twisted-land were to declare that the only abapi-approved driver was, say, pymysql, we could just detect that and call it a day, but that seems kind of draconian, given that any of the drivers I listed work just fine in practice.

jameshilliard:
@ldanielburr Well importing the error if present probably isn't really harmful since we can handle the specific exception codes, we may want to be doing that anyways.

jameshilliard:
This issue doesn't appear to be limited to mysql, I saw it come up with psycopg2cffi recently.

```python
  File "/apppath/pypy35/site-packages/twisted/python/threadpool.py", line 250, in inContext
    result = inContext.theWork()
  File "/apppath/pypy35/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "/apppath/pypy35/site-packages/twisted/python/context.py", line 122, in callWithContext
  File "/apppath/pypy35/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
  File "/apppath/common/dbconnection.py", line 34, in _runInteraction
    conn.rollback()
  File "/apppath/pypy35/site-packages/twisted/enterprise/adbapi.py", line 52, in rollback
    self._connection.rollback()
  File "/apppath/pypy35/site-packages/psycopg2cffi/_impl/connection.py", line 42, in check_closed_
    raise exceptions.InterfaceError('connection already closed')
psycopg2cffi._impl.exceptions.InterfaceError: connection already closed
```

markrwilliams:
Note that the psycopg2_cffi error implies [reconnect was turned off](https://github.com/twisted/twisted/blob/91a72d075db94a78da07d8d601fad090d91d5504/src/twisted/enterprise/adbapi.py#L52)

src/twisted/enterprise/adbapi.py
```diff
@@ -463,18 +463,44 @@ def _close(self, conn):
...
         trans = self.transactionFactory(self, conn)
-        try:
-            result = interaction(trans, *args, **kw)
-            trans.close()
-            conn.commit()
-            return result
-        except:
-            excType, excValue, excTraceback = sys.exc_info()
+        if self.reconnect and self.dbapiName == 'MySQLdb':
+            from MySQLdb import OperationalError
```
markrwilliams, line 467: As you noted in the PR's discussion, it would be effectively impossible to maintain a list of all Python MySQL drivers and what exceptions each raises when a connection has ended but can be re-established.  That's the biggest problem with this approach :frowning_face:

src/twisted/enterprise/adbapi.py
```diff
@@ -463,18 +463,44 @@ def _close(self, conn):
...
-                conn.rollback()
+                result = interaction(trans, *args, **kw)
+                trans.close()
+                conn.commit()
+                return result
+            except OperationalError as e:
+                # 2006: MySQL server has gone away
+                # 2013: Lost connection to MySQL server
+                # 1213: Deadlock found when trying to get lock
+                if e[0] not in (2006, 2013, 1213):
```
markrwilliams, line 477: This is another problem: different MySQL versions might have different error codes that indicate an action could be retried.  It'd be hard to keep this list up to date as well.

markrwilliams (changes requested):
Thanks so much for working on this issue, and for your patience in waiting for a review!

Unfortunately it's not possible to use this particular approach.  I think there's a more general one, though, that would be easier to add to Twisted: make `cp_reconnect` accept a strategy in addition to `True` or `False`.  Besides the corresponding `Reconnect` and `Fail` strategies, there could also be a `CheckException` strategy that took a callable.  This callable could then accept a `Failure` instance created from an exception encountered during `_runInteraction` and return either `True`, which would cause `ConnectionPool` to retry the action, or `False`, which would re-raise the exception.  Here's what that might look like:

```python
def operationalError(interactionFailure):
    interactionFailure.trap(MySQLdb)
    return interactionFailure.value[0] in (2006, 2013, 1213)

pool = ConnectionPool('MySQLdb', db=..., cp_reconnect=CheckException(operationalError)
```

This would move the database specific logic out of Twisted and into application code, without requiring that code to fill out the `ConnectionPool` paperwork necessary to retry the action or terminate the connection.

Do you think that would work for your use case?

markrwilliams:
I'm going to close this PR for now - please feel free to re-open it if you have the time to address [the review](https://github.com/twisted/twisted/pull/771#pullrequestreview-110277645).  Thanks again for your contribution and your patience!

PR #773, opened 2017-04-30
[#9127] http auth session wrapper py3 fix

tardyp:
fix for https://github.com/buildbot/buildbot/issues/3162


tardyp:
@rodrigc fyi a py3 fix. I am not sure why the buildbot tests are not running.

rodrigc (changes requested):
Sometimes these nativeString and networkString functions are a necessary evil, but can we fix this somehow without using them?

Something like:

```
    l.append(k + b"=" + quoteString(v))
```

rodrigc (approved):

PR #774, opened 2017-05-02
Prepare for OpenSSL 1.1.1: Enable TLS 1.3 ciphers

tiran:
TLS 1.3 does not use any of the existing cipher suites from TLS 1.2 and
earlier. A new set of cipher suites was introduced that only specify
bulk encryption, AEAD and PRF algorithms.

I expect that OpenSSL 1.1.1 will enable TLS 1.3 by default. If both
server and client have TLS version 1.3 enabled but have not configured
necessary cipher suites, handshake is going to fail.

Signed-off-by: Christian Heimes <christian@python.org>

### DIFF
--- a/src/twisted/topfiles/9128.feature
+++ b/src/twisted/topfiles/9128.feature
@@ -1,0 +1,1 @@
+twisted.internet._sslverify now enables TLS 1.3 cipher suites.
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1897,6 +1897,8 @@ class OpenSSLAcceptableCiphers(object):
 # - disable NULL authentication, MD5 MACs and DSS for security reasons.
 #
 defaultCiphers = OpenSSLAcceptableCiphers.fromOpenSSLCipherString(
+    "TLS13-AES-256-GCM-SHA384:TLS13-CHACHA20-POLY1305-SHA256:"
+    "TLS13-AES-128-GCM-SHA256:"
     "ECDH+AESGCM:ECDH+CHACHA20:DH+AESGCM:DH+CHACHA20:ECDH+AES256:DH+AES256:"
     "ECDH+AES128:DH+AES:RSA+AESGCM:RSA+AES:"
     "!aNULL:!MD5:!DSS"


Lukasa:
@tiran Thanks for this! By default this needs a [topfile](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles) and a Twisted issue. If you're not up for doing that, I can take this over from you, add those parts, and then bully another committer into reviewing. Otherwise, I'm happy to take the lead on review.

Lukasa (changes requested):
Agh, meant the above to be a review.

lae-quay:
How will we know if this works or not?

exarkun:
How will we know if this works or not?

tiran:
Thanks @Lukasa 

I have created https://twistedmatrix.com/trac/ticket/9128 and added a topfile. Is feature ok? It's technically both a feature and a bugfix for a bug that does not exist yet.

exarkun:
> I have created https://twistedmatrix.com/trac/ticket/9128 and added a topfile. Is feature ok? It's technically both a feature and a bugfix for a bug that does not exist yet.

It's just a feature.

I've added the "review" keyword to the trac ticket.  Please do that in the future whenever you have a PR that you want someone to look at.  Thanks.

Lukasa:
To know whether this works or not right now all we can do is run Twisted against a custom PyOpenSSL that can work with TLSv1.3. I don't even know if that is actually possible yet, though I suppose I should try. Regardless, OpenSSL has currently committed to those names in the cipher suite list, so I think this is a safe bet.

exarkun:
> run Twisted

Can you be more specific?  Twisted is a big pile of library code.  What specific thing can be run?  Is there a way to set up an HTTPS server and query it with an HTTPS client?  Something simpler?

exarkun:
(If it's not clear, the perspective I'm approaching this from is that there's no automated test coverage for this change.)

Lukasa:
>  Is there a way to set up an HTTPS server and query it with an HTTPS client? Something simpler?

In principle, yes, sure: use `twist web` to serve the files with an endpoint string that enables HTTPS, and then point `openssl s_client` against it with the appropriate ciphers enabled and see whether it works.

To *automate* that is moderately difficult, but only moderately: using some of the current interactive TLSMemoryBIO tests would also be absolutely fine for that. The *problem* with it is that to effectively interactively test this would require, again, the custom PyOpenSSL linked against OpenSSL 1.1.1 (as yet unreleased, maybe not even zero effort to build as cryptography may not yet link against it).

exarkun:
> The problem with it is that to effectively interactively test this would require, again, the custom PyOpenSSL linked against OpenSSL 1.1.1 (as yet unreleased, maybe not even zero effort to build as cryptography may not yet link against it).

It sounds like some CI build environment should be set up like this.   Perhaps not *now* (though I don't know why *not* now).  If we want the configuration to work, though, it should really be tested in CI.

rodrigc:
@Lukasa in future, for consistency, point folks to: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

which has all the details about creating a ticket, creating topfile, etc., in one place.

src/twisted/topfiles/9128.feature
```diff
@@ -0,0 +1 @@
+twisted.internet._sslverify now enables TLS 1.3 cipher suites.
```
Lukasa, line 1: So, sorry about this, but generally we try not to reference the complete path of private modules. It'd be better to say something like:

`twisted.protocols.tls.TLSMemoryBIOProtocol and anything that uses it indirectly including the TLS client and server endpoints now enables TLS 1.3 cipher suites.`

Lukasa (changes requested):

Lukasa:
Closed in favour of #787.

PR #775, opened 2017-05-06
[9131] Don't raise AlreadyCancelled for already cancelled timeouts.

adiroiban:
Scope
--------

This fixes the AlreadyCancelled which are raised when the timeout cancellation is called for already cancelled timeouts.

Changes
-----------

Update the code to handle (by ignore) the AlreadyCancelled error.

There were no unit tests for TimeoutProtocol ... so I started a test case and found out that things break apart if the TimeoutProtocol is initialized with None... so I have updated the implementation.

How to test
---------------

I have not managed to reproduce this in production or with manual tests.

I am getting this error with my test suite where for full end to end tests I have a tearDown which will make sure that at the end of the tests the reactor is clean... similar to what there is in ReactorBuilder.unbuildReactor

To simplify my end-to-end tests I have this  threaded reactor and custom test cases https://github.com/chevah/compat/blob/master/chevah/compat/testing/threaded_reactor.py#L15  so that the reactor runs in the background (non blocking) an I can do tests using some blocking (ro)bots / clients

I am only getting the error for TimeoutMixin ...but since I was working on this, I also updated the code for the protocol and wrote the tests.


glyph:
Oops, it's been a while.  This needs `fix-for-towncrier.py` run on it 😕.

glyph (approved):
This looks correct to me.

adiroiban:
Thanks for the review... this was hard :)

The twisted dev process is changing and I might did a bad merge.

Please double check my merge.

glyph:
Thanks for asking for the double-checking! The merge looks good to me; right number of parents, right "handedness" (the left-hand commit is the previous master), commit message closed the right ticket… it doesn't have the "long description" but that's totally OK, that's optional.

PR #776, opened 2017-05-07
[#9126]  Split out twisted.python.url into separate package (hyperlink)

mahmoud:
(From https://twistedmatrix.com/trac/ticket/9126)

The topfile summarizes the enhancements:

> twisted.python.url and twisted.python._url were modified to rely on hyperlink, a new package based on the Twisted URL implementation. Hyperlink adds support for IPv6 (fixing #8069), correct username/password encoding, better scheme/netloc inference, improved URL.click() behavior (fixing #8184), and more. For full docs see hyperlink.readthedocs.io and the CHANGELOG in the hyperlink GitHub repo.

And the commit message the strategy:

> clear out _url and import everything from hyperlink>=17.1.1, which was added to the setup.py install requirements. all tests passing. leaving the tests in place for the moment to ensure that all Twisted APIs continue to function. slight repr changes were updated in the tests (old reprs continue to function.

Code review away! Happy to make changes and answer any questions. :)


src/twisted/python/url.py
```diff
@@ -8,7 +8,7 @@
 @see: L{URL}
 """
 
-from ._url import URL
+from hyperlink import URL
```
markrwilliams, line 11: `twisted.python.url` should be [deprecated](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#how-to-deprecate-apis).  Since it's a submodule, you can use [`twisted.python.deprecate.deprecatedModuleAttribute`](https://twistedmatrix.com/documents/current/api/twisted.python.deprecate.deprecatedModuleAttribute.html).

You can read through the @hawkowl's `constantly` [PR](https://github.com/twisted/twisted/pull/498/files#diff-c7a7c545fe518e01b7e63566b82de9f3) to see how this works in practice.   Don't forget to test the deprecation!

Note that the documentation is a little behind the times.   Import `Version` from `incremental`, and write it as `Version("Twisted", "NEXT", 0, 0)`.  `incremental` will then automatically rewrite `"NEXT"` to the version that will then be released.  This ensures your code doesn't have to change if it misses the next release.

src/twisted/topfiles/9126.feature
```diff
@@ -0,0 +1 @@
+twisted.python.url and twisted.python._url were modified to rely on hyperlink, a new package based on the Twisted URL implementation. Hyperlink adds support for IPv6 (fixing #8069), correct username/password encoding, better scheme/netloc inference, improved URL.click() behavior (fixing #8184), and more. For full docs see hyperlink.readthedocs.io and the CHANGELOG in the hyperlink GitHub repo.
```
markrwilliams: You should break each bug fixed into its own topfile.   Whoever merges this should put all the tickets into the `Fixes` line, as @rodrigc [did](https://github.com/twisted/twisted/commit/e116bbe34dc3a3855a674043bc5f30d7e6417c7f) in #630.  Something like:

```
Fixes: ticket:8069 , ticket:8184 , ticket:9126
```

It'd be great if we could figure out _all_ the tickets this closes, but we can start with the ones you've identified.

markrwilliams (changes requested):
I'm always happy to see Twisted become usable in more places.  Sometimes the best way to do that is to split it up into new packages!

Do we want `hyperlink` to live under the Twisted organization?  Even if we do that doesn't appear to be a prerequisite to merging this PR because neither Automat nor incremental are Twisted projects.

src/twisted/topfiles/9126.feature
mahmoud: I did a search for twisted.python.url in the Trac and these were the only ones open. I could open one on the username/password encoding just to close it, I suppose.

mahmoud:
Re: organizational ownership, like I said in the Trac, I think I'm fine with whichever. I expect the code to be pretty stable. Aside from a few minor items in the TODO, I don't see major features/changes coming down the line that would require a wide contributor base, for instance.

mahmoud:
Changes made, rereview at your leisure @markrwilliams

src/twisted/python/test/test_url.py
```diff
@@ -5,12 +5,18 @@
...
-from ..url import URL
+import warnings
+
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore", category=DeprecationWarning)
+    from ..url import URL
+
 unicode = type(u'')
 from unittest import TestCase
+from twisted import trial
```
mithrandi: This import doesn't ensure that `twisted.trial.unittest` is imported, so the `unittest` attribute of the `twisted.trial` module might not be there when you want to use it. I would instead write something like `from twisted.trial.unittest import TestCase as TwistedTestCase`.

mithrandi: Actually just `from twisted.trial.unittest import SynchronousTestCase`, as glyph points out; we don't need any `Deferred` handling here.

mahmoud:
@mithrandi I changed the test case inheritance, per your review. Does this need to go through the buildbots again? (also i think the os x buildbot is just dead at this point)

src/twisted/python/test/test_url.py
```diff
@@ -793,6 +799,25 @@ def test_technicallyTextIsIterableBut(self):
             URL(path=u'foo')
         self.assertEqual(
             str(raised.exception),
-            "expected iterable of text for path, got text itself: {}"
+            "expected iterable of text for path, not: {}"
             .format(repr(u'foo'))
         )
+
+
+class UrlDeprecationTests(SynchronousTestCase):
```
markrwilliams: URL, not Url

src/twisted/python/test/test_url.py
```diff
@@ -793,6 +799,25 @@ def test_technicallyTextIsIterableBut(self):
...
+    """
+    L{twisted.python.constants} is deprecated.
+    """
+    def test_urlDeprecation(self):
+        """
+        L{twisted.python.constants} is deprecated since Twisted 16.5.
```
markrwilliams: ~16.5~ NEXT

markrwilliams (commented):
Two little changes, then I'll push to `twisted/twisted`.  If it's all green then I think this is ready to merge.

mithrandi:
The buildbots only build branches in the `twisted/twisted` repo; I just pushed this up there (if you have commit access, note the `admin/pr_as_branch` script which automates this process) so we can get a build (the build results will reflect here since they're commit statuses, not branch statuses, and the commit is the same one as on this PR)

glyph:
Have @markrwilliams' changes been dealt with?

mahmoud:
@glyph, I believe so!

src/twisted/topfiles/8069.bugfix
```diff
@@ -0,0 +1 @@
+The transition to the hyperlink package adds IPv6 support to twisted.python.url.URL.
```
adiroiban: I don't see the point of advertising something about a deprecated functionality.

src/twisted/topfiles/9126.feature
adiroiban: I don't think that this should be a .feature.

I think that this should be a .removal fragment informing that twisted.python.url is deprecated and people should use new Hyperlink package.

adiroiban (changes requested):
Changes looks good.

This should be a .removal ticket as what it really does is deprecate twisted.python.url.

-------

I don't see the change from src/twisted/python/urlpath.py

at least in trunk it has `from twisted.python.url import URL as _URL` ... we should update all existing imports to directly import from Hyperlink.

You can find me on IRC and I can help move this into the Twisted organization

markrwilliams (approved):
My review comments have been addressed.  I'll take over the PR and address @adiroiban's concerns.

markrwilliams:
@adiroiban The unit tests should still import `URL` from `twisted.python.url` until the code is removed, right?  That way at least we get coverage on the change over to it.

src/twisted/topfiles/8069.bugfix
markrwilliams: It's deprecated but not removed.  That means interested users get a fix for [8069](http://twistedmatrix.com/trac/ticket/8069) even if they don't import `URL` from `hyperlink`.  I think this description should mention the deprecation ticket, though.

adiroiban:
> @adiroiban The unit tests should still import URL from twisted.python.url until the code is removed,

Yes. The units test should import from twisted.python.url but only for the purpose of testing the deprecation warning. 

All the other test can be removed now,

---------------

For other deprecation tickets, when we were scheduling the removal of an API, without planing any drop-in replacement, we would leave the tests until the deprecation deadline is hit.

In this way, if there are bugs (critical) in the  code which is to be deprecated we could still fix it and have tests

---------

Here is a different story as in this PR we are removing all the code from src/twisted/python/_url.py ... so even if we import twisted.python.url ... all the other non-deprecation tests will now be for the hyperlink code and not for the code from src/twisted/python/_url.py

src/twisted/python/_url.py
```diff
@@ -7,787 +7,7 @@
...
-        removed.
-
-        @param name: The name of the query parameter to remove.
-        @type name: L{unicode}
-
-        @return: a new L{URL} with the parameter removed.
-        """
-        return self.replace(query=((k, v) for (k, v) in self.query
-                                   if k != name))
+__all__ = ["URL"]
```
adiroiban, line 13: maybe we should add a comment here... in case someone will somehow direclty react this code and will don't know what happens... the comment  should describe that this is deprecated and hyperlink is used instead.

src/twisted/python/_url.py
```diff
@@ -7,787 +7,7 @@
...
 """
 
-try:
-    from urlparse import urlsplit, urlunsplit
-    from urllib import quote as urlquote, unquote as urlunquote
-except ImportError:
-    from urllib.parse import (urlsplit, urlunsplit,
-                              quote as urlquote,
-                              unquote_to_bytes as urlunquote)
+from hyperlink._url import URL
```
adiroiban, line 10: Why are we importing from a private API ?

I think that hyperlink should make a vow and expose a public API :)

src/twisted/python/__init__.py
```diff
@@ -17,6 +17,13 @@
     "Please use constantly from PyPI instead.",
     "twisted.python", "constants")
 
+
+deprecatedModuleAttribute(
+    Version("Twisted", "NEXT", 0, 0),
```
adiroiban, line 22: Just a side comment: is this NEXT thing documented somewhere?

I think that is important to have a docs/core/development/policy/compatibility-policy.rst talk about this.

And with this, maybe we can rename it to `~NEXT-RELEASE~` or something to make it clear that is a placeholder.

Also... why not  just `Version("Twisted", "NEXT")` why do we need the other zeros there?

adiroiban (changes requested):
still need changes

src/twisted/python/_url.py
glyph: Hyperlink already has, this should be hyperlink.URL as documented on http://hyperlink.readthedocs.io/en/latest/api.html

src/twisted/python/_url.py
glyph: Doesn't the deprecation message already say that?

src/twisted/python/__init__.py
markrwilliams: > Just a side comment: is this NEXT thing documented somewhere?

Yup! [Right here](https://github.com/hawkowl/incremental#updating)  It may be possible to change this (though I don't, for the record), but that change would be in incremental.

glyph: It's documented in Incremental's README: https://github.com/hawkowl/incremental.

If you want to change it (almost certainly not going to happen) you should file a bug there.

(Are we going to adopt incremental into the twisted org, btw @hawkowl ?  I don't love all these factored-out packages living in peoples' personal repos...)

adiroiban: I was talking about our current documentation in docs/core/development/policy/compatibility-policy.rst

src/twisted/python/test/test_url.py
```diff
@@ -5,12 +5,18 @@
...
 """
-
 from __future__ import unicode_literals
 
-from ..url import URL
+import warnings
+
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore", category=DeprecationWarning)
+    from ..url import URL
```
adiroiban: why the relative import?

src/twisted/python/test/test_url.py
```diff
@@ -793,6 +799,25 @@ def test_technicallyTextIsIterableBut(self):
...
+            "expected iterable of text for path, not: {}"
             .format(repr(u'foo'))
         )
+
+
+class URLDeprecationTests(SynchronousTestCase):
+    """
+    L{twisted.python.constants} is deprecated.
+    """
+    def test_urlDeprecation(self):
```
adiroiban, line 807: Is this test working? I am trying to understand why ? since the URL should be already imported earlier

src/twisted/python/test/test_url.py
markrwilliams: The previous import was also relative.  This change's real intent is to suppress the `DeprecationWarning`.

src/twisted/python/test/test_url.py
```diff
@@ -178,7 +184,7 @@ def test_repr(self):
             repr(URL(scheme=u'http', host=u'foo', path=[u'bar'],
                      query=[(u'baz', None), (u'k', u'v')],
                      fragment=u'frob')),
-            "URL.fromText(%s)" % (repr(u"http://foo/bar?baz&k=v#frob"),)
+            "URL.from_text(%s)" % (repr(u"http://foo/bar?baz&k=v#frob"),)
```
adiroiban, line 183: this should not be changed... as it breaks the API

src/twisted/python/test/test_url.py
```diff
@@ -793,6 +799,25 @@ def test_technicallyTextIsIterableBut(self):
             URL(path=u'foo')
         self.assertEqual(
             str(raised.exception),
-            "expected iterable of text for path, got text itself: {}"
+            "expected iterable of text for path, not: {}"
```
adiroiban, line 798: I am ok with this exception... no need to be that strict on repr... at least for now :)

src/twisted/python/test/test_url.py
markrwilliams: OK @adiroiban pointed out in `#twisted-dev` that `deprecatedModuleAttribute` is _supposed_ to emit a deprecation warning once and only once, when the module's attribute is used.  You would think this would trigger the deprecation warning and break [the deprecation test](https://github.com/twisted/twisted/pull/776/files#diff-19a383d67f518cb991c253b231d176eeR807). However, it doesn't!  Neither of these do:

```python
from ..url import URL
from twisted.python.url import URL
```

But this does:

```python
from twisted.python import url
URL = url.URL
```

Which leads me to believe that the import system's attribute access mechanism isn't just `getattr` (or perhaps [`t.p.deprecate`'s heuristic is wrong?](https://github.com/twisted/twisted/blob/0aef33642cbe04cac558a4aaceeb2a6a03de403b/src/twisted/python/deprecate.py#L455-L462))

I'm going to remove the suppression because it's actually useless!

adiroiban:
This still needs-changes due to the API change... but I think that otherwise (as long as the tests are green) this is a good one.

Regarding all the discussion regarding the deprecation I hope that we can include that info here https://twistedmatrix.com/trac/ticket/8478

markrwilliams:
@adiroiban The API _is_ compatible - only the `__repr__` differs.

src/twisted/python/test/test_url.py
markrwilliams: This is just the `__repr__` - [`URL.fromText`](https://github.com/mahmoud/hyperlink/blob/4f56ee8b7fcc3e6ca4ba13dbb8b55368cfca0433/hyperlink/_url.py#L1098-L1100) still exists.

adiroiban: true :) my bad

adiroiban:
This is good to go. Thanks! 

Looking forward for having the code moved to Twisted org.

The fact that you need to copy this branch inside twisted ... and then run all the tests in unfortunate , but required. ... not sure how to improve this... other than allow any PR (from any source) to trigger tests based on a whitelist

markrwilliams:
@adiroiban Thanks!  I'll take [your comment](https://github.com/twisted/twisted/pull/776#issuecomment-302766842) as an approving review.

adiroiban:
yep... sorry. I am still confused between Trac and GitHub approval messages. 
AFAIK, Trac is the official channel and I think that I marked this PR as reviewed and approved there,

PR #778, opened 2017-05-12
Prevent a "exceptions.ValueError: need more than 1 value to unpack" t…

itshane:
…raceback when the syntax of the Set-Cookie header is wrong.

Fixes https://twistedmatrix.com/trac/ticket/9136

When the Set-Cookie header contains ['HttpOnly;Secure'] value, it cannot split the value appropriately and throws a ValueError traceback.
The change just omits the wrong part of the header, therefore, a response can be gracefully processed.


Lukasa:
Thanks for this, this looks like a really good approach to this issue. However, it needs some enhancements per our [review process](https://twistedmatrix.com/trac/wiki/ReviewProcess) before we can merge:

1. You need to open a ticket on twistedmatrix.com to track the work. This ticket should explain the issue.
2. You'll need to write a unit test to demonstrate the issue and to demonstrate that your patch fixes the issue.
3. You'll need a topfile explaining the patch.

If you need any help with any of this, just let me know.

itshane:
@Lukasa: I've amended the pull request with the requested changes, please take a look when possible.

src/twisted/web/client.py
```diff
@@ -411,14 +411,21 @@ def _cancelTimeout(self, result, timeoutCall):
         return result
 
     def gotHeaders(self, headers):
+        """
+        Parse the response HTTP headers.
+
+        @param headers: The response HTTP headers.
+        @type headers: L{list}
```
Lukasa: `headers` cannot possibly be a `list`: it has to be string-indexable. The rest of the code suggests it's a `dict`.

Lukasa (changes requested):
Ok, I've left a note about some docs stuff. When that gets sorted out I'll push to the builders. You'll want to re-submit your ticket for review.

itshane:
@Lukasa: Updated.

rodrigc (approved):
OK, @Lukasa's feedback was incorporated.

PR #779, opened 2017-05-13
Ensure HTTP path is of type bytes to avoid 404 errors in py3

lthibault:
Under python 3, running `Resource.putChild` with a `path` argument that is not of type `bytes` results in a 404 when visiting the link.  Note that absurd types are also silently accepted (e.g.:  `some_resource.putChild(42, some_other_resource)`.

This patch does a simple type-check and throws a `TypeError` on failure. 

Trac ticket [here](https://twistedmatrix.com/trac/ticket/9135#ticket).

### DIFF
--- a/src/twisted/web/resource.py
+++ b/src/twisted/web/resource.py
@@ -216,6 +216,10 @@ class Resource:
 
         @see: L{IResource.putChild}
         """
+        if not isinstance(path, bytes):
+            emsg = "HTTP paths must be of type bytes (got {0})"
+            raise TypeError(emsg.format(type(path)))
+
         self.children[path] = child
         child.server = self.server
 


mithrandi (changes requested):
Thank you for the contribution!

Unfortunately I think making this raise a hard exception would be a backwards incompatible change as per the [compatibility policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html); instead we would need to make this emit a deprecation warning first, which could later become an exception. Also, the new code path is missing [test coverage](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#testing-deprecation-code) as codecov will hopefully report once the Travis CI run has completed.

lthibault:
Hello,

Why would this be backwards-incompatible?  Under Python 2.7.12, at least, `bytes` is an alias to `string`.

Surely there’s no _working_ code that’s passing non-byte objects to `path`, is there?

>Also, the new code path is missing test coverage <https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#testing-deprecation-code> as codecov will hopefully report once the Travis CI run has completed.

Yikes, I hadn’t thought of that.  Should be fairly trivial to add, though :)

mithrandi:
Any code that currently calls `putChild` with something other than bytes will, essentially, do nothing at the moment. While presumably that was not the author's original intent when writing the code, if some existing code happens to do this (quite possibly by accident), having it change from doing nothing to raising an exception would unfortunately make the code stop working.

lthibault:
Can’t it be argued that this is a feature and not a bug?  i.e.:  a silent bug is made explicit.

I don’t mean to be adversarial, and I completely respect your call.  It just so happens that I would have (probably) made a different one, and so I wanted to be sure to express my thoughts :)

Cheers,
Louis

rodrigc:
(1)  You need to add a test to cover the new code path.

(2)  Probably you should add to the docstring for putChild to indicate that the argument needs to take bytes only

(3)  This change breaks some existing tests: https://travis-ci.org/twisted/twisted/jobs/231939388

mithrandi:
@lthibault The compatibility policy has some [exceptions](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#bug-fixes-and-gross-violation-of-specifications) for such cases, but I think this case doesn't fall under that.

It might be useful to contrast this issue with a prior issue where calling `setHeader()` with unicode accidentally caused UTF-16 or UTF-32 encoded data to be written out (due to the implementation details of the buffer interface on unicode objects at the time), normally resulting in a bunch of NULs unexpectedly truncating the HTTP request. While changing this to raise an exception instead was also incompatible, writing a bunch of NULs in the HTTP header was clearly a gross specification violation, having a negative impact on the rest of the functionality in the application.

However, there is an ["escape hatch"](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#procedure-for-exceptions-to-this-policy) for exceptions to the policy; if you think this issue warrants an exception, then feel free to follow that process (ie. post to the mailing list). I don't think getting some extra opinions involved here would hurt, and I'm certainly not the Final Boss of Compatibility or anything like that :laughing: 

My reasoning for going the normal deprecation route rather than the exception route is that most of the instances of the bug exposed by this exception are likely to either be brand new code (too bad that we're not fixing it immediately, but we're still getting there), or long-standing code where the bug isn't fatal for the whole application, while throwing an exception would be. Consider this strawman: if someone ported to Python 3 and forgot to change one call like `root.putChild('debug', DebugInfoResource())`, then their debug resource would now be mysteriously missing. This is no doubt something they would appreciate knowing about and fixing, but I don't think we need to break their whole application to do it, considering that they've managed without that resource so far.

Regarding the test failures, it looks like the failing test in question is (somewhat harmlessly) broken and should just be fixed: it passes a "native" (`''`) string in a `getChild` call which should be a byte string, but of course becomes a unicode string on Python 3.

rodrigc:
I've done a major porting job of code in buildbot which uses part of Twisted web.
The patch that @lthibault submitted would have helped a lot find places where unicode was passed to putChild() where bytes should have been passed instead.
In fact, as I was porting, I put my own temporary error in the putChild() code to catch all these places in buildbot.

I think this patch is on the right track, but it can be implemented better,
and it needs test coverage, so that we can verify if we pass in the wrong type, the correct exception is thrown.

wiml:
This would have saved me quite a bit of debugging time as well.

I've created a branch off this PR that converts it into a DeprecationWarning and adds a unit test, on the theory that the deprecation can be changed back to lthibault's TypeError after a sufficient delay.

https://github.com/twisted/twisted/tree/9135-typecheck-url-path

Another alternative would be for it to be a FutureWarning on python2 and a TypeError on python3. 

Thoughts?

rodrigc:
@wiml if you change your patch to emit DeprecationWarning on py2, and throw TypeError on py3, I'll  approve and merge it.  This would have saved me many hours when I was trying to get buildbot to work on Python 3.

PR #781, opened 2017-05-15
[#5739] Document the contentType member of t.w.static.File

adiroiban:
Scope
--------

This adds documentation for the contentType (and contentEncoding) instance members of the File and DirectoryListing.

Changes
-----------

Add API doc together with an example in the narrative documentation.

I have never used the contentEncoding so I don't know a practical example for it.... maybe if you have an old `SOMETHING.log.compress` file... but I hope that by giving an example for contentType, users will know how to use the contentEncoding

How to test
---------------

Check that changes make sense.

You should build both the API docs and narrative docs and check that all is rendered OK.
 ... I am still reading the dev docs to see how to build it


wsanchez (approved):

wsanchez:
Reviewed:
 * You are missing a topfile.
 * There are some pyflakes3 errors reported as well, but those looks like false positives,given that they are complaining about code that was not altered.

PR #782, opened 2017-05-18

rodrigc:


src/twisted/python/test/test_release.py
```diff
@@ -861,7 +861,7 @@ def test_statusNotClean(self):
         no pending modifications returns C{False}.
         """
         reposDir = self.makeRepository(self.tmpDir)
-        reposDir.child('some-file').setContent("something")
+        reposDir.child('some-file').setContent(b"something")
```
acabhishek942, line 864: is it not required to make `'some-file'` as bytes?

src/twisted/python/test/test_release.py
```diff
@@ -872,7 +872,7 @@ def test_remove(self):
         """
         reposDir = self.makeRepository(self.tmpDir)
         testFile = reposDir.child('some-file')
```
acabhishek942, line 874: same here

src/twisted/python/_release.py
```diff
@@ -455,16 +456,16 @@ def _writeSection(self, fileObj, header, tickets):
             reverse[description].sort()
         reverse = reverse.items()
         # result is a tuple of (descr, tickets)
-        reverse.sort(key=lambda result: result[1][0])
+        reverse = sorted(reverse, key=lambda result: result[1][0])
```
acabhishek942: why is this change necessary?

src/twisted/python/_release.py
```diff
@@ -455,16 +456,16 @@ def _writeSection(self, fileObj, header, tickets):
...
 
-        fileObj.write(header + '\n' + '-' * len(header) + '\n')
+        fileObj.write((header + u'\n' + u'-' * len(header) + u'\n').encode("utf-8"))
         for (description, relatedTickets) in reverse:
             ticketList = ', '.join([
                 '#' + str(ticket) for ticket in relatedTickets])
             entry = ' - %s (%s)' % (description, ticketList)
```
acabhishek942: I think it would be better to use `format` method here and in other applicable places to keep the code consistent

src/twisted/python/test/test_release.py
rodrigc: No it is not required to make ```'some-file'``` bytes.  Only the argument for ```setContent()``` must be bytes.

src/twisted/python/_release.py
rodrigc: This code has been deleted, so this change is no longer there.

acabhishek942:
Please update the patch. Automated tests are not hitting all the lines

rodrigc:
Coverage is not reaching all the lines because sphinx is not installed on the builders where coverage is statistics are uploaded to codecov.  I mentioned the same thing here: https://github.com/twisted/twisted/pull/713#discussion_r104746647

acabhishek942:
> Coverage is not reaching all the lines because sphinx is not installed on the builders where coverage is statistics are uploaded to codecov. I mentioned the same thing here: #713 (comment)

In this case the patch looks good to me. Please merge once all the builds are green

PR #783, opened 2017-05-19
9139 the0id handling unsupported keys

the0id:
Fixing a bug where Twisted fails when a client tries to authenticate with an unknown key type.
Ticket: https://twistedmatrix.com/trac/ticket/9139


src/twisted/conch/ssh/userauth.py
```diff
@@ -263,21 +263,25 @@ def auth_publickey(self, packet):
...
-            c = credentials.SSHPrivateKey(self.user, algName, blob, b,
-                    signature)
-            return self.portal.login(c, None, interfaces.IConchUser)
-        else:
-            c = credentials.SSHPrivateKey(self.user, algName, blob, None, None)
-            return self.portal.login(c, None,
-                    interfaces.IConchUser).addErrback(self._ebCheckKey,
-                            packet[1:])
 
+        try:
```
hawkowl: Can we make this try block cover just where keys.BadKeyError can come from? Or can it happen in multiple places here?

src/twisted/conch/ssh/userauth.py
```diff
@@ -263,21 +263,25 @@ def auth_publickey(self, packet):
...
+                        signature)
+                return self.portal.login(c, None, interfaces.IConchUser)
+            else:
+                c = credentials.SSHPrivateKey(self.user, algName, blob, None, None)
+                return self.portal.login(c, None,
+                        interfaces.IConchUser).addErrback(self._ebCheckKey,
+                                packet[1:])
+        except keys.BadKeyError:
+            log.msg("Unsupported key type: %s", algName)
+            return defer.fail(UnauthorizedLogin)
```
hawkowl: This needs to be an instance of UnauthorizedLogin, not just the class (for Python 3)

src/twisted/topfiles/9139.bugfix
```diff
@@ -0,0 +1 @@
+A patch to handle when a client provides an unspported authentication key type.
```
hawkowl: This should be in the form of "<fqpn> now <x>" or so, such as "twisted.conch.ssh.userauth.SSHUserAuthServer now gracefully handles unsupported authentication key types."

hawkowl (dismissed):

src/twisted/conch/ssh/userauth.py
the0id: Thanks, I think this fixed the testing error that I didn't understand. :+1:

src/twisted/conch/ssh/userauth.py
the0id: It only happens when keys.Key.fromString(blob) is called.  I made the try/except cover the rest just in case something else went wrong, but I have no problem covering just that line.

the0id:
@hawkowl , could you help me out here?

If I'm reading the Travis error right it's saying I don't have a topfile.  What am I doing wrong?

markrwilliams:
@the0id Surprise!  I merged @hawkowl's [PR](https://github.com/twisted/twisted/pull/763) that moved Twisted over to [towncrier](https://github.com/hawkowl/towncrier).  This means we now put our news snippets into `newsfragments/` instead of `topfiles/`.  There should be no other changes.

Fortunately @hawkowl is awesome and put a very helpful script in [`admin/bin`](https://github.com/twisted/twisted/blob/trunk/admin/fix-for-towncrier.py) that recursively renames any and all `topfiles/` to `newsfragments/` and stages them for commit.  Run this, `git commit -m 'topfiles -> newsfragments'` or something, push, and Travis should be happy.

Maybe I should send this information out to the mailing list?

the0id:
@markrwilliams thanks for clearing that up.

Updating everyone on the mailing list is a good idea, but I'll be honest with you. I would have ignored it and still not known what to do. :)

PR #785, opened 2017-05-20
[#9141] Deprecate twisted.protocols.dict

hawkowl:
https://twistedmatrix.com/trac/ticket/9141


markrwilliams (approved):
@hawkowl correctly pointed out that this doesn't have meaningful tests.  If somebody wants to keep this, at the _very least_ it would need a lot of test.  Given that nobody wants to write those this is ready to land.

src/twisted/python/test/test_url.py
```diff
@@ -802,11 +802,11 @@ def test_technicallyTextIsIterableBut(self):
 
 class URLDeprecationTests(SynchronousTestCase):
     """
-    L{twisted.python.constants} is deprecated.
+    L{twisted.python.url} is deprecated.
```
glyph, line 805: What is this change doing in this PR?

markrwilliams: I'm guessing @hawkowl couldn't help herself and cleaned up a mistake I introduced.

(I'm OK with this being in this PR)

hawkowl: yes

rodrigc (approved):

PR #786, opened 2017-05-22
[#9143] Fix logger in httpfactory on py3

hawkowl:
https://twistedmatrix.com/trac/ticket/9143

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ dist/
 venv/
 htmlcov/
 .coverage*
+coverage.xml
 *~
 *.lock
 apidocs/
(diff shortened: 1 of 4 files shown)


rodrigc (approved):

.gitignore
```diff
@@ -13,6 +13,7 @@ dist/
 venv/
 htmlcov/
 .coverage*
+coverage.xml
```
markrwilliams, line 16: Why do we have to add this?  Gunk left over by the `codecov` command's invoking `coverage xml`?

hawkowl:
@rodrigc it's also wrong, I don't think I put it up for review yet :)

PR #787, opened 2017-05-22
Prepare for OpenSSL 1.1.1: Enable TLS 1.3 ciphers

Lukasa:
See [this issue](https://twistedmatrix.com/trac/ticket/9128). Supersedes #774.


markrwilliams (approved):
@exarkun Cory pointed out that OpenSSL ignores cipher suites it doesn't recognize. If OpenSSL changes the names of TLS 1.3's cipher suites, this PR will have no effect.  If OpenSSL changes what those names _mean_, then it certainly won't ignore them and this PR would cause a regression. If OpenSSL makes no changes, then this PR is sufficient to allow Twisted to speak TLS 1.3, though it will not take advantage of any features other than the new cipher suites.

I think the adversarial case, where OpenSSL changes what `TLS13-AES-256-GCM-SHA384` means, is unlikely enough that the outcomes reduce to a null change or a positive change.  I'm going to merge this.

PR #789, opened 2017-05-22
[WIP] I believe that IPv4-only systems are still a real thing.

Lukasa:
See https://twistedmatrix.com/trac/ticket/9144


src/twisted/conch/test/test_conch.py
```diff
@@ -49,6 +49,25 @@
...
+def _has_ipv6(host):
+    """ Returns True if the system can bind an IPv6 address."""
+    sock = None
+    has_ipv6 = False
+
+    try:
+        sock = socket(AF_INET6)
+        sock.bind((host, 0))
+        has_ipv6 = True
+    except Exception:
```
alex: This should catch a specific exception, not everything.

src/twisted/conch/test/test_conch.py
```diff
@@ -678,6 +701,8 @@ def test_localToRemoteForwardingV6(self):
         """
         Forwarding of arbitrary IPv6 TCP connections via SSH.
         """
+        if not HAS_IPV6:
+            raise unittest.SkipTest("Requires IPv6 support")
```
alex: I believe using teh `.skip` attribute is the common scheme.

src/twisted/conch/test/test_conch.py
```diff
@@ -49,6 +49,25 @@
...
+        has_ipv6 = True
+    except Exception:
+        pass
+
+    if sock:
+        sock.close()
+    return has_ipv6
+
+
+HAS_IPV6 = _has_ipv6('::1')
```
alex: Inline `::1` into the function, taht's the only way it's ever called.

src/twisted/internet/test/test_tcp.py
```diff
@@ -57,11 +57,14 @@
     useSSL = True
 
 try:
-    socket.socket(socket.AF_INET6, socket.SOCK_STREAM).close()
+    s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+    s.bind(('::1', 0))
 except socket.error as e:
     ipv6Skip = str(e)
+    s.close()
```
alex: `s` is not guarnateed to exist here.

alex (changes requested):

src/twisted/internet/test/test_tcp.py
```diff
@@ -56,12 +56,16 @@
...
 try:
-    socket.socket(socket.AF_INET6, socket.SOCK_STREAM).close()
+    s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+    s.bind(('::1', 0))
 except socket.error as e:
     ipv6Skip = str(e)
 else:
     ipv6Skip = None
+if s is not None:
+    s.close()
```
alex, line 68: Deduplicate this code with the other one. We only need one implementation of IPV6 checking.

src/twisted/internet/test/test_udp.py
```diff
@@ -31,6 +32,36 @@
...
+    return has_ipv6
+
+
+HAS_IPV6 = _has_ipv6()
+
+
+def skipWithoutIPv6(f):
+    @functools.wraps(f)
+    def inner(*args, **kwargs):
+        if not HAS_IPV6:
```
alex: Convention appears to be to assign to `skip` on `f` if it should be skipped.

alex (changes requested):

src/twisted/internet/test/test_tcp.py
Lukasa: We're going to not do that this time around because we'd rather add this check as productized code, so @markrwilliams is going to do that while we merge this to unblock you.

alex (approved):

PR #790, opened 2017-05-22
[#9143] twisted.web.server.Server, Next-Generation Site

hawkowl:
https://twistedmatrix.com/trac/ticket/9143


src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+    _requestFactory = attr.ib()
+    _sessionFactory = attr.ib()
+    _logger = attr.ib()
+    _reactor = attr.ib()
+
+
+    def buildProtocol(self, addr):
+
+        @attr.s
+        class _TrashGarbageSiteDontUse(object):
```
adiroiban, line 67: I think that this needs a docstring

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+        p.callLater = self._reactor.callLater
+        p.timeOut = self._timeout
+        p.requestFactory = self._requestFactory
+        p.site = _TrashGarbageSiteDontUse()
+        return p
+
+
+    def doStart(self):
+        """
+        Start.
```
adiroiban, line 119: docstring not very helpful

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+    _timeout = attr.ib()
+    _resource = attr.ib()
+    _requestFactory = attr.ib()
+    _sessionFactory = attr.ib()
+    _logger = attr.ib()
+    _reactor = attr.ib()
+
+
+    def buildProtocol(self, addr):
+
```
adiroiban, line 65: this also needs docstring

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def doStop(self):
+        """
+        Stop.
+        """
+
+
+    def log(self, request):
+
```
adiroiban, line 130: missing docstring

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+            uri=_escape(request.uri),
+            protocol=_escape(request.clientproto),
+            code=request.code,
+            length=request.sentLength or u"-",
+            referrer=referrer,
+            agent=agent,
+        )
+
+
+    # IProtocolNegotiationFactory
```
adiroiban, line 153: why this comment?

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+    # Set up the Logger source correctly
+    logger.source = server
+    return server
+
+
+
+def makeCombinedLogFormatFileForServer(server, logFile,
+                                       logPublisher=globalLogPublisher):
+    """
+    Make a Combined Log Format file, logging from C{server}.
```
adiroiban, line 234: This line does not add any new information, as is just the English version of makeCombinedLogFormatFileForServer :)

adiroiban (changes requested):
Thanks for working on this :)

---------

For me "Next-Generation Web Sites for the Internet Zone Age." make no sense, but if this new API is so great I think that it should be promoted by creating an example in the narative documentation.

I have created a site object only once for my project and not touched it since then.

It would be nice to see a diff in the current narrative documentation to understand how this API can make the life easier :)


Otherwise, it will be hidden inside the Twisted API and people new to Twisted will continue to read the docs and use the "old way"

-----

I think that this needs an update with trunk. This should trigger the twistedchecker errors, at least for the missing docstrings.

and there is a line changed in src/twisted/web/tap.py which does not have coverage, and this was observed by codecov :)

I think that we should request a review only after all tests are green, of if a review is needed even with red tests there should be a comment explaining why the exception.

src/twisted/web/_newserver.py
exarkun: Also a new name.

glyph:
Bump - is this something you're still thinking about @hawkowl ?

hawkowl:
@glyph yes! did you want to talk about it?

glyph:
Mainly I wanted to check in about real support for X-Forwarded-For since right now putting Twisted behind a (non-"PROXY") load balancer is such a mess.

src/twisted/web/_newserver.py
```diff
@@ -0,0 +1,257 @@
...
+            baseProtocols.insert(0, b'h2')
+
+        return baseProtocols
+
+
+
+def makeServer(resource, displayTracebacks=True, compressResponses=True,
+               timeout=_REQUEST_TIMEOUT, requestFactory=Request,
+               sessionFactory=Session, logger=None,
+               trustedReverseProxyIPs=None, reactor=None):
```
wsanchez, line 170: Should this be a class method on `Server`?

PR #791, opened 2017-05-23
There is no Travis macOS.

Lukasa:
This script is not run on the macOS builders, so we simply don't need any of this.

GOODBYE OLD FRIEND.

See https://twistedmatrix.com/trac/ticket/9145

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -2,28 +2,6 @@
 set -e
 set -x
 
-if [[ "$(uname -s)" == 'Darwin' ]]; then
-    curl -O https://bootstrap.pypa.io/get-pip.py
-    python get-pip.py --user
-    python -m pip install --user virtualenv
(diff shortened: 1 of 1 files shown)


rodrigc (approved):
We used to run on Travis's OS X builders, but they were too slow and unreliable, so we stopped doing so.

markrwilliams (approved):
The only Darwin I care about is in Australia.

(More seriously, we can always resurrect this from git if we use Travis' macOS builders ever again.)

glyph:
It strikes me that this code might be useful.  However, it's clearly not useful *here*.  Is there any repository to which such hacks might be usefully committed, rather than adding and removing them from each project individually as it happens to want to enable/disable these features?

Lukasa:
Not that I'm aware of, though I'd be happy to see such a place.

PR #792, opened 2017-05-23
Set namespace to "<unknown> instead of raising KeyError when unable to determine namespace from calling context.

wsanchez:
See https://twistedmatrix.com/trac/ticket/7930

---
This change is [Reviewable](https://reviewable.io/reviews/twisted/twisted/792)


src/twisted/logger/test/test_logger.py
```diff
@@ -76,6 +76,16 @@ def test_namespaceDefault(self):
         self.assertEqual(log.namespace, __name__)
 
 
+    def test_namespaceOMGItsTooHard(self):
+        """
+        L{CannotDetermineNamespaceError} is raised when unable to determine the
```
glyph: One should write docstrings in the active voice.  (i.e.: is raised?  raised by whom?)

src/twisted/newsfragments/7930.bugfix
```diff
@@ -0,0 +1,2 @@
+Raise CannotDetermineNamespaceError instead of KeyError when unable to
```
glyph: As per the examples in https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles this should refer to a fully-qualified python name; what is it that raises the exception?

glyph (changes requested):
Thanks for following up on this fix.

I feel a little bit weird about `Logger` being documented and expected to raise an exception when called from a weird context.  I think this should be documented in more depth; at the very least a `@raises` in its class docstring or something.

Plus, the tests are failing.  Please fix and resubmit.

wsanchez:
Since Glyph is feeling squeamish about `Logger()` raising on purpose instead of on accident when called from a crazy context, I'll try setting the namespace to `"<unknown>"` instead.

glyph:
> Since Glyph is feeling squeamish about Logger() raising on purpose instead of on accident when called from a crazy context, I'll try setting the namespace to "<unknown>" instead.

Thanks, appreciated :).

I should note for posterity that this squeamishness is very nearly exclusive to `Logger`; in almost every case, blowing up is the right thing to do, it's just that `Logger` is the one thing that you use to figure out how you blew up when that happens :)

wsanchez:
@glyph: Cool.  Putting back up for review.

glyph (approved):

glyph:
👍 Looking good. Will merge.

PR #794, opened 2017-05-23
Make usage synopsis accurate and consistent when running twisted or twisted.trial with `python -m`

svvitale:
Sort the help argument output so the order is consistent under Python 3.x.  Detect when we’re being run using “-m twisted” or “-m twisted.trial” and use it to build an accurate usage message.  Update tests to patch sys.argv appropriately to match the output of the spawned process.

### DIFF
--- a/src/twisted/newsfragments/9133.bugfix
+++ b/src/twisted/newsfragments/9133.bugfix
@@ -1,0 +1,1 @@
+Detect when we’re being run using “-m twisted” or “-m twisted.trial” and use it to build an accurate usage message.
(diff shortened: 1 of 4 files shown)


hawkowl:
@svvitale Oh, I forgot, you'll need a topfile. You can just put a text file in `src/twisted/newsfragments/` called <ticketnumber>.bugfix, of a description of the bugfix.

glyph:
@svvitale This is a great fix, thank you!

svvitale:
Thanks!  Most of the credit goes to @hawkowl, I just exercised my ability to follow directions :)

PR #795, opened 2017-05-24
[#9137] Fix IResolverSimple not getting bytes on Py2

hawkowl:
https://twistedmatrix.com/trac/ticket/9137


src/twisted/internet/test/test_base.py
```diff
@@ -126,7 +129,7 @@ def test_timeout(self):
         """
         If L{socket.gethostbyname} does not complete before the specified
         timeout elapsed, the L{Deferred} returned by
-        L{ThreadedResolver.getHostByBame} fails with L{DNSLookupError}.
```
markrwilliams, line 129: WHOOPS

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,32 @@ def fakeGetHostByName(name):
...
+        class FakeResolver(object):
+            def getHostByName(self, name, timeouts=()):
+                calls.append(name)
+                return Deferred()
+
+        class JustEnoughReactor(ReactorBase):
+            def installWaker(self):
+                pass
+
+        fake = FakeResolver()
```
markrwilliams, line 178: `verifyObject(IResolverSimple, fake)` would be nice here.

markrwilliams (dismissed):
Looks good!  It'll be nice to close [9103](http://twistedmatrix.com/trac/ticket/9103) so you don't have to write things like `JustEnoughReactor`.

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,32 @@ def fakeGetHostByName(name):
...
+        reactor = JustEnoughReactor()
+        reactor.installResolver(fake)
+        rec = FirstOneWins(Deferred())
+        reactor.nameResolver.resolveHostName(rec, u"xn--pxaix.localhost")
```
glyph: 1. please use `.invalid` rather than `.localhost`, or `something.example.com`, RFCs are great
2. please make the test a bit more rigorous in terms of `resolveHostName` being passed *both* the ascii-as-unicode given here *and* the actual cyrillic text, making sure that the underlying `FakeResolver` is passed the same value in both cases.

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,32 @@ def fakeGetHostByName(name):
...
+
+        self.assertEqual(len(calls), 1)
+        self.assertIs(type(calls[0]), bytes)
```
glyph: Given that the interface of `IResolverSimple` (a) is formally documented as taking `str` and (b) is _practically_ "whatever `gethostbyname` in the stdlib takes" (which is to say: native `str`), this should _really_ be asserting that it is native `str` both on Python 2 and 3.

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,32 @@ def fakeGetHostByName(name):
         result.put(IOError("The I/O was errorful"))
 
 
+    def test_resolverGivenBytes(self):
```
markrwilliams: This needs a skip for Python 3!

markrwilliams:
Also, there should be no need for documentation changes, since [`IResolverSimple.getHostByName` declares that it wants `str`](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IResolverSimple.html).

src/twisted/internet/test/test_base.py
markrwilliams: ...or better yet it should assert it receives `str`, which would ensure it matches the documentation.

src/twisted/internet/_resolver.py
```diff
@@ -169,6 +170,10 @@ def resolveHostName(self, resolutionReceiver, hostName, portNumber=0,
 
         @return: see interface
         """
+        # If it's Python 2, encode this to IDNA.
+        if not _PY3 and isinstance(hostName, unicode):
+            hostName = hostName.encode('idna')
```
glyph: 1. This should really be using `_idnaBytes` from our own idna wrapper so that (as advertised) we will use the `idna` PyPI package when it is available.
2. I'd really like to remove the conditional and deal with IDNA regardless of Python version; either it's our responsibility to convert non-ASCII into IDNA or not, this is separate from what Python type the ASCII is represented by.
3. As the interface is specified, GAIResolver really ought to be doing this as well.  It's not a regression that it isn't, but it's still a defect, and having these behave inconsistently isn't good.

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,32 @@ def fakeGetHostByName(name):
...
+        """
+        L{ThreadedResolver.getHostByName} is passed L{bytes}.
```
glyph: It should be passed `str`.

glyph (dismissed):
A few things need to be adjusted about this, please fix and resubmit.  Most of them are small though.

src/twisted/internet/test/test_base.py
hawkowl: That's done in `installResolver`.

src/twisted/internet/_resolver.py
```diff
@@ -169,6 +170,9 @@ def resolveHostName(self, resolutionReceiver, hostName, portNumber=0,
 
         @return: see interface
         """
+        # Make it an str
+        hostName = nativeString(hostName)
```
markrwilliams: `nativeString` requires its input be ASCII, right?  So `шанти-шанти.рф`, the domain from issue's reproducing script, will fail with a `UnicodeEncodeError`:

```python
# -*- coding: utf-8 -*-

from twisted.internet.base import ThreadedResolver


reactor.installResolver(ThreadedResolver(reactor))
agent = Agent(reactor)

d = agent.request(
    'GET',
    #'https://шанти-шанти.рф',
    'https://xn----7sbb4ac0ad0be6cf.xn--p1ai/',
    None)

def cbResponse(ignored):
    print('Response received: %r' % ignored)
d.addCallback(cbResponse)

def cbShutdown(ignored):
    print("cbShutdown: %r" % ignored)
    reactor.stop()
d.addBoth(cbShutdown)

reactor.run()
````

```
  File "issue9137.py", line 19, in <module>
    None)
  File "/home/mrw/src/python/twisted/src/twisted/web/client.py", line 1639, in request
    parsedURI.originForm)
  File "/home/mrw/src/python/twisted/src/twisted/web/client.py", line 1416, in _requestWithEndpoint
    d = self._pool.getConnection(key, endpoint)
  File "/home/mrw/src/python/twisted/src/twisted/web/client.py", line 1302, in getConnection
    return self._newConnection(key, endpoint)
  File "/home/mrw/src/python/twisted/src/twisted/web/client.py", line 1314, in _newConnection
    return endpoint.connect(factory)
  File "/home/mrw/src/python/twisted/src/twisted/internet/endpoints.py", line 2091, in connect
    self._wrapperFactory(protocolFactory)
  File "/home/mrw/src/python/twisted/src/twisted/internet/endpoints.py", line 903, in connect
    EndpointReceiver, self._hostText, portNumber=self._port
  File "/home/mrw/src/python/twisted/src/twisted/internet/_resolver.py", line 174, in resolveHostName
    hostName = nativeString(hostName)
  File "/home/mrw/src/python/twisted/src/twisted/python/compat.py", line 435, in nativeString
    return s.encode("ascii")
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-4: ordinal not in range(128)
```

File: src/twisted/internet/test/test_base.py [hunk 9]
```diff
@@ -155,6 +158,38 @@ def fakeGetHostByName(name):
...
+        class JustEnoughReactor(ReactorBase):
+            def installWaker(self):
+                pass
+
+        fake = FakeResolver()
+        reactor = JustEnoughReactor()
+        reactor.installResolver(fake)
+        rec = FirstOneWins(Deferred())
+        reactor.nameResolver.resolveHostName(
+            rec, u"example.invalid")
```
markrwilliams: These domains should have non-ASCII characters.

markrwilliams (dismissed):
The problem reported in the issue still exists :(

src/twisted/internet/_resolver.py
rodrigc: I agree with @glyph's point about removing the conditional and dealing with IDNA regardless of Python version

src/twisted/internet/test/test_base.py
hawkowl: So, @glyph and I originally decided "nope, just make them pass idna". Turns out, Agent will DE-IDNA IT, and then SEND UNICODE TO THE API. So, I kinda have to now. Buh bow, Hawkie loses.

src/twisted/internet/_resolver.py
```diff
@@ -169,6 +171,19 @@ def resolveHostName(self, resolutionReceiver, hostName, portNumber=0,
...
+            hostName = hostName.encode('ascii')
+        except UnicodeEncodeError:
+            # If it's not just ASCII, IDNA it. We don't want to give a Unicode
+            # string with non-ASCII in it to Python 3, as if anyone passes that
+            # to a Python 3 stdlib function, it will probably use the wrong
+            # IDNA version and break absolutely everything
+            hostName = _idnaBytes(hostName)
+
+        # Make sure it's passed down as a native str, to maintain the interface
+        hostName = nativeString(hostName)
```
markrwilliams, line 185: I think this is correct.  `SimpleResolverComplexifier` implements [`IHostnameResolver`](http://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IHostnameResolver.html), so its [`resolveHostname`](http://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IHostnameResolver.html#resolveHostName) method accepts `unicode` objects.

If `hostName = u'192.168.1.1'`, then by line 186 `hostName = '192.168.1.1'` on Python 2 and Python 3, so that `SimpleResolverComplexifier.resolveHostName` always receives a `str`.  The same is true for any pure-ASCII domain name.

If `hostName = u'шанти-шанти.рф'`, then by line 186 `hostName = 'xn----7sbb4ac0ad0be6cf.xn--p1ai'`.  This is also a native string.

Resolvers that call the system's resolver, like `ThreadedResolver`, will resolve these values even on Python 3.  As the comment states, Python 3 uses the built in `idna` codec to encode native strings to bytes, and that codec encodes codepoints that fit within a single byte as their ASCII equivalents.  That is, the `idna` codec is idempotent.

The net result is that the issue's reproducing script now works!

src/twisted/internet/test/test_base.py
```diff
@@ -155,6 +158,45 @@ def fakeGetHostByName(name):
...
+        reactor.installResolver(fake)
+        rec = FirstOneWins(Deferred())
+        reactor.nameResolver.resolveHostName(
+            rec, u"example.example")
+        reactor.nameResolver.resolveHostName(
+            rec, "example.example")
+        reactor.nameResolver.resolveHostName(
+            rec, u"v\xe4\xe4ntynyt.example")
+        reactor.nameResolver.resolveHostName(
+            rec, u"\u0440\u0444.example")
```
markrwilliams, line 189: Should these again be `.invalid`?  Also it might be a good idea include an IDNA encoded native string, e.g. `" xn----7sbb4ac0ad0be6cf.xn--p1ai"`, so that this also serves as a regression test.

markrwilliams (approved):
This looks great!  Great job sorting out an unpleasantly thorny issue.

I included my own notes on how the `resolveHostName` code works for posterity.  The comment is very helpful!

Two small suggestions on the test.  Please address and then merge this!

src/twisted/internet/test/test_base.py
hawkowl: Nah, they're not actually invalid domain names.

PR #798, opened 2017-05-29
Hostname and TCP client endpoints take float timeouts.

Lukasa:
See https://twistedmatrix.com/trac/ticket/9151


src/twisted/newsfragments/9151.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet.endpoints.HostnameEndpoint and twisted.internet.endpoints.TCP4Client endpoint documentation updated to correctly reflect that the timeout argument takes a float as well as an int.
```
twm: I think that this file should be ``.doc`` since there are no functional changes here.

twm (commented):
Code changes look good. The topfile should be renamed before merge, though.

Thanks!

src/twisted/newsfragments/9151.bugfix
Lukasa: Oh hey, I didn't know that .doc was a thing. Good spot!

PR #799, opened 2017-06-02
Allow multiple unregister from TLSMemoryBIOProtocol

Lukasa:
See https://twistedmatrix.com/trac/ticket/9156


hawkowl (approved):
When it goes green, land.

PR #800, opened 2017-06-03
[#9158] switch conch to using sign/verify methods

reaperhulk:
I'm not sure what to put for the news fragment, and the twisted project will need to decide if they want to actually bump their dep or potentially just allow deprecation warnings from cryptography.

refs https://github.com/pyca/cryptography/issues/3659


alex (approved):

alex:
Code looks good. Please add a newsfile as well.

src/twisted/conch/newsfragments/9158.misc
```diff
@@ -0,0 +1 @@
+Bumped minimum cryptography version to 1.5.
```
mithrandi: The contents of .misc files is ignored, so usually they are left empty.

glyph:
As for the news fragment I think this can be a `.misc` (it's "just" keeping up with the deprecation cycle of another library?) or a brief `.bugfix` note just noting the raised minimum version.

reaperhulk:
@glyph I have it as a misc now, so if that's good enough 👍

PR #801, opened 2017-06-03
Appease pyflakes on Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9159


src/twisted/internet/abstract.py
```diff
@@ -13,22 +13,27 @@
...
+    # Python 3
+
     def _concatenate(bObj, offset, bArray):
         # Python 3 lacks the buffer() builtin and the other primitives don't
         # help in this case.  Just do the copy.  Perhaps later these buffers can
         # be joined and FileDescriptor can use writev().  Or perhaps bytearrays
         # would help.
         return bObj[offset:] + b"".join(bArray)
-else:
-    def _concatenate(bObj, offset, bArray):
```
markrwilliams, line 28: You could keep the `if _PY3` logic as it was and still satisfy pyflakes by retrieving `buffer` from [`__builtin__`](https://docs.python.org/2/library/__builtin__.html#module-__builtin__):

```python
else:
    from __builtin__ import buffer
    def _concatenate(bObj, offset, bArray):
```

src/twisted/python/compat.py
```diff
@@ -841,7 +841,10 @@ def _bytesRepr(bytestring):
 else:
     _tokenize = tokenize.generate_tokens
 
-
+if _PY3:
```
markrwilliams: See above about using `__builtin__`.

markrwilliams (approved):
One suggestion that would reduce the size of the diff.  Please address and then merge.

PR #802, opened 2017-06-04
[#6938] Do binary logging in LogFile + HTTP

hawkowl:
https://twistedmatrix.com/trac/ticket/6938


glyph (approved):
I thought real, real hard about the compatibility surface here and I *think* this looks great.

In any case, it will be my fault now when it isn't.  Please land when the builders go green.

PR #803, opened 2017-06-05
Avoid double-firing the hard shutdown.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/9157


src/twisted/web/test/test_http2.py
```diff
@@ -2570,3 +2570,20 @@ def test_losingConnectionWithNoAbortTimeOut(self):
         reactor.advance(2**32)
         self.assertTrue(transport.disconnecting)
         self.assertFalse(transport.disconnected)
+
+    def test_connectionLostAfterForceClose(self):
```
hawkowl: This needs two spaces between method definitions

hawkowl (approved):
Fix the nitpick then land.

glyph:
🎉  thank you 🎉

psi29a:
When is the next release of Twisted with this included? :)

hawkowl:
Soon™ (worst case, next week at PyCon AU there will be a prerelease)

psi29a:
Thank you! :)

PR #738, opened 2017-03-06
Port mailmail to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8487

PR #805, opened 2017-06-06
9161 py3imap4

glyph:
twisted.mail.imap4 now works on Python 3.


rodrigc (commented):
I think you have a case in this patch of b"%s" % something which won't work on Python 3.3, which is still being used on the builders.

glyph:
> I think you have a case in this patch of b"%s" % something which won't work on Python 3.3, which is still being used on the builders.

Which builders?

glyph:
None of the build statuses on this PR are for 3.3, and I don't see a 3.3 on https://buildbot.twistedmatrix.com/boxes-supported?branch=trunk&num_builds=10 anywhere.

rodrigc:
.travis.yml on trunk has a python 3.3 build

glyph:
> .travis.yml on trunk has a python 3.3 build

Yes, but this is a configuration mistake that is irrelevant because we always explicitly specify which environment to run, no?  (As it happens I'm running the py33 tests locally and they pass.)

rodrigc:
The .travis.yml file explicitly specifies using python 3.3 for the 3.3 build.

glyph:
> The .travis.yml file explicitly specifies using python 3.3 for the 3.3 build.

Ah, I see the build now.  (It's still passing, FWIW, but that means the lines in question are uncovered. :))

rodrigc:
Lulz!

glyph:
OK, tests are now failing (as they should be).

glyph:
This explains why the 3.3 build was passing despite invalid syntax ;)

rodrigc:
@markrwilliams can you look at: https://github.com/twisted/twisted/pull/738 ?  maybe we can combine efforts as you move forward with your porting work

markrwilliams:
@rodrigc Sure!

glyph:
May I suggest prioritizing pyflakes fixes so that the tests start passing (periodically) again?  I don't think the issue is fully fixed but maybe this 4kloc branch should just land at this point, and additional test additions could be smaller and slightly more reviewable?

markrwilliams:
@glyph I see two choices: we can land this after satisfying pyflakes, or I can split this into at least two PRs: one that adds tests, and one that actually ports `imap4` to Python 3.

I'd like to split this up for a couple of reasons:

1. No one is going to be able to review a PR this large.
2. We need 100% test coverage before we can say the port actually works.
3. There are unrelated bugs fixed in this branch that should be in their _own_ PRs, like [9207](https://github.com/twisted/twisted/pull/805/commits/8f5ac1a0c2a543ea0d088c52bd068e49380d719a).

Running `twisted.mail.test.test_imap` from this branch against `twisted.mail.imap4` from trunk results in 7 failures and 7 errors, so I don't think it's going to be a terrific amount of effort to split this up.  What d you think?

src/twisted/mail/imap4.py
```diff
@@ -4784,26 +5035,35 @@ def collapseNestedLists(items):
     """
     pieces = []
     for i in items:
+        if isinstance(i, unicode):
+            # anything besides ASCII will have to wait for an RFC 5738
+            # implementation.
```
glyph: Could this be a link to a Twisted ticket?

src/twisted/mail/test/test_imap.py
```diff
@@ -5348,6 +7242,108 @@ def timedOut(failure):
...
+        def expireTime():
+            c.advance(self.server.POSTAUTH_TIMEOUT * 2)
+
+        d = self.connected.addCallback(strip(login))
+        d.addCallback(strip(expireTime))
+
+        # The loopback method's Deferred fires the connection is
+        # closed, and the server closes the connection as a result of
+        # expireTime.
+        return defer.gatherResults([d, self.loopback()])
```
glyph, line 7264: It is somewhat disheartening to see so many new tests returning `Deferreds` when they really shouldn't need to; however, I can see that it's following the local convention in this test suite, and it's better to have coverage than not.  Can you file a ticket to replace these with `IOPump` and make them not return `Deferred`s?

tox.ini
```diff
@@ -100,7 +100,7 @@ commands =
 
     ; Run the tests wrapped using coverage.
     {withcov,coverage}: python {toxinidir}/admin/_copy.py {toxinidir}/admin/zz_coverage.pth {envsitepackagesdir}/zz_coverage.pth
-    {withcov,coverage}: coverage erase
+;    {withcov,coverage}: coverage erase
```
glyph: Please revert this and let's discuss whether it's appropriate on a separate ticket :).

src/twisted/python/_setup.py
```diff
@@ -364,10 +364,15 @@ def _checkCPython(sys=sys, platform=platform):
 _isCPython = _checkCPython()
 
 notPortedModules = [
+    "twisted.internet.glib2reactor",
```
glyph: Why are we *adding* to this list in this branch?

glyph: To make this more actionable: if this is not necessary in this branch, please revert this hunk out.

tox.ini
markrwilliams: Whoops - that should not be there!

src/twisted/python/_setup.py
markrwilliams: This is some kind of merge issue :(

src/twisted/mail/imap4.py
markrwilliams: https://twistedmatrix.com/trac/ticket/9258 !

src/twisted/mail/test/test_imap.py
markrwilliams: https://twistedmatrix.com/trac/ticket/9259 !

rodrigc:
@markrwilliams I can confirm that with this fix, the IMAP example on the front page of https://twistedmatrix.com now works with Python 3.6. (I tested with *strport="ssl:imap.gmail.com:993"*)

Do you have edit permissions on the front page?  In that example, the *user* and *password* need to be prefixed with *b* for bytes.  Other than that, the example works.  It would be good to fix that example on the wiki.

markrwilliams:
@rodrigc Thank you so much for your thoroughness!  I edited the front page - please check I got it right!

rodrigc:
Looks good.
If I replace **strport="ssl:imap.gmail.com:993"** with **strport="tls:imap.gmail.com:993"**,
I get an error: `IMAP4 client interaction failed [('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')]`

I did a trace with wireshark and found this error:

**TLSv1.2 93 Alert (Level: Fatal, Description: Unknown CA)**

Is that a problem on gmail's side, or in the Twisted code?

markrwilliams:
@rodrigc I think `tls` not working is an issue with your OpenSSL installation, but I think `ssl` a bad endpoint and we should get rid of it.  It does not verify server certificates against _any_ trust root unless you specify a directory of CA certificates (e.g. `ssl:imap.gmail.com:993:caCertsDir=/etc/ssl/certs`) or a hostname (e.g., `ssl:imap.gmail.com:993:hostname=imap4.gmail.com`). In the former case, it will read CA certs out of the provided directory, while in the latter case it use [`t.i.ssl.platformTrust`](https://twistedmatrix.com/documents/current/api/twisted.internet.ssl.platformTrust.html) to determine the location of those certs.

It will also not offer SNI or perform SNI validation without the `hostname` keyword argument.

This is a needlessly complicated API with unsafe defaults.  It's so bad that the same [issue](http://bugs.python.org/issue21308) [prompted](https://www.python.org/dev/peps/pep-0466/) the release of [Python 2.7.9](https://www.python.org/downloads/release/python-279/).

Regardless, `tls` verifies that a trusted CA certificate has signed a server's certificate, whereas `ssl` does not.  Upgrading your system's trust store - `security/ca_root_nss`  on FreeBSD, `ca-certificates` on Debian-based distributions - might help, but you might also have to upgrade OpenSSL.

Let me know if that works!

rodrigc:
I am using OS X, and `/etc/ssl/certs` does not exist.  Also `/usr/local/etc/openssl/certs`
is empty.  So trying this did not work.  There is some discussion about this in:

https://github.com/twisted/twisted/commit/71282ea1076374e04c5bcc772fc12d989021df27
https://twistedmatrix.com/trac/ticket/6372
https://twistedmatrix.com/trac/ticket/5446

I don't know how to get this to work.

markrwilliams:
@rodrigc I don't have a Mac, but I poked around inside a few wheels and confirmed with @alex that `pyopenssl` reads `/usr/local/etc/openssl/certs` via `cryptography`, and that this directory can be populated by installing OpenSSL with brew.

You might need to upgrade `pyopenssl` and `cryptography` as well.

Let me know if all this makes the `tls` endpoint work.

rodrigc:
@alex do you know the steps required to get the Twisted tls endpoint to work on Mac?

It would be good to have a web link with the steps needed to get the tls endpoint
to work on Mac (if we don't have one already).  Also, it would be good to refer to this link from the *Mail Client* example web page.

I think that many people use Mac to code things in Python,
and it is quite reasonable to assume that many newcomers to Twisted who use
Mac will copy and paste the *Mail Client* example to try things out.

glyph:
The caveats for macOS are documented here: https://twistedmatrix.com/trac/ticket/9209

PR #806, opened 2017-06-06
Remove some items from notPortedModules

rodrigc:
https://twistedmatrix.com/trac/ticket/9162


src/twisted/python/_setup.py
```diff
@@ -363,12 +363,6 @@ def _checkCPython(sys=sys, platform=platform):
 _isCPython = _checkCPython()
 
 notPortedModules = [
-    "twisted.internet.glib2reactor",
```
markrwilliams, line 366: This is uncovered in the tests because glib isn't available.  How do you know this is ported?  Or is the idea that removing these allows them to be ported.

src/twisted/python/_setup.py
```diff
@@ -363,12 +363,6 @@ def _checkCPython(sys=sys, platform=platform):
...
-    "twisted.internet.gtk2reactor",
```
markrwilliams, line 367: Same as above.

src/twisted/python/_setup.py
```diff
@@ -363,12 +363,6 @@ def _checkCPython(sys=sys, platform=platform):
...
-    "twisted.internet.pyuisupport",
```
markrwilliams, line 368: Does this have any actual tests?

src/twisted/python/_setup.py
```diff
@@ -363,12 +363,6 @@ def _checkCPython(sys=sys, platform=platform):
...
-    "twisted.internet.test.process_connectionlost",
-    "twisted.internet.test.process_gireactornocompat",
```
markrwilliams, line 370: Tests again skipped on the builder because of missing dependencies.

src/twisted/python/_setup.py
```diff
@@ -363,12 +363,6 @@ def _checkCPython(sys=sys, platform=platform):
...
-    "twisted.internet.tksupport",
```
markrwilliams, line 371: Does this even have tests?

markrwilliams (changes requested):
Are some of these being removed so they can be ported?

rodrigc:
For the reactors, those require 3rd party libraries to be installed and new configurations on the builders to actually run them to gather the coverage information.  Most of those reactors use 3rd party GUI frameworks, so it is highly unlikely that the Twisted buildbots will run tests for them, since there are no GUI tests that work and run on the Twisted buildbots.
There is no point in keeping those in this list, and that will facilitate further work on this modules, and possibly automated GUI testing, if someone is motivated to do that.

For these two:
-    "twisted.internet.test.process_connectionlost",
-    "twisted.internet.test.process_gireactornocompat",

They should have been removed from this list in 2b73df411313d147b0b4b775fed98d379afa7b42 , but looks like that was an oversight and they were missed.

markrwilliams:
@rodrigc We should test those GUI reactors or remove them.  [I filed a ticket for that](https://twistedmatrix.com/trac/ticket/9168#ticket).

markrwilliams (approved):
We should deal with the GUI reactors at some point.  But this looks good to merge.

PR #807, opened 2017-06-07
Replace .3only files with .py files

rodrigc:
https://twistedmatrix.com/trac/ticket/9165


exarkun (changes requested):
The motivation is unclear.  Please explain.

rodrigc:
The motivation is to have .py files throughout the Twisted tree,
and not special files only for Python 3.

glyph (changes requested):
> The motivation is to have .py files throughout the Twisted tree,
> and not special files only for Python 3.

That doesn't seem to be a "motivation", you're just describing what you're doing here.  Changing these into `.py` files is going to cause any tooling on python 2 which tries to parse `.py` files to emit `SyntaxError` tracebacks, to no particular benefit.  Without a better explanation of why this is helpful, I think the ticket should just be closed.

rodrigc:
Having the .3only files adds complexity to the installer and MANIFEST.in files, and I am trying to push to reduce this complexity as Twisted is getting closer to having more tests passing on Python 3.

In terms of the tooling, the pylint build was the only thing failing, and I fixed that by switching the pylint build in .travis.yml to run under Python 3.5.

The only other tooling that does not work under Python 3 is pydoctor which still runs only on
python 2 due to epydoc,
but that does not fail with this change.

The .3only files in the tree are only in tests

glyph:
> Having the .3only files adds complexity to the installer and MANIFEST.in files,

By the "installer" are you referring to something other than `MANIFEST.in` itself?

> and I am trying to push to reduce this complexity as Twisted is getting closer to having more tests passing on Python 3.

Getting _all_ tests passing on py3 is a prerequisite to dropping py2, which is probably when this will be appropriate :).

> In terms of the tooling, the pylint build was the only thing failing, and I fixed that by switching the pylint build in .travis.yml to run under Python 3.5.

Sure, that fixes it in our build environment, but the code should lint under *both* 2 and 3 until 2 is no longer a concern, and this makes that more complex to automatically verify.

More importantly though other tools might try to import files from twisted in an integrated environment; this could break things like modulegraph, or invalidate some assumptions from distutils or setuptools.  It's hard to prove a negative here, so maintaining the "all the .py files in site-packages ought to be at least parseable" invariant seems worthwhile to avoid confusing breakages elsewhere.

> The only other tooling that does not work under Python 3 is pydoctor which still runs only on
> python 2 due to epydoc,
> but that does not fail with this change.

That makes me suspicious it is failing silently, as it *should* be complaining it can't parse these files.

> The .3only files in the tree are only in tests

And that should definitely remain the case :).  The only reason to have 3-only syntax within the codebase is for integration tests with runtime features invoked by that syntax.

rodrigc:
I don't agree that doing this should be tied to dropping py2 support.

With my change, all the py2 builders and travis builders still work, and that should continue
to work for as long as py2 is supported in Twisted.

The .3only files were introduced in the tests because pylint was run on a py2 builder in travis,
and there was no way to run pylint on tests which introduced 3-only syntax.

I think that running pylint on 3, and making sure that the py2 builders do not fail, and eliminating
3only files is OK moving forward for Twisted.

In terms of tooling, I mentioned pydoctor (by the way, I emailed you about looking at https://github.com/twisted/pydoctor/pull/134 again, but got no response from you).

You mentioned modulegraph, but that supports Python 3.5: https://pypi.python.org/pypi/modulegraph/

glyph:
While the date where we *should* actually do this fast approaches, this particular branch probably won't make it due to the merge conflicts and such.  So GCing the open PR for that reason.  As usual, not a rejection, etc.  Further discussion should probably happen on the ticket :)

PR #809, opened 2017-06-10
Fix test_tap on Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9169


src/twisted/words/test/test_tap.py
```diff
@@ -12,17 +12,17 @@ class WordsTapTests(unittest.TestCase):
...
+    joeWrong = credentials.UsernamePassword(b'joe', b'bar')
 
 
     def setUp(self):
         """
         Create a file with two users.
         """
         self.filename = self.mktemp()
-        self.file = open(self.filename, 'w')
+        self.file = open(self.filename, 'wb')
```
markrwilliams, line 25: This should be bytes because [`twisted.cred.checkers.FilePasswordDB` reads its data as bytes](https://github.com/twisted/twisted/blob/trunk/src/twisted/cred/checkers.py#L211), and opening this file as text would mean the contents would be whatever the [system's preferred locale is](https://docs.python.org/3/library/functions.html#open), which might not be ASCII congruent.

markrwilliams (approved):
Builds are green and the diff makes sense.  LGTM.

PR #811, opened 2017-06-10
twisted.web.client documentation examples work on Python 3

markrwilliams:
They also consistently use http://httpbin.org/, because it returns
more informative responses than example.com, google.com, or yahoo.com
and is no more unreliable than them.

https://twistedmatrix.com/trac/ticket/9172


rodrigc (approved):
Before merging this,
fix ```docs/web/howto/client.rst by replacing stringprod.py``` with ```bytesprod.py``` so that the documentation build passes

PR #812, opened 2017-06-11
[#7945] Twisted C Extension Package

hawkowl:
https://twistedmatrix.com/trac/ticket/7945


cext/src/_twisted_c_extensions/_setup.py
```diff
@@ -0,0 +1,195 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+# pylint: disable=I0011,C0103,C9302,W9401,W9402
+
+"""
+Setuptools convenience functionality.
+
+This file must not import anything from Twisted, as it is loaded by C{exec} in
+C{setup.py}. If you need compatibility functions for this code, duplicate them
```
glyph: "duplicate them here" <- this seems like a recipe for a nightmare of copy/pasting.

src/twisted/newsfragments/7945.feature
```diff
@@ -0,0 +1 @@
+Twisted's cross-platform C extensions (mainly used for twisted.python.sendmsg) are now distributed in a separate package, removing the requirement for a C compiler on Linux/MacOS/FreeBSD for the majority of Twisted's functionality. This new package can be installed by ``pip install Twisted[c]``.
```
glyph: I really don't like `[c]` as a name for this.

setuptools extras are already inscrutable enough without having their names refer to random implementation details (in this case, that the functionality in question happens to be written in C).  Many of Twisted's other extras require code written in C, but we're not calling it `twisted[libcrypto,libssl]` because the _reason_ you install those dependencies is `tls`.

In this case, the reason that you (as a user) want this code is that it makes `IUNIXTransport.sendFileDescriptor` work on Python 2.7.  So:

- Could we call the extra `sendfd27`?
- Could we update the docs for `IUNIXTransport.sendFileDescriptor`?
- I'm pretty sure this should be included in one of the aggregates, like `all_non_platform` .

(The fact that our own test suite needs it to test `failure` is a dumb implementation detail that we can keep secret, as far as the extras are concerned.)

cext/src/_twisted_c_extensions/_setup.py
hawkowl: @glyph This is already the case for `_setup.py` in Twisted, because it can't import Twisted yet :)

hawkowl:
Okay, gonna put this up for re-review.

src/twisted/python/_setup.py
```diff
@@ -68,9 +68,13 @@
...
         "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
     ],
 )
 
+# THIS MUST REMAIN IN SYNC!
+twistedVersion = "17.5.0"
+
 
 _dev = [
```
markrwilliams: Should this include `twistedcextensions` to include `_raiser`?

cext/src/_twistedcextensions/_setup.py
```diff
@@ -0,0 +1,195 @@
...
+    # This is a workaround for distutils behavior; ext_modules isn't
+    # actually used by our custom builder.  distutils deep-down checks
+    # to see if there are any ext_modules defined before invoking
+    # the build_ext command.  We need to trigger build_ext regardless
+    # because it is the thing that does the conditional checks to see
+    # if it should build any extensions.  The reason we have to delay
+    # the conditional checks until then is that the compiler objects
+    # are not yet set up when this code is executed.
+    arguments["ext_modules"] = extensions
+    # Use custome class to build the extensions.
```
markrwilliams: Typo

docs/installation/howto/optional.rst
```diff
@@ -37,7 +36,9 @@ The following optional dependencies are supported:
 
 * **serial** - the `pyserial`_ package to work with serial data.
 
-* **all_non_platform** - installs **tls**, **conch**, **soap**, and **serial** options.
+* **sendfd27** - the ``twistedcextensions`` package that provides support for sending file descriptors over UNIX sockets on Python 2.7.
```
markrwilliams, line 42: `dev` will also need to be updated if the `dev` extra includes `twistedcextensions`.

src/twisted/python/_setup.py
```diff
@@ -68,9 +68,13 @@
         "Programming Language :: Python :: 3.3",
         "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
     ],
 )
 
+# THIS MUST REMAIN IN SYNC!
+twistedVersion = "17.5.0"
```
markrwilliams: Could this be changed to:

`twistedVersion = 'Twisted NEXT'.split(None, 1)[-1]`

It's ugly but it would allow `incremental` to managed the version, right?

docs/core/development/policy/release-process.rst
```diff
@@ -108,25 +108,27 @@ How to do a release candidate
...
 3. In your Git repo, fetch and check out the new release branch.
-4. Run ``python -m incremental.update Twisted --rc``
-5. Commit the changes made by Incremental.
-6. Run ``towncrier``.
-7. Commit the changes made by towncrier - this automatically removes the NEWS newsfragments.
-8. Bump copyright dates in ``LICENSE``, ``twisted/copyright.py``, and ``README.rst`` if required
-9. Push the changes up to GitHub.
-10. Run ``python setup.py sdist --formats=bztar -d /tmp/twisted-release`` to build the tarballs.
-11. Copy ``NEWS.rst`` to ``/tmp/twisted-release/`` for people to view without having to download the tarballs.
+4. Run ``python -m incremental.update Twisted --rc`` to update Twisted's version and update the constant in ``src/twisted/python/setup.py`` to match.
```
markrwilliams: This will need to change if the constant can be made manipulable by `incremental`.

markrwilliams (changes requested):
One typo and some suggestions around the `dev` extra.

Also, we'll need to update https://twistedmatrix.com/trac/wiki/TwistedDevelopment#Creatingyourworkenvironment to include installing `twistedcextensions.`

Thanks so much for this multi-year effort - I'm excited to get this merged!

src/twisted/python/_setup.py
hawkowl: Nope! Because that NEXT will get rewritten, and then it won't be NEXT anymore.

src/twisted/python/_setup.py
hawkowl: done

docs/installation/howto/optional.rst
hawkowl: ughhh we can't have it in dev because it breaks builders and stuff

exarkun:
Superseded by now-merged https://github.com/twisted/twisted/pull/1446

PR #815, opened 2017-06-12
AXFR AttributeError

markrwilliams:
https://twistedmatrix.com/trac/ticket/9174

Failed TCP connections for AXFR queries no longer raise an AttributeError.

AXFRController now has a pending list that contains a one-element
tuple Deferred returned by Resolver.lookupZone, and DNSClientFactory
only requires the first element of that tuple be a Deferred.

Note that lookupZone cannot use Resolver.queryTCP because it requires
AXFRController, and Resolver.queryTCP assumes that it can reuse an
existing connection whose controller is that Resolver instance.

See Changelog in a237ef3123b86c1a915d078f12adad63b9bfcb24.


glyph (approved):
Improvement to coverage, clear test, short patch, clear bugfix: everything about this looks good.  I'll land it momentarily.

PR #816, opened 2017-06-14
Fix stacktrace in twisted.smtp on buildbot

rodrigc:
https://twistedmatrix.com/trac/ticket/9180

### DIFF
--- a/src/twisted/mail/newsfragments/9180.bugfix
+++ b/src/twisted/mail/newsfragments/9180.bugfix
@@ -1,0 +1,1 @@
+Sending a list of recipients with twisted.smtp.SenderFactory has been fixed.  This fixes a problem found when running buildbot.
--- a/src/twisted/mail/smtp.py
+++ b/src/twisted/mail/smtp.py
@@ -17,8 +17,9 @@ import socket
 import os
 import random
 import binascii
-import email.utils
 import warnings
+
+from email.utils import parseaddr
 
 from zope.interface import implementer
 
@@ -176,7 +177,7 @@ def quoteaddr(addr):
     if isinstance(addr, bytes):
         addr = addr.decode('ascii')
 
-    res = email.utils.parseaddr(addr)
+    res = parseaddr(addr)
 
     if res == (None, None):
         # It didn't parse, use it as-is
@@ -1896,7 +1897,7 @@ class SMTPSenderFactory(protocol.ClientFactory):
                 if not isinstance(_email, bytes):
                     _email = _email.encode('ascii')
 
-                toEmailFinal.append(email)
+                toEmailFinal.append(_email)
             toEmail = toEmailFinal
 
         self.fromEmail = Address(fromEmail)
--- a/src/twisted/mail/test/test_smtp.py
+++ b/src/twisted/mail/test/test_smtp.py
@@ -759,14 +759,38 @@ class TimeoutTests(unittest.TestCase, LoopbackMixin):
         return d
 
 
-    def test_SMTPClient(self):
+    def test_SMTPClientRecipientBytes(self):
         """
         Test timeout for L{smtp.SMTPSenderFactory}: the response L{Deferred}
         should be errback with a L{smtp.SMTPTimeoutError}.
         """
         onDone = defer.Deferred()
         clientFactory = smtp.SMTPSenderFactory(
-            'source@address', 'recipient@address',
+            'source@address', b'recipient@address',
+            BytesIO(b"Message body"), onDone,
+            retries=0, timeout=0.5)
+        return self._timeoutTest(onDone, clientFactory)
+
+
+    def test_SMTPClientRecipientUnicode(self):
+        """
+        Use a L{unicode} recipient.
+        """
+        onDone = defer.Deferred()
+        clientFactory = smtp.SMTPSenderFactory(
+            'source@address', u'recipient@address',
+            BytesIO(b"Message body"), onDone,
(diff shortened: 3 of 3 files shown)


src/twisted/mail/newsfragments/9180.bugfix
```diff
@@ -0,0 +1 @@
+Sending a list of recipients with twisted.smtp.SenderFactory has been fixed.  This fixes a problem found when running buildbot.
```
exarkun, line 1: This fragment doesn't follow the style guidelines.

src/twisted/mail/test/test_smtp.py
```diff
@@ -759,14 +759,38 @@ def check(ign):
...
-            'source@address', 'recipient@address',
+            'source@address', b'recipient@address',
+            BytesIO(b"Message body"), onDone,
+            retries=0, timeout=0.5)
+        return self._timeoutTest(onDone, clientFactory)
+
+
+    def test_SMTPClientRecipientUnicode(self):
+        """
+        Use a L{unicode} recipient.
```
exarkun, line 777: For what?

src/twisted/mail/test/test_smtp.py
```diff
@@ -759,14 +759,38 @@ def check(ign):
...
+        """
+        onDone = defer.Deferred()
+        clientFactory = smtp.SMTPSenderFactory(
+            'source@address', u'recipient@address',
+            BytesIO(b"Message body"), onDone,
+            retries=0, timeout=0.5)
+        return self._timeoutTest(onDone, clientFactory)
```
exarkun, line 784: How about factoring the duplication between this and the previous test out?  This will make it clear what's varying.

src/twisted/mail/test/test_smtp.py
```diff
@@ -759,14 +759,38 @@ def check(ign):
...
+
+
+    def test_SMTPClientRecipientList(self):
+        """
+        Use a L{list} of recipients.
+        """
+        onDone = defer.Deferred()
+        clientFactory = smtp.SMTPSenderFactory(
+            'source@address', (u'recipient1@address', b'recipient2@address'),
```
exarkun, line 793: Oh wow what an awful API.  This should probably be deprecated and removed.

src/twisted/mail/smtp.py
```diff
@@ -1896,7 +1897,7 @@ def __init__(self, fromEmail, toEmail, file, deferred, retries=5,
                 if not isinstance(_email, bytes):
                     _email = _email.encode('ascii')
 
-                toEmailFinal.append(email)
+                toEmailFinal.append(_email)
```
exarkun, line 1900: Maybe it's not a good idea to have variables named `email` and `_email` in the same function?

exarkun (commented):
Thanks.  Some changes would be nice.  Though I see you merged this already without any apparent review?

Are you treating your Twisted commit privileges responsibly?

src/twisted/mail/newsfragments/9180.bugfix
rodrigc: @exarkun thanks for the additional feedback.  @tardyp reviewed the request.

I may not be as talented as some of the other Twisted devs, but I do take my commit privileges responsibly, to the best of my ability.

exarkun: Where is tardyp's review?  The  convention is for the review to be left on the PR.  Failing that, the trac ticket.  I don't see it in either place.  Am I missing it somewhere?

tardyp: We did discuss that over IRC

glyph:
It's important to go through all the parts of the review process as documented on https://twistedmatrix.com/trac/wiki/ReviewProcess#Reviewers:Howtoreviewachange.

In this case, the parts that were skipped are:

- "write a detailed comment about all potential improvements to the branch", and
- "Remove the 'review' keyword from the ticket".

These are pretty critical pieces of the process.  We should improve the clarity of the language here - the comment in question is assumed to have to be either on the PR or the trac ticket, comments elsewhere (not in email, and especially not on IRC, where they'll be entirely lost to other interested readers) don't count.  But in any case, the `review` keyword was not attached or removed, so there's no record of the code review, so this is a blatantly out-of-policy merge.

Since you, the author ( @rodrigc ) and not the reviewer ( @tardyp ) , are the project committer in this case, as per https://twistedmatrix.com/trac/wiki/ReviewProcess#Whocanreview, "the committer may accept it and land their change, but they are then responsible for the adequacy of the review".  This review policy is in place to allow new reviewers to become familiar with our process so they can become committers - it is both out of process and unhelpful to that reviewer if they aren't actually following it.  Usually if you have a non-committer review your changes, you should provide some feedback on the quality of the review in order to help give them confidence to eventually join the project.

Hopefully this explanation makes it a little clearer why these aspects of the process are important, and what exactly is required.

tardyp:
Well, I cannot say I did an official review, as I have no familiarity with that code.
We discussed the problem over IRC, and I said that the fixed looked good, combined with the fact that we had people reporting the fix was fixing their problem.

It looks like @rodrigc was overzealous merging that fix thinking that 17.5.0 was breaking Buildbot email feature. I think he can followup with another PR addressing the review comments.

After a more thorough review, I doubt that this is recent regression, as the problem comes from https://github.com/twisted/twisted/commit/e73c9c55c8d27d5d5afc333e93ec03d01f4ec5c0 which is 10 month old.

PR #817, opened 2017-06-14
MessageSet tests + improvements

markrwilliams:
https://twistedmatrix.com/trac/ticket/9177

(This branch originally was also going to include https://twistedmatrix.com/trac/ticket/9178 and https://twistedmatrix.com/trac/ticket/9179, but it's larger than I anticipated so I'm putting it into review now.)


src/twisted/mail/test/test_imap.py
```diff
@@ -296,6 +296,86 @@ def cbProduced(result):
 
 
 
+class MessageSetTests(unittest.SynchronousTestCase):
+    """
```
markrwilliams, line 300: [This commit](https://codecov.io/gh/twisted/twisted/src/9f61890d97c41065dbb5cece2bd4af986c027363/src/twisted/mail/imap4.py) decreased [the previous commit's coverage](https://codecov.io/gh/twisted/twisted/src/562496b261a71b3372c7dd8015f556d32f1285cb/src/twisted/mail/imap4.py); [Two](https://codecov.io/gh/twisted/twisted/src/9f61890d97c41065dbb5cece2bd4af986c027363/src/twisted/mail/imap4.py#L279) [lines](https://codecov.io/gh/twisted/twisted/src/9f61890d97c41065dbb5cece2bd4af986c027363/src/twisted/mail/imap4.py#L282) aren't covered because this commit no longer passes the output of [`parsedIdList` to `MessageSet`](https://github.com/twisted/twisted/pull/817/commits/9f61890d97c41065dbb5cece2bd4af986c027363?diff=unified#diff-904151008804a75ba425f52b818531c8L399), which [doesn't make sense](https://twistedmatrix.com/trac/ticket/9178).

src/twisted/mail/imap4.py
```diff
@@ -209,33 +209,70 @@ def clean(self):
...
-                h = max(oldh, h)
-                self.ranges[i - 1] = None
-                self.ranges[i] = (l, h)
 
-            oldl, oldh = l, h
+            mergedRanges[-1] = (min(previousLow, low),
+                                max(previousHigh, high))
 
-        self.ranges = [r for r in self.ranges if r]
+        self.ranges = [(None if low is self._infinity else low,
```
glyph: I can't help but notice that this fairly complex comprehension occurs twice.  Put it into a function?

src/twisted/mail/imap4.py
```diff
@@ -209,33 +209,70 @@ def clean(self):
...
+
+
+    def _noneInRanges(self):
+        """
+        Is there a L{None} in our ranges?
+
+        @return: L{True} if L{None} is in some range in ranges and
+            L{False} if otherwise.
+        """
+        # The cleanup method merges overlapping or consecutive ranges.
```
glyph: Please put this lovely comment into the docstring.

glyph: Although: what is "the cleanup" method? `L{MessageSet.cleanup}`?

src/twisted/mail/imap4.py
```diff
@@ -209,33 +209,70 @@ def clean(self):
...
+        # *), so this is the same as 2. but with a z that is greater
+        # than everything.
+        #
+        # The result is that there is a maximum of two Nones, and one
+        # of them has to be the high element in the last tuple in
+        # ranges.  That means checkin self.ranges[-1][-1] suffices
+        return self.ranges[-1][-1] is None
 
 
     def __contains__(self, value):
```
glyph: While you're in here, an `@type` on `value` would be very helpful.

glyph:
If I were reviewing this I'd probably give it a passing review.  Big jump in test coverage, comprehensible design, etc.  What do you think it still needs?

src/twisted/mail/imap4.py
```diff
@@ -127,27 +172,42 @@ def _setLast(self, value):
...
+              largest number in use.  Must be set before attempting to
+              use the MessageSet as a container.
+
+              @raises: L{ValueError} if a largest value has already
+                  been set.
+              '''
         return _getLast, _setLast, None, doc
     last = property(*last())
 
+    def _swap(self, this, that, ifIs):
```
glyph: This doesn't use `self` at all and should therefore be a top-level function, not a method.

src/twisted/mail/imap4.py
```diff
@@ -209,33 +275,76 @@ def clean(self):
         Clean ranges list, combining adjacent ranges
         """
 
-        self.ranges.sort()
+        ranges = sorted((self._swap(low, self._infinity, ifIs=None),
```
glyph: This same, fairly complex double-swap expression happens both at the top and the bottom of this function. It should be pulled out to a separate method (although I gather the `sorted` and `list` invocations can stay here.)

src/twisted/mail/imap4.py
```diff
@@ -209,33 +275,76 @@ def clean(self):
...
+                         self._swap(high, self._infinity, ifIs=None))
+                         for low, high in self.ranges)
 
-        oldl, oldh = None, None
-        for i,(l, h) in enumerate(self.ranges):
-            if l is None:
+        mergedRanges = [(float('-inf'), float('-inf'))]
```
glyph, line 302: perhaps this should be bound to a global called `_inf`?

(I also vaguely seem to remember that `1e999` is more portable than `inf` as a constant, but none of our buildbots are complaining, so…)

glyph (approved):
Overall this looks quite good.  The buildbot error on the docs builder is unfortunately real, https://buildbot.twistedmatrix.com/builders/documentation/builds/2972/steps/api-documentation/logs/new%20pydoctor%20errors but the appveyor failure is spurious (since scrolling takes forever there, traceback recorded at https://twistedmatrix.com/trac/ticket/8879#comment:8).

src/twisted/mail/imap4.py
markrwilliams: It looks like `inf`, `-inf`, and `nan` became portable [a while ago](https://bugs.python.org/issue1635):

> The float() function will now turn the string nan into an IEEE 754 Not A Number value, and +inf and -inf into positive or negative infinity. **This works on any platform with IEEE 754 semantics**. (Contributed by Christian Heimes; issue 1635.)

(from https://docs.python.org/2/whatsnew/2.6.html, emphasis added.)

So I think the infinities are safe...

markrwilliams: I left these as literals because they're used only in this function as the base case for the loop that follows, whereas `MessageSet._infinity`'s unique identity allows `_swap` to reliably identify it.

PR #822, opened 2017-06-16
Use serial.serialutil.to_bytes() in twisted.internet._win32serialport

altendky:
In Python 3 I do not expect serial bytes to be a `str`, let alone an
`object.__repr__()` string.  Example data received looked like this.

Python 2: `<type 'str'>, "'\\x01'"`

Python 3: `<class 'str'> '<memory at 0x048285E0>'`

https://twistedmatrix.com/trac/ticket/9186


wsanchez (dismissed):
Looks correct, but needs tests.

hawkowl:
@altendky Your other PR is merged now, so it should unblock this one. :)

altendky:
@wsanchez not sure if you will need to resolve your change request or not but there are tests now.

PR #823, opened 2017-06-19
Added pyserial 3.x support on Windows

altendky:
https://twistedmatrix.com/trac/ticket/8159


altendky:
~~Unintended push to wrong branch.~~

glyph:
(Code review is over on Trac - https://twistedmatrix.com/trac/ticket/8159#comment:14)

src/twisted/internet/_win32serialport.py
```diff
@@ -56,40 +59,54 @@ def __init__(self, protocol, deviceNameOrPortNumber, reactor,
         self._finishPortSetup()
 
 
+    @property
+    def _serialHandle(self):
+        # Try to detect pyserial version to support internal API changes
+        if hasattr(self._serial, '_port_handle'):
```
exarkun: `hasattr` can mask unexpected exceptions.  I'd suggest something like the following here:

```
serial = self._serial
try:
    # >= 3.0
    return serial._port_handle 
except AttributeError:
    # <= 2.7
    return serial.hComPort
```

src/twisted/internet/test/test_win32serialport.py
```diff
@@ -0,0 +1,137 @@
...
+        # Validate kwargs
+        kwargs = port._serial.captured_kwargs
+        self.assertEqual(9600,                kwargs["baudrate"])
+        self.assertEqual(serial.EIGHTBITS,    kwargs["bytesize"])
+        self.assertEqual(serial.PARITY_NONE,  kwargs["parity"])
+        self.assertEqual(serial.STOPBITS_ONE, kwargs["stopbits"])
+        self.assertEqual(0,                   kwargs["xonxoff"])
+        self.assertEqual(0,                   kwargs["rtscts"])
+        self.assertEqual(None,                kwargs["timeout"])
+        port.connectionLost(reason='Cleanup')
```
exarkun: `reason` should be a `Failure`.  The cheapest technically-correct way to do this is `connectionLost(Failure(Exception("Cleanup")))`.

altendky:
@hawkowl, I believe I made the changes you wanted.  Now requires pySerial >= 3.0 in _setup.py and on Windows only supports 3.x.  That is a point to be made, I am only aware of this incompatibility between 2.x and 3.x affecting Windows so the >=3.0 may be more strict than it needs to be on other platforms.

glyph:
Thank you @altendky for keeping up with all these review changes!

PR #824, opened 2017-06-20
Handle the HTTP * request URI.

Lukasa:
See https://twistedmatrix.com/trac/ticket/9190


src/twisted/web/server.py
```diff
@@ -500,6 +505,27 @@ def getRootURL(self):
...
+        Rather than bother the user with this, we simply fast-path it back to
+        an empty 200 OK. Any non-OPTIONS verb gets a 405 Method Not Allowed
+        telling the client they can only use OPTIONS.
+        """
+        if self.method == b'OPTIONS':
+            self.setResponseCode(http.OK)
+        else:
+            self.setResponseCode(http.NOT_ALLOWED)
+            self.setHeader(b'Allow', b'OPTIONS')
+
```
wsanchez, line 538: You sure this is right?

I suspect that `*` is not a valid resource name (outside of an `OPTIONS` request), in which case `BAD_REQUEST` seems more appropriate.  If it's a valid resource name, then the server should get a crack at handling the request and not get short-circuited, no?

Lukasa: I think we're squarely in the realm of interpretation here.

My reading of 7231 is that we could go either to 400 or 405. My interpretation is that, given that the request is sensibly parseable and understandable, 400 is not really quite right. But I don't believe this strongly: I almost wrote the patch as a 400.

Regardless, I don't really think the user should get a look. It's never something to which you can sensibly send a response to.

So I guess what I'm saying is that I can be talked around to 400 pretty easily.

wsanchez: I dunno… 400 seems pretty correct:

> something that is perceived to be  a client error

`*` as a URL is clearly a client error.

As to whether the user should get a look, I have to disagree.  This is rare enough to fall into the "almost certainly don't care" category, but an application should always have a hook into handling the request if they want to.

A lot of applications choose to explain the error in the body with some HTML or text, for example.  Perhaps you want to redirect bad requests to [YouTube](https://www.youtube.com/watch?v=dQw4w9WgXcQ).  Or you're that guy that sends non-ASCII header names and also use `*` as a magic thing.  You know.

wsanchez: A 405, on the other hand is for

> method is not supported for the requested resource

I'd argue that no resource has been specified here, because your client is stupid and sent an invalid request. Bad, bad request!

wsanchez (approved):
All that said, this is clearly an improvement.

I'd encourage the change to 400, but if you're pretty sure 405 is better, I'm still cool with merging.

src/twisted/web/server.py
Lukasa: I'm certainly willing to go over to 400. I do have a question about how we should allow the user to get a look though: what API would you want for this?

PR #825, opened 2017-06-21
Omit default content-type for responses with no body.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/9191

### DIFF
--- a/src/twisted/web/newsfragments/9191.bugfix
+++ b/src/twisted/web/newsfragments/9191.bugfix
@@ -1,0 +1,1 @@
+twisted.web.server.Request and all web servers that use it now no longer send a default Content-Type header on responses that do not have a body (i.e. that set Content-Length: 0 or that send a 204 status code).
--- a/src/twisted/web/server.py
+++ b/src/twisted/web/server.py
@@ -194,11 +194,20 @@ class Request(Copyable, http.Request, components.Componentized):
         """
         if not self.startedWriting:
             # Before doing the first write, check to see if a default
-            # Content-Type header should be supplied.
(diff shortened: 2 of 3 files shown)


wsanchez (approved):

glyph:
Hmm. Why didn't codecov have an opinion on this?

Lukasa:
Maybe because a builder failed?

glyph:
Wait do *all* the builders have to succeed for coverage to be reported?

Lukasa:
Not sure. It certainly *can* be required.

rodrigc:
I think @adiroiban knows more about this since he set up most of the codecov integration.

For example, I think if a project has 6 submissions to codecov within a short time interval,
and then the next time codecov receives 4 submissions, then codecov will wait until it receives the "missing" codecov submissions before generating the final coverage report.  This isn't always consistent.

adiroiban:
See https://docs.codecov.io/docs/merging-reports#section-how-does-codecov-know-when-to-send-notifications-

From my experience, for projects with big coverage files, the observed behaviour is not the one described in the documentation.

I think that when sending reports of over 3 MB the codecov workers might crash and you will never see a report.

glyph:
I wonder if mentioning @codecov here might help. Really? 3MB coverage reports just crash it?

glyph:
So, `buildbot/` statuses are never reported as "pending", which might explain some of our flakiness in terms of what gets reported.

adiroiban:
Good observation. There might be an older version in use... or some other errors.
I am also using buildbot (latest eight branch) and the pending status is reported as soon as the build starts

PR #826, opened 2017-06-23
761-signal-exit

jlitzingerdev:
Pull request for https://twistedmatrix.com/trac/ticket/761


jlitzingerdev:
Doh.  That missed coverage was a stupid copy/paste error on my part in the test case.  Fixing.

jlitzingerdev:
At this point, assuming I properly interpreted the results of codecov (still learning the tool), it looks like the only missed lines in the diff are those that were skipped (e.g. the sigBreak path).  If desired I can easily fake that path, but wasn't sure what the appropriate action was considering that signal will/will not exist on all platforms.

src/twisted/application/app.py
```diff
@@ -681,3 +690,17 @@ def startApplication(application, save):
         reactor.addSystemEventTrigger('after', 'shutdown', p.save, 'shutdown')
     reactor.addSystemEventTrigger('before', 'shutdown',
                                   service.IService(application).stopService)
+
+
+
+def exitWithSignal(sig):
```
glyph: I don't think this needs to be public API.  We should be conservative about exporting new public API, preferring to go with private APIs at first and exposing them only deliberately.

src/twisted/application/app.py
```diff
@@ -681,3 +690,17 @@ def startApplication(application, save):
...
+    """
+    Force the application to terminate with the specified signal by replacing
+    the signal handler with the default and sending the signal to ourselves.
+
+    @param sig:  Signal to use to terminate the process with C{os.kill}.
+    @type sig:  C{int}
+
+    """
+    signal.signal(sig, signal.SIG_DFL)
```
glyph: Rather than mutating global signal state, can't this just be accomplished with `os._exit(0b10000000 | sig)`?

src/twisted/application/twist/test/test_twist.py
```diff
@@ -201,3 +201,102 @@ def run(self):
             )
         )
         self.assertEqual(runners[0].runs, 1)
+
+
+
+class CrippledRunner(object):
```
glyph: 1. "Crippled" could be construed as insulting to disabled people, so please don't use the term in the codebase.
2. It's also not a standard technical term; I think you mean "dummy"? http://xunitpatterns.com/Mocks,%20Fakes,%20Stubs%20and%20Dummies.html

src/twisted/application/twist/test/test_twist.py
```diff
@@ -201,3 +201,102 @@ def run(self):
...
+                TwistOptions.__init__(self)
+
+
+            def installReactor(self, name):
+                reactor = MemoryReactor()
+                reactor.exitSignal = 0
+                return reactor
+
+
+        self.patch(_twist, "TwistOptions", TestTwistdOptions)
```
glyph: This level of patching suggests that this is not a good testing strategy :-\.

src/twisted/application/twist/test/test_twist.py
```diff
@@ -201,3 +201,102 @@ def run(self):
...
+        class TestTwistdOptions(TwistOptions):
+
+            def __init__(self):
+                TwistOptions.__init__(self)
+
+
+            def installReactor(self, name):
+                reactor = MemoryReactor()
+                return reactor
+
```
glyph: (nitpick: too many blank lines. I wonder why twistedchecker isn't complaining?)

src/twisted/internet/base.py
```diff
@@ -473,6 +473,8 @@ class ReactorBase(object):
     @ivar _registerAsIOThread: A flag controlling whether the reactor will
         register the thread it is running in as the I/O thread when it starts.
         If C{True}, registration will be done, otherwise it will not be.
+
+    @ivar exitSignal: The signal that cause the reactor to stop, or 0.
```
glyph: This attribute appears on some implementations of the reactor now but not others, it's never documented on an interface.

glyph (commented):
Thanks for submitting this fix!  It mostly looks good, but I do have some concerns about exposing a bunch of new public API as a result.  Could you fix this to be an implementation detail somehow without exposing it?

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: No problem here.  Btw, I used it in a similar context to where it is used elsewhere in the code (see test_twistd).

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: It does, I just missed it, will fix in v2.

src/twisted/application/app.py
jlitzingerdev: @glyph  Yes but I don't think that is a valid thing to do in all cases?.  If setting the exit status to 128 + signal was always the right thing to do on all platforms, this would (I think) work.  I actually went this route initially, but changed after reading some of the referenced articles in the original ticket.  My concern is that, while the GNU C library certainly obeys the above behavior, I haven't found any spec/guarantee to indicate this would be consistent with other implementations.  Thus, rather than assume it is ok, I chose to let the OS do what it would normally do had the signal not been caught in the first place.  Comment 8 in the ticket, and the linked article talk a little about calling exit.

In approaching this bug I looked for some consistent, authoritive reference on valid exit status', and realized it is somewhat of a platform/implementation specific mess.  Therefore, lacking a better alternative, I wanted to let the OS take control and do what it would normally do.

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: Fair,  I have an alternate idea, but I'll raise questions in the ticket.

src/twisted/internet/base.py
jlitzingerdev: This is one of the bigger challenges I had, will discuss in the ticket.

src/twisted/application/app.py
exarkun: > My concern is that, while the GNU C library certainly obeys the above behavior, I haven't found any spec/guarantee to indicate this would be consistent with other implementations. 

Correct.  There is no POSIX API for _constructing_ an exit code - only deconstructing it.  You probe the system to create some  values and then inspect them to see how the system composes the two pieces of information and then compose them yourself that way ... but ...  you could also just re-deliver the signal. :)

src/twisted/application/twist/test/test_twist.py
```diff
@@ -201,3 +201,102 @@ def run(self):
...
+class TwistExitTests(twisted.trial.unittest.TestCase):
+
+    def setUp(self):
+        self.exitWithSignalCalled = False
+
+        def startService(service):
+            pass
+
+        self.patch(MultiService, "startService", startService)
```
exarkun: This looks kind of sad.  What's going on?  I suggest refactoring something so you don't have to do this.

jlitzingerdev: Thanks!
The intent was to let Twist invoke the plugin/service machinery normally, but prevent the service from actually starting.  I took the patching approach because it was already used in the test module to achieve the same goal (as well as additional goals specific to that test).  I'm fairly certain it isn't the only way to test these changes without running the service so yeah, agree, I'll refactor.

jlitzingerdev: Actually, changing the plugin called (e.g. to words) would also eliminate the need for that patching as well.  When the web plugin's makeService executes it calls strports.service, which imports the default reactor if one isn't passed in and results in the reactory being dirty when the test exits.

Regardless, I'd prefer something a little more deterministic than relying on the current implementation of whatever plugin is chosen for the test.

src/twisted/application/app.py
```diff
@@ -392,8 +395,14 @@ def startReactor(self, reactor, oldstdout, oldstderr):
 
         @see: L{runReactorWithLogging}
         """
+        if reactor is None:
+            from twisted.internet import reactor
         runReactorWithLogging(
             self.config, oldstdout, oldstderr, self.profiler, reactor)
+        try:
+            self._exitSignal = reactor._exitSignal
+        except AttributeError:
```
glyph: I think we should be _extremely_ cagey about `AttributeError`s, especially in the presence of an `except: pass`.

For example, imagine a future scenario in which someone adds `__slots__` to ApplicationRunner, but forgets to add this attribute.  Of course, in *this* case, a test would nicely fail and explain what had gone wrong, but if someone cribs from this example they'll produce this error elsewhere.

So I'd rewrite this

```python
try:
    exitSignal = reactor._exitSignal
except:
    exitSignal = None
self._exitSignal = exitSignal
```

and then you can also eliminate the redundant cvar.

src/twisted/application/app.py
```diff
@@ -681,3 +690,17 @@ def startApplication(application, save):
...
+
+
+def _exitWithSignal(sig):
+    """
+    Force the application to terminate with the specified signal by replacing
+    the signal handler with the default and sending the signal to ourselves.
+
+    @param sig:  Signal to use to terminate the process with C{os.kill}.
+    @type sig:  C{int}
+
```
glyph: Just as a stylistic note: this blank line really shouldn't be here.

src/twisted/application/twist/_twist.py
```diff
@@ -99,6 +100,12 @@ def run(twistOptions):
             fileLogObserverFactory=twistOptions["fileLogObserverFactory"],
         )
         runner.run()
+        reactor = twistOptions["reactor"]
+        try:
+            if reactor._exitSignal:
+                _exitWithSignal(reactor._exitSignal)
+        except:
+            pass
```
glyph: `except: pass` is pretty much always wrong; it makes debugging impossible.  What exceptions are you actually expecting here?  Why is it important to catch them?

src/twisted/application/twist/test/test_twist.py
```diff
@@ -191,3 +191,80 @@ def run(self):
...
+        def startLogging(_):
+            """
+            Prevent Runner from adding new log observers or other
+            tests outside this module will fail.
+
+            @param _: Unused self param
+            """
+            pass
+
+        self.patch(Runner, 'startLogging', startLogging)
```
glyph: `patch` on a global object is always a form of technical debt.

Is there any possible way you can inject a `Runner` instance into `Twist`?  This could be a follow-up if you don't want to set up this plumbing in this PR, I just don't want to lose track of it.

src/twisted/application/app.py
glyph: Furthermore…

Given that `_exitSignal` is a private attribute, it's always possible that some 3rd-party reactor is using it.

As such, an unambiguous way to signal [namespace ownership](https://glyph.twistedmatrix.com/2017/05/the-sororicide-antipattern.html) would be helpful here.  Interfaces can do this.

Basically, you make a private interface, `_ISupportsExitSignaling`, make every reactor in Twisted that uses this functionality provide it, and then here you can do `if _ISupportsExitSignaling.providedBy(reactor)` before accessing this attribute.  This is unambiguous, and makes it easier for individual reactors to document what they're doing when they provide this attribute.  How does this sound?

src/twisted/internet/test/test_base.py
```diff
@@ -369,3 +369,65 @@ def test_str(self):
...
+    """
+    Subclass of ReactorBase to capture signals delivered to the
+    reactor for inspection.
+    """
+
+    def installWaker(self):
+        """
+        Required method, unused.
+        """
+        pass
```
glyph: `pass` is not necessary here, the docstring is sufficient.  (Sorry, twistedchecker ought to check for this.)

src/twisted/test/test_twistd.py
```diff
@@ -1911,5 +1956,83 @@ def test_hooksNotCalled(self):
 
 
 
+class ExitWithSignalTests(unittest.TestCase):
+
```
glyph: There should not be a blank line here.

src/twisted/test/test_twistd.py
```diff
@@ -1911,5 +1956,83 @@ def test_hooksNotCalled(self):
...
+        """
+        CrippledApplicationRunner._exitSignal = 2
+        self.patch(twistd.app, '_exitWithSignal', self.fakeExitWithSignal)
+        self.patch(twistd, '_SomeApplicationRunner', CrippledApplicationRunner)
+        twistd.runApp(self.config)
+        self.assertTrue(self.exitWithSignalCalled)
+
+
+    def tearDown(self):
+        CrippledApplicationRunner._exitSignal = 0
```
glyph: Is this expected to be set by the system under test?  If not, it would be more appropriate to use `.patch` for this.

src/twisted/test/test_twistd.py
```diff
@@ -1911,5 +1956,83 @@ def test_hooksNotCalled(self):
...
+        self.assertEquals(fakeKillArgs[0], os.getpid())
+        self.assertEquals(fakeKillArgs[1], signal.SIGINT)
+
+
+    def test_normalExit(self):
+        """
+        _exitWithSignal is not called if the runner does not exit with a
+        signal.
+        """
+        self.patch(twistd.app, '_exitWithSignal', self.fakeExitWithSignal)
```
glyph: This is still an awful lot of patching.  Is there any way you can think of to reduce it in the future?

glyph (commented):
I think this is definitely approaching mergeable, but I'd like you to at least consider the remaining patches and the `_exitSignal` namespace confusion issue.  If you think it would be too much work to address those we can probably merge it as-is with some follow-up tickets.

src/twisted/application/twist/_twist.py
jlitzingerdev: In general, I agree, and I really didn't like it when I added it, though I did have a purpose in mind.  I'll explain why, and if you feel it isn't necessary I'm more than happy to remove it.

The purpose here was to preserve the existing exit semantics of twist in the event _exitWithSignal throws an exception.  While that shouldn't happen (technically, I'm not confident that os.kill wont ever throw on other platforms, I'm primarily familiar with POSIX and Linux in particular), my inability to fully verify it can't happen led me to add the catch-all.

Specifically, if an exception gets thrown, I assume it'll go up to the interpreter, and from there may or may not affect the exit status of the program.  Rather than have this new code trigger that, I caught the exception and let the program exit as it does today.

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: If memory serves, we don't even have to inject a dummy instance.  I did an interation with a Dummy log factory and the tests will work just fine.  The problem was that downstream logging specific tests failed.  I inferred, from a quick glance at the failure, that those tests were expecting only a certain number of global observers.  We can likely avoid that by manually mutating the global observers if that is possible.  Let me redo it that way and see exactly what those errors were.

src/twisted/test/test_twistd.py
jlitzingerdev: I didn't make heavy changes in v2 to the test_twistd, so I'm going to have to say, not sure?  The goal here is to make sure _exitWithSignal is called.  Calling the real function would deliver a signal to the test, which would be bad, so it seems like some patching is required with the current implementation.  I could push the patching back to `os.kill()`, which might loosely be considered a platform integration point where patching is more appropriate, thoughts?

Not saying that's the only way, just the immediately obvious one.  I'll take a look.

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: Specifically, one of the errors when a dummy logger instance is passed is:
```
twisted.trial.unittest.FailTest: 4 != 1

twisted.test.test_log.LogTests.test_erroneousErrors
```
I'll look into what the failing test are expecting, but I suspect removing the created observers in my test will do the trick.

src/twisted/internet/test/test_base.py
jlitzingerdev: I was thinking about trying to add the ability to check a patchfile (similar to the Linux kernel's checkpatch utility) to twistedchecker.  The benefit there being that it would potentially ignore previous errors in other unrelated parts of the file.  If I go down that road I'll look at adding this too if nobody beats me to it. 

Will correct in V3.

src/twisted/test/test_twistd.py
jlitzingerdev: Should twistedchecker flag this?  It didn't for me (though I did miss another "too many blank lines" warning later in the class.

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: Actually, no, that's not at all what is happening.  I'll look at either using the existing runner or injecting one.

src/twisted/application/app.py
jlitzingerdev: I like this, it's much clearer, so yes, I'm for this implementation.

src/twisted/test/test_twistd.py
jlitzingerdev: The real need is to inject an input into the SUT, so a Stub ApplicationRunner is likely more appropriate.

src/twisted/application/twist/test/test_twist.py
jlitzingerdev: @glyph I've looked in to this a bit, and there aren't many good options with the existing code.  Runner is imported into _twist, it isn't an option that can be easily injected.  The default runner can be used, but somewhere the call to `globalLogBegginer.beginLoggingTo` needs to be prevented or downstream tests fail.

WIth the existing code, the following seem like the options:
1.  Leave it as is.
2.  Patch the Runner itself, as is done in other tests.
3.  Patch globalLogBeginner in twisted.application.runner._runner with a new LogBeginner for each test.
4.  Patch 'beginLoggingTo' on globalLogBeginner as is done in test_twistd.

Options 3 and 4 allow the use of the default runner.

To inject a runner, the runner should probably be sourced from TwistOptions, rather than imported globally, and be selectable.  While I am more than willing to do this change, it does feel like it is significant enough to warrant its own ticket.  Specfiically, it seems like a fairly heavy side effect for an unrelated change.

Thoughts?

jlitzingerdev:
@markrwilliams @rodrigc Is the Appveyor failure something related to my PR specifially?  If I read it correctly it is complaining about an openssh install, so it doesn't look like it, but if there's something I should look at let me know, otherwise I'll try to keep merging every day or so.

rodrigc:
See:
 https://twistedmatrix.com/pipermail/twisted-python/2017-August/031596.html

jlitzingerdev:
Thanks, I'll keep an eye for resolution and keep up to date accordingly.

PR #757, opened 2017-03-29
[#9098] Add source-specific multicast support

falanger:

PR #827, opened 2017-06-26
[#9098] Add source-specific multicast support

jpschaumloeffel:
Picked up an (abandoned?), still unmerged pull request (https://github.com/twisted/twisted/pull/757) that was not merged due to a small whitespace error.

I'd simply like to get that PR merged, so I started by fixing the whitespace problem.
Will also try to address additional issues with this, if the need arises.


src/twisted/internet/udp.py
```diff
@@ -504,9 +516,33 @@ def _joinAddr2(self, interface, addr, join):
...
+        addr = socket.inet_aton(addr)
+        interface = socket.inet_aton(interface)
+        source = socket.inet_aton(source)
+        if join:
+            cmd = socket.IP_ADD_SOURCE_MEMBERSHIP
+        else:
+            cmd = socket.IP_DROP_SOURCE_MEMBERSHIP
+        try:
+            self.socket.setsockopt(socket.IPPROTO_IP, cmd, addr + interface + source)
+        except socket.error as e:
```
rodrigc, line 534: This line is never touched during the tests.  Can you add a test which somehow triggers this socket.error exception
so that we can test the error handling part of this code?

src/twisted/internet/udp.py
```diff
@@ -484,14 +484,26 @@ def setTTL(self, ttl):
...
     def joinGroup(self, addr, interface=""):
         """Join a multicast group. Returns Deferred of success."""
-        return self.reactor.resolve(addr).addCallback(self._joinAddr1, interface, 1)
+        return self.reactor.resolve(addr).addCallback(self._joinAddr1, "", interface, 1)
 
 
-    def _joinAddr1(self, addr, interface, join):
-        return self.reactor.resolve(interface).addCallback(self._joinAddr2, addr, join)
+    def joinSourceGroup(self, addr, source, interface=""):
+        """Join a source-specific multicast group. Returns Deferred of success."""
```
rodrigc, line 491: The twistedchecker script prefers that docstring comments not be on one line, so this should be: should be:

```
"""
Join a source-specific multicast group. Returns Deferred of success.
"""
```

src/twisted/internet/udp.py
```diff
@@ -504,9 +516,33 @@ def _joinAddr2(self, interface, addr, join):
...
+
+
     def leaveGroup(self, addr, interface=""):
         """Leave multicast group, return Deferred of success."""
-        return self.reactor.resolve(addr).addCallback(self._joinAddr1, interface, 0)
+        return self.reactor.resolve(addr).addCallback(self._joinAddr1, "", interface, 0)
+
+
+    def leaveSourceGroup(self, addr, source, interface=""):
+        """Leave multicast group, return Deferred of success."""
```
rodrigc, line 544: See other feedback about docstring

src/twisted/test/test_udp.py
```diff
@@ -632,20 +645,58 @@ def cbJoined(ignored):
...
             result.addCallback(lambda ign: passthrough)
             return result
+
         joined.addCallback(cleanup)
 
         return joined
 
+    def test_multicastSourceSpecific(self):
+        """
+        Test that a source specific multicast group can be joined and messages sent to and
```
rodrigc, line 667: Remove *Test that a* from docstring.  Twisted prefers to follow docstring conventions for tests https://jml.io/pages/test-docstrings.html

rodrigc (changes requested):
These tests are failing on Windows:

```
[ERROR]
Failure: twisted.internet.error.MulticastJoinError: ('\xe1\x00\x00\xfa', '\x7f\x00\x00\x01', '\x00\x00\x00\x00', 10042, 'An unknown, invalid, or unsupported option or level was specified in a getsockopt or setsockopt call')
```

Source Specific Multicast is supported on Windows:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738586(v=vs.85).aspx

 so this should be fixed.  This is the structure to be used:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738704(v=vs.85).aspx

src/twisted/internet/udp.py
```diff
@@ -504,9 +516,33 @@ def _joinAddr2(self, interface, addr, join):
...
+
+        addr = socket.inet_aton(addr)
+        interface = socket.inet_aton(interface)
+        source = socket.inet_aton(source)
+        if join:
+            cmd = socket.IP_ADD_SOURCE_MEMBERSHIP
+        else:
+            cmd = socket.IP_DROP_SOURCE_MEMBERSHIP
+        try:
+            self.socket.setsockopt(socket.IPPROTO_IP, cmd, addr + interface + source)
```
rodrigc, line 533: This socket option doesn't look right.  *addr + interface + source* is not the correct order for *struct mreq* which is what this parameter takes.  There are other examples out there which show how to use struct.pack for correctly packing the structure to be used in this socket option.  Here is one:

https://stackoverflow.com/questions/603852/multicast-in-python

rodrigc (changes requested):

src/twisted/test/test_udp.py
wiml:
```suggestion
        A source specific multicast group can be joined and messages sent to and
```

PR #829, opened 2017-06-30
Handle coroutines in _flattenElement

wsanchez:
https://twistedmatrix.com/trac/ticket/9199

---
This change is [Reviewable](https://reviewable.io/reviews/twisted/twisted/829)


glyph (approved):
Thanks.  As we move into the post-async/await future it is nice to have all these things starting to work :-).

I think we can land this but it needs some follow-ups.

Before it can land:

- The NEWS file should really clarify that it can only support coroutines *that yield Deferreds* right now, and that e.g. "normal" asyncio coroutines will not work.
- the test case should really add an `await` case to keep the documentation honest.  (This is "in addition to" the current test case, not "instead of" - having the just-`return` case is interesting as well.)

Afterwards, it also needs:

- we should have a separate issue for enhancing this to support Future-awaiting coroutines.  (This is an issue on this code and not just in general, because `ensureDeferred` probably isn't the right API to magically Future-ify things, so we might need to change this.)
- in the meanwhile, `ensureDeferred` really needs to emit a more intelligible error message if you screw this up.
- the narrative documentation needs to provide an example of using a coroutine, especially since "Deferred coroutine" is a slightly odd creature at this point.

Please land after addressing and filing some tickets for the next steps.

PR #830, opened 2017-06-30
Make sure we pass unicode, not bytes to CreateProcess()

rodrigc:
https://twistedmatrix.com/trac/ticket/9200


opalmer (commented):
Looks good :+1:  

On Windows file system paths are Unicode by default so it's a bit redundant to do `.decode(sys.getfilesystemencoding())`. The docs for Python 2 mention that getfilesystemencoding() is mainly intended for string -> byte conversion and for Python 3 it apparently changes to always be 'utf-8'. But based on the docs it shouldn't hurt either so I'd leave it as you have it in case there are edge cases out there. 

References:
https://docs.python.org/2/library/sys.html#sys.getfilesystemencoding
https://docs.python.org/3/library/sys.html#sys.getfilesystemencoding
https://www.python.org/dev/peps/pep-0529/
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247.aspx

rodrigc:
@altendky Thanks for https://github.com/twisted/twisted/pull/831 , it boosted the coverage of my PR from 95% to 100%

glyph:
Thank you @altendky !

opalmer (approved):
Sorry, forgot to hit approve on my side.

glyph:
What happened to break it again :-( it looks like 95% now.

rodrigc:
If I try to look at https://codecov.io/gh/twisted/twisted/pull/830 , I get a 500 error

rodrigc:
I see 100% patch coverage here: https://codecov.io/gh/twisted/twisted/pull/830

glyph:
Thanks for investigating @rodrigc.

Hey @codecov - any idea why we keep getting PRs like this where the comment and the build status don't get updated?

PR #831, opened 2017-06-30
Pin codecov==2.0.5 until fixed

altendky:
https://twistedmatrix.com/trac/ticket/9196
https://github.com/codecov/support/issues/402


rodrigc:
Wow, thanks for tracking this down.  I've recently been working on https://github.com/twisted/twisted/pull/830 and was wondering why codecov was not getting results from the appveyor builds.

altendky:
Yeah, we'll see.  I don't know if this is the sort of thing that gets accepted but that's what the codecov/support#402 original reporter did with conda.  :[

rodrigc (approved):
You definitely fixed the problem.

Without your fix: https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-3370/job/n4khorbeg540wo7m
fails to upload to codecov.

With your fix: https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-3371/job/1exsed6g7bc5g0ig
succeds upload to codecov.

Thanks for tracking this down and submitting the fix!!

altendky:
Sweet, I broke some CGI tests... on one system...  That should be a fun one for me to track down next week.  :[  I was getting so excited to get back to my actual bug fixes.

rodrigc:
You didn't break those tests, looks like there is some instability on that builder.

rodrigc:
Thanks for the patch, you are now at http://twistedmatrix.com/highscores/ !

glyph:
Why don't I see a codecov status on this build? I was all excited to see the coverage go up.

Do we have to wait until there's a baseline on `trunk` to compare against?

Normally I'd suggest pinning with `!=` rather than `==` for a dependency like this which is a client for a service with their own API, but it looks like several releases have had the issue now...

altendky:
I assume no coverage because nothing that gets coverage checked was changed? And yes, to releases I believe. At least it seems pretty simpler for them to fix unless there was a good reason to disable AppVeyor.

PR #833, opened 2017-07-01
8692 notify finish fast

glyph:
Fix for https://twistedmatrix.com/trac/ticket/8692


src/twisted/web/http.py
```diff
@@ -1784,7 +1784,7 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
     A receiver for HTTP requests.
 
     The L{HTTPChannel} provides L{interfaces.ITransport} and
-    L{interfaces.IConsumer} to the L{Request} objects it creates. It also
+    L{interfaces.IConsumer} to the L{Request} objects it creates.  It also
```
glyph, line 1816: Please pardon the irrelevant whitespace changes here; this was *almost* a https://github.com/glyph/python-docstring-mode -compliant docstring, and I re-wrapped it; it's like someone went through and just deleted the double-spacing after it had been already reflowed.  (Maybe an earlier version had a bug?)

Lukasa (approved):
This looks reasonable to me. 

Sorry about introducing the original regression. 🙁

src/twisted/web/http.py
```diff
@@ -2048,22 +2071,30 @@ def allContentReceived(self):
...
 
 
-    def rawDataReceived(self, data):
-        self.resetTimeout()
-
+    def dataReceived(self, data):
+        """
+        Data was received from the network.  Process it.
+        """
         # If we're currently handling a request, buffer this data. We shouldn't
```
mithrandi: This comment seems to be seriously misleading now, could you update it to describe the new behaviour? (We have _not_ necessarily paused the transport yet)

src/twisted/web/http.py
```diff
@@ -2048,22 +2071,30 @@ def allContentReceived(self):
...
         # have received it (we've paused the transport), but let's be cautious.
```
njsmith: This comment is no longer quite true.

glyph:
> This looks reasonable to me.

Doesn't count unless it's in the doobleydoo - https://twistedmatrix.com/trac/ticket/8692

mithrandi (approved):
Could you expand IRequest.notifyFinish's docstring to mention that pausing the channel from the application side will render it ineffective until the channel is resumed? While this isn't technically a new issue at all, it seems to fit with the theme of this branch.

I think this branch is ready to merge after the two issues mentioned are resolved, so I'm approving.

glyph:
> Sorry about introducing the original regression. 🙁

No worries. HTTP/2 was a massive, risky project, and I'm very glad we undertook it.  The nature of risk is sometimes there are downsides.  I think we have generally learned a lot, and the pace of these has slowed down somewhat :-).

The regression was unfortunate, but as I noted on the trac ticket, the implementation was considerably more maintainable (and, apparently, reviewable!) than it would have been in the past.  I strongly suspect it has somewhat higher test coverage percentage these days as well.

glyph:
Thanks @mithrandi - will do!

glyph:
My addressing of the review feedback is complete, so if anyone comes by and sees a passing test status, feel free to merge it.

src/twisted/web/http.py
```diff
@@ -1875,6 +1933,7 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
     _requestProducerStreaming = None
     _waitingForTransport = False
     _abortingCall = None
+    _optimisticEagerReadSize = 0x4000
```
hawkowl, line 1936: how big is this

glyph: 0b100000000000000

hawkowl: omg glyph

glyph: not sure how else to put it. ord('@') << 0o10?

hawkowl: G L Y P H

glyph: oh wait sorry did you want something in *decimal*?

hawkowl: I may be a millennial but that doesn't mean I can read hexadecimal

glyph: 50735 ^ 34351

glyph:
hello codecov

untitled

src/twisted/web/http.py
hawkowl: glyph why

njsmith:
Arguably the best value for the trigger size would just be 1 (or `0x1` or `0o1` or `0xcafe ** 0x10000 % 65537` or not written at all because you can just unconditionally call `pauseProducing` from `dataReceived`), i.e., all non-pipelining clients get good behavior and all pipelining clients get shut down.

glyph:
> Arguably the best value for the trigger size would just be 1 (or 0x1 or 0o1 or 0xcafe ** 0x10000 % 65537 or not written at all because you can just unconditionally call pauseProducing from dataReceived), i.e., all non-pipelining clients get good behavior and all pipelining clients get shut down.

We should probably open a new ticket in Trac to discuss.

njsmith:
[ugh fine](https://twistedmatrix.com/trac/ticket/9203)

PR #834, opened 2017-07-03
Correctly fire notifyFinish with HTTP/2 errors.

Lukasa:
See: https://twistedmatrix.com/trac/ticket/9208

### DIFF
--- a/src/twisted/web/_http2.py
+++ b/src/twisted/web/_http2.py
@@ -32,6 +32,7 @@ import h2.events
 import h2.exceptions
 
 from twisted.internet.defer import Deferred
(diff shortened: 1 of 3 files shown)


Lukasa:
Hrm, not sure what codecov is mad about yet, but I bet it's my callbacks that contain nothing more than `self.fail`.

glyph:
Codecov is routinely giving me 500s for this past week or so.  If you could look at coverage yourself I'd appreciate it, but I'm going to remove the required status because frankly it's just not producing trustworthy numbers right now.

glyph:
I think codecov is just broken (still 500ing on this PR) and this LGTM.

PR #836, opened 2017-07-12
Join key and cert .pem files with a newline between

altendky:
https://twistedmatrix.com/trac/ticket/7530


rodrigc:
`twisted.protocols.test.test_tls.TLSMemoryBIOTests.test_getPeerCertificate` is failing after this change.  Can you fix it?

rodrigc:
@alex can you review this?

src/twisted/newsfragments/7530.bugfix
```diff
@@ -0,0 +1 @@
+Allow for key and cert .pem files that do not have a a trailing newline
```
glyph: This is slightly inaccurate because `ssl.PrivateCertificate.loadPEM` still requires them; the requirement is just omitted in one place (`ssl:` endpoints) and not everywhere.

Perhaps you should just fix this issue inside `loadPEM` so that it doesn't have to be a more complex and restricted NEWS entry?

altendky: Unless we change crypto/openssl to handle the two keys with no newline between or parse it apart ourself right before passing it to crypto, it seems we do need the fix where it is.  But, I'll agree that may well not fix all cases.  I'm digging a bit to see if i can find, test, and fix the others.

altendky: My understanding is that the issue was not about lacking a newline at the end of a standalone file per se but about how we go about merging two files into one when their paths are provided in an `ssl:` endpoint.  Previously we created a combined end/start line like this.

    -----END PRIVATE KEY----------BEGIN CERTIFICATE-----

I added a test loading just the `cert.pem.no_trailing_newline` file via `ssl.Certificate.loadPEM()` and it worked fine.  \*crosses fingers that the build servers say the same thing\*

Did I miss the issue you are pointing out?  If someone provides a (I presume) malformed file with the line above instead of two separate lines wouldn't that be their problem to resolve?  We are only responsible for not creating an invalid file by not creating the bogus line above.  That's what the patch does.

If I have to I suppose I can dive into the RFCs for PEM files.

glyph: I guess I just misunderstood the change :).  But my original point still stands: this change is to a specific area of Twisted (the `ssl:` endpoint, I believe) but the changelog entry here is phrased as if it's universal (i.e. as if nothing that previously parsed a "key" or "cert" PEM file would work if it didn't have a trailing newline).

Since you've correctly explained the change here (the bug doesn't exist elsewhere), please just update the news fragment to be clearer.  The standard for writing them does say to use the FQPN to identify the affected code, although here the endpoint prefix is probably what users really want to know about.

glyph:
`$ ./admin/pr_as_branch 836 7530 altendky-do_not_require_trailing_newline_for_cert` run.

src/twisted/newsfragments/7530.bugfix
```diff
@@ -0,0 +1 @@
+Made `ssl:` endpoints accept key.pem files without trailing newlines
```
markrwilliams: This actually affects the file specified by the `certKey`.  Could you change this to something like:

> The `ssl:` endpoint now accept `certKey` PEM files without trailing newlines.

markrwilliams (approved):
Appveyor failed because I changed the branch name too quickly.  Sorry!

One small request for clarification in the newsfragment.  Please update this so we can merge it.

Thanks!

PR #838, opened 2017-07-17
Fix test_ident on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9221


rodrigc:
@tehasdf can you review this?

tehasdf (approved):
LGTM - the `0:1` change takes advantage of the fact that on python 3, just indexing into a bytes returns an int, but slicing returns a bytes - same as on python 2.

BTW @rodrigc - you know i'm not a member here or anything, so not sure why call me ;P

rodrigc:
@tehasdf Thanks for the review.  According to https://twistedmatrix.com/trac/wiki/ReviewProcess#Whocanreview , a Twisted non-committer (you) is allowed to review the change of a Twisted committer (me).

PR #839, opened 2017-07-18
Fix twisted.web.test_domhelpers on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9222


rodrigc:
@tehasdf since you helped with https://github.com/twisted/twisted/pull/842, I was wondering if you could please code-review this Python 3 porting patch?

src/twisted/web/test/test_domhelpers.py
```diff
@@ -122,15 +122,16 @@ def test_clearNode(self):
 
     def test_get(self):
         doc1 = self.dom.parseString('<a><b id="bar"/><c class="foo"/></a>')
+        doc = self.dom.Document()
         node=domhelpers.get(doc1, "foo")
         actual=node.toxml()
```
tehasdf: since you're fixing spaces around assignments, do it here as well? (and perhaps other places in the test functions)

src/twisted/web/microdom.py
```diff
@@ -811,14 +811,17 @@ def gotTagStart(self, name, attributes):
         attributes = _unescapeDict(attributes)
         namespaces = self.nsstack[-1][0]
         newspaces = {}
+        keysToDelete = []
         for k, v in attributes.items():
```
tehasdf, line 903: there's another iteration-with-del like this in this function, just below

src/twisted/web/microdom.py
```diff
@@ -298,9 +297,9 @@ def appendChild(self, child):
 
     def writexml(self, stream, indent='', addindent='', newl='', strip=0,
                  nsprefixes={}, namespace=''):
-        stream.write('<?xml version="1.0"?>' + newl)
+        stream.write(u'<?xml version="1.0"?>' + newl)
         if self.doctype:
-            stream.write("<!DOCTYPE "+self.doctype+">" + newl)
+            stream.write(u"<!DOCTYPE " + self.doctype + u">" + newl)
```
tehasdf: can we use formatting instead of concat?

src/twisted/web/sux.py
```diff
@@ -149,17 +149,19 @@ def dataReceived(self, data):
         stateTable = self._buildStateTable()
         if not self.state:
             # all UTF-16 starts with this string
-            if data.startswith('\xff\xfe'):
-                self._prepend = '\xff\xfe'
+            if data.startswith(b'\xff\xfe'):
```
tehasdf: we could merge the two `.startswith()` branches into a single `.startswith((b'\xff\xfe', b'\xfe\xff'))`

src/twisted/web/microdom.py
```diff
@@ -422,16 +421,16 @@ def cloneNode(self, deep=0, parent=None):
...
+        stream.write(u"" + self.nodeValue)
+        stream.write(u"]]>")
 
 def _genprefix():
     i = 0
     while True:
         yield  'p' + str(i)
         i = i + 1
-genprefix = _genprefix().next
+genprefix = next(_genprefix())
```
tehasdf: `genprefix` used to be a function yielding the next prefix, but now this is just a value from the generator..

(this was supposed to go into the review i just submitted but i didn't actually click the "add comment" button :( )

src/twisted/web/test/test_domhelpers.py
rodrigc: ok

src/twisted/web/microdom.py
rodrigc: ok

src/twisted/web/microdom.py
rodrigc: ok

src/twisted/web/sux.py
rodrigc: What would the advantage of that be?  self._prepend is set to a different value in each branch.

src/twisted/web/microdom.py
rodrigc: Does this matter, or should this be fixed in a different way?

src/twisted/web/sux.py
tehasdf: right, would need to also set `self._prepend = data[0:2]` or the like; not necessary, but the branches are the same...

src/twisted/web/microdom.py
tehasdf: this likely means the part that uses genprefix isn't exercised in tests, as this would be a TypeError then
it is called here: https://github.com/twisted/twisted/blob/c94a97f6b5c8bb7bd8392cf086adbc79218fda1c/src/twisted/web/microdom.py#L632
i think we can have `genprefix = _genprefix()` and change the calling line to `prefix = next(genprefix)`

rodrigc: `genprefix()` is called in `twisted.web.test.test_xml` so it is being exercised in the tests.

If you suggestion is to do `genprefix = _genprefix(); prefix=next(genprefix)`, then why bother keeping `_genprefix()`?

src/twisted/web/sux.py
rodrigc: OK

src/twisted/web/microdom.py
tehasdf: hmm, i can't see where is that called in `test_xml`, and it would be odd that it wouldn't throw an exception: since currently `genprefix = next(_genprefix())`, that is already a string, so calling `genprefix()` like on the line i linked is of course an error

indeed, the previous version used `genprefix = _genprefix().next` so that `genprefix` was a callable that returned a new, unique string each time it was called

this was because only one `_genprefix()` generator was ever created, and it was repeatedly `.next()`-ed. 
That's why i suggested also only creating one `_genprefix()` generator, and `next()`-ing it repeatedly where needed. This way, subsequent calls to `next(genprefix)` would get different strings, since `genprefix` was the same generator each time

(which is to say, we can't do eg. `prefix = next(_genprefix())` because then we'd get the same string every time)

src/twisted/web/microdom.py
```diff
@@ -367,9 +366,9 @@ class Comment(CharacterData):
     def writexml(self, stream, indent='', addindent='', newl='', strip=0,
                  nsprefixes={}, namespace=''):
         val=self.data
-        if isinstance(val, UnicodeType):
-            val=val.encode('utf8')
-        stream.write("<!--%s-->" % val)
+        if isinstance(val, bytes):
+            val = val.decode('utf8')
+        stream.write(u"<!--{}-->".format(val))
```
glyph: There is a minor interface consistency issue here; previously, this `writexml` method would have accepted an `io.BytesIO`; now, it requires a `io.StringIO`.  Normally I'd say that we should follow the example of the standard library, and their precedent says that `writexml` should be in terms of native `str`s and therefore text going forward.

However, *the only reason that this module exists at all* is the capricious interface changes in the stdlib's XML support that broke almost every XML-based application over the years, so interface stability is a particularly sore point here :).  Can you fix this to be permissive, and use `twisted.python.compat.ioType` to allow for either str- or bytes-based file-like objects?

src/twisted/web/microdom.py
```diff
@@ -623,7 +622,7 @@ def writexml(self, stream, indent='', addindent='', newl='', strip=0,
             # namespace.  Nothing extra to do here.
             bext(self.tagName)
 
-        j = ''.join
```
glyph: Same as above comment.  Please feel free to use any old hacky workaround to achieve this; wrapping an underlying `BytesIO` in a `TextIO` wrapper and leaving the rest of the code exactly as you've changed it would be fine.

glyph (approved):
Overall this looks like a great change - nice to see 6 more lines coming out of `_setup.py` :).

However, in addition to the `writexml` type-signature issue, if you could address the couple of coverage issues to get the build status in better shape that would be great.  They're all pretty straightforward just-cover-one-more-branch issues, so no need for another review cycle, I think.

Unfortunately, `genprefix` is a public name, so please do address @tehasdf's comments as well.  It's a bit silly, and it should be deprecated, but there it is.

tehasdf (approved):

PR #840, opened 2017-07-22
9223 trial tutorial code

avandeursen:
Fix for TRAC ticket [9223](https://twistedmatrix.com/trac/ticket/9223).

Makes the code used in the [trial tutorial](http://twistedmatrix.com/documents/17.5.0/core/howto/trial.html) compatible with Python 3 (lazy map, floor division, multiple inheritance) and the latest version of Twisted (transfering bytes rather than strings).


docs/core/howto/listings/trial/calculus/base_3.py
```diff
@@ -1,16 +1,20 @@
...
+    def _robust_int(self, arg):
         try:
-            return map(int, args)
+            return int(arg)
         except ValueError:
-            raise TypeError("Couldn't coerce arguments to integers: %s" % args)
+            raise TypeError("Couldn't coerce argument to integer: %s" % arg)
+
+    def _make_ints(self, *args):
+        return map(self._robust_int, args)
```
rodrigc: Please replace this with a list comprehension, for consistency on Python 2 and 3:

```
return [int(arg) for arg in args]
```
Also, there is no need for the `_robust_int` function.  It is not adding any value, since the existing `ValueError` exception is fine.

docs/core/howto/listings/trial/calculus/client_1.py
```diff
@@ -18,7 +18,7 @@ def lineReceived(self, line):
     def _sendOperation(self, op, a, b):
         d = defer.Deferred()
         self.results.append(d)
-        line = "%s %d %d" % (op, a, b)
```
rodrigc, line 21: Since you are modernizing the code, please replace all % modifiers with `.format()` in your pull request.
So this will be:

```
"{} {} {}".format(op, a, b)
```

docs/core/howto/listings/trial/calculus/client_1.py
```diff
@@ -18,7 +18,7 @@ def lineReceived(self, line):
...
+        line = ("%s %d %d" % (op, a, b)).encode('utf-8')
```
rodrigc: Additionally, for consistent behavior on Python 2 and 3, any place you are doing an encode on a string, please prefix with *u*, so this will be:

```
u"{} {} {}".format(op, a, b).encode('utf-8')
```

rodrigc (changes requested):

avandeursen:
OK, thanks, I will take care of this shortly.

docs/core/howto/listings/trial/calculus/base_2.py
```diff
@@ -11,4 +11,4 @@ def multiply(self, a, b):
         return a * b
 
     def divide(self, a, b):
-        return a / b
+        return a // b
```
rodrigc, line 17: One more thing.  Since you are doing division here, please add at the top of the file:

```from __future__ import division```

See: https://docs.python.org/2/library/__future__.html

avandeursen:
Thanks for the suggestions.

I took care of all, except removing `TypeError`, though I agree it is technically superfluous.

The tutorial in `docs/core/howto/trial.rst` introduces this exception to illustrate testing exceptions and flushing logged errors. So removing it from the code would also entail adjusting the tutorial text itself, as well as the test code.

I'm not opposed to this, but I thought it would be better to leave the tutorial itself the same as much as as possible.

rodrigc:
Ah, yes, I see `TypeError` is used extensively in the tutorial.  Good point.

rodrigc (approved):
Good job!  Thanks for going through the tutorial and cleaning things up.  That helps newcomers.

avandeursen:
Thanks! Bumped into this since I am such a newcomer :-)

PR #841, opened 2017-07-22
Ignore PyPy's implementation differences in base object class, fix 9225

dansut:
Fix for [#9225](https://twistedmatrix.com/trac/ticket/9225) by causing `twisted.python.reflect.accumulateMethods()` to ignore the base object class, which normally has no methods but under PyPy looks like it does due to implementation differences.

Should fix 3 tests:  
- test_reflect.AccumulateMethodTests.test_baseClass
- test_reflect.AccumulateMethodTests.test_ownClass
- test_reflect.PrefixedMethodsTests.test_onlyObject

### DIFF
--- a/src/twisted/python/reflect.py
+++ b/src/twisted/python/reflect.py
@@ -134,7 +134,8 @@ def accumulateMethods(obj, dict, prefix='', curClass=None):
     if not curClass:
         curClass = obj.__class__
     for base in curClass.__bases__:
-        accumulateMethods(obj, dict, prefix, base)
+        if base is not object:
+            accumulateMethods(obj, dict, prefix, base)
 
     for name, method in curClass.__dict__.items():
         optName = name[len(prefix):]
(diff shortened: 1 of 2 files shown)


src/twisted/python/reflect.py
```diff
@@ -134,7 +134,8 @@ def accumulateMethods(obj, dict, prefix='', curClass=None):
     if not curClass:
         curClass = obj.__class__
     for base in curClass.__bases__:
-        accumulateMethods(obj, dict, prefix, base)
+        if base is not object:
```
rodrigc: I don't understand this change.  Can you add a comment here as to why this change is correct?
Does this change work on Cpython 2.7 and Cpython 3.x?

rodrigc (changes requested):

rodrigc (approved):

src/twisted/python/reflect.py
rodrigc: OK, I did some tests on CPython 2.7 and CPython 3.6 and this change works.
It needs a comment though.

dansut: Sorry, should have mentioned I had run tests in various versions

PR #842, opened 2017-07-23
Port twisted.python.rebuild to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8213


rodrigc:
@dansut @tehasdf please review this.  This is based on patches originally submitted by @tehasdf in http://twistedmatrix.com/trac/ticket/8213

src/twisted/python/rebuild.py
```diff
@@ -156,8 +178,10 @@ def rebuild(module, doLog=1):
     values = {}
     if doLog:
         log.msg('  (scanning %s): ' % str(module.__name__))
+    ClassType = getattr(types, 'ClassType', None)
```
tehasdf: this can actually be removed?

otherwise lgtm

tehasdf (approved):

src/twisted/python/rebuild.py
rodrigc: Yup, you are right.

rodrigc:
@dansut please review this

src/twisted/python/rebuild.py
```diff
@@ -21,6 +21,23 @@
 
 lastRebuild = time.time()
 
+def _isClassType(t):
+    """
+    Compare to types.ClassType in a py2/3-compatible way
+
+    Python 2 used comparison to types.ClassType to check for old-style classes.
+    Python 3 has no concept of old-style classes, so if ClassType doesn't exist,
```
dansut: Comment line too long

src/twisted/python/rebuild.py
```diff
@@ -93,10 +110,14 @@ def latestClass(oldClass):
...
         return newClass
     except TypeError:
-        if newClass.__module__ == "__builtin__":
+        if newClass.__module__ in ("__builtin__", "builtins"):
             # __builtin__ members can't be reloaded sanely
             return newClass
-        ctor = getattr(newClass, '__metaclass__', type)
+
+        ctor = type(newClass)
+        # ctor is the metaclass in both Python 2 and 3, except if it was old-style
```
dansut: Comment line too long and doesn't start with capital

src/twisted/test/test_rebuild.py
```diff
@@ -43,55 +45,49 @@ class RebuildTests(unittest.TestCase):
...
         from twisted.python.util import sibpath
         import shutil, time
         shutil.copyfile(sibpath(__file__, "myrebuilder1.py"),
                         os.path.join(self.fakelibPath, "myrebuilder.py"))
         from twisted_rebuild_fakelib import myrebuilder
         a = myrebuilder.A()
-        try:
-            object
-        except NameError:
+        from twisted.test import test_rebuild
```
dansut: Once the try except has been removed this import seems oddly placed, but see further comment on where it is used to create C ..

src/twisted/test/test_rebuild.py
```diff
@@ -43,55 +45,49 @@ class RebuildTests(unittest.TestCase):
...
             pass
-        else:
-            from twisted.test import test_rebuild
-            b = myrebuilder.B()
-            class C(myrebuilder.B):
-                pass
-            test_rebuild.C = C
-            C()
+        test_rebuild.C = C
+        C()
```
dansut: Are we actually testing anything here or just exercising code?
Does this matter?
This bit of code initially confused me, then after thinking it through I think I see the intention but it seems to me the lack of related assert suggests to me this is no test (unfinished) - maybe I'm missing something but if so then a comment explaining would be nice.

twistedchecker also unhappy about C as class name

src/twisted/test/test_rebuild.py
```diff
@@ -43,55 +45,49 @@ class RebuildTests(unittest.TestCase):
...
-            self.assertEqual(b2.b(), 'c')
-            self.assertEqual(b.b(), 'c')
+        b2 = myrebuilder.B()
+        self.assertEqual(b2.b(), 'c')
+        self.assertEqual(b.b(), 'c')
         self.assertEqual(i.a(), 'd')
         self.assertEqual(a.a(), 'b')
-        # more work to be done on new-style classes
+        # More work to be done on new-style classes
         # self.assertEqual(c.b(), 'c')
```
dansut: Was this commented out line intended to test the above C class?

dansut (commented):
My first Twisted code review so a review of any points in my review is welcome - I tried to follow the guidelines.

There are a bunch of other twisted checker infractions but I'm guessing the rule of thumb is we're only reviewing bits the PR changes..?

Beyond the code it seems that there needs to be more thorough tests created to get the patch codecov up. Specifically I see there is no coverage for the Sensitive mixin at all, I guessed that wasn't touched with this change but codecov is obviously not happy and I wasn't sure if it should be done as part of this PR..?

src/twisted/test/test_rebuild.py
rodrigc: This is just exercising code.  I don't think it matters.

The classname `C` is OK here, because this is just creating a class as part of test code, it is not a class that is exported as part of a library.

src/twisted/test/test_rebuild.py
rodrigc: I think the test is unfinished because I think the intention was that after doing  `rebuild.rebuild(myrebuilder)`, then `class C(myrebuilder.B)`, would inherit from the rebuilt `myrebuilder.B`, and then `C.b()` would return `'c'` instead of `'b'`.

However it looks like this feature was not finished.
Finishing this off is outside the scope of this pull request.

dansut: Understood on the PR scope - I was just wondering if these unfinished bits of code in here (that only appear to exercise code that is already exercised by other tests, and make the purpose of the test slightly confusing) could just be removed, with "tidying up" being within scope of the PR... ?

This test appears to have been written to support Python versions when the whole new/old-style issue was different (pre-Py3) and it looks to me like the testing needs addressed here actually need to be rethought as a whole, and this fragment of non-functional test does not really help in this respect.

Not that I am any sort of expert, I'm just feeling out the process and depth/scope of PR reviews here - so am looking for feedback on my comments as much as anything else.

rodrigc: OK, I updated the test to remove some of those unfinished bits. Please review the updated patch.

dansut (approved):
I think this is an improvement.
I'll have a look at improving testing as part of 9223, which I've been working on a PR for but waiting for this one to land.

PR #847, opened 2017-07-23
Fix test_setup.py with setuptools 36.2.1

rodrigc:
http://twistedmatrix.com/trac/ticket/9231

See:
https://github.com/pypa/setuptools/issues/1098

### DIFF
--- a/src/twisted/newsfragments/9231.bugfix
+++ b/src/twisted/newsfragments/9231.bugfix
@@ -1,0 +1,1 @@
+twisted.python.test.test_setup now passes with setuptools 36.2.1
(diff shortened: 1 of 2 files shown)


rodrigc:
@benoit-pierre can you code review this?  When the Twisted builds go green, I need to get this in to unbreak the build.
I went with a function, and renamed the variables a bit differently to adhere to Twisted coding standards, which are not the same as pep8.

benoit-pierre (approved):

rodrigc:
@benoit-pierre thank you for the code review, and for walking me through the solution.

alex (approved):

PR #848, opened 2017-07-27
cannot add ranges in python3

fchapoton:
so replace range(A)+range(B) by list(range(A))+list(range(B)) twice

### DIFF
--- a/twisted/mail/imap4.py
+++ b/twisted/mail/imap4.py
@@ -460,7 +460,7 @@ class IMailboxListener(Interface):
 # Some definitions (SP, CTL, DQUOTE) are also from the ABNF RFC -
 # <https://tools.ietf.org/html/rfc2234>.
 _SP = ' '
-_CTL = ''.join(chr(ch) for ch in range(0x21) + range(0x80, 0x100))
+_CTL = ''.join(chr(ch) for ch in list(range(0x21)) + list(range(0x80, 0x100)))
 
 # It is easier to define ATOM-CHAR in terms of what it does not match than in
 # terms of what it does match.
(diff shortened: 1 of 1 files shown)


rodrigc:
In future, please follow all the steps at https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch when submitting a patch, including opening  a ticket at twistedmatrix.com

rodrigc:
There is a larger effort to port imap4 to Python 3 here:
https://github.com/twisted/twisted/pull/805

You might want to combine efforts with that and contribute your patch to the *9161-py3imap4* branch

twisted/mail/imap4.py
```diff
@@ -460,7 +460,7 @@ def newMessages(exists, recent):
 # Some definitions (SP, CTL, DQUOTE) are also from the ABNF RFC -
 # <https://tools.ietf.org/html/rfc2234>.
 _SP = ' '
-_CTL = ''.join(chr(ch) for ch in range(0x21) + range(0x80, 0x100))
+_CTL = ''.join(chr(ch) for ch in list(range(0x21)) + list(range(0x80, 0x100)))
```
rodrigc, line 463: This works.  Another way to fix this is to do something like:

```
from itertools import chain

_CTL = ''.join(chr(ch) for ch in chain(range(0x21),  range(0x80, 0x100)))
```

That allows you to combine the ranges into one iterable without allocating multiple lists.

rodrigc:
This has already been fixed in trunk: https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/imap4.py#L577

PR #849, opened 2017-07-28
[#9240] Run tests against python 3.6

michael-k:
https://twistedmatrix.com/trac/ticket/9240


michael-k:
I don't know why the py27-topfile tests failed on travis.  I didn't touch the py27 config.

rodrigc (changes requested):
In future, please follow https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch when submitting a patch to Twisted.

You missed two steps in the process mentioned at that link.  Specifically:

Step 6 which mentions that you need to add a newsfragment file

Step 11 which mentions that you need to add *review* keyword to the twistedmatrix.com ticket.  To move things along, I added the *review* keyword for you.

Can you add the newsfragment, as mentioned in the process above?
Adding the newsfragment will fix the topfile test.

rodrigc (changes requested):
Also, can you add Python 3.6 here: https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_setup.py#L70

That way this info will be added to:
https://pypi.python.org/pypi/Twisted

after the next release of Twisted.

michael-k:
And now I missed step number 15.  Maybe I should just close this PR.

rodrigc:
Try not to lose hope.  Sometimes submitting a patch to Twisted seems like running through an obstacle course, but once you get the hang of it, it is not too bad.

glyph:
@michael-k Twisted sets a high bar for contributions, and the process is part of that - but don't give up, *everybody* who contributes to Twisted (myself, the founder of the project, included) gets review feedback that they have to address!

glyph:
@rodrigc Thank you for your tireless review efforts recently :).

michael-k:
@rodrigc I've addressed the issues you mentioned.  Anything else I should do?

Note that the work done in twisted/twisted#858 overlaps with my changes.

rodrigc:
@hawkowl please review this.  It looks like this is overlapping with https://github.com/twisted/twisted/pull/858

hawkowl:
LGTM. I'll move the other stuff (Windows support) into another ticket.

hawkowl:
(just letting it run on the builders first)

.travis.yml
```diff
@@ -36,23 +36,28 @@ matrix:
...
     # For now all non-trial tests are in a single job to reduce the time spent
     # on starting separate jobs.
     - python: 2.7
       env: TOXENV=narrativedocs,apidocs,pyflakes,newsfragment,manifest-checker
-    - python: 3.5
+    - python: 3.6
       env: TOXENV=pyflakes3
     # Twistedchecker is running as a separate job so that we can ignore if it
     # fails.
-    - python: 3.5
```
rodrigc: twistedchecker cannot run on Python 3.6
This must remain on Python 3.5.
If you look at the logs: https://travis-ci.org/twisted/twisted/jobs/261731629 , this error is there:

```
AttributeError: 'TreeRebuilder3k' object has no attribute 'visit_joinedstr'
```

This is fixed upstream in newer versions of pylint and astroid: https://github.com/PyCQA/pylint/issues/1113

However, twistedchecker cannot use these new versions, because it monkey patches some internals
of astroid.  So we need to keep the twistedchecker stuff on Python 3.5 for now,
until we can update twistedchecker to work with newer versions of astroid.

rodrigc (changes requested):

.travis.yml
hawkowl: haha omg why does it return error code 0 when it totally doesn't work...

PR #850, opened 2017-07-31
Added --add-header to twist web

### LINKED ISSUE #9241: Add a --add-header flag to twist web
|[alex's avatar](https://github.com/alex)| @alex reported|
|-|-|
|Trac ID|trac#9241|
|Type|enhancement|
|Created|2017-07-28 12:17:46Z|

This would make it easy to add headers to all responses. This is useful for things like "adding an HSTS header to all responses in a static web app, without writing code".

```
trac-id__9241 9241
type__enhancement enhancement
reporter__alex alex
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__closed closed
resolution__fixed fixed
component__core core
keywords__None None
time__1501244266950695 1501244266950695
changetime__1501595055861628 1501595055861628
version__None None
owner__Alex Alex

```
</details>

alex:
Fixes #9241

### DIFF
--- a/src/twisted/web/tap.py
+++ b/src/twisted/web/tap.py
@@ -228,6 +228,7 @@ class _AddHeadersResource(resource.Resource):
         return self._originalResource.getChildWithDefault(name, request)
 
 
+
 def makeService(config):
     s = service.MultiService()
     if config['root']:
@@ -240,7 +241,6 @@ def makeService(config):
 
     if isinstance(root, static.File):
         root.registry.setComponent(interfaces.IServiceCollection, s)
-
 
     if config['extraHeaders']:
         headers = [h.split(b':', 1) for h in config['extraHeaders']]
--- a/src/twisted/web/tap.py
+++ b/src/twisted/web/tap.py
@@ -15,7 +15,7 @@ from twisted.internet import interfaces, reactor
 from twisted.python import usage, reflect, threadpool
 from twisted.spread import pb
 from twisted.web import distrib
-from twisted.web import server, static, script, demo, wsgi
+from twisted.web import resource, server, static, script, demo, wsgi
 from twisted.web import twcgi
 
 
@@ -65,6 +65,7 @@ demo webserver that has the Test class from twisted.web.demo in it."""
         usage.Options.__init__(self)
         self['indexes'] = []
         self['root'] = None
+        self['extraHeaders'] = []
 
 
     def opt_index(self, indexName):
@@ -174,6 +175,10 @@ demo webserver that has the Test class from twisted.web.demo in it."""
         self['root'].ignoreExt(ext)
 
 
+    def opt_add_header(self, header):
+        self['extraHeaders'].append(header)
+
+
     def postOptions(self):
         """
         Set up conditional defaults and check for dependencies.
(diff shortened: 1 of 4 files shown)


src/twisted/web/tap.py
```diff
@@ -224,6 +242,10 @@ def makeService(config):
     if isinstance(root, static.File):
         root.registry.setComponent(interfaces.IServiceCollection, s)
 
+    if config['extraHeaders']:
+        headers = [h.split(b':', 1) for h in config['extraHeaders']]
```
derwolfe: It feels like the logic for parsing the `--add-headers` option should live outside of `makeService`. The main reason for this would be ease of testability. I do think I see why this can't live on the options class: you are using `opt_add_header` which will take each instance of `--add-header "H1:HV1" --add-header "H2:HV2"` and join them together in a list.

Some examples of how to test option parsing logic can be found in [twisted/web/test/test_tap.py](https://github.com/twisted/twisted/blob/trunk/src/twisted/web/test/test_tap.py). For other examples, you could look at how conch's options or mail's options are tested (not that I'm sure they are canonical, but they do seem to conform to the tests that I've written for my own tap files).

src/twisted/web/tap.py
```diff
@@ -174,6 +175,10 @@ def opt_ignore_ext(self, ext):
         self['root'].ignoreExt(ext)
 
 
+    def opt_add_header(self, header):
```
derwolfe, line 178: This could use a docstring. :)

Lukasa (changes requested):
Thanks for this @alex! Can you add a news file for me please? I also think we should probably have tests for this functionality.

alex:
@Lukasa There is a newsfile. Do you have a suggestion as to where to look for the appropriate tests to imitate :-)

derwolfe (changes requested):
@alex thanks for putting this PR up! This definitely would have saved me from writing some code a while back. 

The main blocker that I can see is testing. I'd suggest adding some tests that 
1. exercise the new functionality of the options class. This would make it so you could confidently show that using `--add-header` does what you expect.
2. figure out a way to exercise the new service code. I'm not exactly sure what the best way forward would be here. I think it would be useful to know that given an HTTP request, you are able to get an HTTP response out that has the headers you expect. But, I could see there being an argument that this might be too high level of a test to live in `test_tap`. If that is the case, or the test is too complex, maybe it would be sufficient to show that the resource being used as part of `makeService` is an instance of `_AuthHeadersResource`?

Also, please feel free to ask for another review. While I do think I understand how twisted's options logic is tested and how to test `makeService`, I'm not super confident that my suggestions for testing are complete :)

alex:
@Lukasa, @derwolfe: Added tests!

Lukasa:
So while I'm running all the builders, I have a note here: this doesn't work if you specify the same header more than once. That if, is you do `--add-header "K1: V1" --add-header "K1: V2"` you will get only `K1: V2` in the output.

I *think* for now this limitation is fine: it's easily worked around by users, if they're interested in doing so. It would be good if we could document it somewhere, just so people know that it's a deliberate omission and not an oversight.

alex:
@Lukasa Repeated headers are handled now.

Lukasa (approved):
Cool, I'm going to get the buildbot to test this and then we should be good to go.

PR #852, opened 2017-08-02
9243 westfeld add deployment docs pip

westfeld:
see https://twistedmatrix.com/trac/ticket/9243


src/twisted/newsfragments/9243.doc
```diff
@@ -0,0 +1 @@
+documentation now explains how to deploy twistd plugins using python packaging and pip
```
exarkun: instead of "documentation" perhaps say `The "Writing a twistd plugin" howto`

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
 
 
 
+Deploy your application using python packages
+---------------------------------------------
+
+
+
```
exarkun, line 317: The existing formatting in this file is an artifact of a conversion tool from a different format.  There's no real reason to try to mimic it.  One blank line separating sections is plenty.  Content should also be formatted following http://rhodesmill.org/brandon/2012/one-sentence-per-line/

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
 
 
 
+Deploy your application using python packages
```
exarkun: titlecase section headings I think

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+      - ``myplugin.py`` - Plugin file
+
+
+
+
+
+.. code-block:: python
+
+
+    from distutils.core import setup
```
exarkun: I think `setuptools` is preferred by everyone now?  It provides nice things like `find_packages`, anyway, and no one really wants to maintain a manual list of packages.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+    setup(
+        name='MyApplication',
+        version='0.1dev',
+        packages=['myproject', 'twisted.plugins'],
+        license='Creative Commons Attribution-Noncommercial-Share Alike license',
+        long_description="Insert your long description here",
+        install_requires=[
+            'twisted',
+            ],
+    )
```
exarkun, line 364: Perhaps it would be worth highlighting the part of this that's actually relevant to twistd plugins... The `twisted.plugins` entry in `packages`?

Also, does this really work?  For a long time I've had the impression that it does not.  Instead, obscure alternate options like `py_modules` are required.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+.. code-block:: python
+
+
+    from distutils.core import setup
+
+    setup(
+        name='MyApplication',
+        version='0.1dev',
+        packages=['myproject', 'twisted.plugins'],
+        license='Creative Commons Attribution-Noncommercial-Share Alike license',
```
exarkun: Maybe the same license as Twisted here? :)

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+To create the python package from the directory the standard setup tools
+can be used:
+
+
+
+
+.. code-block:: console
+
+
+    python setup.py sdist
```
exarkun: wheels are the best, right?  Any reason to recommend sdist instead of wheel here?

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
 
 
+Deploy your application using python packages
+---------------------------------------------
+
+
+
+To deploy your application one possibility is to wrap it up in a python package.
+For this you need to write a special file ``setup.py``, which contains metadata
+of the package. You would have to extend the layout of your files like this:
```
exarkun, line 320: The directory layout for plugins is described in the plugins document.  If those docs aren't clear, perhaps they should be improved.  Then this document could refer to that one.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,107 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+
+    - **__init__.py**
+
+
+  - **twisted**
+
+
+    - **plugins**
+
+      - ``myplugin.py`` - Plugin file
```
exarkun: This is a "dropin file", technically.  The things inside it are plugins.

exarkun (dismissed):
Thanks.  I've left some comments inline.

docs/core/howto/tap.rst
glyph: My understanding is that the best practice is to consider the `sdist` the "root" thing, which you should always upload, and wheels are derivative artifacts of that.  The differences are mostly only relevant when it comes time to start building native extensions, of course.

glyph: (That said I *totally* just upload wheels to PyPI for a bunch of my smaller projects and direct people to git repos if they want to build from source.)

docs/core/howto/tap.rst
glyph: I am still waiting for someone with real distutils savvy to answer my question here: https://stackoverflow.com/questions/7275295/how-do-i-write-a-setup-py-for-a-twistd-twisted-plugin-that-works-with-setuptools

None of the answers seem really satisfying.

However, I just do the `packages` thing in the various packages I distribute nowadays - see for example https://github.com/glyph/txsni/blob/master/setup.py - and nobody seems to complain.  I don't know if this would cause issues if someone tried to build a Debian packages from these things or not, but increasingly it seems that building for distribution in the global package-manager-managed Linux site-packages is a Worst Practice.

docs/core/howto/tap.rst
westfeld: That is possible, however the unchanged section headings do not use titlecase. For consistency I did not use it, too.

docs/core/howto/tap.rst
westfeld: I see. I tried to stay in this style. I think changing this would be worth another ticket, as the diff will get quite hard to read with so many changes.

twm: I filed https://twistedmatrix.com/trac/ticket/9776 for this.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+
+    from setuptools import setup, find_packages
+
+    setup(
+        name='MyApplication',
+        version='0.1dev',
+        # it is neccesary to extend the found package list with the twisted.plugin
+        # directory. It cannot be automatically detected, because it should not
+        # contain a __init__.py file.
+        packages= find_packages().append('twisted.plugins'),
```
twm: `append()` doesn't return the list.

Try:

```python
packages=find_packages() + ['twisted.plugins'],
```

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+
+.. code-block:: python
+
+
+    from setuptools import setup, find_packages
+
+    setup(
+        name='MyApplication',
+        version='0.1dev',
+        # it is neccesary to extend the found package list with the twisted.plugin
```
twm: Typo: "neccesary" &rarr; "necessary"

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+        # directory. It cannot be automatically detected, because it should not
+        # contain a __init__.py file.
+        packages= find_packages().append('twisted.plugins'),
+        license='MIT License',
+        long_description="Insert your long description here",
```
twm: `license` and `long_description` are technically optional, so I suggest dropping them from this example to reduce noise.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+
+    - **plugins**
+
+      - ``myplugin.py`` - Dropin file containing the actual plugin
+
+
+
+
+
+.. code-block:: python
```
twm, line 348: There's no indication that this is `setup.py` in the text. Perhaps use the `:caption:` option to note that?

```rest
.. code-block:: python
    :caption: ``setup.py``
```

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+
+    pip install MyApplication-0.1dev.tar.gz
+
+
+
+
+
+
+For more information about packaging in Python have a look at the `hitchhiker's
+guide to packaging <https://the-hitchhikers-guide-to-packaging.readthedocs.io/>`_.
```
twm: This link leads to a semi-broken page (no CSS?). We should probably link to [PyPA's guide](https://packaging.python.org/tutorials/packaging-projects/), anyway.

docs/core/howto/tap.rst
```diff
@@ -310,6 +310,110 @@ For a full list of cred plugins supported, see :api:`twisted.plugins <twisted.pl
...
+To create the python package from the directory the standard setup tools
+can be used:
+
+
+
+
+.. code-block:: console
+
+
+    python setup.py sdist
```
twm: Since it is 2020 now, I suggest changing `python` to `python3`.

twm (changes requested):
Thanks for this! There are still a few things that must be addressed before we can merge, however.

I noted a serious issue with the example `setup.py` &mdash; I don't think it actually works as written.

The rest of the comments are minor. In addition to those I noted inline, "Python" should really be capitalized everywhere it appears in prose (including the news fragment).

docs/core/howto/tap.rst
westfeld: For me the link worked, so I left it in there but added the official Python packaging guide link, too.

twm (approved):
Thanks for sticking with this! I've added the PR to the merge queue.

westfeld:
Thanks for finally accepting the PR. It took 3 years, but finally it worked out.

I still think that improving the docs is an important issue and I learned that outlining changes in the mailing list before writing a PR is the better way to go.

PR #854, opened 2017-08-04
Remove "import _preamble" to eliminate pyflakes warnings

rodrigc:
http://twistedmatrix.com/trac/ticket/9247


rodrigc:
@glyph @hawkowl please review this.

rodrigc:
Without this fix, Travis is failing like this:
https://travis-ci.org/twisted/twisted/jobs/260805366

hawkowl:
Preamble is used and has side effects, though.

hawkowl:
Turns out _preamble is broken, so removing it isn't a big deal. I document a workaround in the release notes already (because it's broken)

rodrigc:
@hawkowl in this case should we also remove mention of _preamble from https://github.com/twisted/twisted/blame/trunk/docs/core/development/policy/coding-standard.rst#L380

rodrigc:
@hawkowl and maybe remove mention of it from MANIFEST.in: https://github.com/twisted/twisted/blob/trunk/MANIFEST.in#L8

hawkowl:
@rodrigc the coding standard is outdated anyway (no scripts go in bin/ anymore), we should do that in a different ticket

rodrigc:
@hawkowl OK, I've got a patch here which removes references to _preamble.py: ​https://github.com/twisted/twisted/pull/856

PR #859, opened 2017-08-04
Port twisted.python.short to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9170
https://twistedmatrix.com/trac/ticket/5007

### DIFF
--- a/src/twisted/newsfragments/9170.bugfix
+++ b/src/twisted/newsfragments/9170.bugfix
@@ -1,0 +1,1 @@
+twisted.python.shortcut now works on Python 3 in Windows.
--- a/src/twisted/python/_setup.py
+++ b/src/twisted/python/_setup.py
@@ -405,7 +405,6 @@ notPortedModules = [
     "twisted.python.finalize",
     "twisted.python.hook",
     "twisted.python.rebuild",
-    "twisted.python.shortcut",
     "twisted.python.test.cmodulepullpipe",
     "twisted.python.test.test_pydoctor",
     "twisted.python.test.test_win32",
@@ -414,7 +413,6 @@ notPortedModules = [
     "twisted.test.myrebuilder2",
     "twisted.test.test_hook",
     "twisted.test.test_rebuild",
-    "twisted.test.test_shortcut",
     "twisted.web.soap",
     "twisted.web.test.test_soap",
 ]
--- a/src/twisted/python/shortcut.py
+++ b/src/twisted/python/shortcut.py
@@ -50,36 +50,37 @@ class Shortcut:
             shell.CLSID_ShellLink, None,
             pythoncom.CLSCTX_INPROC_SERVER, shell.IID_IShellLink
         )
-        data = map(None,
-                   ['"%s"' % os.path.abspath(path), arguments, description,
-                    os.path.abspath(workingdir), os.path.abspath(iconpath)],
-                   ("SetPath", "SetArguments", "SetDescription",
-                   "SetWorkingDirectory") )
-        for value, function in data:
-            if value and function:
-                # call function on each non-null value
-                getattr(self, function)(value)
-        if iconpath:
-            self.SetIconLocation(iconpath, iconidx)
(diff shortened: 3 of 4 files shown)


rodrigc:
@opalmer can you please review this?

opalmer:
Took a quick look and at first glance it seems ok. I'll need to take a more in depth look later on but one question I do have though is if you plan to address the failing AppVeyor/freebsd test steps (I didn't get a chance to see why they're failing myself yet)?

rodrigc:
I pushed a few small tweaks to the tests and now all the Appveyor tests build.
I'm not sure why the FreeBSD builds were failing before, since none of this stuff runs on FreeBSD.
Right now the FeeBSD buildbots passed.

opalmer (approved):
Thanks, looks good now. I figured the FreeBSD test issues were due to flake but wasn't sure about the AppVeyor builds since those are Windows.

Looks like there's nothing to do in Trac (no review keywords to remove).

rodrigc:
@opalmer Thanks for the review.  I updated the Trac ticket accordingly

rodrigc:
@opalmer one unrelated question, what advantage does [Pywincffi](https://github.com/opalmer/pywincffi) have version over pywin32?

opalmer:
@rodrigc considering it's my own library that I started in the hope of it becoming an eventual replacement for pywin32 in Twisted I'm probably a bit biased but here's some of the reasons why I wrote it:

* pywin32 does not build wheels today and although there's pypiwin32 it still requires a few workarounds.
* There's no continuous integration or easy process for submitting changes making contributions harder especially for newcomers.
* I know pywin32 has tests but getting some of those tests to run is difficult and there's also not any open source statistics that I know of showing current code coverage.
* pywin32 is great if you need most of the Windows APIs but Twisted only needs [~37](https://github.com/opalmer/pywincffi/issues/69) so you end up with a smaller dependency.
* CFFI itself has become a popular library for many Python modules with C-dependencies. This provides an indirect benefit for pywincffi in that it lowers the barrier for entry. It also has the nice side effect of eliminating the need to write a C-extension.
* The documentation in pywincffi is more consistent, easier to follow and generally tries be give as much info as possible from the source. For example [CreateProcess](https://pywincffi.readthedocs.io/en/0.4.0/modules/pywincffi.kernel32.html#pywincffi.kernel32.process.CreateProcess) has a lot of info pulled from the MSDN pages but quite a few notes on caveats too for those who may not be familiar with Windows APIs.
* All the standards for the project around development are [documented](https://pywincffi.readthedocs.io/en/0.4.0/dev/index.html). Partially to make it easy to know what to expect of the library and try to and keep it consistent over time.

The project's [README](https://github.com/opalmer/pywincffi/blob/master/README.rst) lists a few more points too.

With all that said, there are some disadvantages too:

* It's not a drop in replacement for pywin32. I thought about making it a drop in replacement but after digging into pywin32 it was abstracting away a lot of things which in reality could have hidden bugs. cffi itself already provides methods for type conversion between Python and C and most issues regarding types can be easily work around by explicit casting (which is what pywincffi does).
* It is not as high level as pywin32. This goes along with the point above but as an example pywin32 abstracted away handles in a lot of different ways. In pywincffi you generally just get a windows `HANDLE` object back instead.
* It's not as heavily used. pywincffi has 98% test coverage and most of the tests are exercising the Windows APIs but that's still not the same as using it on a daily basis. I part of pywincffi for work and with some of the traffic I've seen on GitHub there may be two others parts of the library today. Although it would probably be a fine replacement for pywin32 in Twisted it will take time to work out. 

The two above disadvantages are the biggest issues I see with the library. However I also understand not everyone can program in C and fewer still probably can program in C, Python and often have to work on Windows. That's actually one of the big reasons I specifically added documentation around [writing functions](https://pywincffi.readthedocs.io/en/0.4.0/dev/functions.html).

rodrigc:
@opalmer thanks, that's a very thorough description

PR #862, opened 2017-08-13
Fix docs/core/examples on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9251


rodrigc:
@tehasdf can you please help review this?

docs/core/examples/longex.py
```diff
@@ -19,7 +19,7 @@ def connectionMade(self):
         
     def lineReceived(self, line):
         try:
-            numbers = map(long, line.split())
+            numbers = [int(num.decode("ascii")) for num in line.split()]
```
markrwilliams: `int` can parse integers out of byte strings, so it's not necessary to decode `num`.

docs/core/examples/tkinterdemo.py
```diff
@@ -15,7 +15,12 @@
 
 from __future__ import print_function
 
-from Tkinter import Tk, Frame, Button, LEFT
+try:
+    # Python 3
+    from tkinter import Tk, Frame, Button, LEFT
+except ImportError:
+    # Python 2
+    from Tkinter import Tk, Frame, Button, LEFT
```
markrwilliams: This branch will run if the user is running under Python 3, but their Python installation doesn't have `tk` support.  That's confusing.

Replacing this `try`/`except` with `if sys.version_info.major == 2`/`else` would generate the right `ImportError` for each version of Python.

docs/core/benchmarks/linereceiver.py
```diff
@@ -16,7 +17,7 @@ def benchmark(chunkSize, lineLength, numLines):
     bytes = ('x' * lineLength + '\r\n') * numLines
     chunkCount = len(bytes) / chunkSize + 1
     chunks = []
-    for n in xrange(chunkCount):
+    for n in range(chunkCount):
```
markrwilliams: This will fail with a `TypeError` because `chunkCount` becomes a `float` because `/` with two integer operands performs [true division](https://www.python.org/dev/peps/pep-0238/) on [Python 3 and later](https://docs.python.org/3.0/whatsnew/3.0.html#integers).  [The division operator should be switched to the one for floor division (`//`)](https://github.com/twisted/twisted/pull/862/files#diff-2245294c5595eb3baba2363c4767e540L17).

Also, [`deliver` function](https://github.com/twisted/twisted/pull/862/files#diff-2245294c5595eb3baba2363c4767e540R13) should use a `for` loop, because `map` is lazy on Python 3, and [the `math` import](https://github.com/twisted/twisted/pull/862/files#diff-2245294c5595eb3baba2363c4767e540L3) is unused.

docs/core/benchmarks/netstringreceiver.py
```diff
@@ -2,6 +2,7 @@
 # See LICENSE for details.
 from __future__ import print_function
 
+from twisted.python.compat import range
```
markrwilliams: It looks like this benchmark has been broken [since 2012, when `TestNetstring` was moved from `t.t.test_protocols` to `t.p.tests.test_basic`](https://github.com/twisted/twisted/commit/a1ca615b7c5b084d44a4ce7c4f0975a7df9bfaf4). 

We should either delete it or fix it.  It looks like changing `from twisted.test import test_protocols` to `from twisted.protocols.test import test_basic` and updating the reference will do it.

docs/core/examples/courier.py
```diff
@@ -32,9 +32,9 @@
 def trace_dump():
     t,v,tb = sys.exc_info()
     openlog(FILTERNAME, 0, LOG_MAIL)
-    syslog('Unhandled exception: %s - %s' % (v, t))
```
markrwilliams, line 35: This file looks right but I'll try to set up a Courier server to test this.  I wish there were automated tests for all the examples :frowning_face:

docs/core/examples/ftpclient.py
```diff
@@ -47,9 +46,9 @@ def fail(error):
...
     print('Processed file listing:')
     for file in fileListProtocol.files:
-        print('    %s: %d bytes, %s' \
-              % (file['filename'], file['size'], file['date']))
-    print('Total: %d files' % (len(fileListProtocol.files)))
+        print('    {}: {} bytes, {}'.format(
+              file['filename'], file['size'], file['date']))
+    print('Total: {} files'.format(len(fileListProtocol.files)))
 
 def showBuffer(result, bufferProtocol):
```
markrwilliams, line 53: It might be nice if this function decoded `bufferProtocol.buffer.getvalue()` as `charmap`, so that newlines in the buffer would render as newlines in the output.

docs/core/examples/longex2.py
```diff
@@ -75,12 +75,16 @@ class Numbers(basic.LineReceiver):
     """
     def lineReceived(self, line):
         try:
-            numbers = map(long, line.split())
+            numbers = [int(num.decode("ascii")) for num in line.split()]
```
markrwilliams: `int` can parse numbers out of byte strings, so there's no need to `.decode` here.

docs/core/examples/mouse.py
```diff
@@ -60,7 +60,7 @@ def up_right(self):
         log.msg("right")
 
     def move(self, x, y):
-        log.msg("(%d,%d)" % (x, y))
+        log.msg("({},{})".format(x, y))
```
markrwilliams: `t.p.mice` won't be ported to Python 3, so these changes aren't necessary.  In fact, since it's deprecated, maybe we should remove this example?

docs/core/examples/pbinterop.py
```diff
@@ -46,16 +46,16 @@ def remote_local(self, obj):
 
     def _local_success(self, result):
         if result != "hello, world":
-            raise ValueError("%r != %r" % (result, "hello, world"))
```
markrwilliams, line 49: `twisted.spread.jelly` is imported in this module but unused.

docs/core/examples/recvfd.py
```diff
@@ -49,14 +49,14 @@ def fileDescriptorReceived(self, descriptor):
 
     def lineReceived(self, line):
         if self.descriptor is None:
-            print("Received %r without receiving descriptor!" % (line,))
```
markrwilliams, line 52: The module docstring for this has a typo:

> To **runb** this example, run this program with one argument: a path giving the UNIX

markrwilliams (approved):
This looks good!  There are several failures, only one of which (the use of true division in `LineReceiver`) is directly related to this patch.

I also mentioned unused imports because these files aren't checked by `flake8`.

Please fix the failures, address the comments as you see fit, and merge.

PR #864, opened 2017-08-14
Remove gpsfix.py example

rodrigc:
https://twistedmatrix.com/trac/ticket/9253

It uses twisted.protocols.gps which was removed in https://github.com/twisted/twisted/commit/a292874a0d795a46e365817ff974301d791fad79


rodrigc:
@altendky can you code review this?

altendky (approved):
`twisted.protocols.gps` was removed in 680bdbab27f99b43055a657ea8c29191a4822d43 in favor of `twisted.positioning`.

This example certainly doesn't belong as-is but perhaps it should be updated to use the new interface?  I don't know if this is normally handled by conversion or deletion and then maybe addition of a new example later.

`grep -r gpsfix` did only result in hits in `.git` and the news fragment so this seems to be fully removed.

rodrigc:
While not quite the same, there is another example which illustrates  `twisted.positioning`: https://twistedmatrix.com/documents/current/core/howto/positioning.html , so I think we are ok here.
We can always add new examples later.

PR #865, opened 2017-08-16
Fix a few examples with code incompatible with Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9254


rodrigc:
@altendky Can you please code review this?

altendky (commented):
I generally reference this for py2/3 compatibility and `except Exception as e` looks proper. http://python-future.org/compatible_idioms.html#catching-exceptions

`.format()` isn't technically required but I do prefer it.

`print_function` is imported and parentheses properly added as well as `,` -> `end=''`.

Outside the diff there are some uses of `print()` that don't explicitly import `print_function`.  The pull request is titled for fixing Python 3 compatibility but should we really be correcting to Python 2/3 compatibility?

docs/core/howto/defer-intro.rst
new line 61
new line 428: is fine because section is explicitly for 3.5+ only

docs/core/howto/threading.rst
new line 65
new line 80
new line 106: with `print_function` `printResult()` could actually be dropped

I don't know the judgement of others but I could go either way on a .misc vs a .doc news fragment here since it's mostly just technical corrections not content change per-se.

rodrigc:
@altendky can you redo your review but click on the green *Review changes* button at https://github.com/twisted/twisted/pull/865/files and provide your review comments inline?  I'm finding it hard to map your review comments to the proper lines.

altendky:
@rodrigc I was going to do inline but it didn't let me do that for the lines of interest.  When I expand other areas that the diff doesn't touch I can't get the plus icon to click and leave an inline comment.  I'll just copy paste here.  Between line numbers and the context hopefully it will make sense.  Also, my comments were submitted as regular comments through the 'review changes' button.  :]

I don't know where the line is between snippets without imports and those that ought to have them so these points may be non-issues.

---

Uses `print()` but no import of `print_function`.  
docs/core/howto/defer-intro.rst: line 61
```
Sometimes it leads us to encode the order when we don't need to, as in this example::

    total = 0
    for account in accounts:
        total += account.get_balance()
    print("Total balance $%s" % (total,))

But that's normally not such a big deal.
```

---

Imports the reactor but not `print_function` despite using `print()`.
docs/core/howto/threading.rst: line 65
```
For example, to run a method in a non-reactor thread we can do::

    def aSillyBlockingMethod(x):
        import time
        time.sleep(2)
        print(x)

    reactor.callInThread(aSillyBlockingMethod, "2 seconds have passed")
    reactor.run()

``callInThread`` will put your code into a queue, to be run by the next available thread in the reactor's thread pool.
This means that depending on what other work has been submitted to the pool, your method may not run immediately.
```

---

Does not import `print_function` despite using `print()`.  But no other imports so...
docs/core/howto/threading.rst: line 80
```
.. note::
    Keep in mind that ``callInThread`` can only concurrently run a fixed maximum number of tasks, and all users of the reactor are sharing that limit.
    Therefore, you should not submit *tasks which depend on other tasks in order to complete* to be executed by ``callInThread``.
    An example of such a task would be something like this::

        q = Queue()
        def blocker():
            print(q.get() + q.get())
        def unblocker(a, b):
            q.put(a)
            q.put(b)

    In this case, ``blocker`` will block *forever* unless ``unblocker`` can successfully run to give it inputs; similarly, ``unblocker`` might block forever if ``blocker`` is not run to consume its outputs.
    So if you had a threadpool of maximum size X, and you ran ``for each in range(X): reactor.callInThread(blocker)``, the reactor threadpool would be wedged forever, unable to process more work or even shut down.
```

---

Imports the reactor but not `print_function` despite using `print()`.  Also, in py2 'classic' you would need `printResult()` since `print` is a statement and couldn't be added directly as a callback.  But, with `print_function` being a callable (afaik) `printResult()` is technically unneeded.  Though, it may make the example more clear and therefor worth keeping.
docs/core/howto/threading.rst: line 106
```
To get a result from some blocking code back into the reactor thread, we can use :api:`twisted.internet.threads.deferToThread <deferToThread>` to execute it instead of callFromThread.

    from twisted.internet import reactor, threads

    def doLongCalculation():
        # .... do long calculation here ...
        return 3

    def printResult(x):
        print(x)

    # run method in thread and get result as defer.Deferred
    d = threads.deferToThread(doLongCalculation)
    d.addCallback(printResult)
    reactor.run()

Similarly, you want some code running in a non-reactor thread wants to invoke some code in the reactor thread and get its result, you can use :api:`twisted.internet.threads.blockingCallFromThread <blockingCallFromThread>`::
```

rodrigc:
To address your comments, keep in mind that these are code snippets inside documentation
that are meant to illustrate a concept.  They are not full-fledged working examples.
The intent of my changes was to remove syntax that is absolutely an error in Python 3.

In Python 2, for  most usages,  `print(foo)` will do something that is acceptable even if `__future__.print_function` is not imported, so for the purposes of an illustrative example, it is OK.
The places where I explicitly imported `__future__.print_function` was where the Python 3 usage would be absolutely not do the right thing on Python 2, such as `print("foo", end='')`.

As you mentioned, some of the examples do not show all the required imports, but since it is part of an example, and not a full program, that's probably OK.

rodrigc:
While I consider my previous comments valid, it isn't too big a deal to add the imports in the places that you referenced, so I did so in an updated patch.

I left the `printResult()` intact, because that is meant to illustrate the writing a callback.

Please review the updated patch.

Thanks.

altendky:
I agree that some would work in py2 as you pointed out, some are definitely snippets, and I presume you added the import to the code blocks with other imports. So, I think we're good. But it's hard to look over on my phone so I'll check it when I get back to my computer in the morning.

altendky (approved):

PR #867, opened 2017-08-16
Allow UTF-8 characters (non-ASCII) in From: and To: fields in e-mail headers (RFC 6531)

rodrigc:
http://twistedmatrix.com/trac/ticket/9250

Original problem reported in:
https://github.com/buildbot/buildbot/issues/3499


rodrigc:
This patch was tested by @chrwen-omicron here:
https://github.com/buildbot/buildbot/issues/3499#issuecomment-322710656

who tested sending e-mail with a *From:* header that contained non-ASCII characters.

src/twisted/mail/smtp.py
```diff
@@ -175,24 +175,25 @@ def quoteaddr(addr):
...
     else:
-        return  b'<' + res[1].encode('ascii') + b'>'
+        return  b'<' + res[1].encode('utf-8') + b'>'
 
 COMMAND, DATA, AUTH = 'COMMAND', 'DATA', 'AUTH'
 
 
 # Character classes for parsing addresses
-atom = br"[-A-Za-z0-9!\#$%&'*+/=?^_`{|}~]"
+atom = u"[-A-Za-z0-9!\#$%&'*+/=?^_`{|}~\u00a0-\U001000ff]"
```
tardyp: I would add a reference link to where you did find the magic value \u00a0-\U001000ff

tardyp (commented):
some minor comments. unicode decode looks good

rodrigc:
@tardyp please code review this.

tardyp:
This looks good to me. I can see with codecov plugin that there are some of the code not covered, but I don't know if your coverage setup is correct.
I would prefer a true twisted expert also review those changes.

rodrigc:
@tardyp I added some more tests and all lines in the patch are now covered: https://codecov.io/gh/twisted/twisted/pull/867/diff

src/twisted/mail/smtp.py
```diff
@@ -533,7 +542,8 @@ def do_MAIL(self, rest):
             return
         # Clear old recipient list
         self._to = []
-        m = self.mail_re.match(rest)
+
+        m = self.mail_re.match(rest.decode("utf-8"))
```
markrwilliams: This doesn't comply with RFC 6531.  [A client may send UTF-8 encoded sequences if and only if an ESMTP server advertises the `SMTPUTF8` extension](https://tools.ietf.org/html/rfc6531#section-3.2);

> This document applies when an SMTPUTF8-aware SMTP client or server
> supports the SMTPUTF8 extension.  **For all other cases**, and for
> addresses and messages that do not require an SMTPUTF8 extension,
> **SMTPUTF8-aware SMTP clients and servers do not change the behavior
> specified in RFC 5321** [[RFC5321](https://tools.ietf.org/html/rfc5321)].

Where RFC 5321 obsoletes RFC 2821, the RFC called out in the previous implementation.

In this PR, the `SMTP` and `ESTMP` server always expects UTF-8 encoded messages without advertising `SMTPUTF8`.  This contravenes the RFC's requirements.  Furthermore, the RFC only pertains to ESMTP, as SMTP does not support extensions.

The good news is this is easy to fix: first, instead of decoding as `utf-8`, decode these strings as `self._encoding`.  `SMTP._encoding` should be `ascii`, and `ESMTP._encoding` can be `utf-8`.  Then, add `SMTPUTF8` to [the extensions returned by ESMTP.extensions](https://github.com/twisted/twisted/blob/a8f654d642b6a2a88524dfe690a6a04814752506/src/twisted/mail/smtp.py#L1628-L1639).  This will make both SMTP and ESMTP compliant with my reading of the relevant RFCs.

src/twisted/mail/smtp.py
```diff
@@ -1888,14 +1898,14 @@ def __init__(self, fromEmail, toEmail, file, deferred, retries=5,
         assert isinstance(retries, (int, long))
 
         if isinstance(toEmail, unicode):
-            toEmail = [toEmail.encode('ascii')]
+            toEmail = [toEmail.encode('utf-8')]
```
markrwilliams: This also violates RFC 6531, because a  [a client may send UTF-8 encoded sequences if and only if an ESMTP server advertises the `SMTPUTF8` extension](https://tools.ietf.org/html/rfc6531#section-3.2).

This implementation has both `SMTPSender` and `ESMTPSender` *always* send UTF-8 encoded sequences, which violates the RFC.  `SMTPSender` should never send UTF-8, while `ESMTPSender` should only send UTF-8 when [`ESMTP.esmtpState_serverConfig` encounters the `SMTPUTF8` extension](https://github.com/twisted/twisted/blob/a8f654d642b6a2a88524dfe690a6a04814752506/src/twisted/mail/smtp.py#L1429-L1443).  `ESMTPSenderFactory` can then encode its addresses as UTF-8, and raise an informative exception if the server doesn't advertise `SMTPUTF8`.

markrwilliams (changes requested):
This PR does not implement RFC 6531, but it can be made to with relatively minor changes.

Please make those changes and resubmit for review.

src/twisted/mail/smtp.py
rodrigc: Third parties such as buildbot are creating [ESMTPSenderFactory](https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/smtp.py#L2055) directly, which in turn calls [SMTPSenderFactory.__init__()](https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/smtp.py#L1867).
`SMTPSenderFactory.__init__()` is encoding the addresses.  This is all done before the `EHLO` message is sent to the server in order query the ESMTP extensions, such as SMTPUTF8.  So SMTPSenderFactory is encoding the addresses before it receives SMTPUTF8.

Please clarify in the code how we can get this to work.
Thanks.

markrwilliams: @rodrigc 
```python
class SMTPSenderFactory(...):
     ..
    _encoding = 'utf-8'
    ...
    def __init__(self, ...):
        ...
        if isinstance(toEmail, unicode):
            toEmail = [toEmail.encode(self._encoding)]
        elif isinstance(toEmail, bytes):
            toEmail = [toEmail]
        else:
            toEmailFinal = []
            for _email in toEmail:
                if not isinstance(_email, bytes):
                    _email = _email.encode(self._encoding)

                toEmailFinal.append(_email)
            toEmail = toEmailFinal
        ...
```

```python
class ESMTPSenderFactory(SMTPSenderFactory):
    ...
    _encoding = 'utf-8'
```

```python
class ESMTPClient(SMTPClient):
    ...
    def esmtpState_serverConfig(self, code, resp):
        ...
        if b'SMTPUTF8' not in items:
            self.sendError(UTF8Required(...))
```

SMTP clients still encode addresses as ASCII.  ESMTP clients encode them preemptively as UTF-8, but if the server can't accept UTF-8 addresses, the Sender Factory's results Deferred errbacks with a new exception `UTF8Required`.  This gives clients an informative error message and shuts down the connection without violating the RFC and causing potentially undefined behavior.

rodrigc: In your example, did you mean to put:

```
class SMTPSenderFactory(...):
     ..
    _encoding = 'ascii'
    ...
```

?

rodrigc: For the case of ESMTP, if the ESMTPClient encodes addresses as unicode,
but they fall in the ascii range, should the client still throw an exception, or should
it just change the encoding to ascii, and re-encode and send it along?

markrwilliams: @rodrigc Whoops, yeah, I meant UTF-8 for `SMTPSenderFactory`.  It might be neat to try to re-encode the address as ASCII, but it seems kind of tricky and would be worth a log message certainly.

It'd be better to defer encoding until the extensions have come in but it will require a fair amount of refactoring.

rodrigc: Also, if b'SMTPUTF8' is *not* supported by the server, would the exception be bettter named something like *UTF8NotSupported*?

rodrigc: The concern I have is that for people who are now using ESMTPClient with ascii addresses against
older servers which do not support the SMTPUTF8 option, then things go through.  Throwing this exception will be new unexpected behavior.

rodrigc: Maybe what I can do is modify the ESMTPClient so that if SMTPUTF8
is not in the options on the server, it inspects the bytes of the addresses,
and if any non-ASCII characters are in the address, it throws an exception.

tardyp: Note that In the case of Buildbot we want to anyway send that email, even if we have to replace non-ascii characters by ascii equivalent. Probably most of the client would have same need, but some other want to error out i that case, or at least warn in a UI. So we should probably raise that exception, and document an easy and portable way to translate that email into somthing that can be send via that old smtp server

rodrigc:
OK, I pushed a patch which works.  It does the following:

1. In SMTPClient, self._encoding is set to  "ascii"
2. In ESMTPClient,  self._encoding is set to "utf-8".
3. In ESMTPClient, after sending the EHLO, it looks for *SMTPUTF8* as an advertised option.  If the option is not there, self._encoding is set to "ascii", and an message is logged.
4.  In SMTPClient.sendLine(), it tries to verify that the bytes being sent comply with self._encoding.  If not, the connection is closed and an error is raised.

rodrigc:
I used the following test program:

```
from twisted.mail.smtp import sendmail
from twisted.python.log import startLogging
from twisted.internet.task import react

startLogging(open("/tmp/foobar.txt", "w"))

def main(reactor):
    d = sendmail(smtphost="192.168.1.13", from_addr=u"\N{SNOWMAN}foo@twistedmatrix.com", to_addrs=[u"rodrigc@dibbler.crodrigues.org"],
             msg=u"UNGA BUNGA".encode("utf-8"),
             senderDomainName=None,
             port=25,
             reactor=reactor,
             username=None,
             password=None,
             requireAuthentication=False,
             requireTransportSecurity=False)
    return d

react(main)
```

and tested against two mail servers on FreeBSD:

1.  Postfix 3.2.2 _with_ SMTPUTF8 offered
2.  Sendmail 8.15.2 _without_ SMTPUTF8 offered

With Postfix, when I sent an e-mail with non-ASCII UTF-8 characters in the address,
the email went through.  With Sendmail, the client raised an error and closed the connection.

src/twisted/mail/_except.py
```diff
@@ -138,17 +138,20 @@ def __init__(self, code, resp, log=None, addresses=None, isFatal=False,
...
+            return self.__bytes__().decode("utf-8")
     else:
         def __str__(self):
             return self.__bytes__()
 
 
     def __bytes__(self):
         if self.code > 0:
-            res = [networkString("%.3d %s" % (self.code, self.resp))]
+            res = [u"{:03d} {}".format(self.code, self.resp).encode("utf-8")]
```
markrwilliams: Why is this UTF-8?  This is used by `SMTPClient`, so it shouldn't be UTF-8.  This could using `self._encoding` to allow `ESMTPClientError` to inherit from it.

src/twisted/mail/_except.py
```diff
@@ -138,17 +138,20 @@ def __init__(self, code, resp, log=None, addresses=None, isFatal=False,
...
         else:
-            res = [self.resp]
+            if isinstance(self.resp, unicode):
+                res = [self.resp.encode("utf-8")]
```
markrwilliams, line 152: As above - this shouldn't be UTF-8.

src/twisted/mail/smtp.py
```diff
@@ -175,24 +175,32 @@ def quoteaddr(addr):
         return b'<' + bytes(addr) + b'>'
 
     if isinstance(addr, bytes):
-        addr = addr.decode('ascii')
+        addr = addr.decode('utf-8')
```
markrwilliams: This can't be universally UTF-8 because it's also used by `SMTPClient`, right?

src/twisted/mail/smtp.py
```diff
@@ -175,24 +175,32 @@ def quoteaddr(addr):
...
 
     res = parseaddr(addr)
 
-    if res == (None, None):
+    if res == ('', ''):
         # It didn't parse, use it as-is
-        return  b'<' + bytes(addr) + b'>'
+        return  b'<' + str(addr).encode('utf-8') + b'>'
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -175,24 +175,32 @@ def quoteaddr(addr):
...
     else:
-        return  b'<' + res[1].encode('ascii') + b'>'
+        return  b'<' + res[1].encode('utf-8') + b'>'
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -218,26 +226,30 @@ def __init__(self, addr, defaultDomain=None):
         if isinstance(addr, Address):
             self.__dict__ = addr.__dict__.copy()
             return
-        elif not isinstance(addr, bytes):
-            addr = str(addr).encode('ascii')
+        elif isinstance(addr, bytes):
+            addr = addr.decode('utf-8')
```
markrwilliams: As above - this is used by `SMTPClient`.

src/twisted/mail/smtp.py
```diff
@@ -218,26 +226,30 @@ def __init__(self, addr, defaultDomain=None):
...
+        elif addr is None:
+            addr = u""
 
-        self.addrstr = addr
+        self.addrstr = addr.encode("utf-8")
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
 
-        self.local = b''.join(local)
-        self.domain = b''.join(domain)
-        if self.local != b'' and self.domain == b'':
+        self.local = u''.join(local)
+        self.domain = u''.join(domain)
+        if self.local != u'' and self.domain == u'':
             if defaultDomain is None:
-                defaultDomain = DNSNAME
+                defaultDomain = DNSNAME.decode("utf-8")
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
             self.domain = defaultDomain
+        self.local = self.local.encode("utf-8")
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
+        self.domain = self.domain.encode("utf-8")
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
+        Remove RFC 6531 quotes from address.
         """
         res = []
 
-        if not isinstance(addr, bytes):
-            addr = str(addr).encode('ascii')
-
-        atl = filter(None, self.tstring.split(addr))
+        if isinstance(addr, bytes):
+            addr = addr.decode('utf-8')
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
-        def __str__(self):
-            return self.__bytes__()
+        return b''.join([r.encode("utf-8") for r in res])
 
-
-    def __bytes__(self):
+    def __str__(self):
         if self.local or self.domain:
-            return b'@'.join((self.local, self.domain))
+            return u'@'.join((self.local.decode("utf-8"),
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
+                              self.domain.decode("utf-8")))
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -259,55 +271,53 @@ def __init__(self, addr, defaultDomain=None):
...
         else:
-            return b''
+            return u''
+
 
+    def __bytes__(self):
+        return self.__str__().encode("utf-8")
```
markrwilliams: As above.

src/twisted/mail/smtp.py
```diff
@@ -574,7 +586,7 @@ def do_RCPT(self, rest):
         if not self._from:
             self.sendCode(503, b"Must have sender before recipient")
             return
-        m = self.rcpt_re.match(rest)
+        m = self.rcpt_re.match(rest.decode("utf-8"))
```
markrwilliams: Should be `self._encoding`, not `"utf-8"`.

src/twisted/mail/smtp.py
```diff
@@ -919,6 +931,8 @@ class SMTPClient(basic.LineReceiver, policies.TimeoutMixin):
     # None, perform no timeout checking.
     timeout = None
 
+    _encoding = "ascii"
+
     def __init__(self, identity, logsize=10):
         if isinstance(identity, unicode):
             identity = identity.encode('ascii')
```
markrwilliams: Should be `self._encoding`, not `"ascii"`.

src/twisted/mail/smtp.py
```diff
@@ -1791,7 +1826,7 @@ class SenderMixin:
     def getMailFrom(self):
         if not self.done:
             self.done = 1
-            return str(self.factory.fromEmail)
+            return unicode(self.factory.fromEmail)
```
markrwilliams, line 1842: This is going to return the repr of `b'some@address.com'` on Python 3, and decode `self.factory.fromEmail` as ASCII on Python 2.  This should probably be `unicode(self.factory.fromEmail, encoding=self._encoding)`.

src/twisted/mail/test/test_smtp.py
```diff
@@ -1511,6 +1517,11 @@ def test_strWithNegativeCode(self):
         """
         err = smtp.SMTPClientError(-1, b"foo bar")
         self.assertEqual(str(err), "foo bar")
+        err = smtp.SMTPClientError(-1, u"\N{SNOWMAN} smiling")
+        if _PY3:
+           self.assertEqual(str(err), u"\N{SNOWMAN} smiling")
+        else:
+           self.assertEqual(str(err), b"\xe2\x98\x83 smiling")
```
markrwilliams: This would be better:

```python
expected = u"\N{SNOWMAN} smiling"
if _PY3:
    errorAsString = str(err, encoding='utf-8')
else:
    errorAsString = str(err)
    expected.encode('utf-8')
self.assertEqual(stringError, expected)
```

because it is less redundant, makes it clear that the final result should have been encoded as UTF-8, and doesn't depend on an implicit default of UTF-8.

src/twisted/mail/test/test_smtp.py
```diff
@@ -1639,6 +1650,59 @@ def test_requireTLSAndHELOFallbackSucceedsIfOverTLS(self):
...
+        """
+        self.clientProtocol.dataReceived(
+            b"250-AUTH LOGIN\r\n"
+            b"250 STARTTLS\r\n")
+        self.assertEqual(self.clientProtocol._encoding, "ascii")
+
+        # Try to send UTF-8 text when client encoding is ascii
+        # will result in an error which causes the transport to disconnect.
+        self.clientProtocol.sendLine(u"RCPT TO:<прив@example.com>".encode("utf-8"))
+        self.assertTrue(self.clientProtocol.transport.disconnecting)
```
markrwilliams: Is it possible to assert that an exception is raised?

markrwilliams (changes requested):

src/twisted/mail/test/test_smtp.py
rodrigc: I couldn't figure out how to get the exception here, but I found that self.clientProtocol.transport.disconnecting is set to True on failure, due to the exception thrown when trying to decode().

src/twisted/mail/test/test_smtp.py
rodrigc: That doesn't work:

```
  File "/Users/crodrigues/twisted_15/src/twisted/mail/test/test_smtp.py", line 1523, in test_strWithNegativeCode
    errorAsString = str(err, encoding='utf-8')
builtins.TypeError: decoding to str: need a bytes-like object, SMTPClientError found
```

src/twisted/mail/smtp.py
rodrigc: That doesn't work:

```
          File "/Users/crodrigues/twisted_15/src/twisted/mail/smtp.py", line 1829, in getMailFrom
            return unicode(self.factory.fromEmail, encoding=self._encoding)
        builtins.TypeError: decoding to str: need a bytes-like object, Address found
```

src/twisted/mail/_except.py
rodrigc: I don't see how this can be done easily.  These exceptions don't have access to self._encoding, so every single exception inheriting from SMTPClientError needs to be modified to take an encoding argument in its constructor....and there are lots.

I'm relying on this code in SMTPClient.sendLine() to fail if any invalid characters are in the stream:
```
        try:
            line.decode(self._encoding)
        except UnicodeDecodeError:
            self.sendError(
                SMTPProtocolError(-1,
                    "Server supports {} encoding, invalid "
                    "characters detected in sent line:\n{} .".format(
                        self._encoding, repr(line)),
                    self.log.str()))
            return
```

I've used that successfully to test sending invalid ASCII characters to a Sendmail server which does not support SMTPUTF8.

src/twisted/mail/_except.py
rodrigc: See above

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

src/twisted/mail/smtp.py
rodrigc: ok

rodrigc:
@jlquinn was able to successfully test this branch of Twisted as part of a buildbot install here:

https://github.com/buildbot/buildbot/issues/3633#issuecomment-330932589

PR #868, opened 2017-08-17
Fix nested list indentation

twm:
While looking at the docs I noticed that a few lists were rendered with structures like:

```html
<ol><li>...</li></ol>
<blockquote>
  <ul>
    <li>...</li>
  </ul>
</blockquote>
<ol start="2"><li>...</li></ol>
```

This is noticeable visually because the nested list has excessive left indentation. This happens because of a conflict between the reST syntax for nested lists and blockquotes. Specifically, if you under-indent a nested list, docutils takes it as a top-level blockquote containing a list. For example, this syntax is off:

```rst
- foo
 - bar (not enough indentation)
```
 While this is correct:

```
- foo

  - bar
```

There is also [the opposite problem](https://twistedmatrix.com/documents/current/installation/howto/optional.html): over-indented nested lists adding blockquotes.

The simple rule is that the content of a list must be indented exactly to match the first character of text.

I did a pass over all of the .rst documents in the `docs/` directory and fixed this in a few places, and also changed some enumerated lists to use the `#.` syntax instead of explicit numerals, as this makes it easier to see when the syntax is off (because the numbering restarts).

This addresses https://twistedmatrix.com/trac/ticket/9256


rodrigc:
Please follow all the steps at: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch
* need a newsfragment
* need to set *review* keyword on Trac ticket

twm:
@rodrigc Yup, on the way (I was interrupted mid-push).

docs/core/howto/ssl.rst
```diff
@@ -19,20 +19,20 @@ To create a TLS client, use :api:`twisted.internet.endpoints.SSL4ClientEndpoint
 TLS provides transport layer security, but it's important to understand what "security" means.
 With respect to TLS it means three things:
 
-    1. Identity: TLS servers (and sometimes clients) present a certificate, offering proof of who they are, so that you know who you are talking to.
-    2. Confidentiality: once you know who you are talking to, encryption of the connection ensures that the communications can't be understood by any third parties who might be listening in.
-    3. Integrity: TLS checks the encrypted messages to ensure that they actually came from the party you originally authenticated to.
-       If the messages fail these checks, then they are discarded and your application does not see them.
+1. Identity: TLS servers (and sometimes clients) present a certificate, offering proof of who they are, so that you know who you are talking to.
```
markrwilliams: Why not make this `#.` as well?

markrwilliams (changes requested):
Documentation improvements are always fantastic!

The affected docs look good when rendered locally.

The news fragment should be a `.doc`, because even though they're small, they're very useful changes and worth recording.   Please update it.

I've pushed this branch to the builders, so you can see if there are any issues in the meantime.

docs/core/howto/ssl.rst
twm: I could go either way, as the text preceding the list says "three things", but okay. Probably best to encourage this syntax everywhere.

twm:
Thank you for the review! I believe that I have addressed all of your feedback.

markrwilliams (approved):

PR #869, opened 2017-08-19
Fix "twistd -n --uid 1000 web" if no --gid flag is specified

rodrigc:
http://twistedmatrix.com/trac/ticket/4442

This fix is inspired by work originally done by @exarkun on this branch but wasn't merged:
https://github.com/twisted/twisted/tree/initgroups-default-group-4442

WIth this fix it is now possible to do:

```
twistd -n --uid 1000 web
```


src/twisted/scripts/_twistd_unix.py
```diff
@@ -68,7 +69,7 @@ class ServerOptions(app.ServerOptions):
     def opt_version(self):
         """Print version information and exit.
         """
-        print('twistd (the Twisted daemon) %s' % copyright.version)
+        print('twistd (the Twisted daemon) {}'.format(copyright.version))
```
markrwilliams: Why change all this formatting from `%` to `.format`?  It isn't necessary for this change and makes the diff harder to read. Worst of all, though, some of these changes affect lines that have no test coverage, so we won't know if they've introduced a regression.  I read through them and am pretty sure they don't, but I would feel more confident if this PR did not include any string formatting changes.

src/twisted/scripts/_twistd_unix.py
```diff
@@ -439,6 +442,8 @@ def startApplication(self, application):
             uid = process.uid
         if gid is None:
             gid = process.gid
+        if uid is not None and gid is None:
+            gid = pwd.getpwuid(uid).pw_gid
```
markrwilliams, line 450: Please update [the command line help for `--gid`](https://github.com/twisted/twisted/blob/8f34b9b8b5f2a14750e45e636de22e5bfa1044a6/src/twisted/scripts/_twistd_unix.py#L55) so that it mentions that `--gid` is now optional, and that it defaults to the primary group of the user specified by `--uid`.

markrwilliams (changes requested):
Thanks for reviving this very nice improvement!

src/twisted/scripts/_twistd_unix.py
rodrigc: Updated.

src/twisted/scripts/_twistd_unix.py
rodrigc: Normally I would update this in a separate PR, but this change is relatively small and doesn't break anything.  you can look at this entire PR, commit by commit at https://github.com/twisted/twisted/pull/869/commits and see that I isolated the format changes.

rodrigc:
@markrwilliams I have added more tests so all the lines in this patch are covered.  Please review.

src/twisted/python/test/test_util.py
```diff
@@ -847,29 +847,31 @@ class InitGroupsTests(unittest.TestCase):
...
-
     def test_initgroupsInStdlib(self):
         """
         Calling L{util.initgroups} will call the underlying stdlib
         implmentation.
         """
         calls = []
-        util._initgroups = lambda x, y: calls.append((x, y))
+
+        mockInitgroups = lambda x, y: calls.append((x, y))
```
markrwilliams: Functions with names should be defined with `def`.

src/twisted/test/test_twistd.py
```diff
@@ -359,6 +360,27 @@ def test_badAttributeWithConfiguredLogObserver(self):
...
+        if platformType == "win32":
+            name = "(the Twisted Windows runner)"
+        else:
+            name = "(the Twisted daemon)"
+        expectedOutput = ('twistd {} {}\n{}\n'.format(
+            name, copyright.version, copyright.copyright))
+
+        out = NativeStringIO()
+        stdout = self.patch(sys, 'stdout', out)
+        config = twistd.ServerOptions()
```
markrwilliams: `ServerOptions` could accept a file to point `print` at, with a default of `sys.stdout`.  Then you could pass this `out` file in without having to monkey patch anything.

src/twisted/test/test_twistd.py
```diff
@@ -359,6 +360,27 @@ def test_badAttributeWithConfiguredLogObserver(self):
...
+        e = self.assertRaises(SystemExit, config.parseOptions, ['--version'])
+        self.assertIs(e.code, None)
+        stdout.restore()
```
markrwilliams: Why call `restore` here? `SynchronousTestCase.patch` adds `MonkeyPatcher.restore` to the cleanup functions run after the test.

src/twisted/test/test_twistd.py
```diff
@@ -416,6 +437,20 @@ def kill(pid, sig):
         self.assertFalse(os.path.exists(pidfile))
 
 
+    def test_unexpectedOSError(self):
+        """
+        Raise an unexpected OSError when killing the pid.
+        """
+        pidfile = self.mktemp()
+        with open(pidfile, "w") as f:
+            f.write(str(os.getpid() + 1))
```
markrwilliams: Why not write out a fixed value here?

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +899,47 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+            Ignore everything and pass
+            """
+
+        runner = UnixApplicationRunner({})
+        switchUID = self.patch(_twistd_unix, 'switchUID', switchUIDFail)
+        exc = self.assertRaises(SystemExit, runner.shedPrivileges, 35,
+                                200, None)
+        self.assertEqual(exc.code, 1)
+        switchUID.restore()
+        switchUID = self.patch(_twistd_unix, 'switchUID', switchUIDPass)
```
markrwilliams: This should be moved into a separate test.

src/twisted/test/test_twistd.py
```diff
@@ -416,6 +437,20 @@ def kill(pid, sig):
         self.assertFalse(os.path.exists(pidfile))
 
 
+    def test_unexpectedOSError(self):
+        """
+        Raise an unexpected OSError when killing the pid.
```
markrwilliams: `OSError` should be a link: `L{OSError}`.

The docstring should describe the outcome, e.g.: `An unexpected L{OSError} when checking the validity of a PID in a C{pidfile} terminates the process via L{SystemExit}`

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +899,47 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+        switchUID.restore()
+        switchUID = self.patch(_twistd_unix, 'switchUID', switchUIDPass)
+        runner.shedPrivileges(35, 200, None)
+        switchUID.restore()
+
+
+    def test_uidWithoutGid(self):
+        """
+        Set the uid without settig the gid and then call
+        L{UnixApplicationRunner.startApplication}.
```
markrwilliams: The docstring should describe the outcome, e.g.: `Starting an application with L{UnixApplicationRunner} configured with a UID and no GUID results in a warning`.

Implied in this docstring is that it doesn't _fail_, the way it would before this PR.

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +899,47 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+        options = twistd.ServerOptions()
+        options.parseOptions([
+            '--nodaemon',
+            '--uid', os.getuid()])
+        application = service.Application("test_setupEnvironment")
+        self.runner = UnixApplicationRunner(options)
+        runner = UnixApplicationRunner(options)
+        runner.startApplication(application)
+        warningsShown = self.flushWarnings()
+        self.assertEqual(1, len(warningsShown))
```
markrwilliams: This test should make an assertion about the contents of the warnings, since this is a side effect of `shedPrivileges` that might change.

markrwilliams (dismissed):
The additional test coverage is fantastic!

Some feedback on the tests.

src/twisted/test/test_twistd.py
rodrigc: Are you talking about modifying ServerOptions?  There are two implementations,
one for Windows and one for Unix, so any modification would need to be made to both places.

src/twisted/python/test/test_util.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

markrwilliams:
@rodrigc Hm, I don't see any new commits.

src/twisted/test/test_twistd.py
markrwilliams: Yeah.  The change to the initializer could be done in [a shared base class](https://github.com/twisted/twisted/blob/487f9786d6cc01d5fc1cac5b815db19aed998d28/src/twisted/application/app.py#L592-L594), but all the `print` functions in each subclass would have to gain a new `file` argument.  Changing the initializer to accept a default argument should be safe even though this is a public API, because subclasses and callers don't need to change.

The problem with patching `sys.stdout` is that anything in the Python process can write to it, which means seemingly unrelated changes in far away parts of Twisted might render this test's assertions about the contents of `out` incorrect. So I think the change is worth it.

I would also prefer `checkPID's` signature to include `os.kill` as a default argument (e.g., `def checkPID(pidfile, kill=os.kill)`), but its code path is shallow enough that dependency injection seems less necessary.

src/twisted/python/test/test_util.py
```diff
@@ -847,29 +847,33 @@ class InitGroupsTests(unittest.TestCase):
...
 
         util.initgroups(os.getuid(), 4)
-        self.assertEqual(calls, [(pwd.getpwuid(os.getuid())[0], 4)])
+        self.assertEqual(calls, [(pwd.getpwuid(os.getuid()).pw_name, 4)])
         self.assertFalse(setgroupsCalls)
 
 
+    def test_initgroupsNoneGid(self):
+        """
+        Call L{util.initgroups} with None for gid gives an error.
```
markrwilliams: Call*ing*

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +901,91 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+        application = service.Application("test_setupEnvironment")
+        self.runner = UnixApplicationRunner(options)
+        runner = UnixApplicationRunner(options)
+        runner.startApplication(application)
+
+
+    def test_setUidWithoutGid(self):
+        """
+        Starting an application with L{UnixApplicationRunner} configured
+        with a UID and no GUID, will result in the GUID being
```
markrwilliams: no `,`

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +901,91 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+            self.assertEqual(uid, wantedUid)
+            self.assertEqual(gid, wantedGid)
+
+        def setuid(uid):
+            self.assertEqual(uid, wantedUid)
+
+        def setgid(gid):
+            self.assertEqual(gid, wantedGid)
+
+        from twisted.python import util
```
markrwilliams: Why not put this import at the top of the file?

src/twisted/test/test_twistd.py
```diff
@@ -359,6 +361,27 @@ def test_badAttributeWithConfiguredLogObserver(self):
         self.assertNotIn('\n', e.args[0])
 
 
+    def test_version(self):
+        """
+        Print out the version.
```
markrwilliams: `C{--version} prints the version.`

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +901,91 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
             ['/foo/chroot', '/foo/rundir', True, 56, '/foo/pidfile'])
 
 
+    def test_shedPrivileges(self):
+        """
```
markrwilliams, line 905: Empty docstring - maybe something like `L{UnixApplicationRunner.shedPrivileges} switches the process' user ID.`

src/twisted/test/test_twistd.py
```diff
@@ -865,6 +901,91 @@ def fakeSetupEnvironment(self, chroot, rundir, nodaemon, umask,
...
+        """
+        def switchUIDPass(euid, uid, gid):
```
markrwilliams: This should contain code that allows you to assert it's been called.

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: Fixed

src/twisted/test/test_twistd.py
rodrigc: I added something to set stdout and pass it down to the print().

hawkowl (approved):
LGTM. Merged it forward, go ahead and hit merge when the tests pass if I go to bed before they finish :)

PR #871, opened 2017-08-21
Port twisted.mail.bounce to Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9260


src/twisted/mail/bounce.py
```diff
@@ -7,64 +7,74 @@
...
 
 
-def generateBounce(message, failedFrom, failedTo, transcript=''):
+def generateBounce(message, failedFrom, failedTo, transcript='',
+                   encoding='utf-8'):
     """
     Generate a bounce message for an undeliverable email message.
 
-    @type message: L{bytes}
+    @type message: a file-like object
```
moshez, line 42: Isn't this change backwards incompatible?

src/twisted/mail/bounce.py
```diff
@@ -75,18 +85,23 @@ def generateBounce(message, failedFrom, failedTo, transcript=''):
...
     sz = message.tell()
-    message.seek(0, orig)
+    message.seek(orig, SEEK_SET)
     if sz > 10000:
         while 1:
             line = message.readline()
-            if len(line)<=1:
+            if isinstance(line, bytes):
+                line = line.decode(encoding)
+            if len(line) <= 0:
```
moshez, line 99: This logic is different from the previous logic (<=1)

moshez (changes requested):
A few things that seem to change the logic, independent from porting. Can you clarify?

src/twisted/mail/bounce.py
rodrigc: In addition to porting, I added more tests to increase the code coverage.  As I increased the code coverage, I found a few small bugs.  For example, `message` could never have been bytes.  If you look at this line:
https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/bounce.py#L81

The code is performing `seek()` and `tell()` on `message`, so `message` could never have been bytes.  It had to have been a file like object.  In this change I am just clarifying the doc string to match the reality of what this code has been doing for a long time.

src/twisted/mail/bounce.py
rodrigc: Again, this is a bugfix I found by increasing the code coverage of this module.
I found this line didn't work properly until I changed the logic to <= 0.

jlitzingerdev: The changes @rodrigc introduced actually ensure this code actually does something.  The previous `message.seek(2, 0)` ensured that `sz` would never be > 10000.  I assume the intent was `seek(0, 2)`, which is better expressed in the proposed changes.

jlitzingerdev (approved):
The changes proposed not only update to support Python3, but it also fix at least two bugs:
* A docstring issue
* Actually doing something with large messages

I think this is ready for merging.

PR #873, opened 2017-08-21
update conch/ssh/_cryptography_backports.py

ofek:
https://twistedmatrix.com/trac/ticket/9263

@rodrigc I opened this because I botched the rebase on #761.


src/twisted/conch/ssh/_cryptography_backports.py
```diff
@@ -8,59 +8,30 @@
...
-    hexString = '%x' % (integer,)
-    if length is None:
-        n = len(hexString)
-    else:
-        n = length * 2
-    return binascii.unhexlify(hexString.zfill(n + (n & 1)))
+from binascii import unhexlify
+
+
+if hasattr(int, "from_bytes"):
```
rodrigc: Instead of using hasattr(),
do:
`from twisted.python.compat import _PY3`

then do something like:

```
if _PY3:
     intFromBytes = ...
     intToBytes = ...
else:
     def intFromBytes(....):

     def intToBytes(...):
```

This will make the code slightly more readable.

src/twisted/conch/ssh/_cryptography_backports.py
```diff
@@ -8,59 +8,30 @@
...
+        assert not signed
+
+        # call bytes() on data to allow the use of bytearrays
+        return int(bytes(data).encode('hex'), 16)
+
+
+if hasattr(int, "to_bytes"):
+    def intToBytes(integer, length=None):
+        return integer.to_bytes(
+            length or (integer.bit_length() + 7) // 8 or 1, 'big'
```
rodrigc: If you are doing division stuff, can you add `from __future__ import division` at the top of the file?

src/twisted/conch/ssh/_cryptography_backports.py
```diff
@@ -8,59 +8,30 @@
...
+        )
+else:
+    def intToBytes(integer, length=None):
+        hex_string = '%x' % integer
```
rodrigc: since you are doing new stuff, can you change this to a "".format() style string?

rodrigc: More comments:

Can you add an unit tests for intToBytes and intFrom Bytes in `src/twisted/conch/test`?
This is a major change, so we need a unit test.

rodrigc (changes requested):

src/twisted/conch/ssh/_cryptography_backports.py
ofek: That won't work as https://docs.python.org/3/library/stdtypes.html#int.from_bytes was added in 3.2.

src/twisted/conch/ssh/_cryptography_backports.py
ofek: Floor division is the same in py2 and py3.

src/twisted/conch/ssh/_cryptography_backports.py
ofek: That is significantly slower.

ofek: Sure!

ofek:
Just fyi, update comes from https://github.com/pyca/cryptography/blob/master/src/cryptography/utils.py

src/twisted/conch/ssh/_cryptography_backports.py
rodrigc: It's best to import __future__.division in any file doing division, just to prevent any possible problems in future.

ofek:
@rodrigc Test added

src/twisted/conch/ssh/_cryptography_backports.py
rodrigc: Twisted does not support Python 3.2: https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_setup.py#L66

rodrigc:
Since you are copying intToBytes directly from cryptography, can't you
just delete the implementation in Twisted, and import it directly from cryptography?
(bumping up the cryptography dependency in Twisted's _setup.py)

ofek:
Oh. Hmm, it is my understanding that the whole reason for this is if Cryptography is unavailable. Why do we have this file then?

rodrigc:
_cryptography_backports.py was added in https://github.com/twisted/twisted/commit/3ae764d9dd1ed484f1ba37fc617098ef732ce1f7
as part of http://twistedmatrix.com/trac/ticket/8200 by @mithrandi .

The cryptography dependency has been bumped up at least once since then in https://github.com/twisted/twisted/commit/d2215145438b04cf947bb7c2672192cb969ad01c by @reaperhulk 

For conch, cryptography is a required dependency, so as long as it doesn't break existing code,
I think it is OK to bump the dependency.

ofek:
No version bump is necessary. Does this look fine?

rodrigc:
That's good we don't have to bump the cryptography dependency.

rodrigc:
So now that we can just use `int_to_bytes` from cryptography that is good, and all your changes to do that are good.
However, for `_cryptography_backports.py`.  Instead of whacking it, we should follow the 
document for deprecating things in Twisted:

https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html

So as per that document, we need to do:

1.  Add to _cryptography_backports.py something like this before each function:

```
@deprecated(Version("Twisted", 17, 6, 0), 
                        "use cryptography.utils.int_from_bytes instead")
```

2.  Create a unit test to test this deprecation.  Something like:

```
    def test_intToBytesDeprecated(self):
        """
        L{twisted.conch.ssh.cryptography_backports} is deprecated.
        """
        from twisted.conch.ssh._cryptography_backports import intToBytes
        warnings = self.flushWarnings([self.test_intToBytesDeprecated])
        self.assertEqual(warnings[0]['category'], DeprecationWarning)
        self.assertEqual(warnings[0]['message'], "BLAH")
```

3.  The news fragment should be a `.removal`.

Other than that, your change looks fine and is moving in the right direction.
Thanks.

markrwilliams:
The deprecation policy doesn't cover private APIs, like `_cryptography_backports`.  This PR should just remove it.

rodrigc:
@markrwilliams while it is true that the deprecation policy doesn't cover private APIs,
I've begun to shift to a more cautious approach by going through the Twisted deprecation policy,
even for private APIs.  Just because it is "private", it doesn't mean that people aren't somewhere out there using it.

However, I think we can just move forward with this and remove the file without going through the deprecation policy.

rodrigc:
newsfragment still needs to be a .removal with some text in it, not an empty .misc file.

ofek:
@rodrigc Done.

rodrigc:
py27-nodeps build needs to be fixed:
https://travis-ci.org/twisted/twisted/jobs/267017814

This build can be reproduced by doing the following in Python 2.7 environment:

```
python -n tox -e py27-nodeps-withcov-posix
```

ofek:
@rodrigc How would I fix that? Based on what you mentioned in https://github.com/twisted/twisted/pull/873#issuecomment-323890292 I thought `cryptography` would always be available.

markrwilliams:
@ofek The tests that are erroring out need to be skipped when `cryptography` isn't available.  Like `unittest`, `trial` skips `TestCase`s that have a `skip` class attribute with a string that explains why the case is skipped.  [`test_cftp`, for example, attempts to import `cryptography`; if that fails, it assigns `None` to `cryptography`](https://github.com/twisted/twisted/blob/trunk/src/twisted/conch/test/test_cftp.py#L19-L22), and then [assigns a bunch of `skip`s](https://github.com/twisted/twisted/blob/trunk/src/twisted/conch/test/test_cftp.py#L1503-L1511)

You can copy this pattern for all the failed tests.

glyph:
@rodrigc Compatibility is often a matter of judgement, so if you're the reviewer on this I won't question your judgement call - if I wanted to make it I'd do the code review ;-) - but there are a couple of points I do want to make about the compatibility policy:

1. *as a matter of policy* we should not be requiring compatibility work for every private API.  If there are known public users, especially if they're particularly prominent, we should work with them to migrate to a new API before the private API is removed, but ultimately we need to make sure it's clear to downstreams that it's their responsibility to fix and it's rude to depend on private API.  We don't want to make it totally miserable for devs to make *any* changes to Twisted; there's already way too much accidentally-exposed public API that we have to maintain, we shouldn't add much to that burden
2. As Google SRE learned, one should not *much* exceed your SLO https://cloudplatform.googleblog.com/2017/01/availability-part-deux--CRE-life-lessons.html - we should take real actual public API very seriously, but we should occasionally *intentionally* break unsupported / unused private APIs, so that developers will learn what they can rely on and what they can't; as long as we are fairly certain that no prominent public projects are relying on those private APIs for legitimate reasons (like the lack of any alternative implementation strategy).  We already do set pretty aggressive targets for compatibility of *public* API, I think.

All that said, it's important to keep our users happy and their software working, so a conservative approach to changes that we know will break stuff is always fine.  Just wanted to ensure that the other side of the coin was also reflected here.

rodrigc:
@ofek: try this: https://github.com/ofek/twisted/pull/1

ofek:
Well done @rodrigc that did it! Just checking on Windows now.

rodrigc:
One more for you: https://github.com/ofek/twisted/pull/2

ofek:
Can https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-3809 be restarted? Something happened on their end from what I can tell.

ofek:
All tests now pass.

rodrigc:
appveyor has a known problem: https://twistedmatrix.com/pipermail/twisted-python/2017-August/031596.html

ofek:
Oh I was unaware, thanks.

rodrigc:
This is looking good, but due to the tooling involved with Twisted, I cannot merge a patch from your repo to the Twisted repo because your branch is named *trunk*.
Please move all these changes to another branch, using naming convention specified in https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

and submit another pull request.
Other than that, I think this is good.

ofek:
Created #876

PR #874, opened 2017-08-22
#9265 Hash Verification Prompt fails in Python 3

the0id:
In Python 3 you can't open `/dev/tty` without setting `buffering=0`.


markrwilliams:
@the0id Don't forget a `newsfragment` in `src/twisted/conch/newsfragments`!

the0id:
@markrwilliams It's hard to remember newsfragments when dinner is waiting.

src/twisted/conch/newsfragments/9265.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conch.client.default.py verifyHostKey now opens /dev/tty with no buffer to be compatible with Python 3.
```
markrwilliams: A stray `.py` suffix!  Also, please mention that the `conch` command line program works under Python 3 now as well.

markrwilliams (approved):
This definitely makes `conch` work on Python 3.

Please fix the newsfragment so we can merge!

rodrigc (approved):
I could not reproduce the problem on OS X that this bug is trying to fix.

However, I logged into a Linux host (Fedora 23), and could definitely reproduce the problem.  Without this fix, if I tried to use conch with Python 3 under Fedora, the prompt never came up, and conch would fail.

With this fix, the prompt comes up, and I could successfully use conch with Python 3 and Linux.

Good job!

PR #875, opened 2017-08-23
Complete test lifecycle

markrwilliams:
http://twistedmatrix.com/trac/ticket/9267


markrwilliams:
cc @akaptur re https://github.com/twisted/treq/issues/205

mithrandi (approved):
:+1: Both of these bugs are pretty nasty, good job finding/fixing them. `tearDown` is documented in a number of places as being run even if the test fails, so not running it is definitely a bug.

markrwilliams:
Credit to @akaptur for finding this!

PR #877, opened 2017-08-24
Handle queries for unknown record types as best we can.

mithrandi:
https://twistedmatrix.com/trac/ticket/9095


src/twisted/names/common.py
```diff
@@ -67,8 +67,8 @@ def query(self, query, timeout=None):
         try:
             method = self.typeToMethod[query.type]
         except KeyError:
-            return defer.fail(failure.Failure(NotImplementedError(
```
markrwilliams, line 70: This is useful information to developers, not clients.  It'd be nice to log this failure.  Bonus points for using `twisted.logger.Logger` so you can set the log level down to `debug` or the like!

Also, this leaves the decision of what to do with the query type up to subclasses and it's possible that some subclass relies on this check.

[The documentation implies](https://twistedmatrix.com/documents/current/names/howto/custom-server.html#a-server-which-computes-responses-dynamically) users are *not* supposed subclass `Resolver`, and certainly they're not supposed to call `_lookup`, which is a private method.  That leaves the `Resolver`s inside Twisted itself.

Fortunately, all the resolvers that inherit from `ResolverBase` have `_lookup` methods that do the right thing when given a type they don't understand:

- `t.n.authority.FileAuthority`: works, because `test_domain_com` in `test_names` is a subclass of `FileAuthority` and your new tests pass.  Specifically, its `_lookup` method returns either `AuthoritativeDomainError` or `DomainError`:https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/authority.py#L204-L216 ...both of which `DNSServerFactory` turns into `NXDOMAIN` (known as `ENAME` inside Twisted): https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/server.py#L334-L335
- `t.n.secondary.SecondaryAuthority`: Either responds with `DomainError` if it's not the Server of Authority, or does what `FileAuthority` does:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/secondary.py#L158-L161
- `t.n.cache.CacheResolver`: works, because it fails with a `DomainError` on a unknown query type:
 https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/cache.py#L61-L67
- `t.n.client.Resolver`: this, of course, proxies through some other server's response: https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/client.py#L403-L409 
- `t.n.root.Resolver`: Behaves correctly because `t.n.client.Resolver` does.  Its `_lookup` calls `_discoverAuthority`: 
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/root.py#L114-L126
...which calls `_query`:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/root.py#L129-L159
...which uses `_resolverFactory`:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/root.py#L82-L111
...which is `t.n.client.Resolver` by default:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/root.py#L69-L71
- `t.n.resolver.ResolverChain`: responds to failures via `FailureHandler`:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/resolve.py#L80-L83  ...which in turn retries the query with the next resolver in the chain on `DomainError`:
https://github.com/twisted/twisted/blob/727e1e759a77f9b21d3e7145c463f3aa12abdff7/src/twisted/names/resolve.py#L27-L30

It's nice that all the tests pass, too.

src/twisted/names/test/test_names.py
```diff
@@ -594,6 +595,28 @@ def test_recordMissing(self):
...
 
+    def test_unknownTypeNXDOMAIN(self):
+        """
+        Requesting a record of unknown type where no records exist for the name
+        in question results in L{DomainError}.
+        """
+        testDomain = test_domain_com
+        testDomainName = b'nonexistent.prefix-' + testDomain.soa[0]
+        f = self.failureResultOf(
+            testDomain.query(Query(name=testDomainName, type=76)))
```
markrwilliams: Did you chose 76 because it falls within [the unassigned (63-98) `RRTYPE` range](https://www.iana.org/assignments/dns-parameters/dns-parameters-4.csv)?  If so, that's fine but obscure.

A clearer approach would be to use a value derived from the supported record types:

```python
unsupportedType = max(common.typeToMethod) + 1
f = self.failureResultOf(
    testDomain.query(Query(name=testDomainName, type=unsupportedType)))
```

This would also allow the test to pass in a future where IANA assigns 76 to some data `RRTYPE`!

src/twisted/names/test/test_names.py
```diff
@@ -594,6 +595,28 @@ def test_recordMissing(self):
...
+        self.assertIsInstance(f.value, DomainError)
+
+
+    def test_unknownTypeMissing(self):
+        """
+        Requesting a record of unknown type where other records exist for the
+        name in question results in an empty answer set.
+        """
+        answer, authority, additional = self.successResultOf(
+            my_domain_com.query(Query(name=u'my-domain.com', type=76)))
```
markrwilliams: As above.

markrwilliams (changes requested):
A request for some logging and an improvement to the test's clarity.

Also, a long comment justifying the change for posterity.

src/twisted/names/test/test_names.py
mithrandi: The maximum supported type in Twisted is 255 (ALL_RECORDS), so this would produce an illegal value (256). Oh, this is a 16-bit value, not an 8-bit value. Hmm. I guess that would work, then.

mithrandi: Okay, made this change.

src/twisted/names/test/test_names.py
mithrandi: Changed.

src/twisted/names/common.py
mithrandi: I added some logging; thanks for double-checking all the `ResolverBase` subclasses!

markrwilliams (approved):

PR #878, opened 2017-08-26
Port pop3 to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9269


rodrigc:
@markrwilliams please review

src/twisted/mail/imap4.py
```diff
@@ -5902,7 +5901,8 @@ def __bytes__(self):
 #            if self.peek:
 #                base += '.PEEK'
             if self.header:
-                base += b'[%s%s%s]' % (part, separator, self.header,)
+                base += (b'[' + part + separator +
+                         str(self.header).encode("ascii") + b']')
```
evilham, line 5905: Is `str(self.header).encode("ascii")` needed? From what I gather, the Header class in L5923 is used, wouldn't `bytes(self.header)` be equivalent and clearer?
`str(self.header)` is calling `twisted.python.compat.nativeString` which basically does `.decode("ascii")`.

src/twisted/mail/pop3client.py
```diff
@@ -899,7 +900,7 @@ def _apop(self, username, password, challenge):
             On a successful login, it returns the server response minus
             the status indicator.
         """
-        digest = md5(challenge + password).hexdigest()
+        digest = md5(challenge + password).hexdigest().encode("utf-8")
```
evilham: Since this is hex, I'd use `ascii` as encoding. Not really important, but it hints at the limited symbol space.

src/twisted/mail/protocols.py
```diff
@@ -82,7 +81,7 @@ def receivedHeader(self, helo, origin, recipients):
         fromUser = (b"from " + helo[0] + b" ([" + helo[1] + b"]" +
                  heloStr + authStr)
         by = (b"by " + self.host + b" with " + self.protocolName +
-              b" (" + networkString(longversion) + b")")
+              b" (" + longversion.encode("ascii") + b")")
```
evilham, line 84: Why doesn't `networkString` already do the trick?

src/twisted/mail/test/test_imap.py
```diff
@@ -3007,12 +3031,12 @@ def test_tooManyArgumentsForCommand(self):
...
             self.transport.value(),
-            networkString(
-                "001 BAD Illegal syntax:"
-                " Too many arguments for command: %r"
-                "\r\n"
-                % (b"C",)
-            ))
+            (b"001 BAD Illegal syntax:" +
+             b" Too many arguments for command: " +
+             repr(b'C').encode("utf-8") +
```
evilham, line 3037: Am I missing something here or is this equivalent to `b" Too many arguments for command: C\r\n"`?

src/twisted/mail/test/test_pop3.py
```diff
@@ -512,59 +580,64 @@ def setUp(self):
...
-        self.assertTrue(s.getvalue().find("SASL CRAM-MD5") >= 0)
+        p.lineReceived(b"CAPA")
+        self.assertTrue(s.getvalue().find(b"SASL CRAM-MD5") >= 0)
 
-        p.lineReceived("AUTH CRAM-MD5")
+        p.lineReceived(b"AUTH CRAM-MD5")
         chal = s.getvalue().splitlines()[-1][2:]
         chal = base64.decodestring(chal)
-        response = hmac.HMAC('testpassword', chal).hexdigest()
+        response = hmac.HMAC(b'testpassword', chal).hexdigest().encode("utf-8")
```
evilham: Same as somewhere else, here I'd use `.encode("ascii")`

src/twisted/mail/test/test_pop3client.py
```diff
@@ -586,7 +595,8 @@ def cbDisconnected(ign):
                 ['CAPA', 'STLS', 'CAPA', 'QUIT'])
```
evilham: wouldn't input be bytes? And therefore `[b'CAPA', ...`

src/twisted/mail/pop3.py
```diff
@@ -360,7 +364,9 @@ def formatUIDListLines(msgs, getUidl):
     for i, m in enumerate(msgs):
         if m is not None:
             uid = getUidl(i)
-            yield '%d %s\r\n' % (i + 1, uid)
+            if not isinstance(uid, bytes):
+                uid = str(uid).encode("utf-8")
+            yield intToBytes(i + 1) + b' ' + uid + b'\r\n'
```
evilham, line 369: I *think* if `uid` consists of only ascii chars, lines 367 and 368 can be replaced with `networkString`. But it's not entirely clear from the docs, it says "printable characters". Do have more info on this?

src/twisted/mail/pop3.py
```diff
@@ -588,11 +596,11 @@ def processCommand(self, command, *args):
         command = command.upper()
         authCmd = command in self.AUTH_CMDS
         if not self.mbox and not authCmd:
-            raise POP3Error("not authenticated yet: cannot do " + command)
-        f = getattr(self, 'do_' + command, None)
+            raise POP3Error(b"not authenticated yet: cannot do " + command)
+        f = getattr(self, 'do_' + command.decode("utf-8"), None)
```
evilham, line 601: I would suggest adding a `test_BadCommand` test. One where `command.decode("utf-8")` explodes.
E.g.: `b"\x81".decode("utf-8")`, so command `b"\x81"`.

src/twisted/mail/pop3.py
```diff
@@ -945,11 +957,13 @@ def ebMessages(err):
                 if i < 1:
                     raise ValueError()
             except ValueError:
-                self.failResponse("Invalid message-number: %r" % (i,))
+                if not isinstance(i, bytes):
```
evilham, line 961: Isn't this what `intToBytes` is for? We know `i` is an integer here.

src/twisted/mail/pop3.py
```diff
@@ -962,7 +976,10 @@ def ebMessage(err):
                                 "raise IndexError for out-of-bounds message numbers: "
                                 "raise ValueError instead.",
                                 PendingDeprecationWarning)
-                        self.failResponse("Invalid message-number: %r" % (i,))
+                        invalidNum = i
+                        if invalidNum and not isinstance(invalidNum, bytes):
```
evilham, line 982: Sam as above, wouldn't `intToBytes` suffice?

src/twisted/mail/pop3.py
```diff
@@ -1493,7 +1514,9 @@ def sendLong(self, command, params):
         @param params: Command arguments.
         """
         if params:
-            self.sendLine('%s %s' % (command, params))
+            if not isinstance(params, bytes):
+                params = str(params).encode("utf-8")
```
evilham, line 1521: Same observation as above, a `test_BadCommand` that breaks `.encode("utf-8")` is interesting

src/twisted/mail/pop3.py
```diff
@@ -1534,7 +1557,7 @@ def _dispatch(self, command, default, *args):
         Command X is dispatched to handle_X() if it exists.  If not, it is
         dispatched to the default handler.
 
-        @type command: L{bytes}
+        @type command: L{str}
```
evilham: I actually expect this to be `bytes`. What lead you to change it?

src/twisted/mail/pop3.py
```diff
@@ -1545,7 +1568,9 @@ def _dispatch(self, command, default, *args):
         @param args: Arguments to the handler function.
         """
         try:
-            method = getattr(self, 'handle_'+command, default)
+            if isinstance(command, bytes):
+                command = command.decode("utf-8")
+            method = getattr(self, 'handle_' + command, default)
```
evilham: Same as above, I picture this looking a lot like L600:
https://github.com/twisted/twisted/blob/8095f968910471e017e508244391b30e099cdd93/src/twisted/mail/pop3.py#L600

src/twisted/mail/pop3.py
```diff
@@ -1595,7 +1620,7 @@ def apopAuthenticate(self, user, password, magic):
         @type magic: L{bytes}
         @param magic: The challenge provided by the server.
         """
-        digest = md5(magic + password).hexdigest()
+        digest = md5(magic + password).hexdigest().encode("utf-8")
```
evilham: As somewhere else, I'd use `.encode('ascii')` here.

evilham (changes requested):
Sorry it took me longer than expected Craig.

src/twisted/mail/imap4.py
rodrigc: On Python 2 and 3, `str(foo)` called `foo.__str__()`.
On Python 2, `bytes` is an alias for `str`, so on Python 2 `bytes(foo)`  calls `foo.__str__()`.
On Python 3, `bytes` is distinct from `str`, and `bytes(foo)`  calls `foo.__bytes__()`.

I've been burned by this before, so str(foo).encode("ascii") is safe to use on Python 2 and 3.

rodrigc: During my Python 3 porting in Twisted, I've gradually started moving away from using `networkString` and `nativeString`.  They were useful to get the original Twisted port to Python 3 off the ground, but now I prefer to be more explicit about the encode/decode being used.

src/twisted/mail/pop3client.py
rodrigc: Sure, I can do that.

src/twisted/mail/protocols.py
rodrigc: See above, I'm moving away from using `networkString`

src/twisted/mail/test/test_imap.py
rodrigc: Yes, I think you are right.

src/twisted/mail/pop3.py
rodrigc: Good point.

src/twisted/mail/pop3.py
rodrigc: Good point

src/twisted/mail/test/test_pop3client.py
rodrigc: Good point.

src/twisted/mail/pop3.py
rodrigc: I'm moving away from `networkString`

src/twisted/mail/pop3.py
rodrigc: `intToBytes` doesn't quite work well here, if the value is a letter for instance or some random class for instance.

src/twisted/mail/pop3.py
rodrigc: No, see above.

src/twisted/mail/test/test_imap.py
rodrigc: No, this needs to be like this because in imap4.py, there is this:

```
        if line:
            # Too many arguments
            raise IllegalClientResponse("Too many arguments for command: " + repr(line))
```

so the value of this is different on Python 2 and Python 3.
Maybe this can be cleaned up in a follow-up PR, but I don't want to fix that here.

src/twisted/mail/imap4.py
evilham: I see :). It makes sense, had assumed that `*String` were "the preferred way" of doing this.
I'd have been burned by the `bytes` thing in Python 2/3, thank you for explaining!

evilham (approved):
Looks good to me (tm) :)

rodrigc:
@evilham Thanks.  The Python 2/3 bytes/unicode stuff is not intuitively obvious for non-trivial codebases like Twisted for newcomers like I was. :)  I learned a lot though.  You might be interested in this presentation which I gave:

https://www.slideshare.net/CraigRodrigues1/the-onward-journey-porting-twisted-to-python-3 https://www.youtube.com/watch?v=ZbYvtibFymM

Zectbumo:
I put in a pull request #11898

PR #881, opened 2017-08-28
Port mailmail to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/8487

Moved from: https://github.com/twisted/twisted/pull/738


rodrigc:
@adiroiban can you review this?  This is moved from https://github.com/twisted/twisted/pull/738 which you did an initial review of.

adiroiban:
I see that some tests are failing and diff coverage is not 100%.

I think that if this should be reviewed with missing coverage and failing tests, the review should be requested with a comment explaining why the review should be done, even if branch is red :)

thanks!

rodrigc:
@adiroiban: with respect to coverage, if I look at https://codecov.io/gh/twisted/twisted/pull/881/diff , I see that this patch has 100% patch coverage.  Looks like maybe there is a problem from codecov.io updating the status of the PR in github.

With regard to the builder failure, I've seen one of the FreeBSD builders intermittently failing (unrelated to the code change in this PR).
I forced a rebuild on that builder, and it passed.
So now, I see that this builder has 38 out of 38 passing checks.

src/twisted/mail/scripts/mailmail.py
```diff
@@ -37,13 +39,15 @@
...
-    sys.stderr.write(str(message) % args + '\n')
+_logObserver = textFileLogObserver(sys.stderr)
+_log = Logger(observer=_logObserver)
+
 
 
 
 class Options:
     """
+
```
adiroiban: Minor comment. Why the empty line. Why not add a docstring ... or leave it as it is?

src/twisted/mail/scripts/mailmail.py
```diff
@@ -142,7 +147,7 @@ def parseOptions(argv):
         'date': [],
     }
 
-    buffer = StringIO.StringIO()
+    buffer = NativeStringIO()
```
adiroiban, line 151: I don't know how `email.utils.parseaddr` works, but do we need native string here?

Can we always go with bytes or unicode?

But if you don't know and tests are passing I think that is fine.

If someone cares about Unicode handling, she/he can add Unicode specific tests and see if it breaks or not.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+        data = out.getvalue()
+        self.assertEqual(data, "mailmail version: {}\n".format(version))
+
+
+    def test_backGroundDelivery(self):
+        """
+        The I{-odb} flag specifies background delivery.
+        """
+        stdin = NativeStringIO('\n')
+        self.patch(sys, 'stdin', stdin)
```
adiroiban, line 134: do we need to patch stdin? I don't see any check on the patched stdin and I don't see how it affects the test

re-reading the code I see why this is needed, but maybe is better to refactor these 2 lines into something like

self.sendNoEmail() or self.prepareEmptySend or something like that.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+        The I{-t} flags specifies that recipients should be obtained
+        from headers.
+        """
+        stdin = NativeStringIO(
+            'To: Curly <invaliduser2@example.com>\n'
+            'Cc: Larry <invaliduser1@example.com>\n'
+            'Bcc: Moe <invaliduser3@example.com>\n'
+            '\n'
+            'Oh, a wise guy?\n')
+        self.patch(sys, 'stdin', stdin)
```
adiroiban, line 160: maybe the self.path() opeartion can be extracted into a DSL and have a method like self.sendMessage() or self.messageReceived()

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+        The I{-F} flags specifies the From: value.
+        """
+        self.patch(sys, 'stderr', self.out)
+        stdin = NativeStringIO(
+            'To: invaliduser2@example.com\n'
+            'Subject: A wise guy?\n\n')
+        self.patch(sys, 'stdin', stdin)
+        o = parseOptions(["-F", "Larry <invaliduser1@example.com>", "-t"])
+        self.assertEqual(o.sender, "Larry <invaliduser1@example.com>")
+        # Test that -F flag is overridden by From: value in header
```
adiroiban: I think that this should be a separate test with a dedicated docstring.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+            '\n'
+            'Oh, a wise guy?\n')
+        self.patch(sys, 'stdin', stdin)
+        o = parseOptions("-t")
+        self.assertEqual(len(o.to), 3)
+
+
+    def test_setFrom(self):
+        """
+        The I{-F} flags specifies the From: value.
```
adiroiban: I think that this needs to be extended.

When message has no `From: ` header, one can be set using the -I option

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+        sys.stdin = NativeStringIO(
+            'To: Curly <invaliduser4@example.com>\n'
+            'From: Shemp <invaliduser4@example.com>\n')
+        o = parseOptions(["-F", "Groucho <invaliduser5@example.com>", "-t"])
+        self.assertEqual(o.sender, "invaliduser4@example.com")
+
+
+    def test_run(self):
+        """
+        L{twisted.mail.scripts.mailmail.run}
```
adiroiban: This docstring does not add any information :) pydoctor should be smart and you can use just `L{mailmail.run}`

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+        """
+        self.addCleanup(setattr, sys, 'argv', sys.argv)
+        self.addCleanup(setattr, sys, 'stdin', sys.stdin)
+        self.patch(sys, 'stderr', self.out)
+        sys.argv = ("test_mailmail.py", "invaliduser2@example.com", "-oep")
+        sys.stdin = NativeStringIO('\n')
+        mailmail.run()
+
+    if platformType == "win32":
+        test_run.skip = "win32 lacks support for getuid()"
```
adiroiban: maybe this should be: mailman is not supported on Windows... but with the missing docstring I don't know what is the purpose of this teset

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +102,252 @@ def test_copyAliasedSender(self):
...
+
+
+    def test_readInvalidConfig(self):
+        """
+        Read an invalid configuration file.
```
adiroiban: the docstring does not add any new information on top of what is expressed by `test_readInvalidConfig` ... so I consider that this test is missing the docstring and I don't know what to check :)

adiroiban (changes requested):
Changes looks good.

Just a few minor comments.

Please look at updating the docstring for the tests.

One test docstring are ok, this can be merged as it is a big improvement over what we have in trunk.

Thanks!

src/twisted/mail/scripts/mailmail.py
rodrigc: email.utils.parseaddr() fails with bytes on Python 3, so we need to use native string.
Also, a lot of mailmail is based on stdout and stdin, so we need to use NativeStringIO when overriding sys.stdout and sys.stdin in order for the code to work on Python 2 and 3.

src/twisted/mail/test/test_mailmail.py
rodrigc: we need to patch stdin, because of this code in mailmail:

```
o.body = NativeStringIO(buffer.getvalue() + sys.stdin.read())
```
If you don't patch stdin, then the test just hangs, because it is waiting for stdin input.

src/twisted/mail/test/test_mailmail.py
rodrigc: OK

src/twisted/mail/test/test_mailmail.py
rodrigc: OK

rodrigc:
@adiroiban I have incorporated your feedback, and also updated this PR to fix new errors introduced by pycodestyle.  Can this now be merged?

src/twisted/mail/scripts/mailmail.py
```diff
@@ -8,20 +8,22 @@
...
-
 try:
-    import cStringIO as StringIO
-except:
-    import StringIO
+    # Python 3
+    from configparser import ConfigParser
+except ImportError:
+    # Python 2
+    from ConfigParser import ConfigParser
```
adiroiban, line 20: This is fine... minor comment: Can we just use six.moves ?

src/twisted/mail/scripts/mailmail.py
```diff
@@ -283,9 +296,10 @@ def loadConfig(path):
         for (host, up) in p.items('identity'):
             parts = up.split(':', 1)
             if len(parts) != 2:
-                log("Illegal entry in [identity] section: %s", up)
+                _log.error("Illegal entry in [identity] section: {section}",
+                           section=up)
                 continue
-            p.identities[host] = parts
+            c.identities[host] = parts
```
adiroiban, line 302: Just a question... why is this changed from p to c ? Did we had a bug here?

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+deny=37,38
+order=deny""")
+        self.assertEqual(config.allowUIDs, [35, 36])
+        self.assertEqual(config.denyUIDs, [37, 38])
+
+        config = self._loadConfig("""
+[groupaccess]
+allow=gid1,41
+order=allow""")
+        self.assertEqual(config.allowGIDs, [41])
```
adiroiban: please consider splitting into separate tests with dedicated docstrings.
I am not sure why gid1 is ignored here... I guess because is not a number, but please be explicit in the tests.. as there is no code to test that the test code is valid :)

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+                         "invalidgid1")
+        self.assertRegex(self.out.getvalue(),
+                         'Illegal entry in \\[identity\\] section: funny')
+
+    if platformType == "win32":
+        test_readInvalidConfig.skip = ("mailmail.run() does not work on win32"
+                                       " due to lack of support for getuid()")
+
+
+    def _loadConfig(self, config):
```
adiroiban: may rename this to `getConfigurationFromText` ... or something like that

No need to make it private as this is testing code.

For code which should be reused by multiple tests is better to have it defined as public.

We can keep private methods only for the code which is specific for a single test and should not be reused anywhere else.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+        """
+        stdin = NativeStringIO(
+            'To: Curly <invaliduser2@example.com>\n'
+            'Cc: Larry <invaliduser1@example.com>\n'
+            'Bcc: Moe <invaliduser3@example.com>\n'
+            '\n'
+            'Oh, a wise guy?\n')
+        self.patch(sys, 'stdin', stdin)
+        o = parseOptions("-t")
+        self.assertEqual(len(o.to), 3)
```
adiroiban, line 162: why check only the length and not the actual value?

maybe, due to a bug BCC is parsed as an empty string... and the lenght would still be 3

src/twisted/mail/test/test_mailmail.py
```diff
@@ -19,17 +29,43 @@ class OptionsTests(TestCase):
...
+        """
+        Override some things in mailmail, so that we capture C{stdout},
+        and do not call L{reactor.stop}.
+        """
+        self.out = NativeStringIO()
+        # Override the mailmail logger, so we capture stderr output
+        from twisted.logger import textFileLogObserver, Logger
+        logObserver = textFileLogObserver(self.out)
+        self.patch(mailmail, '_log', Logger(observer=logObserver))
+        self.host = None
```
adiroiban, line 44: Do we need this as an instance member? Same for self.ident.

Is this patching required for every test from this testcase?
Maybe this can be refactored it into a method (ex `patchMailman`) and explicitly call it for the tests which need the patching.

Also, `self.out` looks more like `self.mailmanLogOuput`.
Please be more explicit about what `out` is. ... stdout / log output ... or network output

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+    def test_senderror(self):
+        """
+        L{twisted.mail.scripts.mailmail.senderror} sends mail back to the
+        sender if an error occurs while sending mail to the recipient.
+        """
+        def sendmail(host, sender, recipient, body):
+            self.assertRegex(sender, "postmaster@")
+            self.assertEqual(recipient, ["testsender"])
+            self.assertRegex(body.getvalue(), "ValueError")
+            return Deferred()
```
adiroiban, line 373: why return a deferred here?

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+        self.assertIsNone(mailmail.failed)
+
+    if platformType == "win32":
+        test_run.skip = ("mailmail.run() does not work on win32 due to "
+                         "lack of support for getuid()")
+
+
+    def test_readInvalidConfig(self):
+        """
+        Read an invalid configuration file.  Error messages
```
adiroiban: Minor comment
`Read an invalid configuration file.`  for me does not add any new information on top of why I can understand by reading `def test_readInvalidConfig(self):`

Maybe:

```
Error messages for illegal UID value, illegal GID value, and illegal identity entry are sent to stderr.
```

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+        """
+        stdin = NativeStringIO(
+            'To: Curly <invaliduser4@example.com>\n'
+            'From: Shemp <invaliduser4@example.com>\n')
+        self.patch(sys, 'stdin', stdin)
+        o = parseOptions(["-F", "Groucho <invaliduser5@example.com>", "-t"])
+        self.assertEqual(o.sender, "invaliduser4@example.com")
+
+
+    def test_run(self):
```
adiroiban: maybe rename to `test_runErrorsToStderr`... but I don't understand the scope of this test.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
...
+        """
+        The I{-odb} flag specifies background delivery.
+        """
+        stdin = NativeStringIO('\n')
+        self.patch(sys, 'stdin', stdin)
+        o = parseOptions("-odb")
+        self.assertTrue(o.background)
+
+
+    def test_foreGroundDelivery(self):
```
adiroiban: just curious. why these cases test_fore**G**roundDelivery ? :) same for backgroud

src/twisted/mail/newsfragments/8487.bugfix
```diff
@@ -0,0 +1 @@
+twistd.mail.scripts.mailmail has been ported to Python 3
```
adiroiban: Is this a bug? I would say that is a feature :) Don't forget the full stop at the end to be explicit that the sentence ends there and that it was not accidentally left unfinished.

src/twisted/mail/test/test_mailmail.py
```diff
@@ -73,3 +109,271 @@ def test_copyAliasedSender(self):
         """
         exc = self.assertRaises(SystemExit, parseOptions, ['-om'])
         self.assertEqual(exc.args, ("Unsupported option.",))
+
+
+    def test_version(self):
+        """
+        The I{--version} option displays the version and raises
+        L{SystemExit}.
```
adiroiban: Maybe: raises  L{SystemExit with `None` as the exit code so.

In this way, we don't need to read this implementation detail as a comment inside the test's implementation.

adiroiban (approved):
Thanks a lot for working on this.
Changes look good... please check the change from **p** to **c**. That one is important.


I am approving this as this is better than what we have now in trunk.

Not sure who is using this... so I would say that spending more time on this is a waste of time :)

If you are using this in production, then I am -1 as I would say that it needs a bit more cleanup. 

@rodrigc I know that you try to help with Python3 porting, but I think that old code which is not used, not documented and without good test coverage should just die :)

-------------

I know that pycodestyle now complains about the indentation.
I prefer this indentation as it help with long lines.

```
self.someLongVariable = someInstance.someMethod(
    someArgument="some_input_value",
    otherArgument=otherInputValue,
    )
```

vs ... and this line has 81 chars

````
self.someLongVariable = someInstance.someMethod(someArgument="some_input_value",
                                                otherArgument=otherInputValue)
```

src/twisted/mail/scripts/mailmail.py
rodrigc: Well I think using six would be a good thing, but the original Twisted Python 3 porting effort avoided the use of third party python3 porting libraries.  In hindsight I think that was a mistaken strategy, but what is done is done.

src/twisted/mail/scripts/mailmail.py
rodrigc: I found this bug as I added more test coverage, so I fixed it here as you can see.

src/twisted/mail/newsfragments/8487.bugfix
rodrigc: Well I could argue that since no new functionality is being added by this PR,
the fact that mailmail does not work on Python 3 is a bug, so this PR is a bugfix.

src/twisted/mail/scripts/mailmail.py
adiroiban: well, we already depend on six...as I guess that all our dependencies depend on six :)

I think that is not to late to start using six, rather than going further on this wrong way :)

src/twisted/mail/scripts/mailmail.py
adiroiban: I was expecting to see some info about this bug in the release notes...as I guess that this also affects Pyhon2.
The current release notes only talk about python3.

PR #882, opened 2017-09-01
Fix twisted.trial._dist.test.test_workertrial on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8944


moshez:
Updated branch to fix pydoctor failure. If build passes, it should be good.

moshez (approved):

PR #884, opened 2017-09-01
docs.zope.org is dead, so use zopeinterface.readthedocs.io instead.

markrwilliams:
https://twistedmatrix.com/trac/ticket/9276


rodrigc (approved):
I verified this link is correct.
Thanks for fixing.

PR #886, opened 2017-09-04
Fix --auth=unix when used to run a conch server on Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9130

This fixes the following on Python 3, tested on Linux Fedora 24:

```
twistd -n conch -d /etc/ssh/ --auth=unix
```

### DIFF
--- a/src/twisted/conch/unix.py
+++ b/src/twisted/conch/unix.py
@@ -29,7 +29,7 @@ from twisted.conch.interfaces import ISession, ISFTPServer, ISFTPFile
 from twisted.cred import portal
 from twisted.internet.error import ProcessExitedAlready
 from twisted.python import components, log
-from twisted.python.compat import nativeString
+from twisted.python.compat import _bytesChr as chr, nativeString
 
 try:
     import utmp
--- a/src/twisted/cred/test/test_strcred.py
+++ b/src/twisted/cred/test/test_strcred.py
@@ -270,21 +270,11 @@ class CryptTests(unittest.TestCase):
         """
         password = "sample password ^%$"
 
-        # Use MD5 method to encrypt password
-        # The correct string was manually generated with crypt.crypt()
-        cryptMd5Correct = '$1NZEcnLLqFwo'
-        cryptMd5Incorrect = '$1NZEcnwo'
-        # Use traditional crypt method to encrypt password
-        # The correct string was manually generated with crypt.crypt()
-        cryptCryptCorrect = 'SSYv0E4TgOn8U'
-        cryptCryptIncorrect = 'SSYv04ETOgn8U'
-        self.assertTrue(cred_unix.verifyCryptedPassword(cryptMd5Correct,
+        cryptedCorrect = crypt.crypt(password, None)
+        cryptedIncorrect = "$1x1234"
+        self.assertTrue(cred_unix.verifyCryptedPassword(cryptedCorrect,
                                                         password))
-        self.assertFalse(cred_unix.verifyCryptedPassword(cryptMd5Incorrect,
-                                                        password))
-        self.assertTrue(cred_unix.verifyCryptedPassword(cryptCryptCorrect,
-                                                        password))
-        self.assertFalse(cred_unix.verifyCryptedPassword(cryptCryptIncorrect,
+        self.assertFalse(cred_unix.verifyCryptedPassword(cryptedIncorrect,
                                                         password))
 
 
--- a/src/twisted/cred/test/test_strcred.py
+++ b/src/twisted/cred/test/test_strcred.py
@@ -661,10 +661,3 @@ class LimitingInterfacesTests(unittest.TestCase):
         self.assertRaises(SystemExit, options.parseOptions,
                           ['--help-auth-type', 'anonymous'])
         self.assertIn(strcred.notSupportedWarning, newStdout.getvalue())
-
-
-__all__ = [
-    "CheckerOptionsTests", "FileDBCheckerTests", "LimitingInterfacesTests",
-    "SSHCheckerTests", "UnixCheckerTests", "AnonymousCheckerTests",
-    "MemoryCheckerTests", "StrcredFunctionsTests", "PublicAPITests"
-]
--- a/src/twisted/cred/test/test_strcred.py
+++ b/src/twisted/cred/test/test_strcred.py
@@ -14,7 +14,7 @@ from twisted.trial import unittest
 from twisted.cred import credentials, checkers, error, strcred
 from twisted.plugins import cred_file, cred_anonymous, cred_unix
 from twisted.python import usage
-from twisted.python.compat import _PY3, NativeStringIO
+from twisted.python.compat import NativeStringIO
 from twisted.python.filepath import FilePath
 from twisted.python.fakepwd import UserDatabase
 from twisted.python.reflect import requireModule
--- a/src/twisted/cred/test/test_strcred.py
+++ b/src/twisted/cred/test/test_strcred.py
@@ -13,15 +13,11 @@ from twisted import plugin
 from twisted.trial import unittest
 from twisted.cred import credentials, checkers, error, strcred
 from twisted.plugins import cred_file, cred_anonymous
-from twisted.python import usage, compat
+from twisted.python import usage
+from twisted.python.compat import NativeStringIO
 from twisted.python.filepath import FilePath
 from twisted.python.fakepwd import UserDatabase
 from twisted.python.reflect import requireModule
-
-if compat._PY3:
-    from io import StringIO
-else:
-    from io import BytesIO as StringIO
 
 try:
     import crypt
@@ -338,7 +334,7 @@ class FileDBCheckerTests(unittest.TestCase):
         should produce a warning.
         """
         oldOutput = cred_file.theFileCheckerFactory.errorOutput
-        newOutput = StringIO()
+        newOutput = NativeStringIO()
         cred_file.theFileCheckerFactory.errorOutput = newOutput
         strcred.makeChecker('file:' + self._fakeFilename())
         cred_file.theFileCheckerFactory.errorOutput = oldOutput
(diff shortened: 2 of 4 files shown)


src/twisted/plugins/cred_unix.py
```diff
@@ -65,7 +68,9 @@ def checkPwd(self, pwd, username, password):
 
     def checkSpwd(self, spwd, username, password):
         try:
-            cryptedPass = spwd.getspnam(username)[1]
+            if not isinstance(username, StringType):
+                username = username.decode('utf-8')
+            cryptedPass = spwd.getspnam(username).sp_pwdp
```
jlitzingerdev: @rodrigc Sorry for jumping in randomly here, but I'd looked at this for a moment.  Since you have a PR open I thought I'd help however I could.

Looks like the sp_pwd attribute is different between python 2/3.  `sp_pwd` vs `sp_pwdp`
https://docs.python.org/3.5/library/spwd.html
https://docs.python.org/2/library/spwd.html

Which gives me the following when I run on your branch with p2 (exception only)
```
	  File "/home/jlitzing/dev/twisted/twisted-fork/src/twisted/internet/defer.py", line 150, in maybeDeferred
	    result = f(*args, **kw)
	  File "/home/jlitzing/dev/twisted/twisted-fork/src/twisted/plugins/cred_unix.py", line 100, in requestAvatarId
	    checked = self.checkSpwd(spwd, username, password)
	  File "/home/jlitzing/dev/twisted/twisted-fork/src/twisted/plugins/cred_unix.py", line 73, in checkSpwd
	    cryptedPass = spwd.getspnam(username).sp_pwdp
	exceptions.AttributeError: 'spwd.struct_spwd' object has no attribute 'sp_pwdp'
```

src/twisted/cred/test/test_strcred.py
```diff
@@ -264,10 +257,55 @@ def test_unixCheckerFailsPassword(self):
...
+    if not crypt:
+        skip = "Required module is unavailable: crypt"
+
+    def test_verifyCryptedPassword(self):
+        """
+        L{cred_unix.verifyCryptedPassword}
+        """
+        password = "sample password ^%$"
+
+        cryptedCorrect = crypt.crypt(password, None)
```
jlitzingerdev: @rodrigc Looks like another p2/3 change.  The salt parameter isn't optional in p2.  I submitted a PR with one possible fix:
https://github.com/twisted/twisted/pull/887

src/twisted/plugins/cred_unix.py
rodrigc: Fixed.

src/twisted/cred/test/test_strcred.py
rodrigc: Fixed

rodrigc:
@jlitzingerdev since you expressed interest in this, can you code review this change,
as per the review process: https://twistedmatrix.com/trac/wiki/ReviewProcess#Reviewers:Howtoreviewachange ?

jlitzingerdev:
@rodrigc No problem.  I have to step away for a few hours but I can do that later this evening.

src/twisted/conch/unix.py
```diff
@@ -29,7 +29,7 @@
 from twisted.cred import portal
 from twisted.internet.error import ProcessExitedAlready
 from twisted.python import components, log
-from twisted.python.compat import nativeString
+from twisted.python.compat import _bytesChr as chr, nativeString
```
jlitzingerdev, line 32: Should the built-in chr() always have python 2 semantics (certainly yes in the existing use case) in this module?  That seems like it might cause confusion down the road and be easy to miss.  Why not just use _bytesChr() in the one location where chr() is currently used?

src/twisted/plugins/cred_unix.py
```diff
@@ -65,7 +68,14 @@ def checkPwd(self, pwd, username, password):
 
     def checkSpwd(self, spwd, username, password):
         try:
-            cryptedPass = spwd.getspnam(username)[1]
+            if not isinstance(username, StringType):
```
jlitzingerdev: This comment is preference, meaning I'm not advocating for it, but I wanted to point it out for consideration.

In nearly all cases, I dislike magic numbers for an index.  This is one of the rare cases where I actually prefer it.  My reasoning is that because the indexes are explicitly described in the standard library docs they provide a succinct way to access the encrypted password without caring about the python version.  Unless subsequent viewers of this code are familiar with getspnam() then it's likely they'll go to the std library docs anyway.  Were the code to use the old indexing I would suggest a comment describing where the index is documented.

src/twisted/plugins/cred_unix.py
```diff
@@ -52,7 +53,9 @@ class UNIXChecker(object):
 
     def checkPwd(self, pwd, username, password):
         try:
```
jlitzingerdev: This applies to all the methods in this class -- its a shame there are no method docstrings.  The fact that so much of Twisted is documented is one of (my opinion) its strengths.  Should lack of docsstrings live in another ticket?

src/twisted/plugins/cred_unix.py
```diff
@@ -17,22 +17,23 @@
 from twisted.cred.credentials import IUsernamePassword
 from twisted.cred.error import UnauthorizedLogin
 from twisted.internet import defer
+from twisted.python.compat import StringType
 
 
 
 def verifyCryptedPassword(crypted, pw):
-    if crypted[0] == '$': # md5_crypt encrypted
```
jlitzingerdev, line 24: Is there any chance this breaks MD5  encrypted passwords on python 2?  It looks like the parsing here relates to a GLibc implemenation detail (http://man7.org/linux/man-pages/man3/crypt.3.html), and the crypt module in p2 doesn't have those methods.  A secondary question is of course does it matter (security vs compatibility argument)?  I'm not overly familiar with the crypt API, so I don't have test to add right this second, but it might be worth trying.

jlitzingerdev (commented):
This looks good, nice work!  I have two concerns -- MD5 passwords and the renaming of chr() to _bytesChr for the entire module.  The remaining comments were nits about lack of docstrings (which existed prior to this fix), and pure opinion on accessing spwd.

src/twisted/conch/unix.py
rodrigc: There are several places in the code where replacing chr() with byteschr() has been done.
For example, in conch, it has been done in these files:

```
twisted/conch/insults/window.py
twisted/conch/ssh/connection.py
twisted/conch/ssh/session.py
twisted/conch/ssh/transport.py
twisted/conch/ssh/userauth.py
twisted/conch/telnet.py
twisted/conch/test/test_transport.py
twisted/conch/test/test_userauth.py
```

In twisted/conch/unix.py, chr() is only used on one line in that file, so I think it
is OK to replace it for this module.  The pattern has been to ``import _bytesChr as chr`` and then use ``chr()`` so I am following that pattern here.

src/twisted/plugins/cred_unix.py
rodrigc: I'm going to disagree with you on that one.  If someone has to look into the standard library,
these fields are documented quite visibly:

https://docs.python.org/2/library/pwd.html
https://docs.python.org/2/library/spwd.html

and Python 3:

https://docs.python.org/3/library/pwd.html
https://docs.python.org/3/library/spwd.html

Having the field names spelled out is more readable than some index offset.
If someone needs to look up the field, it is quite well documented.

src/twisted/conch/unix.py
jlitzingerdev: Fair enough, I didn't search for existing instances of the same pattern (should have), so I agree.

src/twisted/plugins/cred_unix.py
jlitzingerdev: No worries, like I said, this was opinion on my part.

src/twisted/plugins/cred_unix.py
rodrigc: There are a few functions in this file which do not have docstrings.  I'm not going to add docstrings for all of them, but I agree with you that this could be improved, so I added docstring for `checkPwd`, `checkSpwd`, and `verifyCryptedPassword`.

Those are important functions in this file, so it doesn't hurt to document them a bit better.

src/twisted/plugins/cred_unix.py
rodrigc: On my Python 3 system, I ran this program to generate an MD5 password:

```
import crypt
crypt.crypt("password", crypt.METHOD_MD5)
```

I took the string which was generated: `'$1d2n7Q0.r54s'`

and ran the following program in a Python 2.6.9 virtualenv:

```
import crypt
crypted =  "$1d2n7Q0.r54s"
crypt.crypt("password", "$1d2n7Q0.r54s") == crypted
```

and the Python 2.6.9 program returned `True`.  So this doesn't break
parsing of MD5 passwords.

rodrigc: Other thing to note, is that in the Python 2 documentation:
https://docs.python.org/2/library/crypt.html

there is this:

```
Since a few crypt(3) extensions allow different values, with different sizes in the salt, it is recommended to use the full crypted password as salt when checking for a password.
```

rodrigc: Also, to confirm with the implementation, this is where crypt is called in CPython 2.7:

https://github.com/python/cpython/blob/2.7/Modules/cryptmodule.c#L27

and this the implementation of crypt inside glibc:

https://sourceware.org/git/?p=glibc.git;a=blob;f=crypt/crypt-entry.c;h=2271665af2d21d2aec61ebe712ffc946c035e61e;hb=HEAD#l74

If you notice in the glibc code, the crypt() implementation parses the first
few characters of the salt, and depending on what prefix is there,
it hands it off to the code which parses MD5, SHA512, SHA256, etc.

So the change in this PR is compatible with the underlying implementation.

jlitzingerdev (approved):
All of my concerns have been thoroughly addressed by Craig's comments here.  Nicely done!

PR #888, opened 2017-09-06
Clarify index pages in documentation by putting protocol names in parentheses

rodrigc:
http://twistedmatrix.com/trac/ticket/9204

@alex mentioned in ticket 9204:

> https://twistedmatrix.com/documents/current/index.html
>
> It uses the internal names for twisted packages, which are opaque to newcomers, if I'm not an
> existing twisted person I don't know that Twisted Names is DNS. Even doing as much as putting
> the protocol in parenthesis would go a long way: Twisted Conch (SSH and Telnet).

I agree with this.

### DIFF
--- a/docs/words/index.rst
+++ b/docs/words/index.rst
@@ -3,8 +3,8 @@
 :LastChangedRevision: $LastChangedRevision$
 :LastChangedBy: $LastChangedBy$
 
-Twisted Words
-=============
+Twisted Words (IRC, XMPP)
+=========================
 
 .. toctree::
    :hidden:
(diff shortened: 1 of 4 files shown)


docs/words/index.rst
```diff
@@ -3,8 +3,8 @@
 :LastChangedRevision: $LastChangedRevision$
 :LastChangedBy: $LastChangedBy$
 
-Twisted Words
-=============
+Twisted Words (IRC, XMPP)
```
alex: Consistency nit, write this as "IRC and XMPP" like the other lines.

alex (approved):
Besides that nit, looks great to me!

docs/words/index.rst
rodrigc: Fixed

PR #889, opened 2017-09-08
[#8149] Update zope.interface dependency

hawkowl:
https://twistedmatrix.com/trac/ticket/8149

We let people use Twisted with these versions, but we haven't actually tested any of them for years and years. I picked 4.4.2 as it's decently stable (3+ months old), supports Python 3.6 (which the majority of our Python 3 installs are), and is heavily tested by our test suite. There shouldn't be any downside to requiring this, since we require relatively new things as base dependencies already, and z.i isn't that much of a burden to have new.


rodrigc (approved):
This looks good.
A few imports need to be removed from test_twisted.py to appease pyflakes, but other than that, this is fine.

Fix the pyflakes warnings here: https://travis-ci.org/twisted/twisted/jobs/273252999#L592

then merge.

PR #890, opened 2017-09-09
[#8848] Update some Mac OSX -> macOS references in setuptools extras + docs

hawkowl:
https://twistedmatrix.com/trac/ticket/8848


rodrigc (approved):
This isn't super-critical because the name "OS X" has been around for about 16 years, and it is quite common knowledge what platform is being referred to.

However, this change doesn't hurt anything, so:
Approved.

One suggestion that I would make is for you to clean up references to "OS X" in various comments in the Twisted codebase.

If I do:

```
egrep -ri "OS X|OS X" *
```

I see a bunch of comments that refer to *MacOS X" or *MacOSX" .  Please clean up the references
_in comments only_, i.e. do not change variables or function names.  And then commit.

PR #891, opened 2017-09-12
[#7922] Add parent to IService

moshez:
Addresses #7922


moshez:
@adiroiban I think you already approved, but I want to verify, since it has been two years.

src/twisted/application/service.py
```diff
@@ -92,13 +92,17 @@ class IService(Interface):
...
     Run start-up and shut-down code at the appropriate times.
-
-    @type name:            C{string}
-    @ivar name:            The name of the service (or None)
-    @type running:         C{boolean}
-    @ivar running:         Whether the service is running.
     """
 
+    name = Attribute(
+        "A C{string} which is the name of the service or c{None}.")
```
adiroiban: This should be C{None}. with capital C.

But I don't think that this documentation will work.

We should still document them using epydoc as docstrings.

src/twisted/newsfragments/7922.bugfix
```diff
@@ -0,0 +1 @@
+IService now properly requires the 'runner', 'name' and 'parent' attributes.
```
adiroiban: Please use the full python path for the IService reference twisted.application.service.IService

there might be multiple IService interfaces, and with the full path it is clear what was changed.

adiroiban (dismissed):
I know that I have approved it 2 years ago... but I don't remember why :(

For reference https://www.mail-archive.com/twisted-python@twistedmatrix.com/msg10613.html

-----------

So I think that the API change is OK, but I don't think that the documentation is right.

Please use standard epydoc markup and not Zope interface conventions.

Thanks!

src/twisted/application/test/test_service.py
```diff
@@ -7,14 +7,125 @@
...
+    """
+    Implement IService except for the attributes.
+    """
+    def setName(self, name):
+        """
+        See L{twisted.application.service.IService}.
+
+        @param name: ignored
+        """
+        pass
```
exarkun: `pass` is superfluous here.

src/twisted/application/test/test_service.py
```diff
@@ -7,14 +7,125 @@
...
+
+    def test_noRunning(self):
+        """
+        AlmostService with no running does not implement IService.
+        """
+        del self.almostService.running
+        with self.assertRaises(BrokenImplementation):
+            verifyObject(IService, self.almostService)
+
+
```
exarkun: These tests all seem to set a poor example.  None of these attributes should be written from outside the implementation.

src/twisted/newsfragments/7922.bugfix
exarkun: "Properly"?  "Requires"?  Do those words mean something?

src/twisted/application/service.py
exarkun: There's no such type as "string".  It's either `bytes` or `unicode` and this is the place  to say which.

src/twisted/newsfragments/7922.bugfix
exarkun: in case it's not clear from the previous comment (which I admit is probably the most likely event), the point here is ... As a user, what am I expected to learn from these words?  Was my code broken because of the "improperness" of this interface?  Or was *my* code "improper" if it is _now_ broken as a result of this change?  For that matter, isn't it improper for Twisted to break my code without any backwards compatibility grace period?

And weren't these attributes "required" before?  I doubt you could create a usable implementation of `IService` without having them.

Also, there is no `runner` attribute: it's `running`.

src/twisted/application/test/test_service.py
exarkun: Put another way, should application code  consuming an `IService` mutate these attributes directly?  The tests weakly imply yes and nothing in the documentation suggests otherwise.  (In case you can't guess, my answer to this question is "absolutely not".)

moshez:
Addressed, I believe, @adiroiban and @exarkun comments. Builds still running, please review when green.

src/twisted/application/service.py
```diff
@@ -95,13 +95,13 @@ class IService(Interface):
     """
 
     name = Attribute(
-        "A C{string} which is the name of the service or c{None}.")
+        "A C{str} which is the name of the service or C{None}.")
```
markrwilliams: These should be `L{}`, not `C{}`.

src/twisted/application/service.py
```diff
@@ -95,13 +95,13 @@ class IService(Interface):
...
 
     running = Attribute(
         "A C{boolean} which indicates whether the service is running.")
 
     parent = Attribute(
-        "An C{IServiceCollection} which is the parent or c{None}.")
+        "An C{IServiceCollection} which is the parent or C{None}.")
```
markrwilliams: These should be `L{}`, not `C{}`.

src/twisted/application/test/test_service.py
```diff
@@ -23,13 +23,36 @@ class AlmostService(object):
     """
     Implement IService except for the attributes.
     """
+
+    def __init__(self, name, parent, running):
```
markrwilliams: Someday we should just start using `attrs`...

src/twisted/application/test/test_service.py
```diff
@@ -23,13 +23,36 @@ class AlmostService(object):
...
+        self.name = name
+        self.parent = parent
+        self.running = running
+
+    def makeInvalidByDeletingName(self):
+        """
+        Probably not a wise method to call.
```
markrwilliams: Why does it exist, then?  Please write a doc string that explains the purpose of the method.

src/twisted/application/test/test_service.py
```diff
@@ -23,13 +23,36 @@ class AlmostService(object):
...
+        """
+        del self.name
+
+    def makeInvalidByDeletingParent(self):
+        """
+        Probably not a wise method to call.
```
markrwilliams: Why does it exist, then? Please write a doc string that explains the purpose of the method.

src/twisted/application/test/test_service.py
```diff
@@ -23,13 +23,36 @@ class AlmostService(object):
...
+        """
+        del self.parent
+
+    def makeInvalidByDeletingRunning(self):
+        """
+        Probably not a wise method to call.
```
markrwilliams: Why does it exist, then? Please write a doc string that explains the purpose of the method.

markrwilliams (dismissed):
A few corrections to the docstring markup, and a request for clarification in the tests' docstrings.

glyph (approved):
The previous reviews seem substantive, and appear to have all been addressed.

Branch updated; please merge when the builders go green.

One note though, just because I didn't see any discussion of this: *technically*, adding attributes to an Interface like this is an "incompatible change", because if something were formally adhering to the previous interface (`verifyObject` returned True, etc), and they passed it to an object which only worked with a formally correct provider of this interface, this Twisted release will retroactively invalidate that previous implementation.

I'm not making that objection here because this really was just an oversight on the interface, akin to a Gross Violation Of Specifications mistake, where all the implementations really do require `parent` (and `name` and `running`) already, so it's not like this is actually a new requirement.  Just want to make sure this is noted for future changes.

PR #893, opened 2017-09-18
Fix errback documentation for twisted.names.client.getHostByName()

johnmaguire:
If we follow this to `Resolver._lookup()` and `Resolver.queryUDP()`, we can see that this will always return a `twisted.internet.defer.TimeoutError`, as `twisted.names.error.DNSQueryTimeoutError` is trapped by [`Resolver._reissue()`](https://github.com/twisted/twisted/blob/trunk/src/twisted/names/client.py#L277-L336). It seems to have been this way for a while, possibly over a decade.

### DIFF
--- a/src/twisted/names/client.py
+++ b/src/twisted/names/client.py
@@ -624,8 +624,9 @@ def getHostByName(name, timeout=None, effort=10):
     """
     Resolve a name to a valid ipv4 or ipv6 address.
 
-    Will errback with C{DNSQueryTimeoutError} on a timeout, C{DomainError} or
-    C{AuthoritativeDomainError} (or subclasses) on other errors.
+    Will errback with C{twisted.internet.defer.TimeoutError} on a timeout,
+    C{DomainError} or C{AuthoritativeDomainError} (or subclasses) on other
+    errors.
 
     @type name: C{str}
     @param name: DNS name to resolve.


rodrigc:
Thanks for your submission to Twisted.  In order to submit a patch to Twisted, you must follow the list of steps at:

You are mostly there, but are missing a few items from those steps:
   Step 1:  you need a ticket at https://twistedmatrix.com
   Step 6: you need a topfile newsfragment

Please fix this and notify here when you are done.

wiml:
Hmm, is it better to change the documentation or the implementation here? Would it make sense to change [the return case in `_reissue()`](https://github.com/twisted/twisted/blob/c20f01967702c6941ca796cc4837791754944c1c/src/twisted/names/client.py#L325) to simply return `reason`, which is already known to be a `DNSQueryTimeoutError`?

`DNSQueryTimeoutError` is a subclass of `defer.TimeoutError`, so it wouldn't break existing code that might be checking for `TimeoutError`. And it might occasionally be useful to be able to distinguish between a query timeout and some other timeout.

glyph:
@wiml I'm inclined to say the implementation should be fixed.  (When handling errbacks, `TimeoutError` and `CancelledError` are both things that most savvy consumers should be aware can happen to just about anything even if they're not documented.)

@JohnMaguire I'm going to close this just because there's been no activity in so long, but please feel free to reopen if you still have any interest.

PR #894, opened 2017-09-19
Fix "Twisted Web in 60 Seconds" examples on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9285


wsanchez:
This all seems… correct for the current API, but there are several cases where you take some text and UTF-8 encode it, but do nothing to ensure that the content encoding is UTF-8.  Is that a safe assumption?

rodrigc:
For these tutorial examples, I cut and pasted them directly from the tutorial and into a code editor, where I then ran these examples in Python 2 and Python 3, where the worked.
I think that is OK for these examples.

src/twisted/web/static.py
```diff
@@ -297,12 +297,15 @@ def getChild(self, path, request):
             if fpath is None:
                 return self.childNotFound
 
+        extension = fpath.splitext()[1]
+        if not isinstance(extension, str):
+            extension = extension.decode("utf-8")
```
markrwilliams: Shouldn't this be [`sys.getfilesystemencoding()`](https://docs.python.org/3/library/sys.html#sys.getfilesystemencoding) instead of `utf-8`?  Or better yet, [`os.fsdecode`](https://docs.python.org/3/library/os.html#os.fsdecode)?

markrwilliams (dismissed):
HTML does not default to UTF-8.  [The default `Content-Type` for `t.w.server.Requests` is `text/html`](https://github.com/twisted/twisted/blob/81c031cecd9e1039a4c66326bc2ee824cbd68555/src/twisted/web/server.py#L105), and [RFC 2854, which describes the `text/html`, has this to say about the media type's default encoding](https://tools.ietf.org/html/rfc2854#section-6):

>  The use of an explicit charset parameter is strongly recommended.
   While [[MIME](https://tools.ietf.org/html/rfc2046)] specifies "The default character set, which must be
   assumed in the absence of a charset parameter, is US-ASCII."  [[HTTP](https://tools.ietf.org/html/rfc2616)]
   Section 3.7.1, defines that "media subtypes of the 'text' type are
   defined to have a default charset value of 'ISO-8859-1'".  Section
   19.3 of [[HTTP](https://tools.ietf.org/html/rfc2616)] Using an explicit
   charset parameter will help avoid confusion.
>
>  Using an explicit charset parameter also takes into account that the
   overwhelming majority of deployed browsers are set to use something
   else than 'ISO-8859-1' as the default; the actual default is either a
   corporate character encoding or character encodings widely deployed
   in a certain national or regional community. For further
   considerations, please also see Section 5.2 of [[HTML40](https://www.w3.org/TR/1998/REC-html40-19980424/)].

In other words, HTTP 1.1's or HTML 4.0's default of ISO-8559-1 overrides MIME's default of ASCII according to RFC 2854.  Of course, RFC 2854's claim about HTTP 1.1's default encoding comes from the now deprecated [RFC 2616](https://tools.ietf.org/html/rfc2616#section-3.7.1), which states:

>  The "charset" parameter is used with some media types to define the
   character set (section 3.4) of the data. When no explicit charset
   parameter is provided by the sender, media subtypes of the "text"
   type are defined to have a default charset value of "ISO-8859-1" when
   received via HTTP. Data in character sets other than "ISO-8859-1" or
   its subsets MUST be labeled with an appropriate charset value. See
   section 3.4.1 for compatibility problems.

[RFC 7231, which supersedes RFC 2616's description of HTTP 1.1's content and semantics, removes the default charset for "text" media types.](https://tools.ietf.org/html/rfc7231#section-3.1.1.3).

HTML5, HTML 4's successor, _also_ doesn't specify a default, instead relying on a) the `Content-Type` header, b) a `meta` tag that includes an encoding, or c) [a complicated encoding detection heuristic](https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding).

Firefox and Chrome assume these responses are ISO-8559-1 in the absence of an explicit `charset`.  Users are liable to copy and paste these examples to build Unicode-sensitive applications in Python 3, and using UTF-8 will result in hard-to-debug mojibake.

Twisted Web should probably expose an API that allows users to set the encoding of their content.  In the absence of that, the examples should add the following preamble:

```
<!DOCTYPE html><html><head><meta charset='UTF-8'></head>
```

This turns the response into an HTML5 document whose encoding is UTF-8 (note that the closing `html` tag is optional).  The prose documentation should be edited to explain this preamble.

src/twisted/web/static.py
rodrigc: `os.fsdecode` doesn't exist on Python 2.7

markrwilliams: Would the condition ever be true on Python 2?

rodrigc: Probably not, unless somehow somone passed something in a way such that fpath was somehow unicode on Python 2.

rodrigc:
@wsanchez OK, I switch the examples to use "ascii"

markrwilliams:
@rodrigc Why not explicitly declare the encoding as UTF-8 with a `<meta>` tag?  Everything I've read strongly encourages explicitly declaring the encoding of HTML documents.

rodrigc:
These are simple examples, and I'm just going with the simplest way to get these examples
working.  Using ascii encoding seems fine, since these are really small code snippets.

src/twisted/web/static.py
```diff
@@ -297,12 +299,15 @@ def getChild(self, path, request):
             if fpath is None:
                 return self.childNotFound
 
+        extension = fpath.splitext()[1]
+        if not isinstance(extension, StringType):
+            extension = extension.decode(sys.getfilesystemencoding())
```
markrwilliams: In a previous review, I questioned the choice of UTF-8 here, because I though `fpath`'s type was determined by the file system's encoding.  I was wrong!

`fpath` is a child path of `site.File`, which in turn is a subclass of `twisted.python.filepath.FilePath`.  `FilePath`s [can represent either `unicode` or `bytes` paths, and accessing a `bytes` child from a `unicode` path results in a `bytes` path](http://twistedmatrix.com/documents/17.9.0/api/twisted.python.filepath.FilePath.html):

```
Python 3.6.1 (v3.6.1:69c0db5050, Apr  3 2017, 01:12:50) 
[GCC 6.3.0 20170221] on linux
>>> from twisted.python.filepath import FilePath
>>> path = FilePath('/unicode/path')
>>> path
FilePath('/unicode/path')
>>> path.child(b"somebytes")
FilePath(b'/unicode/path/somebytes')
```

That's exactly what's happening in this code when you run `twist web --path=/some/path`; `self.path` is `str`, because Python 3 decodes `sys.argv` according to your locale.

However, here in `static.File.getChild`, `fpath` is constructed by calling `FilePath.child` or `FilePath.preauthChild` on the `bytes` object `path`:

https://github.com/twisted/twisted/blob/73961416385261fc876bf8f03cc868424890a689/src/twisted/web/static.py#L289-L293

`path` is `bytes` because it's a _URL path_, and _generally_ [these are encoded with UTF-8](https://tools.ietf.org/html/rfc3986#section-2.5).

I now think the better solution is to do this:

```python
def getChild(self, path, request):
    if isinstance(self.path, twisted.python.compat.unicode):
        try:
            path = path.decode('utf-8')
        except UnicodeDecodeError:
            self._logger.failure("Could not decode {path!r} as UTF-8", path=path)
            return self.childNotFound  # or, better yet, return a 400
    ....
```

Since URL paths are usually UTF-8, this prevents `self.child` or `self.preauthChild` from decoding it with whatever `sys.getfilesystemencoding` returns.  It should also be backwards compatible.

markrwilliams (changes requested):

src/twisted/web/static.py
```diff
@@ -274,11 +273,33 @@ def directoryListing(self):
...
+        @return: A resource representing the requested file or
+            directory, or L{NoResource} if the path cannot be
+            accessed.
+        @rtype: An object that provides L{resource.IResource}.
         """
+        if isinstance(self.path, unicode):
+            try:
+                # Request calls urllib.unquote on each path segment,
+                # leaving us with raw bytes.
+                path = path.decode('utf-8')
```
rodrigc: @markrwilliams This part of the patch is wrong, and breaks several builders.
`unicode` does not have a `.decode` method

markrwilliams (approved):

src/twisted/web/static.py
markrwilliams: Whoops!  Thanks!

markrwilliams:
@rodrigc You should probably put this back up for review.

src/twisted/web/static.py
markrwilliams: These are the tests that appear to break:

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/test/test_tap.py#L97

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/test/test_tap.py#L107

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/test/test_tap.py#L120

They pass native strings as the `path` argument to `Resource.getChild`, which is wrong.  URL paths in twisted.web are always bytes:

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http.py#L803-L818

https://github.com/twisted/twisted/blob/trunk/src/twisted/web/server.py#L168-L197

`FilePath`s can describe either unicode _or_ bytes paths. `static.File` is a subclass of `FilePath`, so `self.path` here can be either unicode or bytes. This is supposed to decode the request path only if `self.path` describes a unicode path; the original code is correct, `.decode` needs to be there, and those tests need to use bytes instead of native strings for their request paths.

rodrigc: @markrwilliams You approved this PR.  Is it OK to merge this PR as is, and fix the test_tap.py usages in a follow-up PR, or do you want everything fixed in this PR?

markrwilliams:
@rodrigc Thanks for asking - I would prefer the tests be fixed and the `decode` brought back for this PR.

rodrigc:
For now, I'll merge this as is, and fix things in a subsequent PR, since there are some
interrelated changes that mark things hard to follow in a single PR.

PR #896, opened 2017-09-24
[9287] Add cwd parameter to ProcessMonitor.addProcess

moshez:
Add cwd parameter to ProcessMonitor.addProcess.

Since a naive addition would break backwards compatibility, this needed to change how the process information is stored. We also deprecate directly accessing the process information -- this was never intended functionality, but accidentally left as a public API.


exarkun:
No ticket number, no ticket link, no coherent commit messages?

Please bear in mind other people use and depend on this project, too.  It's frustrating to see this kind of development taking place.

moshez:
It's not up for review yet.

moshez:
https://twistedmatrix.com/trac/ticket/9287

moshez:
PR still needs all diffed lines to be covered by unit tests.

src/twisted/runner/test/test_procmon.py
```diff
@@ -130,6 +131,28 @@ def setUp(self):
...
+                           uid=1, gid=2, env={})
+        representation = repr(self.pm)
+        self.assertIn('foo', representation)
+        self.assertIn('1', representation)
+        self.assertIn('2', representation)
+
+
+    def test_simpleReprLooksGood(self):
+        """
+        Repr does not include unneeded details
```
adiroiban: How can I tell which are unneeded details?

There is no full stop. Is this doctring complete or we should expect more information :)

Together with test_reprLooksGood the docstring of these test don't add to much information on top of what is communicated by the test names.

I don't understand why are you checkin for paranthesis in the repr() result.

src/twisted/runner/test/test_procmon.py
```diff
@@ -503,6 +538,22 @@ def test_stopService(self):
         self.assertEqual({}, self.pm.protocols)
 
 
+    def test_restartAllRestartsOneProcess(self):
+        """
+        L{ProcessMonitor.restartAll} succeeds when there are no processes
```
adiroiban: I am confused by this test. in the test name there is **One** process, but the docstring talks about **no** processes and there is no full stop :)

src/twisted/runner/test/test_procmon.py
```diff
@@ -547,3 +598,56 @@ def test_stopServiceCleanupScheduledRestarts(self):
...
+        """
+        Create reactor and process monitor
+        """
+        self.reactor = DummyProcessReactor()
+        self.pm = ProcessMonitor(reactor=self.reactor)
+
+
+    def test_toTuple(self):
+        """
+        _Process.toTuple is deprecated
```
adiroiban: I think that this needs extra info. I think that I get what you are doing here, but is better to be explict.
The test name and docstring talks about **toTuple** but I don't see any call to toTuple here.

I don't that we should write the docsting assuming that people will know that _Process is a magic class produced by `attr` :)

src/twisted/runner/procmon.py
```diff
@@ -103,14 +128,26 @@ class ProcessMonitor(service.Service):
...
-        self.processes = {}
+        self._processes = {}
         self.protocols = {}
         self.delay = {}
         self.timeStarted = {}
         self.murder = {}
         self.restart = {}
 
 
+    @deprecate.deprecatedProperty(incremental.Version("Twisted", 17, 10, 0))
```
adiroiban: I don't know where it is documented, but I remember seeing something like a placeholder which is used instead of the version.
At release, there is a script which will replace it with the actual version in which it is released

src/twisted/runner/procmon.py
```diff
@@ -103,14 +128,26 @@ class ProcessMonitor(service.Service):
...
+        """
+        Processes as list of tuples
+
+        @return: List of monitored processes as tuples
+        """
+        return {name: process.toTuple()
+                for name, process in self._processes.items()}
+
+
+    @deprecate.deprecated(incremental.Version("Twisted", 17, 10, 0))
```
adiroiban: why are we deprecating pickle support? just asking.

I don't see any reference in the release notes about this deprecation

src/twisted/runner/procmon.py
```diff
@@ -103,14 +128,26 @@ class ProcessMonitor(service.Service):
...
         self.murder = {}
         self.restart = {}
 
 
+    @deprecate.deprecatedProperty(incremental.Version("Twisted", 17, 10, 0))
+    def processes(self):
+        """
+        Processes as list of tuples
+
+        @return: List of monitored processes as tuples
```
adiroiban: is this a list? it looks like dict to me

src/twisted/runner/procmon.py
```diff
@@ -5,11 +5,36 @@
...
+import incremental
+
+from twisted.python import log, deprecate
 from twisted.internet import error, protocol, reactor as _reactor
 from twisted.application import service
 from twisted.protocols import basic
 
+@attr.s(frozen=True)
+class _Process(object):
+
```
adiroiban: is this empy line needed?

src/twisted/runner/procmon.py
```diff
@@ -5,11 +5,36 @@
...
+    """
+    The details of a process to be restarted.
```
adiroiban: I was expecting to see docstring for each member of this class .

-------

minor comment.

details seem generic... but I don't have a better name.

Maybe "parameters" ... but this can be confused with arguments .... maybe "attributes"?

src/twisted/runner/procmon.py
```diff
@@ -119,10 +156,12 @@ def __getstate__(self):
         dct['timeStarted'] = {}
         dct['murder'] = {}
         dct['restart'] = {}
+        del dct['_processes']
+        dct['processes'] = self.processes
         return dct
 
 
-    def addProcess(self, name, args, uid=None, gid=None, env={}):
+    def addProcess(self, name, args, uid=None, gid=None, env={}, cwd=None):
```
adiroiban, line 191: no docstring for the new cwd argument

src/twisted/runner/test/test_procmon.py
```diff
@@ -547,3 +598,56 @@ def test_stopServiceCleanupScheduledRestarts(self):
         # all pending process restarts.
         self.assertEqual(self.pm.protocols, {})
 
+
+
+class DeprecationTests(unittest.SynchronousTestCase):
+
+    """
+    Tests that check functionality that should be deprecated is deprecated.
```
adiroiban, line 609: why do we need this separate test case? why not write them together with the other test for ProcessMonitor?

adiroiban (dismissed):
Thanks for working on this.

I am happy with using the magic _Process class for this is much better than using `proc[0] proc[1] proc[2]`

I feel that the code was documented in a hurry and I am not quite sure that I get what is happening here.

I think that this branch is on the right track but it needs a bit more love :)

Thanks!

adiroiban:
> No ticket number, no ticket link, no coherent commit messages?

@exarkun  I know the pain and the confusion :)

For my project I have a set of keywords which are left as comments and then a GitHub hook will use them to automatically set a label... you can see a review with automatic label management here https://github.com/chevah/brink/pull/113

I tried to propose using it on Twisted but I got no feedback https://twistedmatrix.com/pipermail/twisted-python/2017-October/031694.html

src/twisted/runner/test/test_procmon.py
moshez: When we finally remove the deprecated functionality, it is easy to remove all tests for it.

src/twisted/newsfragments/9287.feature
```diff
@@ -0,0 +1,3 @@
+Add optional cwd argument to twisted.runner.procmon.ProcMon.addProcess
+
+Also deprecate direct introspection of ProcMon's processes: processes should not be directly accessed or pickled.
```
adiroiban: this needs to be in a separate 9287.removal file, since this is  not about a feature.

see http://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles for info about how these fragments are used

src/twisted/runner/procmon.py
```diff
@@ -5,11 +5,50 @@
 """
 Support for starting, monitoring, and restarting child process.
 """
-from twisted.python import log
+import attr
```
adiroiban, line 8: does Twisted depend on attr ?

I don't see it defined in https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_setup.py#L241

I know that attr my be already there from another dependency, but we need to explicitly define it... just in case all the dependencies are moving away from it and then it will no longer be available

src/twisted/runner/procmon.py
```diff
@@ -5,11 +5,50 @@
...
+    args = attr.ib()
+    uid = attr.ib(default=None)
+    gid = attr.ib(default=None)
+    env = attr.ib(default=attr.Factory(dict))
+    cwd = attr.ib(default=None)
+
+    @deprecate.deprecated(incremental.Version("Twisted", "NEXT", 0, 0))
+    def toTuple(self):
+        """
+        Convert to tuple
```
adiroiban: just a note the other devs:

I am not sure if this docstrings adds any extra info on top of `def toTuple(self):` and if it makes sense to use it

So in the current form it's just useless documentation.

Maybe use this oportunity to add info about what to use instead of this deprecated code

moshez:
Note that the build failure on the fedora is a flaky test. i've rebuilt, and it passed (but did not comment on the PR...)

wsanchez (approved):
Nice, this is great work, thanks.

src/twisted/runner/procmon.py
```diff
@@ -5,11 +5,62 @@
...
+    """
+    The parameters of a process to be restarted.
+
+    @ivar args: command-line arguments (including name of command as first one)
+    @type args: C{list}
+
+    @ivar uid: user-id to run process as, or None
+    @type uid: C{int}
+
+    @ivar gid: group-id to run process as, or None
```
twm: "or None meaning inherit"? Same for uid?

src/twisted/runner/procmon.py
twm: This should be updated to NEXT like the one on line 43?

Can you point me at the docs for this functionality so that I can add it to the wiki page?

src/twisted/runner/procmon.py
```diff
@@ -164,15 +233,15 @@ def removeProcess(self, name):
...
+        del self._processes[name]
 
 
     def startService(self):
         """
         Start all monitored processes.
         """
         service.Service.startService(self)
-        for name in self.processes:
+        for name in self._processes:
```
twm: Should this (and stopService) be reentrant? I think that you may need to do `for name in list(self._processes)`.

src/twisted/runner/procmon.py
markrwilliams: I think https://github.com/twisted/incremental#updating is the citation.

src/twisted/runner/procmon.py
moshez: It's there now :)

src/twisted/runner/procmon.py
moshez: Good catch, adding.

src/twisted/runner/procmon.py
moshez: Yup, fixing.

src/twisted/runner/procmon.py
moshez: Added

PR #897, opened 2017-09-24
[9288] Create a new session when previous session expired.

adiroiban:
Scope
=====

This tries to fix https://twistedmatrix.com/trac/ticket/9288 so that an exception is not raised when requesting the session associated with a request.

Changes
=======

When the previous session has expired, the code will generate a new session on the fly.

-----------

Not sure how to fix the design for an expired session. The fact that Session.touch() raises AlreadyCalled for an expired session does not seem to me like the right API.

Maybe Session touch should catch these errors and raise a dedicated exception

-------------

Not sure if this is the right way to implement it... this is Sunday 11PM fixing.

The possibilities (as described by Exarkun):

1. You get back the same Session object as you had before, but with its lifetime extended as though you've just `touch()`'d it.
2. You get back a new Session object but with the same uid as the one you had before with a life time of `sessionTimeout` starting at the point of the `getSession` call.
3. You get back a new Session object with a different uid - basically, a whole new session, as if the previous one never existed.

The current fix is option 3 Please double check :)

------

As a drive by, I have fixed the docstring of an existing test... I think it was my fault:)

How to test
=========

I got into this by having a one GET request which creates the session and very long running streamed POST request which takes longer than 15 minutes.... After all the POSTed data is received the session is retrieved... but I ended up with an exception.

I have created a unit test. Let me know if you think that a functional/integration/end to end test is needed


src/twisted/web/newsfragments/9288.bugfix
```diff
@@ -0,0 +1,2 @@
+twisted.web.server.Request.getSession() now returns a new session if the
```
exarkun: Doesn't need the `()`

src/twisted/web/server.py
```diff
@@ -453,6 +454,14 @@ def getSession(self, sessionInterface=None, forceNotSecure=False):
 
         session = getattr(self, sessionAttribute)
 
+        if session:
```
exarkun: This is probably better written as `if session is not None:` (and the existing line below better written as `if session is None:`

src/twisted/web/test/test_web.py
```diff
@@ -751,6 +751,41 @@ def test_retrieveNonExistentSession(self):
...
+        request = server.Request(d, 1)
+        request.site = site
+        request.sitepath = []
+
+        def SessionFactoryWithClock(site, uid):
+            """
+            Forward to normal session factory, but inject the clock.
+            """
+            session = OriginalSessionFactory(site, uid)
+            session._reactor = clock
```
exarkun: If this code doesn't know what `OriginalSessionFactory` is then it doesn't really know that that it's valid to set `_reactor` on it,

src/twisted/web/test/test_web.py
```diff
@@ -751,6 +751,41 @@ def test_retrieveNonExistentSession(self):
...
+
+        # The site is patch to allow injecting a clock to the session.
+        OriginalSessionFactory = site.sessionFactory
+        site.sessionFactory = SessionFactoryWithClock
+
+        initialSession = request.getSession()
+
+        # When the session is requested after the session timeout,
+        # no error is raised and a new session is returned.
+        clock.advance(OriginalSessionFactory.sessionTimeout )
```
exarkun: extra space near the end

src/twisted/web/test/test_web.py
```diff
@@ -751,6 +751,41 @@ def test_retrieveNonExistentSession(self):
...
+        newSession = request.getSession()
+
+        self.assertIsNot(initialSession, newSession)
+        self.addCleanup(newSession.expire)
```
exarkun: It's moderately pointless to do `addCleanup` at the _end_ of a test method.  The point is that you do it _before_ any exceptions that could occur and prevent you from doing the cleanup.  The test runner then calls the cleanups you registered whether the test succeeds or fails.  It's basically like a `finally` statement for the whole test method.  You should call `addCleanup` as soon as the session is created. 

But, also, since the IReactorTime is a `Clock` there's probably no actual reason to expire the session.  No resources are being leaked by not expiring it.

src/twisted/web/test/test_web.py
```diff
@@ -751,6 +751,41 @@ def test_retrieveNonExistentSession(self):
...
+        site.sessionFactory = SessionFactoryWithClock
+
+        initialSession = request.getSession()
+
+        # When the session is requested after the session timeout,
+        # no error is raised and a new session is returned.
+        clock.advance(OriginalSessionFactory.sessionTimeout )
+        newSession = request.getSession()
+
+        self.assertIsNot(initialSession, newSession)
```
exarkun: This is the most important question for the whole PR ... what should be true about the result of `getSession` after a prior session has expired?

The possibilities:

1. You get back the same Session object as you had before, but with its lifetime extended as though you've just `touch()`'d it.
2. You get back a new Session object but with the same uid as the one you had before with a life time of `sessionTimeout` starting at the point of the `getSession` call.
3. You get back a new Session object with a different uid - basically a whole new session, as if the previous one never existed.

Option 3 makes the most sense to me.  Once the old session expires it seems like all remaining behavior should be as though it never existed in the first place.  But ...

If you are handling a request using two libraries, call them A and B, and near the beginning of handling the request library A asks for the session (creating it, perhaps) and near the end of handling the request library B asks for the session, then what is the desired behavior if enough time separates those two events sufficiently for the session to expire?  Do you _want_ to be able to get two different sessions while handling a single request?

It seems like it might be even more reasonable, on a particular request, the session could never exist and then not exist.  Either the request is associated with a session or it is not.  It shouldn't be associated with two sessions, should it?

So perhaps I will propose that instead this assertion should be exactly flipped?  `assertIs(initialSession, newSession)`?

What do you think?  Does this example scenario sound compelling?  Does it differ from any real-world cases you've encountered?

exarkun (changes requested):
Thanks.  Some comments left inline and a kind of big question about the desirable behavior.

adiroiban:
Thanks for the wonderful review. And this was fast :) Much appreciated.

> This is probably better written as `if session is not None`:

True 


> If this code doesn't know what `OriginalSessionFactory`

Can you please suggest what name to use or what approach to use here for the test?

Maybe we should add Site._reactor and the site should then always pass its reactor to the session.

I did not want to extend the scope of this PR, but if you are happy to review such a change, I can create a new PR or extend the scope of this PR. 

> It's moderately pointless to do addCleanup
> But, also, since the IReactorTime is a `Clock`.

True. The cleanup should come first, and in this case, it can be removed.

I first wrote this test for my project. There I have a SpecializedTestCase which has a clock instances used by all tests and this instance is checked at tearDown to make sure it is clean.
For me, this is good practice and it help find some bugs by just writing simple unit tests.
Maybe tests using task.Clock should use a dedicated TestCaseWithClock, which can do the same checks at teardown.

-----------

Looking at the code, I think Session.expireCallbacks should be private 

https://github.com/twisted/twisted/blob/3af1e3250c26ba9e3deb2062e3e56566693bbaf3/src/twisted/web/server.py#L678

Also, checking the Session.expire() implementation, not sure what this comment is all about: `# Break reference cycle.`

It looks like that cycle is broken on explicit expire, but is not broken when auto-expired. Not sure if this is a bug or a feature :)

https://github.com/twisted/twisted/blob/3af1e3250c26ba9e3deb2062e3e56566693bbaf3/src/twisted/web/server.py#L692


------------------------

> This is the most important question for the whole PR ... what should be true about the result of getSession after a prior session has expired?

I have implemented only a few things with twisted web, basically just uploading and downloading of files, so I don't know if I am the best person to decide.

I have never worked with a library which ask for the raw Twisted Web session.

I don't know if option 1 and 2 are ok.

When a session expires, Session.expireCallbacks are called. Those callback might do some ireversible actions to the session.

Pretending to get the same session, as if nothing has happend, might create more confusion.

we have +2 for Option 3.

I will raise discussion over the mailing list and see if we get any feedback.

If not, we can go with Option 3, but update the docstring/documentation to talk that the use case that you have presented with library A and library B is not supported.

Thanks again for the review :)

exarkun:
> Can you please suggest what name to use or what approach to use here for the test?

> Maybe we should add Site._reactor and the site should then always pass its reactor to the session.

> I did not want to extend the scope of this PR, but if you are happy to review such a change, I can create a new PR or extend the scope of this PR.

Passing a reactor in to the session factory sounds like a nice improvement to me.  I agree it would be better as a separate PR.

> Maybe tests using task.Clock should use a dedicated TestCaseWithClock, which can do the same checks at teardown.

Something like this might be nice.  Probably an approach using fixtures would be better than a new `TestCase` base class but I guess trial still doesn't support fixtures.  I agree that there's a category of code/tests for which it's nice to see that you haven't leaked resources and it's nice when the test harness checks this for you.  It would be good to separate "use the real reactor" from "get teardown-time checks about leaked resources" and be able to use but not the other.  For this particular PR, I'm not too concerned though.

> It looks like that cycle is broken on explicit expire, but is not broken when auto-expired. Not sure if this is a bug or a feature :)

Yea, oops.  Also, the session's handling of `self.expireCallbacks` is not re-entrant safe (imagine a callback that calls `notifyOnExpire` or `touch` or something).

> I will raise discussion over the mailing list and see if we get any feedback.

Thanks.  It will be good to get some more feedback.

As far as a next review goes, I doubt I'll have any time before next weekend at least.  If you remind me about it Friday or Saturday though I'll probably be able to do another pass.

adiroiban:
> Yea, oops. Also, the session's handling of self.expireCallbacks is not re-entrant safe (imagine a callback that calls notifyOnExpire or touch or something).

Yep. I forgot about this. First time when I saw the error I thought that I am using some on expiring callbacks which are touching the session, but I don't have such code and while digging deeper I forgot about this issue.

I have created a ticket for that at https://twistedmatrix.com/trac/ticket/9290 , but as commented in the ticket, I think that sane code will not try to touch or add another expire callback while being callled :) 

------------

I have fixed the style issues (and will fix twistedchecker-trunk-diff.sh to report such issues).

Not sure how to fix the design for an expired session. The fact that Session.touch() raises AlreadyCalled for an expired session does not seem to me like the right API.

I have updated the PR description to reflect this issue.

I am putting this for review, to see if I can get more feedback :)

needs-review

ldanielburr:
As an interested onlooker, I am wondering how this affects applications that have both authenticated and anonymous users.  If an authenticated user makes a very long-lived request, such that their session expires before the request is completed, then option 3 would seem to make the most sense, as you'd want them to be logged out, with an anonymous session, rather than still connected to the authenticated session.

adiroiban:
Thanks for the feedback so +3 for Option 3.

---------

There is also the option for long-lived request to be implemented in such a way that it will touch the session from time to time.

-----------

My main issue with option 1 or 2 is that the session expiration callbacks are called, yet getSession will return a new session, as nothing happened.

The site or the request might not know how to re-create a new session.

Ex, you have a session which is created during login time.
You create the session and then attach some data to it using login logic/code.

If you try to auto/magically create a new session at the end of the request, you can get it a session, but it will not have the data

adiroiban:
@exarkun just in case you have some time to review it. Thanks!

There was no feedback over the mailing list https://twistedmatrix.com/pipermail/twisted-python/2017-September/031664.html

src/twisted/web/test/test_web.py
```diff
@@ -789,6 +789,54 @@ def test_retrieveNonExistentSession(self):
...
+            """
+            Forward to normal session factory, but inject the clock.
+
+            @param site: The site on which the session is created.
+            @type site: L{server.Site}
+
+            @param uid: A unique identifier for the session.
+            @type uid: C{bytes}
+
+            @param site: The site on which the session is created.
```
glyph: This `@param` is duplicated, please remove it.

glyph (approved):
Hopefully CI will be happier with you this trip around the roulette wheel :).

I think that Session is a deeply problematic interface that should probably eventually be deprecated (it doesn't really work in a multi-process context, among other things) and that's why I've been overhauling it from the ground up in Klein's forms-and-sessions branch.  Nevertheless, keeping it maintained and working as well as it can within the constraints of its design is a good thing, and will make migrating to a future interface a lot easier than if it's left to bitrot, so your work on this is much appreciated :).

It appears that you've dealt with the feedback JP offered in his earlier reviews, so please feel free to land when you can.

SamyCookie:
Hi,
thanks for fixing this !
I have reported a very similar issue with session caching and session.expire() : 
cf. https://twistedmatrix.com/trac/ticket/9385

by slightly editing you're new test case (just for a demonstration of my bug),
cf. https://github.com/twisted/twisted/commit/e0c355419c330449bd73f562cee1cdcd0bea094a I got a crash.

Could you tell me if that behaviour is intended, and if you have a clue to fix this (as it seems you know this part of the code).

Thanks !

adiroiban:
@SamyCookie I don't understand your issue. Sorry.
I also don't understand the changed test. 
Please try to state the issue by highlighting the expected result vs the actual result.

PR #900, opened 2017-09-29
[9292] Add template for GitHub PR.

adiroiban:
Scope
====

This adds a template for the GitHub PR description. See https://twistedmatrix.com/trac/ticket/9292

Changes
======

Added the GitHub template.

I went for `.github` to reduce the numbers of files in root and to make it clear that those files are only for GitHub. (Ex can be removed if we move to something else)

This is based on the template suggested by Craig.

I went for "automated tests" not just "unit tests" to make it clear that integration tests (or other forms of tests) are needed :)

There is the full path to newsfragments.

Maybe in a separate PR we can create src/twisted/newsfragments/README which explains how this folder is managed, and in this way we have the docs in the souce code.

How to test
=========

Unless this is merged in master, the template will not appear for PR requests... so for now, just check that the information is ok.

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -15,8 +15,6 @@ https://twistedmatrix.com/trac/wiki/ContributingToTwistedLabs
 
 Twisted has a Code of Conduct, available at code_of_conduct.md.
 
-**Warning: pull requests are ignored** unless they have an associated ticket in trac.
-
 File a ticket at:
 
 https://twistedmatrix.com/trac/newticket
@@ -24,6 +22,7 @@ https://twistedmatrix.com/trac/newticket
 Twisted uses Trac to keep track of bugs, feature requests, and associated
 patches because GitHub doesn't provide adequate tooling for its community.
 
+Contributions are managed using GitHub's Pull Requests.
 For a PR to be accepted it needs to have:
 
 * all Travis CI tests passing
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,0 +1,11 @@
+## Remove this paragraph
+
+Please have a look at our developer documentation before submitting your Pull Request.
+
+https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch
+
+## Contributor Checklist:
+
+* [ ] There is an associated ticket in Trac. Create a new one at https://twistedmatrix.com/trac/newticket
+* [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )
+* [ ] I have updated the automated tests.
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -23,6 +23,7 @@ exclude appveyor.yml
 prune bin
 prune admin
 prune .travis
+prune .github
 
 # Include test-running utilities for downstream packagers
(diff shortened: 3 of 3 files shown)


rodrigc (approved):
This looks fine to me.  It is short and succinct enough that it has all the relevant points backed by the URLs that people should look at.

Keep in mind, that no matter what we do, there will
be people who file pull requests, and will skim over even this checklist and not read it. :)

However, it's better to have this than not have this.

Before merging this, get @moshez's feedback to make sure he is on board.

CONTRIBUTING
```diff
@@ -15,15 +15,14 @@ https://twistedmatrix.com/trac/wiki/ContributingToTwistedLabs
 
 Twisted has a Code of Conduct, available at code_of_conduct.md.
 
-**Warning: pull requests are ignored** unless they have an associated ticket in trac.
```
moshez, line 18: Thanks @adiroiban for moving so quickly on this!

FWIW -- currently the only way to kick the auto-test machinery is to open a pull request, so it would be nice if there was a way to say "this is not ready for review yet". Maybe we can suggest an explicit "WIP" or "In-progress" tag in the template for such things?

adiroiban: I know the pain.

Not sure what is a "tag" :)

Labels will not work as only contributors can change them.

---------

For now, things should be "simple". Set the "review" keyword on the Trac ticket. This will show in the review report https://twistedmatrix.com/trac/report/25

--------

What I am doing for other repos, is using a "needs-review" keyword in a comment.

There is a webhook listening for comments, and when needs-review is there it auto updates Trac and can do other things... like setting a label.

moshez: Usually I've seen this done by a prefix on the PR (e.g., "WIP:" means "work in progress"). If we all agree that nobody reviews until it's marked "review" in trac that's fine -- we should make it explicit, for contributors and reviewers alike. Nobody likes their half-done work judged :)

adiroiban: The Trac keyword thing is described at 

http://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch 

and this link is part of the new template.

----------

I saw the WIP in the title, but I don't like it.

You request once a review... but then changes are required, should the reviewer set WIP back to the title and once is ready for re-review the author remove the WIP title?

For me, the 'needs-review' work best, as once all is ready for re-review, you will leave a comment addresing what whas changed and at the end you leave the keyword.

No need to scroll to the top, click edit button and change the title :)

adiroiban:
I will merge this.

I think that other discussions in terms of changing the current process should be made on the mailing list.

Thank for the review.

rodrigc:
This is an incremental improvement and is OK to merge as you have done.

The additional issues that @moshez brought up are valid, but definitely need to be discussed on
this mailing list.

Shortly after the migration to GitHub, @glyph brought up this proposal:

https://twistedmatrix.com/pipermail/twisted-python/2016-May/030333.html

It would probably worth resurrecting that discussion, and get some agreement on
the mailing list about improving the patch submission workflow for Twisted.

adiroiban:
Yes. This should be done over the ML.

And I agree that they are valid.

I don't like closing a PR to mark that it needs more work.
I think that we should use labels.

The label management can be automated by a txghbot (https://github.com/markrwilliams/txghbot)

Even with the process proposed by Glyph, we would still want to use the bot... and since we have the bot, why not put it in full use and map the workflow to labels.

PR #901, opened 2017-09-29
Update physical constant to 2014 CODATA value

exarkun:
Fixes https://twistedmatrix.com/trac/ticket/9293

### DIFF
--- a/src/twisted/internet/protocol.py
+++ b/src/twisted/internet/protocol.py
@@ -374,7 +374,8 @@ class ReconnectingClientFactory(ClientFactory):
     # Phi = 1.6180339887498948 # (Phi is acceptable for use as a
     # factor if e is too large for your application.)
 
-    # molar Planck constant times c, joule meter/mole
+    # This is the value of the molar Planck constant times c, joule
+    # meter/mole.  The value is attributable to
     # https://physics.nist.gov/cgi-bin/cuu/Value?nahc|search_for=molar+planck+constant+times+c
     jitter = 0.119626565582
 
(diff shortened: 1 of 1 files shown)


Julian:
Fine copy pasting skills.

This lgtm, but I feel like either here or the ticket should have some sort of note on what functional thing this does.

You accepted the ticket, which means it looks like you understand what it does :) -- if so can you note that either here or on the ticket (or even better on the commit?) so that someone can figure out why someone would care about the small change in the constant?

(ISTM that really all this does is slightly change the randomness distribution that will be used for the reconnecting, which really shouldn't change any end user behavior in a noticeable way and that this is just a "pedantry" change to have an accurate constant value -- if that's the case, probably just saying "yes that's all" on this ticket is good enough for me :)

adiroiban (approved):
Looks good

For 

```
src/twisted/internet/protocol.py (75.0%):
    377: W9402: (Used for checking comment format issues. ), : The first letter of comment should be capitalized
```

I think that we should remove W9402 from twistedcheker as its heuristic can get many false positives.

exarkun:
> Fine copy pasting skills.

Skills honed through long practice.  I did perform the change three different ways using three different tools and verified I achieved the same result each time, just to be sure.

> this is just a "pedantry" change to have an accurate constant value -- if that's the case, probably just saying "yes that's all" on this ticket is good enough for me :)

Yes.  Precisely.

> Looks good

Thanks!

> I think that we should remove W9402 from twistedcheker as its heuristic can get many false positives.

I agree.  I'll bodge it here for now.

PR #902, opened 2017-09-30
[9295] HTTP client closes the connection in a clean way on long headers from server.

adiroiban:
Scope
=====

https://twistedmatrix.com/trac/ticket/9295

This fixes an internal error when server responds with a long header.

Changes
=======

loseConnection was implemented on TransportProxyProducer as this is called by the parser.

In an ideal world, the HTTP11Protocol should have a different signal when headers are too long.
... but twisted/protocols/basic.py(648)lineLengthExceeded() only calls loseConnection without giving any reason... and loseConnection does not accept a reason.

If something needs to be improved, I think that this should be done in a separate ticket.

---------

The is a `pragma nocover` there as the assertion should not be called under normal circumstances.

I don't know how else to write that test.

----------

I don't understand why TransportProxyProducer._stopProxying and HTTPClientParser._responseDeferred are private, while they are called from external places.

As a drive by, I have made stopProxying public, as it is used as public anyway... so keeping it private is just confusing.

How to test
=========

Demo server code:

```
from twisted.web import server, resource
from twisted.internet import reactor, endpoints

class Simple(resource.Resource):
    isLeaf = True
        request.setHeader('some-long-header', 'value' * 10000)
        return "<html>Hello, world!</html>"

site = server.Site(Simple())
print("Listening on http://locahost:8080")
endpoint = endpoints.TCP4ServerEndpoint(reactor, 8080)
endpoint.listen(site)
reactor.run()
```

Demo client code

```

agent = Agent(reactor)

d = agent.request(
    b'GET',
    b'http://localhost:8080',
    None,
    )

def cbResponse(ignored):
    print('Response received')
d.addCallback(cbResponse)


def ebResponse(failure):
    print('Failure received %r' % failure)
d.addErrback(ebResponse)

def cbShutdown(ignored):
    reactor.stop()
d.addBoth(cbShutdown)

reactor.run()
```

-------------------

On trunk, you should get:

```
$ python web_client.py 
Failure received <twisted.python.failure.Failure twisted.web._newclient.ResponseFailed: [<twisted.python.failure.Failure exceptions.AttributeError: 'TransportProxyProducer' object has no attribute 'loseConnection'>]>
```

On this branch, it should just be closed

```
Failure received <twisted.python.failure.Failure twisted.web._newclient.ResponseFailed: [<twisted.python.failure.Failure twisted.internet.error.ConnectionDone: Connection was closed cleanly.>]>
```


src/twisted/web/_newclient.py
```diff
@@ -1347,6 +1347,14 @@ def pauseProducing(self):
...
 
 
+    def loseConnection(self):
+        """
+        Proxy the request to lose the connection to the underlying producer,
+        unless this proxy has been stopped.
+        """
+        if self._producer is not None:
+            self._producer.loseConnection()
+
```
jlitzingerdev: style nit: twistedchecker expects 3 lines here so this will add a new error.

src/twisted/web/_newclient.py
```diff
@@ -1568,7 +1576,7 @@ def _disconnectParser(self, reason):
             # transport.  Stop proxying from the parser's transport to the real
             # transport before telling the parser it's done so that it can't do
             # anything.
-            self._transportProxy._stopProxying()
+            self._transportProxy.stopProxying()
```
jlitzingerdev, line 1594: Beyond just making the method public, is there much value in the method at all?  The producer is set to None immediately after it is called, is it common for something else to keep a reference?  Aside from that question making it public does seem sane, though somewhat unrelated to the problem at hand.

jlitzingerdev (changes requested):
Review feedback.

Thanks for your work in tracking down this error, the classes involved here have interesting relationships with each other.

My two cents -- While the proposed change does silence the error, I don't think it's the complete solution to the problem.  The request still fails (somewhat mysteriously now) and the user's situation isn't much improved.

That said, another core problem is that HTTPClientParser expects to treat its transport like an ITransport, which means loseConnection() should be implemented by TransportProxyProducer based on the context in which it is used (as the transport of a protocol).  I've no issue with how that is done here.

Rather than reject without trying to provide any thoughts on what I think the complete solution might look like, here goes:

After investigating, it seems to me that the real problem is that silently applying LineReceiver.MAX_LENGTH to an HTTP request isn't necessarily appropriate. From what I understand HTTP itself doesn't specify a limit, and when I test using Firefox and Safari on iOS both clients happily complete the request.  At the very least it should be configurable at the client level so that the user knows *why* the request failed and can make an informed decision.

One path forward would be:
1.  Provide an option at the client level to explicitly set a max header length and set MAX_LENGTH accordingly.
2.  Override lineLengthExceeded in HTTPClientParser.
3.  Call the errback for the request with a failure indicating that the reason for the failure was a header that was too large.

Ultimately the goal I have in mind is allowing the issuer of the request to clearly understand why it failed and choose whether a large header is a valid failure case for their use.

src/twisted/web/_newclient.py
adiroiban: If you don't have stopProxying, you will have to access the private `self._transportProxy._producer.stopProxying()`

I think that there is a lot of value in this method as it encapsulated a well-defined action on the transport.

For now we are only (conditionally) stop producing, but in the future there might be other internal actions to be done for stopProxying

With a public method, the other users don't have to care about _producer and whether it is None or not.

adiroiban:
Thanks for the review.

I think that adding configurable behaviour is out of the scope of this PR.
The HTTP parser is using line receiver so anybody can use the standard LineReceiver "configuration".

What we have here is HTTP parser which is already calling loseConnection on the transfer.
It might call it when headers are too long, but it might call it in another context.

For the scope of this, the problem is that the parser is using TransportProxyProducer as transport and that TransportProxyProducer is not implementing loseConnection.

For this PR, I don't want to go into details of when the HTTP parser call loseConnection on its transport.

> Ultimately the goal I have in mind is allowing the issuer of the request to clearly understand why it failed and choose whether a large header is a valid failure case for their use.

I am not saying that your questions are not valid or that your goal is not valid, but I think that they should be fixed in another ticket.

jlitzingerdev (approved):
Assuming that this gets a green build and no NAKs pointing out misses on my part, these changes look good to me.  My original concerns were actually discussed in ticket 8570 (http://twistedmatrix.com/trac/ticket/8570).  In addition, the fixes in this PR don't change behavior  (the connection still closes), they just don't throw an unrelated error.

I would ask that you consider (read - this is not a deal breaker to me) fixing the twistedchecker issue I mentioned, but that is mainly speaking as a non-committer who routinely digs through twistedchecker outputs before submission.  Not adding to that would be nice.

adiroiban:
Thanks! Much appreciated.

Well, there are conflicts and issues, so I have to solve that and put this back for another review.

All tests, including twistedchecker should be green before merge.

src/twisted/web/_newclient.py
adiroiban: Thanks. I don't know why twistedchecker has not raised this issue

adiroiban:
I hope that tests will be green now and this should be ready for review :)

jlitzingerdev:
@adiroiban Is that Fedora failure legitimate or something with buildbot?  I'm happy to review again if it can be safely ignored (tangential question: does the ticket need to be in review?).

adiroiban:
The fedora issue looks serious... I retry and it still fails.

adiroiban:
@exarkun has found the reason of this issue https://twistedmatrix.com/trac/ticket/9355#comment:4 so this is not related to the changes from this branch

putting it back for review

jlitzingerdev:
Looks like https://twistedmatrix.com/trac/ticket/9382 is also related.  Regardless, assuming nobody picks it up prior I'll be able to get to this review in the next day or three.

jlitzingerdev:
These changes look good.  The style nit I raised in my last review has been addressed and I haven't seen any objections to making stopProxying public.  For good measure I also ran treq and klein's tests and aside from something odd (Success?! was the literal result) in klein, didn't see any issues raised with this change.

jlitzingerdev (approved):
These changes look good.  The style nit I raised in my last review has been addressed and I haven't seen any objections to making stopProxying public.  For good measure I also ran treq and klein's tests and aside from something odd (Success?! was the literal result) in klein, didn't see any issues raised with this change.

adiroiban:
Thanks a lot for the review. Much appreciated :)

PR #904, opened 2017-10-02
[9299] Patch SMTP Authentication for Python3

gmorell:
https://twistedmatrix.com/trac/ticket/9299


rodrigc (commented):
This fix is not quite correct.  auth.challengeResponse() must return bytes.  If you put this change in, then there are lots of places that need to change in similar ways.

A better fix is to encode self.username and self.password as it is passed into sendmail().  That way LOGINAuthenticator will still return bytes.

rodrigc:
Moved to https://github.com/twisted/twisted/pull/932

PR #906, opened 2017-10-03
Don't install C sources, etc

tdsmith:
Files included from MANIFEST.in but excluded by exclude_package_data are
included in sdists but excluded from bdists and actual installation.

https://twistedmatrix.com/trac/ticket/7495


tdsmith:
This happens to lack unit test coverage. A useful functional test would be that an sdist includes a .c file and a wheel does not, but I'm not sure if that's in the useful scope of twisted's testing.

moshez:
🚗👋
Have you seen the tests in https://github.com/twisted/twisted/blob/trunk/src/twisted/python/test/test_setup.py#L33 ? Are they useful as inspiration?

tdsmith:
Thanks for taking a look! It seems hard to test this without actually calling setuptools.setup() and seeing what happens, and nothing in test_setup is quite willing to cross that line. :) I could pull it out into a module variable and assert that it contains what it contains but I don't think that really improves its testedness. Or I could try to dive into actually producing s/bdist objects and see if anything catches on fire!

adiroiban:
@tdsmith I understand that this is hard to do automated test.

Since there are not automated tests, I think that it would help to update the PR description with some information about how do to manual testing for this and what should be the expected results.

We can merge this without tests and if there are problems people could refer to this PR and check that level of testing was done and what was missing from the test plan.

Thanks!

morgendave (changes requested):
@tdsmith Can you add how to manual test this in the description then we can merge this. Thanks a lot!

hawkowl:
```~/c/twisted (7495-excludepackagedata|✔) $ 
ls build/tests/lib/python2.7/site-packages/twisted/internet/iocpreactor/
abstract.py   const.pyc     interfaces.py   reactor.py   setup.pyc  udp.py
abstract.pyc  __init__.py   interfaces.pyc  reactor.pyc  tcp.py     udp.pyc
const.py      __init__.pyc  notes.txt       setup.py     tcp.pyc
~/c/twisted (7495-excludepackagedata|✔) $ ls appveyor.yml `^C
~/c/twisted (7495-excludepackagedata|✔) $ 
ls ~/venvs/commands/lib/python2.7/site-packages/twisted/internet/iocpreactor/
abstract.py   const.py     __init__.pyc    iocpsupport/  reactor.pyc  tcp.py   udp.pyc
abstract.pyc  const.pyc    interfaces.py   notes.txt     setup.py     tcp.pyc
build.bat     __init__.py  interfaces.pyc  reactor.py    setup.pyc    udp.py
```

(the first does not have the iocpsupport sources -- which is this branch -- my locally installed venv with mainline does, so this passes)

PR #907, opened 2017-10-04
[9300] arnimarj wsgi race

arnimarj:
A fix and a unit-test for https://twistedmatrix.com/trac/ticket/9300.

The test is unfortunately not deterministic, and relies on delaying pending writes from the wsgi application, whilst closing the underlying transport.


src/twisted/web/wsgi.py
```diff
@@ -506,11 +507,13 @@ def run(self):
         except:
             def wsgiError(started, type, value, traceback):
                 err(Failure(value, type, traceback), "WSGI application error")
-                if started:
-                    self.request.loseConnection()
-                else:
-                    self.request.setResponseCode(INTERNAL_SERVER_ERROR)
-                    self.request.finish()
+                if not self._requestFinished:
```
adiroiban: minor comment,
to reduce indentation and document the intend maybe this can be written as
```
def wsgiError(started, type, value, traceback):
    err(Failure(value, type, traceback), "WSGI application error")
    if self._requestFinished:
        # Some comment about what happend here
        return
    if started:
      EXISTING_CODE_HERE
```

src/twisted/web/wsgi.py
```diff
@@ -446,9 +446,10 @@ def write(self, data):
         # Thing at some point in the future.
 
         def wsgiWrite(started):
-            if not started:
-                self._sendResponseHeaders()
-            self.request.write(data)
+            if not self._requestFinished:
```
adiroiban: see the other comment regarding indentation.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1848,6 +1851,115 @@ def cbRendered(ignored):
         return d
 
 
+    @inlineCallbacks
+    def test_pendingRequestOperationsRespectConnectionLoss(self):
```
adiroiban, line 1964: this is a huge test with a lot going on there, but no docstring :(

Please consider writing as many docstring as possible so people reading this could assert whether the testing strategy is valid.

Do we really need threads and the real system clock to write this test?

Why not use just a twisted.internet.task.Clock

--------


I guess that you can write this by triggering a write in a thread and before the reactor is spinned, close the connection on the server side... this should trigger the client side connection closing and then when the reactor is spun for the write, it will find that the request is done.

adiroiban (changes requested):
Thanks for looking into this.

That huge tests needs more love :)

What do you mean  by "The test is unfortunately not deterministic"

Does it mean that if  we run this test with current trunk, it will not always fail?

I think that this should be split into a unit test in which we just focus on a write after the connection is closed or something like this.

With the current tests as it is now, if in 6 months it will start to fail, I am not sure that the person looking at this test will know how to approch it and how to fix it.

arnimarj:
The previous testing strategy was hopelessly broken and  complicated. I've redesigned in such a way that pending writes are held back until after the connection is closed. I'll do a better write up once I've cleared my head.

I appreciate the review, thanks.

arnimarj:
Just to re-iterate on the bug in question. The issue is that the transport is used even after it has been closed. The ordering of events which triggers the bug is:

1. the `wsgi` application returns a string
2. the returned string is scheduled to be written (using `reactor.callFromThread`)
3. the transport is closed
4. the write from 2. is performed on the now-closed transport.

The faulty step is 4, where the current status of the transport is not checked before the write is performed. This PR fixes that by checking the `_WSGIResponse._requestFinished` flag before every write.

The test forces this ordering of events by blocking the write in 2. until the transport has been closed. It does so by patching the `_WSGIResponse.write()` function and blocks on a `threading.Event()` which is set once the transport is closed.

I hope this is clearer than my previous attempt @adiroiban.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+                self.reactor.callFromThread(writeResult.errback, Failure())
+
+        self.patch(_WSGIResponse, 'write', _WSGIResponse_write)
+
+        # Run and wait for the request to fail.
+        request = self.lowLevelRender(
+            requestFactory, applicationFactory, DummyChannel,
+            'GET', '1.1', [], [''])
+
+        yield self.failUnlessFailure(request.notifyFinish(), ConnectionLost)
```
adiroiban: I don't know why we have the failUnlessFailure alias, (it is only used in a single other place)
I think that is better to use assertFailure http://twistedmatrix.com/documents/current/api/twisted.trial.unittest.TestCase.html#assertFailure

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+        _write = _WSGIResponse.write
+
+        # Signal for the delayed write allowing it to proceed.
+        connectionLost = threading.Event()
+
+        # Result of the (delayed) write.
+        writeResult = Deferred()
+
+        def _WSGIResponse_write(self, data):
+            # Block write until transport is closed.
```
adiroiban: minor comment about style. Can we have it as a docstring?

And it can be written in a generic form describing how this write method is rigged and how it is expected to behave.

I had to read this code a couple of time to understand how writeResult is used and what is the design of the test.

Maybe include this into a second paragraph in the top level docstring.

I saw the comment on the PR and I know the conversation from there, but I think that the code should be readable without having to consult all the PR comments.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+        If there are pending transport writes after
+        C{Request.notifyFinish} is called, the transport should
+        not be written to.
+        """
+
+        self.enableThreads()
+        trackedRequest = [None]
+
+        def applicationFactory():
+            def application(environ, startResponse):
```
adiroiban, line 1997: missing docstring.
May say that `applicationFactory` this is needed by the WSGIResource API

and then `def application(environ, startResponse):` can have a docstring describing what this application does.

Like:
Minimal application with always succeeds with 200 and returns the response in a single chunk.

----------

I am not familiar with WSGI so to understand this code I had to look at

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+                startResponse('200 OK', [])
+                self.reactor.callFromThread(
+                    trackedRequest[0].connectionLost,
+                    Failure(ConnectionLost("No more connection")))
+                yield b'write'
```
adiroiban: maybe the response can be self documented and have it like `b'single chunk response'`

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+
+            return application
+
+        def requestFactory(*args, **kwargs):
+            # We need to manually close the request instantiated
```
adiroiban: maybe this can have a docstring describing what type of requests does this factory produce.

And I don't understand the comment.

I don't see how we are closing the request here, but rather  just keep generate a request for which we keep a reference.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+            # by self.lowLevelRender().
+            self.assertIsNone(trackedRequest[0])
```
adiroiban: maybe add a message here, like "Something went wrong as this tests is expected to produce a single request."

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+            trackedRequest[0] = Request(*args, **kwargs)
+            return trackedRequest[0]
+
+        # Keep a copy of _WSGIResponse.write before we patch it.
+        _write = _WSGIResponse.write
```
adiroiban: maybe this can be named as _nonRiggedWSGIResponseWrite , or something like that to hint that this is the non patched method (or _nonPatchedWSGIResponseWrite or _originalWSGIResponseWrite)

In this way. I hope that the comment is no longer needed

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+
+        # Signal for the delayed write allowing it to proceed.
+        connectionLost = threading.Event()
```
adiroiban: can we have another name for this. Using the same name as the API method is a bit confusing.

Maybe just call it connectionLostSyncPoint or something like that.

The fact that we use an event to wait for something can be added to the top level docstring where the testing strategy is described.

src/twisted/newsfragments/9300.bugfix
```diff
@@ -0,0 +1 @@
+There is no longer a race condition in _WSGIResponse where pending writes would fail on already closed transports.
```
adiroiban: I think that the release note should be targeted to end users (for twisted lib these are developers) and only talk about public stuff.

I expect that no WSGI app developer will use _WSGIResponse or should care about it.

From what I can see, this is a fix to the WSGIResource as this is what end users will have in their code.

And the release note should use the full name: twisted.web.wsgi.WSGIResource

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+        not be written to.
+        """
+
+        self.enableThreads()
+        trackedRequest = [None]
+
+        def applicationFactory():
+            def application(environ, startResponse):
+                startResponse('200 OK', [])
+                self.reactor.callFromThread(
```
adiroiban, line 2000: I don't understand why we need to call from thread.

I had applied this patch to your branch and the test still pass

```
diff --git a/src/twisted/web/test/test_wsgi.py b/src/twisted/web/test/test_wsgi.py
index fdae8de..ca85fa4 100644
--- a/src/twisted/web/test/test_wsgi.py
+++ b/src/twisted/web/test/test_wsgi.py
@@ -1957,8 +1957,7 @@ class ApplicationTests(WSGITestsMixin, TestCase):
         def applicationFactory():
             def application(environ, startResponse):
                 startResponse('200 OK', [])
-                self.reactor.callFromThread(
-                    trackedRequest[0].connectionLost,
+                trackedRequest[0].connectionLost(
                     Failure(ConnectionLost("No more connection")))
                 yield b'write'
```

so the test is either broken or this callFromThread is not needed

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+        def _WSGIResponse_write(self, data):
+            # Block write until transport is closed.
+            connectionLost.wait()
+
+            try:
+                r = _write(self, data)
+                self.reactor.callFromThread(writeResult.callback, None)
+                return r
+            except:
+                self.reactor.callFromThread(writeResult.errback, Failure())
```
adiroiban: I feel that this code will be not executed in the tests and it has a very generic exception handling.

I think that it needs a comment and at least `except Exception` to not catch Ctrl+C or other higher level exceptions

Without a comment, someone later might see that it has no coverage and might decide to delete it.

I saw that without this, the test will block.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,68 @@ def cbRendered(ignored):
...
+        # Run and wait for the request to fail.
+        request = self.lowLevelRender(
+            requestFactory, applicationFactory, DummyChannel,
+            'GET', '1.1', [], [''])
+
+        yield self.failUnlessFailure(request.notifyFinish(), ConnectionLost)
+
+        # Unblock write and wait for it to finish.
+        connectionLost.set()
+        yield writeResult
```
adiroiban: I think that we should also check what was written to the network, not only that write was successful.

it might be successful,  but have some extra info written to the network

src/twisted/web/wsgi.py
```diff
@@ -455,6 +455,10 @@ def write(self, data):
         # Thing at some point in the future.
 
         def wsgiWrite(started):
+            if self._requestFinished:
+                # Self.request.notifyFinish() has triggered, and
+                # transport is potentially closed.
```
adiroiban, line 459: Maybe: `Do nothing as this is late write requested.`

but is it ok to just ignore this case?

I was expecting that this might be due to some bad code on the end users, so we should at least log an error message

adiroiban (changes requested):
The code looks much better. Thanks a lot!

In summary, I was expecting to see a log for the "lost" write call as I don't expect that a well-written application will call write after request is done.

I see that there are 3 changes, one is a rewrite.
I am expecting at least 2 new tests... or 3 if the existing code has no tests.

The coverage report is broken now.
I hope it will be fixed soon and you can see the missing coverage.

I am happy with the testing strategy.

Thanks again for all your effort.

src/twisted/newsfragments/9300.bugfix
arnimarj: Updated the class reference, and added a note about the symptoms before the fix.

src/twisted/web/test/test_wsgi.py
arnimarj: Added a more specific docstring.

src/twisted/web/test/test_wsgi.py
arnimarj: The way Twisted handles running wsgi applications is by running the request handler in a thread pool, so this code will never run on the I/O thread.

src/twisted/web/test/test_wsgi.py
arnimarj: Yes.

src/twisted/web/test/test_wsgi.py
arnimarj: A updated the doc. For the test to schedule a disconnect for the request issues by `lowLevelRender` it needs a reference to it (which we store in `trackedRequest[0]`).

src/twisted/web/test/test_wsgi.py
arnimarj: Added a comment.

src/twisted/web/test/test_wsgi.py
arnimarj: Went with `_originalWSGIResponseWrite`.

src/twisted/web/test/test_wsgi.py
arnimarj: Renamed `connectionLost` to `connectionLostSyncPoint`.

Added a description of the test to the docstring, but found the mentioning of threading.Event to not be needed.

src/twisted/web/test/test_wsgi.py
arnimarj: Agreed. Added a note about the reasoning behind this.

src/twisted/web/test/test_wsgi.py
arnimarj: Wasn't aware it was an alias. Fixed.

src/twisted/web/test/test_wsgi.py
arnimarj: The point is what it wasn't written, because the write was only attempted after the connection was lost. The original bug was that the attempted write did not check whether the transport was still open (and subsequently failed).

src/twisted/web/wsgi.py
arnimarj: It is not OK, since we don't know (and have no direct control over) when this code will run. Since the wsgi app is run in a thread pool, all data returned must be handed over to the main I/O thread (using callFromThread), and whilst that data is waiting to be written, the transport may be closed (the race in question).

src/twisted/web/test/test_wsgi.py
adiroiban: but maybe test that transport is empty... or update the implementation to yield some data before the sync point

in this way we can check that only the initial data was written and no other unwanted data.

arnimarj:
Did a second iteration on the test. I'll address the missing coverage a bit later.

arnimarj:
@adiroiban, I assumed the codecov issue was a blunder on my part, but like you said, it seems to be broken for now.

The PR is in slightly better shape, hope you can take another pass at it.

src/twisted/newsfragments/9300.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a race condition in twisted.web.wsgi.WSGIResource which would emit "Unhandled error in Deferred" log messages for wsgi apps.
```
adiroiban: Not sure if this is better. The idea is that the release note should help someone decide if they were previously affected by this error or not.

A race condition in twisted.web.wsgi.WSGIResource caused by writes being done after the request was finished no longer occur. The condition was triggering  "Unhandled error in Deferred" type of errors.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,91 @@ def cbRendered(ignored):
...
+                self.reactor.callFromThread(
+                    trackedRequest[0].connectionLost,
+                    Failure(ConnectionLost("No more connection")))
+
+                yield b'single chunk response'
+
+            return application
+
+        def requestFactory(*args, **kwargs):
+            # A factory which returns a twisted.web.server.Request
```
adiroiban: minor comment: why not use docstrings here?

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,91 @@ def cbRendered(ignored):
...
+            signalled, and puts the result/failure
+            in C{writeResult}.
+            @return: None
+            """
+            connectionLostSyncPoint.wait()
+
+            try:
+                r = _originalWSGIResponseWrite(self, data)
+                self.reactor.callFromThread(
+                    wsgiResponseWriteResult.callback, None)
```
adiroiban: minor comment. maybe instead of "None" we should pass "data" to the wsgiResponseWriteResult and then we can check what we got here.
I expect that is little place for failure, but if its not too hard I prefer to assert as much as possible when dealing with patched/rigged objects.

src/twisted/web/test/test_wsgi.py
```diff
@@ -1941,6 +1943,91 @@ def cbRendered(ignored):
...
+                return r
+            except Exception:
+                # The above code is not expected to fail, but if it does the
```
adiroiban: maybe

```
# The above code is not expected to fail.
# We need this handling here as if the code does fail,
# without this part the test will hang.
  ```

**Review by adiroiban — CHANGES_REQUESTED** [review]
Looks good. For now, only new minor comments.

I see that you agree that there should be a log in wsgiWrite but I don't see any change in that.

Please consider adding a logging before returning from write, without writing anything.

So to have this merged I would like to see:

* A log in wsgiWrite when it does not write anything
* Tests for the other places where no action is done

Thanks!

File: src/twisted/web/test/test_wsgi.py (hunk 20 above)
Comment (arnimarj) [reply src/twisted/web/test/test_wsgi.py]: Agreed.

File: src/twisted/web/test/test_wsgi.py (hunk 19 above)
Comment (arnimarj) [reply src/twisted/web/test/test_wsgi.py]: Replaced `None` with `data` and checked for equality at the end of the test.

File: src/twisted/web/test/test_wsgi.py (hunk 18 above)
Comment (arnimarj) [reply src/twisted/web/test/test_wsgi.py]: Done.

File: src/twisted/newsfragments/9300.bugfix (hunk 17 above)
Comment (arnimarj) [reply src/twisted/newsfragments/9300.bugfix]: Done.

File: src/twisted/web/test/test_wsgi.py (hunk 15 above)
Comment (arnimarj) [reply src/twisted/web/test/test_wsgi.py]: I check now that the write was attempted, and that nothing was actually written to the transport.

**Comment by arnimarj** [comment]
Hi @adiroiban . Addressed all the points, except the one regarding logging when writes are skipped.

Let me explain. There are 3 places (for the current trunk) in `wsgi.py` which attempt to write to the transport: 

1. `wsgiWrite()`
2. `wsgiError()`
3. `wsgiFinish()`

Of the 3, only the last one (correctly) checks for the `_requestFinished` flag before writing to the transport. It currently does NOT log that it skipped the write when `_requestFinished` is `False`.

The PR fixes 1. and 2. to mimic the behavior of 3. I assumed that being consistent with current behavior would be the better option.

**Comment by arnimarj** [comment]
Is there a way for me to restart certain travis jobs? There are a couple failing because of external factors (e.g. failing github.com connections).

**Comment by adiroiban** [comment]
> The PR fixes 1. and 2. to mimic the behavior of 3. I assumed that being consistent with current behavior would be the better option.

`wsgiFinish` tries to write only the header and only if the response was not done yet... so I think that is ok to not log something here. I think that this is fine.

`wsgiError` does send a generic log (failure) message, before handling the state of the request. I think that this is also fine.

I can try to get someone else to weight into this, but I want to have a log for `wsgiWrite` and my impression was that you are also expecting to see a log there.

I don't think that keeping consistency with `wsgiFinish` is a valid point. wsgiFinish is for something which is done ... `wsgiWrite` is for something which we still expect to be active.

So wsgiFinish has a different purpose, and I don't find it valid to try to keep the same behaviour in wsgiWrite :)

**Comment by adiroiban** [comment]
> Is there a way for me to restart certain travis jobs?

I don't know... as you are not owner/collaborator I guess that Travis does not allow you to mess with the builds. I have restarted them.

**Comment by arnimarj** [comment]
Hi. Added a log message. Not having the same behavior is a fair point, but I would like to add one final argument...

There is already a mechanism for the application to know that the request did not finish successfully (`notifyFinish()`), so why log on pending work where there is no expectation of it finishing.

The log message would be emitted for every chunk of data, and a particularly bad (admittedly contrived) case would be an app emitting multiple single byte chunks, each of which may end up as a log message.

In any case, I'd be happy see this merged as is. Thanks for you patience.

**Comment by rodrigc** [comment]
@arnimarj The trick to restarting Travis jobs if you don't have admin access is to close the PR on github, and then re-open the PR.

File: src/twisted/web/wsgi.py [hunk 21]
```diff
@@ -455,6 +455,16 @@ def write(self, data):
...
+                message = 'skipping write scheduled after transport closed'
+                self._log.error(message,
+                    system='wsgi',
+                    isError=True,
```
adiroiban: maybe @wsanchez can help here.

are there any conventions for how to add extra data to an log message.

I was expecting that log.error() will take care of isError ... and self._log will take care of `system` so that we don't need to pass these arguments for each log message.

based on that reasoning system and isError are not adding any new information to the message

arnimarj: I wasn't clear on the convention either. I copied a similar construct from the `_ErrorStream` class which may be used the the `wsgi` app for error logging.

adiroiban:
> There is already a mechanism for the application to know that the request did not finish successfully (notifyFinish()), so why log on pending work where there is no expectation of it finishing.

my reasoning for this log is to help identify erroneous code.
My understanding is that under normal circumstances this branch of the code should not be reached.

I don't understand how notifyFinish() can help here.
My understanding is that by the time this write is called, notifyFinish() was already signaled.

----

This need more tests to be merged :)

arnimarj:
You misunderstand that only buggy applications will hit this bug.

I believe the example I gave in the original bug demonstrates it (https://twistedmatrix.com/trac/attachment/ticket/9300/twisted_bug.py).

 EDIT:

I have a slightly updates version:

```python
#!/usr/bin/python

import time, requests, threading

from twisted.internet import reactor, threads, defer, endpoints
from twisted.web.wsgi import WSGIResource
from twisted.web.server import Site

def sleep_a_bit():
	d = defer.Deferred()
	reactor.callLater(10.0, d.callback, None)
	return d

def failing_get(url):
	try:
		requests.get(url, timeout = (1.0, 0.00001))
	except requests.exceptions.ReadTimeout:
		pass

class Counter(object):
	def __init__(self):
		self.count = 0
		self.lock = threading.Lock()

	def inc(self):
		with self.lock:
			self.count += 1

	def get(self):
		with self.lock:
			return self.count

def sleepyApplicationFactory(counter):
	def application(environ, startResponse):
		try:
			startResponse('200 OK', [])
			time.sleep(1.0)
			yield b'some string'
		finally:
			counter.inc()

	return application


@defer.inlineCallbacks
def startup():
	try:
		n_requests = 1
		counter = Counter()

		resource = WSGIResource(reactor, reactor.getThreadPool(), sleepyApplicationFactory(counter))
		site = Site(resource)
		port = yield endpoints.TCP4ServerEndpoint(reactor, 8888).listen(site)

		# submit all requests
		for i in range(n_requests):
			threads.deferToThread(failing_get, 'http://127.0.0.1:8888/')

		# wait for the request handlers to finish
		while counter.get() < n_requests:
			yield threads.deferToThread(time.sleep, 0.1)

		yield port.stopListening()
	except Exception as e:
		print(e)
	finally:
		reactor.stop()

reactor.callWhenRunning(startup)
reactor.run()
```

The application is simple enough. With Twisted 17.9.0 you will get this:

```
WSGI application error
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/python/context.py", line 87, in callWithContext
    self.contexts.pop()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/web/wsgi.py", line 522, in run
    self.started = True
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/web/wsgi.py", line 500, in run
    self.write(elem)
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/web/wsgi.py", line 455, in write
    self.reactor, wsgiWrite, self.started)
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/internet/threads.py", line 122, in blockingCallFromThread
    result.raiseException()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/python/failure.py", line 385, in raiseException
builtins.AttributeError: 'NoneType' object has no attribute 'writeHeaders'

Unhandled Error
  File "/home/arni/dohop/repo/admin/yandex/twisted_bug.py", line 72, in <module>
    reactor.run()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/internet/base.py", line 1243, in run
    self.mainLoop()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/internet/base.py", line 1252, in mainLoop
    self.runUntilCurrent()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/internet/base.py", line 851, in runUntilCurrent
    f(*a, **kw)
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/there should be an error log in `wsgiFsite-packages/twisted/web/wsgi.py", line 510, in wsgiError
    self.request.loseConnection()
  File "/home/arni/dohop/virtualenvs/py363/lib/python3.6/site-packages/twisted/web/http.py", line 1474, in loseConnection
    self.channel.loseConnection()
builtins.AttributeError: 'NoneType' object has no attribute 'loseConnection'
```

Actually now, I believe that the error log in `wsgiWrite` should be in `wsgiFinish`.

adiroiban:
> I have a slightly updates version:

I guess that sleep_a_bit is not needed... or is the example incomplete?

If you think that log is not needed, I am fine with that.

Basically, if the server takes longer to respond and client closes the connection before the server will respond it is normal/expect for the write method to be called after the connection is closed.

------

So you can remove the log and add the missing tests and this should be ready for merge :)

Thanks!

arnimarj:
Hi. The reason for the `sleep_a_bit` is to make sure all the `wsgiWrite`/`wsgiFinish`es scheduled on the background thread are allowed to complete (and fail) before the reactor is shut down. The updated example has an explicit counter that it polls to wait for the pending work.

Your understanding is correct, the there may be multiple scheduled (but not complete) `wsgiWrite`/`wsgiFinish` after a connection is lost.

Thanks again for the reviews, I'll be very happy to have contributed a tiny bit.

arnimarj:
Just wanted to ping you @adiroiban to check if you find this ready to be merged.

src/twisted/web/wsgi.py
```diff
@@ -518,18 +522,26 @@ def wsgiError(started, type, value, traceback):
                     "WSGI application error",
                     failure=Failure(value, type, traceback)
                 )
+                if self._requestFinished:
+                    # Self.request.notifyFinish() has triggered, and
+                    # transport is potentially closed
+                    return
```
adiroiban, line 527: Do we have a test dedicated to this branch?

adiroiban:
> So you can remove the log and add the missing tests and this should be ready for merge :)

I see a change in run with error, but I don't see any test for that.

The test is needed for your own good... since I assume that you are using this code, the test should help prevent introducing a regression during future work/refactor.

Note that I have never wrote a wsgi app or done any wsgi related works... and I don't really care about wsgi in twisted (yet).

I was just trying to do a review for the general parts hoping that meanwhile someone who cares more about wsgi in Twisted could do a final review

src/twisted/web/wsgi.py
arnimarj: I added code which does another request where the response handler purposefully raises an exception. It hits this branch.

arnimarj:
Re-opening to force a new travis run.

adiroiban:
That is one big (towards huge test). I think that it needs to be split into 2 tests, each testing a separate part.

Have you tested this on py3.5 and 3.6 ... looks like it stalls on those systems. Does it pass on your system?

And the long lines need to be fixed.

The comments about comment not starting with a capital letter should  be ignored (and update the travis rule to ignore them)

PR #908, opened 2017-10-06
Avoid using recursion when building tracebacks

richvdh:
The list of frames in a Failure may be very long - particularly if the failure
was due to a stack overflow. It is unhelpful if our error-handling functions
themselves fail with stack overflows when handling a stack overflow situation.

To fix one particular failure mode, avoid using recursion to handle the frames
list.

Fixes https://twistedmatrix.com/trac/ticket/9301.

### DIFF
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -100,7 +100,7 @@ def _Traceback(frames):
     @param frames: [(methodname, filename, lineno, locals, globals), ...]
     """
     assert len(frames) > 0, "Must pass some frames"
-    # we deliberately avoid using recursion here, as the frames list may be
+    # We deliberately avoid using recursion here, as the frames list may be
     # long.
     tb = None
     for frame in reversed(frames):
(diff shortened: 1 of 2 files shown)


richvdh:
I don't *think* the CI failures on python 3.6 are my doing, but feel free to tell me how I'm wrong...

markrwilliams:
@richvdh Definitely not your fault - @rodrigc [recently fixed this](https://github.com/twisted/twisted/commit/e58956fe12af7b41f6c1735ba7b14cba69e82a5e).

src/twisted/python/failure.py
```diff
@@ -90,30 +90,41 @@ class NoCurrentExceptionError(Exception):
 
 
 
-class _Traceback(object):
+def _Traceback(frames):
```
markrwilliams, line 93: I'm a little worried about switching this to a function, but it's a private API and it doesn't seem to be called anywhere outside of `twisted.python.failure`.

markrwilliams (approved):
This looks reasonable - thanks for the contribution!

This is ready to merge when the builds succeed.

PR #911, opened 2017-10-07
Fix more bytes/unicode issues in twisted.web

rodrigc:
https://twistedmatrix.com/trac/ticket/9135


rodrigc:
@markrwiliams please look

src/twisted/web/distrib.py
```diff
@@ -362,7 +365,7 @@ def getChild(self, name, request):
             sub = 0
         try:
             pw_name, pw_passwd, pw_uid, pw_gid, pw_gecos, pw_dir, pw_shell \
-                     = self._pwd.getpwnam(username)
+                     = self._pwd.getpwnam(username.decode("utf-8"))
```
markrwilliams: This is UTF-8 because name comes from a URL path?

src/twisted/web/resource.py
```diff
@@ -240,7 +243,7 @@ def render(self, request):
 
         @see: L{IResource.render}
         """
-        m = getattr(self, 'render_' + nativeString(request.method), None)
+        m = getattr(self, u'render_' + request.method.decode("ascii"), None)
```
markrwilliams: `getattr` works in terms of native strings, not unicode strings.  Why did you change this?

src/twisted/web/script.py
```diff
@@ -119,11 +119,16 @@ class ResourceScriptDirectory(resource.Resource):
     """
     def __init__(self, pathname, registry=None):
         resource.Resource.__init__(self)
+        if isinstance(pathname, bytes):
+            pathname = pathname.decode("utf-8")
```
markrwilliams: This should use `sys.getfilesystemencoding`, not UTF-8.

src/twisted/web/script.py
```diff
@@ -161,7 +166,8 @@ def render(self, request):
         will NOT be handled with print - standard output goes to the log - but
         with request.write.
         """
-        request.setHeader(b"x-powered-by", networkString("Twisted/%s" % copyright.version))
+        request.setHeader(b"x-powered-by",
+            u"Twisted/{}".format(copyright.version).encode("utf-8"))
```
markrwilliams: [RFC 7230 says header values should be ASCII](https://tools.ietf.org/html/rfc7230#page-25).  Why has this been changed to UTF-8?

src/twisted/web/script.py
```diff
@@ -175,8 +181,8 @@ def render(self, request):
             io = NativeStringIO()
             traceback.print_exc(file=io)
             output = util._PRE(io.getvalue())
-            if _PY3:
-                output = output.encode("utf8")
+            if not isinstance(output, bytes):
+                output = output.encode("utf-8")
```
markrwilliams, line 188: The docstring for `PythonScript` should say that it requires the script to be UTF-8.

src/twisted/web/static.py
```diff
@@ -42,7 +41,7 @@
 dangerousPathError = resource.NoResource("Invalid request URL.")
 
 def isDangerous(path):
-    return path == b'..' or b'/' in path or networkString(os.sep) in path
+    return path == b'..' or b'/' in path or os.sep.encode("utf-8") in path
```
markrwilliams: This should use `sys.getfilesystemencoding`.

src/twisted/web/static.py
```diff
@@ -57,7 +56,7 @@ def __init__(self, data, type):
 
 
     def render_GET(self, request):
-        request.setHeader(b"content-type", networkString(self.type))
+        request.setHeader(b"content-type", self.type.encode("utf-8"))
```
markrwilliams: Headers should be ASCII.

src/twisted/web/static.py
```diff
@@ -290,15 +289,15 @@ def getChild(self, path, request):
...
-                path = path.decode('utf-8')
-            except UnicodeDecodeError:
-                log.err(None,
-                        "Could not decode path segment as utf-8: %r" % (path,))
-                return self.childNotFound
+        if not isinstance(path, bytes):
+            raise TypeError(
+                "{} is type: {}, not bytes".format(path, type(path)))
+        try:
+            path.decode('utf-8')
```
markrwilliams: You need to use the decoded path: https://github.com/twisted/twisted/pull/894#discussion_r143305534

src/twisted/web/static.py
```diff
@@ -451,8 +450,8 @@ def _contentRange(self, offset, size):
         @return: The value as appropriate for the value of a Content-Range
             header.
         """
-        return networkString('bytes %d-%d/%d' % (
-            offset, offset + size - 1, self.getFileSize()))
+        return u'bytes {}-{}/{}'.format(
```
markrwilliams, line 464: Header value should be ASCII, and this goes into a header.

src/twisted/web/static.py
```diff
@@ -477,7 +476,8 @@ def _doSingleRangeRequest(self, request, startAndEnd):
             # request is unsatisfiable.
             request.setResponseCode(http.REQUESTED_RANGE_NOT_SATISFIABLE)
             request.setHeader(
-                b'content-range', networkString('bytes */%d' % (self.getFileSize(),)))
+                b'content-range',
+                u'bytes */{}'.format(self.getFileSize()).encode("utf-8"))
```
markrwilliams: Header values should be ASCII.

src/twisted/web/static.py
```diff
@@ -515,9 +515,10 @@ def _doMultipleRangeRequest(self, request, byteRanges):
         matchingRangeFound = False
         rangeInfo = []
         contentLength = 0
-        boundary = networkString("%x%x" % (int(time.time()*1000000), os.getpid()))
+        boundary = u"{:x}{:x}".format(
```
markrwilliams, line 529: Header values should be ASCII.

src/twisted/web/static.py
```diff
@@ -515,9 +515,10 @@ def _doMultipleRangeRequest(self, request, byteRanges):
...
+            int(time.time()*1000000), os.getpid()).encode("utf-8")
         if self.type:
-            contentType = self.type
+            contentType = self.type.encode("utf-8")
```
markrwilliams: Header values should be ASCII.

src/twisted/web/static.py
```diff
@@ -527,26 +528,28 @@ def _doMultipleRangeRequest(self, request, byteRanges):
...
             contentLength += len(partSeparator)
             rangeInfo.append((partSeparator, partOffset, partSize))
         if not matchingRangeFound:
             request.setResponseCode(http.REQUESTED_RANGE_NOT_SATISFIABLE)
             request.setHeader(
                 b'content-length', b'0')
             request.setHeader(
-                b'content-range', networkString('bytes */%d' % (self.getFileSize(),)))
+                b'content-range',
+                u'bytes */{}'.format(self.getFileSize()).encode("utf-8"))
```
markrwilliams: Header values should be ASCII.

src/twisted/web/static.py
```diff
@@ -567,9 +570,15 @@ def _setContentHeaders(self, request, size=None):
             size = self.getFileSize()
         request.setHeader(b'content-length', intToBytes(size))
         if self.type:
-            request.setHeader(b'content-type', networkString(self.type))
+            contentType = self.type
+            if isinstance(contentType, unicode):
+                contentType = contentType.encode("utf-8")
```
markrwilliams: Header values should be ASCII.  Also, is the `type` instance variable supposed to allow native strings?  Is that why this can be unicode?

src/twisted/web/static.py
```diff
@@ -1057,13 +1065,11 @@ def render(self, request):
...
-        header = "Directory listing for %s" % (
-            escape(unquote(nativeString(request.uri))),)
+        header = u"Directory listing for {}".format(
+            escape(unquote(request.uri.decode("utf-8"))))
 
         done = self.template % {"header": header, "tableContent": tableContent}
-        if _PY3:
-            done = done.encode("utf8")
-
+        done = done.encode("utf-8")
```
markrwilliams, line 1089: This can cause an encoding error with UTF-8 on Python 2, because `done` will first be decoded as ASCII.  The `_PY3` check should remain.

src/twisted/web/test/test_static.py
```diff
@@ -59,6 +59,18 @@ def test_invalidMethod(self):
...
 
+    def test_isDangerous(self):
+        """
+        L{static.isDangerous} returns L{True} if path
+        contains b'..', b'/' or L{os.sep}.
+        """
+        self.assertTrue(static.isDangerous(b".."))
+        self.assertTrue(static.isDangerous(b"foo/bar"))
+        self.assertTrue(static.isDangerous(
+            b"foo" + os.sep.encode("utf-8") + b"bar"))
```
markrwilliams: This should use `sys.getfilesystemencoding`

src/twisted/web/test/test_static.py
```diff
@@ -330,7 +342,7 @@ def cbRendered(ignored):
             self.assertEqual(b''.join(request.written), content)
             self.assertEqual(
                 request.responseHeaders.getRawHeaders(b'content-length')[0],
-                networkString(str(len(content))))
+                str(len(content)).encode("utf-8"))
```
markrwilliams: Header values should be ASCII.

src/twisted/web/test/test_static.py
```diff
@@ -865,7 +908,7 @@ def test_stopProducingClosesFile(self):
         L{StaticProducer.stopProducing} closes the file object the producer is
         producing data from.
         """
-        fileObject = StringIO()
+        fileObject = BytesIO()
```
markrwilliams, line 921: This is good.  It's more explicit than pretending `BytesIO` is `StringIO`.

src/twisted/web/test/test_static.py
```diff
@@ -1803,10 +1848,12 @@ def test_usesGlobalInitFunction(self):
         # something, somewhere, calls mimetypes.init. Yay global
         # mutable state :)
         if getattr(inspect, "signature", None):
+            # Python 3
```
markrwilliams, line 1861: This is good.

src/twisted/web/test/test_xmlrpc.py
```diff
@@ -661,17 +663,18 @@ def setUp(self):
 
 
     def test_authInfoInURL(self):
-        url = "http://%s:%s@127.0.0.1:%d/" % (
-            nativeString(self.user), nativeString(self.password), self.port)
-        p = xmlrpc.Proxy(networkString(url))
+        url = u"http://{}:{}@127.0.0.1:{}/".format(
+            self.user.decode("utf-8"), self.password.decode("utf-8"),
+            self.port)
+        p = xmlrpc.Proxy(url.encode("utf-8"))
```
markrwilliams: Why is this URL encoded as UTF-8, when the others are encoded as ASCII?

The best thing to do would be to use `hyperlink.URL.from_text`, then `to_url()`, then encode _that_ as ASCII.  That will encode each component of the URL according to its type.  @mahmoud Can you confirm this?

src/twisted/web/xmlrpc.py
```diff
@@ -13,8 +13,8 @@
 
 from __future__ import division, absolute_import
 
-from twisted.python.compat import _PY3, intToBytes, nativeString, urllib_parse
-from twisted.python.compat import unicode
+from twisted.python.compat import (_PY3, intToBytes,
```
markrwilliams: Please move this to a separate PR.

markrwilliams (changes requested):
Thank you for working on this.

There are changes around header values unrelated to either https://twistedmatrix.com/trac/ticket/8985 or https://twistedmatrix.com/trac/ticket/9135, and unrelated improvements to `t.w.xmlrpc`.

Please move the `xmlrpc` work to another PR, and refrain from including unrelated changes in subsequent PRs.  It makes them much harder to review.

src/twisted/web/resource.py
rodrigc: `getattr` works with unicode strings on Python 2 and Python 3.  I confirmed by trying it on both.

markrwilliams: On Python 2, `getattr` will silently encode unicode strings with the ASCII codec.  So this will decode `request.method` as ASCII, concatenate it with u`render_`, then re-encode it as ASCII.

On Python 3, prefixing a string literal with `u` has no effect, and `request.method.decode('ascii')` is equivalent to `nativeString`.

So, yes it works, but `getattr`'s actual API works in terms of native strings. The original code aligned with that API by concatenating a native string with `nativeString`.

src/twisted/web/script.py
rodrigc: copyright.version always has characters which fall under the ascii range when encoding with utf-8.

src/twisted/web/script.py
rodrigc: PEP 3120 says the default encoding is UTF-8, so that is a reasonable default: https://docs.python.org/3/reference/lexical_analysis.html#lexical-analysis

src/twisted/web/test/test_static.py
rodrigc: This value will always be in the ascii range even when encodign with "utf-8".

src/twisted/web/static.py
rodrigc: I found that when I used the decoded path here, it didn't work due to cascading dependencies in the call chain of this function, and I needed to keep it as bytes.

src/twisted/web/static.py
rodrigc: No, this is always needed on Python 2 and 3.  Now, done is always unicode, due to header being unicode,
but bytes must be returned from this function.

rodrigc:
@markrwilliams I've updated this patch based on your review comments.  Please look.

src/twisted/web/resource.py
```diff
@@ -240,7 +243,10 @@ def render(self, request):
 
         @see: L{IResource.render}
         """
-        m = getattr(self, 'render_' + nativeString(request.method), None)
+        method = u'render_' + request.method.decode("ascii")
+        if not isinstance(method, str):
```
jlitzingerdev, line 247: Am I mistaken, if method happens to be unicode, this will raise an AttributeError where before nativeString() would have just done the right thing?

src/twisted/web/test/requesthelper.py
```diff
@@ -353,7 +352,7 @@ def setHost(self, host, port, ssl=0):
         if port == default:
```
jlitzingerdev, line 352: This comment applies to all the `intToBytes` conversions -- is this safe?  Specifically, intToBytes does provide some validation of the input where these changes will not.  For example, with trunk on both python2 and 3, the following:

```
from twisted.web.http import HTTPChannel, Request

ch = HTTPChannel()
r = Request(ch)
r.setHost(b'http://www.example.com', 'abc')
```
Will raise a TypeError, however with these changes, no error will be raised.

jlitzingerdev (changes requested):
Here's my review by request.  There is a ton of work here, nicely done!

It also seems like the scope of changes is far wider than just fixing putChild(), are there other related tickes or is it general Python3 cleanup?  Specifically, a theme of these changes seems to be to eliminate the use of several (nativeString, networkString, intToBytes) compat functions

My concerns/requested changes, in summary (with details in the review):

Requested Changes:
misc newsfragments don't really seem appropriate.  Even at the most basic level this modifies putChild's behavior in such a way that a client whose code silently did the wrong thing on python 3 will now raise a TypeError.  It would be nice if this were called out in the changelog.

Concerns
1.  I'm worried about the intToBytes removals, specifically the new code is a lot more forgiving than intToBytes.
2.  The scope of changes also seems to include a general move away from the twisted.python.compat functions (nativeString, networkString, intToBytes), was this motivated by a general discussion/ticket, or was it just to improve conformance to the coding standard by being explicit?
3.  There are a lot of good but seemingly unrelated changes in this patchset, were those necessary to fix the 9135/8985 or were other issues addressed that should be called out in a ticket or summary somewhere?

If you can look at this I will happily re-review, however, I don't think my green light would be enough to merge.  I strongly believe someone with deeper knowledge of twisted.web should review this to ensure there aren't lurking complications with some of the changes.

src/twisted/web/resource.py
rodrigc: No, you are mistaken.  With Python 2.7, getattr() works with a unicode argument.
On Python 3, getattr() will fail with a bytes argument.  So it is safe to use getattr() with a unicode arg on Python 2.7 and Python 3.  See: https://github.com/python/cpython/blob/2.7/Objects/object.c#L1172

rodrigc:
It's hard to pull out changes into separate tickets, because this code is interdependent.

src/twisted/web/test/requesthelper.py
rodrigc: There is not a lot of input validation done on all these arguments to begin with.
I added https://github.com/twisted/twisted/pull/911/commits/38ee5f3ca983c486521c9cffad5cb8afdbc3f54a which will fail on the testcase you provided.

rodrigc:
The functions in `twisted.python.compat` were written a long time ago, during the initial work to port Twisted to Python 3.
They are OK, and got the ball rolling, but are not always the best.
After porting many Twisted modules to Python 3, I am more comfortable with being explicit
in some areas related to bytes/unicode encoding, and not depending on all the functions in twisted.python.compat.
Some of the functions such as networkString() are OK, but not the best.

rodrigc:
Using a local checkout of Twisted from this branch, plus a local checkout of buildbot,
I used this script to checkout treq, klein, and builtbot into a Python 2.7 virtualenv:

```
#!/bin/sh

python2 -m virtualenv testit-2.7
. testit-2.7/bin/activate
python -m pip install --upgrade pip
python -m pip install --upgrade 'treq[dev,test]'
python -m pip install --upgrade 'klein'
cd $HOME/buildbot4
python -m pip install --upgrade -r requirements-ci.txt
python -m pip uninstall -y Twisted
cd $HOME/twisted6
python setup.py develop
```

and this script for a Python 3.6 virtualenv:

```
#!/bin/sh

python3 -m venv testit-3.6
. testit-3.6/bin/activate
python -m pip install --upgrade pip
python -m pip install --upgrade 'treq[dev,test]'
python -m pip install --upgrade 'klein'
cd $HOME/buildbot4
python -m pip install --upgrade -r requirements-ci.txt
python -m pip uninstall -y Twisted
cd $HOME/twisted6
python setup.py develop
```

In each virtualenv, I then ran these commands to test treq, klein, and buildbot against this Twisted branch:

```
trial treq
trial klein
trial buildbot.test
trial buildbot_worker.test
```

The tests passed.

src/twisted/web/test/requesthelper.py
jlitzingerdev: Thanks!  It was primarily the change in validation that drew my attention, not that I was expecting your changes to add new validation.

src/twisted/web/resource.py
jlitzingerdev: So I am, thank you for the clarification!

jlitzingerdev:
@rodrigc You have done a ton of work and testing with this PR, thank you!  My concerns have been addressed, did you have any thoughts on my request for a non misc newsfragment (I realize you haven't resubmitted for review, sorry if this is jumping ahead)?

As I'd like to help see this over the finish line I updated Twisted to this branch (in a sandbox) to test any impact to one of my use cases (HTTPS long polling on an embedded system) and its unit tests.  Everything worked as expected.  I have two others (TLS mutual auth and an HTTP server) I can test in a bit that can hopefully add some confidence.

jlitzingerdev:
Why not introduce new compatibility functions as Glyph requested this in 3 in his review of 8985 (https://twistedmatrix.com/trac/ticket/8985#comment:15)?

It seems relevant for two reasons:

1.  Since this PR deviates from existing approaches to compatibility, it raises the question for non-committers --  which way is "correct"?  The documented, advertised compatibility module, or these deviations buried in source.  Odds are unless one is aware of this thread the former will be selected.  Does that mean proliferation of the "incorrect" way of doing things?   That seems bad.

2.  While there is no question there are improvements to what is in trunk, there are also changes well outside the scope of 8985 and 9135, with the goal of fixing bytes/str issues.  Given that, and the move away from existing compatibility functions, the introduction/deprecation seems directly applicable to what this PR is trying to achieve -- better bytes/str handling for Python 3.

rodrigc:
Moving to https://github.com/twisted/twisted/pull/967

PR #912, opened 2017-10-11
Improve traceback objects returned from Failure.getTracebackObject

richvdh:
This PR attempts to make the traceback returned by `getTracebackObject` look a bit more like a real one, by including `f_lineno` and `f_back` fields in the frame objects, and including the whole stack.

When trying to ascertain why an exception is being thrown, it's often helpful if we can log the whole stack, rather than just the bit between point the exception was thrown and that where it was caught. Applications may have configured their log observers to do just that, so we should make the traceback objects look as much like the real thing as possible.

Fixes https://twistedmatrix.com/trac/ticket/9305.


wiml (approved):
LGTM

PR #913, opened 2017-10-12
[9306] Update openssl.org link to SSL_CTX_load_verify_locations

skylerbunny:
The link provided in this file is far out of date and currently returns a 404. The link above tracks openssl's master branch for documentation of the method.

If instead this wants to track a specific version of OpenSSL, this link can be used instead, since I'm fairly sure that 1.1.0 is the (currently) targeted build:

https://www.openssl.org/docs/man1.1.0/ssl/SSL_CTX_load_verify_locations.html

## Contributor Checklist:


glyph:
Hi @skylerbunny!

Thanks for contributing to Twisted; however, if you want someone to actually *look* at your contribution, you need to add the “review” keyword in trac to get into the list at https://twisted.reviews/

Would you like someone to review this change?

rodrigc:
The existing link in the file, https://www.openssl.org/docs/ssl/SSL_CTX_load_verify_locations.html , doesn't 404 for me.  It redirects to the existing link.  I don't think this change is needed.

skylerbunny:
Ugh. Of course it did. So here's what it looks like happened:

In the last couple weeks, openssl.org itself changed a redirect to a 404 that's been broken since at least 2016, which is what prompted me to create this ticket on the assumption that they'd simply 'drastically reformatted their website's map and didn't care about this broken redirect'.

It used to go to:
https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_load_verify_locations.html
It now goes to:
https://www.openssl.org/docs/manmaster/man3/SSL_CTX_load_verify_locations.html

So I agree - it doesn't 404 anymore. The only reason to change this now (It's still a reason, but not as strong of one) would be if we want to point to a specific SSL version that Twisted targets. (This points to 'master'.) It's probably not important enough to matter.

glyph:
@skylerbunny Nevertheless, your attention to detail is appreciated :). I do think we should probably still make this change, albeit for the different reason you stated.

glyph (approved):
Please link to the 1.1.1 version; similar comment on trac. Thanks for resurrecting this @twm !

PR #843, opened 2017-07-23
Make twisted.python.failure.Failure inherit from BaseException

rodrigc:
http://twistedmatrix.com/trac/ticket/5519

PR #914, opened 2017-10-13
Fix 9307 - raise self (Failure) in python2, wrapped exception in python3

jwtrhs:
Trac bug reference: https://twistedmatrix.com/trac/ticket/9307

Looks like this was last working in 16.6.0, but became broken in 17+.


exarkun (dismissed):
Thanks.  Apparently this behavior has no automated test coverage (if you can believe that).  Can you please add a couple unit tests as well?  Thanks again.

jwtrhs:
No worries, I've added a couple test cases around expected trap() behaviour.

jwtrhs:
Actually turns out there were already some tests in src/twisted/test/test_failure.py but they only covered the Python3 expected functionality. Modified the test to only run on Python3 and added a new test for Python2.

src/twisted/test/test_failure.py
```diff
@@ -65,6 +65,9 @@ def test_trapRaisesWrappedException(self):
         expected types, L{failure.Failure.trap} raises the wrapped
         C{Exception}.
         """
+        if not _PY3:
+            raise SkipTest('Only relevant on Python 3.')
```
markrwilliams: It would be nice if this message were more informative.  Twisted raises the wrapped exception in Python 3 because before [5519](https://github.com/twisted/twisted/pull/843/files), `Failure` did not inherit from `BaseException`, and Python 3 can only `raise` instances of `BaseException`.  Now that `Failure` _is_ a `BaseException`, this behavior persists because it's the API, and as this PR demonstrates, it's bad to break that API.  Even just a `@see` directed at [the documentation](http://twisted.readthedocs.io/en/latest/core/howto/python3.html#twisted-python-failure) would be great.

src/twisted/test/test_failure.py
```diff
@@ -75,6 +78,24 @@ def test_trapRaisesWrappedException(self):
         self.assertIs(exception, untrapped)
 
 
+    def test_trapRaisesSelf(self):
+        """
+        If the wrapped C{Exception} is not a subclass of one of the
+        expected types, L{failure.Failure.trap} raises itself.
+        """
+        if _PY3:
+            raise SkipTest("Only relevant on Python 2.")
```
markrwilliams: See above - Python 2 has raised the `Failure` instance for a long time.

markrwilliams (changes requested):

markrwilliams (approved):

markrwilliams:
@jwaterhouse Looks like twistedchecker doesn't like the format of the comments: https://travis-ci.org/twisted/twisted/jobs/288064570#L537

jwtrhs:
Took a couple tries, but finally figured out the comment checker was picking up the '#' in the @see URL as a comment and emitting a couple errors. But have put in a workaround and flagged the issue here:

https://github.com/twisted/twistedchecker/issues/135

markrwilliams:
@jwaterhouse This is fine for now.  Thanks so much for your contribution!

PR #915, opened 2017-10-16
Port twisted.news to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9309


glyph:
Oh wow this is a bunch of stuff.  Should this supersede #1144 ?

wsanchez:
Need to fix various lint warnings in Travis.

hawkowl:
Since twisted.news is effectively unmaintained enough that it doesn't have a robust test suite, I don't think that we can effectively port it to the quality of the rest of the codebase. Hence, I'm going to close this and the related ticket. If it's useful on Python 3, any porting should be in a separate package from here on.

PR #918, opened 2017-10-21
Line only receiver connection lost 6557

markrwilliams:
https://twistedmatrix.com/trac/ticket/6557

### DIFF
--- a/src/twisted/protocols/basic.py
+++ b/src/twisted/protocols/basic.py
@@ -481,7 +481,7 @@ class LineOnlyReceiver(protocol.Protocol):
         Called when the maximum line length has been reached.
         Override if it needs to be dealt with in some special way.
         """
-        return error.ConnectionLost('Line length exceeded')
+        return self.transport.loseConnection()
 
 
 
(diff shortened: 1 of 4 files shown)


src/twisted/protocols/basic.py
```diff
@@ -481,7 +481,7 @@ def lineLengthExceeded(self, line):
         Called when the maximum line length has been reached.
         Override if it needs to be dealt with in some special way.
         """
-        return error.ConnectionLost('Line length exceeded')
+        return self.transport.loseConnection()
```
moshez, line 484: The "return" here is redundant, and somewhat misleading.

moshez (approved):
Thanks for taking care of this!
No blocking requests, but do consider removing the redundant "return".

src/twisted/protocols/basic.py
markrwilliams: Agreed.  Unfortunately [`LineReceiver` does this as well](https://github.com/requests/requests/issues/4006#issuecomment-300228176).  I'd rather deal with both in a ticket that merges their implementations, or better yet one that finally eliminates the long-deprecated return value of `dataReceived`.

PR #919, opened 2017-10-22
9280 web remove legacy log

jlitzingerdev:
This is the branch for 9280.  The ticket is located here:
https://twistedmatrix.com/trac/ticket/9280

Some items of note:
The following uses of twisted.python.log were not changed:
* html.py - The only usage was inside a deprecated function.
* Usage of twisted.python.logfile.

The strategy I took was to validate that either the existing log messages (for info level) or failures, are still available to observers registered with the global publisher using the new loggers.  The assumption was that the existing code was "correct", and I wanted to prove I didn't break anything.

Unfortunately, this means that existing tests that didn't care about logging now care.  This may not be ideal, as it means changes in messages could break the tests.  However, rather than assume it didn't matter, I took the opposite approach.  If the new code validating logging in the tests should be removed, I'm happy to do so.

There is one simple test I added in test_distrib.  I added it to verify that the failure was logged appropriately because it wasn't previously covered by any tests.  A better test would be to actually exercise the failure of the component a bit better, but the work looked significant enough that I took the easy approach first.  As above, I can change if desired.


jlitzingerdev:
The last two build errors are a bit strange as neither is directly related to any changes I've made in twisted.web.  As I assume review will take a while, I'm still putting the ticket in review and will monitor this as I periodically merge trunk.

jlitzingerdev:
There are two failures in the above code coverage, both in test code.  The first is in a new helper I added in proto_helpers, and the other is a modified line of code in test_wsgi.py that wasn't covered in the first place (I verified coverage on trunk).  I know the policy is 100% coverage, but how does that apply to test code, especially if it wasn't covered in the first place?

src/twisted/test/proto_helpers.py
```diff
@@ -925,3 +927,64 @@ def _check():
...
+@implementer(ILogObserver)
+class EventLoggingObserver(Sequence):
+    """
+    L{ILogObserver} That stores its events in a list for later inspection.
+    This class is similar to L{LimitedHistoryLogObserver} save that the
+    internal buffer is public and intended for external inspection.  The
+    observer implements the sequence protocol to ease iteration of the events.
+
+    @ivar _events: The events captured by this observer
+    @type _events: C{list}
```
rodrigc: This should be `L{list}` as specified in the coding standard: http://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

src/twisted/web/test/test_wsgi.py
```diff
@@ -49,7 +52,9 @@ def callInThread(self, f, *a, **kw):
             # callInThread doesn't let exceptions propagate to the caller.
             # None is always returned and any exception raised gets logged
             # later on.
-            err(None, "Callable passed to SynchronousThreadPool.callInThread failed")
+            self._log.failure(
```
rodrigc, line 55: Can you add a new test to this file which creates a `SynchronousThreadPool` and passes it a function which throws an exception?  That will cover this line which is uncovered.

rodrigc (changes requested):

src/twisted/test/proto_helpers.py
jlitzingerdev: Yep, will do, sorry for the miss.

src/twisted/web/test/test_wsgi.py
jlitzingerdev: I can, and will worst case, but let me see if I can get anything more relevant to the context under test that'll cover this line.

src/twisted/test/proto_helpers.py
```diff
@@ -925,3 +927,64 @@ def _check():
...
+
+    def __len__(self):
+        return len(self._events)
+
+
+    def __getitem__(self, index):
+        return self._events[index]
+
+
+    def __contains__(self, value):
```
rodrigc: This __contains__ function is not getting covered.  Can you add unit test which exercises this function?

jlitzingerdev: @rodrigc I knew that last night but hadn't decided on the best approach before my internet went out.  In thinking it over, since Sequence already has __contains__ in terms of __getitem__, (and since this isn't doing anything special), there is really no value in this code.  If you have no objections I'll just remove it, thoughts?

rodrigc: Removing it would be easier, go ahead

rodrigc (approved):
Looks good, thanks!

glyph:
Landed since @rodrigc approved but I don't believe @jlitzingerdev is a project member :)

jlitzingerdev:
@glyph Thanks for merging, @rodrigc thanks for review.  And assuming 'project member' means one of owner/maintainer/one-with-commit-privs, that is correct, I've only contributed on a handful of tickets/reviews.

rodrigc:
@jlitzingerdev thanks again for your contributions to Twisted.  You put a lot of work into this patch.

jlitzingerdev:
@rodrigc No problem, happy to help, I enjoy working with and on Twisted!  I mentioned on IRC that I plan to go after the remaining uses, assuming no objections when I file each ticket, on a package by package basis.

jlitzingerdev:
@glyph @rodrigc In looking at https://github.com/twisted/treq/issues/185 I found that changes made in this PR suppress the stack trace.  Unless that is an intention of the new log system for cases like this then I may have introduced a regression.

The change that supresses the trace is:
1cc32833ef6d5ca5f397795f9739db69aa53a083

I'm going to dive into why later, but I wanted to give both of you as reviewer/merger a heads up.

Note that the error still occurs, but now, instead of a trace, one sees:
```
[<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')]>]
Unexpected exception from twisted.web.client.FileBodyProducer.stopProducing
```

jlitzingerdev:
As near as I can tell, the visible stack trace is a feature of the legacy log system provided by twisted.python.log.py:textFromEventDict().  The same information can be obtained by processing the failure included with the event in a similar manner, but doesn't happen by default.  So the question -- is a regression?

PR #923, opened 2017-11-06
[9314] Make t.w.s.Request hashable again

twm:
This fixes a regression that appears to have been introduced in https://github.com/twisted/twisted/commit/9a5f758a1a2c49c0b6bc82787eecca068fc527c8. I am not sure if this is the optimal solution. In particular:

* This implementation calls `id(self)`, which I vaguely recall is bad on PyPy. Is this a problem?
* Should `__hash__` only be defined under Python 2? Or even only when old-style classes are in use?
* Should `__hash__` be added to `_IDeprecatedHTTPChannelToRequestInterface` since that requires `__eq__` and `__ne__`?

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9314


src/twisted/web/http.py
```diff
@@ -1516,6 +1516,15 @@ def __ne__(self, other):
         return NotImplemented
 
 
+    def __hash__(self):
+        """
+        A C{Request} is hashable so that it can be used as a mapping key.
+
+        @return A C{int} based on the instance's identity.
+        """
+        return id(self)
```
markrwilliams, line 1549: This is OK under PyPy (see review comment) and fine otherwise.

Please add this to the `_IDeprecatedHTTPChannelToRequestInterface`.

markrwilliams (changes requested):
Thanks so much for digging into this nasty bug!  Since I introduced the regression it's only fair that I review the fix.

> * This implementation calls `id(self)`, which I vaguely recall is bad on PyPy. Is this a problem?

Not with PyPy's current `incminimark` GC.  PyPy is a) not reference counted and b) used moving garbage collectors, so that a collection might change the memory address of an object.  That would be bad for `id` if it weren't for [a trick pulled by `incminimark`](http://doc.pypy.org/en/release-2.4.x/garbage_collection.html#minimark-gc):

> The objects move once only, so we can use a trick to implement id() and hash(). If the object is not in the nursery, it won’t move any more, so its id() and hash() are the object’s address, cast to an integer. If the object is in the nursery, and we ask for its id() or its hash(), then we pre-reserve a location in the old stage, and return the address of that location. If the object survives the next minor collection, we move it there, and so its id() and hash() are preserved. If the object dies then the pre-reserved location becomes free garbage, to be collected at the next major collection.

(The [implementation](https://bitbucket.org/pypy/pypy/src/bab05da3f317dfe5036c8eb020a65fff8df9e46c/rpython/memory/gc/incminimark.py?at=default&fileviewer=file-view-default) calls these pre-reserved locations "shadow" objects.)

> * Should `__hash__` only be defined under Python 2? Or even only when old-style classes are in use?
> * Should __hash__ be added to _IDeprecatedHTTPChannelToRequestInterface since that requires __eq__ and __ne__?

I think it should part of the interface and explicitly defined and documented with a link to your ticket. 

The purpose of this interface was to cover _all_ the ways `Request` is used.  `t.w.http` happens to use `!=` so I put `__eq__` and `__ne__` in, but `Request` is a public API.  If I'd realized the current approach had broken `Request`'s hashability, I would have included `__hash__`.

Now it's true that new style classes "solve" this by always defining `__hash__`, whereas old style classes don't.  This difference affects Python 2.7's [`PyObject_Hash`](https://github.com/python/cpython/blob/2.7/Objects/object.c#L1104-L1126)'s algorithm:

1.Check if the object has a `__hash__` (`tp_hash` gets resolved to Python-level `__hash__` methods via [`fixupslotdefs`](https://github.com/python/cpython/blob/4376a2275967971a2f9586028347d6ed4ed6ef13/Objects/typeobject.c#L6457-L6468)); if so, call it and return the result;
2. Check if the object has no comparison methods defined; if doesn't, return its address;
3. Fail with `NotImplemented`.

So instances of this are hashable in Python 2:

```python
class Hashable:
    def __hash__(self): return 1
    def __cmp__(self, other): return NotImplemented
    def __eq__(self, other): return NotImplemented
    def __ne__(self, other): return NotImplemented
    def __gt__(self, other): return NotImplemented
    def __lt__(self, other): return NotImplemented
```
But instances of this aren't:

```python
class NotHashable:
    def __eq__(self, other): return self is other
```

[On Python 3, however, an object is only hashable if it has a `__hash__`](https://github.com/python/cpython/blob/646e4c8fcc55fc8c4befb9d8a071d31f34fbf562/Objects/object.c#L760-L779).  Note that in _both_ cases `__eq__`,`__cmp__`, etc. are _not_ required for hashability, but are required if you want ever instances from a dictionary.

This is all too subtle _not_ to be covered by the interface so long as Twisted supports Python 2.

twm:
@markrwilliams Thank you for the review! I have added `__hash__` to the interface, so I think this is ready for a second pass.

glyph:
@twm Did you intend to submit this for review? I don't see the keyword in trac.

twm:
@glyph Yes, I did. I had somehow gotten the notion that we were accepting the GitHub labels as well as the review keyword now, but looking back over the mailing list I'm not sure where I got that idea. Maybe I misremembered https://twistedmatrix.com/pipermail/twisted-python/2017-October/031680.html? Anyway, the review keyword has been applied.

src/twisted/web/test/test_http.py
```diff
@@ -2986,6 +2986,17 @@ def test_ne(self):
...
 
+    def test_hashable(self):
+        """
+        A L{http.Request} is hashable.
+        """
+        req = http.Request(DummyChannel(), False)
+        try:
+            hash(req)
+        except TypeError as e:
+            self.fail("http.Request must be hashable: {}".format(e))
```
markrwilliams: I think it's OK to drop the enclosing `try` `except`.  Other tests that call functions that raise an exception on failure, such as those that check interface compliance with `zope.interface.verify.verifyObject`, don't use `fail` to produce a nicer error message, and eliminating this would boost coverage.

src/twisted/web/test/test_web.py
```diff
@@ -499,6 +499,18 @@ def test_interface(self):
...
+    def test_hashable(self):
+        """
+        L{server.Request} instances are hashable, thus can be put in a mapping.
+        """
+        request = server.Request(DummyChannel(), True)
+
+        try:
+            hash(request)
+        except TypeError as e:
+            self.fail("server.Request is not hashable: {}".format(e))
```
markrwilliams: As above.

markrwilliams (approved):
This looks ready to merge!  Please consider the minor comments about improving coverage and then merge.

twm:
Thank you for the review, @markrwilliams! I've removed the try...except blocks.

markrwilliams:
@twm Thank you for finding and fixing this!

PR #924, opened 2017-11-07
Fix bytes issues on Python 3 for "trial -j" 

rodrigc:
https://twistedmatrix.com/trac/ticket/9264

I did some debugging, and without this fix, there are a bunch of errors related to bytes vs. unicode that are happening with *trial -j* such as this one:

```
<twisted.python.failure.Failure builtins.TypeError: Unicode value for key   
b'testCase' not allowed:
'twisted.test.test_abstract.AddressTests.test_emptyString'>
```

These errors are not making it into the logs, which makes things quite difficult to debug.

I went through a few round of fixing the bytes vs. unicode issues with respect to trial and AMP, and got things to work.


rodrigc:
Approved by @glyph here: https://twistedmatrix.com/trac/ticket/9264#comment:5

PR #925, opened 2017-11-08
Work around buggy Cygwin behavior w.r.t. broken pipes.

embray:
Ensure that the parent end of the stdin pipe gets closed on Cygwin and
doesn't cause the reactor to hang.  See
https://twistedmatrix.com/trac/ticket/9323

### DIFF
--- a/src/twisted/internet/process.py
+++ b/src/twisted/internet/process.py
@@ -103,7 +103,7 @@ def unregisterReapProcessHandler(pid, process):
 
 def detectLinuxBrokenPipeBehavior():
     """
-    On some Linux version, write-only pipe are detected as readable. This
+    On some Linux versions, write-only pipe are detected as readable. This
     function is here to check if this bug is present or not.
 
     See L{ProcessWriter.doRead} for a more detailed explanation.
@@ -111,6 +111,10 @@ def detectLinuxBrokenPipeBehavior():
     @return: C{True} if Linux pipe behaviour is broken.
     @rtype : L{bool}
     """
+
+    if not sys.platform.startswith('linux'):
+        return False
+
     r, w = os.pipe()
     os.write(w, b'a')
     reads, writes, exes = select.select([w], [], [], 0)
@@ -126,6 +130,34 @@ def detectLinuxBrokenPipeBehavior():
 
 
 brokenLinuxPipeBehavior = detectLinuxBrokenPipeBehavior()
+
+
+
+def detectCygwinBrokenPipeBehavior():
+    """
+    On Cygwin, polling on the write end of a pipe whose read end has been
+    closed does not set the POLLERR bit.
+
+    In this case we should not use the "enableReadHack" flag of
+    L{ProcessWriter.doRead} or else a select()/poll() on that file descriptor
+    will block.
+
+    @return: C{True} if Cygwin pipe behaviour is broken.
+    @rtype : L{bool}
+    """
+
+    if sys.platform != 'cygwin':
+        return False
+
+    r, w = os.pipe()
+    os.close(r)
+    reads, writes, excs = select.select([w], [], [], 0)
+    os.close(w)
+    return not reads
+
+
+
+brokenCygwinPipeBehavior = detectCygwinBrokenPipeBehavior()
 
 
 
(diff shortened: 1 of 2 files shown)


src/twisted/internet/process.py
```diff
@@ -129,6 +133,34 @@ def detectLinuxBrokenPipeBehavior():
...
+
+    r, w = os.pipe()
+    os.close(r)
+    reads, writes, excs = select.select([w], [], [], 0)
+    os.close(w)
+    return not reads
+
+
+
+brokenCygwinPipeBehavior = detectCygwinBrokenPipeBehavior()
```
adiroiban, line 160: Do we really need to have these 2 new methods part of the public API of process?

I know that we have that for brokenLinuxPipeBehavior ... but I don't see why people using Twisted should care about this :)

src/twisted/internet/process.py
```diff
@@ -980,6 +1013,9 @@ def maybeCallProcessEnded(self):
         #  all readers have indicated EOF
         # This insures that we've gathered all output from the process.
         if self.pipes:
+            if (brokenCygwinPipeBehavior and self.lostProcess and
+                    list(self.pipes.keys()) == [0]):
```
adiroiban, line 1017: I think that this needs a comment, or refactor this long condition into a dedicated boolean variable.

I know that the ticket description is comprehensive but when people will read this code, it will take extra effort to look at the commit id, and then detect in which commit it  was merged in trunk and from there get the ticket id, and then go to Trac to read the discussions.

src/twisted/internet/process.py
```diff
@@ -761,7 +793,8 @@ def pipe():
 
             if childFDs[childFD] == "w":
                 writer = self.processWriterFactory(reactor, self, childFD,
-                                        parentFD, forceReadHack=True)
+                                        parentFD,
+                                        forceReadHack=not brokenCygwinPipeBehavior)
```
adiroiban, line 797: maybe this can be left as it is, and overwrite this forceReadHack only for Cygwin, inside the factory/protocol implementation

src/twisted/internet/process.py
```diff
@@ -129,6 +133,34 @@ def detectLinuxBrokenPipeBehavior():
...
+    """
+
+    if sys.platform != 'cygwin':
+        return False
+
+    r, w = os.pipe()
+    os.close(r)
+    reads, writes, excs = select.select([w], [], [], 0)
+    os.close(w)
+    return not reads
```
adiroiban, line 156: I don't quite understand why we get this from here.

From the Trac ticket, my understand is that Cygwin always has this behaviour as there versions of Cygwin in which the pipes are not "broken"?

The docstring on this method does not contain "On some Cygwin" versions

src/twisted/internet/process.py
```diff
@@ -163,7 +195,7 @@ def __init__(self, reactor, proc, name, fileno, forceReadHack=False):
             self.enableReadHack = False
         elif forceReadHack:
             self.enableReadHack = True
-        else:
+        elif not brokenCygwinPipeBehavior:
```
adiroiban, line 198: my suggestion is to put this into a separate `if brokenCygwinPipeBehavior` and overwrite forceReadHack together with a comment

adiroiban (changes requested):
Thanks for working on this.

As a must, I think that we need:

* see if new members should be public
* see if we need the  function at all
* see how to test the changes

-----------

AFIK we don't have any slaves for cygwin so I don't know how are we going to do automated and continuous tests for this.

Have you used python + cygwin on appveyour 

It would be great to see a cygwin + python build for Twisted so that we can cover this platform on a regular basis.

Right now I don't have time to investigate how to run python on cygwin and then how to run in on appveyor but I am happy to do extra hours to review such work.

-------

You can also try to get someone over IRC or mailing list to manually run a one time test this of this on Cygwin and review that changes are working.

but it would be best to have automated and continous tests for Cygwin.

src/twisted/internet/process.py
embray: Yeah, I debated going either way on that.  Tbh I think the `forceReadHack` argument should not be passed on here at all, and instead allow the implementation to check whether or not it's needed for the relevant platform.

glyph:
See ticket closure explanation.

PR #926, opened 2017-11-08
9324 memoryview buffer

markrwilliams:
`lazyByteSlice` can use `memoryview`s now because:

1) PyOpenSSL uses cryptography now, which in turn uses cffi's
   `from_buffer` to slurp in binary data. from_buffer fully supports the
   new buffer protocol used by `memoryview`.

2) It's used in `FileDescriptor` implementations. For PyOpenSSL to be
   involved, Twisted would have to use wrapped sockets, but it uses
   memory BIOs above `FileDescriptor`s instead.

`twisted.internet.abstract._concatenate` can use `memoryview`s on Python 3.4+
because `memoryview`s can be joined with bytes, even though
`memoryview(...) + bytes(...)` still doesn't work.  Using a `memoryview`
prevents the slice from copying in the same was a buffer does under
Python 2.

This fixes a major performance regression under Python 3.

### DIFF
--- a/src/twisted/internet/abstract.py
+++ b/src/twisted/internet/abstract.py
@@ -30,6 +30,8 @@ if _PY3:
         def _concatenate(bObj, offset, bArray):
             return b''.join([memoryview(bObj)[offset:]] + bArray)
 else:
+    from __builtin__ import buffer
+
     def _concatenate(bObj, offset, bArray):
         # Avoid one extra string copy by using a buffer to limit what
         # we include in the result.
--- a/src/twisted/internet/abstract.py
+++ b/src/twisted/internet/abstract.py
@@ -13,21 +13,13 @@ from socket import AF_INET, AF_INET6, inet_pton, error
 from zope.interface import implementer
 
 # Twisted Imports
-from twisted.python.compat import unicode, lazyByteSlice
+from twisted.python.compat import unicode, lazyByteSlice, _PY3
 from twisted.python import reflect, failure
 from twisted.internet import interfaces, main
 
 import sys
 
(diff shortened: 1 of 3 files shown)


src/twisted/internet/abstract.py
```diff
@@ -26,13 +28,15 @@ def _concatenate(bObj, offset, bArray):
...
         return buffer(bObj, offset) + b"".join(bArray)
 except ImportError:
-    # Python 3
-    def _concatenate(bObj, offset, bArray):
-        # Python 3 lacks the buffer() builtin and the other primitives don't
-        # help in this case.  Just do the copy.  Perhaps later these buffers can
-        # be joined and FileDescriptor can use writev().  Or perhaps bytearrays
-        # would help.
-        return bObj[offset:] + b"".join(bArray)
+    if (sys.version_info.major, sys.version_info.minor) == (3, 3):
```
moshez: Can we just change everything to do version checking instead of a combination of version checking and import probing?

src/twisted/news/9324.bugfix
```diff
@@ -0,0 +1 @@
+Writing large amounts of data no longer implies repeated, expensive copying under Python 3.  Python 3's write speeds are now as fast as Python 2's.
```
moshez: Wrong place, should be in newsfragments
(Our classic trap)

moshez (approved):
Thank you for working on it!
One technical comment you really have to address (placement of news fragment) and one suggestion for streamlining the decision process.

Other than those, it looks good to go. Please address and merge.

PR #928, opened 2017-11-21
Remove _OpenSSLECCurve in favor of pyOpenSSL's own curve.

markrwilliams:
http://twistedmatrix.com/trac/ticket/7033

### DIFF
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1630,7 +1630,7 @@ class OpenSSLCertificateOptions(object):
 
     def _getEllipticCurve(self, name):
         """
-        A patchable wrapper for L{crypto.get_elliptic_curve}
+        A patchable wrapper for L{OpenSSL.crypto.get_elliptic_curve}
 
         @param name: The name of the elliptic curve to look up.
         @type name: L{unicode}
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1592,8 +1592,8 @@ class OpenSSLCertificateOptions(object):
         self.dhParameters = dhParameters
 
         try:
-            self._ecCurve = _OpenSSLECCurve(_defaultCurveName)
-        except NotImplementedError:
+            self._ecCurve = self._getEllipticCurve(_defaultCurveName)
+        except ValueError:
             self._ecCurve = None
 
         if acceptableCiphers is None:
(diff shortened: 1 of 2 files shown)


src/twisted/internet/_sslverify.py
```diff
@@ -1693,7 +1705,7 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
 
         if self._ecCurve is not None:
             try:
-                self._ecCurve.addECKeyToContext(ctx)
+                ctx.set_tmp_ecdh(self._ecCurve)
             except BaseException:
```
adiroiban, line 1709: minor comment: can we have a more specific exception here... just so that we don't accidentally mask other errors.

very minor comment:

I was thinking that we already do an eliptic curve check at `__init__` , maybe we should also move a check for `set_tmp_ecdh` there so that we check for support in a single place.

Regarding the best effort. Is this what most of the  people want?

If I am setting up a context with some settings, I am expecting to get an error if those settings can't be fulfilled so that I am aware of this as soon as possible.

adiroiban (approved):
It looks good. Thanks. Only very minor comments.

I am happy with the existing test_ecSuccessWithRealBindings 

Please update with master and merge, once all tests have passed.

src/twisted/internet/_sslverify.py
markrwilliams: > minor comment: can we have a more specific exception here... just so that we don't accidentally mask other errors.

It'd be nice to narrow this down so that it didn't catch, say, `SystemExit`, but I can't do that because OpenSSL's own documentation doesn't explain what errors `SSL_CTX_set_tmp_ecdh` can return.  That means I can't trace upwards to determine how `cryptography` and pyOpenSSL represent it.  I could dig through the code for various versions of OpenSSL, but I might get it wrong.  To my mind, then, changing this would risk introducing a regression in the name of a minor improvement, which I don't want to do. 

> I was thinking that we already do an eliptic curve check at __init__ , maybe we should also move a check for set_tmp_ecdh there so that we check for support in a single place.

This PR's goal is to just to eliminate the use of private APIs.   It's easier to see that this change is equivalent if it leaves the structure the same.  I think a future PR should clean this logic up.

> Regarding the best effort. Is this what most of the people want?

This code dates back to 2014, when fewer people used OpenSSLs with elliptic curve cryptography support compiled in, and the newest OpenSSLs didn't offer as wide as an array of ECDHE algorithms.  There was also no easy way to check what ECDHE algorithms, if any, were supported by the underlying OpenSSL installation.  The code this PR updates struck a balance between these constraints by attempting to select a commonly supported and safe curve, and assumed that ECC simply wasn't available if that failed.

Cryptography in 2017 is very different than in 2014. I think Twisted should require at least OpenSSL 1.1, which would ensure contexts were auto-configured with a variety of better curves than P256, but changing defaults and minimums is outside the scope of this PR.  

The way forward is #927, and I think that will be improved if this gets merged because it won't rely on private APIs in pyOpenSSL and `cryptography`.

PR #929, opened 2017-11-23
Enable test_inlineCallbacksTracebacks on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/9336

### DIFF
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1415,19 +1415,19 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         newFailure = self.failureResultOf(d)
         tb = traceback.extract_tb(newFailure.getTracebackObject())
 
-        self.assertEqual(len(tb), 2)
+        if _PY3:
+            self.assertEqual(len(tb), 3)
+            self.assertIn('test_defer', tb[2][0])
+            self.assertEqual('getDivisionFailure', tb[2][2])
+            self.assertEqual('1/0', tb[2][3])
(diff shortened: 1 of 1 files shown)


src/twisted/test/test_defer.py
```diff
@@ -1415,19 +1415,19 @@ def ic(d):
         newFailure = self.failureResultOf(d)
         tb = traceback.extract_tb(newFailure.getTracebackObject())
 
-        self.assertEqual(len(tb), 2)
+        if _PY3:
+            self.assertEqual(len(tb), 3)
+            self.assertIn('test_defer', tb[2][0])
+            self.assertEqual('getDivisionFailure', tb[2][2])
+            self.assertEqual('1/0', tb[2][3])
```
jlitzingerdev, line 1422: Out of curiosity, did you consider using the filename, line, and name attributes of the FrameSummary object?  As the objects are new in 3.5 that wouldn't work if Twisted supported versions pre 3.5, but I'm drawing a blank on whether that's true (doesn't look like anything pre 3.5 in python3 is in the build list).

Not a concern, it just might be more readable than the indicies into the FrameSummary object.

jlitzingerdev (approved):
These changes look good to me.  I did have one question, not a deal breaker, about using attributes of the FrameSummary object instead of the indicies, but as both are valid and the previous code used indicies I wouldn't consider it a blocker to a merge.

rodrigc:
The list of supported Python versions are in `src/twisted/python/_setup.py`.  Right now it is Python 2.7 and Python 3.3+

PR #931, opened 2017-11-25
Fix shell completion on Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9303

This now works on Python 3:
`twistd --_shell-completion zsh:1`

### DIFF
--- a/src/twisted/python/_shellcomp.py
+++ b/src/twisted/python/_shellcomp.py
@@ -28,6 +28,7 @@ Options howto.
 import itertools, getopt, inspect
 
 from twisted.python import reflect, util, usage
+from twisted.python.compat import ioType, unicode
 
 
 
@@ -157,6 +158,8 @@ class ZshBuilder(object):
         self.options = options
         self.cmdName = cmdName
         self.file = file
+        if file and ioType(file) == unicode:
+            self.file = file.buffer
 
 
     def write(self, genSubs=True):
@@ -269,6 +272,8 @@ class ZshArgumentsGenerator(object):
         self.options = options
         self.cmdName = cmdName
         self.file = file
+        if file and ioType(file) == unicode:
+            self.file = file.buffer
 
         self.descriptions = {}
         self.multiUse = set()
--- a/src/twisted/python/_shellcomp.py
+++ b/src/twisted/python/_shellcomp.py
@@ -22,7 +22,7 @@ functions for your own commands. You should only need to change the first line
 to something like C{#compdef mycommand}.
 
 The main public documentation exists in the L{twisted.python.usage.Options}
-docstring, the L{twisted.python.usage.Completions} docstring, and the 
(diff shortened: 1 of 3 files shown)


rodrigc:
@teratorn can you try this patch?

src/twisted/python/_shellcomp.py
```diff
@@ -157,6 +158,8 @@ def __init__(self, options, cmdName, file):
         self.options = options
         self.cmdName = cmdName
         self.file = file
+        if file and ioType(file) == unicode:
```
adiroiban: can the `file` ever be false ? 

Maybe we need a comment here to explain why we go with this strange file.buffer thing.

I would prefer to have the docstring updated to specify that `file` should be a BytesIO.

Then the caller of ZshBuilder should take care of callling it with the right IO.

So the fix should be in `shellComplete(config, cmdName, words, shellCompFile)` or in the caller of shellComplete

In this way, we will only fix the input stream in a single place and it will be fixed at the entry point, and not somewhere in the middle of this processing.

adiroiban:
I run a manual test and it looks good. It fails on master and passed on this branch

I am approving this as it is better than what we have in trunk.

Feel free to update based on the comments or merge as it is.

----------------

I see that the zsh completion tests are already passing on trunk 

```
  ZshTests
    test_accumulateAdditionalOptions ...                                   [OK]
```

https://buildbot.twistedmatrix.com/builders/debian9-py3.5/builds/513/steps/epoll/logs/stdio 

With that I assume that there are no tests for the reported defect.

But I think that is fine. If something breaks in zsh completion I expect that people will notice and any bug will not have a big impact on any production system

-----

See my inline comment.

I would like to have the `shellCompFile` "patched" only once. Maybe in  shellComplete or even better in twisted.python.usage.Option.parseOptions


Thanks!

adiroiban (approved):

teratorn:
@rodrigc looks good to me

src/twisted/python/_shellcomp.py
rodrigc: There are some tests which do things like:
```
argGen = _shellcomp.ZshArgumentsGenerator(opts, 'ace', None)
```
so apparently that is allowed.

rodrigc: After going through other parts of the Twisted code, especially Twisted written before Python 3, that when an argument of C{file} is specified, it really means a C{file} with bytes-semantics, such as BytesIO.  That's not clear until you dig into things.

However, the way that shellComplete() is used, `sys.stdout` is usually passed as the file argument.   isinstance(sys.stdout, io.BytesIO) == False even on Python 2, so it is not quite correct to say that this parameter should be a BytesIO.  It must be a file with bytes semantics.

So I think moving this check up into shellComplete() is fine.

PR #933, opened 2017-11-25
Port twisted.mail.test.test_mail to Python 3

rodrigc:
http://twistedmatrix.com/trac/ticket/9337


src/twisted/mail/alias.py
```diff
@@ -40,14 +42,15 @@ def handle(result, line, filename, lineNo):
...
+    parts = [p.strip() for p in line.split(b':', 1)]
     if len(parts) != 2:
         fmt = "Invalid format on line %d of alias file %s."
         arg = (lineNo, filename)
         log.err(fmt % arg)
     else:
         user, alias = parts
-        result.setdefault(user.strip(), []).extend(map(str.strip, alias.split(',')))
+        result.setdefault(user.strip(), []).extend(
+            [a.strip() for a in alias.split(b',')])
```
wiml, line 53: I don't think it's necessary to collect the addresses into a list before passing them to `extend()`; it can take a generator directly

src/twisted/mail/alias.py
```diff
@@ -220,7 +223,7 @@ def createMessageReceiver(self):
         @rtype: L{IMessage <smtp.IMessage>} provider
         @return: A message receiver.
         """
-        return self.domain().exists(str(self.alias))
+        return self.domain().exists(str(self.alias).encode("ascii"))
```
wiml, line 226: Given the set of types that `self.alias` is documented to be, this seems like the wrong way to convert to a bytestring. 

`IDomain.exists()` is documented to take only a `User`, not a bytestring — either that documentation or this code must be wrong.

src/twisted/mail/alias.py
```diff
@@ -723,18 +726,18 @@ def __init__(self, items, *args):
         self.aliases = []
         while items:
             addr = items.pop().strip()
-            if addr.startswith(':'):
+            if addr.startswith(b':'):
                 try:
-                    f = open(addr[1:])
-                except:
+                    f = open(addr[1:], "rb")
+                except Exception:
```
wiml, line 732: Is it appropriate to swallow the error like this without propagating the error to the caller?

I think either the `except:` clause should be narrowed to only "invalid filenames", or the log message should include the exception that was raised. (Perhaps both: narrow to `OSError` and `ValueError`, and also report the exception?)

src/twisted/mail/maildir.py
```diff
@@ -483,14 +488,16 @@ def createTempFile(self):
...
+                _generateMaildirName()).asBytesMode().path
             try:
                 self.fh = self.osopen(self.tmpname, attr, 0o600)
                 return None
             except OSError:
                 tries += 1
                 if tries > 500:
-                    self.defer.errback(RuntimeError("Could not create tmp file for %s" % self.mbox.path))
+                    self.defer.errback(RuntimeError(
+                        b"Could not create tmp file for " + self.mbox.path))
```
wiml, line 500: `t.p.filepath.FilePath.path` is not necessarily `bytes`, so this may raise.

src/twisted/mail/maildir.py
```diff
@@ -96,14 +97,15 @@ def initializeMaildir(dir):
     @type dir: L{bytes}
     @param dir: The path name for a user directory.
     """
-    if not os.path.isdir(dir):
-        os.mkdir(dir, 0o700)
+    dirPath = FilePath(dir)
+    if not dirPath.isdir():
+        os.mkdir(dirPath.path, 0o700)
```
wiml, line 102: Is there a reason to convert to `FilePath` and immediately back to string/bytes? Py3 filesystem calls are string/bytes flexible so it seems like the old code would work fine.

If we're using `t.p.filepath.FilePath`, though, should we be using its `createDirectory()` method here instead of the `os` module?

src/twisted/mail/maildir.py
```diff
@@ -132,7 +134,7 @@ def __init__(self, address, fp, *a, **kw):
         @type kw: L{dict}
         @param kw: Keyword arguments for L{FileMessage.__init__}.
         """
-        header = "Delivered-To: %s\n" % address
+        header = u"Delivered-To: {}\n".format(address).encode("ascii")
```
wiml, line 137: I think this should be `b"Delivered-To: {}\n".format(address.__bytes__())` or the equivalent, rather than converting to Unicode and back.

src/twisted/mail/maildir.py
```diff
@@ -237,7 +240,7 @@ def exists(self, user, memo=None):
             return lambda: self.startMessage(user)
         try:
             a = self.alias[user.dest.local]
-        except:
+        except Exception:
```
wiml, line 243: Could this be `except KeyError:`, or do we really intend to swallow all exceptions here?

src/twisted/mail/maildir.py
```diff
@@ -261,12 +264,13 @@ def startMessage(self, user):
...
             name, domain = user.dest.local, user.dest.domain
-        dir = self.userDirectory(name)
+        dirPath = FilePath(self.userDirectory(name))
         fname = _generateMaildirName()
-        filename = os.path.join(dir, 'tmp', fname)
-        fp = open(filename, 'w')
-        return MaildirMessage('%s@%s' % (name, domain), fp, filename,
-                              os.path.join(dir, 'new', fname))
+        filename = dirPath.child('tmp').child(fname).asBytesMode().path
+        fp = open(filename, 'wb')
```
wiml, line 270: Same comment as with `initializeMaildir()`: I think if we're converting to `FilePath`s, it's appropriate to use the FilePath methods instead of the `os` module.

In any event, the `asBytesMode()` seems suspect to me.

src/twisted/mail/maildir.py
```diff
@@ -261,12 +264,13 @@ def startMessage(self, user):
...
+        return MaildirMessage(
+            u'{}@{}'.format(name, domain).encode("ascii"), fp, filename,
```
wiml, line 272: `name`/`domain` may be either `str` or `bytes` here; this will produce garbled results if they are `bytes`

src/twisted/mail/maildir.py
```diff
@@ -483,14 +488,16 @@ def createTempFile(self):
         tries = 0
         self.fh = -1
         while True:
-            self.tmpname = os.path.join(self.mbox.path, "tmp", _generateMaildirName())
+            self.tmpname = FilePath(self.mbox.path).child("tmp").child(
+                _generateMaildirName()).asBytesMode().path
```
wiml, line 492: Is `asBytesMode()` needed?

Can we use `FilePath.open()` here?

src/twisted/mail/relay.py
```diff
@@ -45,8 +45,8 @@ def exists(self, user):
         """
         if self.willRelay(user.dest, user.protocol):
             # The most cursor form of verification of the addresses
-            orig = filter(None, str(user.orig).split('@', 1))
-            dest = filter(None, str(user.dest).split('@', 1))
+            orig = [c for c in str(user.orig).split('@', 1) if c]
+            dest = [c for c in str(user.dest).split('@', 1) if c]
```
wiml, line 49: Isn't this splitting-and-counting equivalent to `'@' in user.orig and '@' in user.dest`, or am I missing something?

src/twisted/mail/relaymanager.py
```diff
@@ -412,7 +414,7 @@ def addMessage(self, message):
         if message not in self.relayed:
             self.waiting[message] = 1
             if self.noisy:
-                log.msg('Set ' + message + ' waiting')
+                log.msg(b'Set ' + message + b' waiting')
```
wiml, line 417: I think it would be better to convert `message` to unicode  (e.g. by `repr` ?) than to emit the log text as a bytestring.

src/twisted/mail/relaymanager.py
```diff
@@ -563,7 +567,7 @@ def notifySuccess(self, relay, message):
         @param message: The path of the file holding the message.
         """
         if self.noisy:
-            log.msg("success sending %s, removing from queue" % message)
+            log.msg(b"success sending " + message + b", removing from queue")
```
wiml, line 570: I think it would be better to convert `message` to unicode  (e.g. by `repr` ?) than to emit the log text as a bytestring.

src/twisted/mail/relaymanager.py
```diff
@@ -578,15 +582,15 @@ def notifyFailure(self, relay, message):
         @param message: The path of the file holding the message.
         """
         if self.noisy:
-            log.msg("could not relay " + message)
+            log.msg(b"could not relay " + message)
```
wiml, line 585: I think it would be better to convert `message` to unicode  (e.g. by `repr` ?) than to emit the log text as a bytestring.

src/twisted/mail/relaymanager.py
```diff
@@ -607,7 +611,7 @@ def notifyDone(self, relay):
         """
         for message in self.manager.managed.get(relay, ()):
             if self.noisy:
-                log.msg("Setting " + message + " waiting")
+                log.msg(b"Setting " + message + b" waiting")
```
wiml, line 614: I think it would be better to convert `message` to unicode  (e.g. by `repr` ?) than to emit the log text as a bytestring.

src/twisted/mail/relaymanager.py
```diff
@@ -776,7 +780,7 @@ def _checkStateMX(self):
             if len(parts) != 2:
                 log.err("Illegal message destination: " + to)
                 continue
-            domain = parts[1]
+            domain = parts[1].encode("ascii")
```
wiml, line 783: Domains should probably be encoded as `"idna"`

src/twisted/mail/relaymanager.py
```diff
@@ -841,11 +845,11 @@ def _ebExchange(self, failure, factory, domain):
         @type domain: L{bytes}
         @param domain: A domain.
         """
-        log.err('Error setting up managed relay factory for ' + domain)
+        log.err(b'Error setting up managed relay factory for ' + domain)
         log.err(failure)
 
         def setWaiting(queue, messages):
-            map(queue.setWaiting, messages)
+            [queue.setWaiting(msg) for msg in messages]
```
wiml, line 852: There's no need to collect the results in a list; perhaps an ordinary for-loop would be clearer here

src/twisted/mail/relaymanager.py
```diff
@@ -841,11 +845,11 @@ def _ebExchange(self, failure, factory, domain):
         @type domain: L{bytes}
         @param domain: A domain.
         """
-        log.err('Error setting up managed relay factory for ' + domain)
+        log.err(b'Error setting up managed relay factory for ' + domain)
```
wiml, line 848: Same comment as above about leaving the text as text and converting the `domain` to textual form for logging

src/twisted/mail/relaymanager.py
```diff
@@ -1096,9 +1100,9 @@ def _cbMX(self, answers, domain, cnamesLeft):
                 exchanges.append((record.preference, record))
 
         if exchanges:
-            exchanges.sort()
+            exchanges = sorted(exchanges, key=lambda k: str(k))
```
wiml, line 1103: This seems incorrect: we should sort numerically on the preference field, not lexicographically on the repr. Maybe `key=lambda k: k[0]` ?

Is there a reason to use `sorted()` instead of sorting in-place?

wiml (changes requested):
alias.py line 39: on py3, filenames may be either bytes or str; see https://docs.python.org/3/library/os.path.html

maildir.py line 253: the docstring for the `user` parameter to `startMessage()` says `bytes`, but the code says it can be either `str` or `User`; and elsewhere it is called with a `User`

src/twisted/mail/alias.py
thijstriemstra: This can be done in a different ticket, it is not related to porting to Python 3.

PR #934, opened 2017-11-26
[9340] SMTPSenderFactory should not mutate the Failure receive at connection lost.

adiroiban:
Scope
=====

This fixes https://twistedmatrix.com/trac/ticket/9340

Changes
=======

The connectionLost handling was updated to produce a new error instead of mutating the failure received from upstream code.

I have refactored to tests to reuse as much code as possible.

The unit tests were updated to pass a failure as we expect to see in real code instead of direclty the exception.

There was no coverage for the case when failure is done before finishing, so I wrote tests for all 3 branches.

I have not written an assertion for the side-effect part, but I expect that since that code is now triggered as part of the automated tests, it there are any errors we would observer it in other tests.

How to test
==========

Check the changes make sense.


src/twisted/mail/smtp.py
```diff
@@ -1937,11 +1937,19 @@ def _processConnectionError(self, connector, err):
...
-            # If we were unable to communicate with the SMTP server a ConnectionDone will be
-            # returned. We want a more clear error message for debugging
+
             if err.check(error.ConnectionDone):
-                err.value = SMTPConnectError(-1, "Unable to connect to server.")
-            self.result.errback(err.value)
+                # If we were unable to communicate with the SMTP server,
+                # a ConnectionDone will be returned.
+                # We want a more clear error message for debugging.
+                reason = SMTPConnectError(-1, "Unable to connect to server.")
```
markrwilliams, line 1945: I'd prefer to see this `raise`d so it includes a traceback.

src/twisted/mail/test/test_smtp.py
```diff
@@ -937,17 +980,74 @@ def test_removeCurrentProtocolWhenClientConnectionFailed(self):
...
+        sut = self.getFactory(sentDeferred=sentDeferred, retries=0)
+        connector = self.connect(reactor, sut)
+
+        sut.clientConnectionLost(
+            connector, Failure(error.ConnectionDone("Bye.")))
+
+        failure = self.failureResultOf(sentDeferred)
+        self.assertIsInstance(failure.value, smtp.SMTPConnectError)
+        # The delivery deferred is no longer available on the factory.
+        self.assertFalse(hasattr(sut, 'result'))
```
markrwilliams, line 1011: Thanks for testing something that wasn't tested before!  Unfortunately, [`hasattr` sometimes lies under Python 2](https://hynek.me/articles/hasattr/).  Please replace this with `getattr`.

src/twisted/mail/test/test_smtp.py
```diff
@@ -937,17 +980,74 @@ def test_removeCurrentProtocolWhenClientConnectionFailed(self):
...
+        error = RuntimeError(b'error-marker')
+        reactor = MemoryReactor()
+        sentDeferred = defer.Deferred()
+        sut = self.getFactory(sentDeferred=sentDeferred, retries=0)
+        connector = self.connect(reactor, sut)
+
+        sut.clientConnectionLost(connector, Failure(error))
+
+        failure = self.failureResultOf(sentDeferred)
+        self.assertIs(error, failure.value)
```
markrwilliams, line 1029: Shouldn't this also check that `result` is unset?

src/twisted/mail/test/test_smtp.py
```diff
@@ -937,17 +980,74 @@ def test_removeCurrentProtocolWhenClientConnectionFailed(self):
...
+        sut = self.getFactory(sentDeferred=sentDeferred, retries=3)
+        connector = self.connect(reactor, sut)
+        sut.sendFinished = True
+
+        sut.clientConnectionLost(
+            connector, Failure(error.ConnectionDone("Bye.")))
+
+        # It does not touches the deferred used to signal the message
+        # completion.
+        self.assertNoResult(sentDeferred)
```
markrwilliams, line 1048: Should this also test that the factory has a `result` attribute?

markrwilliams (approved):
Thanks for removing unexpected mutation of shared state!

Please consider the comments before merging.

twm:
@adiroiban Will you be able to address Mark's comments on this PR at some point?

PR #935, opened 2017-11-28
9210 ecdh selection 1

markrwilliams:
A continuation of #927, which was closed for mysterious reasons.

### DIFF
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1591,10 +1591,14 @@ class OpenSSLCertificateOptions(object):
             self._options |= SSL.OP_NO_TICKET
         self.dhParameters = dhParameters
 
-        try:
-            self._ecCurve = _OpenSSLECCurve(_defaultCurveName)
-        except NotImplementedError:
-            self._ecCurve = None
+        self._ecCurve = None
+        # Only select a curve for OpenSSL < 1.0.2, in 1.0.2+ we use the
+        # automatic mode that supports multiple curves
+        if SSL.OPENSSL_VERSION_NUMBER < 0x10002000:
+            try:
+                self._ecCurve = _OpenSSLECCurve(_defaultCurveName)
+            except NotImplementedError:
+                pass;
 
         if acceptableCiphers is None:
             acceptableCiphers = defaultCiphers
@@ -1690,6 +1694,14 @@ class OpenSSLCertificateOptions(object):
         if self.dhParameters:
             ctx.load_tmp_dh(self.dhParameters._dhFile.path)
         ctx.set_cipher_list(self._cipherString.encode('ascii'))
+
+        # OpenSSL 1.0.2 introduced the SSL_CTX_set_ecdh_auto so that you don't
+        # need to set a supported key. OpenSSL 1.1.0 removed the function
+        # again and it's always enabled.
+        try:
+            self._lib.SSL_CTX_set_ecdh_auto(ctx, 1);
+        except BaseException:
+            pass  # ECDHE support is best effort only.
 
         if self._ecCurve is not None:
             try:
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1875,8 +1875,8 @@ class _ChooseDiffieHellmanEllipticCurve(object):
                 self._ecCurve = openSSLcrypto.get_elliptic_curve(
                     _defaultCurveName)
             except:
-                # get_elliptic_curve raises a ValueError when the
-                # curve does not exist.
+                # The get_elliptic_curve method raises a ValueError
+                # when the curve does not exist.
                 self.configureECDHCurve = self._configureOpenSSL101NoCurves
             else:
                 self.configureECDHCurve = self._configureOpenSSL101
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1874,7 +1874,7 @@ class _ChooseDiffieHellmanEllipticCurve(object):
             try:
                 self._ecCurve = openSSLcrypto.get_elliptic_curve(
                     _defaultCurveName)
-            except:
+            except ValueError:
                 # The get_elliptic_curve method raises a ValueError
                 # when the curve does not exist.
                 self.configureECDHCurve = self._configureOpenSSL101NoCurves
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -12,6 +12,7 @@ from constantly import Names, NamedConstant
 from hashlib import md5
 
 from OpenSSL import SSL, crypto
+from OpenSSL._util import lib as pyOpenSSLlib
 
 from twisted.python import log
 from twisted.python._oldstyle import _oldStyle
@@ -1319,7 +1320,6 @@ class OpenSSLCertificateOptions(object):
 
     _defaultMinimumTLSVersion = TLSVersion.TLSv1_0
 
-
     @_mutuallyExclusiveArguments([
         ['trustRoot', 'requireCertificate'],
         ['trustRoot', 'verify'],
@@ -1591,14 +1591,11 @@ class OpenSSLCertificateOptions(object):
             self._options |= SSL.OP_NO_TICKET
         self.dhParameters = dhParameters
 
-        self._ecCurve = None
-        # Only select a curve for OpenSSL < 1.0.2, in 1.0.2+ we use the
-        # automatic mode that supports multiple curves
-        if SSL.OPENSSL_VERSION_NUMBER < 0x10002000:
-            try:
-                self._ecCurve = self._getEllipticCurve(_defaultCurveName)
-            except ValueError:
-                self._ecCurve = None
+        self._ecChooser = _ChooseDiffieHellmanEllipticCurve(
+            SSL.OPENSSL_VERSION_NUMBER,
+            openSSLlib=pyOpenSSLlib,
+            openSSLcrypto=crypto,
+        )
 
         if acceptableCiphers is None:
             acceptableCiphers = defaultCiphers
@@ -1630,18 +1627,6 @@ class OpenSSLCertificateOptions(object):
             )
 
         self._acceptableProtocols = acceptableProtocols
-
-
-    def _getEllipticCurve(self, name):
-        """
-        A patchable wrapper for L{OpenSSL.crypto.get_elliptic_curve}
-
-        @param name: The name of the elliptic curve to look up.
-        @type name: L{unicode}
-
-        @return: A pyOpenSSL elliptic curve object.
-        """
-        return crypto.get_elliptic_curve(name)
 
 
     def __getstate__(self):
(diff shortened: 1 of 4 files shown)


markrwilliams:
@kroeckx Sorry for the inconvenience of a new PR.  We'll get this merged soon.

markrwilliams:
@adiroiban Can you review this now?  This includes #928.  You might want to look at the [now-deceased #927 for context](https://github.com/twisted/twisted/pull/927#issuecomment-347438828).

adiroiban:
I think that this should be reviewed by the people who have initially reviewed #927 .
For example get a review from @kroeckx to confirm that this PR is solving the initial issues.

Thanks!

markrwilliams:
@kroeckx I have confirmed with Wireshark that this PR presents more than one ECDHE curve in a ClientHello when used with OpenSSL >= 1.0.2.  Please take a look at the code when you can.

kroeckx:
The code looks fine, since it's mostly what I wrote. If the client hello from both 1.0.2 and 1.1.0 show multiple curves, things work as expected.

I currently don't have time to test this myself.

src/twisted/internet/_sslverify.py
```diff
@@ -1703,6 +1707,14 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
             ctx.load_tmp_dh(self.dhParameters._dhFile.path)
         ctx.set_cipher_list(self._cipherString.encode('ascii'))
 
+        # OpenSSL 1.0.2 introduced the SSL_CTX_set_ecdh_auto so that you don't
+        # need to set a supported key. OpenSSL 1.1.0 removed the function
+        # again and it's always enabled.
+        try:
+            self._lib.SSL_CTX_set_ecdh_auto(ctx, 1);
+        except BaseException:
```
adiroiban: I think that here we should have an explicit exception.
I am looking at his code, but I have no idea where the `self._lib` came from.

Maybe this should be `OpenSSL._util.lib.SSL_CTX_set_ecdh_auto` ?

I think that we need to new test to check that on other systems it raises the expected exception.

I ran some tests for 

```
pyOpenSSL OpenSSL 0.9.8j-fips 07 Jan 2009 (9470127)
pyOpenSSL 16.0.0

pyOpenSSL OpenSSL 1.0.1i-fips 6 Aug 2014 (268439711)
pyOpenSSL 17.0.0

>>> import OpenSSL
>>> OpenSSL._util.lib.SSL_CTX_set_ecdh_auto
AttributeError: 'module' object has no attribute 'SSL_CTX_set_ecdh_auto'


```

adiroiban: I have tried on 1.1.0 and the method is still there, so the comment is misleading :)

```
pyOpenSSL OpenSSL 1.1.0f  25 May 2017 (269484143)
pyOpenSSL 17.0.0

>>> import OpenSSL
>>> import OpenSSL.SSL
>>> OpenSSL.SSL.OPENSSL_VERSION_NUMBER
269484143
>>> OpenSSL._util.lib.SSL_CTX_set_ecdh_auto
<built-in method SSL_CTX_set_ecdh_auto of CompiledLib object at 0x7f7705e34c58>

**Review by adiroiban — CHANGES_REQUESTED** [review]
Thanks for working on this.

I bite the bullet and reviewed the code and ran some tests.

See inline comments.

I think that we should only catch AttributeError and maybe isolate it as

```
try:
  ecdh_auto = OpenSSL._util.lib.SSL_CTX_set_ecdh_auto
except AttributeError:
  pass
else:
  ecdh_auto(ctx,1)
```

In this way we should reduce the risk of catching an accidental error.

And it looks like it works even with older stuff like

```
pyOpenSSL OpenSSL 0.9.7d 17 Mar 2004 (+ security fixes for: REMOVED)
pyOpenSSL 0.13

>>> import OpenSSL
>>> OpenSSL._util.lib.SSL_CTX_set_ecdh_auto
AttributeError: 'module' object has no attribute '_util'
```

File: src/twisted/internet/_sslverify.py (hunk 1 above)
Comment (kroeckx) [reply src/twisted/internet/_sslverify.py]: OpenSSL 1.1.0 changed SSL_CTX_set_ecdh_auto to a define that does nothing. It always was a define, but in 1.0.2 it called SSL_CTX_ctrl().

Comment (kroeckx) [reply src/twisted/internet/_sslverify.py]: The self._lib thing comes from the _OpenSSLECCurve class, I just copied it not understanding what I was doing at the time. I also pointed that out that I had no idea what I was doing with regard to this function when I created the merge request.

**Comment by kroeckx** [comment]
So what is the state of this?

**Comment by adiroiban** [comment]
> So what is the state of this?

@kroeckx this was sent back to the author as the proposed changes are of bad quality, especially the `self._lib` part.

I assume that Mark was busy and did not had time to check on this.
I am happy to review any patch :)

**Comment by markrwilliams** [comment]
Correct.  It's the holiday season in the US and I'm busy.

Inasmuch as I've had time to devote to this, I've been looking into the feasibility of adding Buildbot environments that can run Twisted's test suite against OpenSSL 1.0.1 and 1.0.2.  I had hoped that this wouldn't be that much work; unfortunately, it's enough work that I can't do it without delaying this patch even more.

We have Python 3.3 building against OpenSSL 1.0.1, so I can add a test for that.  If somebody else wants to pick this up before the end of the week, please do.  Otherwise I'll address @adiroiban's comments this weekend.

**Comment by markrwilliams** [comment]
I think this addresses the bug, but I expect it to have coverage issues.  [See Trac for the gory details](https://twistedmatrix.com/trac/ticket/9210#comment:15).

File: tox.ini [hunk 2]
```diff
@@ -79,7 +79,7 @@ setenv =
    TOX_INI_DIR = {toxinidir}
 
    ; Skip twistedchecker warnings that are not yet required in the required run
-   txchecker-travis-required: TWISTEDCHECKER_SKIP_WARNINGS = W9208,C0302
+   txchecker-travis-required: TWISTEDCHECKER_SKIP_WARNINGS = W9208,C0302,C0103,C9302
```
markrwilliams, line 87: This deactivates C0103 (invalid name) C9302 (invalid method name), because the tests fake OpenSSL APIs like `set_tmp_ecdh` that violate our coding standard.

These should be removed once https://github.com/twisted/twistedchecker/issues/120 and https://github.com/twisted/twistedchecker/issues/76 are closed.

adiroiban: maybe add a FIXME comment here with reference to those tickets

alex:
Notes:

- There's way too much indirection here, there should not be an `ecChooser` and such, there should just be a function which takes a ctx and an OpenSSL version and configures it correctly
- Right now the tests don't really test that anything works, they test that various combinations of mocks work correctly. I think you'd be way better off making in memory TLS connections and introspecting the configured curve.
- There's a ton of bare `except` clauses that should be turned into something precise.

adiroiban:
> There's way too much indirection here, there should not be an ecChooser and such, there should just be a function which takes a ctx and an OpenSSL version and configures it correctly

I am fine with the current ecChooser, which for me is the same think as the function which Alex wants.
I find it easier to read and document.

> Right now the tests don't really test that anything works

When I saw all the extra testing objects and knowing how small the initial patch was, I was thinking http://www.osnews.com/story/19266/WTFs_m 

But on a second thought :)

I know that writing integration/functional is a lot of work,  so I don't know what to say.
We can ask for those tests and don't have this merge, or merge it in good faith :)

As long as it is a bit better than what we have in master, I think that we should merge it.

-------

+1 for not using bare exceptions... I remember rejecting this PR for an error in the way a bare exception was used.

src/twisted/internet/_sslverify.py
```diff
@@ -1850,6 +1835,103 @@ def fromOpenSSLCipherString(cls, cipherString):
...
+        self._openSSLcrypto = openSSLcrypto
+        if openSSLVersion >= 0x10100000:
+            self.configureECDHCurve = self._configureOpenSSL110
+        elif openSSLVersion >= 0x10002000:
+            self.configureECDHCurve = self._configureOpenSSL102
+        else:
+            try:
+                self._ecCurve = openSSLcrypto.get_elliptic_curve(
+                    _defaultCurveName)
+            except:
```
adiroiban: we should have this in twistedchecker ...

pycodestyle already catches these erros so maybe we should look at adding pycodestyle with selected errors.

Twisted code is not pep8 but some checks in pycodestyle are not pep-8 specific and we should use them

markrwilliams:
> There's way too much indirection here, there should not be an ecChooser and such, there should just be a function which takes a ctx and an OpenSSL version and configures it correctly

[`_ChooseDiffieHellmanEllipticCurve`](https://github.com/twisted/twisted/pull/935/files#diff-df501bac724aab523150498f84749b88R1838) is equivalent to a closure that takes some injectable dependencies, including the OpenSSL version, and returns a function that takes an `SSLContext` and configures it correctly.  I could eliminate this class in favor of a module-level function conditionally defined on the OpenSSL version, but:

1. It would be slightly harder to remove when we drop support for OpenSSL 1.0.1.  With this patch, removal hinges on eliminating the syntactically delineated `_ChooseDiffieHellmanEllipticCurve` class, not a semantically delineated if statement;
2. [It would be harder to test](https://glyph.twistedmatrix.com/2007/07/functional-functions-and-python.html), and as I'll discuss next, this patch is already hard to test;

> Right now the tests don't really test that anything works, they test that various combinations of mocks work correctly. I think you'd be way better off making in memory TLS connections and introspecting the configured curve.

[This patch includes the same test](https://github.com/twisted/twisted/pull/935/files#diff-a0749b8aa4682956a79714cc4a69efd8R1602) that [CPython has](https://github.com/python/cpython/commit/0bebbc33faae7ac10e7a7980b260e786f05d81bf), but it only asserts that _some_ elliptic curve was and will be used for key exchange (note that the [CPython OpenSSL 1.0.2 compatability patch](https://bugs.python.org/issue29697) I'm reproducing here included no additional tests.) [As far as I've read](https://crypto.stackexchange.com/questions/11310/with-openssl-and-ecdhe-how-to-show-the-actual-curve-being-used) OpenSSL does not have the necessary introspection APIs to allow user code to determine the actual curve a connection uses for ECDH.  Unfortunately that's exactly what this ticket is about; right now, Twisted sets one curve on `SSLContext`s when it should allow the default set of multiple curves when that's possible.  The best I can do is test that if Twisted calls the right pyOpenSSL APIs, it will create `SSLContext`s with the right curve(s).  [I have integration tests in another repository](https://github.com/markrwilliams/twisted-ssl-environments) that validate Twisted actually does this.  I'd love to see builders that ran Twisted under all the supported major versions of OpenSSL, but setting that up is outside the scope of this PR.

> There's a ton of bare except clauses that should be turned into something precise.

I carried these forward from the original code [on the assumption that the previous review was conducted by people who knew both pyOpenSSL and OpenSSL better than I do.](https://twistedmatrix.com/trac/ticket/6586)  If they thought that `except:` was what should be caught to keep ECDH support optional, then I'll do the same.  A separate ticket with suitable fanfare on the mailing list can change that and remove or refine these `except`s, unless it's possible to determine precisely what exceptions can be raised.  It wasn't clear to me at the time by reading OpenSSL and pyOpenSSL's code.  I'd be happy to take direction here.

adiroiban:
@markrwilliams I am on the same page with you regarding ecChooser :)

My only concern for this PR is the use of `except:` . I think that even if the old code was using it, as a drive by we can use at least `except: Exception` so that we don't catch SystemExit or keyboard interrupt.

And I think that the bare `except:` was there due to negligence :) I don't think that for OpenSSL compat purpose anybody wanted to catch a SystemExit :)

I guess that people are not discussion about this as pycodestyle (ex pep8) tool will nag about this and everybody will just update the code :)

But, as we saw in this PR, even a BaseException is not enough as it can hide errors https://github.com/twisted/twisted/pull/935/commits/3259b7c1f0262b00b3acaae9f7dfa6764fcd1634#r154221072

This is why for the scope of this PR I am -1 of using bare `except: `

PR #936, opened 2017-11-28
Add toBytes() helper method

rodrigc:
https://twistedmatrix.com/trac/ticket/9343

We can replace:

```
if isinstance(foo, unicode):
    foo = foo.encode("ascii")
```

with:

```
foo = toBytes(foo)
```

### DIFF
--- a/src/twisted/newsfragments/9343.feature
+++ b/src/twisted/newsfragments/9343.feature
@@ -1,0 +1,1 @@
+twisted.python.compat.toBytes is a convenience function for converting unicode strings to bytes.
--- a/src/twisted/python/compat.py
+++ b/src/twisted/python/compat.py
@@ -822,6 +822,32 @@ def _bytesRepr(bytestring):
         return 'b' + repr(bytestring)
 
 
+
+def toBytes(s, encoding="ascii", errors="strict"):
+    """
+    Convert L{unicode} string to L{bytes}.
+
+    @param s: The string to convert.
+    @param encoding: The encoding to pass to L{str.encode}.
+    @param errors: The error handling scheme to pass to L{str.encode).
+
+    @raise TypeError: The input is not L{unicode} or L{bytes}.
+
+    @return: The encoded string.  If I{s} is L{bytes} or L{None}, just return
+             I{s}.
+    @rtype: L{bytes}
+    """
+    if s is None:
+        return s
+    elif isinstance(s, bytes):
+        return s
+    elif isinstance(s, unicode):
+        return s.encode(encoding=encoding, errors=errors)
+    else:
+        raise TypeError("Expected {} to be unicode or bytes, not: {}".format(
+            s, type(s)))
+
+
 if _PY3:
     _tokenize = tokenize.tokenize
 else:
@@ -868,5 +894,6 @@ __all__ = [
     "intern",
     "unichr",
     "raw_input",
-    "_tokenize"
+    "_tokenize",
+    "toBytes"
 ]
--- a/src/twisted/python/compat.py
+++ b/src/twisted/python/compat.py
@@ -829,9 +829,9 @@ def toBytes(s, encoding="ascii", errors="strict"):
 
     @param s: The string to convert.
     @param encoding: The encoding to pass to L{str.encode}.
-    @param errors: The error handling scheme to pass to L{str.encode).
+    @param errors: The error handling scheme to pass to L{str.encode}.
 
-    @raise TypeError: The input is not L{unicode} or L{bytes}.
+    @raise TypeError: The input is not L{unicode}, L{bytes} or L{None}.
 
     @return: The encoded string.  If I{s} is L{bytes} or L{None}, just return
              I{s}.
@@ -844,8 +844,8 @@ def toBytes(s, encoding="ascii", errors="strict"):
     elif isinstance(s, unicode):
         return s.encode(encoding=encoding, errors=errors)
     else:
-        raise TypeError("Expected {} to be unicode or bytes, not: {}".format(
-            s, type(s)))
+        raise TypeError("Expected {} to be unicode, bytes, or None not: "
+            "{}".format(s, type(s)))
 
 
 if _PY3:
--- a/src/twisted/test/test_compat.py
+++ b/src/twisted/test/test_compat.py
@@ -928,7 +928,7 @@ class FutureBytesReprTests(unittest.TestCase):
 
 class ToBytesTests(unittest.TestCase):
     """
-    L{toBytes}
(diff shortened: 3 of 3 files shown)


src/twisted/python/compat.py
```diff
@@ -822,6 +822,32 @@ def _bytesRepr(bytestring):
...
 
 
+
+def toBytes(s, encoding="ascii", errors="strict"):
+    """
+    Convert L{unicode} string to L{bytes}.
+
+    @param s: The string to convert.
+    @param encoding: The encoding to pass to L{str.encode}.
+    @param errors: The error handling scheme to pass to L{str.encode).
```
jlitzingerdev: There is an error on this line that breaks API doc generation for this param.  Specifically, L{str.encode) <- should be a }.

src/twisted/python/compat.py
```diff
@@ -822,6 +822,32 @@ def _bytesRepr(bytestring):
         return 'b' + repr(bytestring)
 
 
+
+def toBytes(s, encoding="ascii", errors="strict"):
+    """
+    Convert L{unicode} string to L{bytes}.
+
+    @param s: The string to convert.
+    @param encoding: The encoding to pass to L{str.encode}.
```
jlitzingerdev, line 831: Why no `@type` for the parameters?  Especially since s can be None as well as bytes or unicode.

jlitzingerdev (changes requested):
Thanks for this, I like this API, It provides a nice way of converting to bytes with sensible (to me) defaults.  It could likely replace the implementation of networkString (not verified) and provides more customization when needed.

There is one certain issue, the apidocs are broken due to an unbalanced {, as I've noted below.

I would like to understand, why treat None specially and not raise a TypeError?  Is there a motivating recommendation or use-case?  Assuming there is, shouldn't the docstrings also reference None when they reference acceptable types (e.g. the TypeError doc).

src/twisted/python/compat.py
```diff
@@ -822,6 +822,32 @@ def _bytesRepr(bytestring):
...
+    @param errors: The error handling scheme to pass to L{str.encode).
+
+    @raise TypeError: The input is not L{unicode} or L{bytes}.
+
+    @return: The encoded string.  If I{s} is L{bytes} or L{None}, just return
+             I{s}.
+    @rtype: L{bytes}
+    """
+    if s is None:
+        return s
```
exarkun, line 841: I thought it was supposed to raise `TypeError`.

src/twisted/test/test_compat.py
```diff
@@ -921,3 +923,46 @@ def test_bytesReprPrefix(self):
         ``b`` to the returned repr on both Python 2 and 3.
         """
         self.assertEqual(_bytesRepr(b'\x00'), "b'\\x00'")
+
+
+
+class ToBytesTests(unittest.TestCase):
+    """
+    L{toBytes}
```
exarkun: What about it?

src/twisted/python/compat.py
```diff
@@ -822,6 +822,32 @@ def _bytesRepr(bytestring):
...
+    elif isinstance(s, bytes):
+        return s
+    elif isinstance(s, unicode):
+        return s.encode(encoding=encoding, errors=errors)
+    else:
+        raise TypeError("Expected {} to be unicode or bytes, not: {}".format(
+            s, type(s)))
```
exarkun: There's lots of funny formatting like this throughout the diff.

src/twisted/newsfragments/9343.feature
```diff
@@ -0,0 +1 @@
+twisted.python.compat.toBytes is a convenience function for converting unicode strings to bytes.
```
exarkun, line 1: Putting it in `twisted.python.compat` suggests this is only to do with bridging the Python 2/3 divide.  And so it will eventually go away because some behavior it offers is native to or natural in Python 3 ... but I don't think this is actually the case.  What Python 3-only construct idiomatically replaces `toBytes`?

src/twisted/python/compat.py
```diff
@@ -868,5 +894,6 @@ def _bytesRepr(bytestring):
     "intern",
     "unichr",
     "raw_input",
-    "_tokenize"
+    "_tokenize",
+    "toBytes"
```
exarkun, line 898: trailing commas are nice

src/twisted/test/test_compat.py
```diff
@@ -921,3 +923,46 @@ def test_bytesReprPrefix(self):
...
+    """
+
+    def test_toBytesUnicode(self):
+        """
+        If L{unicode} is passed to L{toBytes}, the L{unicode} is converted
+        to L{bytes} and returned.
+        """
```
exarkun, line 938: This looks like about 4 tests mashed into one.

exarkun (changes requested):
Thanks.  It would be nice to have a clearer demonstration of the utility of such a method.  Perhaps you can highlight a few places in Twisted which would be simplified by the use of it.

src/twisted/newsfragments/9343.feature
rodrigc: This does not need to live in `twisted.python.compat`, since this is a generally useful utility function for converting unicode to bytes, and should not be tied to `twisted.python.compat` which is a "temporary" bridging for Python 2/3.

Where is a better place for it `twisted.python.util`?

rodrigc: `toBytes()` can be used all over the place.  There are places in `twisted.web` where it could be used, and would help eliminate Python 3-porting problems I have seen in 3rd party code which use twisted.web.  For example, here is one failure I found in txkube due to use of `networkString`:

https://travis-ci.org/LeastAuthority/txkube/jobs/304997908#L945

exarkun: `networkString` raising that exception may indicate there's a _bug_ in txkube though.  Silently converting between unicode and bytes when the programmer didn't _intend_ to it is one of the worst features of Python 2.  `networkString` fails in this case on purpose.  If `networkString` were replaced with an API like `toBytes` then it would be harder to find this kind of programming error (because the exception would not necessarily be raised).  So I'd recommend being very careful where this conversion function is used: it could easily take us back to the bad-old-times of carelessly mixed bytes/unicode.

src/twisted/python/compat.py
rodrigc: I don't understand your feedback.

rodrigc:
@exarkun based on your feedback, I created a new PR where I put the `toBytes()` function
somewhere else instead of in `twisted.python.compat`

src/twisted/python/compat.py
adiroiban: I would say that we should enforce trailing commas as a code style :) to reduce future diffs

PR #937, opened 2017-11-29
Fix a few bytes/unicode issues in conch script

rodrigc:
https://twistedmatrix.com/trac/ticket/9344

Fix a few bytes/unicode issues in the conch script, and fix a few twistedchecker style issues.

Fix one case where if you do:

```
conch localhost "nonexistentcommand"
```

instead of getting this:

```
Unhandled Error
                                                   File "/Users/crodrigues/twisted4/src/twisted/python/context.py", line 122, in callWithContext
                                                             File "/Users/crodrigues/twisted4/src/twisted/python/context.py", line 85, in callWithContext
                                                                                                                                                             return func(*args,**kw)
                         File "/Users/crodrigues/twisted4/src/twisted/conch/ssh/service.py", line 45, in packetReceived
                                                                                                                           return f(packet)
                                                                                                                                             File "/Users/crodrigues/twisted4/src/twisted/conch/ssh/connection.py", line 273, in ssh_CHANNEL_EXTENDED_DATA
                                                                                                 log.callWithLogger(channel, channel.extReceived, typeCode, data)
                                     File "/Users/crodrigues/twisted4/src/twisted/python/log.py", line 103, in callWithLogger
                               File "/Users/crodrigues/twisted4/src/twisted/python/log.py", line 86, in callWithContext
                             File "/Users/crodrigues/twisted4/src/twisted/python/context.py", line 122, in callWithContext
  File "/Users/crodrigues/twisted4/src/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
  File "/Users/crodrigues/twisted4/src/twisted/conch/scripts/conch.py", line 468, in extReceived
    sys.stderr.write(data)
builtins.TypeError: write() argument must be str, not bytes
```

You get something like:

```
zsh:1: command not found: nonexistentcommand
```


rodrigc:
@wiml can you try this patch?

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -816,35 +816,45 @@ def setAttrs(self, attrs):
...
+
+    def __next__(self):
         if self.filesCache:
             return self.filesCache.pop(0)
         d = self.read()
         d.addCallback(self._cbReadDir)
         d.addErrback(self._ebReadDir)
         return d
 
+    next = __next__
```
wiml: `_ebReadDir` modifies the `next` attribute at the end of the listing; it will have to either modify `__next__` as well or one of these should call the other

wiml:
I added a comment to the diff.

While you're looking at this, could you take a look at fixing [#7322](https://twistedmatrix.com/trac/ticket/7322) as well? That's the other bug I currently have to work around in order to use the directory iterator, and it seems like it should also have a nearly trivial fix.

rodrigc:
Please look at the updated patch, where I deal with __next__ and next.

Do you have a small testcase that we can use to trigger  [#7322](https://twistedmatrix.com/trac/ticket/7322)?

wiml:
Here's a test case you can add to `test_filetransfer.py`/`OurServerOurClientTests` to demonstrate both [#7322](https://twistedmatrix.com/trac/ticket/7322) and [#9349](https://twistedmatrix.com/trac/ticket/9349):

```python
    @defer.inlineCallbacks
    def testOpenDirectoryIterator(self):
        """
        Check that the object returned by
        L{filetransfer.FileTransferClient.openDirectory} can be used
        as an iterator.
        """

        # This function is a little more complicated than it would be
        # normally, since we need to call _emptyBuffers() after
        # creating any SSH-related Deferreds, but before waiting on
        # them via yield.
        
        self._emptyBuffers()
        openDir = yield d
        
        filenames = set()
        try:
            for f in openDir:
                self._emptyBuffers()
                (filename, _, fileattrs) = yield f
                filenames.add(filename)
        finally:
            d = openDir.close()
            self._emptyBuffers()
            yield d
            
        self._emptyBuffers()
        
        self.assertEqual(filenames,
                         set([b'.testHiddenFile', b'testDirectory',
                              b'testRemoveFile', b'testRenameFile',
                              b'testfile1']))

```

rodrigc:
@wiml can you try this updated patch, which addresses [#7322](https://twistedmatrix.com/trac/ticket/7322)? also?

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -856,12 +856,8 @@ def _cbReadDir(self, names):
 
 
     def _ebReadDir(self, reason):
-        reason.trap(EOFError)
-        def _():
-            raise StopIteration
-        self.__next__ = _
-        self.next = self.__next__
```
wiml: This change removes the monkeypatching of `next`/`__next__` (which I think is good, stylistically) but it changes the behavior of the iterator. It no longer remembers that it has reached EOF, and calling `next` again will cause a protocol violation (issuing `READDIR` on a closed handle). Python iterators [are required](https://docs.python.org/3/library/stdtypes.html#iterator.__next__) to allow calling `next` after StopIteration (and should continue raising StopIteration).

wiml:
I think the last change (ba968f180) breaks the iterator protocol. I have an alternate suggestion, I'll post it as soon as I have a chance to test it.

wiml:
Try this fix instead: wiml/twisted@ad321dcc0

rodrigc:
@wiml I merged your change, and it looks good.  Is this OK to merge?

wiml:
I might not be using github's viewer correctly, but it doesn't look like this PR contains wiml/twisted@ad321dcc0 or fixes the problem that it is supposed to fix — the [iterator protocol requires](https://docs.python.org/3/library/stdtypes.html#iterator-types) that you can continue calling `next` on a finished iterator and keep safely getting a `StopIteration`.

rodrigc:
@wiml can you take another look?  I've merged your patch.

wiml:
@rodrigc  - That looks good to me and works in my use case.

wiml (approved):

PR #939, opened 2017-11-30
9333 default observer fix

jlitzingerdev:
This is a fix for 9333: https://twistedmatrix.com/trac/ticket/9333

Some items of note for reviewers:

 1.  The new formatEventWithTracebacks API was deliberately kept private to avoid adding new public APIs to maintain. Additionally, its easy to make it public if so specified in review.

2.  The existing formatEventAsClassicLogText has similar code that can be replaced. Rather than assume that was desired, I kept the change as focused as possible. Easy to change if desired.

As noted in the ticket, there are cases where duplicate tracebacks result, specifically if observers are appended to the globalLogPublisher, rather than overwrite with a beginLoggingTo call on the globalLogBeginner. This is documented in the logger docs.


hawkowl:
Pushed to the buildbots.

hawkowl:
The OSX build failures are spurious. They pass in circlecli.

src/twisted/logger/_format.py
```diff
@@ -287,3 +287,33 @@ def formatWithCall(formatString, mapping):
...
+    """
+    eventText = formatEvent(event)
+    try:
+        traceback = event["log_failure"].getTraceback()
+        if isinstance(traceback, bytes):
+            traceback = traceback.decode('utf-8')
+    except KeyError:
+        return eventText
+    except:
+        traceback = u"(UNABLE TO OBTAIN TRACEBACK FROM EVENT)\n"
```
markrwilliams: This should be lower case and contain the exception message, like [its counterpart in the legacy logging system](https://github.com/twisted/twisted/blob/104394ddb76775c57c18897160a5ad055617b51a/src/twisted/python/log.py#L447).

src/twisted/logger/_format.py
```diff
@@ -287,3 +287,33 @@ def formatWithCall(formatString, mapping):
...
+    @rtype: L{unicode}
+    """
+    eventText = formatEvent(event)
+    try:
+        traceback = event["log_failure"].getTraceback()
+        if isinstance(traceback, bytes):
+            traceback = traceback.decode('utf-8')
+    except KeyError:
+        return eventText
+    except:
```
markrwilliams: This should be `except Exception:` - we don't want to catch `SystemExit`!

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+    """
+    A captured error for use in format tests.
+    """
+
+
+
+class FormatEventWithTracebackTests(unittest.TestCase):
+    """
+    Tests for L{formatEventWithTraceback}, all of which ensure that the
+    returned type is a UTF-8.
```
markrwilliams: This is an incomplete sentence.  I think it should be, "... all of which ensure that the returned type is UTF-8 decoded text."

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        An event with a C{log_failure} key will have a traceback appended.
+        """
+        f = Failure(CapturedError("This is a fake error"))
+        event = {
+            "log_format": u"This is a test log message"
+        }
+        event["log_failure"] = f
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('Traceback', eventText)
```
markrwilliams: A better test would be:

```python
try:
    raise CapturedError("This is a fake error")
except CapturedError:
    f = Failure()
...
self.assertIn(f.getTraceback().decode('utf-8'), eventText)
...
```

That's because it would exercise the code with a real traceback.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        the accompanying failure.
+        """
+        f = Failure(CapturedError("This is a fake error"))
+        event = {
+            "log_format": u""
+        }
+        event["log_failure"] = f
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('Traceback', eventText)
```
markrwilliams: As above.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+
+    def test_formatUnformattableWithTraceback(self):
+        """
+        An event with an unformattable value in the C{log_format} key still
+        has a traceback appended.
+        """
+        event = {
+            "log_format": "{evil()}",
+            "evil": lambda: 1 / 0,
+            Unformattable(): "gurk",
```
markrwilliams: This is extraneous to the stated purpose of the test, which is served by `{evil()}`.  Please remove it.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        """
+        f = Failure(CapturedError("This is a fake error"))
+        event = {
+            "log_format": u"This is a test log message"
+        }
+        event["log_failure"] = f
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('Traceback', eventText)
+        self.assertIn('This is a fake error', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        self.assertIn('This is a test log message', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        """
+        f = Failure(CapturedError("This is a fake error"))
+        event = {
+            "log_format": u""
+        }
+        event["log_failure"] = f
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('Traceback', eventText)
+        self.assertIn('This is a fake error', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        event = {
+            "log_format": "{evil()}",
+            "evil": lambda: 1 / 0,
+            Unformattable(): "gurk",
+        }
+        f = Failure(CapturedError("This is a fake error"))
+        event["log_failure"] = f
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('MESSAGE LOST', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        self.assertIn('Traceback', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        self.assertIn('This is a fake error', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+        If a traceback cannot be appended, a message indicating this is true
+        is appended.
+        """
+        event = {
+            "log_format": u""
+        }
+        event["log_failure"] = object()
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn("(UNABLE TO OBTAIN TRACEBACK FROM EVENT)", eventText)
```
markrwilliams: This should be a unicode literal - also see above comment about the message in this case.

src/twisted/logger/test/test_format.py
```diff
@@ -429,3 +430,94 @@ class Unformattable(object):
...
+    def test_formatEventNonCritical(self):
+        """
+        An event with no C{log_failure} key will not have a traceback appended.
+        """
+        event = {
+            "log_format": u"This is a test log message"
+        }
+        eventText = formatEventWithTraceback(event)
+        self.assertIsInstance(eventText, unicode)
+        self.assertIn('This is a test log message', eventText)
```
markrwilliams: This should be a unicode literal.

src/twisted/logger/test/test_global.py
```diff
@@ -356,3 +357,16 @@ def test_warningsModule(self):
...
+
+    def test_failuresAppendTracebacks(self):
+        """
+        The string resulting from a logged failure contains a traceback.
+        """
+        f = Failure(Exception("this is not the behavior you are looking for"))
+        log = Logger(observer=self.publisher)
+        log.failure('a failure', failure=f)
+        msg = self.errorStream.getvalue()
+        self.assertIn('a failure', msg)
```
markrwilliams, line 370: This (and what follows) should _not_ be a unicode literal :)

That's because [FileLogObserver encodes unicode messages to UTF-8 when it's given a binary file](https://github.com/twisted/twisted/blob/104394ddb76775c57c18897160a5ad055617b51a/src/twisted/logger/_file.py#L33-L37), and [it is on Python 2](https://github.com/jlitzingerdev/twisted/blob/a0a1c7c2bd7b7a2f396ea6ddb0453c131ccec4fc/src/twisted/logger/test/test_global.py#L65).

markrwilliams (approved):
This is an important improvement and a well-done PR.  Thank you so much!

Please address the review comments before resubmitting for review.

markrwilliams (changes requested):
This review is to block merging until comments are addressed.

src/twisted/logger/_format.py
jlitzingerdev: No problem.  I modeled formatEventWithTraceback after formatEventAsClassicLogText (above it), which also uses all caps and an bare except.  Is that something that should be fixed (in a different ticket/PR)?

markrwilliams: Yes, I think `formatEventAsClassicLogText` should get changed via a new ticket.  I appreciate that you started with a goal of consistency, as well!

src/twisted/logger/test/test_format.py
jlitzingerdev: Makes sense.  What I submit will be slightly different as eventText is always unicode and getTraceback() is the native string type (the decode call fails on P3).

src/twisted/logger/test/test_format.py
jlitzingerdev: @markrwilliams I agree, with concerns.  It tests a valid "unformattable" path, but also neglects one.  The first path would be caught by your suggested change, eliminating the `Unformattable()`.  The second path, exercised by the `Unformattable()` key occurs when the __repr__ call for the event throws.  My question -- do we only need to exercise the former, or should I split this into two tests, one for each?  For the sake of completeness, that seems like the right thing to do, thoughts?

src/twisted/logger/_format.py
jlitzingerdev: @markrwilliams As with the bare except, I modeled this after the existing similar example in `formatEventAsClassicLogText`, rather than the legacy log system.  Should `formatEventAsClassicLogText` be changed to match the legacy (and concatenate the error) in a separate ticket?

src/twisted/logger/test/test_global.py
jlitzingerdev: @markrwilliams Am I correct in interpreting that this code should remain as-is, to exercise the different paths for a native string on Python2/Python3?  Or, did you want me to ensure the encoding was forced on both?

markrwilliams: Sorry for being unclear!  This was more of a note to myself confirming that this code is correct to use a native string.

src/twisted/logger/_format.py
```diff
@@ -287,3 +287,33 @@ def formatWithCall(formatString, mapping):
...
+    @param event: A logging event.
+    @type event: L{dict}
+
+    @return: A formatted string with a traceback if applicable.
+    @rtype: L{unicode}
+    """
+    eventText = formatEvent(event)
+    try:
+        traceback = event["log_failure"].getTraceback()
+        if isinstance(traceback, bytes):
```
jlitzingerdev, line 311: I have concerns with this code.  Failure's getTraceback() method will return a native string depending on the Python version, so str/bytes Python 2.  The assumption that the bytes can sanely decoded as utf-8 is not one I've been able to verify and is probably wrong.

Looking at `formatEventAsClassicLogText`, which I used as a guide, it doesn't do any conversion but seems (I may be misreading) to rely on the fact that bytes and unicode can be joined in Python2 (https://github.com/twisted/twisted/blob/2e7b403c03e9b9b891a5eb36297e494f541d5d6a/src/twisted/logger/_format.py#L192).  That may (haven't confirmed) expose it to the same issue as 9442 due to the ascii conversion.  Should I confirm that I'll open a ticket.

So while this might be fine, I'd like to do a bit more diligence before this is merged, thus, I'm removing it from review.

jlitzingerdev: Related to the concern with `formatEventAsClassicLogText`:
https://twistedmatrix.com/trac/ticket/9456#ticket

jlitzingerdev:
Closing, no longer relevant.

PR #940, opened 2017-12-01
Zsh completions now provided for twist command

teratorn:
http://twistedmatrix.com/trac/ticket/9338


adiroiban:
Travis and Appveyor are green. Codecov reporting is broken.
I will merge this as it is.

Many thanks @teratorn for your contribution. ...I don't know when this will be released ... but at least it should be in trunk :)

PR #941, opened 2017-12-13
doc: explain xmlrpc.Proxy's queryFactory

ktdreyer:
Give an example of how to debug raw XML-RPC responses using xmlrpc.Proxy.

https://twistedmatrix.com/trac/ticket/9350


adiroiban (changes requested):
Thanks for working I think.

This does not look bad :)

The API comment is nice, but a narrative example would be best :)

I like the example that you sent to the Trac ticket and I think that it would be very useful to have it as part of documentation examples from docs/web/examples/

http://twistedmatrix.com/documents/current/web/examples/index.html

Try to have the code (with full docstring and comment) in docs/web/examples/ and consider where this can be 

To make the example easy to discover please consider whether you can introduce it with a few sentences (or a dedicated section) here http://twistedmatrix.com/documents/current/web/howto/xmlrpc.html

------

Note that you need to add a release notes fragment.
In this way, your change will be advertised in the next public release.

------

Let me know if something is not clear and I would be happy to help (release fragment, writing narrative documentation)

ktdreyer:
@adiroiban thank you for the review. I think I've addressed your comments now.

wiml:
@adiroiban this looks good to me and I think your earlier comments have been addressed. OK to land (once CI is happy)?

adiroiban:
This looks good, but with this example, we might want to make the QueryFactory class public.

The private stuff is there so that Twisted library users don't use them.
If we document something, it means that we expect other peoples to use it .. and then it make sense to have it private.

adiroiban:
thanks @wiml  for the follow and @ktdreyer for pushing your suggestion

adiroiban:
So... instead of documenting the usage of a private code... even for debubging, this PR should make it public and document it.

If we make it private, we might change it in the future without any notice, and people using it might get broken code without any warning.

ktdreyer:
I'm using `_QueryFactory` in another project https://github.com/ktdreyer/txkoji to implement a custom `Marshaller` that supports long ints (`<i8></i8>`), so it would be great to formally make this a public API.

ktdreyer:
This PR has evolved several times, so I'll just open a new one: https://github.com/twisted/twisted/pull/1091

PR #942, opened 2017-12-14
Add toBytes() helper method

rodrigc:
https://twistedmatrix.com/trac/ticket/9343

We can replace:

```
if isinstance(foo, unicode):
    foo = foo.encode("ascii")
```

with:

```
foo = toBytes(foo)
```


rodrigc:
Moving to https://github.com/twisted/twisted/pull/961

PR #943, opened 2017-12-18
Remove support for Python 3.3

rodrigc:
https://twistedmatrix.com/trac/ticket/9352

Python 3.3 has been marked as EOL on September 29, 2017:

​https://www.python.org/dev/peps/pep-0398/#x-end-of-life

I would suggest that Twisted drop support for Python 3.3.

Twisted would now support:

Twisted 2.7, and Twisted 3.4-3.6+

According to this spreadsheet: https://docs.google.com/spreadsheets/d/1JSX8fBmPb84emTmV0Kmyf0_r6R0kZM0h9Wdm91tn7Kg/edit#gid=0

Python 3.3 came with OpenSUSE 13.1.
Python 3.4 is in Debian 8, Ubuntu 14.04, Fedora 21


src/twisted/__init__.py
```diff
@@ -7,16 +7,7 @@
...
-    # Don't allow the user to run a version of Python we don't support.
-    import sys
-
-    version = getattr(sys, "version_info", (0,))
-    if version < (2, 7):
-        raise ImportError("Twisted requires Python 2.7 or later.")
-    elif version >= (3, 0) and version < (3, 3):
-        raise ImportError("Twisted on Python 3 requires Python 3.3 or later.")
-
+from twisted._checkrequirements import _checkRequirements
```
adiroiban: can we move this to _setup.py and not call each time we import twisted? ... and then we don't need the extra exec code in setup.py.

NEWS.rst
```diff
@@ -6,6 +6,9 @@ http://twistedmatrix.com/trac/ticket/<number>
 Twisted 17.9.0 (2017-09-23)
 ===========================
 
+This is the last Twisted release where Python 3.3 is supported, on any
```
adiroiban, line 9: This looks like we are re-writing history here.
I think that we should not touch previous release notes.

So we should just update the release notes for the next (future) release and drop support after that...
but I am not familiar with the process of dropping a python version.

src/twisted/_checkrequirements.py
```diff
@@ -0,0 +1,14 @@
+# -*- test-case-name: twisted -*-
+
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+def _checkRequirements():
+    # Don't allow the user to run a version of Python we don't support.
```
adiroiban: I know that this is a copy/paste, but maybe take the oportunity to clean this code a bit and for example convert this comment into a docstring.

And why the local import?

adiroiban (commented):
Changes looks good, but I don't like the fact that we are changing the release notes for older releases.

So this is an abstention

NEWS.rst
rodrigc: When Python 2.6 support was dropped, the NEWS.rst file was modified after the fact to update this: http://twistedmatrix.com/trac/ticket/8651

Since that seemed to be OK, I did it here.

Dropping support for a Python release happens so infrequently, that I thought it was OK.

src/twisted/__init__.py
rodrigc: OK, I moved the check into _setup.py

src/twisted/_checkrequirements.py
rodrigc: OK, I made it a docstring

src/twisted/python/_setup.py
```diff
@@ -198,11 +197,27 @@ def __init__(self, *args, **kwargs):
 
 
 
+def _checkPythonVersion():
+    """
+    Fail if we detect a version of Python we don't support.
+    """
+    import sys
```
adiroiban: I think that this import can be removed

adiroiban (approved):
Only minor comments.

Since I am the only one commenting on the release notes I think that is OK to do it this way and hurry up the removal.

I think that this was under review for long enough.
Based on "no news is good news" I guess that nobody is against the change and we should merge it.

Many thanks for working on this.

PR #948, opened 2017-12-28
9353 archer0001 iocpinvalidsends

archer0001:
https://twistedmatrix.com/trac/ticket/9353


src/twisted/internet/iocpreactor/abstract.py
```diff
@@ -188,48 +188,49 @@ def _resumeWriting(self):
...
 
     def _cbWrite(self, rc, numBytesWritten, evt):
-        if self._handleWrite(rc, numBytesWritten, evt):
+        if self._handleWrite(rc, numBytesWritten, evt) and \
+        self._writeScheduled is None:
             self.doWrite()
 
 
-    def _handleWrite(self, rc, numBytesWritten, evt):
+    def _handleWrite(self, rc, numBytesWritten, evt, updateOffset=False):
```
adiroiban, line 196: Please add docstring for the new updateOffset argument, even if this is a private method.

src/twisted/internet/iocpreactor/abstract.py
```diff
@@ -188,48 +188,49 @@ def _resumeWriting(self):
 
 
     def _cbWrite(self, rc, numBytesWritten, evt):
-        if self._handleWrite(rc, numBytesWritten, evt):
+        if self._handleWrite(rc, numBytesWritten, evt) and \
+        self._writeScheduled is None:
```
adiroiban, line 192: Please use parentheses for breaking long lines.
Also, either encapsulate the logic in a variable or add a comment to this conditional.

Ex
```
is_crash = self._someCheck() and self._otherThing
if is_crash:
    self.doDomething()
```
or

```
if self._someCheck() and self._otherThing:
    # Bungulu has crashed, so we clean before propagating the error.
    self.doDomething()
```

adiroiban (changes requested):
Happy to see some fixed for heavy load usage :)

I don't understand the changes here.

Please consider adding more docstrings.

In the same time, the tests are huge without any details about the "patching strategy"

See the failing tests from https://tm.tl/#9358 maybe it can be used as a functional/system tests so that we can try this change without mocking.

All Appveyor build are failing.
Can you please update your code to have Appveyour tests green and then request a new review.

Thanks again :)

src/twisted/internet/iocpreactor/abstract.py
archer0001: I can certainly document this better.  I don't understand the parenthesis  suggestion.  Are you suggesting breaking down the one if statement to two?

adiroiban: Breaking long lines

```
if self._handleWrite(rc, numBytesWritten, evt) and \
        self._writeScheduled is None:
    CODE_BLOCK_HERE
```

vs

```
if (
    self._handleWrite(rc, numBytesWritten, evt) and
    self._writeScheduled is None
        ):
    CODE_BLOCK_HERE
```

archer0001: I can make that change!  Adding a comment to explain the logic as well.

glyph:
As commented in trac, I believe another fix for this has been landed / released.

However, you probably know more about IOCP than I do, so if I'm wrong about that please reopen!

Thanks for this contribution to Twisted; sorry it took so long to fix!

PR #955, opened 2018-01-27
9374  Fixing circular references in t.p.policies and t.p.tls

IlyaSkriblovsky:
https://twistedmatrix.com/trac/ticket/9374

ProtocolWrapper creates the circular reference between itself and its wrappedProtocol. Also TLSMemoryBIOProtocol creates the circular reference between itself and its _tlsConnection.

Both circles are never broken and still alive after connections are closed until garbage collector collects them.

This seems like a performance issue especially for apps handling many short-lived TLS connections. For default GC thresholds this will cause GC to be called very often wasting CPU time. For loosened GC thresholds (for the sake of performance) this leads to eating RAM very quickly.

Discussed in mailing list: ​https://twistedmatrix.com/pipermail/twisted-python/2018-January/031798.html

### DIFF
--- a/src/twisted/newsfragments/9374.bugfix
+++ b/src/twisted/newsfragments/9374.bugfix
@@ -1,1 +1,1 @@
-t.p.p.ProtocolWrapper and t.p.t.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed
+twisted.protocols.policies.ProtocolWrapper and twisted.protocols.tls.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed.
--- a/src/twisted/newsfragments/9374.bugfix
+++ b/src/twisted/newsfragments/9374.bugfix
@@ -1,1 +1,1 @@
-twisted.protocols.policies.ProtocolWrapper and twisted.protocols.tls.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed.
+twisted.protocols.policies.ProtocolWrapper and twisted.protocols.tls.TLSMemoryBIOProtocol no longer create circular references that keep protocol instances in memory after connection is closed.
--- a/src/twisted/newsfragments/9374.bugfix
+++ b/src/twisted/newsfragments/9374.bugfix
@@ -1,0 +1,1 @@
+t.p.p.ProtocolWrapper and t.p.t.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed
--- a/src/twisted/protocols/policies.py
+++ b/src/twisted/protocols/policies.py
@@ -124,7 +124,7 @@ class ProtocolWrapper(Protocol):
         self.factory.unregisterProtocol(self)
         self.wrappedProtocol.connectionLost(reason)
 
-        # breaking reference cycle between self and wrappedProtocol
+        # Breaking reference cycle between self and wrappedProtocol
         self.wrappedProtocol = None
 
 
--- a/src/twisted/protocols/policies.py
+++ b/src/twisted/protocols/policies.py
@@ -124,6 +124,9 @@ class ProtocolWrapper(Protocol):
         self.factory.unregisterProtocol(self)
         self.wrappedProtocol.connectionLost(reason)
 
+        # breaking reference cycle between self and wrappedProtocol
+        self.wrappedProtocol = None
+
 
 
 class WrappingFactory(ClientFactory):
--- a/src/twisted/protocols/policies.py
+++ b/src/twisted/protocols/policies.py
@@ -124,7 +124,7 @@ class ProtocolWrapper(Protocol):
         self.factory.unregisterProtocol(self)
         self.wrappedProtocol.connectionLost(reason)
 
-        # Breaking reference cycle between self and wrappedProtocol
+        # Breaking reference cycle between self and wrappedProtocol.
         self.wrappedProtocol = None
 
 
--- a/src/twisted/protocols/policies.py
+++ b/src/twisted/protocols/policies.py
@@ -128,7 +128,6 @@ class ProtocolWrapper(Protocol):
         self.wrappedProtocol = None
 
 
-
 class WrappingFactory(ClientFactory):
     """
     Wraps a factory and its protocols, and keeps track of them.
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -1003,7 +1003,7 @@ class TLSMemoryBIOTests(TestCase):
         gc.disable()
 
         try:
-            class ConnectionCloserProtocol(Protocol):
+            class CloserProtocol(Protocol):
                 def dataReceived(self, data):
                     self.transport.loseConnection()
 
@@ -1013,24 +1013,24 @@ class TLSMemoryBIOTests(TestCase):
                     self.transport.write(b'hello')
 
             origTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
-            origServerProtos = nObjectsOfType(ConnectionCloserProtocol)
+            origServerProtos = nObjectsOfType(CloserProtocol)
 
             authCert, serverCert = certificatesForAuthorityAndServer()
             serverFactory = TLSMemoryBIOFactory(
                 serverCert.options(), False,
-                Factory.forProtocol(ConnectionCloserProtocol)
+                Factory.forProtocol(CloserProtocol)
             )
             clientFactory = TLSMemoryBIOFactory(
                 optionsForClientTLS(u'example.com', trustRoot=authCert), True,
                 Factory.forProtocol(GreeterProtocol)
             )
             loopbackAsync(
-                TLSMemoryBIOProtocol(serverFactory, ConnectionCloserProtocol()),
+                TLSMemoryBIOProtocol(serverFactory, CloserProtocol()),
                 TLSMemoryBIOProtocol(clientFactory, GreeterProtocol())
             )
 
             newTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
-            newServerProtos = nObjectsOfType(ConnectionCloserProtocol)
+            newServerProtos = nObjectsOfType(CloserProtocol)
             self.assertEqual(newTLSProtos, origTLSProtos)
             self.assertEqual(newServerProtos, origServerProtos)
         finally:
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -927,6 +927,33 @@ class TLSMemoryBIOTests(TestCase):
         return disconnectDeferred
 
 
+    from twisted.internet import defer
+    @defer.inlineCallbacks
+    def test_loseConnectionAfterConnectionLost(self):
+        """
+        If TLSMemoryBIOProtocol.loseConnection is called after connectionLost,
+        it does nothing.
+        """
+        tlsClient, tlsServer, handshakeDeferred, disconnectDeferred = (
+            self.handshakeProtocols())
+
+        # Make sure connectionLost calls _shutdownTLS, but loseConnection
+        # doesnt call it for the second time.
+        calls = []
+        def _shutdownTLS(shutdown=tlsClient._shutdownTLS):
+            calls.append(1)
+            return shutdown()
+        tlsServer._shutdownTLS = _shutdownTLS
+        tlsServer.write(b'x')
+        tlsClient.loseConnection()
+        # wait until tlsServer.connectionLost is called
+        yield disconnectDeferred
+        self.assertEqual(calls, [1])
+
+        tlsServer.loseConnection()
+        self.assertEqual(calls, [1])
+
+
     def test_unexpectedEOF(self):
         """
         Unexpected disconnects get converted to ConnectionLost errors.
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -985,6 +985,9 @@ class TLSMemoryBIOTests(TestCase):
         TLSMemoryBIOProtocol doesn't leave circular references that make
         it alive after connection is closed.
         """
+        def nObjectsOfType(type):
+            return sum(1 for x in gc.get_objects() if isinstance(x, type))
+
         gc.disable()
 
         try:
@@ -1000,6 +1003,9 @@ class TLSMemoryBIOTests(TestCase):
                 def connectionLost(self, reason):
                     reason.trap(ConnectionDone)
 
+            origTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
+            origServerProtos = nObjectsOfType(DummyServerProtocol)
+
             authCert, serverCert = certificatesForAuthorityAndServer()
             serverFactory = TLSMemoryBIOFactory(
                 serverCert.options(), False,
@@ -1014,11 +1020,8 @@ class TLSMemoryBIOTests(TestCase):
                 TLSMemoryBIOProtocol(clientFactory, DummyClientProtocol())
             )
 
-            def nObjectsOfType(type):
-                return sum(1 for x in gc.get_objects() if isinstance(x, type))
-
-            self.assertEqual(nObjectsOfType(TLSMemoryBIOProtocol), 0)
-            self.assertEqual(nObjectsOfType(DummyServerProtocol), 0)
+            self.assertEqual(nObjectsOfType(TLSMemoryBIOProtocol), origTLSProtos)
+            self.assertEqual(nObjectsOfType(DummyServerProtocol), origServerProtos)
         finally:
             gc.enable()
 
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -7,6 +7,8 @@ Tests for L{twisted.protocols.tls}.
 
 from __future__ import division, absolute_import
 
+import gc
+
 from zope.interface.verify import verifyObject
 from zope.interface import Interface, directlyProvides, implementer
 
@@ -42,7 +44,7 @@ from twisted.internet.interfaces import (
 
 from twisted.internet.error import ConnectionDone, ConnectionLost
 from twisted.internet.defer import Deferred, gatherResults
-from twisted.internet.protocol import Protocol, ClientFactory, ServerFactory
+from twisted.internet.protocol import Protocol, ClientFactory, ServerFactory, Factory
 from twisted.internet.task import TaskStopped
 from twisted.protocols.loopback import loopbackAsync, collapsingPumpPolicy
 from twisted.trial.unittest import TestCase, SynchronousTestCase
@@ -307,8 +309,9 @@ class DeterministicTLSMemoryBIOTests(SynchronousTestCase):
         """
         client, server, pump = handshakingClientAndServer(b"untrustworthy",
                                                           True)
+        wrappedServerProtocol = server.wrappedProtocol
         pump.flush()
-        self.assertEqual(server.wrappedProtocol.received, [])
+        self.assertEqual(wrappedServerProtocol.received, [])
 
 
 
@@ -975,6 +978,49 @@ class TLSMemoryBIOTests(TestCase):
             self.assertTrue(reason[0].check(Error), reason[0])
         disconnectDeferred.addCallback(disconnected)
         return disconnectDeferred
+
+
+    def test_noCircularReferences(self):
+        """
+        TLSMemoryBIOProtocol doesn't leave circular references that make
+        it alive after connection is closed.
+        """
+        gc.disable()
+
+        try:
+            class DummyServerProtocol(Protocol):
+                def dataReceived(self, data):
+                    self.transport.loseConnection()
+
+
+            class DummyClientProtocol(Protocol):
+                def connectionMade(self):
+                    self.transport.write(b'hello')
+
+                def connectionLost(self, reason):
+                    reason.trap(ConnectionDone)
+
+            authCert, serverCert = certificatesForAuthorityAndServer()
+            serverFactory = TLSMemoryBIOFactory(
+                serverCert.options(), False,
+                Factory.forProtocol(DummyServerProtocol)
+            )
+            clientFactory = TLSMemoryBIOFactory(
+                optionsForClientTLS(u'example.com', trustRoot=authCert), True,
+                Factory.forProtocol(DummyClientProtocol)
+            )
+            loopbackAsync(
+                TLSMemoryBIOProtocol(serverFactory, DummyServerProtocol()),
+                TLSMemoryBIOProtocol(clientFactory, DummyClientProtocol())
+            )
+
+            def nObjectsOfType(type):
+                return sum(1 for x in gc.get_objects() if isinstance(x, type))
+
+            self.assertEqual(nObjectsOfType(TLSMemoryBIOProtocol), 0)
+            self.assertEqual(nObjectsOfType(DummyServerProtocol), 0)
+        finally:
+            gc.enable()
 
 
 
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -44,7 +44,8 @@ from twisted.internet.interfaces import (
 
 from twisted.internet.error import ConnectionDone, ConnectionLost
 from twisted.internet.defer import Deferred, gatherResults
-from twisted.internet.protocol import Protocol, ClientFactory, ServerFactory, Factory
+from twisted.internet.protocol import (Protocol, ClientFactory, ServerFactory,
+                                       Factory)
 from twisted.internet.task import TaskStopped
 from twisted.protocols.loopback import loopbackAsync, collapsingPumpPolicy
 from twisted.trial.unittest import TestCase, SynchronousTestCase
@@ -1020,8 +1021,10 @@ class TLSMemoryBIOTests(TestCase):
                 TLSMemoryBIOProtocol(clientFactory, DummyClientProtocol())
             )
 
-            self.assertEqual(nObjectsOfType(TLSMemoryBIOProtocol), origTLSProtos)
-            self.assertEqual(nObjectsOfType(DummyServerProtocol), origServerProtos)
+            newTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
+            newServerProtos = nObjectsOfType(DummyServerProtocol)
+            self.assertEqual(newTLSProtos, origTLSProtos)
+            self.assertEqual(newServerProtos, origServerProtos)
         finally:
             gc.enable()
 
--- a/src/twisted/protocols/test/test_tls.py
+++ b/src/twisted/protocols/test/test_tls.py
@@ -1000,41 +1000,39 @@ class TLSMemoryBIOTests(TestCase):
             """
             return sum(1 for x in gc.get_objects() if isinstance(x, type))
 
+        self.addCleanup(gc.enable)
         gc.disable()
 
-        try:
-            class CloserProtocol(Protocol):
-                def dataReceived(self, data):
-                    self.transport.loseConnection()
+        class CloserProtocol(Protocol):
+            def dataReceived(self, data):
+                self.transport.loseConnection()
 
 
-            class GreeterProtocol(Protocol):
-                def connectionMade(self):
-                    self.transport.write(b'hello')
+        class GreeterProtocol(Protocol):
+            def connectionMade(self):
+                self.transport.write(b'hello')
 
-            origTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
-            origServerProtos = nObjectsOfType(CloserProtocol)
+        origTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
+        origServerProtos = nObjectsOfType(CloserProtocol)
 
-            authCert, serverCert = certificatesForAuthorityAndServer()
-            serverFactory = TLSMemoryBIOFactory(
-                serverCert.options(), False,
-                Factory.forProtocol(CloserProtocol)
-            )
-            clientFactory = TLSMemoryBIOFactory(
-                optionsForClientTLS(u'example.com', trustRoot=authCert), True,
-                Factory.forProtocol(GreeterProtocol)
-            )
-            loopbackAsync(
-                TLSMemoryBIOProtocol(serverFactory, CloserProtocol()),
-                TLSMemoryBIOProtocol(clientFactory, GreeterProtocol())
-            )
+        authCert, serverCert = certificatesForAuthorityAndServer()
+        serverFactory = TLSMemoryBIOFactory(
+            serverCert.options(), False,
+            Factory.forProtocol(CloserProtocol)
+        )
+        clientFactory = TLSMemoryBIOFactory(
+            optionsForClientTLS(u'example.com', trustRoot=authCert), True,
+            Factory.forProtocol(GreeterProtocol)
+        )
+        loopbackAsync(
+            TLSMemoryBIOProtocol(serverFactory, CloserProtocol()),
+            TLSMemoryBIOProtocol(clientFactory, GreeterProtocol())
+        )
 
-            newTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
-            newServerProtos = nObjectsOfType(CloserProtocol)
-            self.assertEqual(newTLSProtos, origTLSProtos)
-            self.assertEqual(newServerProtos, origServerProtos)
-        finally:
-            gc.enable()
+        newTLSProtos = nObjectsOfType(TLSMemoryBIOProtocol)
+        newServerProtos = nObjectsOfType(CloserProtocol)
+        self.assertEqual(newTLSProtos, origTLSProtos)
+        self.assertEqual(newServerProtos, origServerProtos)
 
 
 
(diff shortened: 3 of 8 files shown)


exarkun:
> This seems like a performance issue especially for apps handling many short-lived TLS connections. For default GC thresholds this will cause GC to be called very often wasting CPU time.

It doesn't matter how many cycles you actually have in your process: (on CPython) the garbage collector runs with the same frequency regardless, and does the same amount of work.  The _purpose_ of the garbage collector is to _find_ the cycles among all the objects that exist in your process.  It has slightly more work to do actually collecting the cycles that it finds but this is basically negligible.

>  For loosened GC thresholds (for the sake of performance) this leads to eating RAM very quickly.

Since GC CPU considerations are orthogonal, this is basically orthogonal too (that is, you may choose to  tweak your GC settings but doing so has little or nothing to do with these cycles).

The difference that will be made is that some memory may be releasable to the OS slightly earlier if the cycles are broken earlier.  In practice, I suspect it would be hard to observe this effect since memory is almost never released back to the OS anyway and over-commit means that unused memory is practically as good as released memory anyway.

IlyaSkriblovsky:
Sorry, but I'm not agree that GC in CPython has fixed run frequency. `gc.set_threshold` manual says:

> In order to decide when to run, the collector keeps track of the number object allocations and deallocations since the last collection. When the number of allocations minus the number of deallocations exceeds threshold0, collection starts.

So we can affect collection frequency by:

1. **Setting higher GC thresholds.** That's what my team does for cpu-intensive processes. 
2. **Carefully breaking unused reference cycles.** By doing so, `number of allocations minus the number of deallocations` can be almost stable or growing very slowly and GC will be started rarely. But every single leaked reference cycle in a tight loop or in a frequently-repeated task (such as receiving tls connection) makes this value to grow rapidly. If this happen with loosened GC thresholds (see 1), and if reference cycle includes the protocol instance with heavy business-level data attached to it, the app's RAM usage can how quickly.

For example, I have the server that receives tons of short-lived TLS connections with loosened GC thresholds. Before this fix it took around 1.2Gb of RAM after 1 day after being started. With this fix it takes around 200Mb and not growing.

exarkun:
> Sorry, but I'm not agree that GC in CPython has fixed run frequency. gc.set_threshold manual says:

So it does.  I had forgotten about that trait.  Thanks for the doc reference and sorry about the incorrect post.

> For example, I have the server that receives tons of short-lived TLS connections with loosened GC thresholds. Before this fix it took around 1.2Gb of RAM after 1 day after being started. With this fix it takes around 200Mb and not growing.

It would be great if you could contribute this as a repeatable benchmark.

IlyaSkriblovsky:
Here is the benchmark script: https://gist.github.com/IlyaSkriblovsky/4dd3abfd5f67c64b13f1c673f56466f9

(the script is Py3-only because I don't know direct way to get collection count in Py2)

The script simulates 10k (set by `N` constant) TLS connections to connect-and-drop server. If `loose` cmdline arg it specified then it uses loosened GC thresholds. It prints out the number of times garbage collection was run and a peak RAM usage (using `psutil`).

Results on my machine:

GC thresholds  | Wihout fix | With fix
---------------------|---------------|--------------
default (700, 10, 10) | 500 colls, 50mb | 150 colls, 45mb
(100000, 200, 100) | 3 colls, 288mb | 1 coll, 89mb

So, with default GC settings there is small effect on RAM usage, but several times lower GC run count. With high GC thresholds there is significant difference in RAM usage.

rodrigc:
@IlyaSkriblovsky please follow all the steps at http://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch .  You've done most of the steps, but are missing step 11.

IlyaSkriblovsky:
Did it, thanks

rodrigc:
Looks like some of the earlier Windows failures with this PR were due to a broken version of pypiwin32: https://github.com/mhammond/pywin32/issues/1151 .  I triggered another build now that the broken version has been removed from pypiwin32

src/twisted/test/test_sslverify.py
```diff
@@ -267,20 +268,28 @@ def dataReceived(self, data):
         def connectionLost(self, reason):
             self.lostReason = reason
 
+    clientWrappedProto = ListeningClient()
+    serverWrappedProto = GreetingServer()
+
+    cF = protocol.Factory()
```
adiroiban: I would prefer not to use abbreviations and call this clientFactory. ... why not clientFact ? :)

...same for clientWrappedProto ... why not cWP ?

------

I see that the existing code uses abbreviations, but I don't think that this is a good reason to continue to write new code in this way :)

src/twisted/protocols/tls.py
```diff
@@ -196,7 +198,9 @@ def makeConnection(self, transport):
         Connect this wrapper to the given transport and initialize the
         necessary L{OpenSSL.SSL.Connection} with a memory BIO.
         """
-        self._tlsConnection = self.factory._createConnection(self)
+        # Using weakref.proxy to avoid circular references
+        selfProxy = weakref.proxy(self)
```
adiroiban: just asking. Is using a weakref the only way to solve this?

Why not break it in `connectionLost` like done in the other place?

-------

Using weakref can make the tests harder to read / write as you will need to take extra care on the scope of an instance..

...and I would like to write Python without having to care about the lifecycle of the objects.

src/twisted/newsfragments/9374.bugfix
```diff
@@ -0,0 +1 @@
+t.p.p.ProtocolWrapper and t.p.t.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed
```
adiroiban: please use full names and end the sentence with a full stop :)

src/twisted/protocols/policies.py
```diff
@@ -124,6 +124,9 @@ def connectionLost(self, reason):
         self.factory.unregisterProtocol(self)
         self.wrappedProtocol.connectionLost(reason)
 
+        # Breaking reference cycle between self and wrappedProtocol
```
adiroiban: Use full for a sentence.
In this way, it is clear that the sentence has an end and that it was not left unfinished by accident.

src/twisted/protocols/test/test_tls.py
```diff
@@ -42,7 +44,8 @@
 
 from twisted.internet.error import ConnectionDone, ConnectionLost
 from twisted.internet.defer import Deferred, gatherResults
-from twisted.internet.protocol import Protocol, ClientFactory, ServerFactory
+from twisted.internet.protocol import (Protocol, ClientFactory, ServerFactory,
```
adiroiban: this is fine, but to reduce future dif size, I would prefer to have this list alphabetically ordered and each import on a line.


```
from twisted.internet.protocol import (
    ClientServerFactory,
    Factory,
    Protocol,
    ServerFactory,
    )
```

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +981,54 @@ def disconnected(ign):
         return disconnectDeferred
 
 
+    def test_noCircularReferences(self):
+        """
+        TLSMemoryBIOProtocol doesn't leave circular references that make
+        it alive after connection is closed.
```
adiroiban: `make it alive` What does it mean?

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +981,54 @@ def disconnected(ign):
...
+        """
+        def nObjectsOfType(type):
```
adiroiban: even if this is test code please consider adding a docstring to describe the purpose of this function.

There are no tests for the test code so in order to check whether the test code itself is correct we can only read the code.

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +981,54 @@ def disconnected(ign):
...
+            return sum(1 for x in gc.get_objects() if isinstance(x, type))
+
+        gc.disable()
+
+        try:
+            class DummyServerProtocol(Protocol):
```
adiroiban: Why is  this protocol "dummy" ? What means dummy?

This looks like a fully working protocol which is closing the connecting first time it receives data.

I have used these types of protocol in production code :)

Please consider adding some docstring to document this code.

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +981,54 @@ def disconnected(ign):
...
+                def dataReceived(self, data):
+                    self.transport.loseConnection()
+
+
+            class DummyClientProtocol(Protocol):
```
adiroiban: Again, why dummy?
This looks like a fully functional protocol which will ignore all received data and will send a `hello` message on a connection.

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +981,54 @@ def disconnected(ign):
...
+                def connectionMade(self):
+                    self.transport.write(b'hello')
+
+                def connectionLost(self, reason):
+                    reason.trap(ConnectionDone)
```
adiroiban: Just asking.  I have not executed (yet) the tests. Can the connection be closed for other reasons?
Please document why ConnectionDone is important here.

adiroiban (changes requested):
Thanks for working on this.

Happy to see performance improvements :)

Is there a test for new the code in ProtocolWrapper ?

Can this be implemented without weakref?

-------

I would like to see fewer (to none) abbreviations in the code and more docstrings, especially for the test part.

At 8AM and in the context of this PR the code is easy to ready. Hacking at 11PM and getting into some strange bugs, might prove that the code is not that easy to read.
Especially the dummy protocols and the Failure.trap in connectionLost

IlyaSkriblovsky:
Thanks for review!
I've fixed some wording and naming issues.

I would be happy to explicitly break `TLSMemoryBIOProtocol._tlsConnection` reference cycle without weakrefs. But `twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback` scenario causes `TLSMemoryBIOProtocol.dataReceived` to be called _after_ `TLSMemoryBIOProtocol.connectionLost`, causing null reference access if `_tlsConnection` is deleted in `connectionLost`. I've tried to dig into this to check whether this is a real-life case or a strange consequence of synthetic testing using `IOPump`. But the issue seems to be requiring deeper understanding of Twisted's TLS internals than I currently have :(

`test_surpriseFromInfoCallback` is the only test that failing if `_tlsConnection` is explicitly deleted in `connectionLost`.

src/twisted/newsfragments/9374.bugfix
```diff
@@ -0,0 +1 @@
+twisted.protocols.policies.ProtocolWrapper and twisted.protocols.tls.TLSMemoryBIOProtocol no longer create circular references that still alive after connection is closed.
```
adiroiban: `references that still alive after` :) please also update this part

IlyaSkriblovsky: done!

adiroiban:
> But twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback scenario causes TLSMemoryBIOProtocol.dataReceived to be called after TLSMemoryBIOProtocol.connectionLost,

AFIK `dataReceived` should not be called after `connectionLost`. 

@exarkun Can you please confirm if there should be any case in which dataReceived  is called after connectionLost ?

I see this info in the API

```
class IProtocol(Interface):

    def dataReceived(data):
        """
        Called whenever data is received.

        Use this method to translate to a higher-level message.  Usually, some
        callback will be made upon the receipt of each complete protocol
        message.

        Please keep in mind that you will probably need to buffer some data
        as partial (or multiple) protocol messages may be received!  We
        recommend that unit tests for protocols call through to this method
        with differing chunk sizes, down to one byte at a time.

        @param data: bytes of indeterminate length
        @type data: L{bytes}
        """

    def connectionLost(reason):
        """
        Called when the connection is shut down.

        Clear any circular references here, and any external references
        to this Protocol.  The connection has been closed. The C{reason}
        Failure wraps a L{twisted.internet.error.ConnectionDone} or
        L{twisted.internet.error.ConnectionLost} instance (or a subclass
        of one of those).

        @type reason: L{twisted.python.failure.Failure}
        """
```

I can see that the API explicitly talks about cleaning circular references here, but does not make any statement about when dataReceived can be called... but since circularReferences are broken in connectionLost, I expect that this should be the last call for a certain instance.

The fact that `test_surpriseFromInfoCallback` fails might reveal a bug in the existing code.

I will give it a try and see why dataReceived is called after connectionLost in this test.

Thanks!

IlyaSkriblovsky:
`test_surpriseFromInfoCallback` creates the server and the client on the top of `twisted.test.iosim.FakeTransport` transports. Then an exception is simulated during TLS handshake and one side calls his `TLSMemoryBIOProtocol.failVerification()` → `abortConnection()` → `connectionLost()`. But then another side gets its `TLSMemoryBIOProtocol._flushSendBIO()` called writing something to its transport. But `FakeTransport` stills receives data even when it is closed, so the written data successfully triggers `dataReceived` on the first side.

May be `FakeTransport` should check whether it is closed before accepting data in its `write()` method? If I put `if self.disconnected: return` as the first line of `iosim.FakeTransport.write()`, then all test seem to be passing with explicit deleting of `_tlsConnection` in `connectionLost()`. But I'm not sure that this is safe and reasonable thing to do.

Sorry if my explanations are confusing or inaccurate.

adiroiban:
For the TLS/SSL connections there is one corner case which might get into the way of this patch... and this might be an existing bug... see https://github.com/twisted/twisted/blob/trunk/src/twisted/protocols/tls.py#L346

When you connect to a TLS/SSL socket and the connection is closed right away without sending any data, the connection is not actually closed.

---------

I have checked the code and I see that connectionLost is called twice ...and before dataReceived, so I guess that there is a bug in iosim :)

IlyaSkriblovsky:
> connectionLost is called twice

It is probably called once for the server and once for the client. But then one of the sides calls `_flushSendBIO` and writes something to its transport (`FakeTransport`) and `IOPump.pump()` sends it into other side's `dataReceived()`.

But I'm not sure where is the source of the bug: whether `TLSMemoryBIOProtocol` should not write to its transport after its `connectionLost()` called, or `FakeTransport` should not accept more `write()`s after it is marked as disconnected.

adiroiban:
> It is probably called once for the server and once for the client. But then one of the sides calls _flushSendBIO and writes something to its transport (FakeTransport) and IOPump.pump() sends it into other side's dataReceived().

True. I am stupid and confused :)

Let me dig more :)

------------

Also, if I run a single tests, I see 14 errors reported by trial.

```
$ ./build/bin/trial twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback

SNIP

exceptions.ZeroDivisionError: division by zero

twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
twisted.test.test_sslverify.ServiceIdentityTests.test_surpriseFromInfoCallback
Ran 1 tests in 191.540s

FAILED (errors=14)
```

IlyaSkriblovsky:
> Also, if I run a single tests, I see 14 errors reported by trial.

That's because `test_surpriseFromInfoCallback` tests bogous function passed as info_callback to pyOpenSSL. And this callback seems to be called many times during TLS handshake. So the single test causes 14 deferred failures.

I like the idea of adding check for `self.disconnected` to `iosim.FakeTransport.write()`. This seems to solve the issue with nulled `TLSMemoryBIOProtocol._tlsConnection` in `connectionLost()` without breaking any other tests. But it would be cool if someone more experienced in Twisted internals can confirm that this is safe.

exarkun:
> I like the idea of adding check for self.disconnected to iosim.FakeTransport.write(). This seems to solve the issue with nulled TLSMemoryBIOProtocol._tlsConnection in connectionLost() without breaking any other tests. But it would be cool if someone more experienced in Twisted internals can confirm that this is safe.

All real `ITransport` implementations allow `transport.loseConnection` followed by `transport.write` or `protocol.connectionLost` followed by `transport.write`.  In both these cases, they allow it and silently drop the `write` data on the floor.  Since when a _real_ connection has been lost the data _can't_ be delivered, the data is not delivered in these cases.

It would be a good thing to make any test double for `ITransport` duplicate this behavior (on the premise that test doubles which do not replicate the actual behavior of the thing for which they are a double lead to false confidence in the test suite since they invalidate the tests they are used in by only telling you what the code under test would do if you used it against something _different_ from the real implementation).  Therefore, `FakeTransport.write` should indeed be changed to drop the data on the floor if called after the fake connection has been lost.

Three bonus points for also adding tests which demonstrate that `FakeTransport` and other `ITransport` implementations _both_ behave this way (iow, expanded the surface of `ITransport` / `FakeTransport` which is "verified" by the test suite).

adiroiban:
@IlyaSkriblovsky I see that you are member of Twisted team.
In this case please create branches directly in the main repo.
In this way,  buildbot tests are automatically triggered and we can see the results from OSX.

Also, please consider writing tests for any code that you write, including tests for the testing infrastructure, as suggested by Jean Paul... and with the docstring to document that calling write after connection lost is acceptable :)

Thanks!

IlyaSkriblovsky:
@adiroiban I'm member of Twisted team because I'm was contibuting to TxMongo, but I haven't ever contributed to Twisted itself. Okay, I will create branches in the main repo next time.

I will try to write tests for `FakeTransport` and other `ITransport` implementers.

adiroiban:
> I'm member of Twisted team because I'm was contributing to TxMongo

That is fine, you are good to create branches in Twisted :).
AFIK trunk is protected by GitHub, so you should not worry.

The main reason why we don't allow anybody to commit to the main repo, is that commits will trigger builds and we don't want to end up with PR which are executing malicious code on our buildslaves.

IlyaSkriblovsky:
I've added a test for `FakeTransport.write()` dropping data after `loseConnection()`.

> Three bonus points for also adding tests which demonstrate that FakeTransport and other ITransport implementations both behave this way

Should I also find other fake `ITransport` implementations in testing code and add same tests (`write` after `loseConnection`) for them too? I can try, but it seems a bit outside of the scope of this particular PR.

adiroiban:
> Should I also find other fake ITransport implementations in testing code and add same tests (write after loseConnection) for them too? I can try, but it seems a bit outside of the scope of this particular PR.

I think that this can be done outside, so that we can merge this as it is.
A future ticket might also update the docstring for IProtocol to be explicit that is possible to have dataReceived after connectionLost

Thanks!

IlyaSkriblovsky:
I suspect that AppVeyor build failures are not related to the code changed by me. Is there any known issues with AppVeyor testing? Or am I wrong and I should re-check my changes?

adiroiban:
Hm. They might be due to this PR. They are SSH failures... I guess that due to protocol wrapper.

AFAIk trunk is green ... this is the latest build for trunk https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-4597

adiroiban:
@IlyaSkriblovsky you can push your branch to the Twisted main repo and it will trigger the buildbot tests.
Among them there are a few Windows builders... if they are green, then this is an Appveyor issue... but I think that this is due to the changes from here.

Does those tests pass on your local machine?

IlyaSkriblovsky:
Seems like I haven't right to push to twisted/twisted
image

I live on linux, but I will try to find Windows to test on

IlyaSkriblovsky:
I've tried to run tests on a Windows machine and get the same failure in `twisted.conch.test.test_conch.OpenSSHClientForwardingTests` test on trunk...

```
(venv) ilya@WINSERVER C:\Users\ilya\twisted
$ git status
On branch trunk
Your branch is up to date with 'origin/trunk'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        src/twisted/internet/iocpreactor/iocpsupport.cp36-win_amd64.pyd
        src/twisted/test/raiser.cp36-win_amd64.pyd

nothing added to commit but untracked files present (use "git add" to track)

(venv) ilya@WINSERVER C:\Users\ilya\twisted
$ trial twisted.conch.test.test_conch.OpenSSHClientForwardingTests.test_localToRemoteForwarding
twisted.conch.test.test_conch
  OpenSSHClientForwardingTests
    test_localToRemoteForwarding ...                                     [FAIL]
                                   [ERROR]
                                   [ERROR]

[FAIL]
  File "c:\users\ilya\twisted\src\twisted\trial\_synctest.py", line 432, in assertEqual
  File "c:\python36\lib\unittest\case.py", line 829, in assertEqual
  File "c:\python36\lib\unittest\case.py", line 822, in _baseAssertEqual
twisted.trial.unittest.FailTest: None != b'test\n'

twisted.conch.test.test_conch.OpenSSHClientForwardingTests.test_localToRemoteForwarding
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
DelayedCalls: (set twisted.internet.base.DelayedCall.debug = True to debug)
<DelayedCall 0x5086550 [29.890799283981323s] called=0 cancelled=0 _BaseBaseClient.failIfNotConnected(TimeoutError('',))>

twisted.conch.test.test_conch.OpenSSHClientForwardingTests.test_localToRemoteForwarding
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
Selectables:
<<class 'twisted.internet.tcp.Client'> to ('127.0.0.1', 52717) at 50864a8>

twisted.conch.test.test_conch.OpenSSHClientForwardingTests.test_localToRemoteForwarding
Ran 1 tests in 0.312s

FAILED (failures=1, errors=2)
```

And I also added AppVeyor build on trunk in my fork and it failed the same way:
https://ci.appveyor.com/project/IlyaSkriblovsky/twisted/build/job/x46o39p54jkhgpyw

IlyaSkriblovsky:
Okay, I figured out that `twisted.conch.test.test_conch.OpenSSHClientForwardingTests` were failing due to `twisted.conch.test.keydata.privateRSA_openssh` key was 1024bit RSA, but 1024bit RSA keys are not supported anymore in OpenSSH 7.6: https://www.openssh.com/releasenotes.html

>  * Refuse RSA keys <1024 bits in length and improve reporting for keys
>   that do not meet this requirement.

I've replaced them with 2048bit and these tests are now passing on AppVeyor.

I have checked the logs of last succeded trunk build (https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-4597) and it seems that `choco install openssh` installs the same 7.6.0.1 version as it installs for me today. So I have no idea why it worked then and doesn't work anymore neither for this branch, nor for trunk :confused:

But replacing test RSA keys with 2048 made many other conch-related test failing, mostly because hardcoded fingerprints. If you are ok with replacing RSA keys with 2048bit ones, I will look into all these places and update accordingly (but it is probably the matter of distinct PR).

adiroiban:
@IlyaSkriblovsky thanks for looking at the SSH issue. If this is not related and is already failing in trunk please create a ticket. If you don't have time to work on this one, please assign it to me. Is important to have trunk green.

But the whole test environment should be set up with pinned dependencies, so that if we want to run the test for a branch from 1 year ago, it will still pass... so I don't know why the change.

We can discuss on the separate ticket.


----

> Seems like I haven't right to push to twisted/twisted

Right. I added you to @twisted/twisted-contributors ... I have misinterpreted the `member` badge.
Please let me know if you still can't push.

IlyaSkriblovsky:
> But the whole test environment should be set up with pinned dependencies, so that if we want to run the test for a branch from 1 year ago, it will still pass... so I don't know why the change.

appveyor.yml just installs the latest openssh: https://github.com/twisted/twisted/blob/trunk/appveyor.yml#L62

```yaml
install:
  # Install OpenSSH through chocolatey
  - "choco install openssh -confirm"
  - "refreshenv"
```

But I've compared build logs and it seems that both this [succeded build](https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-4597) and [this failing one](https://ci.appveyor.com/project/IlyaSkriblovsky/twisted/build/trunk-7) are using the same openssh version...  But windows builds are certaily failing on trunk both on my local machine and AppVeyor. And other PRs are hit by this too: https://github.com/twisted/twisted/pull/966

I've created #968 to fix RSA keys issue

IlyaSkriblovsky:
two build slaves of buildbot failed with asyncio reactor:
https://buildbot.twistedmatrix.com/builders/fedora24-py3.5/builds/1239
https://buildbot.twistedmatrix.com/builders/fedora25-py3.5/builds/1235

The reason is that `twisted.test.test_ssl.BufferingTests` do following in its `tearDown`:
```python
    def tearDown(self):
        # . . .
        if self.clientProto.transport is not None:
            self.clientProto.transport.loseConnection()
```

But `self.clientProto.transport` (which is `TLSMemoryBIOProtocol`) is already closed at that moment, so it's `loseConnection()` is called twice. On epoll reactor this leads to this order of method calls for `TLSMemoryBIOProtocol`:
```
loseConnection
loseConnection
connectionLost
connectionLost
```
and it is ok. But on asyncio reactor the order is following:
```
loseConnection
connectionLost
loseConnection
connectionLost
```

`connectionLost` breaks reference cycle by clearing `_tlsConnection`, so the second `loseConnection` does null dereference.

What is the general requirements for `loseConnection()`? Is it acceptable to call it twice? If no, how the client code is supposed to check if a transport is already closed?

exarkun:
> What is the general requirements for loseConnection()? Is it acceptable to call it twice? If no, how the client code is supposed to check if a transport is already closed?

I would consider application code that calls it more than once sloppy but it is not disallowed to do so.  When this happens, `connectionLost` should nevertheless only be called once.  So it sounds like both epoll and asyncio reactors have a bug related to this?

> connectionLost breaks reference cycle by clearing _tlsConnection, so the second loseConnection does null dereference.

Not really important yet I don't want to let it go. :)  You can't do a null dereference in Python.  You can attempt to access an attribute of an object which doesn't exist.  This is a _lot_ nicer than a null dereference.

src/twisted/protocols/test/test_tls.py
```diff
@@ -977,6 +985,58 @@ def disconnected(ign):
...
+            """
+            Return the number of instances of a given type in memory.
+
+            @param type: Type whose instances to find.
+
+            @return: The number of instances found.
+            """
+            return sum(1 for x in gc.get_objects() if isinstance(x, type))
+
+        gc.disable()
```
exarkun: fwiw you can get rid of the whole `try/finally` with `self.addCleanup(gc.enable)` right here.

IlyaSkriblovsky: Thanks, I forgot about it

IlyaSkriblovsky:
> When this happens, connectionLost should nevertheless only be called once. So it sounds like both epoll and asyncio reactors have a bug related to this?

Sorry, I was wrong. Two `loseConnection()` calls is simply one for a server and another for a client :man_facepalming:
So the test failure is directly related to this PR: after we clear `_tlsConnection` in `connectionLost`, subsequent call to `loseConnection` will raise on attempting to access it. I will dig into it.

> You can't do a null dereference in Python. You can attempt to access an attribute of an object which doesn't exist. This is a lot nicer than a null dereference.

You are right, sorry my C-ism. But to be even more presice, `None` is an object too! So it is attempt to access an attribute of _existing_ object, but not the one _you meant to_ :)

exarkun:
> > You can't do a null dereference in Python. You can attempt to access an attribute of an object which doesn't exist. This is a lot nicer than a null dereference.

> You are right, sorry my C-ism. But to be even more presice, None is an object too! So it is attempt to access an attribute of existing object, but not the one you meant to :)

Ah yes, sorry, that's what I meant but my statement was ambiguous (I meant the attribute doesn't exist on the object, not that the object doesn't exist :smile: ).  Thanks for the clarification.

glyph:
Should this be back in review? It seems like all the blockers are dealt with but there’s a lot of discussion to review here :)

IlyaSkriblovsky:
Yes, please review it again.

I've added trivial check if TLSMemoryBIOProtocol already received
`connectionLost()` in its `loseConnection()`. I hope it is reasonable check
to do.

glyph:
Sorry, I wasn't asking because I was about to review it :)

The place Twisted reviewers (such as yourself, congrats on your addition to the team :)) should look to see what needs reviewing is here:

https://twisted.reviews

Unfortunately this is what Trac is still required for, tracking this state.  Please stick the 'review' keyword back on the Trac ticket so that it will show up there.

IlyaSkriblovsky:
Sorry my silliness. Still getting familiar with the process. I've added "review" keyword and reassigned the ticket to nobody.

glyph:
> Sorry my silliness. Still getting familiar with the process. I've added "review" keyword and reassigned the ticket to nobody.

No worries. The process is definitely a bit clunky. One day somebody will have enough free time to migrate the rest of it to github and it won't require manually pushing so many buttons and twirling so many dials :)

src/twisted/protocols/tls.py
```diff
@@ -402,12 +402,15 @@ def connectionLost(self, reason):
...
+        # Breaking reference cycle between self._tlsConnection and self.
+        self._tlsConnection = None
+
 
     def loseConnection(self):
         """
         Send a TLS close alert and close the underlying connection.
         """
-        if self.disconnecting:
+        if self.disconnecting or not self.connected:
```
adiroiban, line 413: is there a test for this change? :)

src/twisted/protocols/policies.py
```diff
@@ -124,6 +124,9 @@ def connectionLost(self, reason):
         self.factory.unregisterProtocol(self)
         self.wrappedProtocol.connectionLost(reason)
 
+        # Breaking reference cycle between self and wrappedProtocol.
+        self.wrappedProtocol = None
```
adiroiban, line 128: do we have tests for this change?

Besides updating the existing tests for policies, I was expecting a new test (or update of an existing test) for connectionLost in which of which the docstring describe that on connection lost the wrappedProtocol is None

adiroiban (changes requested):
Thanks a lot. Happy to see a green build :)
Looks good... but I think that it is missing a few tests.

Other than that, it should be ready for merging soon :)

adiroiban:
@IlyaSkriblovsky before pushing changes to this, please merge with trunk.
This will get the Circle-CI configuration.

Also, before pushing a code, try to run pyflakes and pycodestyle on local code to reduce the load on our testing servers.... Travis and Appveyor are free... but in the end, somebody has to pay for their usage

IlyaSkriblovsky:
@adiroiban you are right, sorry. Will do as you said. I've cancelled Travis build, but I can't cancel AppVeyor.

IlyaSkriblovsky:
@adiroiban test failure looks like unrelated to this PR because all buildbot builds was green until today changes, but today I only added some tests without changing implementation. Is it possible to restart this particular build to check if failure is permanent?

adiroiban:
@IlyaSkriblovsky I have restarted it.

You can also restart it... check for credentials here  https://twistedmatrix.com/trac/wiki/ContinuousIntegration/DeveloperWorkflow#WriteAccess :)

adiroiban:
but that is a flaky test... I saw it failing in other branches

IlyaSkriblovsky:
Thank you. I'm going to put this back in review.

adiroiban (approved):
Looks good.

My only minor comment is about the short `cF`, `sF`, `serverWrappedProto`,  `cProto` ... ex why `cProto`  and not `cP` ... or why `cF` and not `cFactory` or `cFact` ...

I would like to see non-abreviated names... but I will not block the merge on that :)

Feel free to merge :)

Thanks a lot for the great work!

IlyaSkriblovsky:
I've renamed `cF` and `sF` to `clientFactory` and `serverFactory`. I've just used `cF` and `sF` because saw these names in some another place with analogous code. But you are right, this is not justifying reason :)

So, waiting for testing results after updating to trunk and merging.

IlyaSkriblovsky:
I'm not sure why this build failed: https://buildbot.twistedmatrix.com/builders/osx10.10-py2.7/builds/1350
It can't install tox!? But this build successfully installed it: https://buildbot.twistedmatrix.com/builders/osx10.10-py2.7-coverage/builds/1340

What is going on?
This is certainly not in the scope of this PR, but I'm a bit afraid of breaking the trunk.

adiroiban:
@IlyaSkriblovsky feel free to merge it. This is a known issue (see the conversation from the mailing list)

It was green before...but while testing the OSX built, I picked your branch for testing. Sorry :)

The OSX is executed over circle-ci so we are covered.

Thanks!

IlyaSkriblovsky:
@adiroiban Done, thanks!

PR #957, opened 2018-02-13
[9377] Don't remove trailing spaces from SSH version string comment.

adiroiban:
Scope
====

See https://twistedmatrix.com/trac/ticket/9377 for a description of the bug.

Conch should keep the SSH version string as provided by the peer, even if it has trailing spaces.

Changes
=======

Update the code to only remove LF character

As a drive by, I have updated the comments to inform that Twisted also support simple Unix new line separate and  wrote a test to explicitly cover this behavior

How to test / What to check
======================

For backward compatibility, I left the non-RFC behaviour in place and just documented.
Let me know if you think that Twisted conch should be more strict and reject peers which only use unix new line... this will be done in a separate ticket as we will need to update some tests.

I wrote unit tests and I think that they are good enough.

To manual test this in the wild you need a Bitvise server which advertise the version with spaces

See https://github.com/mscdex/ssh2/issues/488#issuecomment-329085866 for more details.

### DIFF
--- a/src/twisted/conch/ssh/transport.py
+++ b/src/twisted/conch/ssh/transport.py
@@ -725,11 +725,18 @@ class SSHTransportBase(protocol.Protocol):
         if not self.gotVersion:
             if self.buf.find(b'\n', self.buf.find(b'SSH-')) == -1:
                 return
+
+            # RFC 4253 section 4.2 ask for strict `\r\n` line ending.
+            # Here we are a bit more relaxed and accept implementations ending
+            # only in '\n'.
+            # https://tools.ietf.org/html/rfc4253#section-4.2
             lines = self.buf.split(b'\n')
             for p in lines:
                 if p.startswith(b'SSH-'):
                     self.gotVersion = True
-                    self.otherVersionString = p.strip()
+                    # Since the line was split on '\n' and most of the time
+                    # it uses '\r\n' we may get an extra '\r'.
+                    self.otherVersionString = p.rstrip('\r')
                     remoteVersion = p.split(b'-')[1]
                     if remoteVersion not in self.supportedVersions:
                         self._unsupportedVersionReceived(remoteVersion)
--- a/src/twisted/conch/ssh/transport.py
+++ b/src/twisted/conch/ssh/transport.py
@@ -737,7 +737,7 @@ class SSHTransportBase(protocol.Protocol):
                     self.gotVersion = True
                     # Since the line was split on '\n' and most of the time
                     # it uses '\r\n' we may get an extra '\r'.
-                    self.otherVersionString = p.rstrip('\r')
+                    self.otherVersionString = p.rstrip(b'\r')
                     remoteVersion = p.split(b'-')[1]
                     if remoteVersion not in self.supportedVersions:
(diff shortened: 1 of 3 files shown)


src/twisted/conch/ssh/transport.py
```diff
@@ -726,11 +726,18 @@ def dataReceived(self, data):
         if not self.gotVersion:
             if self.buf.find(b'\n', self.buf.find(b'SSH-')) == -1:
                 return
+
+            # RFC 4253 section 4.2 ask for strict `\r\n` line ending.
+            # Here we are a bit more relaxed and accept implementations ending
+            # only in '\n'.
+            # https://tools.ietf.org/html/rfc4253#section-4.2
```
markrwilliams, line 733: I'm always dubious of RFC violations but it's apparently pretty popular to end lines with just '\n'. [openssh even addresses this](https://github.com/openssh/openssh-portable/blob/b5974de1a1d419e316ffb6524b1b277dda2f3b49/ssh_api.c#L333-L341).

I appreciate you taking the time to comment the existing behavior!

src/twisted/conch/ssh/transport.py
```diff
@@ -726,11 +726,18 @@ def dataReceived(self, data):
...
             lines = self.buf.split(b'\n')
             for p in lines:
                 if p.startswith(b'SSH-'):
                     self.gotVersion = True
-                    self.otherVersionString = p.strip()
+                    # Since the line was split on '\n' and most of the time
+                    # it uses '\r\n' we may get an extra '\r'.
+                    self.otherVersionString = p.rstrip('\r')
```
markrwilliams: Should this remove the last character if it happens to be `\r`?  A single instance of that is the only thing that have been part of the line delimiter.

markrwilliams (approved):
I'd like to see something stricter than stripping all trailing `\r`, but the tests and comments make this worthwhile as is.

src/twisted/conch/ssh/transport.py
adiroiban: I don't understand the question.

If we have a line line, `SSH-bla-bla\r\rn' ... I think that it should just left a '\r' ... but I don't know if anybody is using that in production... and implementing that would require more work and tests.. so why bother? :)

src/twisted/conch/ssh/transport.py
```diff
@@ -726,11 +726,18 @@ def dataReceived(self, data):
...
+            # only in '\n'.
+            # https://tools.ietf.org/html/rfc4253#section-4.2
             lines = self.buf.split(b'\n')
             for p in lines:
                 if p.startswith(b'SSH-'):
                     self.gotVersion = True
-                    self.otherVersionString = p.strip()
+                    # Since the line was split on '\n' and most of the time
+                    # it uses '\r\n' we may get an extra '\r'.
+                    self.otherVersionString = p.rstrip(b'\r')
```
markrwilliams, line 740: Sorry I wasn't clear; my point is that given:
`SSH banner\r\r\r`
this produces
`SSH banner`
whereas OpenSSH would parse it as:
`SSH banner\r\r`.

Maximum compatibility would require removing only the last `\r`.

PR #960, opened 2018-02-20
twisted.trial._dist's AMP Commands use Unicode on Python 3.

markrwilliams:
They continue to use String on Python 2 so that twisted.trial._dist
can continue to work in terms of native strings.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. Create a new one at https://twistedmatrix.com/trac/ticket/9379


markrwilliams:
If this is acceptable, it will need a different news fragment because of https://github.com/twisted/twisted/blob/trunk/src/twisted/newsfragments/9264.bugfix

It should go in after #959

adiroiban (commented):
While the diff looks great, I agree with @exarkun  (https://twistedmatrix.com/trac/ticket/9379#comment:3) that protocol should have the same behavior, regardless of python version.

But, it this code used outside of Twisted? I see this code is in a private _dist package.

I have never used this functionality, so I guess is up to the people using it to decide the tradeoff between clean code vs protocol consistency

adiroiban (dismissed):
Based on IRC dicussion
```
(07:50:36 PM) runciter: adiroiban: it's not used outside of twisted, no
(07:51:01 PM) runciter: adiroiban: however i believe the concern in the ticket is that even if it's used only inside twisted, as a wire protocol it should allow different versions of python to communicate  
(07:51:05 PM) adiroiban: so as long as is private we can hack it as we like :)
(07:51:25 PM) runciter: adiroiban: that was my initial perspective as well - as i said in the ticket i saw it as merely incidental to disttrial
(07:51:26 PM) adiroiban: but is that a real world case?
(07:51:35 PM) runciter: adiroiban: it isn't yet, but it's hard to take back a change to a wire protocol
(07:51:45 PM) runciter: especially one that isn't versioned, like this one
(07:52:20 PM) runciter: and if i'm being honest i'd love it if i could stop using tox with twisted projects
(07:52:42 PM) runciter: so i don't want to preclude it becoming a real world case
(07:53:26 PM) runciter: as usual, though, file system paths are hard
(07:53:58 PM) adiroiban: so I would say -1 then and close this
(07:54:40 PM) runciter: adiroiban: ok, fair enough
```

src/twisted/trial/_dist/test/test_worker.py
```diff
@@ -172,15 +170,15 @@ def test_runErrorWithFrames(self):
         results = []
         errorClass = fullyQualifiedName(ValueError)
         d = self.worker.callRemote(managercommands.AddError,
-                                   testName=self.testName, error=b'error',
-                                   errorClass=errorClass.encode("ascii"),
-                                   frames=[b"file.py", b"invalid code", b"3"])
+                                   testName=self.testName, error='error',
+                                   errorClass=errorClass,
+                                   frames=["file.py", "invalid code", "3"])
```
markrwilliams, line 175: This covers https://twistedmatrix.com/trac/ticket/9436.

glyph (approved):
The previous review indicated the code was of sufficient quality, but then rejected the change due largely to concerns about protocol compatibility.  Compatibility concerns are indeed serious, so I'm glad they were considered, but I think the decision was deeply misguided in this case.

First and foremost:

> (07:51:35 PM) runciter: adiroiban: it isn't yet, but it's hard to take back a change to a wire protocol
> (07:51:45 PM) runciter: especially one that isn't versioned, like this one

1. It's currently not possible to connect to any software speaking this protocol over a network connection.  All the classes being modified here are in a private implementation detail within Trial.  By definition, this is as much an "implementation detail" as it is possible for any code to be.
2. Versioning is a red herring; all the commands have names, all the arguments are tagged, and we could easily add a negotiation phase to connecting; if there were anything to connect to (which, see point 1, there isn't).
3. The *wire* protocol here, if it were a concern (it isn't, see point 1) is not changing.  The octets sent over the wire will be exactly the same for any python-2 compatible code, since python-2 identifiers can only be ASCII.  So the only *possible* concern here would be the *API* for the protocol changing in a way that some client code might be confused about: but then, again, the API does not qualify for a compatibility contract.
4. If we were to take this code and make it public, as has been intimated we might, would there be a concern then, with differing behavior between python 2 and python 3?  In most cases, concern about the distinction between text and bytes is valid; you really want to stick to the appropriate type consistently across versions.  But this is the one case where that concern is explicitly invalid: the type of "python identifier" (i.e., `.__name__`) in the 2/3 compatible subset of Python is "native string", so not only *can* a faithful 2/3 compatible reproduction of this logic use native strings, it *must* do so in order to communicate on the wire about identifiers in order to work with APIs like `namedAny`, `qual`, `twisted.python.modules.PythonModule`, etc etc.  Jean-Paul suggested that python 2 could do extra work to convert these identifiers to the appropriate type after it receives them from AMP, but… that's exactly what AMP Argument objects do. Declaratively convert the bytes on the wire into the type desired by the application.

---

More generally, I am a little disappointed that this code got this review and remained in this state, and therefore will miss 18.6.0.  *Le mieux est l'ennemi du bien*.  While we should always respect compatibility concerns and long-term API contracts, when raising the spectre of such a concern, we must be careful to ask "where is the surface that touches the application where this might matter".  We already have an extremely strict and formal definition of compatibility; we should not block forward progress - it's important to remember that `trial -j` is just totally unusable on python 3 without this change, and more work is probably required before it's truly useful - based upon vague concerns that maybe some private APIs *might* be public one day, or some bytes passed only over a local pipe *might* be passed over a socket one day.  The one valid bit of feedback here might have been to add something to a docstring so as to be sure to revisit these decisions in the event that we *do* ever make this protocol public.

So, in the future: reviewers, please try to be permissive within the existing policy, and allow important fixes to get made even if their internals aren't necessarily ideal.  And authors, please push back on feedback like this, and ask if it's really necessary to meet our stated quality bar or if it's just vague concerns about things that might one day happen.  I promise that a `trial -j`-replaces `tox` future is never going to get here if we can't even get it to work for basic use-cases on our supported platforms & python versions :)

adiroiban:
Thanks @glyph for the review.
I closed it after talking with Mark over IRC.
My understanding was that at that time he agreed that is ok to close it.

I asked the question about the importance of this code
> But, it this code used outside of Twisted? I see this code is in a private _dist package.
>
> I have never used this functionality, so I guess is up to the people using it to decide the tradeoff between clean code vs protocol consistency

See the IRC conversation https://github.com/twisted/twisted/pull/960#pullrequestreview-106708763

There, I was ok to merge it.

--------
 
Note that I have not reviewed the PR after it was reopened.... maybe we can configure GitHub to discard any previous reviews on new commits.

-----

Regarding the PR authors pushing for reviews... not sure if that is ok
We definitively need more people to review.
I don't I think that is ok for the author to push for reviews.

glyph:
> Thanks @glyph for the review.
> I closed it after talking with Mark over IRC.
> My understanding was that at that time he agreed that is ok to close it.

Yep!  And he shouldn't have :).

> I asked the question about the importance of this code
> 
> But, it this code used outside of Twisted? I see this code is in a private _dist package.
> 
> I have never used this functionality, so I guess is up to the people using it to decide the tradeoff between clean code vs protocol consistency
> See the IRC conversation #960 (review)
> 
> There, I was ok to merge it.
> 
> Note that I have not reviewed the PR after it was reopened.... maybe we can configure GitHub to discard any previous reviews on new commits.

Yeah, the whole "dismiss review" thing is a weird workflow.

> Regarding the PR authors pushing for reviews... not sure if that is ok
> We definitively need more people to review.
> I don't I think that is ok for the author to push for reviews.

I'm suggesting that the author advocate for their change by describing it in accurate terms, not to push for a specific outcome just because they want it to land.

glyph:
@adiroiban To clarify my comment, one of the things I probably should have stressed here: every individual person performed their role mostly correctly:

1. JP voiced a technical concern we might want to consider
2. you did a review on a ticket that you didn't necessarily have the full context on, to try to help move things along
3. mark paid attention to review feedback

according to the process, these are all good things to do, individually.  So I'm not trying to call anyone out for having broken the process or done anything *wrong*; I'm just suggesting is that we need to be mindful of the ultimate outcome as well.

Mostly, we should not be too fearful of making changes that might break for reasons we don't understand.  Once we've analyzed the change to the best of our ability (everybody concerned responded to JP's feedback saying that the mechanism changing is private) we should make the change, and if we need to roll it back or some unforseen consequence happens, c'est la vie.  We can only do our best, no better.

adiroiban:
@glyph thanks for the follow-up.

I think that there is a bigger problem here.
In that, there is not much of a community for Twisted.

I feel that everyone (including me) is on his/her own and is not available to dedicate time to the work done by others.

For this PR, what I can tell, is that nobody else, other than Mark cared about this `_dist` change.
I tried to help, but I didn't care if this was merged or not. 
Starting today, you also care about this PR and we can see that once we have 2 people, things started to move.

When I see that I am the only one who wants to see a change, I stopped pushing for changes to Twisted and just keep a fork.

HTTP server is a thing which I think is used by many people, yet I can give this small example https://github.com/twisted/twisted/pull/897
Now, for that example I can see that it no longer has the "review" keyword, but look in Trac history, I don't know why https://twistedmatrix.com/trac/ticket/9288

This is not a rant and I know that everybody is busy with life and work and other stuff.

I feel that you care about Twisted community so I just wanted to send my feedback.

-----
There are a few reviewers, yet we ask a lot from the reviewer.

My suggestion is to simplify the infrastructure and make it easier to contribute and to review.
Use only GitHub... Require only Travis/Appveyor/Circle-CI for a merge.

I know that using Trac is not that hard, but working with it is not fun.
When you are doing something for which you don't have any direct benefit, I think that being able to do it easy and without stress is an important thing.
Otherwise, is too easy to give up.

Maybe, it's worth creating a survey to see why people are not reviewing or collaborating on Twisted.
Does Twisted still have some money?
Maybe it can be used to hire a consultant to do the survey.
If there is no money, maybe we can try to raise them.

-------

Maybe is just me, but I feel that something is broken with the Twisted community.
I don't know exactly why and I can't think of something that can be done to change the things.

PR #961, opened 2018-02-20
Create a utility function for converting unicode to bytes

rodrigc:
http://twistedmatrix.com/trac/ticket/9343

There are several places in the Twisted source where a pattern like:

```
if isinstance(s, unicode):
    s = s.encode("ascii")
```

This attempts to encapsulate this pattern in a function.

Based on this feedback from an earlier PR, I did not put this in `twisted.python.compat`: https://github.com/twisted/twisted/pull/936#discussion_r153816654

My motivation for this is to fix some problems I found in `twisted.web` when I found this problem when trying to port https://github.com/LeastAuthority/txkube to Python 3:
https://github.com/twisted/twisted/pull/936#discussion_r153823007

Specifically if I have:
```
networkString(foo)
```
if `foo` is of type `bytes` on Python 2, it will work fine, but if `foo` is of type `bytes` on Python 3,
this `networkString` will throw an exception.

If this code goes in, I would also like to delete the `_ensureBytes()` private function in https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http.py#L1154

### DIFF
--- a/src/twisted/newsfragments/9343.feature
+++ b/src/twisted/newsfragments/9343.feature
@@ -1,1 +1,1 @@
-twisted.python.bytes.toBytes is a convenience function for converting unicode strings to bytes.
+twisted.python.bytes.ensureBytes is a convenience function for converting unicode strings to bytes.
--- a/src/twisted/newsfragments/9343.feature
+++ b/src/twisted/newsfragments/9343.feature
@@ -1,0 +1,1 @@
+twisted.python.bytes.toBytes is a convenience function for converting unicode strings to bytes.
--- a/src/twisted/newsfragments/9343.feature
+++ b/src/twisted/newsfragments/9343.feature
@@ -1,1 +1,1 @@
-twisted.python.bytes.ensureBytes is a convenience function for converting unicode strings to bytes.
+twisted.python.bytes.ensureBytes was added for converting Unicode to bytes.
--- a/src/twisted/python/bytes.py
+++ b/src/twisted/python/bytes.py
@@ -25,13 +25,9 @@ def ensureBytes(s, encoding="ascii", errors="strict"):
     @return: The encoded string.  If I{s} is L{bytes} just return I{s}.
     @rtype: L{bytes}
     """
-    if isinstance(s, bytes):
-        return s
-    elif isinstance(s, unicode):
+    if isinstance(s, unicode):
         return s.encode(encoding=encoding, errors=errors)
-    else:
-        raise TypeError("Expected {} to be unicode or bytes not: "
-            "{}".format(s, type(s)))
+    return s
 
 
 
--- a/src/twisted/python/bytes.py
+++ b/src/twisted/python/bytes.py
@@ -1,0 +1,37 @@
+# -*- test-case-name: twisted.python.test.test_bytes -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+Utility for converting between L{unicode} and L{bytes}
+"""
+
+from __future__ import division, absolute_import, print_function
+
+from twisted.python.compat import unicode
+
+
+def ensureBytes(s, encoding="ascii", errors="strict"):
+    """
+    Convert L{unicode} string to L{bytes}.
+
+    @param s: The string to convert.
+    @param encoding: The encoding to pass to L{str.encode}.
+    @param errors: The error handling scheme to pass to L{str.encode}.
+
+    @raise TypeError: The input is not L{unicode}, L{bytes}.
+
+    @return: The encoded string.  If I{s} is L{bytes} just return I{s}.
+    @rtype: L{bytes}
+    """
+    if isinstance(s, bytes):
+        return s
+    elif isinstance(s, unicode):
+        return s.encode(encoding=encoding, errors=errors)
+    else:
+        raise TypeError("Expected {} to be unicode or bytes not: "
+            "{}".format(s, type(s)))
+
+
+
+__all__ = ['ensureBytes']
--- a/src/twisted/python/bytes.py
+++ b/src/twisted/python/bytes.py
@@ -16,10 +16,11 @@ def ensureBytes(s, encoding="ascii", errors="strict"):
     Convert L{unicode} string to L{bytes}.
 
     @param s: The string to convert.
+    @type s: L{unicode} or L{bytes}
     @param encoding: The encoding to pass to L{str.encode}.
     @param errors: The error handling scheme to pass to L{str.encode}.
 
-    @raise TypeError: The input is not L{unicode}, L{bytes}.
+    @raise TypeError: The input is not L{unicode} or L{bytes}.
 
     @return: The encoded string.  If I{s} is L{bytes} just return I{s}.
     @rtype: L{bytes}
--- a/src/twisted/python/test/test_bytes.py
+++ b/src/twisted/python/test/test_bytes.py
@@ -14,7 +14,8 @@ from twisted.trial.unittest import TestCase
 
 class EnsureBytesTests(TestCase):
     """
-    L{twisted.python.bytes.ensureBytes} is used to convert L{unicode} to L{bytes}.
+    L{twisted.python.bytes.ensureBytes} is used to convert L{unicode} to
+    L{bytes}.
     """
 
     def test_ensureBytesUnicode(self):
--- a/src/twisted/python/test/test_bytes.py
+++ b/src/twisted/python/test/test_bytes.py
@@ -64,6 +64,6 @@ class EnsureBytesTests(TestCase):
     def test_ensureBytesTypeError(self):
         """
         L{ensureBytes} will raise L{TypeError} if passed anything which
-        is not L{unicode}, L{bytes}, or L{None}.
+        is not L{unicode} or L{bytes}.
         """
         self.assertRaises(TypeError, ensureBytes, ["foo", "bar"])
--- a/src/twisted/python/test/test_bytes.py
+++ b/src/twisted/python/test/test_bytes.py
@@ -1,0 +1,69 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+Tests for L{twisted.python.bytes}.
+"""
+
+from __future__ import division, absolute_import, print_function
+
+from twisted.python.bytes import ensureBytes
+from twisted.trial.unittest import TestCase
+
+
+
+class EnsureBytesTests(TestCase):
+    """
+    L{twisted.python.bytes.ensureBytes} is used to convert L{unicode} to L{bytes}.
+    """
+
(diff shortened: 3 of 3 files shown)


markrwilliams:
I agree with [@exarkun's previous review](https://github.com/twisted/twisted/pull/936#discussion_r153824653).  It's hard to say from the txkube build you linked to, but because [the traceback](https://travis-ci.org/LeastAuthority/txkube/jobs/304997908#L945) [comes from `txkube.test.test_network.NetworkKubernetesFromContextTests.test_client_chain_certificate`](https://travis-ci.org/LeastAuthority/txkube/jobs/304997908#L951) [and goes through `twisted.web.static.Data`](https://github.com/twisted/twisted/blob/twisted-17.9.0/src/twisted/web/static.py#L59) I'm guessing that [this line](
https://github.com/LeastAuthority/txkube/blob/0a69bb37a177e927173181ae59e2d8082fa468dd/src/txkube/test/test_network.py#L294) was the culprit:

```python
root.putChild(b"", Data(b"success", b"text/plain"))
```

This is in fact a bug in txkube, since [`Data` expects the content type to be a native string](https://github.com/twisted/twisted/blob/c16491942672d5d4d589894fc2a878eb80abed44/src/twisted/web/test/test_static.py#L41).  Extending `Data`'s API to accept native strings _or_ bytes _or_ text seems bad because [HTTP headers have a specific encoding rules](https://tools.ietf.org/html/rfc7230#section-3.2.4).

I guess we already decided that [HTTP cookies can be native strings _or_ bytes _or_ text](http://twistedmatrix.com/trac/ticket/8067) as well as [HTTP header keys and values](http://twistedmatrix.com/trac/ticket/8129).  In that case the correct solution is to remove `networkString` from `Data`'s `setResponseHeader` call.

Where else would this `toBytes` helper be useful?

adiroiban:
Thanks for working on this API. I think that is important to have good support for bytes between py2 and py3 and I am happy to see work done in this area :)

> if foo is of type bytes on Python 2, it will work fine, but if foo is of type bytes on Python 3,
> this networkString will throw an exception.

I see that networkString API asks for `{str}` as input.

Maybe networkString API can be updated to also accept Py3 `bytes` on top of str, to make it easier to use it.

What do you think?

rodrigc:
@adiroiban I thought about modifying `networkString` but went with a new function for the following reasons
1.  `twisted.python.compat` was a module written to facilitate porting Twisted to Py3.  Since Twisted now works 95%+ on Py3, I think it has served its purpose.  Long term I don't really want to put much effort into adding new stuff to `twisted.python.compat` or modifying the functions in it.  When the Py3 porting work for Twisted began, this module needed to be written, but in hindsight, some of the mainstream libraries for facilitating Py3 porting maybe should have been used (six, future, etc.).    However, what is done is done.

2.  `networkString` isn't a very good function name.  The name doesn't reflect that this function takes in str or bytes and returns bytes.  Whether or not the output goes over the network is secondary.

3.  `networkString` has some quite involved behavior and tries to deal with Py2 vs. Py3 differences.  For most cases that I've needed, it is overkill.

4.  I attended this presentation ( https://www.youtube.com/watch?v=aetoXWRt24k ) given by Roy Williams ( @rowillia ) about Lyft's Py3 porting efforts.  Roy mentioned that Lyft uses an `ensure_bytes` function: https://github.com/lyft/syx/blob/master/syx/__init__.py#L23 

That seems much simpler and more suited to what I want than `networkString`

So for these reasons, and inspired by Lyft, I went ahead and wrote a new `ensureBytes` function instead of modifying `networkString`.

rodrigc:
@markrwilliams in addition to the original examples I mentioned, there are lots of places in the
Twisted code that do:

```
if isinstance(s, unicode):
    s = s.encode("utf-8")
```

that are candidates to be replaced by `ensureBytes`.

Using:
```
grep -r 'if isinstance(' * | grep unicode
```

shows some of the possible places.

src/twisted/python/bytes.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @return: The encoded string.  If I{s} is L{bytes} just return I{s}.
+    @rtype: L{bytes}
+    """
+    if isinstance(s, bytes):
+        return s
+    elif isinstance(s, unicode):
+        return s.encode(encoding=encoding, errors=errors)
+    else:
+        raise TypeError("Expected {} to be unicode or bytes not: "
```
adiroiban: Not sure if we need this here.
It is documented that it needs Unicode or bytes. When people pass something else, they are on their own.

src/twisted/python/test/test_bytes.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+Tests for L{twisted.python.bytes}.
+"""
+
+from __future__ import division, absolute_import, print_function
```
adiroiban, line 9: just a minor comment... to make it easier to test py3 stuff on py27 I also import unicode_literals

src/twisted/python/test/test_bytes.py
```diff
@@ -0,0 +1,70 @@
...
+            b'')
+
+
+    def test_ensureBytesUnicodeEncodeError(self):
+        """
+        L{UnicodeEncodedError} will be raised if the input string cannot be
+        converted using the encoding's rules.
+        """
+        self.assertRaises(UnicodeEncodeError,
+            ensureBytes, u'\N{SNOWMAN}', encoding="ascii")
```
adiroiban: minor comment of style... why different indentation style?

I prefer the one used in test_ensureBytesErrorsParameter so this would be

```
self.assertRaises(
    UnicodeEncodeError,
    ensureBytes, u'\N{SNOWMAN}', encoding="ascii")
```

src/twisted/python/test/test_bytes.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertEqual(ensureBytes(u"hello"), b"hello")
+
+
+    def test_ensureBytesEncodingParameter(self):
+        """
+        The I{encoding} parameter is passed to L{str.encode} and selects the
+        codec to use when converting L{unicode} to L{bytes}.
+        """
+        self.assertEqual(ensureBytes(u'\N{SNOWMAN}', encoding="utf-8"),
+                         b'\xe2\x98\x83')
```
adiroiban: minor comment on indentation style.
why not

```
self.assertEqual(
    ensureBytes(u'\N{SNOWMAN}', encoding="utf-8"),
    b'\xe2\x98\x83')
```

src/twisted/python/test/test_bytes.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_ensureBytesErrorsParameter(self):
+        """
+        The I{errors} parameter is passed to L{str.encode} and specifies the
+        response when the input string cannot be converted according to the
+        encoding’s rules.
+        """
+        self.assertEqual(
+            ensureBytes(u'\N{SNOWMAN}', encoding="ascii", errors="ignore"),
+            b'')
```
adiroiban: A minor question of style.

To make it easier to read test failures, have the expectation first. So assertions are called like self.assertSomething(expected, actual)

src/twisted/newsfragments/9343.feature
```diff
@@ -0,0 +1 @@
+twisted.python.bytes.ensureBytes is a convenience function for converting unicode strings to bytes.
```
adiroiban: Minor comment. Maybe:

```
twisted.python.bytes.ensureBytes was added for converting Unicode to bytes.
```

src/twisted/python/bytes.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+
+from __future__ import division, absolute_import, print_function
+
+from twisted.python.compat import unicode
+
+
+def ensureBytes(s, encoding="ascii", errors="strict"):
+    """
+    Convert L{unicode} string to L{bytes}.
```
adiroiban, line 16: maybe the docstring should add a warning that this should not be used for py2/py3 porting.

adiroiban (approved):
I think that we can have this.

From @exarkun 
> So I'd recommend being very careful where this conversion function is used: it could easily take us back to the bad-old-times of carelessly mixed bytes/unicode.

With or without this helper, people might write mixed code.

I guess that diligent reviews are needed, with or without this helper.

And maybe with having this helper is easier to do a later audit of the code... I don't know :)

So I think that we should merge it and see how it is used and abused :)

Thanks @rodrigc  for going to all this pain :)

src/twisted/python/bytes.py
rodrigc: @adiroiban
are you OK if I change `ensureBytes` to look like https://github.com/lyft/syx/blob/master/syx/__init__.py#L23 with no `raise TypeError`?

Also, I think I want to move this function inside `twisted.python.__init__.py` instead
of having it in a separate bytes.py file.  Are you OK with that?

exarkun: No.  No implementations in `__init__.py`.  And `twisted.python` doesn't expose an API except for the modules it containers.

adiroiban: I am ok with the implementation from syx.
I am +1 for not raising TypeError and is ok to add an encoding mode (with strict as default)... but I guess that we can add the argument later if required.
For now we can document that it goes with strict encoding.

-1 for moving in `__ini__ `. We can keep it in a separate module, as is now in the PR

Thanks

markrwilliams:
@rodrigc Thanks for [the pointer to uses of this in Twisted](https://github.com/twisted/twisted/pull/961#issuecomment-367250447).  I'm sorry I'm only noticing it now.

Many occurrences of `if isinstance(value, unicode): value = value.decode(ENCODING)` don't justify the addition of `ensureBytes`:

1) Often `ENCODING` is not UTF-8; for example, [FTP has to send lines](https://github.com/twisted/twisted/blob/d23799b6f642fa8efecc651d9ddb7e5c57bf6ea6/src/twisted/protocols/ftp.py#L759-L760) [as `latin1`](https://github.com/twisted/twisted/blob/d23799b6f642fa8efecc651d9ddb7e5c57bf6ea6/src/twisted/protocols/ftp.py#L2473); [IMAP4 has to send its nested  ASCII](https://github.com/twisted/twisted/blob/trunk/src/twisted/mail/imap4.py#L5038-L5042);
2) Places where native strings are used but richer Python objects would be more appropriate (basically all of `imap4`);
3) And many were [just](https://github.com/twisted/twisted/blob/trunk/src/twisted/conch/client/options.py#L76-L77) [the easiest](https://github.com/twisted/twisted/blob/trunk/src/twisted/protocols/sip.py#L632-L633) [way](https://github.com/twisted/twisted/blob/trunk/src/twisted/web/microdom.py#L77-L78) at the time -- why should any of these be UTF-8?

Furthermore, the `txkube` case demonstrates how this obscures actual bugs.

PR #962, opened 2018-02-20
AppVeyor skip config for newsfragment changes

exarkun:
https://twistedmatrix.com/trac/ticket/9380

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,12 @@ test: off
 # commits on random branches without an associated PR.
 skip_branch_with_pr: true
 
+# Don't build branch if the only change in the HEAD commit is a news fragment.
+# https://www.appveyor.com/docs/how-to/filtering-commits/
+only_commits:
+  files:
+    - '**/newsfragments/*'
+
 # Not sure if the cache works.
 # Note the use of single quotes around the entire line, when environment
 # variables are used.


adiroiban:
This is not a review... just some comments.

I am waiting to see how Appveyor react :)

... but if you commit the news fragment last, and it will not trigger a build, I expect that it will not have the "green" commit status.

And Travis-CI are still executed...and they can also take time.

-----

There is the option to add 'no-ci' in the commit message, to fully skip CI and maybe we can have no-appveyor no-travis ... etc.. but I don't know if this makes sense.

You have the test to run them :) for any change.

---------

What I would like to see is a quick command which can be executed locally by each developer and which will do all these static checks, in order to not waste time triggering tests for a commit which will fail.

----------

What I am doing for my project which is using Buildbot is to run tests in stages.
The static linters and a few other fast tests are executed as stage 1... and only after they pass, the heavy tests are triggered.

Travis got support for stages ... so maybe we can try it. I have not tried Travis yet https://docs.travis-ci.com/user/build-stages/

So we can create a quick-test tool which can be used as stage 1 (and should execute in less than 30 secondes) and only after it passes the other tests are triggered.

markrwilliams:
@adiroiban I would love linting for this that I could run locally!

Also note that people have worked around Travis' feature deficit by using private APIs:

https://github.com/twisted/twisted/commit/bc5868e466c13134af68dc69b349eb52bdd7a57e

So maybe that's an option as well.

markrwilliams:
Also note that AppVeyor only considers `[skip ci]` in "commit titles":

https://www.appveyor.com/docs/how-to/filtering-commits/#skip-directive-in-commit-message

So attempting to use `[skip ci]` for both Travis and AppVeyor seems easy to get wrong.

adiroiban:
@exarkun it looks like Apveyor runs the test on the merged PR version... and reports status for the pre-merged commit.

The merge will always contain more files, so I think that Appveyor will alwasy run it.

I have disabled Appveyour tests for the "push" version of tests run.

markrwilliams (approved):
I can't figure out if AppVeyor reliably bestows a green checkmark to a commit that contains only ignored files.

The [newsfragment-only commit in this PR](https://github.com/twisted/twisted/pull/962/commits/13895e37df1c04305f47615f0ef42d5a061b70d8) links to [a successful AppVeyor build](https://ci.appveyor.com/project/adiroiban/twisted/build/trunk-4533) for [a different merge commit that isn't even in the branch](https://github.com/twisted/twisted/commit/bc5868e466c13134af68dc69b349eb52bdd7a57e).  @exarkun certainly didn't create it.

Maybe this merge commit is how AppVeyor handles commits that only contain ignored files?

Despite this uncertainty, our managed CI build queues are a scarce enough resource that I think we should merge this and see what happens.

exarkun:
> The merge will always contain more files, so I think that Appveyor will alwasy run it.

Yes.  This seems to mean there's basically nothing to be gained here.  Perhaps we could skip Windows builds for doc-only changes or something but that case seems a lot more obscure and I don't find compelling.  Probably should close this without merging.

adiroiban:
I think that this can be closed without merging.

My suggestion:

* Improve tools for running test on local system before a push (ex write scripts for running a few quick tests) 
* Try running test in stages... with the logic done from Travis-CI. Travis-CI will trigger both Buildbot and Appveyor.

I can help with setting up the stages in Travis and doing the triggers for the Builbot and Appveyor builds

exarkun:
I don't think the stages idea really does much to help.  It just means you get incomplete builds more often.  Improving tools for local checking would certainly be great.  AFAICT it's not actually possible to do all the checks locally in less than an hour (mostly spent thinking or running garbage commands) or so.

PR #963, opened 2018-02-20
If selected port is not really free, try another

exarkun:
https://twistedmatrix.com/trac/ticket/9381

### DIFF
--- a/src/twisted/internet/test/test_tcp.py
+++ b/src/twisted/internet/test/test_tcp.py
@@ -26,7 +26,9 @@ from twisted.python import log
 from twisted.trial.unittest import SkipTest, TestCase
 from twisted.internet.error import (
     ConnectionLost, UserError, ConnectionRefusedError, ConnectionDone,
-    ConnectionAborted, DNSLookupError, NoProtocol)
+    ConnectionAborted, DNSLookupError, NoProtocol,
+    ConnectBindError,
+)
 from twisted.internet.test.connectionmixins import (
     LogObserverMixin, ConnectionTestsMixin, StreamClientTestsMixin,
     findFreePort, ConnectableProtocol, EndpointCreator,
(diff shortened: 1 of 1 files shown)


src/twisted/internet/test/test_tcp.py
```diff
@@ -630,9 +632,23 @@ def makeConnection(self, transport):
         clientFactory.protocol = CheckAddress
 
         def connectMe():
-            reactor.connectTCP(
-                fakeDomain, server.getHost().port, clientFactory,
-                bindAddress=(self.interface, port))
+            while True:
```
markrwilliams, line 635: While unlikely, it's possible that this could run for a while.  Perhaps it would be good to retry only for a certain amount of time, after which the test would be `fail`ed?  This assumes you can trust `time.time`.

src/twisted/internet/test/test_tcp.py
```diff
@@ -630,9 +632,23 @@ def makeConnection(self, transport):
...
+                ignored, port = findFreePort(self.interface, self.family)[:2]
+                bindAddress = (self.interface, port)
+                log.msg("Connect attempt with bindAddress {}".format(bindAddress))
```
markrwilliams: [`twistedchecker` says this line is too long.](https://travis-ci.org/twisted/twisted/jobs/343852852#L540)

markrwilliams (approved):
A suggestion for limiting the retry loop.  Feel free to ignore it.

Please address `twistedchecker`'s complaint before merging.

src/twisted/internet/test/test_tcp.py
exarkun: If the host is so hosed that it's impossible to ever connectTCP without a ConnectBindError then I don't know what's going on.  I guess I'll leave this as-is.

PR #965, opened 2018-02-23
Bugfix: slot data stack is badly cleaned in templating

SamyCookie:
see ticket #9383 in trac. - https://twistedmatrix.com/trac/ticket/9383

In the templating module, more precisely in the '_flattenElement' method, when getting through elements in the tree, some cleanings of the slot data stack are missing in some cases.

### DIFF
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -241,6 +241,7 @@ def _flattenElement(request, root, write, slotData, renderFactory,
 
         if not root.tagName:
             yield keepGoing(root.children)
+            slotData.pop()
             return
 
         write(b'<')
@@ -273,6 +274,8 @@ def _flattenElement(request, root, write, slotData, renderFactory,
             write(b'</' + tagName + b'>')
         else:
             write(b' />')
+
+        slotData.pop()
 
     elif isinstance(root, (tuple, list, GeneratorType)):
         for element in root:
--- a/src/twisted/web/newsfragments/9383.bugfix
+++ b/src/twisted/web/newsfragments/9383.bugfix
@@ -1,0 +1,1 @@
+twisted.web._flatten._flattenElement, slot data stack is cleaned after flattening a node tree.
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -433,6 +433,28 @@ class SerializationTests(FlattenTestCase, XMLAssertionMixin):
         return self.assertFlattensTo(t, b'<p><em>four&gt;</em></p>')
 
 
+    def test_slotDataPropagation(self):
+        """
+        slot data propagation is only made to child nodes.
+        Before #9383 fix, slot datas were not correctly clean during tree
+        traversal and some elements that are not directly child nodes can
+        reach forbidden slot data.
+        """
+        span1 = tags.span(slot('test'))
+        span2 = tags.span(slot('test', default='17'))
+        tree = tags.div(
+            tags.p(span1).fillSlots(test='42'),
+            tags.p(span2)
+        )
+        result = (
+            b'<div>'
+            b'<p><span>42</span></p>'
+            b'<p><span>17</span></p>'  # '42' was written before #9383
+            b'</div>'
+        )
+        return self.assertFlattensTo(tree, result)
+
+
     def test_unknownTypeRaises(self):
         """
         Test that flattening an unknown type of thing raises an exception.


src/twisted/web/newsfragments/9383.bugfix
```diff
@@ -0,0 +1 @@
+in templating module, slot data stack is badly cleaned after flattening a node tree
```
adiroiban: Please start this with the full name of the public class/module/function which is affected...and end the sentence with a full stop :)

src/twisted/web/test/test_flatten.py
```diff
@@ -431,6 +431,18 @@ def test_serializeDeferredSlots(self):
         t = tags.p(slot('test'))
         t.fillSlots(test=succeed(tags.em('four>')))
         return self.assertFlattensTo(t, b'<p><em>four&gt;</em></p>')
+    
+    
+    def test_slotDataPropagation(self):
+        """ Test that verify if a slot data inserted in a node is correctly propagated
```
adiroiban: See this for how to write docstrings https://jml.io/pages/test-docstrings.html

See the docstring from for the other tests.

No need for "test that", "check that". All tests are written to test something :)

What means "correctly"?

src/twisted/web/test/test_flatten.py
```diff
@@ -431,6 +431,18 @@ def test_serializeDeferredSlots(self):
...
+        and can be retrieved in the child nodes and ONLY in the child nodes
+        (not in sibling nodes for instance).
+        """
+        span1 = tags.span( slot('test') )
+        span2 = tags.span( slot('test', default='17') )
+        tree = tags.div( tags.p(span1).fillSlots(test='42'), tags.p(span2) )
+        result = b'<div><p><span>42</span></p><p><span>23</span></p></div>'
+        return self.assertFlattensTo(tree, result)
```
adiroiban: See code formatting for the other Twisted code. http://twistedmatrix.com/documents/current/core/development/policy/index.html

I am not sure which is the system under test here. Which function or code is tested?

src/twisted/web/_flatten.py
```diff
@@ -229,50 +229,51 @@ def keepGoing(newRoot, dataEscaper=dataEscaper,
         write(b'-->')
     elif isinstance(root, Tag):
         slotData.append(root.slotData)
+
```
adiroiban: Minor comment
Do we need this change?
I know that you might not like the spacing..but other dev might not like your style...so I don't want to start an edit war :)

src/twisted/web/_flatten.py
```diff
@@ -229,50 +229,51 @@ def keepGoing(newRoot, dataEscaper=dataEscaper,
...
             rootClone = root.clone(False)
             rootClone.render = None
             renderMethod = renderFactory.lookupRenderMethod(rendererName)
             result = renderMethod(request, rootClone)
             yield keepGoing(result)
-            slotData.pop()
-            return
-
-        if not root.tagName:
+
```
adiroiban: I don't what to say about these changes.

I would prefer to have `return` here, and maybe use something like `yield keepGoingFinal(result)` which will call slotData.pop()

With the current changes, slotData.pop() is called after many new branches... so this would need more than 1 new test.

adiroiban (changes requested):
Thanks for looking into this issue.

See inline comments.

It is not clear what path / paths are tested by the new tests.

I see that slotData.pop() is not called after more than one execution paths... but I only see a new tests.

Please consider updating the docstring of the tests so that I know what is the intention of that test.

In the same time, the Appveyour and Travis tests are failing at twisted.web.test.test_flatten.SerializationTests.test_slotDataPropagation

which is the new tests that you wrote.

New and existing tests should pass across all the supported environments before merging or being able to approve this change.

Thanks!

src/twisted/web/newsfragments/9383.bugfix
SamyCookie: done

src/twisted/web/test/test_flatten.py
SamyCookie: I tried to be more explicit !

SamyCookie:
@adiroiban sorry for the loooooooooooooong response.

I have updated my commits, removed the polemic one, and made the test pass (I don't know how I failed that).
Is this all more understandable ?

src/twisted/web/newsfragments/9383.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web._flatten._flattenElement, slot data stack is cleaned after flattening a node tree.
```
adiroiban, line 1: this talks about a private API.

What public API is targeted by this change?

src/twisted/web/test/test_flatten.py
```diff
@@ -433,6 +433,28 @@ def test_serializeDeferredSlots(self):
...
+        span1 = tags.span(slot('test'))
+        span2 = tags.span(slot('test', default='17'))
+        tree = tags.div(
+            tags.p(span1).fillSlots(test='42'),
+            tags.p(span2)
+        )
+        result = (
+            b'<div>'
+            b'<p><span>42</span></p>'
+            b'<p><span>17</span></p>'  # '42' was written before #9383
```
adiroiban, line 452: no need to document the broken implementation.
just try to explain what is the expected result.
so I think that this comment can be removed.

src/twisted/web/test/test_flatten.py
```diff
@@ -433,6 +433,28 @@ def test_serializeDeferredSlots(self):
         return self.assertFlattensTo(t, b'<p><em>four&gt;</em></p>')
 
 
+    def test_slotDataPropagation(self):
+        """
+        slot data propagation is only made to child nodes.
+        Before #9383 fix, slot datas were not correctly clean during tree
+        traversal and some elements that are not directly child nodes can
+        reach forbidden slot data.
```
adiroiban, line 441: What is a forbidden slot data?

I have never worked with this API...so I am not sure about the terminology.

But to me, it looks like this test is about.

```
Slot data is applied only to the targeted tag.
```

No need to talk about previous bugs or invalid behavior.
Just document the correct behavior as good as possible.

src/twisted/web/test/test_flatten.py
```diff
@@ -433,6 +433,28 @@ def test_serializeDeferredSlots(self):
...
+        """
+        span1 = tags.span(slot('test'))
+        span2 = tags.span(slot('test', default='17'))
+        tree = tags.div(
+            tags.p(span1).fillSlots(test='42'),
```
adiroiban, line 446: not sure why we need the P tag for this test.
Is it really needed?

adiroiban (changes requested):
Thanks for the update.
Don't worry about the delay.
Looks good.

I still see 2 `slotData.pop()`, yet there is only a new test.

Is the new test covering both changes?

To make it easy to follow the implementation, I would expect to have 2 tests, each test with is own docstring.

I have never  used the flatten API so my review might be wrong.

My suggestion is to reach other Twisted developers ove IRC or mailling list and see if someone else can take a look at this.

src/twisted/web/newsfragments/9383.bugfix
SamyCookie: `twisted.web.template.flatten` and `twisted.web.template.flattenString`

PR #966, opened 2018-02-26
Fix syntax error under Python 3.7 for async param

lopuhin:
Syntax error under Python 3.7 fixed for ``twisted.conch.manhole`` and ``twisted.main.imap4``. ``async_`` keyword argument is used by default instead of ``async``, but ``async`` is also supported for backwards compatibility.

Twisted issue: https://twistedmatrix.com/trac/ticket/9384


lopuhin:
One build failed due to ``Invalid argument name "async_"`` (https://travis-ci.org/twisted/twisted/jobs/346467875#L537) - probably it does not like a trailing underscore. AFAIK ``async_`` is a commonly used way to replace ``async`` argument, but I can change it to something like ``is_async`` as well - what would be better?

rodrigc:
Try using `_async` for the variable and `getAsyncParam()` for the function.  The https://github.com/twisted/twistedchecker program seems to accept that.

lopuhin:
> Try using _async for the variable and getAsyncParam() for the function.

Thanks @rodrigc , trying it in 0ad85ce92

lopuhin:
Thanks, looks like it fixed the build!

lopuhin:
All scrapy tests [pass](https://travis-ci.org/scrapy/scrapy/jobs/349230683) under python 3.7 with this branch. Travis build passed at https://github.com/twisted/twisted/pull/966/commits/30a298ab4e00db645d8552043cc4d0c6c187a927, the last build failed due to an unrelated job erroring out. OS X build was stuck at "Waiting for status to be reported" both times. Is there anything that needs to be fixed?

rodrigc (approved):
Without your patch,
if I run the following command in the top level directory of the trunk branch in Twisted:

```
python -m tox -e pycodestyle
```

I get a few warnings like:

```
src/twisted/mail/imap4.py:1105:45: W606 'async' and 'await' are reserved keywords starting with Python 3.7
```

so your patch definitely fixes the problem.
This is very helpful, since this is a syntax error on Python 3.7.

src/twisted/conch/manhole.py
```diff
@@ -151,8 +153,9 @@ def _ebDisplayDeferred(self, failure, k, obj):
         return failure
 
 
-    def write(self, data, async=False):
-        self.handler.addOutput(data, async)
+    def write(self, data, _async=None, **kwargs):
```
adiroiban: instead of _async, can this be renamed to `asAsync=True|False` ? ...

src/twisted/mail/imap4.py
```diff
@@ -1090,8 +1090,9 @@ def sendNegativeResponse(self, tag = None, message = b''):
         self._respond(b'NO', tag, message)
 
 
-    def sendUntaggedResponse(self, message, async=False):
-        if not async or (self.blocked is None):
+    def sendUntaggedResponse(self, message, _async=None, **kwargs):
+        _async = get_async_param(_async, **kwargs)
```
adiroiban: shouldn't this trigger a deprecation warning?

I expect that in the future code should be updated to just use `_async`...

but since this is public API I am -1 for using `_async` as an argument name

and since this is used as a boolean flag, I am +1 for naming it like `isAsync` or some other name which implies that this is a boolean

src/twisted/conch/manhole.py
lopuhin: +1, I like ``asAsync`` better than ``_async``. I've seen ``async_`` used in other libraries and it matches the PEP-8, but this name does not pass Twisted style check.

adiroiban (changes requested):
Thanks @lopuhin for working on this and @rodrigc for the review :)

I am -1 on this due to the introduction if `_async` as a public API and it looks like @lopuhin agreed on doing the full transition and issue a deprecation warning.

adiroiban:
The patch looks  like fixing all the `W606 'async' and 'await' are reserved keywords starting with Python 3.7` issue. I was running it as
```
$ python admin/pycodestyle-twisted.py --select=W606 src/twisted/

or with tox

$ tox -e pycodestyle '--select=W606 src/twisted/'
```

lopuhin:
Thanks for review @adiroiban , I changed parameter name to ``isAsync`` in ed675f2  and added a deprecation warning in 1634587. Is it fine that the warning does not specify a twisted version when it's deprecated? If it's better to specify it, which version shall it use?

adiroiban:
> If it's better to specify it, which version shall it use?

I don't know what version to put in deprecations. @exarkun / @hawkowl / @glyph . Any suggestion?

I think that you cat put a placeholder like `NEXT`, and it will be automatically  be updated... but I am not sure how this works and I don't see the documentation for it.

For deprecation, please use these helpers http://twistedmatrix.com/documents/current/api/twisted.python.deprecate.html

and the dev docs at http://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#how-to-deprecate-apis

Please add docstrings to every test :)

src/twisted/python/compat.py
```diff
@@ -829,6 +830,34 @@ def _bytesRepr(bytestring):
 
 
 
+def get_async_param(isAsync=None, **kwargs):
```
adiroiban: I think that this should be private...as in, it should only be used inside Twisted core and not used by other projects... so this should be `def _get_async_param`

exarkun:
I intended for the release management tooling to deal with the "NEXT" idea: https://twistedmatrix.com/trac/ticket/3266

That effort seems pretty well stalled.  So I suppose it must be a manual release manager task, hopefully not undocumented but who knows. :/

src/twisted/test/test_compat.py
```diff
@@ -921,3 +921,32 @@ def test_bytesReprPrefix(self):
...
+        self.assertEqual(get_async_param(isAsync=False), False)
+        self.assertEqual(get_async_param(isAsync=True), True)
+        self.assertEqual(
+            get_async_param(isAsync=None, **{'async': False}), False)
+        self.assertEqual(
+            get_async_param(isAsync=None, **{'async': True}), True)
+        self.assertRaises(TypeError, get_async_param, False, {'async': False})
+
+
+    def test_get_async_param_deprecation(self):
```
adiroiban: The test should be for the public API itself, and not for this (what should be) private code.

src/twisted/newsfragments/9384.bugfix
```diff
@@ -0,0 +1,3 @@
+Syntax error under Python 3.7 fixed for twisted.conch.manhole and
+twisted.main.imap4. isAsync keyword argument is used by default instead of
+async, async keywoard argument is deprecated.
```
adiroiban: The deprecation info should be put into a separate file named src/twisted/newsfragments/9384.removal

In this way, the depreciation part of the release notes will contain the info about `async` keyword.
Otherwise, the bugfix part will contain both bugfix and deprecation.

rodrigc (changes requested):
This line is not being hit during the tests:

https://codecov.io/gh/twisted/twisted/compare/6a1d600241c273b693186f2a166b8e9c5717634e...93864a6674733b48e3264c1fe993858da0cd5c5f/diff#D6-856

Can you fix the code or fix your test so that this line is hit during the tests?

rodrigc:
@lopuhin  Can you fix the test?

lopuhin:
@rodrigc sure, sorry that I stopped working on it - I plan to resume on Friday, but if someone wants to finish it sooner I'm totally fine with that too :)

hroncok:
@lopuhin How can I help?

felixyin:
找到这个src/twisted/conch/manhole.py文件，参照下面链接替换一下即可，原因是async是3.7的关键词：
https://github.com/twisted/twisted/commit/30a298ab4e00db645d8552043cc4d0c6c187a927

glyph:
Replying to @exarkun from march:

> That effort seems pretty well stalled. So I suppose it must be a manual release manager task, hopefully not undocumented but who knows. :/

https://github.com/twisted/incremental#updating

To do this within Twisted, simply write "Twisted NEXT".

hawkowl:
Running tests on full builders. If it makes the tests run on 3.7 for imap/manhole, I'm happy.

lopuhin:
Thanks a lot @hawkowl for finishing and merging the fix, sorry for letting it stall!

PR #969, opened 2018-03-17
9389 Add external pycodestyle checker.

adiroiban:
Scope
=====

Trac ticket https://twistedmatrix.com/trac/ticket/9389

This adds an external tool for running pycodestyle on Twisted code.
Previously the pycodestyle fork was part of twistedcheker... but twistedchecker is slow and it looks like contributors are not using twistedchecker that often on local system.

I hope that by making it easier to run pycodestyle on local system, we reduce the load on the CI systems.

I also hope that with a more standard pycodestyle reporting we can use tools like https://github.com/hhatto/autopep8 to automatically update the code.

Here is a statistic with current errors https://gist.github.com/adiroiban/ccfb1283bd7d8641e726a06e36c40cde

I hope that soon we can clean the whole code (as we did with pyflakes) and then contributors should no longer care about existing errors (as there are none:) 

Changes
=======

Created a fork of pycodestyle with custom blank_lines.
I placed it in 'admin' and called it `pycodestyle-twisted.py`.
I went for latest `master` as I hope that in this way we can the latest checks...and I hope that we will have soon a version of pycodestyle which works out of the box with the Twisted blank lines.

I added a pycodestyle tox environment for Travis and for local usage.
The Travis-CI version only checks the diff... and any change should be PEP8 complaint.

I have removed the txchecker-travis-all, as it was set it ignore. And since Travis reports a green build even when it fails, I don't know if anybody was checking it... so it is just a waste of time.

As a drive-by, I added admin/README file, as I am confused about the scope of `admin` vs `bin/admin`

While testing the changes, I fixed a few files.
For example we no longer have mixed tabs and spaces (which were found in a single file) 

How to test
=========

Check that Travis-CI build fails, when the diff has codestyle validations.

--------------

Run local for a specific file and with specific arguments
```
$ admin/pycodestyle-twisted.py --help
$ admin/pycodestyle-twisted.py --select=E272 src/twisted/web/test/test_static.py
```

Run local and check only for changes since trunk.
```
$ git diff trunk... | admin/pycodestyle-twisted.py --diff
```

-----------

Run from Tox
```
$ tox -e pycodestyle admin/pycodestyle-twisted.py  --select=E272 src/twisted/web/test/test_static.py
$ tox -e cipycodestyle
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -7,5 +7,6 @@ https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch
 ## Contributor Checklist:
 
 * [ ] There is an associated ticket in Trac. Create a new one at https://twistedmatrix.com/trac/newticket
+* [ ] The changes pass minimal style checks (see: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted )
 * [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )
 * [ ] I have updated the automated tests.
--- a/admin/pycodestyle-twisted.py
+++ b/admin/pycodestyle-twisted.py
@@ -1,6 +1,32 @@
 #!/usr/bin/env python
 
 """
+Copyright © 2006-2009 Johann C. Rocholl <johann@rocholl.net>
+Copyright © 2009-2014 Florent Xicluna <florent.xicluna@gmail.com>
+Copyright © 2014-2018 Ian Lee <IanLee1521@gmail.com>
+
+Licensed under the terms of the Expat License
+
+Permission is hereby granted, free of charge, to any person
+obtaining a copy of this software and associated documentation files
+(the "Software"), to deal in the Software without restriction,
+including without limitation the rights to use, copy, modify, merge,
+publish, distribute, sublicense, and/or sell copies of the Software,
+and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
 A patched version for pycodestyle which uses the blank line conversion from
 Twisted.
 
@@ -106,5 +132,6 @@ pycodestyle._checks['logical_line'][blank_lines] = _previousValue
 
 
 if __name__ == '__main__':
+    # Calling main in a similar was as the upstream script.
     sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
     sys.exit(pycodestyle._main())
--- a/admin/pycodestyle-twisted.py
+++ b/admin/pycodestyle-twisted.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
 
 """
-Copyright © 2006-2009 Johann C. Rocholl <johann@rocholl.net>
-Copyright © 2009-2014 Florent Xicluna <florent.xicluna@gmail.com>
-Copyright © 2014-2018 Ian Lee <IanLee1521@gmail.com>
+Copyright 2006-2009 Johann C. Rocholl <johann@rocholl.net>
+Copyright 2009-2014 Florent Xicluna <florent.xicluna@gmail.com>
+Copyright 2014-2018 Ian Lee <IanLee1521@gmail.com>
 
 Licensed under the terms of the Expat License
 
--- a/admin/pycodestyle-twisted.py
+++ b/admin/pycodestyle-twisted.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+
+"""
+A patched version for pycodestyle which uses the blank line conversion from
+Twisted.
+
+Upstream integration is pending.
+See https://github.com/PyCQA/pycodestyle/issues/732
+"""
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+    unicode_literals,
+    )
+
+import re
+import sys
+
+import pycodestyle
+
+
+
+def blank_lines(logical_line, blank_lines, indent_level, line_number,
+                blank_before, previous_logical,
+                previous_unindented_logical_line, previous_indent_level,
+                lines):
+    r"""Separate top-level function and class definitions with two blank lines.
+
+    Method definitions inside a class are separated by a single blank line.
+
+    Extra blank lines may be used (sparingly) to separate groups of related
+    functions.  Blank lines may be omitted between a bunch of related
+    one-liners (e.g. a set of dummy implementations).
+
+    Use blank lines in functions, sparingly, to indicate logical sections.
+
+    Okay: def a():\n    pass\n\n\ndef b():\n    pass
+    Okay: def a():\n    pass\n\n\nasync def b():\n    pass
+    Okay: def a():\n    pass\n\n\n# Foo\n# Bar\n\ndef b():\n    pass
+    Okay: default = 1\nfoo = 1
+    Okay: classify = 1\nfoo = 1
+
+    E301: class Foo:\n    b = 0\n    def bar():\n        pass
+    E302: def a():\n    pass\n\ndef b(n):\n    pass
+    E302: def a():\n    pass\n\nasync def b(n):\n    pass
+    E303: def a():\n    pass\n\n\n\ndef b(n):\n    pass
+    E303: def a():\n\n\n\n    pass
+    E304: @decorator\n\ndef a():\n    pass
+    E305: def a():\n    pass\na()
+    E306: def a():\n    def b():\n        pass\n    def c():\n        pass
+    """
+    top_level_lines = 3
+    method_lines = 2
+    if line_number < 3 and not previous_logical:
+        return  # Don't expect blank lines before the first line
+    if previous_logical.startswith('@'):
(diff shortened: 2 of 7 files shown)


adiroiban:
Ubuntu 14.04 was failing due to a flaky test related to client-side passive port initialization.
I have also observed this and I am working on patch to refactor it and use endpoints.

tox.ini
```diff
@@ -124,6 +129,11 @@ commands =
 
     manifest-checker: check-manifest --ignore "docs/_build*,docs/historic/*,admin*,bin/admin*,twisted/topfiles/*.Old"
 
+    pycodestyle: python {toxinidir}/admin/pycodestyle-twisted.py {posargs:admin src/twisted}
+    ; Travis-CI version is executing in diff mode without any input arguments.
+    cipycodestyle: /bin/sh -c "git diff trunk... | python {toxinidir}/admin/pycodestyle-twisted.py --diff"
```
twm: I suggest renaming this `pycodestyle-diff` as it will be useful outside of CI. When I run `tox -e pycodestyle` I get thousands of lines of output to sift through, so it's not much use on its own.

Additionally, for interactive use it would be useful to do `git diff trunk` instead of `git diff trunk...` so that uncommitted changes to tracked files are checked.

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -7,5 +7,6 @@ https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch
 ## Contributor Checklist:
 
 * [ ] There is an associated ticket in Trac. Create a new one at https://twistedmatrix.com/trac/newticket
+* [ ] The changes pass minimal style checks (see: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted )
```
twm, line 10: You will update the example at this link to incorporate a `pycodestyle` env?

    $ tox -e pyflakes,newsfragment,manifest-checker

admin/pycodestyle-twisted.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
+
```
twm: This is missing a copyright notice. Also shouldn't we reproduce the pycodestyle license https://github.com/PyCQA/pycodestyle/blob/master/LICENSE here in this file as `blank_lines()` was partially copied from that repository? Not that we are distributing this script in our releases, but best to overcommunicate on these legal things.

admin/pycodestyle-twisted.py
```diff
@@ -0,0 +1,110 @@
...
+# Replace the upstream blank_lines implementation, with our fork.
+_previousValue = pycodestyle._checks['logical_line'][pycodestyle.blank_lines]
+del pycodestyle._checks['logical_line'][pycodestyle.blank_lines]
+pycodestyle._checks['logical_line'][blank_lines] = _previousValue
+
+
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
```
twm: I don't understand what the purpose of this replacement is. Something to do with Windows? Can you explain?

twm (changes requested):
Hi Adi,

Overall I think that this is a good idea. Twistedchecker is flaky enough for me that I rely on Travis to run it, which really slows down getting patches into mergeable shape, so I am keen on getting a more reliable linting setup.

I did note a few things that I think should change, and a question I had. Hopefully the buildmaster will be back up soon, as I couldn't investigate the Fedora test failures without it.

Thanks for all your work on improving the Twisted development process, Adi!

.github/PULL_REQUEST_TEMPLATE
adiroiban: yes... but I hope that soon we will have a newer and faster command... as the whole tox rig is slow :)

admin/pycodestyle-twisted.py
adiroiban: done. Let me know if this is ok... otherwise feel free to update and push your suggested changes :)

admin/pycodestyle-twisted.py
adiroiban: This is just copied from upstream... I think that this is how entrypoints scripts are installed .

see all the scripts from a virtualenv bin folder.

adiroiban:
> I suggest renaming this pycodestyle-diff as it will be useful outside of CI. When I run tox -e pycodestyle I get thousands of lines of output to sift through, so it's not much use on its own.

I tried renaming before... but it looks like tox is stupid... it will also call the `pycodestyle`... we we can't use the dash (-)

See with pycodestyle-diff, before pycodestyle... as otherwise it calls pycodestyle and then pycodestyle-diff

```
$ tox -e pycodestyle-diff
GLOB sdist-make: /home/adi/chevah/twisted/setup.py
pycodestyle-diff inst-nodeps: /home/adi/chevah/twisted/build/dist/Twisted-17.9.1.dev0.zip
pycodestyle-diff installed: attrs==17.4.0,Automat==0.6.0,constantly==15.1.0,hyperlink==18.0.0,idna==2.6,incremental==17.5.0,pycodestyle==2.3.1,six==1.11.0,Twisted==17.9.1.dev0,zope.interface==4.4.3
pycodestyle-diff runtests: PYTHONHASHSEED='258055152'
pycodestyle-diff runtests: commands[0] | python -c import sys; print(sys.prefix)
/home/adi/chevah/twisted/build/pycodestyle-diff
pycodestyle-diff runtests: commands[1] | python -c import sys; print(sys.exec_prefix)
/home/adi/chevah/twisted/build/pycodestyle-diff
pycodestyle-diff runtests: commands[2] | python -c import sys; print(sys.executable)
/home/adi/chevah/twisted/build/pycodestyle-diff/bin/python
pycodestyle-diff runtests: commands[3] | python --version
Python 2.7.12
pycodestyle-diff runtests: commands[4] | /bin/sh -c git diff trunk... | python /home/adi/chevah/twisted/admin/pycodestyle-twisted.py --diff
pycodestyle-diff runtests: commands[5] | python /home/adi/chevah/twisted/admin/pycodestyle-twisted.py admin src/twisted
admin/fix-for-src-mv.py:7:22: W291 trailing whitespace
```

-------------

I went with `pycodestylediff` ... but I hope you will not have to use it :)

I was thinking that for dev you will not use tox... as it is slow... 100x slower :)... so my guess was that nobody would use tox for that and did not spend much time polishing that env.

```
$ time tox -e pycodestylediff
real	0m9.905s

$ time /bin/sh -c "git diff trunk... | python /home/adi/chevah/twisted/admin/pycodestyle-twisted.py --diff"
real	0m0.091s
```

-------

So I think that tox is somehow ok for Travis and Buildbot, but for local dev is just slow.
Maybe there is a way to speed it

rodrigc:
twistedchecker replaces the internals of pylint with a new `blank_lines` implementation.
Because twistedchecker is patching internals, it is stuck on a specific version of pylint,
and does not work on anything higher than Python 3.5.

This patch replaces the internals of pycodestyle with a `blank_lines` implementation.
This is still patching internals, but at least this is moving to a newer codebase.

At some point, it would be nice if https://github.com/PyCQA/pycodestyle/issues/732 was solved, so that pycodestyle gave an API to provide a customized `blank_lines` implementation.

rodrigc (approved):
Despite what I mentioned here: https://github.com/twisted/twisted/pull/969#issuecomment-374069472

 I think this is a good step forward.

adiroiban:
Thank @rodrigc for the review.

@twm Are you happy with the latest changes?

Not sure how we should handle reviews in GitHub. Should we ignore previous reviews and merge it as long as there is one person happy with the latest review?

---------

Yes. This is not a big step forward, other than you can now run only pycodestyle and don't need to spin twistedcker for that.

This uses the latest (unreleased) pycodestyle code.
This is somehow better as the "configurable" parts are extracted into 2 variables... so I hope it will be easier to accept upstream.

The pycodestyle API is always changing... and I don't know if it has a public API.

adiroiban:
@rodrigc I have checked the pycodestyle API ... what we can do, is to disable all E3XX erros and register our own blank_lines codes.
In this way, we are not patching internals and only touching private API.
Will see what is the feedback from pycodestyle maintainers

admin/pycodestyle-twisted.py
twm: I think that license information belongs in a comment, not a docstring, but I'm not going to be too picky about it here.

tox.ini
```diff
@@ -25,8 +25,8 @@
 ; coverage results to codecov.io. It should be called after running the
 ; standard coverage environment.
 ;
-; A non-default `txchecker-travis` environment is used to run twistedchecker
-; on travis in --diff mode
+; A non-default `txchecker-travis` and `cipycodestyle` environments are
```
twm: This comment needs to be updated for "pycodestylediff" (which is a fine name, BTW.)

twm (approved):
Hi @adiroiban,

I think the only change that must be made before merge is to fix the comment that mentions "pycodestyleci". However I suggest adding the `skip_install = true` flag to the `pycodestyle` and `pycodestylediff` testenvs. This avoids installing Twisted, which cuts 8 seconds (!) off of the time on my system. The same can also be applied to any testenv that doesn't require a Twisted install.

There is also a [`skipsdist`](https://tox.readthedocs.io/en/latest/example/general.html?highlight=skipsdist#avoiding-expensive-sdist) option, but it is unfortunately global rather than per-testenv. This is a shame as generating the sdist accounts for virtually all of the time it takes to run `tox -e pycodestylediff` on my system.

We should also [set some environment variables](https://github.com/twisted/klein/pull/191/files) to prevent Pip and virtualenv from hitting the network, but that can come in another PR.

Thanks again!

twm:
I filed https://twistedmatrix.com/trac/ticket/9392 to get those environment variables set.

admin/pycodestyle-twisted.py
adiroiban: I can do that :)

adiroiban:
Thanks a lot @twm for the great review.
I will implement your suggestions... but for me tox is still slow... so I am using a simple virtualenv for now.

I am not using tox, so my knowledge of the tox capabilities is limited.

But I think that for local dev, not using tox is fine and fast :)

twm:
@adiroiban I like to use Tox to initialize the virtualenv like so:

    tox -e py27-nodeps-nocov-posix --develop

The `--develop` flag does a `pip install -e` install of Twisted, so the virtualenv doesn't need to be recreated on every test run. Then I run the specific tests I am iterating on with a file watcher.  For example when working on #971 I ran:

    find src -name '*.py' | entr build/py27-nodeps-nocov-posix/bin/python  -m twisted.trial --reactor=default --reporter=verbose twisted.application

That subset of tests runs quickly enough for a pleasant iteration cycle.

adiroiban:
Thanks for the info. I have updated the wiki dev page with the `tox --develop` option... I am also using it in this way for py3.

It looks like sphinx documentation needs Twisted install to read the version and release notes fragment also needs it to get the script.
The plan is to merge all these scripts into a single tox environment for faster runs :)... but in a separate PR

adiroiban:
Thanks @twm and @rodrigc for the review. This is now merged.
Hope the circle-ci PR will also be merged soon as otherwise it will continue to fail...

and I see some flaky tests on Fedora

PR #970, opened 2018-03-17
9359 providing hyperlink url on request object

morgendave:
Provide web.server.request to have its hyperlink.URL as URLPath is superceded by hyperlink.URL. 
Pre Process is similar to what URLPath was doing. 
Ticket link: https://twistedmatrix.com/trac/ticket/9359


glyph:
Thanks for submitting this! This is definitely something twisted should do. Sadly don’t have time for a full review, but I do have a few pointers based on previous discussions and work:

- this should use `DecodedURL`, as that’s the interface that allows applications to store and retrieve application-data from URLs rather than quoted bytes 
- the method name shouldn’t contain the name of the library; hyperlink is obviously just the way URLs are represented in Python now 😉 
- the URL should contain *all* available information about the url the request is to, particularly including the host and scheme, probably including any basic auth as well

src/twisted/newsfragments/9359.feature
```diff
@@ -0,0 +1 @@
+Provide web.server.request to have its hyperlink.URL, as URLPath is superceded by hyperlink.URL.
```
Julian: Minor wording here: it's a bit more natural to say something like "twisted.web.server.request now provides access to its hyperlink.URL."

src/twisted/web/server.py
```diff
@@ -522,6 +522,27 @@ def URLPath(self):
         return urlpath.URLPath.fromRequest(self)
 
 
+    def hyperlinkURL(self):
```
Julian: As @glyph mentioned, I'd just call this `url`, assuming that's not already taken (I think it isn't?)

src/twisted/web/server.py
```diff
@@ -522,6 +522,27 @@ def URLPath(self):
...
+        """
+        Providing hyperlink url on request object, as URLPath is
```
Julian: The first line of a docstring should fit on exactly one line typically (both for tooling reasons and also to ensure things are well-distilled in purpose).

Here you might try `Return the URL of this request.`, and rely on someone reading the `rtype` if they want to know what type that means.

src/twisted/web/server.py
```diff
@@ -522,6 +522,27 @@ def URLPath(self):
...
+        superceded by hyperlink.URL.
+
+        @return: a hyperlink.URL object from request prePathURL
```
Julian: It's a bit simpler to just say `return: the URL` or something that leaves out more details here. That way, someone who doesn't even know what prePathURL is can still very likely understand what this method does.

src/twisted/web/test/test_web.py
```diff
@@ -865,6 +865,19 @@ def test_noDefaultContentTypeOn204Response(self):
         self.assertNotIn(b'content-type', response.lower())
 
 
+    def test_hyperlinkRequestURL(self):
+        """
+        L{Request.prePathURL} quotes special characters in the URL segments to
```
Julian: This test seems oddly specific -- I'd expect us to have at least one test that does an equality comparison directly from `request.url` to a manually constructed `hyperlink.URL` -- then if there's some additional quoting behavior that needs checking yeah additional test cases might be in order, but I think we need the simple one too.

src/twisted/web/server.py
```diff
@@ -522,6 +522,27 @@ def URLPath(self):
...
+        @rtype: L{hyperlink.URL}
+        """
+
+        # Same pre process as URLPath do.
```
Julian: If it's literally the same, maybe it's worth putting in a helper function?

Julian:
Thanks!

Looks like a good start, left a few more comments.

One more general thing: I think we might want to consider deprecating URLPath -- it would make sense to me to do that alongside this change, but if not alongside this, maybe in a new ticket.

glyph:
Let's do that in a separate ticket — smaller changes are better :)

morgendave:
Sure, updating based on the comments. For name it would just be url? Or something like schemed_url?

Julian:
That'd be what I call it yeah.

glyph:
It looks like you've accidentally checked in an Incremental `.egg` directory.  Might want to remove that commit :)

morgendave:
All updated based on the comments, the part doing commonly would remain since URLPath seems to be removed later on.

src/twisted/web/server.py
```diff
@@ -522,6 +522,26 @@ def URLPath(self):
...
+        """
+        Return URL object of this request object.
+
+        @return: the URL
+        @rtype: L{hyperlink.DecodedURL}
+        """
+
+        # Pre process to decode from bytes data
+        _allascii = b"".join([chr(x).encode('ascii') for x in range(1, 128)])
+        quoted = quote(self.prePathURL(), safe=_allascii)
```
Julian, line 535: So, this throws away the query string (and fragment) -- I don't think that's correct behavior, though that seems to be what URLPath currently does as well.

glyph: Let's not repeat that mistake :)

twm:
Two quick notes:

1. This should really be a `hyperlink.DecodedURL` rather than a `hyperlink.URL`. Then it can effectively replace `request.args` in the GET case.
2. The above is blocked by https://github.com/python-hyper/hyperlink/issues/129, where Hyperlink doesn't code `+` as space.

PR #90, opened 2016-06-02
Add onNewConnection that is called when a new connection is created.

dreid:

PR #971, opened 2018-03-18
8375 ClientService(prepareConnection=...)

twm:
A continuation of the work @dreid began in #90.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/8375
* [x] There is a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )

Closes #90.


src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,18 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
             this attribute will not be serialized, and the default value (the
             reactor) will be restored when deserialized.
         @type clock: L{IReactorTime}
+
+        @param onNewConnection: A single argument L{callable} that may return a
+            Deferred. It will be called with the L{protocol
+            <interfaces.IProtocol>} after a new connection is made.
```
adiroiban: ~should this callable return something~ ?

To rephrase the previous question.

Maybe : `A single argument L{callable} that may return a Deferred or None.`

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,18 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
...
+        @type onNewConnection: L{callable}
         """
         clock = _maybeGlobalReactor(clock)
         retryPolicy = _defaultPolicy if retryPolicy is None else retryPolicy
 
         self._machine = _ClientMachine(
             endpoint, factory, retryPolicy, clock,
-            log=self._log,
+            onNewConnection, log=self._log,
```
adiroiban: Maybe also call here using keyword argument?

src/twisted/newsfragments/8375.feature
```diff
@@ -0,0 +1 @@
+twisted.application.internet.ClientService now accepts a function to initialize a connection as the onNewConnection argument
```
adiroiban: Missing a full stop.
Is the function used to initialize the connection or is called when the connection is initialized?

src/twisted/newsfragments/8375.feature
```diff
@@ -0,0 +1 @@
+twisted.application.internet.ClientService now accepts a function which is used to initialize a new connection as the onNewConnection argument.
```
adiroiban: `is used to initialize a new connection` ... maybe 

twisted.application.internet.ClientService now accepts as the `onNewConnection` argument a function which is called when a new connection is made.

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,20 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
             this attribute will not be serialized, and the default value (the
             reactor) will be restored when deserialized.
         @type clock: L{IReactorTime}
+
+        @param onNewConnection: A single argument L{callable} that may return a
+            L{Deferred} or L{None}. It will be called with the L{protocol
+            <interfaces.IProtocol>} after a new connection is made. Should it
+            raise an exception or the L{Deferred} fail, the connection attempt
+            is treated as a failure.
+        @type onNewConnection: L{callable}
```
adiroiban: Why is this "onNewConnection" and not "onConnectionMade" ?

What does "new" mean?

I think that the docstring should be explicit that it is called for each connection (including calling it multiple times) on reconnection after a timeout.... or maybe just call onConnectionMade and this should be straightforward :)

And it would help to have a test in which we check whether it is called or not on reconnection.

src/twisted/newsfragments/8375.feature
adiroiban: `is used to initialize a new connection` ... I don't know if I get what it takes to initialize a new connection?

By reading only this I can understand that the function is in charge of creating/making that connection.

Maybe:

twisted.application.internet.ClientService now has the onNewConnection argument which accepts a function called when a connection is made.

src/twisted/application/test/test_internet.py
```diff
@@ -1031,3 +1031,78 @@ def test_stopServiceOnStoppedService(self):
 
         self.assertIsNone(self.successResultOf(firstStopDeferred))
         self.assertIsNone(self.successResultOf(secondStopDeferred))
+
+
+    def test_onNewConnectionCalledWhenServiceStarts(self):
+        """
+        The C{onNewConnection} callable is called when the
+        L{ClientService.startService} is called.
```
adiroiban: Is this true?
I was expecting onNewConnection to be called only when a connection is done.

If you call ClientService.startService and the connection fails, onNewConnection is not called.

So maybe

onNewConnection is called after ClientService.startService once the connection is made.

src/twisted/application/test/test_internet.py
```diff
@@ -1031,3 +1031,78 @@ def test_stopServiceOnStoppedService(self):
...
+        """
+        newConnections = [0]
+        def onNewConnection(_proto):
+            newConnections[0] += 1
+
+        cq, service = self.makeReconnector(onNewConnection=onNewConnection)
```
adiroiban: Since the docstring talks about startService,
either use

```
cq, service = self.makeReconnector(onNewConnection=onNewConnection, startService=True)
```

or better, explicit about  what is going on here

```
clock = Clock()
cq, service = self.makeReconnector(
    onNewConnection=onNewConnection,
    startService=False,
    fireImmediatly=False,
    clock=clock)

service.startService()

# The initial connection fails.
cq.connectQueue[0].errback(Exception('failed-connection'))

# Not yet called.
self.assertEqual(0, newConnections[0])

# The connection is made on the retry.
clock.advance(AT_LEAST_ONE_ATTEMPT)
cq.connectQueue[1].callback(None)

self.assertEqual(1, newConnections[0])
```

src/twisted/application/test/test_internet.py
```diff
@@ -1031,3 +1031,78 @@ def test_stopServiceOnStoppedService(self):
...
+
+
+    def test_onNewConnectionCalledWhenServiceStarts(self):
+        """
+        The C{onNewConnection} callable is called when the
+        L{ClientService.startService} is called.
+        """
+        newConnections = [0]
+        def onNewConnection(_proto):
+            newConnections[0] += 1
```
adiroiban: It would also help to have a test which checks that whatever onNewConnection returns, in the end, its result is ignored and you get the original protocol

```
newConnectionCalls = []
def onNewConnection(protocol):
    newConnectionCalls.append(protocol)
    return object()

then in the test

self.assertEqual(1, len(newConnectionCalls))
self.assertIs(newConnectionCalls[0], cq.constructedProtocols[0])
```

adiroiban (dismissed):
Thanks for working on this.

Please check my inline comments.

The  summary:

* onNewConnection vs onConnectionMade
* Have an integration or functional test to also check the condition in which it is not called (ex on connection failure) :)
* Document and test that the result from onNewConnection is ignored an if it want to make a change, it should trigger side effects

twm:
Thank you for the review, @adiroiban! I think that I have addressed all of the feedback:

* I renamed `onNewConnection` to `prepareConnection`. This is more suggestive of the intended purpose of the callback. Dropping the "on" prefix seems more in keeping with the naming of things in Twisted in general, and without the prefix there is no implication that this is merely an observer. The suggestion of "onConnectionMade" would seem to imply the same semantics as IProtocol.connectionMade, but the semantics of this callback are quite different (in particular it returns a Deferred).
* There is enough flexibility inherent in the API design here that I don't think that we can come up with a succinct name which conveys the full semantics of the parameter. We'll have to make up for this with documentation, so I expanded the docstring of the parameter to better describe the ways it interacts with the reconnection loop and clarify that it happens before a `whenConnected()` Deferred resolves.
* I added several test cases suggested by @adiroiban.

The build is failing on the RHEL 7 buildbot worker with a failure to download the appdirs distribution. There are similar failures on this worker for other PRs for different distributions (e.g., #1003 is failing to install pyasn1-modules), so I am marking this for review despite the failure.

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,31 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
             this attribute will not be serialized, and the default value (the
             reactor) will be restored when deserialized.
         @type clock: L{IReactorTime}
+
+        @param prepareConnection: A single argument L{callable} that may return
```
glyph, line 1062: This is the one blocker:  the `prepareConnection` parameter needs a "since" declaration explaining when it's available since.  See https://github.com/twisted/incremental#updating for an explanation about how to phrase this so it will be picked up by the release machinery.

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,31 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
...
+            a L{Deferred}. It will be called once with the L{protocol
```
glyph, line 1063: Some non-blocking, but useful stuff:

- What should this `Deferred` fire with?
- If it can be anything (which is implied) will its callback type be preserved?
- Will errors be handled?  Logged?

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,31 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
...
+            <interfaces.IProtocol>} each time a new connection is made.  It may
+            call methods on the protocol to prepare it for use (e.g.
+            authenticate) or validate it (check its health).
+
+            The C{prepareConnection} callable may raise an exception or return
+            a L{Deferred} which fails to reject the connection.  A rejected
```
glyph, line 1069: Any particular exception type it's supposed to use to reject the connection "cleanly"?

src/twisted/application/internet.py
```diff
@@ -1029,13 +1058,31 @@ def __init__(self, endpoint, factory, retryPolicy=None, clock=None):
...
+            connection is not used to resolve an L{Deferred} returned by
```
glyph: Deferreds are generally said to be "fired" rather than "resolved" (which is the language one uses with JavaScript Promises). This isn't a dealbreaker, use this language if you prefer it, it's just a little inconsistent with the rest of Twisted.

glyph (approved):
Please address the one blocker here, documenting the version of the addition, and then I'd say this is approved to land.  (~~I don't see the little "member" flag next to your name, perhaps we should give you commit, but if you need a project member to land it, please just indicate this is pre-approved and doesn't need another review round.~~ looks like I was looking at the wrong place in the UI, land away :))

twm:
@glyph Thank you for the review! I have adjusted the terminology of the docstrings per your suggestion. I also submitted #1020 to better-document the syntax for `@since`. I will merge as soon as we have a solution to the Travis sendmsg test issues.

glyph:
@twm My understanding of the Travis issue is that it's unrelated. If it is actually caused by this change then by all means let's figure it out, of course.

Since it looked intermittent and unrelated to me though, I've re-run the builder and if it succeeds this time let's fix it separately.

glyph:
Bleh, I forgot that Travis doesn't save old logs before hitting the rebuild button. In any case, looks like that issue was unrelated, so please merge away.

twm:
@glyph I suppose the fourth time's the charm. :) I had given up mashing the button as it seemed to be failing consistently. Thanks again!

glyph:
When this sort of thing happens repeatedly on Travis, sometimes it's that we get unlucky and put onto a machine with a broken kernel or hardware configuration; weird APIs like `sendmsg` can be particularly sensitive to bad kernel versions, incorrect tuning, or overly strict container configuration.  Eventually they fix the issue and the intermittent failure stops.

PR #972, opened 2018-03-19
9391 Run macOS tests and static checkers on CircleCI.

adiroiban:
Scope
====

https://twistedmatrix.com/trac/ticket/9391

This tries to enable macOS builds on Circle-CI so that for external PR we have tests exected on Linux (Travis-CI) / Windows (Appveyor) and macOS (Circle-CI)

As a drive-by, while spinning the Circle-CI it also enabled the static checkers on Circle-CI and removes them from Travis. In this way we should have faster builds.

On Circle-CI we should have 4 parallel builds on docker. 

Changes
=======

Added an initial circle-ci config for Python 2.7 with coverage.

You can visualize here the workflow https://circleci.com/workflow-run/5e07fd3e-97a7-4316-9760-0d816dba6738

For now, I paid for 1 month of Seed macOS plan.

I have invited all "teammates" with email from https://circleci.com/team/gh/twisted

I sent an email to Circle-ci asking for a free ride :) ... will see how it gets.
Otherwise, I am ok to sponsor the seed plan.

I have moved some static checkers from Travis-CI to Circle-CI to reduce the load on Travis.

Also to reduce the number of builds on Circle-CI on macOS, the macOS builds are triggered only after the static checkers pass.
I know that is not idea, but just by working on this PR I have consumed 10% of the macOS time.

As a drive by, I have removed the installation of pydoctor from the test tox env as it should not be needed.
Only the env which builds the apidocs should need it.

How to check
===========

Check that the commit status and their names.

Check that the Circle-CI workflow makes sense. The macOS build is started only after static checkers.
So the macOS build is delayed by 7 minutes...but the Appveyor and Travis are also late... and we need to save macOS time.

For now, we have 500min per month which translates in about 100 builds per month.
At this point, I would not add more builds to Circle-CI...so no Python3 on macOS.

I see that as I run more builds... they are slower... last one was 6:50...while there were also builds which took 4:20 

I already used 6% of the quota by testing this PR :)

The test output is big... not sure if we should run in quiet mode... or produce XML test results.

Once this is merge, we can mark the Circle-CI as required to replace buildbot/osx10.10-py2.7


.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+      xcode: "9.0"
+
+    working_directory: ~/repo
+
+    steps:
+      # Get the source.
+      - checkout
+
+      - run:
+          name: prepare environment
```
markrwilliams: Can this say "Prepare macOS environment" so there's some indication of the build's platform on its page?

markrwilliams (approved):
On the one hand, CircleCI's build summary is harder to read than Travis' and Buildbots'.

On the other hand, I think [the CircleCI builds will fail meaninglessly until some configuration is merged.](https://circleci.com/gh/twisted/twisted/12#config/containers/0).

At worst merging this configuration will have no effect, so I'm :+1: for it.   Please make the build's more informative before merging.

I'm :-1: on removing the macOS builder until we see how CircleCI performs.

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
+# CircleCI 2.0 configuration file for running Python on macOS.
```
markrwilliams, line 1: Can we also disable email notifications of failed builds?

adiroiban: I have disabled them from the web UI. I don't know how todo it from config.

.circleci/config.yml
adiroiban: Sure :)...I am  testing a bit with Circle-CI to see if I can control the way commit status is reported

adiroiban:
@markrwilliams Thanks for your review.

Please give it a completely new review as I have changed a lot.

We now have workflow https://circleci.com/workflow-run/a6e86196-29f2-4185-851b-087cf961dd5d
Some travis-ci builds were moved to Circle-CI.

I have also talked with @exarkun over IRC and it looks like stopping builds due to pyflakes (and other errror) is a good thing.

```
(12:14:27 PM) exarkun: If the static checkers actually catch problems that cause builds to fail, that seems like a good idea
(12:14:52 PM) exarkun: Or, well
(12:15:05 PM) exarkun: I guess there are two goals and they conflict with each other so one has to win and the other has to lose (afaict)
(12:15:22 PM) exarkun: Actually, one goal and one constraint that briefly appeared to be a goal
(12:15:54 PM) exarkun: One goal is fast builds.  The constraint is 500 minutes of time per month (suggesting a goal of minimizing "stupidly" failing builds).
(12:15:55 PM) adiroiban: well.. some pyflakes error will cause the tests to fail... but not all :)
(12:16:20 PM) exarkun: Yea.  Let's just call it half.  If you forget to define a name that's a "stupid" build to run on CI.
(12:16:29 PM) adiroiban: yes :)
(12:16:47 PM) exarkun: If you forget to clean up an unused import, is that a "stupid" build?  Hmm.  You're gonna have to run another build for the fix for that... So you could argue it is a stupid build.
(12:16:47 PM) adiroiban: and we should have the same rule for travis/appevor/buildbot
(12:17:00 PM) adiroiban: since these are executed for any PR... I guess that we can expect anything to be pushed
(12:17:14 PM) exarkun: I like circleci's workflows features.  I guess travis has "staged builds" or something now that's similar but I haven't used it yet.
(12:17:26 PM) adiroiban: yes... travis has stages
(12:17:36 PM) exarkun: Okay I guess it makes sense
(12:17:48 PM) exarkun: Run static checks before you push or you're wasting CI resources
(12:17:54 PM) adiroiban: now... I need to see how to configure the paralelization in circle-Ci
(12:17:58 PM) exarkun: CI should punish you if you forget because punishment is the best
(12:18:14 PM) exarkun: er I mean because then you'll learn to run static checks before you push and save yourself time and save the project resources
```

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+      - run:
+          name: Check pyflakes.
+          command: |
+            tox -r -e pyflakes
+
+
+  #
+  # Documentation tests are slow so we execute them in a separate job.
+  #
+  documentation:
```
markrwilliams, line 61: Why build documentation in CircleCI?  Won't that just burn through our allotted time faster?

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+      - run:
+          name: Check the API documentation.
+          command: |
+            tox -r -e apidocs
+
+
+  #
+  # We run pyflakes with Python 3 .
+  #
+  pyflakes3:
```
markrwilliams, line 91: Shouldn't this also be a static check, like `pyflakes` under Python 2?

.travis.yml
```diff
@@ -33,26 +33,13 @@ matrix:
       env: TOXENV=py35-alldeps-withcov-posix,codecov-publish
     - python: 3.6
       env: TOXENV=py36-alldeps-withcov-posix,codecov-publish
-    # For now all non-trial tests are in a single job to reduce the time spent
-    # on starting separate jobs.
-    - python: 2.7
-      env: TOXENV=narrativedocs,apidocs,pyflakes,newsfragment,manifest-checker
```
markrwilliams: I'm not sure it makes sense to move these checks off Travis to CircleCI when the latter imposes a lower usage limit.  If we don't pay CircleCI (or get some free usage from them!) then we won't build any of these linting steps?

markrwilliams (dismissed):

.circleci/config.yml
adiroiban: This is for Linux ... not the macOS and on CIrcle-CI we get 4 concurent builds.. and on Travis we get 5 AFAIK.

So since we have 4 "worked" available on Circle-CI I was thinking to use them here to reduce the load on Travis-CI.

Does that make sense?

I will try to create a documentation explaining how we run the automated tests.

.circleci/config.yml
adiroiban: I don't understand the question. This is a separate job as it needs a separate docker image.

.travis.yml
adiroiban: The limit is only for macOS builds. For Linux, we got 4 concurent linux container which can be used.

So I thought that is a good think to make use of them.

And with this workflow, if the static checker fails, we don't spin the macOS and save some macOS build time.

.circleci/config.yml
adiroiban: Yes... This could have been a pre-requisite for the macOS build... but I thought that the python2 pyflakes should be enough ... so pyflakes3 and documentation builds are executed after that, to not delay the macOS any longer.

This is a compromise

adiroiban: In the same times, this needs python3, so it would run as a separate job anyway.

In theory, we can create a python2+python3 docker image and run the static checkers with that image...but at this stage, I prefer to use the Circle-CI Docker images.

adiroiban:
Thanks @markrwilliams for the review.

While the trial tests are a bit harder to read on Circle-Ci, I think that static checkers tests are easier to read on Circle-CI (the way is set up now).

I also like that on Circle-CI we can get a separate a status for each build, so you can go directly to its page.
On travis, you see all jobs, and then click again to get the details for the failing build.

Please give it another try :)

.circleci/config.yml
markrwilliams: So the time limit you (graciously) paid for only applies to the Mac OS builds?

.circleci/config.yml
markrwilliams: I guess I don't understand the pricing plan for CircleCI; if this doesn't gate the macOS build, it seems like it will waste money.  But it looks like you're only running the tests under Python 2.7 on macOS, so this is OK if it runs under Linux and doesn't spend any of the time budget.

.circleci/config.yml
markrwilliams: I'm still getting notification emails :disappointed:

tox.ini
```diff
@@ -97,7 +97,6 @@ commands =
     python --version
 
     ; Install PyDoctor here so it DOESNT overwrite Twisted
-    py27-{alldeps,tests}: pip install --no-deps epydoc pydoctor
```
markrwilliams, line 100: Why is this gone?

markrwilliams (approved):
This looks good!  The statuses on this PR are much more informative.

There's a change I don't understand in the `tox.ini`.  Please let me know what's up with that and stop the emails (please!) before merging.

tox.ini
adiroiban: see the description of this pr

> As a drive by, I have removed the installation of pydoctor from the test tox env as it should not be needed.
> Only the env which builds the apidocs should need it.

It was removed as it was just wasting time installing pydoctor when not needed

adiroiban:
> so this is OK if it runs under Linux and doesn't spend any of the time budget.
The static checkers run on Linux dockers, and they use  4 separate dockers.

In Circle-CI you have to pay for 2 parallel plans. Linux Docker and macOS

--------

Now I don't know why I don't see the Circle-Ci statuses for the latest commit :)

adiroiban:
> Now I don't know why I don't see the Circle-Ci statuses for the latest commit :)

It was disabled by @glyph 
Unless this branch is merged, Circle-CI is smart and will try to run with some default steps..and I don't know how to disable that

PR #974, opened 2018-03-21
[#9393] Make the threadpool tests pass on pypy

hawkowl:
https://twistedmatrix.com/trac/ticket/9393

### DIFF
--- a/src/twisted/test/test_threadpool.py
+++ b/src/twisted/test/test_threadpool.py
@@ -172,9 +172,6 @@ class ThreadPoolTests(unittest.SynchronousTestCase):
         # Sanity check - no threads should have been started yet.
         self.assertEqual(tp.threads, [])
 
-        # this holds references obtained in onResult
-        refdict = {} # name -> ref value
-
         onResultWait = threading.Event()
         onResultDone = threading.Event()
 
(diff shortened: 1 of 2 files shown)


src/twisted/test/test_threadpool.py
```diff
@@ -182,6 +182,9 @@ def test_threadCreationArgumentsCallInThreadWithCallback(self):
 
         # result callback
         def onResult(success, result):
+            # Spin the GC, which should now delete worker and unique if it's
```
markrwilliams, line 191: This is fantastic.  For extra credit, mention that `gc.collect()` has to happen here because the relevant objects are still in the thread's queue.

src/twisted/test/test_threadpool.py
```diff
@@ -227,6 +230,9 @@ class Dumb(object):
         self.assertIsNone(onResultRef())
         self.assertIsNone(resultRef[0]())
 
+        # The callback shouldn't have been able to resolve the references.
```
markrwilliams, line 239: :+1:

markrwilliams (approved):
On the one hand tests that exercise collection semantics are of dubious value on PyPy; on the other hand testing for memory leaks, especially with threads, is valuable.

I agree that this captures the intent of test.  Please consider expanding the already useful comment before merging.

PyPy 3 here we come!

markrwilliams:
Also :1st_place_medal: for this fix.

PR #975, opened 2018-03-23
8241 ipv6 client ip

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. http://twistedmatrix.com/trac/ticket/8241

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1386,7 +1386,7 @@ class Request:
         @returns: the client IP address
         @rtype: C{str}
         """
-        if isinstance(self.client, address.IPv4Address):
+        if isinstance(self.client, (address.IPv4Address, address.IPv6Address)):
             return self.client.host
         else:
             return None
--- a/src/twisted/newsfragments/7704.bugfix
+++ b/src/twisted/newsfragments/7704.bugfix
@@ -1,0 +1,1 @@
+twisted.web.http.Request.getClientIP now returns the host part of the client's address when connected over IPv6.
--- a/src/twisted/newsfragments/8241.bugfix
+++ b/src/twisted/newsfragments/8241.bugfix
@@ -1,0 +1,1 @@
+twisted.web.http.Request.getClientIP now returns the host part of the client's address when connected over IPv6.
--- a/src/twisted/newsfragments/8241.bugfix
+++ b/src/twisted/newsfragments/8241.bugfix
@@ -1,1 +1,1 @@
-twisted.web.http.Request.getClientIP now returns the host part of the client's address when connected over IPv6.
+twisted.web.wsgi.WSGIResource no longer raises an exception when a client connects over IPv6.
--- a/src/twisted/web/test/requesthelper.py
+++ b/src/twisted/web/test/requesthelper.py
@@ -31,12 +31,15 @@ class DummyChannel:
         port = 80
         disconnected = False
 
-        def __init__(self):
+        def __init__(self, peer=None):
+            if peer is None:
+                peer = IPv4Address("TCP", '192.168.1.1', 12344)
+            self._peer = peer
             self.written = BytesIO()
             self.producers = []
 
         def getPeer(self):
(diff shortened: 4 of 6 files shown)


markrwilliams:
Also http://twistedmatrix.com/trac/ticket/7704

glyph:
(I think) I've disabled CircleCI status updates because the configuration is completely broken.  @adiroiban If you'd like to turn it back on please fix it so it doesn't fail every build :).

src/twisted/web/http.py
```diff
@@ -1386,7 +1386,7 @@ def getClientIP(self):
         @returns: the client IP address
         @rtype: C{str}
         """
-        if isinstance(self.client, address.IPv4Address):
+        if isinstance(self.client, (address.IPv4Address, address.IPv6Address)):
```
adiroiban, line 1389: This is fine... but maybe we should use  `_IPAddress` here... I don't know which is better.

src/twisted/web/http.py
```diff
@@ -1386,7 +1386,7 @@ def getClientIP(self):
...
             return self.client.host
         else:
             return None
```
adiroiban, line 1392: I know that this is existing code, but when is expected to return `None`.

The docstring is not clear.

Can this ever return `None`? In which case? Is there a test for this?

The code coverate, reports that this branch is executed  but I could not found the explicit tests https://codecov.io/gh/twisted/twisted/src/89e71172c2e4b2f923d3f3fa49922d50262394df/src/twisted/web/http.py#L1392

I saw that it is called in twisted.web.test.test_web.test_etagMatched due to a log which is trigerred.

Can we have a test describing why we have this execution branch?

adiroiban (approved):
Thanks for working on this.

It looks much better than trunk :)

Is there a test for the non IPv6 or IPv4 execution branch?

Can we have HTTP requests over non-TCP/IP stack?

I don't know why the OSX failed on installing tox.
I saw that OSX pass after this on a separate branch.
See if they still fail after merging with trunk.

Thanks!

src/twisted/web/http.py
markrwilliams: > Can this ever return None?

It absolutely can!  Try this:

```twist web --port=unix:/tmp/sock --wsgi wsgiref.simple_server.demo_app```

Connect to it like so:

```curl --unix-socket /tmp/sock http://example.invalid/```

Now you'll get a traceback from ```twisted.web```:

```
...
          File "/home/mrw/src/python/twisted/src/twisted/web/wsgi.py", line 303, in __init__
            'REMOTE_ADDR': _wsgiString(request.getClientIP()),
          File "/home/mrw/src/python/twisted/src/twisted/web/wsgi.py", line 85, in _wsgiString
            return string.decode("iso-8859-1")
        builtins.AttributeError: 'NoneType' object has no attribute 'decode'
```

This is mentioned in [a comment on 7704](https://twistedmatrix.com/trac/ticket/7704#comment:2), which was linked to from [this ticket](https://twistedmatrix.com/trac/ticket/8241) and also captured in [5406](https://twistedmatrix.com/trac/ticket/5406).

adiroiban: Great work :) Feel free to merge this as it is. Having a test for this is optional and I hope it will be done in the follow-up ticket

markrwilliams:
If this is merged, its `Fixes` should include two tickets: https://twistedmatrix.com/trac/ticket/7704 and https://twistedmatrix.com/trac/ticket/8241

src/twisted/web/http.py
markrwilliams: I'm reluctant to use a private implementation even though this is inside Twisted.

markrwilliams:
@rodrigc This was in the review queue - be sure to remove it to get review credit.

markrwilliams:
Also, per https://github.com/twisted/twisted/pull/975#issuecomment-375925298, this needs to close 7704 as well.

PR #976, opened 2018-03-24
docs: render deferToThread example as code

ktdreyer:
Prior to this change, the documentation was not rendering this example as a code block. The documentation simply displayed indented text here.

You can see this here: http://twistedmatrix.com/documents/current/core/howto/threading.html#getting-results

Add a double-colon to signify that this is code.

https://twistedmatrix.com/trac/ticket/9396

### DIFF
--- a/docs/core/howto/threading.rst
+++ b/docs/core/howto/threading.rst
@@ -99,6 +99,8 @@ When we run some code, we often want to know what its result was.  For this, Twi
 
 To get a result from some blocking code back into the reactor thread, we can use :api:`twisted.internet.threads.deferToThread <deferToThread>` to execute it instead of callFromThread.
 
+::
+
     from __future__ import print_function
     from twisted.internet import reactor, threads
 


exarkun:
Thanks.  This looks like a good change.  Can you please follow the development process steps described in the PR template?  Thanks.

ktdreyer:
@exarkun thanks for reviewing. I've filed https://twistedmatrix.com/trac/ticket/9396 to track this contribution.

markrwilliams:
@ktdreyer Please add a [newsfragment](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles).  Thanks for improving our documentation!

ktdreyer:
@markrwilliams I've added one. thanks for cluing me in :)

rodrigc (approved):
Thanks for your submission, and thanks for your patience following the Twisted process.

Please have a quick read of http://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch , which is included in the pull request template.

There are a lot of steps there, and it may seem a bit annoying, but honestly, it's not that bad. :)
If you go through that, it makes the review and merge cycle go faster. :)

Thanks again for your submission!

ktdreyer:
I think I see what's happening. My challenge here is that I used [hub](https://github.com/github/hub) to submit my PRs, and I need to run `hub` with some arguments, specifically `hub pull-request -F .github/PULL_REQUEST_TEMPLATE --edit`. Would you consider editing that wiki page to guide contributors who use hub?

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9396
* ~I have updated the automated tests.~ (not applicable)

glyph:
> Would you consider editing that wiki page to guide contributors who use hub?

Thanks for the explanation!  It seems unlikely that folks would manage to find that wiki page, though.  My contention would be that this is a bug in `hub` itself, and it should just be fixed there.  (The whole point of these special files is that they're supposed to be displayed before you submit a PR!)

Would you mind filing such a bug?

ktdreyer:
@glyph I completely agree with your reasoning that this should be fixed in `hub`. Looks like it was added in https://github.com/github/hub/pull/1187 , and it will be available to users in `hub` version 2.3.0.

PR #980, opened 2018-03-25
Do not crash if we try to run wsgi over a Unix socket

rodrigc:
https://twistedmatrix.com/trac/ticket/5406

I tested this patch by using this testcase mentioned by @markrwilliams here: https://github.com/twisted/twisted/pull/975#discussion_r176917494

Start a server with:

```
twist web --port=unix:/tmp/sock --wsgi wsgiref.simple_server.demo_app
```

Connect to it like so:

```
curl --unix-socket /tmp/sock http://example.invalid/
```

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1333,7 +1333,11 @@ class Request:
         host = self.getHeader(b'host')
         if host:
             return host.split(b':', 1)[0]
-        return networkString(self.getHost().host)
+        hostAddress = self.getHost()
+        if isinstance(hostAddress, address.UNIXAddress):
+            return b"unix-socket"
+        else:
+            return networkString(self.getHost().host)
 
 
     def getHost(self):
--- a/src/twisted/web/newsfragments/5304.bugfix
+++ b/src/twisted/web/newsfragments/5304.bugfix
@@ -1,0 +1,1 @@
+Running a wsgi server over a Unix socket now works.  This can be tested with: twist web --port=unix:/tmp/sock  --wsgi wsgiref.simple_server.demo_app
--- a/src/twisted/web/wsgi.py
+++ b/src/twisted/web/wsgi.py
@@ -299,7 +299,6 @@ class _WSGIResponse:
         # *both* Python 2 and Python 3, so says PEP-3333.
         self.environ = {
             'REQUEST_METHOD': _wsgiString(request.method),
-            'REMOTE_ADDR': _wsgiString(request.getClientAddress().host),
             'SCRIPT_NAME': _wsgiString(scriptName),
             'PATH_INFO': _wsgiString(pathInfo),
             'QUERY_STRING': _wsgiString(queryString),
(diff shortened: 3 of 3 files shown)


rodrigc:
@ewdurbin you mentioned here: https://twitter.com/EWDurbin/status/976616064170758145

that you would like to run `twistd` over wsgi for https://pypi.org

If you have time, can you quickly try out this patch by installing in a virtualenv directly from this branch by doing:

```
python -m pip install git+https://github.com/twisted/twisted.git@5406-rodrigc-wsgi-unix-socket
```

and provide feedback?

Thanks.

rodrigc:
@EWDurbin if you install this branch, you will also get a fix for https://twistedmatrix.com/trac/ticket/7704 done by @markrwilliams in https://github.com/twisted/twisted/pull/975

rodrigc:
In ticket:5406, @exarkun referred to this thread https://mail.python.org/pipermail/web-sig/2012-January/005097.html .  The WSGI standard doesn't take into account Unix sockets, so other servers like gunicorn and CherryPy deviate from the standard to work with Unix sockets.

ewdurbin:
@rodrigc thank you! i'll get a canary of this out as soon as I determine how to get pip-compile to allow it :-p

markrwilliams:
This will [break](https://github.com/django/django/blob/master/django/http/request.py#L113) [Django](https://github.com/django/django/blob/master/django/http/request.py#L82).  We need a [different approach](https://twistedmatrix.com/trac/ticket/5406#comment:10).

rodrigc:
@markrwilliams actually, I tried my patch and it did not break Django.

I created a very simple django project called mysite, then started it with:

```
twist web --port=unix:/tmp/sock --wsgi mysite.wsgi.application
```

Then I tried to access it via:

```
curl --unix-socket /tmp/sock http://example.invalid/
```

At first, Django returned an error message:

```
Invalid HTTP_HOST header: 'example.invalid'. You may need to add 'examle.invalid' to ALLOWED_HOSTS.
```

But then in my Django project, I added to mysite/settings.py:

```
ALLOWED_HOSTS = ['example.invalid']
```

I then re-ran the curl command, and Django successfully returned the web page.

Is there some other scenario you are referring to where you would expect Django to fail?

markrwilliams:
Per the links I shared, and as explained in the Trac ticket I linked to, there are two cases that fail:

1. [HttpRequest.get_port](https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpRequest.get_port) for all HTTP requests;
2. [HttpRequest.get_host](https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpRequest.get_host) for HTTP 1.0 requests without a `Host` header.

Here's a reproducer:

https://github.com/markrwilliams/django-demo

markrwilliams:
Here's an incomplete branch that works:

https://github.com/twisted/twisted/compare/5406-wsgi-unix-socket

It needs https://twistedmatrix.com/trac/ticket/9400 and https://twistedmatrix.com/trac/ticket/9403 before it can be completed.

Here's a branch that addresses both of those:

https://github.com/twisted/twisted/compare/9400-getRequestHostname-exception

rodrigc:
@markrwilliams thanks for the Django test case.  I modified my patch, and ran it against
your Django testcase and it works.

glyph (changes requested):
Much as I'd like to land this soon, this branch introduces new behavior but doesn't introduce tests to cover it, so we do need to block on that.

I really like the implementation strategy over on https://github.com/twisted/twisted/compare/5406-wsgi-unix-socket since it includes both unit tests and a referentially transparent `_extractAddresses` with an expansive docstring (although it would be nice if it included even more of the information from https://twistedmatrix.com/trac/ticket/5406#comment:10 since that context is useful).  If it could be fixed up to pass tests, that would be great.

PR #981, opened 2018-03-25
7705 deprecate get client ip

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/7705 and https://twistedmatrix.com/trac/ticket/7707
* [x] I have created newsfragments


src/twisted/web/_auth/digest.py
```diff
@@ -41,7 +41,7 @@ def getChallenge(self, request):
         @return: The L{dict} that can be used to generate a WWW-Authenticate
             header.
         """
-        return self.digest.getChallenge(request.getClientIP())
+        return self.digest.getChallenge(request.getClientAddress().host)
```
glyph, line 44: So, here's the crux of the problem with this change:

`IAddress` promises *nothing*.  This implementation therefore breaks everything but ipv4 and ipv6 addresses - in particular, there's now nothing that can be done to make this work for UNIX peer addresses.

I'm not sure I have a good solution here, since this ambiguity is exactly why `getClientIP` existed in the first place.

markrwilliams: It's true that this PR breaks on UNIX peers just like the code it replaces.  However, truly nothing could be done about them with `getClientIP` which returns `None` for non-IP addresses.

A subsequent PR can at least now use `isinstance` to check if the peer's a UNIX socket, passing a suitable value in its place if it is.  In this case suitable probably means "random value" or even None, because [`getChallenge` passes its argument to `_generateOpaque`](https://github.com/twisted/twisted/blob/d23799b6f642fa8efecc651d9ddb7e5c57bf6ea6/src/twisted/cred/credentials.py#L240), which a) returns a value that by definition is only intelligible to the server and not any client, and b) [can already accept a falsey address](https://github.com/twisted/twisted/blob/d23799b6f642fa8efecc651d9ddb7e5c57bf6ea6/src/twisted/cred/credentials.py#L276-L277).  

I suspect all the places getClientIP was used have similarly lenient requirements, so ad-hoc `isinstance` checks will suffice to support UNIX peers.

markrwilliams: [Here's how `getClientAddress` could be used to support WSGI over UNIX domain sockets](https://github.com/twisted/twisted/compare/5406-wsgi-unix-socket#diff-5535355bbab0934c4750cbedcd92fac7R238).

See also https://github.com/twisted/twisted/compare/9400-getRequestHostname-exception.

tardyp:
Hello,
Please note that this PR breaks buildbot unit tests, which enforce that no warning is raised from the tests.
There is still an instance of getClientIP() here: https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http.py#L2623

PR #982, opened 2018-03-27
[#9399] Remove a flaky test that tests something we don't actually want to do

hawkowl:
https://twistedmatrix.com/trac/ticket/9399


markrwilliams (approved):
It'd be nice to have a replacement test, but I doubt it's worth the increased implementation complexity.

PR #983, opened 2018-03-27
[#9398] Fix a flaky test causing test failures later on pypy

hawkowl:
https://twistedmatrix.com/trac/ticket/9398


src/twisted/names/root.py
```diff
@@ -322,7 +322,7 @@ def bootstrap(resolver, resolverFactory=None):
     """
     domains = [chr(ord('a') + i) for i in range(13)]
     L = [resolver.getHostByName('%s.root-servers.net' % d) for d in domains]
-    d = defer.DeferredList(L)
+    d = defer.DeferredList(L, consumeErrors=True)
```
markrwilliams, line 325: I think this is right, because nothing can add errbacks to any of the Deferreds in `L` without using private APIs.

markrwilliams (approved):
:+1:

PR #985, opened 2018-04-01
[9404] Use only socket.AI_NUMERICHOST in t.i.tcp._resolveIPv6.

adiroiban:
Scope
=====

See https://twistedmatrix.com/trac/ticket/9404

In Solaris (10 and 11) you can't use listenTCP with IPv6 as it fails to get the address info.

This tries to fix that :)

Changes
=======

I have removed the `AI_NUMERICSERV` flag used in _resolveIPv6 since the port is never specified as a service name.
`twisted.internet.interfaces.IReactorTCP` ask for a port number and does not document that service names are supported.

I have emulated the AI_NUMERICSERV... not sure if that is ok, instead of an if/else to check for Solaris.

I have updated pycodestyle to ignore the top module imports, as we have many if/else imports so this will create false positives.

Contributor Checklist
=================

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9404
* [x] The changes pass minimal style checks.
* [x] I have created a newsfragment in src/twisted/newsfragments
* [ ]  Automated tests were not updated as I expect that this refactoring don't need extra tests.

How to test
=========

This code should continue to work and now it also works on Solaris.

```
from twisted.internet.protocol import Factory
reactor.listenTCP(12345, Factory(), interface='::1')
```

### DIFF
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -631,6 +631,15 @@ def _resolveIPv6(ip, port):
     @raise socket.gaierror: if either the IP or port is not numeric as it
         should be.
     """
+    try:
+        int(port)
+    except ValueError:
+        # This raises a similar error to the error raised when using
+        # the socket.AI_NUMERICSERV flag.
+        # The socket.AI_NUMERICSERV flag is not used as it fails on Solaris
+        # even when a numeric port is provided.
+        raise socket.gaierror(-2, 'Name or service not known')
+
     return socket.getaddrinfo(
         ip,
         port,
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -638,7 +638,7 @@ def _resolveIPv6(ip, port):
         # the socket.AI_NUMERICSERV flag.
         # The socket.AI_NUMERICSERV flag is not used as it fails on Solaris
         # even when a numeric port is provided.
-        raise socket.gaierror(-2, 'Name or service not known')
+        raise socket.gaierror(socket.EAI_NONAME, 'Name or service not known')
 
     return socket.getaddrinfo(
         ip,
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -89,9 +89,6 @@ from twisted.internet.error import CannotListenError
 from twisted.internet import abstract, main, interfaces, error
 from twisted.internet.protocol import Protocol
 
-# Not all platforms have, or support, this flag.
-_AI_NUMERICSERV = getattr(socket, "AI_NUMERICSERV", 0)
-
 
 # The type for service names passed to socket.getservbyname:
 _portNameType = (str, unicode)
@@ -613,8 +610,6 @@ class BaseClient(_BaseBaseClient, _TLSClientMixin, Connection):
 
 
 
-_NUMERIC_ONLY = socket.AI_NUMERICHOST | _AI_NUMERICSERV
-
 def _resolveIPv6(ip, port):
     """
     Resolve an IPv6 literal into an IPv6 address.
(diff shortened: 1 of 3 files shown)


adiroiban:
This needs more work... I now see that there is a test which checks that service names are supported twisted.internet.test.test_tcp.SimpleUtilityTests.test_resolveNumericService

adiroiban:
The fix is on the stupid side, but I prefer it than check for Solaris platform.
I am happy to implement it in any other way.
Since Solaris 10/11 is not a supported platform, and this is due to a bug in Solaris I am ok if this is not accepted.

The OSX10.10 slave is failing due to an unrelated general issue.

Thanks!

adiroiban:
Maybe, before merging this, we should have #986 merge, as that one is more important.

src/twisted/internet/tcp.py
```diff
@@ -636,7 +631,23 @@ def _resolveIPv6(ip, port):
...
     """
-    return socket.getaddrinfo(ip, port, 0, 0, 0, _NUMERIC_ONLY)[0][4]
+    try:
+        int(port)
+    except ValueError:
+        # This raises a similar error to the error raised when using
+        # the socket.AI_NUMERICSERV flag.
+        # The socket.AI_NUMERICSERV flag is not used as it fails on Solaris
+        # even when a numeric port is provided.
+        raise socket.gaierror(socket.EAI_NONAME, 'Name or service not known')
```
markrwilliams, line 641: `EAI_NONAME` isn't how `socket.getaddrinfo` fails with glibc 2.26 and a service name instead of a port:

```
>>> import socket
>>> socket.getaddrinfo('::1', 'http', 0, 0, socket.AI_NUMERICSERV)
socket.gaierror: [Errno -8] Servname not supported for ai_socktype
>>> socket.EAI_SERVICE
-8
```

I'm certain that the actual errno for this varies by platform and C library version, to the point where attempting to come up with appropriate failures is not worth the time and effort.

As you point out, though, `_resolveIPv6` is only ever called with a numeric port (courtesy of [other code](https://github.com/twisted/twisted/blob/80e67a6775a190cba77e0d4229d12b9859ba12fc/src/twisted/internet/tcp.py#L1159-L1162)).  As a result, `AI_NUMERICSERV` is not an assertion that the port is numeric; instead, it's a direction to the resolver to [do less work](http://twistedmatrix.com/trac/ticket/5085#comment:48) that might fail:

> If the AI_NUMERICSERV flag is specified, then a non-null servname
> string supplied shall be a numeric port string.  Otherwise, an
> [EAI_NONAME] error shall be returned.  **This flag shall prevent any
> type of name resolution service (for example, NIS+) from being
> invoked.**

([From RFC 3493](https://tools.ietf.org/html/rfc3493#page-23), emphasis mine).

On Linux, for example, [nssswitch.conf](http://man7.org/linux/man-pages/man5/nsswitch.conf.5.html) might direct the glibc resolver backing `getaddrinfo` to first look in NIS+ for service names, even those that are numeric.  `AI_NUMERICSERV`, when available, bypasses that lookup regardless of the configuration.

This cannot be done without `AI_NUMERICSERV`.

Given the actual purpose of `AI_NUMERICSERV` here and the difficulties around coming up with a platform-appropriate exception to raise for non-numeric ports, I'd actually prefer to see a Solaris specific work around for this issue.

markrwilliams (changes requested):
Please see the comment for why a Solaris specific workaround seems better.

src/twisted/internet/tcp.py
adiroiban: Thanks for the review and the explanation. Much appreciated :)

There is this tests which check that socket.EAI_NONAME is raise
https://github.com/twisted/twisted/blob/0dd344d492dee0917f0efbf66534b04a14051aaa/src/twisted/internet/test/test_tcp.py#L2510

I will look at setting up a solaris 11 slave and then resume this.

glyph:
I don't think we are likely to start supporting Solaris again any time soon.  Really, Solaris should just fix this bug on their side.

PR #986, opened 2018-04-01
#7798 Handle non-socket.error in TCP client and invalid port numbers.

### LINKED ISSUE #7798: Handle non-socket.error in tcp client when connection with an address which does not require name resolution.
|-|-|
|Trac ID|trac#7798|
|Type|defect|
|Created|2015-03-05 15:48:42Z|
|Branch|https://github.com/twisted/twisted/tree/ipv6-client-port-overflow-7798-2|

Example code and output is here

https://gist.github.com/adiroiban/3967df30aca9add2eb9a

---

I will come back with a test and a patch

```
trac-id__7798 7798
type__defect defect
reporter__adiroiban adiroiban
priority__normal normal
milestone__ 
branch__branches_ipv6_client_port_overflow_7798_2 branches/ipv6-client-port-overflow-7798-2
branch_author__adiroiban adiroiban
status__new new
resolution__None None
component__core core
keywords__ 
time__1425570522562906 1425570522562906
changetime__1457222224099340 1457222224099340
version__None None
owner__adiroiban adiroiban

```
</details>

adiroiban:
Scope
=====

Fixes #7798

See error example https://gist.github.com/adiroiban/3967df30aca9add2eb9a

This fixes an unhandled error when failing to connect (or listen) due to various reasons.

The previous review is at https://twistedmatrix.com/trac/ticket/7798#comment:19

Change
======

Always call the code from a deferred.
Add tests and update the handling when an invalid port range is provided.

Contributor Checklist
=================

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/7798

How to test
=========

See https://gist.github.com/adiroiban/3967df30aca9add2eb9a


adiroiban:
@markrwilliams thanks for the update

These tests are gonna be ugly as for now Twisted just forwards the OS error, and it looks like each OS has a different error.

My approach was not not reivent the wheel and add PORT number validation in twisted, but just forward any number to the OS call.

Maybe this is not a good idea, and in twisted we can add a simple check for port number and raise a standard error. This would make a lot of tests nicer to read. 

The tests are still nice as they document the low level error for each os...but I think there is not much use in them.


In Ubuntu 18.04 we now have 

```diff
- ['getsockaddrarg: port must be 0-65535.']
+ [-8]
```

glyph:
@adiroiban are you interested in resolving these conflicts and finishing up this PR?  We should probably close it if it's dead

adiroiban:
The error assertion in the test is ugly, as Twisted is just raising the low-level error.

the important part of this PR is to make sure that the errback is called on errors.

this should be ready for review

needs-review


-----------

There was an unrelated flaky test on macos ... but no more of the old cfreactor errors :)

```
[FAIL]
  File "/Users/runner/work/twisted/twisted/.tox/macos-withcov-alldeps/lib/python3.12/site-packages/twisted/internet/test/test_process.py", line 523, in test_errorDuringExec
    self.assertIn("\N{SNOWMAN}".encode(), output.getvalue())
  File "/Users/runner/work/twisted/twisted/.tox/macos-withcov-alldeps/lib/python3.12/site-packages/twisted/trial/_synctest.py", line 509, in assertIn
    raise self.failureException(msg or f"{containee!r} not in {container!r}")
twisted.trial.unittest.FailTest: b'\xe2\x98\x83' not in b''

twisted.internet.test.test_process.PTYProcessTestsBuilder_AsyncioSelectorReactorTests.test_errorDuringExec

```

glyph:
> There was an unrelated flaky test on macos ... but no more of the old cfreactor errors :)

Hmm. This might be #8840 still being an issue

src/twisted/internet/tcp.py
```diff
@@ -669,7 +672,19 @@ def _resolveIPv6(ip, port):
     @raise socket.gaierror: if either the IP or port is not numeric as it
         should be.
     """
-    return socket.getaddrinfo(ip, port, 0, 0, 0, _NUMERIC_ONLY)[0][4]
+    usedPort = port
+    if isinstance(port, int):
+        # On OS X `getaddrinfo` will reject invalid port number while other
```
glyph:
```suggestion
        # On macOS `getaddrinfo` will reject invalid port number while other
```

src/twisted/internet/tcp.py
```diff
@@ -669,7 +672,19 @@ def _resolveIPv6(ip, port):
...
+        # systems accept it as long as it is an int.
+        # We pass a valid port which is later overwritten.
+        usedPort = 1
+
+    result = socket.getaddrinfo(ip, usedPort, 0, 0, 0, _NUMERIC_ONLY)[0][4]
+    # On non OS X `getaddrinfo` will also `resolve` invalid port numbers into a
```
glyph:
```suggestion
    # On every platform except macOS, `getaddrinfo` will also `resolve` invalid port numbers into a
```

src/twisted/internet/test/test_tcp.py
```diff
@@ -122,7 +124,16 @@
 
     getLinkLocalIPv6Addresses = _win32ifaces.win32GetLinkLocalIPv6Addresses
 
+    try:
+        from twisted.internet.iocpreactor.reactor import IOCPReactor
```
glyph, line 129: … why is this never covered? It seems like this line should definitely see coverage

glyph (commented):
I'd really like to see some code that confirms the assumptions as they are stipulated, since they should be pretty deterministic (we can pass an invalid port number and watch what happens), and to fix whatever is going on with the coverage reporting. Why aren't these tests being run?

src/twisted/internet/test/test_tcp.py
adiroiban: This line is covered.

The line 133 is not covered, as we only run the `-nodeps-` tests on Linux.

adiroiban:
Thanks for checking this.

> I'd really like to see some code that confirms the assumptions as they are stipulated, since they should be pretty deterministic (we can pass an invalid port number and watch what happens), and to fix whatever is going on with the coverage reporting. Why aren't these tests being run?

I am not sure what you are asking here.

The test are checking the actual low-level OS API.
There is no mock involved in these tests.

----------

Note that the Linux test take about [4 minutes, while the Windows test take 7 minutes](https://github.com/twisted/twisted/actions/runs/8972917304/usage 
)

This is why, you can see the coverage reported sooner... after only the Linux tests are executed.

glyph:
> I am not sure what you are asking here.
> 
> The test are checking the actual low-level OS API. There is no mock involved in these tests.

The comment makes a claim — that macOS will error in a particular way with erroneous integer port numbers. It would be nice to have a test which does exactly that, resolve an invalid port number, and see if it gets truncated or translated in the way it claims; not to test Twisted, but to test the assumptions upon which the implementation is based.

Not a hard blocker, would just be nice to have.

glyph:
OK, looks like we have another #8840 failure, I'm going to open a revert.

src/twisted/internet/tcp.py
```diff
@@ -669,7 +672,19 @@ def _resolveIPv6(ip, port):
     @raise socket.gaierror: if either the IP or port is not numeric as it
         should be.
     """
-    return socket.getaddrinfo(ip, port, 0, 0, 0, _NUMERIC_ONLY)[0][4]
+    usedPort = port
+    if isinstance(port, int):
+        # On macOS `getaddrinfo` will reject invalid port number while other
```
adiroiban: Thanks @glyph for the extra info.

Now it makes sense.

---------

I will write a separate check to make sure this will continue to fail on macOS

This code was written a long time ago with support for python 2.6

I now see that Linux is also failing

macos

```
% python3 
Python 3.9.6 (default, Nov 10 2023, 13:38:27) 
[Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import socket
>>> _AI_NUMERICSERV = getattr(socket, "AI_NUMERICSERV", 0)
>>> socket.getaddrinfo('127.0.0.1', -1, 0, 0, 0, _AI_NUMERICSERV)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/socket.py", line 953, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno 8] nodename nor servname provided, or not known
>>> 

```


linux
```
$ python3
Python 3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0] on linux
>>> import socket
>>> _AI_NUMERICSERV = getattr(socket, "AI_NUMERICSERV", 0)
>>> socket.getaddrinfo('127.0.0.1', -1, 0, 0, 0, _AI_NUMERICSERV)
  File "/usr/lib/python3.11/socket.py", line 962, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
socket.gaierror: [Errno -8] Servname not supported for ai_socktype
>>> 
```


--------

On Windows, no exception is raised

```
> python.exe
Python 3.11.7 (tags/v3.11.7:fa7a6f2, Dec  4 2023, 19:24:49) [MSC v.1937 64 bit (AMD64)] on win32
>>> import socket
>>> _AI_NUMERICSERV = getattr(socket, "AI_NUMERICSERV", 0)
>>> socket.getaddrinfo('127.0.0.1', -1, 0, 0, 0, _AI_NUMERICSERV)
[(<AddressFamily.AF_INET: 2>, 0, 0, '', ('127.0.0.1', 65535))]
```

glyph: > I now see that Linux is also failing

this kind of detail is exactly why i want to have a mechanical record of the behavior, rather than just a comment :-). Thanks for checking!

adiroiban:
I am going to write a test for socket.getaddrinfo to check that an exception is raised on Linux and macOS

src/twisted/internet/tcp.py
adiroiban: Done. Let me know if this is what you were expecting.

I am not used to write this type of tests... as this is a test for a stdlib function, not our code.

We have tests for _resolveIPv6 with invalid port number, and for me that is all that we should care about :)

adiroiban:
This is ready for review. Thanks!

twm:
Weird, this isn't showing up at https://twisted.reviews for me. please review?

glyph:
> Weird, this isn't showing up at https://twisted.reviews for me. please review?

It shouldn't — the end of the custom query there is `-author:@me` as it's the PRs *you* should review ;-)

adiroiban:
There is this GitHub URL https://github.com/pulls/review-requested which shoudl help detect things pending review... you can then filter based on project

twm:
> > Weird, this isn't showing up at https://twisted.reviews for me. please review?
> 
> It shouldn't — the end of the custom query there is `-author:@me` as it's the PRs _you_ should review ;-)

I don't follow... I didn't author this PR, I was looking for PRs of Adi's to review.

glyph:
> I don't follow... I didn't author this PR, I was looking for PRs of Adi's to review.

Oops, I must have confused myself with multiple tabs or something. Odd. It shows up for me.

twm:
> > I don't follow... I didn't author this PR, I was looking for PRs of Adi's to review.
> 
> Oops, I must have confused myself with multiple tabs or something. Odd. It shows up for me.

And now it shows up for me... :thinking: Maybe it counted as archived because due to staleness?

src/twisted/internet/test/test_tcp.py
glyph: Could we convince codecov not to leave this comment by adding a `# pragma: no branch` here, maybe?

src/twisted/internet/tcp.py
```diff
@@ -656,7 +659,7 @@ def _connectDone(self):
             self.protocol.makeConnection(self)
 
 
-_NUMERIC_ONLY = socket.AI_NUMERICHOST | _AI_NUMERICSERV
+_NUMERIC_ONLY = socket.AI_NUMERICHOST | socket.AI_NUMERICSERV
 
 
 def _resolveIPv6(ip, port):
```
glyph, line 665: Given that the logic has changed a bunch here it would be good to get a type annotation to ensure the signature has not changed.

src/twisted/internet/test/test_tcp.py
```diff
@@ -122,7 +125,16 @@
...
+    except ImportError:  # pragma: no cover
+        # On our CI, we always have iocp-support package available.
+        # This is here for the case in which someone runs the tests on
+        # Windows, without the package installed.
+        IOCPReactor = None  # type: ignore[misc,assignment]
```
glyph, line 134: Hmm. Seeing this comment written out here, it occurs to me that this *is* a case we might want to care about? At least at one point, the constituency of Windows users who could install Twisted but could not compile (or even necessarily install) native code was one we cared about.  If we do care about them, we should probably have CI; if not, we should probably just delete this code.

However that is a long and potentially annoying policy discussion and for this PR the coverage exception is fine, I don't want to get bogged down here, just raise the issue for addressing later.

src/twisted/internet/test/test_tcp.py
```diff
@@ -750,6 +765,189 @@ def whenRun():
             raise SkipTest("Reactor does not support ITLSTransport")
         self.assertEqual(BrokenContextFactory.message, str(results[0]))
 
+    def test_connectPortUnderflow(self):
```
glyph, line 768: New code / new tests should have type annotations.
```suggestion
    def test_connectPortUnderflow(self) -> None:
```

glyph: (If it would be a big burden to annotate this stuff because it's surfacing whole new classes of typing problem, I would say it's fine to just file an issue to note the issue and not block here, particularly given that these are tests rather than new API)

src/twisted/internet/test/test_tcp.py
```diff
@@ -750,6 +765,189 @@ def whenRun():
...
+            firstResolve = result[0]
+            addressResolve = firstResolve[4]
+            portResolve = addressResolve[1]
+            self.assertEqual(65535, portResolve)
+            return
+
+        with self.assertRaises(socket.gaierror):
+            resolve_negative_port()
+
+    def test_getaddrinfo_overflow_port(self):
```
glyph, line 875: Just a coding standard nitpick.
```suggestion
    def test_getaddrinfoOverflowPort(self) -> None:
```

src/twisted/newsfragments/7798.bugfix
```diff
@@ -0,0 +1 @@
+Initiating a TCP connection with an address literals and a port outside of the 1-65535 range in twisted.internet.tcp.BaseClient and twisted.internet.iocpreactor.tcp.Client now triggers an errback instead of an unhandled error. This change might also fix other unhandled errors occurred while connecting using address literals.
```
glyph, line 1: Thinking of `BaseClient` and `Client` as public API is correct but it still makes me grind my teeth a bit 😬 . Could we describe this as `IReactorTCP.connectTCP` implementations.

.github/workflows/test.yaml
```diff
@@ -181,7 +181,7 @@ jobs:
           # We still run some tests with coverage,
           # as there are test with specific code branches for pypy.
           - python-version: 'pypy3.10-v7.3.13'
-            trial-target: 'twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests'
+            trial-target: 'twisted.internet.test.test_tcp twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests'
```
glyph, line 184: … interesting, why this one in particular? I did not clock any PyPy specifics

glyph (approved):
I'd love to see some type annotations here but I will leave it to your discretion whether they're an unreasonable amount of work to fix the bug. As noted inline though, if it is too annoying, please file an issue.

src/twisted/newsfragments/7798.bugfix
adiroiban: Sure.

I have updated to comply with our newsfragment standard.

This code was originally written a long time ago.... 2018 

```suggestion
IReactorTCP.connectTCP via the twisted.internet.tcp.BaseClient and twisted.internet.iocpreactor.tcp.Client implementations,
now triggers an errback instead of an unhandled error when using a TCP connection with an address literals and a port outside of the 1-65535 range. This change might also fix other unhandled errors occurred while connecting using address literals.
```

glyph:
I ack'd the regressions because these pretty clearly look like noise. There's nothing related to TCP at all in most of those benchmarks.

glyph:
Not sure why the robot did not see my review and update the labels accordingly, so I've done it manually.

adiroiban:
This was lost. I will merge it now via auto-merge.

I have also enabled the coverage test on PYPY since it was reported as missing coverage.

The tests was already running on PYPY, but without coverage, since PYPY tests are slow.

PR #987, opened 2018-04-02
[#9406] Remove Python 3.3 checks

hawkowl:
https://twistedmatrix.com/trac/ticket/9406#ticket


markrwilliams (approved):
Thanks for removing newly minted cruft!

PR #989, opened 2018-04-02
Allow 'twisted.scripts.twistd' to be invoked via runpy (python -m).

jaraco:
## Contributor Checklist:

* [X] There is an [associated ticket in Trac](https://twistedmatrix.com/trac/ticket/9408).
* [X] The changes pass minimal style checks.
* [X] I have updated the automated tests (N/A).


jaraco:
I'll create the news fragment after there's some consensus this is the preferred approach.

jaraco:
I've demonstrated the behavior works as expected:

```
$ rwt -q "git+https://github.com/jaraco/twisted@9408-jaraco-runpy_twisted" ldaptor -- -m twisted.scripts.twistd --help | head
:0: UserWarning: You do not have a working installation of the service_identity module: 'No module named 'service_identity''.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.
Usage: twistd [options]
Options:
  -b, --debug          Run the application in the Python Debugger (implies
                       nodaemon),         sending SIGUSR2 will drop into
                       debugger
      --chroot=        Chroot to a supplied directory before running
  -d, --rundir=        Change to a supplied directory before running [default:
                       .]
  -e, --encrypted      The specified tap/aos file is encrypted.
      --euid           Set only effective user-id rather than real user-id.
```

src/twisted/scripts/twistd.py
```diff
@@ -32,3 +32,6 @@ def run():
 
 
 __all__ = ['run', 'runApp']
+
+
+__name__ == '__main__' and run()
```
markrwilliams: This should be an `if` statement, though I'm guessing you did this to avoid lowering code coverage?

markrwilliams (dismissed):
Thanks for your contribution!

I'm curious to know why `twist` (`python -m twisted`) doesn't work for your use case.

We certainly still support `twistd` so it makes sense that we should make `python -m twisted.something` run it.

Making `twisted.scripts.twistd` itself runnable breaks the convention followed by other `twisted.scripts` and makes it a very public API.  I would prefer a new package for `twistd` with its own `__main__`, so that we could instead do `python -m twisted.twistd`.

jaraco:
> I'm curious to know why twist (python -m twisted) doesn't work for your use case.

I didn't try twist because I assumed it was different from `twistd` and `twistd` was what the ldap server prescribed. I'll try to figure out what the relationship is.

jaraco:
It's not obvious to me how one would run `ldaptor` using `python -m twisted`. If I checkout ldaptor and run `python -m twisted test-ldapserver.tac`, I get `Error: Unknown command: test-ldapserver.tac`. How does one run the test-ldapserver using `twist`, or is `twistd` still needed?

src/twisted/scripts/twistd.py
jaraco: I've been using this pattern everywhere because it's more concise. I'm happy to use an if block if that's the preference of the project (and I should have done that from the start, modeling after the patterns already in the project).

jaraco:
>  I would prefer a new package for twistd

I've implemented this suggestion in the latest commit, though if it were my project, I wouldn't create a package for twistd, I'd just create `twisted/twistd.py` as the runpy script (and only move that to its own package if `twisted.twistd` later became something more than this etnry script.

markrwilliams:
> I'd just create `twisted/twistd.py`

@jaraco That seems OK - I'd like to move `_twistd_unix.py` into this new package, but I can create it in a later PR.  Feel free to make this a module.

Please add a newsfragment so i can get this merged!

jaraco:
> I'd like to move _twistd_unix.py into this new package

Cool. Since you have plans for it, no need to make it a module. Cheers.

twm:
@markrwilliams Do you consider this ready to land? This lacks a formal approval.

PR #992, opened 2018-04-07
Conch's client SSH endpoint now extracts the exit-signal payload.

markrwilliams:
Previously,
twisted.conch.endpoints._CommandChannel.request_exit_signal
incorrectly parsed the `exit-signal` payload.  As a result, the
endpoint's protocol's connectionLost received a ConnectionDone
exception instead of ProcessTerminated.

This commit also introduces a dependency on
PyHamcrest (http://pyhamcrest.readthedocs.io), which eases the writing
of sequence containment assertions.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9412
* [x] I have created a newsfragment in src/twisted/conch/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )


markrwilliams:
@wiml Does this address your issue?

src/twisted/conch/test/test_endpoints.py
```diff
@@ -9,9 +9,12 @@
 from struct import pack
 from errno import ENOSYS
 
+import hamcrest
```
markrwilliams, line 12: This is bound to be controversial.

Unlike [testtools](http://testtools.readthedocs.io/en/latest/), [PyHamcrest](https://pyhamcrest.readthedocs.io) doesn't depend on pbr.  Its style also conflicts with Twisted's.

However, it made the test a lot shorter and clearer.

wiml:
That looks like it fixes the issue I have (just looking at the conch code change, not the hamcrest stuff).

src/twisted/python/_setup.py
```diff
@@ -238,6 +238,7 @@ class my_build_ext(build_ext_twisted):
         "incremental >= 16.10.1",
         "Automat >= 0.3.0",
         "hyperlink >= 17.1.1",
+        "PyHamcrest >= 1.9.0",
```
adiroiban, line 241: Why do we need to push this dependency to production?

I was expecting to see it only in `_dev`

https://github.com/twisted/twisted/blob/c0dbc4ee76a27ead51db06e93959b3f75606bf73/src/twisted/python/_setup.py#L75

src/twisted/conch/newsfragments/9412.bugfix
```diff
@@ -0,0 +1 @@
+If the command run by twisted.conch.endpoints.SSHCommandClientEndpoint exits because of a delivered signal, the client protocol's connectionLost is now called with a ProcessTerminated exception instead of a ConnectionDone exception.
```
adiroiban, line 1: This is not what I read from the diff :) I see that the same ProcessTerminated was return in previous code.

adiroiban (changes requested):
I don't understand how this fixed the issue.

The fix might be ok, but I don't understand it.

Can you please add more details?

Thanks!

From the diff, this looks  like just adding the logger, but the reason is the same

-------

Not blocking for this PR.
The new code is fragile to invalid input.

This is not specific to this PR and there are many other places in conch.ssh where it assumes that the remote peer is a well-behaved SSH implementation.

So I don't think that this should be fixed here.

I just want to raise the issue and discuss a fix.

Maybe SSHTransportBase.dispatchMessage should be updated to handle these errors in a generic way.

I would say that it

Try to run the test with this patch

```
diff --git a/src/twisted/conch/test/test_endpoints.py b/src/twisted/conch/test/test_endpoints.py
index a8f2c99..0384773 100644
--- a/src/twisted/conch/test/test_endpoints.py
+++ b/src/twisted/conch/test/test_endpoints.py
@@ -611,7 +611,7 @@ class SSHCommandClientEndpointTestsMixin(object):
                                     # string
             b'\x01',                # Core dumped; boolean
             common.NS(b'message'),  # Error message; string (UTF-8 encoded)
-            common.NS(b'en-US'),    # Language tag; string
+            #common.NS(b'en-US'),    # Language tag; string
         ])
         exc = self._exitStatusTest(b'exit-signal', packet)
         exc.trap(ProcessTerminated)
```

This is what I use for my production code

```
    def dispatchMessage(self, messageNum, payload):
        """
        Send the received message to the appropriate method.

        It is called by `dataReceived` once a complete command was
        received.

        It is the root dispatcher for all the SSH commands.

        @type messageNum: C{int}
        @type payload: c{str}
        """
        deferred = defer.maybeDeferred(
            self._dispatchMessage, messageNum, payload)

        def eb_dispatch_failure(failure):
            """
            Called when we failed to handle the SSH command.
            """
            LOGGING_CODE_GOES_HERE

            # This is a critical error and we can not continue.
            self.sendDisconnect(
                DISCONNECT_CONNECTION_LOST,
                'Internal error while processing %s.' % (messageNum,),
                )

        deferred.addErrback(eb_dispatch_failure)
        # Deferred is returned only to help with testing as there should
        # be no other code wanting to use this result.
        return deferred
```

But I am not sure how to handle this and this is why I have not create a ticket yet.

src/twisted/conch/newsfragments/9412.bugfix
markrwilliams: That code never ran because `struct.unpack` would fail with an exception instead of parsing a well-formed `exit-signal` message:

```python
def request_exit_signal(self, data):
    """
    When the server sends the command's exit status, record it for later
    delivery to the protocol.
    @param data: The network-order four byte representation of the exit
        signal of the command.
    @type data: L{bytes}
    """
    (signal,) = unpack('>L', data)  # raises: struct.error: unpack requires a string argument of length 4
    self._reason = ProcessTerminated(None, signal, None)
```

You can see this on trunk by running:

```
>>> from twisted.conch.endpoints import _CommandChannel
>>> packet = '\x00\x00\x00\x04TERM\x01\x00\x00\x00\x07message\x00\x00\x00\x05en-US'
>>> _CommandChannel(None, None, None, None).request_exit_signal(packet)
  File "src/twisted/conch/endpoints.py", line 229, in request_exit_signal
    (signal,) = unpack('>L', data)
struct.error: unpack requires a string argument of length 4
```

[`log.callWithLogger`](https://twistedmatrix.com/documents/18.4.0/api/twisted.python.log.html#callWithLogger) suppresses the exception inside `SSHTransportBase.dispatchMessage`:

https://github.com/twisted/twisted/blob/9d6dbd7e0584bd10a502b4c99045dbd5e794abc5/src/twisted/conch/ssh/transport.py#L774-L775

Here's a partial traceback courtesy of a `pdb` breakpoint inside `_CommandChannel.request_exit_signal`:

```
  /home/mrw/src/python/twisted/src/twisted/internet/endpoints.py(125)dataReceived()
-> return self._wrappedProtocol.dataReceived(data)
  /home/mrw/src/python/twisted/src/twisted/conch/ssh/transport.py(750)dataReceived()
-> self.dispatchMessage(messageNum, packet[1:])
  /home/mrw/src/python/twisted/src/twisted/conch/ssh/transport.py(775)dispatchMessage()
-> messageNum, payload)
  /home/mrw/src/python/twisted/src/twisted/python/log.py(103)callWithLogger()
-> return callWithContext({"system": lp}, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/log.py(86)callWithContext()
-> return context.call({ILogContext: newCtx}, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/context.py(122)callWithContext()
-> return self.currentContext().callWithContext(ctx, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/context.py(85)callWithContext()
-> return func(*args,**kw)
  /home/mrw/src/python/twisted/src/twisted/conch/ssh/service.py(45)packetReceived()
-> return f(packet)
  /home/mrw/src/python/twisted/src/twisted/conch/ssh/connection.py(322)ssh_CHANNEL_REQUEST()
-> channel.requestReceived, requestType, rest[1:])
  /home/mrw/src/python/twisted/src/twisted/internet/defer.py(150)maybeDeferred()
-> result = f(*args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/log.py(103)callWithLogger()
-> return callWithContext({"system": lp}, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/log.py(86)callWithContext()
-> return context.call({ILogContext: newCtx}, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/context.py(122)callWithContext()
-> return self.currentContext().callWithContext(ctx, func, *args, **kw)
  /home/mrw/src/python/twisted/src/twisted/python/context.py(85)callWithContext()
-> return func(*args,**kw)
  /home/mrw/src/python/twisted/src/twisted/conch/ssh/channel.py(162)requestReceived()
-> return f(data)
> /home/mrw/src/python/twisted/src/twisted/conch/endpoints.py(230)request_exit_signal()
-> (signal,) = unpack('>L', data)
```

Because the exception is suppressed, the code related to `connectionLost` continues to run without setting the `_reason` attribute on the `_CommandChannel` instance.  Eventually `_CommandChannel.closed` runs, finds that its `_reason` attribute is `None`, and calls back the underlying protocol with a default `ConnectionDone` instance:

https://github.com/twisted/twisted/blob/9d6dbd7e0584bd10a502b4c99045dbd5e794abc5/src/twisted/conch/endpoints.py#L233-L243

In other words, the exception raised by `struct.unpack` prevents `_CommandChannel.request_exit_signal` from setting its `_reason` to `ProcessTerminated`, so calls the protocol's `connectionLost` method with `ConnectionDone` instead.

This PR has `request_exit_signal` actually parse the signal number out of the packet and thus set `_reason` to `ProcessTerminated`.  The packet contains additional information that cannot be included in `ProcessTerminated`, like whether or not a core dump occurred. I chose to log so that information so users could see it and the tests could assert the packet was fully parsed.

src/twisted/python/_setup.py
markrwilliams: Because the tests are packaged under `twisted` with the implication that users can run at least a subset of them in a fresh installation.

If I put `pyhamcrest` in the `dev` extra's requirements, then I would have had to put a skip on the tests that use it and a conditional import.  There are enough tests that I think would benefit from `pyhamcrest` that making it a `dev`-only dependency would result in many additional skips. 

It's also a small, pure-Python dependency that's only used in private test modules, so including a dependency on it seemed low risk.

adiroiban (approved):
Thanks for the details.

The description of the PR is better :)

Assuming that all tests are green, this can be merged.

Thanks again!

PR #993, opened 2018-04-08
9400 get request hostname exception

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. http://twistedmatrix.com/trac/ticket/9400 http://twistedmatrix.com/trac/ticket/9403


src/twisted/web/newsfragments/9400.feature
```diff
@@ -0,0 +1 @@
+twisted.web.http.Request.getRequestHost (and thus the deprecated getRequestHostname; see #9403) raises an UnsupportedTransport exception when called on requests that lack a Host header and are bound to non-TCP transports.
```
twm, line 1: This would seem to say that getRequestHostname will raise UnsupportedHost, but actually it raises AttributeError.

src/twisted/web/http.py
```diff
@@ -1318,22 +1318,54 @@ def getAllHeaders(self):
...
+    def getRequestHostname(self):
+        """
+        Get the hostname that the user passed in to the request.
+
+        This will either use the Host: header (if it is available) or
+        the host we are listening on if the header is unavailable.
+
+        @returns: the requested hostname
+
+        @rtype: C{bytes}
```
twm, line 1397: This does not match the return type the interface declares, which is `str`. I would prefer a native string here, as that is the type generally used for IP addresses and hostnames on Python 3.

src/twisted/web/server.py
```diff
@@ -507,7 +507,7 @@ def _prePathURL(self, prepath):
             hostport = ':%d' % port
         prefix = networkString('http%s://%s%s/' % (
             self.isSecure() and 's' or '',
-            nativeString(self.getRequestHostname()),
+            nativeString(self.getRequestHost()),
```
twm, line 510: The immediate conversion of the network string to a native string here is hinting that the method should just return a native string. ;-)

src/twisted/web/iweb.py
```diff
@@ -82,11 +82,30 @@ def getRequestHostname():
...
+        This method is B{deprecated}.  See L{getRequestHost} instead.
 
         @returns: the requested hostname
-        @rtype: C{str}
+
+        @rtype: L{str}
+        """
+
+
+    def getRequestHost():
```
twm, line 96: Isn't adding a method to an interface a backwards-incompatible change?

twm (changes requested):
Thanks for taking this on! I noted some inconsistencies between the interface and implementation that would need to be fixed before merge.

However, I am leery of the overall approach. [Adding a method to an interface is a backwards-incompatible change.](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#interface-changes) I doubt there are many implementations of `IRequest` interface in the wild (it is enormous), but this seems like a much more incompatible change than altering getRequestHostname to raise a more descriptive exception. Raising AttributeError is arguably a straightforward bug &mdash; why go through all this ceremony? Are you aware of any code that would actually break if AttributeError changed in this case? What about if getRequestHostname() raised a subclass of AttributeError?

src/twisted/web/iweb.py
markrwilliams: Agh, yes, it is - I suppose this will require a new interface.

adiroiban: Since getRequestHost is used for now only in the existing getRequestHostname, I think that is safe to leave it out of the interface and just handle it like an implementation detail.

Do we need it as a public method?

Maybe we can create a list of methods for IRequestV2 :) which can include the required methods to allow traversal before all content is received (after all headers are received) and deferred traversal.

twm:
I see that #981 added an `IRequest.getClientAddr()` method to the same interface. I guess we don't have a problem adding things to interfaces? If so, [ITransport.disconnected is on my bucket list](https://twistedmatrix.com/trac/ticket/6606).

adiroiban:
> I see that #981 added an IRequest.getClientAddr() method to the same interface. I guess we don't have a problem adding things to interfaces? If so, ITransport.disconnected is on my bucket list.

There can always be exceptions... especially when things are broken :)

Instead of `getRequestHost` maybe we can have `getServerAddress` ... to match IRequest.getClientAddress()

Now... I would prefer to have `Host - Peer` vs `Server and Client` 

Host - Peer should match the other API.

Now, since getClientAddress was not yet released, maybe we can still rename it.

@glyph @markrwilliams @hawkowl  What do you think about using Host - Peer instead of Server - Client?

Thanks!

glyph:
> There can always be exceptions... especially when things are broken :)

This is definitely a compatibility exception though and should go through the relevant process.

glyph:
> @glyph @markrwilliams @hawkowl What do you think about using Host - Peer instead of Server - Client?

Let's have a nicely designed interface for IRequest at some point in the future (perhaps starting with the txrequest/Klein stuff).  However, for the time being, I'd say that the fallback behavior should *really* be implemented on the existing method, and we should not add a new one, or deprecate the existing one.

glyph:
I realize that this is a small behavioral change, but it's only relevant on spec-non-compliant requests anyway :).

PR #994, opened 2018-04-10
#9101 allow TCP listeners to re-use ports across processes

### LINKED ISSUE #9101: tcp: adding support for SO_REUSEPORT
|[wdauchy's avatar](https://github.com/wdauchy)| @wdauchy reported|
|-|-|
|Trac ID|trac#9101|
|Type|enhancement|
|Created|2017-04-01 05:56:19Z|
|Branch|https://github.com/twisted/twisted/tree/9101-wdauchy-tcp_reuseport|

I'd like to add support for SO_REUSEPORT using tcp sockets

```
trac-id__9101 9101
type__enhancement enhancement
reporter__wdauchy wdauchy
priority__normal normal
milestone__None None
branch__9101_wdauchy_tcp_reuseport 9101-wdauchy-tcp_reuseport
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1491026179574066 1491026179574066
changetime__1507144348657302 1507144348657302
version__None None
owner__None None

```
</details>

moshez:
Fixes #9101


glyph:
hey @moshez . what's up

src/twisted/test/test_tcp.py
```diff
@@ -187,6 +187,21 @@ def test_listen(self):
         self.addCleanup(p1.stopListening)
         self.assertTrue(interfaces.IListeningPort.providedBy(p1))
 
+    def test_multiListen(self):
+        """
+        Test that multiple sockets can listen on the same port
+        """
+        if not hasattr(socket, "SO_REUSEPORT"):
+            raise unittest.SkipTest('SO_REUSEPORT not supported')
```
adiroiban: My understading is that on Windows, SO_REUSEADDR also implements SO_REUSEPORT... so listenMultiple is also supported on Windows.

But on Windows, we might want to use SO_EXCLUSIVEADDRUSE

https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse?redirectedfrom=MSDN

Maybe this test should be extend by making 100 ping connections and check that both factories will receive at least one connection.

We don't know if the OS will do a round-robin between the ports, but I guess that our of 100 connection we should get at least another connection.

At the same time, this is not testing listening to the same port from different twisted processes ... but I guess that this is something for OS and the current level of testing is OK for Twisted. 

Also, I think that we need a test in which when listenMultiple=False we see that  the second listenTCP fails... just to make sure we don't accidentally set listenMultiple all the time

adiroiban:
@glyph do we need to do all the extra interface changes here?

Are there any TCP reactors implementations outside of Twisted ? 

I don't see what will break if we add a new optional argument to the existing reactor.

glyph:
> @glyph do we need to do all the extra interface changes here?

Yes, although we could do them in a slightly less invasive way. We *do* need a new interface, but we don't necessarily need a new method name.

> Are there any TCP reactors implementations outside of Twisted ?

Yes, various GUI reactors (such as the QT reactor) live outside of Twisted.

> I don't see what will break if we add a new optional argument to the existing reactor.

Third-party reactors - or even things short of full-on reactor implementations, such as diagnostic wrappers - getting an unexpected parameter with no warning.

PR #995, opened 2018-04-11
[9418] IReactorTime.getDelayedCalls() return type

twm:
This updates the IReactorTime.getDelayedCalls() docstring to reflect what all the implementations in Twisted return: a list, not a tuple.

This is the simplest change I could make to resolve the issue. There are other possibilities which we could discuss:

* The interface could be documented as returning a sequence, leaving the concrete type up to the reactor implementation.
* All of the implementations could be changed to return a tuple. (I don't suggest this, as it breaks backward compatibility because tuples and lists don't compare equal.)

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9418
* [x] (N/A) I have updated the automated tests.

### DIFF
--- a/src/twisted/internet/interfaces.py
+++ b/src/twisted/internet/interfaces.py
@@ -1267,9 +1267,9 @@ class IReactorTime(Interface):
         """
         Retrieve all currently scheduled delayed calls.
 
-        @return: A tuple of all L{IDelayedCall} providers representing all
+        @return: A list of L{IDelayedCall} providers representing all
                  currently scheduled calls. This is everything that has been
-                 returned by C{callLater} but not yet called or canceled.
+                 returned by C{callLater} but not yet called or cancelled.
         """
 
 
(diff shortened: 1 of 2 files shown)


src/twisted/internet/interfaces.py
```diff
@@ -1267,9 +1267,9 @@ def getDelayedCalls():
         """
         Retrieve all currently scheduled delayed calls.
 
-        @return: A tuple of all L{IDelayedCall} providers representing all
+        @return: A list of L{IDelayedCall} providers representing all
```
markrwilliams, line 1270: If all the implementers return `list`s, we're probably stuck with `list`s.

markrwilliams (approved):
As you mentioned, it's important that the interface specify the concrete types that users can check equality.

I agree that this is the simplest possible change. :+1:

PR #996, opened 2018-04-12
5368 break on emfile

markrwilliams:
## Remove this paragraph

Please have a look at our developer documentation before submitting your Pull Request.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/5368

### DIFF
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -1355,10 +1355,16 @@ class Port(base.BasePort, _SocketCloser):
                         skt, protocol, addr, self, s, self.reactor)
                     protocol.makeConnection(transport)
 
-            if not accepted:
-                self.numberAccepts = self.numberAccepts+20
+            # Scale our synchronous accept loop according to traffic
+            # Reaching our limit on consecutive accept calls indicates
+            # there might be still more clients to serve the next time
+            # the reactor calls us.  Prepare to accept some more.
+            if accepted == self.numberAccepts:
+                self.numberAccepts += 20
+            # Otherwise, don't attempt to accept any more clients than
+            # we just accepted or any less than 1.
             else:
-                self.numberAccepts = accepted
+                self.numberAccepts = max(1, accepted)
         except BaseException:
             # Note that in TLS mode, this will possibly catch SSL.Errors
             # raised by self.socket.accept()
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -927,7 +927,7 @@ class _FileDescriptorReservation(object):
     """
     L{_IFileDescriptorReservation} implementation.
 
-    @param fileFactory: A factory that will be called to reserve a
+    @ivar fileFactory: A factory that will be called to reserve a
         file descriptor.
     @type fileFactory: A L{callable} that accepts no arguments and
         returns an object with a C{close} method.
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -897,7 +897,7 @@ class _FileDescriptorReservation(object):
     descriptor's space, and then close that connection and reopen this
     one.
 
-    Calling L{_FileDescriptorReservation.ensure} opens the reserve
+    Calling L{_FileDescriptorReservation.acquire} opens the reserve
     file descriptor if it is not already open.
     L{_FileDescriptorReservation.release} close this descriptor; a
     subsequent call to L{_FileDescriptorReservation.replace} will
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -9,7 +9,6 @@ End users shouldn't use this module directly - use the reactor APIs instead.
 """
 
 from __future__ import division, absolute_import
-
 # System Imports
 import socket
 import sys
@@ -19,12 +18,10 @@ import struct
 
 import attr
 
-from automat import MethodicalMachine
-
 from zope.interface import Interface, implementer
 
 from twisted.logger import Logger
-from twisted.python.compat import lazyByteSlice, unicode
+from twisted.python.compat import izip, lazyByteSlice, unicode
 from twisted.python.runtime import platformType
 from twisted.python import versions, deprecate
 
@@ -897,15 +894,20 @@ class _IFileDescriptorReservation(Interface):
     descriptor's space, and then close that connection and reopen this
     one.
 
-    Calling L{_FileDescriptorReservation.reserve} opens the reserve
-    file descriptor if it is not already open.
-    L{_FileDescriptorReservation.release} close this descriptor; a
-    subsequent call to
-    L{_FileDescriptorReservation.maybeCloseForReserved} will close the
-    provided file and re-open the reserved file.  A claimed
-    reservation will not replace the provided file with itself.
-    L{_FileDescriptorReservation.reserve} and
-    L{_FileDescriptorReservation.release} are idempotent.
+    Calling L{_IFileDescriptorReservation.reserve} attempts to open
+    the reserve file descriptor if it is not already open.
+    L{_IFileDescriptorReservation.available} returns L{True} if the
+    underlying file is open and its descriptor claimed.
+
+    L{_IFileDescriptorReservation} instances are context managers;
+    entering them releases the underlying file descriptor, while
+    exiting them attempts to reacquire it.  The block can take
+    advantage of the free slot in the process' file descriptor table
+    accept and close a client connection.
+
+    Because another thread might open a file descriptor between the
+    time the context manager is entered and the time C{accept} is
+    called, opening the reserve descriptor is best-effort only.
     """
 
     def available():
@@ -929,24 +931,21 @@ class _IFileDescriptorReservation(Interface):
         """
 
 
-    def maybeCloseForReserved(fileObject):
+    def __enter__():
         """
-        Close the provided file if and only if L{release} has been
-        called to free the reserved file's file descriptor in exchage
-        for re-opening the reserved file.
-
-        @param fileObject: The file object to close iff the
-            reservation is available.
-
-        @return: L{True} when the file object was closed; L{False}
-            otherwise.
+        Release the underlying file descriptor so that code within the
+        context manager can open a new file.
         """
 
 
-    def release():
+    def __exit__(excType, excValue, traceback):
         """
-        Release the reserved descriptor so another file can take
-        its place.
+        Attempt to re-open the reserved file descriptor.  See
+        L{reserve} for caveats.
+
+        @param excType: See L{object.__exit__}
+        @param excValue: See L{object.__exit__}
+        @param traceback: See L{object.__exit__}
         """
 
 
@@ -963,56 +962,9 @@ class _FileDescriptorReservation(object):
         returns an object with a C{close} method.
     """
     _log = Logger()
-    _machine = MethodicalMachine()
 
     _fileFactory = attr.ib()
     _fileDescriptor = attr.ib(init=False, default=None)
-
-    @_machine.state(initial=True)
-    def _unreserved(self):
-        """
-        The reserved file is unreserved.
-        """
-
-    @_machine.state()
-    def _reservationRequested(self):
-        """
-        A request to open the reserved file was made.
-        """
-
-    @_machine.state()
-    def _reserved(self):
-        """
-        The reserved file is reserved.
-        """
-
-    @_machine.input()
-    def _maybeReserve(self):
-        """
-        Attempt to open the reserved file.
-        """
-
-    @_machine.input()
-    def _closeWhenUnreserved(self, fileObject):
-        """
-        Close the file object when the reserved file's descriptor is
-        available because of a prior call to L{release}.
-
-        @param fileObject: The file object to close.
-        """
-
-    @_machine.input()
-    def _reservationFailedWithEMFILE(self):
-        """
-        An attempt to open the reserved file itself failed with
-        C{EMFILE}.
-        """
-
-    @_machine.input()
-    def _fileDesecriptorReserved(self):
-        """
-        The reserved file was opened.
-        """
 
 
     def available(self):
@@ -1030,137 +982,40 @@ class _FileDescriptorReservation(object):
         """
         See L{_IFileDescriptorReservation.reserve}.
         """
-        reserved = self._maybeReserve()
-        if reserved:
-            self._fileDesecriptorReserved()
-        else:
-            self._reservationFailedWithEMFILE()
-
-
-    def maybeCloseForReserved(self, fileObject):
-        """
-        See L{_IFileDescriptorReservation.maybeCloseForReserved}.
-
-        @param fileObject: The file object to close iff the
-            reservation is available.
-
-        @return: L{True} when the file object was closed; L{False}
-            otherwise.
-        """
-        if self._closeWhenUnreserved(fileObject):
-            self.reserve()
-            return True
-        return False
-
-
-    @_machine.input()
-    def release(self):
-        """
-        See L{_IFileDescriptorReservation.maybeCloseForReserved}.
-        Release the reserved descriptor so another file can take
-        its place.
-        """
-
-    @_machine.output()
-    def _openReservationFile(self):
-        """
-        Use the C{_fileFactory} to claim a file descriptor.
-        """
-        try:
-            fileDescriptor = self._fileFactory()
-        except (IOError, OSError) as e:
-            if e.errno == EMFILE:
-                self._log.failure("Failed to open reserved file.")
-                self._fileDescriptor = None
-                return False
+        if self._fileDescriptor is None:
+            try:
+                fileDescriptor = self._fileFactory()
+            except (IOError, OSError) as e:
+                if e.errno == EMFILE:
+                    self._log.failure(
+                        "Could not reserve EMFILE recovery file descriptor.")
+                    self._fileDescriptor = None
+                else:
+                    raise
             else:
-                raise
-        else:
-            self._fileDescriptor = fileDescriptor
-            return True
+                self._fileDescriptor = fileDescriptor
 
 
-    @_machine.output()
-    def _closeReservedFile(self):
+    def __enter__(self):
         """
-        Close the reserved file.
+        See L{_IFileDescriptorReservation.__enter__}.
         """
+        if not self._fileDescriptor:
+            raise RuntimeError(
+                "No file reserved.  Have you called my reserve method?")
         self._fileDescriptor.close()
         self._fileDescriptor = None
 
 
-    @_machine.output()
-    def _closeFile(self, fileObject):
+    def __exit__(self, excValue, excType, traceback):
         """
-        Close the provided file object and return L{True}
-
-        @param fileObject: The file object to close.
-        @return: L{True}
+        See L{_IFileDescriptorReservation.__exit__}.
         """
-        fileObject.close()
-        return True
-
-
-    @_machine.output()
-    def _notClosed(self, fileObject):
-        """
-        Return False.
-
-        @param fileObject: ignored.
-
-        @return: L{False}
-        """
-        return False
-
-
-    @_machine.output()
-    def _returnTrue(self):
-        """
-        An output that always returns L{True}.
-
-        @return: L{True}
-        """
-        return True
-
-    _unreserved.upon(
-        release,
-        enter=_unreserved, outputs=[],
-        collector=_silenceOutputs)
-    _unreserved.upon(
-        _maybeReserve,
-        enter=_reservationRequested, outputs=[_openReservationFile],
-        collector=_nthOutput(0))
-    _unreserved.upon(
-        _closeWhenUnreserved,
-        enter=_unreserved,
-        outputs=[_closeFile],
-        collector=_nthOutput(-1))
-
-    _reservationRequested.upon(
-        _reservationFailedWithEMFILE,
-        enter=_unreserved, outputs=[],
-        collector=_silenceOutputs)
-    _reservationRequested.upon(
-        _fileDesecriptorReserved,
-        enter=_reserved, outputs=[],
-        collector=_silenceOutputs)
-
-    _reserved.upon(
-        release,
-        enter=_unreserved, outputs=[_closeReservedFile],
-        collector=_silenceOutputs)
-    _reserved.upon(
-        _maybeReserve,
-        enter=_reserved, outputs=[_returnTrue],
-        collector=_nthOutput(0))
-    _reserved.upon(
-        _fileDesecriptorReserved,
-        enter=_reserved, outputs=[],
-        collector=_silenceOutputs)
-    _reserved.upon(
-        _closeWhenUnreserved, enter=_reserved,
-        outputs=[_notClosed],
-        collector=_nthOutput(0))
+        try:
+            self.reserve()
+        except Exception:
+            self._log.failure(
+                "Could not reserve EMFILE recovery file descriptor.")
 
 
 
@@ -1187,21 +1042,21 @@ class _NullFileDescriptorReservation(object):
         """
 
 
-    def maybeCloseForReserved(self, fileObject):
+    def __enter__(self):
         """
-        Never close the file in exchange for the reserved file.  See
-        L{_IFileDescriptorReservation.maybeCloseForReserved}.
-
-        @param fileObject: The file; never closed by this method.
+        Do nothing. See L{_IFileDescriptorReservation.__enter__}
 
         @return: L{False}
         """
-        return False
 
 
-    def release(self):
+    def __exit__(self, excValue, excType, traceback):
         """
-        Do nothing.  See L{_IFileDescriptorReservation.release}.
+        Do nothing.  See L{_IFileDescriptorReservation.__exit__}.
+
+        @param excType: See L{object.__exit__}
+        @param excValue: See L{object.__exit__}
+        @param traceback: See L{object.__exit__}
         """
 
 
@@ -1228,6 +1083,66 @@ else:
 # readable, so accept(2) will never be called.  Calling accept(2) on
 # such a listener, however, does not return at all.
 _ACCEPT_ERRORS = (EMFILE, ENOBUFS, ENFILE, ENOMEM, ECONNABORTED)
+
+
+def _accept(listener, reservedFD):
+    """
+    Return a generator that yields client sockets from the provided
+    listening socket until there are none left or an unrecoverable
+    error occurs.
+
+    @param listener: The listening socket.
+    @type listener: L{socket.socket}
+
+    @param reservedFD: A reserved file descriptor that can be used to
+        recover from C{EMFILE} on UNIX-like systems.
+    @type reservedFD: L{_IFileDescriptorReservation}
+
+    @return: A generator that yields C{(socket, addr)} tuples from
+        L{socket.accept}
+    """
+
+    while True:
+        try:
+            client, address = listener.accept()
+        except socket.error as e:
+            if e.args[0] in (EWOULDBLOCK, EAGAIN):
+                # No more clients.
+                return
+            elif e.args[0] == EPERM:
+                # Netfilter on Linux may have rejected the
+                # connection, but we get told to try to accept()
+                # anyway.
+                continue
+            elif e.args[0] == EMFILE and reservedFD.available():
+                # Linux and other UNIX-like operating systems return
+                # EMFILE when a process has reached its soft limit of
+                # file descriptors.  The reserved file descriptor is
+                # available, so it can be released to free up a
+                # descriptor for use by listener.accept()'s clients.
+                # Each client socket will be closed until the listener
+                # returns EAGAIN.
+                with reservedFD:
+                    log.msg("EMFILE encountered;"
+                            " releasing reserved file descriptor.")
+                    clientsToClose = _accept(listener, reservedFD)
+                    for clientToClose, closedAddr in clientsToClose:
+                        clientToClose.close()
+                        log.msg("EMFILE recovery:"
+                                " Closed socket from %r"
+                                " and reopening reserve file descriptor"
+                                % (closedAddr,))
+                    # All pending client connections have been closed.
+                    # Return control back to the reactor.
+                    return
+            elif e.args[0] in _ACCEPT_ERRORS:
+                log.msg("Could not accept new connection (%s)" % (
+                    errorcode[e.args[0]],))
+                return
+            else:
+                raise
+        else:
+            yield client, address
 
 
 
@@ -1401,64 +1316,28 @@ class Port(base.BasePort, _SocketCloser):
                 # win32 event loop breaks if we do more than one accept()
                 # in an iteration of the event loop.
                 numAccepts = 1
-            i = 0
-            while i < numAccepts:
-                # we need this so we can deal with a factory's buildProtocol
-                # calling our loseConnection
-                if self.disconnecting:
-                    return
-                try:
-                    skt, addr = self.socket.accept()
-                except socket.error as e:
-                    if e.args[0] in (EWOULDBLOCK, EAGAIN):
-                        self.numberAccepts = i
-                        break
-                    elif e.args[0] == EPERM:
-                        # Netfilter on Linux may have rejected the
-                        # connection, but we get told to try to accept()
-                        # anyway.
-                        continue
-                    elif e.args[0] == EMFILE and _reservedFD.available():
-                        # Linux and other UNIX-like operating systems
-                        # return EMFILE when a process has reached its
-                        # soft limit of file descriptors.
-                        log.msg(
-                            "EMFILE encountered;"
-                            " releasing reserved file descriptor.")
-                        _reservedFD.release()
-                        # Ensure that we have one more iteration
-                        # available after this.
-                        if i == (numAccepts - 1):
-                            numAccepts += 1
-                        continue
-                    elif e.args[0] in _ACCEPT_ERRORS:
-                        log.msg("Could not accept new connection (%s)" % (
-                            errorcode[e.args[0]],))
-                        break
-                    raise
 
-                i += 1
+            accepted = 0
+            attempts = range(1, numAccepts + 1)
+            clients = _accept(self.socket, _reservedFD)
+            # Limit the infinite stream of clients with the attempts
+            # iterator
+            for accepted, (skt, addr) in izip(attempts, clients):
+                fdesc._setCloseOnExec(skt.fileno())
+                protocol = self.factory.buildProtocol(self._buildAddr(addr))
+                if protocol is None:
+                    skt.close()
+                    continue
+                s = self.sessionno
+                self.sessionno = s + 1
+                transport = self.transport(
+                    skt, protocol, addr, self, s, self.reactor)
+                protocol.makeConnection(transport)
 
-                if _reservedFD.maybeCloseForReserved(skt):
-                    log.msg(
-                        "EMFILE recovery:"
-                        " Closing socket from %r"
-                        " and reopening reserve file descriptor"
-                        % (addr,))
-                else:
-                    fdesc._setCloseOnExec(skt.fileno())
-                    protocol = self.factory.buildProtocol(
-                        self._buildAddr(addr))
-                    if protocol is None:
-                        skt.close()
-                        continue
-                    s = self.sessionno
-                    self.sessionno = s+1
-                    transport = self.transport(
-                        skt, protocol, addr, self, s, self.reactor)
-                    protocol.makeConnection(transport)
-            else:
+            if not accepted:
                 self.numberAccepts = self.numberAccepts+20
+            else:
+                self.numberAccepts = accepted
         except BaseException:
             # Note that in TLS mode, this will possibly catch SSL.Errors
             # raised by self.socket.accept()
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -954,7 +954,7 @@ class _FileDescriptorReservation(object):
 
 
     @_machine.input()
-    def replace(self, fileObject):
+    def maybeClose(self, fileObject):
         """
         Replace a file with the reservation file only if our reserve
         file is unclaimed.
@@ -1054,7 +1054,6 @@ class _FileDescriptorReservation(object):
         """
         if self._fileDescriptor is None:
             self._openFailedWithEMFILE()
-            return False
         return True
 
 
@@ -1074,7 +1073,7 @@ class _FileDescriptorReservation(object):
         acquire, enter=_reserved, outputs=[_openReservationFile],
         collector=_silenceOutputs)
     _unreserved.upon(
-        replace, enter=_reserved,
+        maybeClose, enter=_reserved,
         outputs=[_closeFile, _reopenReservationFile, _replaced],
         collector=_nthOutput(-1))
     _unreserved.upon(
@@ -1088,7 +1087,7 @@ class _FileDescriptorReservation(object):
         acquire, enter=_reserved, outputs=[],
         collector=_silenceOutputs)
     _reserved.upon(
-        replace, enter=_reserved,
+        maybeClose, enter=_reserved,
         outputs=[_notReplaced],
         collector=_nthOutput(0))
     _reserved.upon(
@@ -1327,9 +1326,9 @@ class Port(base.BasePort, _SocketCloser):
                         break
                     raise
 
-                if self._reservedFD.replace(skt):
                 i += 1
 
+                if self._reservedFD.maybeClose(skt):
                     log.msg(
                         "EMFILE recovery:"
                         " Closing socket %r"
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -1212,6 +1212,25 @@ else:
     _reservedFD = _FileDescriptorReservation(lambda: open(os.devnull))
 
 
+# Linux and other UNIX-like operating systems return EMFILE when a
+# process has reached its soft limit of file descriptors. *BSD and
+# Win32 raise (WSA)ENOBUFS when socket limits are reached.  Linux can
+# give ENFILE if the system is out of inodes, or ENOMEM if there is
+# insufficient memory to allocate a new dentry.  ECONNABORTED is
+# documented as possible on all relevant platforms (Linux, Windows,
+# macOS, and the BSDs) but occurs only on the BSDs.  It occurs when a
+# client sends a FIN or RST after the server sends a SYN|ACK but
+# before application code calls accept(2).  On Linux, calling
+# accept(2) on such a listener returns a connection that fails as
+# though the it were terminated after being fully established.  This
+# appears to be an implementation choice (see inet_accept in
+# inet/ipv4/af_inet.c).  On macOS, such a listener is not considered
+# readable, so accept(2) will never be called.  Calling accept(2) on
+# such a listener, however, does not return at all.
+_ACCEPT_ERRORS = (EMFILE, ENOBUFS, ENFILE, ENOMEM, ECONNABORTED)
+
+
+
 @implementer(interfaces.IListeningPort)
 class Port(base.BasePort, _SocketCloser):
     """
@@ -1400,10 +1419,9 @@ class Port(base.BasePort, _SocketCloser):
                         # anyway.
                         continue
                     elif e.args[0] == EMFILE and _reservedFD.available():
-                        # Linux gives EMFILE when a process is not
-                        # allowed to allocate any more file
-                        # descriptors.  *BSD and Win32 give
-                        # (WSA)ENOBUFS.
+                        # Linux and other UNIX-like operating systems
+                        # return EMFILE when a process has reached its
+                        # soft limit of file descriptors.
                         log.msg(
                             "EMFILE encountered;"
                             " releasing reserved file descriptor.")
@@ -1413,27 +1431,7 @@ class Port(base.BasePort, _SocketCloser):
                         if i == (numAccepts - 1):
                             numAccepts += 1
                         continue
-                    elif e.args[0] in (
-                            EMFILE, ENOBUFS, ENFILE, ENOMEM, ECONNABORTED):
-                        # Linux can give ENFILE if the system is out
-                        # of inodes, or ENOMEM if there is
-                        # insufficient memory to allocate a new
-                        # dentry.  ECONNABORTED is documented as
-                        # possible on all relevant platforms (Linux,
-                        # Windows, macOS, and the BSDs) but occurs
-                        # only on the BSDs.  It occurs when a client
-                        # sends a FIN or RST after the server sends a
-                        # SYN|ACK but before application code calls
-                        # accept(2).  On Linux, calling accept(2) on
-                        # such a listener returns a connection that
-                        # fails as though the it were terminated after
-                        # being fully established.  This appears to be
-                        # an implementation choice (see inet_accept in
-                        # inet/ipv4/af_inet.c).  On macOS, such a
-                        # listener is not considered readable, so
-                        # accept(2) will never be called.  Calling
-                        # accept(2) on such a listener, however, does
-                        # not return at all.
+                    elif e.args[0] in _ACCEPT_ERRORS:
                         log.msg("Could not accept new connection (%s)" % (
                             errorcode[e.args[0]],))
                         break
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -923,20 +923,17 @@ class _FileDescriptorReservation(object):
         The reservation file is unreserved.
         """
 
-
     @_machine.state()
     def _reserved(self):
         """
         The reservation file is reserved.
         """
 
-
     @_machine.input()
     def available(self):
         """
         Is the reservation available?
         """
-
 
     @_machine.input()
     def release(self):
@@ -945,13 +942,11 @@ class _FileDescriptorReservation(object):
         its place.
         """
 
-
     @_machine.input()
     def acquire(self):
         """
         Ensure that the reservation file is held.
         """
-
 
     @_machine.input()
     def maybeClose(self, fileObject):
@@ -962,14 +957,12 @@ class _FileDescriptorReservation(object):
         @param fileObject: The file object to close.
         """
 
-
     @_machine.input()
     def _openFailedWithEMFILE(self):
         """
         An internal input that undoes a reservation that failed to
         open the reservation file.
         """
-
 
     def _createFD(self):
         """
@@ -1067,7 +1060,6 @@ class _FileDescriptorReservation(object):
         @return: L{False}
         """
         return False
-
 
     _unreserved.upon(
         acquire, enter=_reserved, outputs=[_openReservationFile],
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -21,7 +21,7 @@ import attr
 
 from automat import MethodicalMachine
 
-from zope.interface import implementer
+from zope.interface import Interface, implementer
 
 from twisted.logger import Logger
 from twisted.python.compat import lazyByteSlice, unicode
@@ -888,8 +888,7 @@ def _nthOutput(n):
 
 
 
-@attr.s
-class _FileDescriptorReservation(object):
+class _IFileDescriptorReservation(Interface):
     """
     An open file that represents an emergency reservation in the
     process' file descriptor table.  If L{Port} encounters C{EMFILE}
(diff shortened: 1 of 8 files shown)


src/twisted/internet/tcp.py
```diff
@@ -853,6 +858,246 @@ def getPeer(self):
...
+        collector=_silenceOutputs)
+    _reserved.upon(
+        replace, enter=_reserved,
+        outputs=[_notReplaced],
+        collector=_nthOutput(0))
+    _reserved.upon(
+        available, enter=_reserved, outputs=[_returnTrue],
+        collector=_nthOutput(0))
+    _reserved.upon(
+        _openFailedWithEMFILE, enter=_unreserved, outputs=[])
```
markrwilliams: This relies on Automat's transitioning to `_reserved` before the outputs for `_unreserved` --`replace`--> `_reserved` finish running.  That isn't a public API; see https://github.com/glyph/automat/issues/41.

src/twisted/internet/tcp.py
```diff
@@ -1038,42 +1285,62 @@ def doRead(self):
...
+                        continue
+                    s = self.sessionno
+                    self.sessionno = s+1
+                    transport = self.transport(
+                        skt, protocol, addr, self, s, self.reactor)
+                    protocol.makeConnection(transport)
             else:
                 self.numberAccepts = self.numberAccepts+20
-        except:
+        except BaseException:
```
markrwilliams: This snuck into the diff so `pycodestyle` complained about it.  Maybe it should be `Exception`.

src/twisted/internet/tcp.py
```diff
@@ -1038,42 +1285,62 @@ def doRead(self):
...
-                        # SYN|ACK but before application code calls accept(2).
-                        # On Linux, calling accept(2) on such a listener
-                        # returns a connection that fails as though the it were
-                        # terminated after being fully established.  This
-                        # appears to be an implementation choice (see
-                        # inet_accept in inet/ipv4/af_inet.c).  On macOS,
-                        # such a listener is not considered readable, so
-                        # accept(2) will never be called.  Calling accept(2) on
-                        # such a listener, however, does not return at all.
+                    elif e.args[0] == EMFILE and self._reservedFD.available():
```
markrwilliams: Let's see if the builders confirm `ENOBUFS`.

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +919,81 @@ def createTestSocket(test, addressFamily, socketType):
...
+class _ExhaustsFileDescriptors(object):
+    """
+    A class that triggers C{EMFILE} by creating as many file
+    descriptors as necessary.
+    """
+    _fileDescriptors = attr.ib(default=attr.Factory(list), init=False)
+
+    def exhaust(self):
+        """
+        Open file descriptors until C{EMFILE} is reached.
```
exarkun: I think it could be handy to log the number of descriptors this allocated to reach the limit.

Maybe also worth noting that even if this method raises an exception, it is important to call `release` to undo any partial efforts.

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +919,81 @@ def createTestSocket(test, addressFamily, socketType):
...
+
+    def release(self):
+        """
+        Release all file descriptors opened by L{exhaust}.
+        """
+        for fd in self._fileDescriptors:
+            try:
+                os.close(fd)
+            except Exception:
+                continue
```
exarkun: Hmmm.  `OSError`?  `IOError`?  `EnvironmentError`?  Ahh, the choices.  What exceptions did you anticipate here?  What exceptions did you observe?  How about squashing OSError for EBADF and logging anything else?

Also, separately, maybe `_fileDescriptors` should be cleared by this method, too?  State, eh. :/

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +919,81 @@ def createTestSocket(test, addressFamily, socketType):
...
+        L{_ExhaustsFileDescriptors.release} releases all those
+        acquired.
+        """
+        self.exhauster.exhaust()
+
+        def failsWithEMFILE():
+            open(os.devnull).close()
+
+        exception = self.assertRaises(IOError, failsWithEMFILE)
+        self.assertEqual(exception.errno, EMFILE)
```
exarkun: If either assertion fails, I think all of the descriptors opened by the exhauster remain open and the test process is totally hosed from this point onward.

Perhaps after `exhaust()` this test should `addCleanup(release)` and there can be a second test that focuses on the release logic (then you can remove the "and" clause from the test method description which is usually a good sign).

src/twisted/internet/test/test_tcp.py
```diff
@@ -977,6 +1054,40 @@ def test_allNewStyle(self):
...
+        soft resource limit.
+        """
+        reactor = self.buildReactor()
+        port = self.getListeningPort(reactor, MyServerFactory())
+        listeningHost = port.getHost()
+        clientFactory = MyClientFactory()
+        self.connectToListener(reactor, listeningHost, clientFactory)
+
+        exhauster = _ExhaustsFileDescriptors()
+        reactor.callWhenRunning(exhauster.exhaust)
```
exarkun: This logic depends on the reactor creating the client socket even if the reactor is not running.  All current reactor implementations seem to do this so it seems like the test can get away with this.  I was imagining this test would use the `socket` module directly for the client side.  However, since this works and is less complicated, it makes sense to do it the way you've written.  At worst, someone will have to do some twiddling at some unspecified future point to account for reactor refactoring (as an aside, the way the *Client* classes work, by initiating async I/O operations as a side effect of `__init__` is terrible; ah, 2001...).

src/twisted/internet/test/test_tcp.py
```diff
@@ -977,6 +1054,40 @@ def test_allNewStyle(self):
...
+        clientFactory.failDeferred.addBoth(stopReactorAndCloseFileDescriptors)
+
+        self.runReactor(reactor)
+
+        self.assertNoResult(clientFactory.failDeferred)
+        self.successResultOf(clientFactory.deferred)
+        self.assertRaises(
+            self.lostConnectionReason,
+            clientFactory.lostReason.raiseException,
+        )
```
exarkun: Does this test ever terminate (aside from a `runReactor` timeout) if the desired case isn't entered?  Let's say the exhauster somehow fails to provoke the EMFILE condition (who knows why).  The connection will be accepted and I guess it will just sit there  doing nothing for a while, right?  Perhaps the test could use a server protocol which notes that the test expectations have been violated and ends the test immediately.

Also, in this case, I don't think `exhauster.release()` will be called and any allocated descriptors will leak.

src/twisted/internet/test/test_tcp.py
```diff
@@ -988,6 +1099,24 @@ def getListeningPort(self, reactor, factory, port=0, interface=''):
...
+
+        @param address: The listening's address.
+        @type address: L{IPv4Address} or L{IPv6Address}
+
+        @param factory: The client factory.
+        @type factory: L{ClientFactory}
+
+        @return: The connector
+        """
+        return reactor.connectTCP('127.0.0.1', address.port, factory)
```
exarkun: Would be nice to have this duplication (`ListenTCPMixin.connectToListener`, `SocketTCPMixin.connectToListener`) factored out.

I suppose this hasn't come up before because the existing tests use the `socket` module instead of Twisted's TCP client API (so they just have a bunch of semi-duplicated `socket`-module code).  That makes me think there may have been some explicit decision not to use Twisted's client APIs to test these Twisted server-side behaviors.

Also, if it makes sense to keep this method, it seems like the `address` parameter is a bit misleading since the connection is always to localhost.

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2686,107 @@ def test_resolveIPv6(self):
         # but, luckily, IP presentation format and what it means to be a port
         # number are a little better specified.
         self.assertEqual(result[:2], ("::1", 2))
+
+
+
+class FileDescriptorReservationTests(SynchronousTestCase):
+    """
+    Tests for L{_FileDescriptorReservation}.
+    """
```
exarkun, line 3100: This whole suite seems like a great candidate for Hypothesis-based testing.  Guess Twisted doesn't use Hypothesis for anything yet, though ... ?

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2686,107 @@ def test_resolveIPv6(self):
...
+
+        self.reserveFD = _FileDescriptorReservation(fakeFileFactory)
+
+
+    def test_acquireOpensFileOnce(self):
+        """
+        Multiple acquisitions without releases open the reservation
+        file exactly once.
+        """
+        self.assertEqual(len(self.fileObjects), 0)
```
exarkun: This assertion looks like it snuck in from a (non-existent) `test_initializationDoesntOpenAnything`.

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2686,107 @@ def test_resolveIPv6(self):
...
+        self.assertEqual(len(self.fileObjects), 1)
+        self.assertTrue(self.fileObjects[0].closed)
+
+
+    def test_acquireAvailableReleaseUnavailable(self):
+        """
+        Acquiring the reservation makes it available; releasing it
+        makes it unavailable.
+        """
+        self.assertFalse(self.reserveFD.available())
```
exarkun: Another assertion for a test about initial conditions?

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2686,107 @@ def test_resolveIPv6(self):
...
+        the reservation available.
+        """
+        self.assertFalse(self.reserveFD.available())
+
+        toClose = io.BytesIO()
+
+        self.reserveFD.acquire()
+        self.reserveFD.release()
+
+        self.assertFalse(toClose.closed)
```
exarkun: Not sure how this _could_ ever be closed at this point.  Just being defensive, I guess.

src/twisted/internet/tcp.py
```diff
@@ -921,6 +1166,7 @@ def __init__(self, port, factory, backlog=50, interface='', reactor=None):
             self.addressFamily = socket.AF_INET6
             self._addressType = address.IPv6Address
         self.interface = interface
+        self._reservedFD = _FileDescriptorReservation(lambda: open(os.devnull))
```
exarkun: It seems unnecessarily consumptive of a limited resource to have one of these reserved FDs _per listening port_.  It seems like it would work just as well to share a single reserve between all ports.

src/twisted/internet/tcp.py
```diff
@@ -1038,42 +1285,62 @@ def doRead(self):
...
+                    elif e.args[0] == EMFILE and self._reservedFD.available():
+                        # Linux gives EMFILE when a process is not
+                        # allowed to allocate any more file
+                        # descriptors.  *BSD and Win32 give
+                        # (WSA)ENOBUFS.
+                        log.msg(
+                            "EMFILE encountered;"
+                            " releasing reserved file descriptor.")
+                        self._reservedFD.release()
+                        continue
```
exarkun: On the last iteration of `for i in range(numAccepts):` this `continue` will jump out of the loop which takes it to the end of `doRead` - the reserve FD will not be re-acquired and the port ends up in a weird state that I don't think it can recover from (wrt EMFILE and the reserve fd).

src/twisted/internet/tcp.py
```diff
@@ -853,6 +858,246 @@ def getPeer(self):
...
+
+
+    @_machine.input()
+    def replace(self, fileObject):
+        """
+        Replace a file with the reservation file only if our reserve
+        file is unclaimed.
+
+        @param fileObject: The file object to close.
+        """
```
exarkun: Can you expand on this?  It's not clear to me from this docstring what this method is meant to do.  What does it mean to "replace a file" in this context?

src/twisted/internet/tcp.py
```diff
@@ -14,10 +14,15 @@
 import socket
 import sys
 import operator
+import os
 import struct
 
+import attr
+import automat
```
exarkun: `from automat import MethodicalMachine` would more closely match the style of this module, I think.

src/twisted/internet/tcp.py
exarkun: Preserving the old behavior in _this_ PR seems like the wiser choice to me.  So stick with `BaseException` (I think that's equivalent to bare-except?).  Of course, if the goal is not to change the behavior, not changing the code might be the _best_ choice... and if pycodestyle is making a mistake here (complaining about this pre-existing code), it's making a mistake and we should ignore it (after filing a bug report for the mistake).

src/twisted/internet/tcp.py
```diff
@@ -1038,42 +1285,62 @@ def doRead(self):
...
+                        # allowed to allocate any more file
+                        # descriptors.  *BSD and Win32 give
+                        # (WSA)ENOBUFS.
+                        log.msg(
+                            "EMFILE encountered;"
+                            " releasing reserved file descriptor.")
+                        self._reservedFD.release()
+                        continue
+                    elif e.args[0] in (
+                            EMFILE, ENOBUFS, ENFILE, ENOMEM, ECONNABORTED):
```
exarkun: Totally optional: It might be nice to hoist this tuple to module scope, along with the comment (which would then have a bit more horizontal space and be easier to read).

src/twisted/internet/tcp.py
```diff
@@ -853,6 +858,246 @@ def getPeer(self):
...
 
+def _silenceOutputs(outputs):
+    """
+    An L{automat} output collector that returns L{None}.
+
+    @param outputs: An outputs iterator
+
+    @return: L{None}
+    """
+    list(outputs)
```
exarkun: I read through the automat docs (all of them, I think) to try to understand this line but I didn't find anything that explains why this is a useful thing to do.  Did I miss something from the docs?

If I didn't miss anything, please file a doc issue against automat to explain the issue this code is addressing and also add a comment here meanwhile.

exarkun (changes requested):
Thanks.  This is a very nice start.  The subject is complex but the code is mostly understandable.  There are some issues with the implementation, I think.  Comments inline.  Happy to discuss if you like.

src/twisted/internet/test/test_tcp.py
markrwilliams: This behavior is indeed pretty terrible.  My understanding is that it's at least part of the reason we're trying to move towards endpoints, whose `connect` and `listen` methods return `Deferred`s...

src/twisted/internet/test/test_tcp.py
markrwilliams: > I suppose this hasn't come up before because the existing tests use the socket module instead of Twisted's TCP client API (so they just have a bunch of semi-duplicated socket-module code). That makes me think there may have been some explicit decision not to use Twisted's client APIs to test these Twisted server-side behaviors.

[It seems like other tests used `socket`s directly to minimize the system under test](https://github.com/twisted/twisted/blob/de597aa01ea3e0db7ff95f31b0fecd74318fd285/src/twisted/internet/test/test_tcp.py#L1096-L1131).  I'll admit to using `connectTCP` here out of laziness: it allows me to use `MyClientFactory` which conveniently fires a `Deferred` on `connectionLost`.

> Also, if it makes sense to keep this method, it seems like the address parameter is a bit misleading since the connection is always to localhost.

[This is an unfortunate bit of platform specificity](https://github.com/twisted/twisted/pull/996/commits/94113c4cf9709f479e2e8f5d458da72066384487).  The other option i saw was to change `getListeningPort`'s default interface to `'127.0.0.1'`, which would cause `port.getHost().address` to be connectable on Windows. This seems like the right thing to do since I don't think tests should listen on my public IP address anyway, but I didn't want change an implicit interface that other tests might rely on.

src/twisted/internet/test/test_tcp.py
markrwilliams: [Not yet](https://github.com/twisted/twisted/pull/677).  I'm inclined to address that in a smaller PR...

src/twisted/internet/test/test_tcp.py
markrwilliams: Sorry - this like the precondition in a Hoare triple.  Alternatively: the test is actually asserting that there's a delta of exactly one, so it should check the starting value.

src/twisted/internet/test/test_tcp.py
markrwilliams: [As above](https://github.com/twisted/twisted/pull/996/files/d29f39116cf69586ab814a366acca24f64fddc1a#r181253658).  If repetition doesn't make this pattern defensible I can eliminate it :smile:

src/twisted/internet/test/test_tcp.py
markrwilliams: [As above](https://github.com/twisted/twisted/pull/996/files/d29f39116cf69586ab814a366acca24f64fddc1a#r181253658).

src/twisted/internet/test/test_tls.py
```diff
@@ -324,6 +325,26 @@ def useIt(reactor, contextFactory):
...
+        @param address: The listening's address.
+        @type address: L{IPv4Address} or L{IPv6Address}
+
+        @param factory: The client factory.
+        @type factory: L{ClientFactory}
+
+        @return: The connector
+        """
+        return reactor.connectSSL(
+            '127.0.0.1', address.port, factory, self.getClientContext())
```
markrwilliams: [As above](https://github.com/twisted/twisted/pull/996/files/d29f39116cf69586ab814a366acca24f64fddc1a#r181253167).  These `connectToListener`s are special cases; the interface should in fact accept whatever address came out of `port.getHost()` so that the UNIX domain socket tests work without modification.  The easiest solution is probably a documentation change.

src/twisted/internet/tcp.py
markrwilliams: [Looks like a `pycodestyle` bug](https://github.com/PyCQA/pycodestyle/issues/749).  I'll climb that hill after this one :frowning:

src/twisted/internet/tcp.py
markrwilliams: Oops!  Thanks for catching this implicit dependency.

I see two ways to address this:

1. Add an additional state variable so that a subsequent `doRead` call can know that the reserve file descriptor should be available;
2. Change [`for i in range(numAttempts)`](https://github.com/twisted/twisted/pull/996/files/d29f39116cf69586ab814a366acca24f64fddc1a#diff-d4edf4dd546357655acc84fff0ab7082L1025) to a while loop and increment `numAttempts` here.

The second option is ugly but maybe less so than the first.

src/twisted/internet/tcp.py
markrwilliams: It really means: tell the caller if the provided file was closed because the reservation was made available by a previous call to `release`.  If it wasn't, then the reservation is still held because `EMFILE` has not occurred.

Naming is hard and I'm bad at it.  Maybe `maybeClose` is a better name?

src/twisted/internet/tcp.py
markrwilliams: See also https://github.com/glyph/automat/issues/28

src/twisted/internet/tcp.py
markrwilliams: https://github.com/glyph/automat/issues/18 is half the issue here and all of the issue with `nthOutput`.  This collector also ensures that the transition it's applied to returns `None`; by default, Automat inputs return a (possibly empty) list of their output method's return values, which makes no sense for, say, `_FileDescriptorReservation.acquire`.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +858,246 @@ def getPeer(self):
...
+        """
+        Return True.
+
+        @param fileObject: ignored.
+
+        @return: L{True}
+        """
+        if self._fileDescriptor is None:
+            self._openFailedWithEMFILE()
+            return False
```
markrwilliams: This is wrong for two reasons: the file given to `replace` _was_ closed, so [user code can't use it](https://github.com/twisted/twisted/pull/996/files/d29f39116cf69586ab814a366acca24f64fddc1a#diff-d4edf4dd546357655acc84fff0ab7082R1324), and why on earth would something named `_returnTrue` ever return `False`.

src/twisted/internet/tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/066e1c7a31a9988c7f379b3c204eaa9996d1c78a

src/twisted/internet/tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/124437a0814e35c6dafddfa21ec2f337960af5e1

src/twisted/internet/tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/124437a0814e35c6dafddfa21ec2f337960af5e1

src/twisted/internet/tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/89cc5c3e9115eeaf18fbbed101cb847ba7302f0f

src/twisted/internet/tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/024d8fb8326a2e2cefe13f2dd664a61993261269

src/twisted/internet/test/test_tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/f36b5e91953ca1ea3630d506008de8d3a71444a6

src/twisted/internet/test/test_tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/f36b5e91953ca1ea3630d506008de8d3a71444a6

src/twisted/internet/test/test_tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/f36b5e91953ca1ea3630d506008de8d3a71444a6

src/twisted/internet/test/test_tcp.py
markrwilliams: Adressed in: https://github.com/twisted/twisted/pull/996/commits/f5ffa9d2c3e4cb05690d679219342f3d770b8b69

Is this excessive?  It feels excessive, but it also seems good to test something that makes the tests fail sanely.

src/twisted/internet/test/test_tcp.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/963e05a2ef24c11b5d78cc385868f0d85602aaac

src/twisted/internet/test/test_tls.py
markrwilliams: Addressed in: https://github.com/twisted/twisted/pull/996/commits/963e05a2ef24c11b5d78cc385868f0d85602aaac

markrwilliams (commented):
[test_closePeerOnEMFILE is a flaky test](
https://buildbot.twistedmatrix.com/builders/ubuntu16.04-py2.7-nodeps/builds/1309/steps/select/logs/stdio):

```
[FAIL]
  File "/buildslave/ubuntu16_04-py2_7-nodeps/Twisted/build/py27-nodeps-nocov-posix/local/lib/python2.7/site-packages/twisted/internet/test/test_tcp.py", line 1349, in test_closePeerOnEMFILE
    connect=self.connectToListener,
  File "/buildslave/ubuntu16_04-py2_7-nodeps/Twisted/build/py27-nodeps-nocov-posix/local/lib/python2.7/site-packages/twisted/internet/test/test_tcp.py", line 1184, in assertPeerClosedOnEMFILE
    noResult, "Server accepted connection; EMFILE not triggered.")
  File "/buildslave/ubuntu16_04-py2_7-nodeps/Twisted/build/py27-nodeps-nocov-posix/local/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 384, in assertFalse
    super(_Assertions, self).assertFalse(condition, msg)
  File "/usr/lib/python2.7/unittest/case.py", line 416, in assertFalse
twisted.trial.unittest.FailTest: Server accepted connection; EMFILE not triggered.

twisted.internet.test.test_tcp.TCPFDPortTestsBuilder_EPollReactorTests.test_closePeerOnEMFILE
```

I can reproduce this with `trial -u` and other transport tests; strace implies that `epoll` objects are getting garbage collected.  That is, given:

`strace trial -u twisted.internet.test.test_unix.UNIXPortTestsBuilder_EPollReactorTests.test_closePeerOnEMFILE |& tee fail.log`

This appears in the trace:

```
dup(0)                                  = 1017
dup(0)                                  = 1018
dup(0)                                  = 1019
dup(0)                                  = 1020
dup(0)                                  = 1021
dup(0)                                  = 1022
dup(0)                                  = 1023
dup(0)                                  = -1 EMFILE (Too many open files)
close(5)                                = 0
close(6)                                = 0
close(10)                               = 0
close(12)                               = 0
close(13)                               = 0
close(14)                               = 0
close(15)                               = 0
close(16)                               = 0
stat("/etc/localtime", {st_mode=S_IFREG|0644, st_size=2845, ...}) = 0
write(3, "2018-04-12 17:57:08-0700 [-] EMF"..., 77) = 77
getsockopt(24, SOL_SOCKET, SO_ERROR, [0], [4]) = 0
connect(24, {sa_family=AF_UNIX, sun_path="./tmp2IDgWm.sock"}, 18) = 0
epoll_ctl(17, EPOLL_CTL_ADD, 24, {EPOLLIN, {u32=24, u64=139753940844568}}) = 0
epoll_wait(17, [{EPOLLIN, {u32=23, u64=139753940844567}}], 4, 29977) = 1
accept(23, {sa_family=AF_UNIX}, [110->2]) = 5
```

File descriptors 5, 6, 10, 12, 13, 14, 15, and 16 all represent `epoll` objects, e.g.;

```
...
epoll_create(1024)                      = 15
pipe([16, 17])                          = 0
...
```

The simplest fix I can think of is to put a `gc.collect()` [at the beginning of `test_closePeerOnEMFILE`](https://github.com/twisted/twisted/pull/996/files#diff-f9525083d254dddffa1c74bceca3e8d7R1343).  This seems silly but probably generally the right thing to do, since who knows what files might be lurking. 

`trial -u` with the collection has also successfully completed 10,000 runs without failing.

src/twisted/internet/tcp.py
exarkun: As I understand it, this explains the `return None` (below).  I'm specifically asking about the `list(outputs)` here though.

https://github.com/glyph/automat/issues/30 seems to vaguely suggest that it used to be necessary to consume `outputs` for correct operation but it may no longer be so (but it's not clear to me whether that issue is really fixed or if it was accidentally closed).  A link to _that_ ticket here seems appropriate.

A link to https://github.com/glyph/automat/issues/18 for the `return None` and `_nthOutput` might be good as well.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +859,359 @@ def getPeer(self):
...
+
+        @raises: Any exception except an L{OSError} or L{IOError}
+            whose errno is L{EMFILE}.
+        """
+
+
+    def maybeCloseForReserved(fileObject):
+        """
+        Close the provided file if and only if L{release} has been
+        called to free the reserved file's file descriptor in exchage
```
exarkun: typoe - `exchage`

markrwilliams:
A better citation inside `libuv` for the claim that a reserved file doesn't work on Windows:

https://github.com/libuv/libuv/blob/v1.x/src/win/stream.c#L49-L65

Versus:

https://github.com/libuv/libuv/blob/v1.x/src/unix/stream.c#L540-L552

src/twisted/internet/tcp.py
```diff
@@ -1038,42 +1418,48 @@ def doRead(self):
...
+                        # soft limit of file descriptors.
+                        log.msg(
+                            "EMFILE encountered;"
+                            " releasing reserved file descriptor.")
+                        _reservedFD.release()
+                        # Ensure that we have one more iteration
+                        # available after this.
+                        if i == (numAccepts - 1):
+                            numAccepts += 1
+                        continue
```
exarkun: This looks like it fixes the issue I raised last time.  I'm a bit concerned about the fragility of the fix (with respect to future code changes), though.  The test suite doesn't cover the case where this matters (and it would be tricky to make it and make sure it continues to hit the right case).

exarkun (changes requested):
Thanks.  Couple comments inline.  Ran out of steam before I got to the test suite.

markrwilliams:
@exarkun suggested out-of-band that the `accept` loop inside `Port.doRead` be moved into its own function.  This would in turn allow the reserved file descriptor logic to move closer to the `accept` call and focus the incidental complexity, making the code easier to analyze.  His sketched API:

```python

def accept(reserve, port):
    while True:
        try:
            return port.accept()
        except EMFile:
            with reserve:
                skt, addr = port.accept()
                skt.close()
```

(From https://gist.github.com/exarkun/64fe848c5f8267367e5b01d5c10a93eb)

Note that the context manager would allow `reserved` to re-open itself if `accept` fails with `EAGAIN`, which the current code doesn't.

This seems good so I'm going to implement it.

markrwilliams:
Ask codecov to reconsider.

src/twisted/internet/tcp.py
exarkun: I suppose this and the other automat-related code should be deleted now.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+
+
+
+@implementer(_IFileDescriptorReservation)
+@attr.s
+class _FileDescriptorReservation(object):
+    """
+    L{_IFileDescriptorReservation} implementation.
+
+    @param fileFactory: A factory that will be called to reserve a
```
exarkun: I think this is supposed to be `ivar`.  If it were on `__init__` then it would be `param`.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+        file descriptor.
+    @type fileFactory: A L{callable} that accepts no arguments and
+        returns an object with a C{close} method.
```
exarkun: Does Twisted do typing?  pep 483, 484, 526?  `typing.Callable[[], ICloseable]`?

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+        See L{_IFileDescriptorReservation.reserve}.
+        """
+        if self._fileDescriptor is None:
+            try:
+                fileDescriptor = self._fileFactory()
+            except (IOError, OSError) as e:
+                if e.errno == EMFILE:
+                    self._log.failure(
+                        "Could not reserve EMFILE recovery file descriptor.")
+                    self._fileDescriptor = None
```
exarkun: redundant assignment here, I guess, since the whole thing is guarded by `self._fileDescriptor is None`.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+
+
+    def reserve(self):
+        """
+        See L{_IFileDescriptorReservation.reserve}.
+        """
+        if self._fileDescriptor is None:
+            try:
+                fileDescriptor = self._fileFactory()
+            except (IOError, OSError) as e:
```
exarkun, line 959: Not an actionable item but I just learned that on Python >=3.3 `IOError is OSError`.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+                self._fileDescriptor = fileDescriptor
+
+
+    def __enter__(self):
+        """
+        See L{_IFileDescriptorReservation.__enter__}.
+        """
+        if not self._fileDescriptor:
+            raise RuntimeError(
+                "No file reserved.  Have you called my reserve method?")
```
exarkun: I think `RuntimeError` is meant for "errors in the runtime".  I think it is commonly misunderstood to mean "an error occurred at runtime" and so this intention may be blurred past the point of relevance now.

But ... there is no built-in `StateError`.  So, eh.  The utility of the exception signature of an `__enter__` is questionable anyway, since exceptions from this method can only be handled in the most awkward of ways.

I do have some vague idea about `reserve` returning a new object which is then guaranteed to have a working `__enter__` method.  That raises new questions about object lifecycle management of that new object though so, despite temptations otherwise, what you have here is probably the right path.

Also, separately ... re-entrancy?  We may be okay right now but I'm not sure.  At the moment, the only caller of `__enter__` is `_accept`.  Inside the context managed by this context manager, this code runs:

```
                    log.msg("EMFILE encountered;"
                            " releasing reserved file descriptor.")
                    clientsToClose = _accept(listener, reservedFD)
                    for clientToClose, closedAddr in clientsToClose:
                        clientToClose.close()
                        log.msg("EMFILE recovery:"
                                " Closed socket from %r"
                                " and reopening reserve file descriptor"
                                % (closedAddr,))
                    # All pending client connections have been closed.
                    # Return control back to the reactor.
                    return
```

`log.msg` invokes arbitrary application code.  Elsewhere, I've suggested these logging calls should actually be outside of the managed block.  We might need to generalize that: _no_ arbitrary application code can be invoked from that managed block.  If arbitrary application code runs, it might steal the file descriptors and that's the end of our EMFILE safety (at least temporarily).  That's not _too_ bad since we've already acknowledged this protection is a best-effort not a guarantee.

At the top of this section, I was worried we would end up trying to call `__enter__` inside the managed block.  I don't _see_ a way for that to happen now, even with arbitrary application code invoked (short of manually instantiating and running a new reactor - which is not supported anyway).  Future reactor implementation changes _could_ change this, perhaps.  They might be somewhat far-fetched though.  The reactor event loop is not re-entrant itself so it's _not_ going to call `Port.doRead` again while `Port.doRead` is already on the stack.  So this might all be worrying for nothing ... but it has a fragile feeling to it.  I'd probably at least annotate the `with reserveFD:` block with a comment about avoiding application code (we know we would _like_ to eventually tell the application when this case has been encountered so having a reminder here for whoever works on that would be nice) and avoiding code that might want to open file descriptors.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+
+    def __exit__(self, excValue, excType, traceback):
+        """
+        See L{_IFileDescriptorReservation.__exit__}.
+        """
+        try:
+            self.reserve()
+        except Exception:
+            self._log.failure(
+                "Could not reserve EMFILE recovery file descriptor.")
```
exarkun: Maybe "re-reserve" in this message?

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+        @param excValue: See L{object.__exit__}
+        @param traceback: See L{object.__exit__}
+        """
+
+
+
+if platformType == 'win32':
+    _reservedFD = _NullFileDescriptorReservation()
+else:
+    _reservedFD = _FileDescriptorReservation(lambda: open(os.devnull))
```
exarkun: If, by chance, you've learned anything about Win32 behaviors while developing this branch, this might be a reasonable place to record them.

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+    @type listener: L{socket.socket}
+
+    @param reservedFD: A reserved file descriptor that can be used to
+        recover from C{EMFILE} on UNIX-like systems.
+    @type reservedFD: L{_IFileDescriptorReservation}
+
+    @return: A generator that yields C{(socket, addr)} tuples from
+        L{socket.socket.accept}
+    """
+
```
exarkun: What's with this whitespace?

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+                # Linux and other UNIX-like operating systems return
+                # EMFILE when a process has reached its soft limit of
+                # file descriptors.  The reserved file descriptor is
+                # available, so it can be released to free up a
+                # descriptor for use by listener.accept()'s clients.
+                # Each client socket will be closed until the listener
+                # returns EAGAIN.
+                with reservedFD:
+                    log.msg("EMFILE encountered;"
+                            " releasing reserved file descriptor.")
```
exarkun: I would say ... no logging calls inside this `with` block.  Log observers seem like prime candidates to steal file descriptors?  (Of course, they might really appreciate being able to steal one in order to get these messages out ... but maybe that's *their* problem.)

Also, `twisted.logger` instead of `log.msg`?  I'm not up on the current story about the Twisted-internal log migration.  I would _guess_ "write all new code with `twisted.logger` and gradually port nearby `twisted.python.log` code."

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+                    clientsToClose = _accept(listener, reservedFD)
+                    for clientToClose, closedAddr in clientsToClose:
+                        clientToClose.close()
+                        log.msg("EMFILE recovery:"
+                                " Closed socket from %r"
+                                " and reopening reserve file descriptor"
```
exarkun: The "reopening" part is redundant inside this loop.

src/twisted/internet/tcp.py
```diff
@@ -1022,58 +1315,29 @@ def doRead(self):
...
-                            errorcode[e.args[0]],))
-                        break
-                    raise
 
+            accepted = 0
+            attempts = range(1, numAccepts + 1)
+            clients = _accept(self.socket, _reservedFD)
+            # Limit the infinite stream of clients with the attempts
+            # iterator
+            for accepted, (skt, addr) in izip(attempts, clients):
```
exarkun: `izip(attempts, clients)` looks a lot like `enumerate(clients, 1)` fwiw (and avoids punishing Python 2 w/ an extra list allocation per doRead call (who knows if that's significant)).

src/twisted/internet/tcp.py
```diff
@@ -1022,58 +1315,29 @@ def doRead(self):
...
+                transport = self.transport(
+                    skt, protocol, addr, self, s, self.reactor)
                 protocol.makeConnection(transport)
-            else:
+
+            if not accepted:
                 self.numberAccepts = self.numberAccepts+20
-        except:
+            else:
+                self.numberAccepts = accepted
```
exarkun: So... `if not accepted` - that means "accept() gave us no connections at all" - then increase numberAccepts by 20.  Otherwise ("we got at least one connection from accept()") set numberAccepts to the number of connections we accepted.

I think maybe this isn't what the old code intended.  I _think_ the intent of the old code was "try to accept up to `numberAccepts`; if you accept that many, next time around try accepting up to 20 more; if you accept fewer, try accepting exactly as many as you accepted this time next time".

Or I guess more succinctly:

```
if accepted == numAccepts:
    self.numberAccepts += 20
else:
    self.numberAccepts = max(1, accepted)
```

or at least something close to that.

src/twisted/internet/test/test_tcp.py
```diff
@@ -11,19 +11,25 @@
 __metaclass__ = type
 
 import errno
+import gc
+import io
+import os
```
exarkun, line 16: old macdonald had a farm

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+
+
+@implementer(_IExhaustsFileDescriptors)
+@attr.s
+class _ExhaustsFileDescriptors(object):
+    """
+    A class that triggers C{EMFILE} by creating as many file
+    descriptors as necessary.
+
+    @param fileDescriptorFactory: A factory that creates a new file
```
exarkun: `ivar`?

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+    _log = Logger()
+    _fileDescriptorFactory = attr.ib(default=lambda: os.dup(0), repr=False)
+    _fileDescriptors = attr.ib(
+        default=attr.Factory(list), init=False, repr=False)
+
+    def exhaust(self):
+        """
+        Open file descriptors until C{EMFILE} is reached.
+        """
+        gc.collect()            # Force a collection to close dangling
```
exarkun: Comments above instead of to the right are preferred I think.

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+                    fd = self._fileDescriptorFactory()
+                except (IOError, OSError) as e:
+                    if e.errno == errno.EMFILE:
+                        break
+                    raise
+                else:
+                    self._fileDescriptors.append(fd)
+        except Exception as e:
+            self.release()
+            raise e
```
exarkun: Does this produce a better report?

```
except Exception:
    self.release()
    raise
```

I think that it does on Python 2, at least.  Maybe Python 3 exception chaining makes it do something usable as-written?

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+
+
+    def test_release(self):
+        """
+        L{_ExhaustsFileDescriptors.release} releases all opened
+        file descriptors.
+        """
+        self.exhauster.exhaust()
+        self.exhauster.release()
+        self.openAFile()        # Does not fail with EMFILE
```
exarkun: Comment above instead of to the right.

exarkun: Maybe a useful variation on this test would be to assert that `count()` returns the same number after one and two `exhaust` / `release` pairs?  Or maybe that would be too fragile (for some reason).

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+
+    def test_fileDescriptorsReleasedOnFailure(self):
+        """
+        L{_ExhaustsFileDescriptors.exhaust} closes any opened file
+        descriptors if an exception occurs during its exhaustion loop.
+        """
+        fileDescriptors = []
+        def failsAfterThree():
+            if len(fileDescriptors) == 3:
+                raise ValueError
```
exarkun: `raise ValueError` produces a vexingly untraceable exception.  I suggest putting a distinctive string in here.  For example, `raise ValueError("test_fileDescriptorsReleasedOnFailure fake open exception")`

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+        exhauster = _ExhaustsFileDescriptors(failsAfterThree)
+        self.addCleanup(exhauster.release)
+
+        self.assertRaises(ValueError, exhauster.exhaust)
+        self.assertEqual(len(fileDescriptors), 3)
+        self.assertEqual(exhauster.count(), 0)
+
+        for fd in fileDescriptors:
+            exception = self.assertRaises(OSError, os.close, fd)
+            self.assertEqual(exception.errno, errno.EBADF)
```
exarkun: Kind of risky since those integers technically lose all meaning as descriptors after they get closed by the exhauster...  Probably be okay and at least the assertion will fail if somehow the integers get re-allocated as descriptors in the interim.

Maybe it would be safer to `os.fstat` them instead of `os.close`ing them though.

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+        connect,
+):
+    """
+    Assert that an L{IListeningPort} immediately closes an accepted
+    peer socket when the number of open file descriptors exceeds the
+    soft resource limit.
+
+    @param testCase: The test case under which to run this assertion.
+    @type testCase: L{trial.unittest.SynchronousTestCase} or
+        L{trial.unittest.TestCase}
```
exarkun: `trial.unittest.TestCase` is-a `trial.unittest.SynchronousTestCase` fwiw.

src/twisted/internet/test/test_tcp.py
```diff
@@ -913,10 +931,367 @@ def createTestSocket(test, addressFamily, socketType):
...
+
+    serverConnectionMade.addCallback(
+        stopReactorIfServerAccepted)
+
+    port = listen(reactor, serverFactory)
+    listeningHost = port.getHost()
+    clientFactory = MyClientFactory()
+    connect(reactor, listeningHost, clientFactory)
+
+    reactor.callWhenRunning(exhauster.exhaust)
```
exarkun, line 1250: I'm slightly worried that the connection created by `connect` could be handled by the port before `exhauster.exhaust` runs.  I think this _could_ happen if there were a Deferred-returning "before" "startup" trigger in the reactor because `callWhenRunning` calls the given callable in the "after" "startup" phase.

This could easily be dealt with by using `addSystemEventTrigger("before", "startup", ...)` instead of `callWhenRunning`, I think.

This also guarantees the exhaust happens before the reactor starts handling I/O events (which begin only after all "before" "startup" triggers have been run).

src/twisted/internet/tcp.py
```diff
@@ -853,6 +856,295 @@ def getPeer(self):
...
+                    raise
+            else:
+                self._fileDescriptor = fileDescriptor
+
+
+    def __enter__(self):
+        """
+        See L{_IFileDescriptorReservation.__enter__}.
+        """
+        if not self._fileDescriptor:
```
exarkun: `if self._fileDescriptor is None` for consistency with other new code and more precise statement of the intent.

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2966,197 @@ def test_resolveIPv6(self):
...
+
+
+    def test_enterFailsWithoutFile(self):
+        """
+        A reservation without an open file used as a context manager
+        raises a L{RuntimeError}.
+        """
+        with self.assertRaises(RuntimeError):
+            with self.reservedFD:
+                "Does not run."
```
exarkun: maybe `side_effect = True` w/ an assertion afterwards to be sure (initially thought `assertFalse` but that relies on correct exception propagation behavior which it would be preferable to isolate this test from).

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2966,197 @@ def test_resolveIPv6(self):
...
+        """
+        class TestException(Exception):
+            """
+            An exception only used by this test.
+            """
+
+        self.reservedFD.reserve()
+        with self.assertRaises(TestException):
+            with self.reservedFD:
+                raise TestException
```
exarkun: `raise TestException()`  (raising a class is a historical language mistake remains shockingly unfixed)

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2966,197 @@ def test_resolveIPv6(self):
...
+
+        class SuppressedException(Exception):
+            """
+            An exception raised by the file descriptor factory.
+            """
+
+        called = [False]
+        def failsWithSuppressedExceptionAfterSecondOpen():
+            if called[0]:
+                raise SuppressedException
```
exarkun: `raise SuppressedException()`

src/twisted/internet/test/test_tcp.py
```diff
@@ -2539,3 +2966,197 @@ def test_resolveIPv6(self):
...
+
+        reservedFD = _FileDescriptorReservation(
+            failsWithSuppressedExceptionAfterSecondOpen)
+
+        reservedFD.reserve()
+        self.assertTrue(reservedFD.available())
+
+        with self.assertRaises(AllowedException):
+            with reservedFD:
+                raise AllowedException
```
exarkun: `raise AllowedException()`

src/twisted/internet/test/test_tcp.py
exarkun: Is this test missing an `assertTrue(self.reservedFD.available())` at the end?  It seems like it is otherwise I don't see how it actually verifies the behavior in the docstring.

src/twisted/python/_release.py
```diff
@@ -35,6 +35,7 @@
     u"https://python-hyper.org/h2/en/stable/objects.inv",
     u"https://python-hyper.org/priority/en/stable/objects.inv",
     u"https://zopeinterface.readthedocs.io/en/latest/objects.inv",
+    u"https://automat.readthedocs.io/en/latest/objects.inv",
```
exarkun, line 38: no longer necessary I guess.  OTOH Twisted uses automat elsewhere so it's probably a good thing to have.

exarkun (changes requested):
Thanks!  This was a lot easier to follow (OTOH it's Monday morning instead of Friday afternoon but I'm sure that has nothing to do with it).

Some comments left inline but nothing that will require major changes, I think.

src/twisted/internet/tcp.py
markrwilliams: Not yet, for better and for worse.

src/twisted/internet/tcp.py
markrwilliams: > I think `RuntimeError` is meant for "errors in the runtime". I think it is commonly misunderstood to mean "an error occurred at runtime" and so this intention may be blurred past the point of relevance now.

It's now documented as a catch-all, both on [Python 2](https://docs.python.org/2/library/exceptions.html#RuntimeError) and [Python 3](https://docs.python.org/3/library/exceptions.html#RuntimeError).  When in Rome...

> I do have some vague idea about reserve returning a new object which is then guaranteed to have a working `__enter__` method. That raises new questions about object lifecycle management of that new object though so, despite temptations otherwise, what you have here is probably the right path.

Also, the context this manages is global state and returning a new context manager whose `__enter__` always succeeds would mask that.  It _does_ seem weird, however.

> Also, separately ... re-entrancy? We may be okay right now but I'm not sure. ...
`log.msg` invokes arbitrary application code. Elsewhere, I've suggested these logging calls should actually be outside of the managed block. We might need to generalize that: no arbitrary application code can be invoked from that managed block. If arbitrary application code runs, it might steal the file descriptors and that's the end of our `EMFILE` safety (at least temporarily). That's not too bad since we've already acknowledged this protection is a best-effort not a guarantee.

Allowing the logging system to steal the recovery FD does seem like an unnecessary trap to set, but when I wrote this it seemed more important to report the behaviors as they happened -- probably because I was running manual integration tests.  Now I think it's better to wrap `_accept` in its _own_ context manager that buffers log messages and flushes them in its `__exit__`.  That would lift logging out of this critical section but still ensure that it occurs.

src/twisted/internet/tcp.py
markrwilliams: True, but it would move the termination condition (`if accepted > numAccepts: break`) inside the for loop, which seemed like a potential source of bugs.

src/twisted/internet/tcp.py
markrwilliams: > I think maybe this isn't what the old code intended. I think the intent of the old code was "try to accept up to numberAccepts; if you accept that many, next time around try accepting up to 20 more; if you accept fewer, try accepting exactly as many as you accepted this time next time".

Whoops!  On a closer reading of the original code that seems exactly right:

```python
numberAccepts = self.numberAccepts
...
for i in range(numberAccepts):
    ...
        if e.args[0] in (EWOULDBLOCK, EAGAIN):
            self.numberAccepts = i
            break
    ...
else:
    self.numberAccepts = self.numberAccepts+20
```

(Excerpted from [here](https://github.com/twisted/twisted/blob/dffc6ab317e06dc6bc3dadc1003228c593ea3cc1/src/twisted/internet/tcp.py#L1019-L1075).)

src/twisted/internet/test/test_tcp.py
markrwilliams: It certainly does on Python 2 - this is left over from some other change, I think.

src/twisted/internet/test/test_tcp.py
markrwilliams: How would that differ from https://github.com/twisted/twisted/pull/996/files/#diff-f9525083d254dddffa1c74bceca3e8d7R1066 ?

src/twisted/internet/test/test_tcp.py
markrwilliams: Excellent point - the check shouldn't have side effects on process-global state.

src/twisted/internet/test/test_tcp.py
markrwilliams: Moving this to `before` `startup` prevents the reactor from creating its signal delivery pipe :frowning_face:  That means this test has an implicit dependency on the internal setup of the reactor.

I guess the alternative is to patch `ReactorBuilder.runReactor` to accept `installSignal` handlers as an optional argument, except that might raise spurious warnings on platforms where `signal` isn't available.  Then again, do we support any of those?

FWIW it appears that reactors don't wait for `startup` events before performing I/O:

https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L753
https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L699-L704

Note the absence of `Deferred`s coordinating with `doIteration` et. al.

src/twisted/internet/test/test_tcp.py
markrwilliams: This was a lame trick to avoid lowering coverage.

src/twisted/internet/tcp.py
exarkun: `islice(enumerate(clients, 1), numAccepts)`?  Whether _that's_ actually an improvement over what's checked in now or not is not so clear to me though.  But hey who doesn't love iterators?

src/twisted/internet/test/test_tcp.py
exarkun: Ah, bummer. :/  If these were Hypothesis tests then you could very easily avoid the need  for such things. :wink:

src/twisted/internet/test/test_tcp.py
exarkun: > Moving this to before startup prevents the reactor from creating its signal delivery pipe ￼

Woops!  Guess there's too much random unspecified stuff going on in the reactor to do much better than what's here, then.  Maybe someday.

> FWIW it appears that reactors don't wait for startup events before performing I/O:

True but it does _run_ them.  So it seemed like this would be okay - because `exhaust` doesn't return a Deferred, it just does its job and then it's done.  So not waiting would be okay.  But given the other problems here, it's moot.

exarkun (approved):
Great.  Thanks.  No coverage report posted yet but the tests look good to me.  Please merge when you're satisfied the test coverage is sufficient.

PR #998, opened 2018-04-14
Fix conch MSG_DEBUG parsing on Python 2

cjwatson:
`bool(b'\x00...'[0])` returns `True`, which isn't what's needed here.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9422
* [x] I have created a newsfragment in src/twisted/conch/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )

### DIFF
--- a/src/twisted/conch/ssh/transport.py
+++ b/src/twisted/conch/ssh/transport.py
@@ -967,7 +967,7 @@ class SSHTransportBase(protocol.Protocol):
         @type packet: L{bytes}
         @param packet: The message data.
         """
-        alwaysDisplay = bool(packet[0])
+        alwaysDisplay = bool(ord(packet[0:1]))
         message, lang, foo = getNS(packet[1:], 2)
         self.receiveDebug(alwaysDisplay, message, lang)
 
--- a/src/twisted/conch/newsfragments/9422.bugfix
+++ b/src/twisted/conch/newsfragments/9422.bugfix
@@ -1,0 +1,1 @@
+twisted.conch.ssh.transport.SSHTransportBase now correctly handles MSG_DEBUG with a false alwaysDisplay field on Python 2 (broken since 8.0.0).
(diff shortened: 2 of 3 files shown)


src/twisted/conch/ssh/transport.py
```diff
@@ -967,7 +967,7 @@ def ssh_DEBUG(self, packet):
         @type packet: L{bytes}
         @param packet: The message data.
         """
-        alwaysDisplay = bool(packet[0])
```
markrwilliams, line 970: On Python 3 this was correct because indexing a `bytes` object results in the integer value for that index:

```
>>> b'\x00'[0], b'\x01'[0]
(0, 1)
```

On Python 2, however, `bytes` is `str`, so indexing a `bytes` object returns another `bytes` object:

```
>>> b'\x00'[0], b'\x01'[0]
('\x00', '\x01')
```

Because items in the returned tuple are truthy this always evaluated to `True` on Python 2.  Whoops!

src/twisted/conch/ssh/transport.py
```diff
@@ -967,7 +967,7 @@ def ssh_DEBUG(self, packet):
...
+        alwaysDisplay = bool(ord(packet[0:1]))
```
markrwilliams, line 970: This is correct on both Python 2 and 3 because both return `bytes` for _slices_ of `bytes`, and `ord` returns the ASCII integral value for single-item `bytes` objects on Python 3.

markrwilliams (approved):
:+1: Thanks so much for improving Conch (on Python 2, too!)

markrwilliams:
I've pushed this to a branch of https://github.com/twisted/twisted so the Buildbot builders will run it.  It can be merged after the builds are green (this sometimes takes a few tries due to some unreliable tests.)

PR #1000, opened 2018-04-15
Include Twisted's version in SSH ident string

cjwatson:
* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9424

### DIFF
--- a/src/twisted/conch/test/test_transport.py
+++ b/src/twisted/conch/test/test_transport.py
@@ -7,8 +7,10 @@ Tests for ssh/transport.py and the classes therein.
 
 from __future__ import absolute_import, division
 
-import struct
 import binascii
+import re
+import string
+import struct
 
 from twisted.python.reflect import requireModule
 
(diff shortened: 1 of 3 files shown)


src/twisted/conch/test/test_transport.py
```diff
@@ -456,11 +459,25 @@ class BaseSSHTransportTests(BaseSSHTransportBaseCase, TransportTestCase):
...
+                         b"SSH-2.0-Twisted_" + twisted_version.encode('ascii'))
+        softwareVersion = version.decode('ascii')[len('SSH-2.0-'):]
+        # This is an inefficient regex, but it's simple to build.
+        softwareVersionRegex = (
+            r'^(' +
+            '|'.join(
+                re.escape(c) for c in string.printable
+                if c != '-' and not c.isspace()) +
+            r')*$')
+        self.assertRegex(softwareVersion, softwareVersionRegex)
```
markrwilliams, line 480: This could also be:

```python
self.assertEqual(set(softwareVersion), set(string.printable) - set(string.whitespace))
```

But a failed assertion would probably be less readable.

markrwilliams (approved):
This also looks good - thanks!

src/twisted/conch/test/test_transport.py
cjwatson: Right, that was my thought too.  (A failed assertion with my version is a bit wall-of-regex, but I think it's reasonably possible for somebody unfamiliar with it to work out what it means.)

PR #1001, opened 2018-04-17
Cascading cancellation for inlineCallbacks

altendky:
https://twistedmatrix.com/trac/ticket/4632

Thanks to @glyph for giving me a branch to resurrect.


glyph:
> Thanks to @glyph for giving me a branch to resurrect.

Very happy to do it! Thanks for your contribution.

src/twisted/internet/defer.py
```diff
@@ -1435,40 +1467,74 @@ def gotResult(r):
...
-    inlineCallbacks helps you write L{Deferred}-using code that looks like a
+    L{inlineCallbacks} helps you write L{Deferred}-using code that looks like a
     regular sequential function. For example::
 
         @inlineCallbacks
         def thingummy():
             thing = yield makeSomeRequestResultingInDeferred()
             print(thing)  # the result! hoorj!
 
+    WARNING: this syntax was introduced in Python 2.5, so this example will not
```
markrwilliams: We don't support any Python older than 2.7 so this can go.

src/twisted/internet/defer.py
```diff
@@ -1360,9 +1360,41 @@ def returnValue(val):
 
 
 
-def _inlineCallbacks(result, g, deferred):
+class _CancellationStatus(object):
```
markrwilliams: This could become an `attrs` class now that Twisted depends on it.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
         d = calling()
         f = self.failureResultOf(d)
         self.assertIn("in erroring", f.getTraceback())
+
+
+
+class TranslateMe(Exception):
+    """
+    Sample exception type.
```
markrwilliams: This docstring should explain the intended purpose of this exception.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+    """
+
+
+
+class TranslateResult(Exception):
+    """
+    Sample exception type.
```
markrwilliams: As above.

src/twisted/internet/defer.py
```diff
@@ -1435,40 +1467,74 @@ def gotResult(r):
...
+    """
+    Make an C{@}L{inlineCallbacks} cancellable.
+
+    @param g: a generator object returned by calling a function or method
+        decorated with C{@}L{inlineCallbacks}
+
+    @return: L{Deferred} for the C{@}L{inlineCallbacks} that is cancellable.
+    """
+    def cancel(it):
+        it.callbacks, tmp = [], it.callbacks
```
markrwilliams, line 1503: Putting `handleCancel` at the front of `callbacks` _and_ using `_PeculiarError` seems redundant, because errbacking the deferred with `_PeculiarError` synchronously invokes `handleCancel`.

At the same time trapping only `_PeculiarError` in `handleCancel` asserts that it was synchronously invoked by `cancel`.  `_PeculiarError`'s docstring should explain this.

src/twisted/internet/defer.py
```diff
@@ -1435,40 +1467,74 @@ def gotResult(r):
...
+        it.addErrback(handleCancel)
+        it.callbacks.extend(tmp)
+        it.errback(_PeculiarError())
+    deferred = Deferred(cancel)
+    status = _CancellationStatus(deferred)
+    def handleCancel(result):
+        result.trap(_PeculiarError)
+        status.deferred = Deferred(cancel)
```
markrwilliams: The control flow here is hard enough to follow that `handleCancel` definitely deserves a comment explaining it.  Here's my attempt:

Calling `cancel` on the `Deferred` returned by an `inlineCallbacks` decorated function only has an effect if that `Deferred` has no result.  That in turn is only possible if the decorated function has `yield`ed a `Deferred` that itself has no result yet.  In that case, [`waiting[0]` will be `True` after `inlineCallbacks` attaches `gotResult`, and the `_CancellationStatus` instance's `waitingOn` will be the `yield`ed result-less `Deferred`](https://github.com/altendky/twisted/blob/b021148a1bf4d626b22ccf2b18a253a77b5edb42/src/twisted/internet/defer.py#L1463-L1479).

Cancelling `waitingOn` propagates cancellation to the `yield`ed `Deferred` as intended.  This cancellation either suspends `waitingOn`'s callback processing by delegating execution to another result-less `Deferred` or executes `gotResult` and thus resumes the decorated generator.  Now, the generator might emit a final object that `_inlineCallbacks` has to communicate to the caller, but `_inlineCallbacks` can't [`callback` or `errback` its `_CancellationStatus` instance's `deferred`](https://github.com/altendky/twisted/blob/b021148a1bf4d626b22ccf2b18a253a77b5edb42/src/twisted/internet/defer.py#L1417-L1460) because it was just cancelled.  The solution is to replace the cancelled `deferred` attribute with a _new_ `Deferred` whose cancellation function is `cancel` that's also returned into `waitingOn`'s callback chain.  The generator then resumes with all the degrees of freedom it had prior to its `Deferred`'s cancellation; it can return a value or raise an exception that callbacks or errbacks `status.deferred`, which resumes the previous `Deferred`'s callback chain; it can even `yield` a `Deferred` without a result that can then be cancelled again.

This explains why `waitingOn`, which `_CancellationStatus.__init__` sets to `None`, is always a `Deferred` by the time this runs, and why `status.deferred` has to be replaced by a new `Deferred` in `handleCancel`.  The reason that `handleCancel` has to be at the beginning of the callback chain of the `Deferred` returned by `inlineCallbacks` is a little more complicated.

A `Deferred` waiting on a `Deferred` pauses its own callback chain until that `Deferred` fires and propagates cancellation downwards: 

```python
d = Deferred()
d.addCallback(lambda _: d.Deferred())
d.addErrback(print)
d.callback(None)  # <Deferred at 0x1234 waiting on Deferred at 0x5678>
d.cancel()  # prints [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.defer.CancelledError'>: ]
```

`inlineCallbacks`, however, contains its own trampoline [so that `yield`ing `Deferred`s with results doesn't overflow the stack.](https://github.com/altendky/twisted/blob/b021148a1bf4d626b22ccf2b18a253a77b5edb42/src/twisted/internet/defer.py#L1399-L1402), which prevents its `Deferred` from directly waiting on those `yield`ed by its generator:

```python
@inlineCallbacks
def f():
    d = Deferred()
    print(d)  # prints: <Deferred at 0x1234>
    yield Deferred()

print(f())  # prints: <Deferred at 0x5678>; note that it isn't waiting on 0x1234!
```

This is indeed the reason cancellation isn't propagated.  Mimicking the exact behavior of `Deferred`s outside of `inlineCallback`s would be difficult; at what point in the callback chain should execution be suspended?  Fortunately the nature of `inlineCallback`s makes the answer easy.  Because it's not possible to add callbacks to the `Deferred` that run before the generator is resumed, propagating cancellation downward should always happen _before_ any callbacks added to the `inlineCallback`s `Deferred`.  That's why `cancel` splices `handleCancel` in at the beginning of `callbacks`.

src/twisted/internet/defer.py
```diff
@@ -1435,40 +1467,74 @@ def gotResult(r):
...
+        awaited.cancel()
+        return status.deferred
+    _inlineCallbacks(None, g, status)
     return deferred
 
 
 
+class _PeculiarError(Exception):
+    """
+    A distinctive exception, only raised by cancelling an inline callback.
```
markrwilliams: See above about explaining this exception's purpose.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+    """
+
+    def __init__(self, actual):
+        Exception.__init__(self)
+        self.actualValue = actual
+
+
+
+class CancellationTests(TestCase):
```
markrwilliams: This should be [`SynchronousTestCase`](https://twistedmatrix.com/documents/18.4.0/api/twisted.trial.unittest.SynchronousTestCase.html) because none of the tests use the global reactor.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+        - G be a generator decorated with C{inlineCallbacks}
+        - D be a L{Deferred} returned by G
+        - C be a L{Deferred} awaited by G with C{yield}
+    """
+
+    def setUp(self):
+        """
+        Set up the list of outstanding L{Deferred}s.
+        """
+        self.thingsOutstanding = []
```
markrwilliams: This needs a more descriptive name than `thing`.  Why not `deferredsOutstanding`?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+
+
+    def tearDown(self):
+        """
+        If any L{Deferred}s are still outstanding, fire them.
+        """
+        while self.thingsOutstanding:
+            self.thingGotten()
```
markrwilliams: `deferred` instead of `thing`?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+
+
+    @inlineCallbacks
+    def sampleInlineCB(self, getChildThing=None):
```
markrwilliams: `Deferred` instead of `Thing`?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+        try:
+            x = yield getChildThing()
+        except TranslateMe:
+            raise TranslateResult()
+        except DontFail as df:
+            x = df.actualValue - 2
+        returnValue(x + 1)
+
+
+    def getThing(self):
```
markrwilliams: `Deferred` instead of `Thing`?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+        A sample function that returns a L{Deferred} that can be fired on
+        demand, by L{CancellationTests.thingGotten}.
+
+        @return: L{Deferred} that can be fired on demand.
+        """
+        self.thingsOutstanding.append(Deferred())
+        return self.thingsOutstanding[-1]
+
+
+    def thingGotten(self, result=None):
```
markrwilliams: `Deferred` instead of `Thing`.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+            d = Deferred()
+            def _eb(result):
+                childResultHolder[0] = result.check(CancelledError)
+                return result
+            d.addErrback(_eb)
+            return d
+        d = self.sampleInlineCB(getChildThing=getChildThing)
+        d.addErrback(lambda result: None)
+        d.cancel()
+        self.assertEqual(
```
markrwilliams: You could use `failureResultOf` instead of `childResultHolder`:

```python
child = defer.Deferred()
d = self.sampleInlineCB(getChildDeferred=lambda: child)
d.cancel()
self.assertRaises(CancelledError, self.failureResultOf(child).raiseException)
```

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+            childResultHolder[0],
+            CancelledError,
+            "no cascade cancelling occurs",
+        )
+
+
+    def test_trapChildCancelledErrorOnCascadeCancelling(self):
```
markrwilliams: Where is `CancelledError` trapped?  It looks like this actually tests that the cascaded `CanelledError` is sent back up to the `inlineCallbacks` `Deferred` where it can be suppressed with an errback.  As a result, there's no logged error about an unhandled error in the `Deferred`.  That wouldn't happen anyway, though, because `d` doesn't go out of scope until after the logging assertion.

The name should be changed to reflect what this actually tests.  It might also be clearer to use `failureResultOf` here:

```python
d = self.sampleInlineCB()
d.cancel()
self.assertRaises(CancelledError, self.failureResultOf(d).raiseException)
```

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+        Exception.__init__(self)
+        self.actualValue = actual
+
+
+
+class CancellationTests(TestCase):
+    """
+    Tests for cancellation of L{Deferred}s returned by L{inlineCallbacks}.
+    For each of these tests, let:
+        - G be a generator decorated with C{inlineCallbacks}
```
markrwilliams: This is informative notation! :100:

It might be nice to wrap the capital letters in `C{}`.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,181 @@ def calling():
...
+        moreDeferred = Deferred()
+        def deferMeMore(result):
+            result.trap(CancelledError)
+            return moreDeferred
+        def deferMe():
+            d = Deferred()
+            d.addErrback(deferMeMore)
+            return d
+        d = self.sampleInlineCB(getChildThing=deferMe)
+        finalResult = []
```
markrwilliams: This test might be clearer with `assertNoResult` and `successResultOf`:

```python
moreDeferred = Deferred()
def deferMeMore(result):
    result.trap(CancelledError)
    return moreDeferred
def deferMe():
    d = Deferred()
    d.addErrback(deferMeMore)
    return d
d = self.sampleInlineCB(getChildThing=deferMe)
d.cancel()
self.assertNoResult(d)
moreDeferred.callback(6543)
self.assertEqual(self.successResultOf(d), 6544)
```

markrwilliams (changes requested):
This is _fantastic_ work.  Thank you so much for reviving the branch.

The implementation itself looks solid and the tests correspond to the different possible code paths, as far as I understand them.  Please address the comments about improving the names of the tests and docstrings and resubmit for review.

src/twisted/internet/defer.py
altendky: So `_PeculiarError` isn't really needed?  It's just a double check?  That's what I'm getting from the code and your explanation.  If true, do we really want to keep it?

src/twisted/internet/defer.py
altendky: I'm guessing you don't want this whole dissertation in there so I made a (massively) abbreviated addition after re-understanding this code (thanks).  Let me know if you did want a good chunk more of this explanation included.

src/twisted/internet/defer.py
markrwilliams: The code as written here doesn't seem to need it, no.  But `callbacks` is a public API so who knows what can happen to it; maybe it's not actually a list but some other weird data structure that a user has monkeypatched in.

In general I favor defensive programming and this in particular seems like a good example of it, so I'm in favor of keeping it.  Its redundancy did give me pause while reviewing so I thought I'd mention it.  I think your docstring updates are sufficient to largely quell my concern.

src/twisted/internet/defer.py
```diff
@@ -1503,12 +1503,23 @@ def cancel(it):
...
         it.addErrback(handleCancel)
         it.callbacks.extend(tmp)
-        it.errback(_PeculiarError())
+        it.errback(_InternalInlineCallbacksCancelledError())
     deferred = Deferred(cancel)
     status = _CancellationStatus(deferred)
     def handleCancel(result):
-        result.trap(_PeculiarError)
+        """
+        Propogate the cancellation of an C{@}L{inlineCallbacks} to the
```
markrwilliams: Typo; should be "propagate".

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -145,3 +147,186 @@ def calling():
...
+            childResultHolder[0],
+            CancelledError,
+            "no cascade cancelling occurs",
+        )
+
+
+    def test_errbackCancelledErrorOnCancel(self):
+        """
+        When C{D} cancelled, CancelledError from cascade cancelled C{C} will be
+        trapped
```
markrwilliams: Nothing is trapped here; the `CancelledError` is sent from `D` to `C`, which doesn't do anything with it, so it's then sent back to `D`.

src/twisted/internet/defer.py
```diff
@@ -1512,6 +1589,12 @@ def thingummy():
         def loadData(url):
             response = yield makeRequest(url)
             return json.loads(response)
+    You can cancel the L{Deferred} returned from your L{inlineCallbacks}
+    generator before it is fired by your generator completing (either by
+    reaching its end, a C{return} statement, or by calling L{returnValue}).
+    This will cause any L{Deferred} being awaited upon in a C{yield} statement
+    to be cancelled, with whatever effects would result from that (often a
+    C{CancelledError} being raised from the C{yield}, but not always).
```
markrwilliams: This is vague.  A `CancelledError` will be raised from the `yield`ed `Deferred` that has been canceled if that `Deferred` does not otherwise suppress it.

src/twisted/internet/defer.py
```diff
@@ -1435,33 +1468,77 @@ def gotResult(r):
...
+        it.callbacks.extend(tmp)
+        it.errback(_InternalInlineCallbacksCancelledError())
+    deferred = Deferred(cancel)
+    status = _CancellationStatus(deferred)
+    def handleCancel(result):
+        """
+        Propogate the cancellation of an C{@}L{inlineCallbacks} to the
+        L{Deferred} it is waiting on.
+
+        :param result: An L{_InternalInlineCallbacksCancelledError} from
```
markrwilliams: Oops - this is reStructuredText and not epytext.  Change this to:

```
@param result:
```

src/twisted/internet/defer.py
```diff
@@ -1435,33 +1468,77 @@ def gotResult(r):
...
+            C{cancel()}.
+        :return: A new L{Deferred} that the C{@}L{inlineCallback} generator
```
markrwilliams: RST instead of epytext.  Change to

```
@return:
```

markrwilliams (changes requested):
A few remaining documentation issues, especially the use of RST instead of epytext which [has caused the twistedchecker build to fail](https://travis-ci.org/twisted/twisted/jobs/377737360#L568).

Thanks again for your contribution and your patience!

src/twisted/internet/defer.py
markrwilliams: The documentation as you've written it, is, I think, sufficient.  `inlineCallbacks` might deserve its own design document somewhere that makes clear what my comment tried to make clear, but that would be a separate ticket and PR.

altendky:
That's just bad.  Submitting a PR for review with a broken build?  :[  Sorry.

altendky:
@markrwilliams, `twisted.python.test.test_sendmsg.SendmsgTests.test_flags` and `twisted.python.test.test_sendmsg.SendmsgTests.test_shortsend` failed in LInux.

https://travis-ci.org/twisted/twisted/jobs/377986053#L3372-L3392

Using `trial -u` I ran `test_shortsend` 36k times(then `^C`) without failure.  Moved on to running the module `twisted.python.test.test_sendmsg`.  Up to 6k runs of the module without failure, I'll leave it running today.  Normally I'd write this off as a random error, but given the things that this PR touches it seems prudent to take it a bit more seriously (at least from my not-so-familiar perspective).

markrwilliams:
@altendky We have enough unreliable tests that it's not certain when builds will succeed or fail :disappointed: 

I've never seen anything in `test_sendmsg` fail like this, though, and I also can't reproduce the issue locally (under Python 3.5 with `-u`).  I noted the failed tests in [our build reliability placeholder](https://twistedmatrix.com/trac/ticket/8879#comment:12) and restarted the Travis job.

markrwilliams (approved):
Thanks for working on the docstrings.

I've pushed the branch to https://github.com/twisted/twisted so the Buildbot builders will run the code.  Let's see what happens with `test_sendmsg` there.

PR #1002, opened 2018-04-17
Remove or update references to Python 3.3

altendky:
https://twistedmatrix.com/trac/ticket/9425

### DIFF
--- a/src/twisted/newsfragments/9425.removal
+++ b/src/twisted/newsfragments/9425.removal
@@ -1,1 +1,0 @@
-Remove and update 3.3 related items in twisted.internet.defer.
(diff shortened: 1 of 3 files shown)


src/twisted/newsfragments/9425.removal
```diff
@@ -0,0 +1 @@
+Remove and update 3.3 related items in twisted.internet.defer.
```
markrwilliams: This should be a `misc` ticket because https://twistedmatrix.com/trac/ticket/9352 already mentioned that 3.3 is unsupported.

markrwilliams (approved):
It looks like this removes all the 3.3 references in `defer`.  Thanks!

PR #1003, opened 2018-04-18
A general description of release blockers, courtesy of @exarkun.

markrwilliams:
* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9426

### DIFF
--- a/docs/core/development/policy/release-process.rst
+++ b/docs/core/development/policy/release-process.rst
@@ -388,6 +388,7 @@ Open questions
 
 - What bugs should be considered release blockers?
 
+  - Generally, fixes for regressions since the last release in the series and difficult-to-mitigate problems with the release toolchain
   - All bugs with a type from the release blocker family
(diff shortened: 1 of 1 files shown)


docs/core/development/policy/release-process.rst
```diff
@@ -388,6 +388,7 @@ Open questions
 
 - What bugs should be considered release blockers?
 
+  - Generally, fixes for regressions since the last release in the series and difficult-to-mitigate problems with the release toolchain
```
adiroiban, line 391: It is strange that we start with "generally" :)

And why a "fix" is a blocker. I was expecting that the bug is the blocker.

I don't understand what "difficult-to-mitigate problems" are.

Can you give an example?

adiroiban (changes requested):
At least for me, this new item in the list does not make things better.

I am confised by fix vs bug part.

Can you please rephrase it?

Thanks!

PR #1008, opened 2018-05-06
9438 py2 bytes directory listing

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9438

### DIFF
--- a/src/twisted/web/static.py
+++ b/src/twisted/web/static.py
@@ -277,8 +277,8 @@ class File(resource.Resource, filepath.FilePath):
         Return a resource that generates an HTML listing of the
         directory this path represents.
 
-        @return: A L{resource.Resource} that renders the directory to
-            HTML.
+        @return: A resource that renders the directory to HTML.
+        @rtype: L{DirectoryLister}
         """
         if _PY3:
             path = self.path
--- a/src/twisted/web/static.py
+++ b/src/twisted/web/static.py
@@ -263,9 +263,31 @@ class File(resource.Resource, filepath.FilePath):
     childNotFound = resource.NoResource("File not found.")
     forbidden = resource.ForbiddenResource()
 
+
     def directoryListing(self):
-        return DirectoryLister(self.path,
-                               self.listNames(),
+        """
+        Return a resource that generates an HTML listing of the
+        directory this path represents.
+
+        @return: A L{resource.Resource} that renders the directory to
+            HTML.
+        """
+        if _PY3:
+            path = self.path
+            names = self.listNames()
+        else:
+            # DirectoryLister works in terms of native strings, so on
+            # Python 2, ensure we have a bytes paths for this
+            # directory and its contents.  We use the asBytesMode
+            # method inherited from FilePath to ensure consistent
+            # encoding of the actual path.  This returns a FilePath
+            # instance even when called on subclasses, however, so we
+            # have to create a new File instance.
+            nativeStringPath = self.createSimilarFile(self.asBytesMode().path)
+            path = nativeStringPath.path
+            names = nativeStringPath.listNames()
+        return DirectoryLister(path,
+                               names,
                                self.contentTypes,
                                self.contentEncodings,
                                self.defaultType)
--- a/src/twisted/web/test/test_static.py
+++ b/src/twisted/web/test/test_static.py
@@ -177,10 +177,6 @@ class StaticFileTests(TestCase):
 
         response = child.render(request)
         self.assertIsInstance(response, bytes)
-    if sys.getfilesystemencoding().lower() not in ('utf-8', 'mcbs'):
-        test_emptyChildUnicodeParent.skip = (
-            "Cannot write unicode filenames with file system encoding of"
-            " %s" % (sys.getfilesystemencoding(),))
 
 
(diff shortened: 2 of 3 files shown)


src/twisted/web/static.py
```diff
@@ -271,9 +271,31 @@ def ignoreExt(self, ext):
...
 
+
     def directoryListing(self):
-        return DirectoryLister(self.path,
-                               self.listNames(),
+        """
+        Return a resource that generates an HTML listing of the
+        directory this path represents.
+
+        @return: A L{resource.Resource} that renders the directory to
```
glyph: Given that `DirectoryLister` is public, this could refer directly to that.

src/twisted/web/static.py
```diff
@@ -271,9 +271,31 @@ def ignoreExt(self, ext):
...
+            HTML.
+        """
+        if _PY3:
+            path = self.path
+            names = self.listNames()
+        else:
+            # DirectoryLister works in terms of native strings, so on
```
glyph, line 287: Do we not need to do the inverse (`.asTextMode()`) on PY3 then?

src/twisted/web/test/test_static.py
```diff
@@ -155,6 +155,34 @@ def test_emptyChild(self):
...
+        self.assertIsInstance(child, static.DirectoryLister)
+
+        nativePath = compat.nativeString(textBase.path)
+        self.assertEqual(child.path, nativePath)
+
+        response = child.render(request)
+        self.assertIsInstance(response, bytes)
+    if sys.getfilesystemencoding().lower() not in ('utf-8', 'mcbs'):
+        test_emptyChildUnicodeParent.skip = (
+            "Cannot write unicode filenames with file system encoding of"
```
glyph: ITYM "non-ASCII"?  Given that the `u''` child is the one we care about, I'm pretty sure that will survive most boneheaded filesystem encodings, and this skip could be removed?

glyph (approved):
The new test looks good, and the proposed change fixes it in a way which seems reasonably compatible.

If you want to address any of the minor feedback inline, feel free to do so either as a tweak before landing or as a follow-up.

src/twisted/web/static.py
markrwilliams: No, but for a bad reason: `File` and `DirectoryLister` are tricky.

[`getChild` now decodes all descendant paths to text](https://github.com/twisted/twisted/blob/twisted-18.4.0/src/twisted/web/static.py#L301-L309), so that by the time the empty path is reached and `directoryListing` called, the path given to `DirectoryLister` is guaranteed to be text, _unless_ of course the path being listed is the `File`'s _own_ path.

Given:

```python
from twisted.web import server, static
from twisted.internet import defer, endpoints, task


@task.react
def main(reactor):
    site = server.Site(static.File(b"/path/to/some/directory/"))
    listener = endpoints.serverFromString(reactor, "tcp:8080")
    d = listener.listen(site)

    @d.addCallback
    def _(_):
        return defer.Deferred()

    return d
```

Assuming that `/path/to/some/directory/child` exists, accessing `http://localhost:8080/child/` or any other descendant creates a `DirectoryLister` with a text path.  Accessing `http://localhost:8080/`, however, creates one with a _bytes_ path.  

Surprise! On Python 3, this is fine.  That's `DirectoryLister` _itself_  [decodes bytes descendants to text](https://github.com/twisted/twisted/blob/twisted-18.4.0/src/twisted/web/static.py#L1004-L1006) on Python 3.

So the _actual_ admitted by `DirectoryLister` is:

| `DirectoryLister` parameter  | Python 3         | Python 2        |
|------------------------------|------------------|-----------------|
| pathname                     | `str` or `bytes` | `str`           |
| dirs                         | `str` or `bytes` | `str`           |
| contentTypes                 | `dict` of `str`  | `dict` of `str` |
| contentEncodings             | `dict` of `str`  | `dict` of `str` |
| defaultType                  | `str`            | str             |

Which is not exactly what I said in the commit message, but does mean this part of the PR is correct (I think.)

src/twisted/web/test/test_static.py
markrwilliams: This is a safeguard I carried over from [another test](https://github.com/twisted/twisted/blob/7ae65314dcc50979e10f70d6ad4dfd1f67d06786/src/twisted/web/test/test_static.py#L312-L339), because who knows what [sys.getfilesystemencoding](https://docs.python.org/2/library/sys.html#sys.getfilesystemencoding), which `FilePath` uses, is going to return.  But as you point out, it should be impossible that `SynchronousTestCase.mkstemp().decode(sys.getfilesystemencoding()) + u''` could return an unencodable path, so this skip is unnecessary.

src/twisted/web/static.py
markrwilliams: Why is it public though. :sob:

src/twisted/web/static.py
glyph: Every day I'm a little more sympathetic towards web2's abandonment of this API :).

whalebot-helmsman:
Is there a plan to release new twisted version with this fix?

PR #1009, opened 2018-05-06
test_closePeerOnEMFILE allows ConnectionLost on client's end.

markrwilliams:
Whether the client's connectionLost method is called with
ConnectionDone or ConnectionLost depends on its internal state, which
the test cannot directly control; for example, a TLS handshake might
not have completed by the time the server disconnects the client.

Make the test more robust by allowing both ConnectionDone and
ConnectionLost.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9439#ticket


exarkun (approved):
Thanks.  This seems reasonable to me.  One adjustment you might like is to use `ConnectionClosed` instead of `(ConnectionDone, ConnectionLost)`.  Feel free to make that change or not and then merge.

PR #1010, opened 2018-05-10
9441 conch public key

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9441


adiroiban (approved):
Thanks for fixing this.

Looks good to merge.

I have reverted the fix and run only the tests and I can see a failure.

In terms of testing, I was expecting to see another test with dedicated docstring like

```
def test_publicAlreadyPublic():
    """
    For public keys, it returns the same instance.
    """
    publicRSAKey = keys.Key.fromString(keydata.publicRSA_openssh)
    self.assertIs(publicRSAKey, publicRSAKey.public())
```

wiml:
@adiroiban : I think that having `public_key()` return `self` is only an implementation detail; at least, that's what the documentation says. I don't think the unit tests should be asserting something that the documentation explicitly says you shouldn't rely on.

adiroiban:
> I don't think the unit tests should be asserting something that the documentation explicitly says you shouldn't rely on.

True.. Thanks. Merged

PR #1012, opened 2018-05-14
[#9448] Correct multipart handling on Python 3.7

hawkowl:
https://twistedmatrix.com/trac/ticket/9448

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -90,7 +90,7 @@ from zope.interface import Attribute, Interface, implementer, provider
 
 # twisted imports
 from twisted.python.compat import (
-    _PY3, long, unicode, intToBytes, networkString, nativeString)
+    _PY3, long, unicode, intToBytes, networkString, nativeString, _PY37PLUS)
 from twisted.python.deprecate import deprecated
 from twisted.python import log
 from twisted.logger import Logger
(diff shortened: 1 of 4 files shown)


src/twisted/web/http.py
```diff
@@ -861,30 +861,56 @@ def requestReceived(self, command, path, version):
...
                         self.args.update({x.encode('iso-8859-1'): y
                                           for x, y in cgiArgs.items()})
+                    elif _PY37PLUS:
+                        # The parse_multipart function on Python 3.7+
+                        # decodes the header bytes as iso-8859-1 and
+                        # decodes the body bytes as utf8 with
+                        # surrogateescape -- we want bytes
+                        self.args.update({
+                            x.encode('iso-8859-1'): \
+                            [z.encode('utf8', "surrogateescape") for z in y]
```
markrwilliams: Body components that represent files are still returned as bytes.  You can confirm this with the following unpleasant snippet:

```python
import io
import pprint
import cgi
from http import server
from http import HTTPStatus
import socketserver


PORT = 8001

class Handler(server.BaseHTTPRequestHandler):
    content = b'''
<html>
<form action="/" method="post" enctype="multipart/form-data">
What files are you sending? <INPUT type="file" name="files"><BR>
<input type="submit">
</form>
'''

    def do_GET(self):
        self.send_response(HTTPStatus.OK)
        self.send_header("Content-type", "text/html")
        self.send_header("Content-length", str(len(self.content)))
        self.end_headers()
        self.wfile.write(self.content)

    def do_POST(self):
        _, pdict = cgi.parse_header(self.headers['Content-type'])
        pdict['CONTENT-LENGTH'] = self.headers['Content-length']
        pdict = {key: value.encode('charmap') for key, value in pdict.items()}
        parsed = cgi.parse_multipart(self.rfile, pdict)
        bio = io.StringIO()
        pprint.pprint(parsed, stream=bio)
        self.wfile.write(bio.getvalue().encode('utf-8'))


with socketserver.TCPServer(("", PORT), Handler) as httpd:
    print("serving at port", PORT)
    httpd.serve_forever()
```

That means you'll need to do something like `[z.encode('utf-8', 'surrogate-escape') if isinstance(z, str) else z for z in y]`.

markrwilliams (dismissed):
Thanks for working on this!  It looks like the code also needs to handle bytes in the parsed multipart data.

src/twisted/web/http.py
hawkowl: sob let me fix it

src/twisted/web/test/test_http.py
```diff
@@ -1871,6 +1871,43 @@ def process(self):
...
+"""
+
+        req = '''\
+POST / HTTP/1.0
+Content-Type: multipart/form-data; boundary=---------------------------738837029596785559389649595
+Content-Length: ''' + str(len(body.replace(b"\n", b"\r\n"))) + '''
+
+
+'''
+        print(str(len(body)))
```
markrwilliams: A wild `print` appeared!

src/twisted/web/test/test_http.py
```diff
@@ -1871,6 +1871,43 @@ def process(self):
         self.assertEqual(processed[0].args, {b"text": [b"abasdfg"]})
 
 
+    def test_multipartFileData(self):
+        """
+        If the request has a Content-Type of C{multipart/form-data}, and the
+        form data is parseable, the form arguments will be added to the
+        request's args.
```
markrwilliams: Something about how `input`s of type `file` in particular are added to the request's args would be good.

markrwilliams (approved):
Thanks for fixing this!

Please address the comments before merging.

PR #1013, opened 2018-05-15
[#9450] port twisted.internet.address to attrs

hawkowl:
https://twistedmatrix.com/trac/ticket/9450#ticket

### DIFF
--- a/src/twisted/internet/address.py
+++ b/src/twisted/internet/address.py
@@ -134,11 +134,15 @@ class UNIXAddress(object):
             return res
     else:
         def __eq__(self, other):
-            return self.name == other.name
+            if isinstance(other, self.__class__):
+                return self.name == other.name
+            return False
 
 
     def __ne__(self, other):
-        return not self.__eq__(other)
+        if isinstance(other, self.__class__):
+            return not self.__eq__(other)
+        return True
 
 
     def __repr__(self):
--- a/src/twisted/internet/address.py
+++ b/src/twisted/internet/address.py
@@ -137,6 +137,10 @@ class UNIXAddress(object):
             return self.name == other.name
 
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
     def __repr__(self):
         name = self.name
         if name:
--- a/src/twisted/internet/address.py
+++ b/src/twisted/internet/address.py
@@ -18,17 +18,9 @@ from twisted.python.runtime import platform
 from twisted.python.compat import _PY3
 
 
-_IPAddress = attr.make_class("_IPAddress", {
-    "type": attr.ib(validator=attr.validators.in_(["TCP", "UDP"])),
-    "host": attr.ib(),
-    "port": attr.ib()
-})
-
-
-
 @implementer(IAddress)
 @attr.s(hash=True)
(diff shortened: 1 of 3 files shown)


src/twisted/internet/address.py
```diff
@@ -7,79 +7,60 @@
...
-
-
-    def __repr__(self):
-        return '%s(%s, %r, %d)' % (
-            self.__class__.__name__, self.type, self.host, self.port)
-
-
-    def __hash__(self):
-        return hash((self.type, self.host, self.port))
+_IPAddress = attr.make_class("_IPAddress", {
```
hynek: Why make_class and not just define the class? To avoid having to write a docstring? :)

hynek: Also, does it make sense to even have _IPAddress if you’re using attrs? You’re saving two lines in each class definition but have scattered information.

src/twisted/internet/address.py
```diff
@@ -117,66 +95,32 @@ class HostnameAddress(FancyEqMixin, object):
...
 
     if getattr(os.path, 'samefile', None) is not None:
         def __eq__(self, other):
             """
-            Overriding C{FancyEqMixin} to ensure the os level samefile
+            Overriding C{attrs} to ensure the os level samefile
             check is done if the name attributes do not match.
             """
-            res = super(UNIXAddress, self).__eq__(other)
+            res = self.name == other.name
```
hynek: You need to check the class type first. I presume that’s what the super call did before and why CI is failing.

src/twisted/internet/address.py
```diff
@@ -188,6 +132,17 @@ def __eq__(self, other):
                     if not _PY3 and not platform.isLinux():
                         raise e
             return res
+    else:
+        def __eq__(self, other):
+            if isinstance(other, self.__class__):
```
hynek, line 136: Basically this.

hynek (dismissed):

src/twisted/internet/address.py
hawkowl: fix'd

src/twisted/internet/address.py
hawkowl: type is a keyword :)

hynek: Nah, it’s fine:

```
In [1]: import attr

In [2]: @attr.s
   ...: class C:
   ...:     type = attr.ib()
   ...:

In [3]: C(42)
Out[3]: C(type=42)
```

src/twisted/internet/address.py
```diff
@@ -117,66 +92,35 @@ class HostnameAddress(FancyEqMixin, object):
...
-        On UNIX, paths are always bytes. However, as paths are L{unicode} on
-        Python 3, and L{UNIXAddress} technically takes a file path, we convert
-        it to bytes to maintain compatibility with C{os.path} on Python 3.
-        """
-        if name is not None:
-            self._name = _asFilesystemBytes(name)
-        else:
-            self._name = None
-
+    name = attr.ib(converter=lambda x: _asFilesystemBytes(x)
```
hynek: You can simplify this with http://www.attrs.org/en/stable/api.html#attr.converters.optional

PR #1016, opened 2018-05-16
[#9442] fix lastColorizedLine regression

hawkowl:
https://twistedmatrix.com/trac/ticket/9442

### DIFF
--- a/src/twisted/newsfragments/9442.bugfix
+++ b/src/twisted/newsfragments/9442.bugfix
@@ -1,0 +1,1 @@
+twisted.python.htmlizer.TokenPrinter now explicitly works on bytestrings.
--- a/src/twisted/python/htmlizer.py
+++ b/src/twisted/python/htmlizer.py
@@ -21,7 +21,7 @@ class TokenPrinter:
 
     currentCol, currentLine = 0, 1
     lastIdentifier = parameters = 0
-    encoding = b"utf-8"
+    encoding = "utf-8"
 
     def __init__(self, writer):
         """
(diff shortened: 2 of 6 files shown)


src/twisted/python/htmlizer.py
```diff
@@ -21,7 +21,7 @@ class TokenPrinter:
 
     currentCol, currentLine = 0, 1
     lastIdentifier = parameters = 0
-    encoding = "utf-8"
+    encoding = b"utf-8"
```
markrwilliams: You can't use a byte string with encode on Python 3, right?  So this is only valid in Python 2 where `b"utf-8"` is the same thing as `"utf-8"`.

Furthermore, [it looks like `tokenize` on Python 3 always emits an `ENCODING` token](https://docs.python.org/3/library/tokenize.html#tokenize.ENCODING):

> The first token returned by tokenize() will always be an ENCODING token.

Overall It looks like the effect of this `bytes` literal and the `hasattr` check below would be more clearly express with `if _PY3`.

src/twisted/newsfragments/9442.bugfix
```diff
@@ -0,0 +1 @@
+twisted.python.htmlizer.TokenPrinter now explicitly works on bytestrings.
```
markrwilliams, line 1: This is definitely true and the underlying cause, but it'd be good to mention `lastColorized`, because it is the issue that led to the report.

markrwilliams (approved):
The use of Python 2 and 3 specific features is less clear than an explicit `_PY3` check; I don't think these differences in behavior are common enough knowledge that their use is obvious.

Also, the news fragment should address the reported issue in addition to the underlying cause.  Kudos for fixing a general problem though!

Please address these issues as you see fit before merging.

src/twisted/newsfragments/9442.bugfix
hawkowl: That's in another topfile :)

markrwilliams: Whoops, sorry!  So it is!

Nothing like a fix of several fixes - yay!

PR #1017, opened 2018-05-16
[#9454] Support MSTS

hawkowl:


hawkowl:
As per @glyph at PyCon: I am unilaterally merging this, as this does not meaningfully alter Twisted.

PR #1018, opened 2018-05-20
[#9449] IPv6 address changes in 3.7

hawkowl:
https://tm.tl/9449

### DIFF
--- a/src/twisted/internet/iocpreactor/tcp.py
+++ b/src/twisted/internet/iocpreactor/tcp.py
@@ -12,8 +12,9 @@ from zope.interface import implementer, classImplements
 from twisted.internet import interfaces, error, address, main, defer
 from twisted.internet.protocol import Protocol
 from twisted.internet.abstract import _LogOwner, isIPv6Address
-from twisted.internet.tcp import _SocketCloser, Connector as TCPConnector
-from twisted.internet.tcp import _AbortingMixin, _BaseBaseClient, _BaseTCPClient
+from twisted.internet.tcp import (
+    _SocketCloser, Connector as TCPConnector, _AbortingMixin, _BaseBaseClient,
+    _BaseTCPClient, _resolveIPv6, _getsockname)
 from twisted.python import log, failure, reflect
 from twisted.python.compat import _PY3, nativeString
 
@@ -449,7 +450,7 @@ class Port(_SocketCloser, _LogOwner):
                                             self.socketType)
             # TODO: resolve self.interface if necessary
             if self.addressFamily == socket.AF_INET6:
-                addr = socket.getaddrinfo(self.interface, self.port)[0][4]
+                addr = _resolveIPv6(self.interface, self.port)
             else:
                 addr = (self.interface, self.port)
             skt.bind(addr)
@@ -540,12 +541,11 @@ class Port(_SocketCloser, _LogOwner):
 
     def getHost(self):
         """
-        Returns an IPv4Address.
+        Returns an IPv4Address or IPv6Address.
 
         This indicates the server's address.
         """
-        host, port = self.socket.getsockname()[:2]
-        return self._addressType('TCP', host, port)
+        return self._addressType('TCP', *_getsockname(self.socket))
 
 
     def cbAccept(self, rc, data, evt):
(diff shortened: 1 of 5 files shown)


src/twisted/conch/scripts/ckeygen.py
```diff
@@ -189,7 +189,7 @@ def changePassPhrase(options):
             'Enter file in which the key is (%s): ' % filename)
     try:
         key = keys.Key.fromFile(options['filename'])
-    except keys.EncryptedKeyError as e:
+    except keys.EncryptedKeyError:
```
glyph: This seems… extraneous, but I'll allow it :)

src/twisted/internet/tcp.py
```diff
@@ -100,6 +100,39 @@ class _TLSServerMixin(object):
 _portNameType = (str, unicode)
 
 
+def _getrealname(addr):
+    """
+    Return a 2-tuple of socket IP and port for IPv4 and a 4-tuple of
+    socket IP, port, flowInfo, and scopeID.  With IPv6 addresses, this
+    preserves the interface portion.
```
glyph: Out of "socket IP, port, flowInfo, and scopeID", none of those things are obviously called "interface".  Can you clarify just a little as to what is being preserved, and why this is important (or what wouldn't preserve it?)  I think for future maintainers understanding that this is papering over a 3.6 / 3.7 difference is probably pretty important.

src/twisted/internet/iocpreactor/tcp.py
```diff
@@ -580,16 +580,24 @@ def handleAccept(self, rc, evt):
                 rAddr = (nativeString(rAddr[0]), rAddr[1])
             assert family == self.addressFamily
 
+            # Build a proper IPv6 address structure, if it exists
```
glyph: By "proper" I think you mean "including scope ID, if necessary"?

src/twisted/internet/tcp.py
```diff
@@ -429,7 +462,12 @@ def _setRealAddress(self, address):
             and the 'host' portion will always be an IP address, not a DNS
             name.
         """
-        self.realAddress = address
+        if len(address) == 4:
+            # IPv6, make sure we have the interface portion
```
glyph: ITYM "scope ID"

src/twisted/internet/tcp.py
```diff
@@ -100,6 +100,39 @@ class _TLSServerMixin(object):
...
+    Return a 2-tuple of socket IP and port for IPv4 and a 4-tuple of
+    socket IP, port, flowInfo, and scopeID.  With IPv6 addresses, this
+    preserves the interface portion.
+
+    @param addr: A 2-tuple for IPv4 information or a 4-tuple for IPv6
+        information.
+    """
+    if len(addr) == 4:
+        # IPv6
+        host = socket.getnameinfo(addr, socket.NI_NUMERICHOST)[0]
```
glyph: I think you really want `socket.NI_NUMERICHOST | socket.NI_NUMERICSERV`; you don't need to do service-name lookup, so this is asking the OS to do some pointless I/O on a part of the address you'll immediately throw away.

glyph (approved):
I'd really like it if the language in some of these comments were tightened up, and explicit references to the specific API change in 3.7 were added so it's a little clearer what information we're preserving and why.  Scope IDs are a weird beast and it's good to be precise when talking about them - the scope ID in an IPv6 address is very much *not quite* the same thing as a physical network interface (although they do frequently correspond, and sometimes interface names are valid scope IDs).

Should the `.misc` be a non-`.misc` that says something about 3.7 compatibility, or is this something that's obscure enough that most people won't notice?

One thing I'm also curious about, though, is why you're so interested in preserving the scope ID within the address literal portion of the 4-tuple, given that the scope ID is *already present* as an integer at the end of that 4-tuple.

However, I'm willing to take it on faith that there's a good compatibility reason you needed to do this, and things break on 3.7 without it.  The test coverage is sufficient and the logic looks correct.  So if you could just please add this info, correct any "interface -> scopeID" language where appropriate, and then land that would be great.

src/twisted/conch/scripts/ckeygen.py
hawkowl: failing txchecker because it hates me >:C

PR #1019, opened 2018-06-02
7927 unhandled error tb

jlitzingerdev:
This changeset addresses ticket 7927 [1].  Much of this code started with a PR for ticket 9333 [2], now closed as a duplicate.  In the interest of preserving the review comments and feedback, the original commit changes and messages are preserved, while the content has been rebased onto trunk.  I'm happy to clean the history up as needed.

Another key part of this is the creation of a new API instead of modifying formatEvent() directly.  The rationale is that modifying formatEvent will likely result in duplicate tracebacks for any users who have made changes to append tracebacks to the exising result of formatEvent.

[1] https://twistedmatrix.com/trac/ticket/7927
[2] https://github.com/twisted/twisted/pull/939

### DIFF
--- a/src/twisted/logger/__init__.py
+++ b/src/twisted/logger/__init__.py
@@ -49,6 +49,7 @@ __all__ = [
     # From ._format
     "formatEvent", "formatEventAsClassicLogText",
     "formatTime", "timeFormatRFC3339",
+    "formatEventWithTraceback",
 
     # From ._flatten
     "extractField",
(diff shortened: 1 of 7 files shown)


jlitzingerdev:
FWIW, I'm able to reproduce the pyflakes3 error locally with tox on trunk.

glyph:
@jlitzingerdev Yeah, the pyflakes error is a misconfiguration of the circle builder, not this branch's fault. We need to be pinning versions there.

PR #1021, opened 2018-06-03
9461 pydoctor warnings

twm:
Clear some pydoctor/epytext warnings I noticed while working on #1020.  The pywin32 errors appear be legitimate but not fixable, as the ActiveState docs there have disappeared: http://docs.activestate.com/activepython/2.7/pywin32/PyWin32.html is a 404.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9461


adiroiban:
Thanks for looking into this.
AFAIK we have a test for documentation and the build should fail if we have pydoctor warnings.
Maybe we can take this opportunity to also fix the test :)

adiroiban:
So it looks like there is no test to check that we don't introduce new warnings in future PR.
I remember that I worked on this at some time... but not sure how and why it was changed.

----

Or maybe it was done only for buildbot

In buildbot I see that we got errors in step 6, yet the build is green

https://buildbot.twistedmatrix.com/builders/documentation/builds/4014

------------


Note that now we also run epydoc build on Circle-CI so maybe as part of TDD process we can look at updating `tox -r -e apidocs` to fail.

What do you think?

twm:
@adiroiban I actually wrote the ticket for this with a deliberately narrow scope. Basically, I wanted to fix the warning I introduced.

Since pydoctor warnings are issued based on external resources over which we have no control, I'm not too keen on making them break the build. I feel that it is important that we seek and destroy false positives. I mean, there hasn't been much merged here in the last month, has there? I note that Travis started reporting random failures of `twisted.python.test.test_sendmsg.CModuleSendmsgTests.test_shortsend` about a month ago. A broken build is discouraging to contributors.

Overall, I would prefer that the perfect not be the enemy of the good. Do these changes look okay to merge?

src/twisted/newsfragments/9461.doc
```diff
@@ -0,0 +1 @@
+Several minor formatting problems in the API documentation have been corrected.
```
glyph, line 1: This might be more a `.misc`, to be honest.

glyph (approved):
This looks to be unambiguously correct. Please land at your convenience!

PR #1023, opened 2018-06-04
9463 _sslverify uses md5 to generate unique session identifiers

mstojcevich:
https://twistedmatrix.com/trac/ticket/9463

Switch to random session identifier generation

Fixes problems w/ FIPS-mode OpenSSL and increases uniqueness of session identifiers.

## Contributor Checklist:

### DIFF
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -38,6 +38,8 @@ from twisted.python.util import FancyEqMixin
 
 from twisted.python.deprecate import deprecated
 
+
+
 class TLSVersion(Names):
     """
     TLS versions that we can negotiate with the client/server.
@@ -172,7 +174,10 @@ def _selectVerifyImplementation():
     return simpleVerifyHostname, SimpleVerificationError
 
 
+
 verifyHostname, VerificationError = _selectVerifyImplementation()
+
+
 
 class ProtocolNegotiationSupport(Flags):
     """
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -5,7 +5,6 @@
 
 from __future__ import division, absolute_import
 
-import itertools
 import warnings
 
 from constantly import Names, NamedConstant
@@ -177,15 +176,6 @@ from twisted.python.failure import Failure
 from twisted.python.util import FancyEqMixin
 
 from twisted.python.deprecate import deprecated
-
-
-def _sessionCounter(counter=itertools.count()):
-    """
-    Private - shared between all OpenSSLCertificateOptions, counts up to
-    provide a unique session id for each context.
-    """
-    return next(counter)
-
 
 
 class ProtocolNegotiationSupport(Flags):
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -9,7 +9,7 @@ import itertools
 import warnings
 
 from constantly import Names, NamedConstant
-from hashlib import md5
+from hashlib import md5, sha256
 
 from OpenSSL import SSL, crypto
 from OpenSSL._util import lib as pyOpenSSLlib
@@ -1698,9 +1698,9 @@ class OpenSSLCertificateOptions(object):
 
         if self.enableSessions:
             name = "%s-%d" % (reflect.qual(self.__class__), _sessionCounter())
-            sessionName = md5(networkString(name)).hexdigest()
+            sessionName = sha256(networkString(name)).digest()
 
-            ctx.set_session_id(sessionName.encode('ascii'))
+            ctx.set_session_id(sessionName)
 
         if self.dhParameters:
             ctx.load_tmp_dh(self.dhParameters._dhFile.path)
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -9,13 +9,14 @@ import itertools
 import warnings
 
 from constantly import Names, NamedConstant
-from hashlib import md5, sha256
+from hashlib import md5
 
 from OpenSSL import SSL, crypto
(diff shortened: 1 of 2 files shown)


src/twisted/newsfragments/9463.feature
```diff
@@ -0,0 +1 @@
+twisted.internet._sslverify now uses SHA256 instead of MD5 for ssl session identifier generation by default.
```
exarkun: Public names are preferred for news fragments.  The subject should be `twisted.internet.ssl.CertificateOptions`, I think.

twm:
Hi @mstojcevich,

Is this ready for review? It may be sitting here because there is no `review` keyword on the ticket, so it doesn't show up in [the review queue](https://twisted.reviews).

mstojcevich:
@twm Yes it is, thanks for letting me know!

src/twisted/internet/_sslverify.py
```diff
@@ -1684,9 +1684,9 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
 
         if self.enableSessions:
             name = "%s-%d" % (reflect.qual(self.__class__), _sessionCounter())
-            sessionName = md5(networkString(name)).hexdigest()
+            sessionName = sha256(networkString(name)).digest()
 
-            ctx.set_session_id(sessionName.encode('ascii'))
+            ctx.set_session_id(sessionName)
```
alex, line 1695: Is there a reason not to just generate random session identifiers: `os.urandom(16)`?

glyph (dismissed):
@mstojcevich While I'd love to just land this as-is and eliminate another usage of MD5, I don't want to replace one piece of cryptogibberish with another.  I like @alex's suggestion that we just use completely random IDs (although I sure wish I could find a page telling me about best practices with respect to TLS session IDs...).  Could you implement that instead?

src/twisted/internet/_sslverify.py
mstojcevich: From what I can tell, not really. The OpenSSL docs mention that there are problems if this is not unique to the application (https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set_session_id_context.html), but I don't see any reason why a random byte sequence and this hash would be more/less application-unique than each other.

Presumably the current strings (a counter based on how many sessions have been created), don't need to be reproducible -- the only benefit that I can think of to using the hash here is that that the first _n_ session IDs used by Twisted will always be the same (but that could also be spun as a downside). The docs also mention that conflicting IDs can cause problems, so I'm curious why this hasn't been a problem before w/ multiple twisted clients in different processes negotiating session IDs w/ the same server.

I'll update this to use random numbers, which I prefer as a solution, but I'm still uneasy due to not fully understanding why this was a hash in the first place. The original commit appears to be lost to time (oldest I can find w/ this file is a merge of an svn branch which I can't find).

exarkun: Here's the first committed revision that included this code:

https://github.com/twisted/vertex/commit/389656bed398f2d4940cb3de1e2c10ee6204c7aa#diff-e2026d749e1abadafbd27bb502844bdcR670

It doesn't shed much light, though.

mstojcevich: It sounds like the worst case of this being non-deterministic is that the session will have to be renegotiated, which isn't a huge deal.
Having a conflicting random generation is extremely unlikely as this will be 32 bytes long, so nothing to really worry about there.
I'll go ahead and make this randomly generate the id.

akgood: I spent some time digging into OpenSSL docs and source code (*not* my favorite pastime, mind) to try to figure out what's actually going on here. It turns out that what we're dealing with here is not the actual TLS SessionID, but the OpenSSL "Session ID Context". (That is, the PyOpenSSL `set_session_id` method is a misnomer, as it in turn calls `SSL_CTX_set_session_id_context`). The OpenSSL manpages do a poor job of explaining what this is; the most coherent explanation I could find is in the O'Reilly OpenSSL book:

> All session must have a session ID context. For the server, session caching is disabled by default unless a call to `SSL_CTX_set_session_id_context` is made. The purpose of the session ID context is to make sure the session is being reused for the same purpose for which it was created. For instance, a session created for an SSL web server should not be automatically allowed for an SSL FTP server. Along the same lines, we can use session ID contexts to exercise a finer-grained control over sessions within our application. For example, authenticated clients could have a different session ID context than unauthenticated ones. The context itself can be any data we choose...

So, to put it another way, we can think of the "session ID context" as a namespace for OpenSSL sessions. As the quote suggests, we probably don't normally want to allow sessions to be resumed across different `SSL_CTX` instances, so making this a unique value per context, by default, seems preferable. As Marcus said, using random generation (assuming a well-behaved CSPRNG) with a length of 32 bytes will accomplish this with very high probability.

mstojcevich:
Any updates on this? No rush, just want to make sure it doesn't slip through the cracks since it's a month old now.

twm (approved):
It looks like all of @glyph's concerns have been addressed here. As far as I can see the only thing preventing merge is some linting failures. Unfortunately the goalposts seem to have moved while this PR was waiting for re-review. :(

I've done a pr_as_branch but it didn't pass when I did so the other day so I don't think it will this time. Unfortunately I can't actually merge the PR with required checks failing (I don't mean as a matter of process, I literally can't as I am not an admin). Otherwise I'd fix the failures myself. Perhaps someone with more powers will do so?

You've been quite patient with us, @mstojcevich, so thank you for that and all your contributions!

mstojcevich:
Thanks for the review! Rebased and hopefully fixed the linter errors (having issues running the linter locally right now though.)

twm:
Thanks @mstojcevich! I will go ahead and merge.

PR #1024, opened 2018-06-05
convert return to bytes, because can`t concatenate string with bytes

januszmarchlewski:
there was a problem with concatenate bytes :  cal.encode('utf-8')  with  "</pre>" string, so b"</pre>" is working


twm:
Hi @jasiu889, thank you for taking the time to fix this! However, we do have a [review process](https://twistedmatrix.com/trac/wiki/ReviewProcess). Could you create a Trac ticket for this so it shows up in the [review queue](https://twistedmatrix.com/trac/report/25)? This change will also need a `.doc` newsfile.

wiml (commented):
This change looks good to me (the example can't run on py3 without it). Agree that the Twisted review process asks for some administrivia that isn't here.

wiml:
I've created a Trac ticket ([#9559](https://twistedmatrix.com/trac/ticket/9559)) and a `pr_as_branch` push to [`9559-jasiu889-patch-2`](https://github.com/twisted/twisted/tree/9559-jasiu889-patch-2), which should take care of the administrivia

wiml (approved):

twm:
Thanks for moving this forward, @wiml! And particularly thank you for making be aware that we can push arbitrary changes to PR'd branches (with the "allow maintainers" box checked I presume), that will help a lot. I will go ahead and merge this.

PR #1025, opened 2018-06-06
[#9446] iocp reactor is duplicating/loosing outgoing data

asodeur:
`iocpreactor.abstract.FileHandle.doWrite()` can get called again before `._handleWrite()` got the chance to update the bytes already written if `.write()` gets called while a `.writeToHandle()` is still in flight.

- Changed `.startWriting()` to only schedule a call to `._resumeWriting()` when `self.writing == False`.
- Made`test_tcp.LargeBufferTests` write the last byte in a separate `.write()`.


wiml:
This looks good to me, although I'm not familiar with the IOCP API at all.

To make sure I understand the code, it looks like `.writing` is supposed to be true if a write is scheduled via `_writeScheduled` *or* if a write is in flight via the completion port callback + deferred chain + etc.. It seems to me that there are some paths through `doWrite()` which could leave `writing` true even if nothing is in progress, which would cause the file handle to become unusable. Does that seem like the case to you?

(If so, perhaps it's worth fixing it in a separate PR so that this one can be committed…)

oberstet:
fwiw, would be cool to land this, as the `select` reactor is limited in number of connections and the `iocp` reactor has this issue on high load (eg https://github.com/crossbario/crossbar/issues/1347) - leaving windows/crossbar users (outside of dev/test use) currently sad ..

src/twisted/test/test_tcp.py
```diff
@@ -1427,11 +1433,15 @@ def testWriter(self):
...
 
         d = defer.gatherResults([wrappedF.deferred, wrappedClientF.deferred])
+
         def check(ignored):
             self.assertTrue(f.done, "writer didn't finish, it probably died")
-            self.assertTrue(clientF.len == self.datalen,
+            self.assertTrue(clientF.len >= self.datalen,
                             "client didn't receive all the data it expected "
                             "(%d != %d)" % (clientF.len, self.datalen))
+            self.assertTrue(clientF.len <= self.datalen,
```
adiroiban, line 1443: What is the point of this assertion?
Why the previous one was not good? :)

src/twisted/test/test_tcp.py
```diff
@@ -1363,15 +1363,21 @@ def test_buildProtocolClient(self):
...
     # massive string to make sure Twisted deals with this fact.
 
     def connectionMade(self):
         # write 60MB
-        self.transport.write(b'X'*self.factory.len)
-        self.factory.done = 1
-        self.transport.loseConnection()
+        self.transport.write(b'X'*(self.factory.len-1))
+
+        def finish():   # write last byte separately to check issue #9446
```
adiroiban: I guess that the inline command can be set as a docstring for this function.
Rather than "check issue #9446", maybe this command can describe the targeted behaviour regardless of the ticket number.

src/twisted/test/test_tcp.py
```diff
@@ -1363,15 +1363,21 @@ def test_buildProtocolClient(self):
 
 
 class LargeBufferWriterProtocol(protocol.Protocol):
-
     # Win32 sockets cannot handle single huge chunks of bytes.  Write one
     # massive string to make sure Twisted deals with this fact.
 
     def connectionMade(self):
         # write 60MB
```
adiroiban: do we still write 60MB? it looks like there are 60MB-1 ... I guess that we can remove this comment

oberstet:
cool, thanks for merging! also pinged OP to check if the issue is actually gone as it surfaced at the user level (crossbar in this case) and `iocp` now works also under high load (backpressure)

src/twisted/test/test_tcp.py
asodeur: Current IOCP reactor shows rather interesting behaviour. ~10% of runs are ok, ~80% will loose data, and the remaining ~10% return duplicate data. Of course a simple `==` assertion would do, but the error message should not state data is missing.

src/twisted/test/test_tcp.py
asodeur: commited change

src/twisted/test/test_tcp.py
asodeur: commited change

src/twisted/test/test_tcp.py
adiroiban: maybe the assertion message can be changed from `didn't received all the data` to `didn't receive the expected data` ... and replace with the dedicated self.assertEqual 

but this is minor comment ...

what I would like to see, is this tests executed without the patch applied and see that it constantly fails
and then when applying the patch we have a solid green test :)

asodeur: Making test fail with probability 1 on the unpatched IOCP reactor will be hard  IMHO. The issue is with the interaction of twisted with the Windows network stack. You had to mock that to get deterministic behaviour. You could run the test multiple times which would obviously reduce chances of false passes.

That said, [this](https://github.com/asodeur/twisted/commit/ca7e96360a1de70cc364c6eb5b0c0b21d75ca4df) is the commit that changed the test. Everything else (including the IOCP reactor) is still unchanged. You might give that a spin to get an idea of the frequency of false passes ...

glyph:
It sounds like we've gotten lots of feedback on this and it improves the situation, everybody feels like it's good, but nobody wants to approve it.

Sounds like it's time for somebody to be the hero that Twisted needs.

src/twisted/test/test_tcp.py
glyph: A mock that emulates the behavior of the IOCP APIs under their various failure modes *would* be a welcome addition, especially to help other maintainers who may want to clean things up but don't know as much about IOCP.  That said, it sounds like that would be a good followup.

glyph (approved):
I'm going to merge this once CI is happy.

PR #1027, opened 2018-06-12
[#9468] protect against double request-finished in twcgi

chrisbarber:
(edit) link to Trac ticket now that I was able to login:

https://twistedmatrix.com/trac/ticket/9468

===

I do not have an existing account on twisted trac, and the [login page](https://twistedmatrix.com/trac/login) is giving me the error: "No handler matched request to /login".

I'm not sure if this fix is appropriate; I just adapted it from twisted.web.wsgi. But I can explain the issue in some detail.

Running a twisted cgi resource (inside crossbar.io, in this case) with a simple script, and GETting it from a browser, cURL, etc. caused no problems. Requesting the page from an f5 load balancer "health check" caused the following traceback (not causing any consequences in this case other than polluting the crossbar log):
```
2018-06-11T15:27:26-0700 [Router      31860] unexpected error in processEnded
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/internet/process.py", line 64, in reapAllProcesses
    process.reapProcess()
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/internet/process.py", line 350, in reapProcess
    self.processEnded(status)
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/internet/_baseprocess.py", line 52, in processEnded
    self.maybeCallProcessEnded()
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/internet/process.py", line 987, in maybeCallProcessEnded
    _BaseProcess.maybeCallProcessEnded(self)
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/internet/_baseprocess.py", line 64, in maybeCallProcessEnded
    proto.processEnded(Failure(reason))
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/web/twcgi.py", line 321, in processEnded
    self.request.unregisterProducer()
  File "/local1/miniconda3/envs/datacube/lib/python3.6/site-packages/twisted/web/http.py", line 913, in unregisterProducer
    self.channel.unregisterProducer()
builtins.AttributeError: 'NoneType' object has no attribute 'unregisterProducer'
```
This can be reproduced by requesting the page like this from python:
```python
import socket
clientsocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
clientsocket.connect(('localhost', 8080))
clientsocket.send('GET /cgi/apptest.py HTTP/1.1\r\n\r\n')
ret = clientsocket.recv(10000)
print("Returned\n" + ret)
```
Sleeping for some time at the end of the script fixes the problem so it seems to be down to timing, and from what I can tell `Request.connectionLost` (which sets the request's `channel` to `None`) is probably getting called before `CGIProcessProtocol.processEnded` (which calls `Request.unregisterProducer` that will fail when `channel` is `None`).

It seems to me that the ending bit of `CGIProcessProtocol.processEnded` is meant to only be called once on the request, so it made sense to adapt the pattern for this from twisted.web.wsgi.

If this fix is on the right track let me know; I'm happy to go through the remaining steps (although I could use a little advice on how to write a test for this-- maybe there is a similar test elsewhere I could use as a model?). If I can get into trac I can write up the issue over there.

## Contributor Checklist:


rodrigc:
The `/login` link you tried to click on is broken.  Try this one instead: https://twistedmatrix.com/trac/github/login

twm:
These changes look correct to me, though I'm not too familiar with this part of the codebase. I'm going to put this into review in Trac so that it appears in the review queue.

src/twisted/web/twcgi.py
```diff
@@ -311,12 +313,24 @@ def processEnded(self, reason):
...
+                    resource.ErrorPage(http.INTERNAL_SERVER_ERROR,
+                        "CGI Script Error", "Premature end of script headers."
+                    ).render(self.request))
+
+        if not self._requestFinished:
+            self.request.unregisterProducer()
+            self.request.finish()
+
+
+    def _finished(self, ignored):
```
glyph, line 331: I would prefer it if we defined callbacks that are only used once like this as inner methods rather than instance methods to illustrate where they can be used, and to avoid polluting the namespace on an (unfortunately) inheritable class.

src/twisted/web/twcgi.py
```diff
@@ -202,6 +202,7 @@ class CGIProcessProtocol(protocol.ProcessProtocol, pb.Viewable):
     headertext = b''
     errortext = b''
     _log = Logger()
+    _requestFinished = False
```
glyph, line 205: It would be nice to document this with an `@ivar`, and populate it on the instance rather than the class.

src/twisted/web/test/test_cgi.py
```diff
@@ -485,6 +486,20 @@ def test_prematureEndOfHeaders(self):
...
+        Ensure that the CGI process ends cleanly when the request connection
+        is lost.
+        """
+        d = DummyChannel()
+        request = http.Request(d, True)
+        protocol = twcgi.CGIProcessProtocol(request)
+        request.connectionLost(
+            failure.Failure(ConnectionLost("Connection done"))
+        )
+        protocol.processEnded(failure.Failure(error.ProcessTerminated()))
```
glyph, line 500: Ideally tests should always have at least one assertion; I realize that this is just a regression test for something that previously caused an exception, so I am not going to be a stickler about it, but it would ideally verify *something* about the behavior of `processEnded`.

glyph (approved):
Looks like a good fix to me, so I'll land it.  Minor feedback inline just for learning purposes :)

PR #1029, opened 2018-06-22
Potential race condition in _startRunCallbacks

zachlowry:
When creating a Deferred in a separate thread, then chaining that deferred into the callback chain of another Deferred, and then adding that deferred to a DeferredList to wait for completion, _runCallbacks can be executed on the chained deferred before the result has been set. This change seems to prevent that issue. I've attached some example code and an example traceback of the error to ticket #9469.


exarkun:
See comment on ticket:9469 (basically, Deferred does not even try to be threadsafe).

PR #1030, opened 2018-06-25
9470 Update combinedLogFormatter for getClientIP deprecation

twm:
`IRequest.getClientIP()` was deprecated in #981. Twisted 18.4.0 logs a deprecation warning on every request because `combinedLogFormatter` was using this API instead of the new `getClientAddr()` method. It took me a while to notice this, as I don't find combined logging terribly useful and disable it in most of my services, but here is a fix nonetheless.

All of the logging test cases use a subclass of `requesthelper.DummyRequest`, so I altered that subclass to have a `getClientIP()` which raises &mdash; nothing inside Twisted should use the deprecated API, after all. I am not sure what effect this will have on coverage, but I see the same technique used in the same file, so... :crossed_fingers: 

## Contributor Checklist:

* [x] There is an associated ticket in Trac: https://twistedmatrix.com/trac/ticket/9470#trac-add-comment

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -2639,6 +2639,7 @@ def combinedLogFormatter(timestamp, request):
     return line
 
 
+
 @implementer(interfaces.IAddress)
 class _XForwardedForAddress(object):
     """
@@ -2652,6 +2653,7 @@ class _XForwardedForAddress(object):
     """
     def __init__(self, host):
         self.host = host
+
 
 
 class _XForwardedForRequest(proxyForInterface(IRequest, "_request")):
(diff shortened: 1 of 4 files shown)


markrwilliams (approved):
Thanks for cleaning up after #981 - I'm sure Buildbot will appreciate it!

Please address [the style issues](https://travis-ci.org/twisted/twisted/jobs/396260978#L554) and merge when green.

twm:
I updated https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted to include pycodestylediff so that these style issues stop sneaking in. Next step: teach my editor about Twisted's style guide.

Thank you for the review, @markrwilliams!

glyph:
FWIW @twm you missed a spot here.  `_XForwardedForRequest`'s docstring still talks about `getClientIP` even though its implementation correctly uses `getClientAddress` :)

twm:
Thanks @glyph, I filed https://twistedmatrix.com/trac/ticket/9491

PR #1034, opened 2018-07-01
Fix reverse proxy resource docstring

dmollerm:
Fix reverse proxy resource docstring and example snippet accordingly
This change reveals a bug when the resource tries to
encode the host, solved as well

## Contributor Checklist:

* [x] There is an associated ticket in Trac: https://twistedmatrix.com/trac/ticket/9192


dmollerm:
AFAIU this PR solves this other ticket as well: https://twistedmatrix.com/trac/ticket/8836

dmollerm:
This has been broken for a while, I'd like to contribute a test preventing this code to break again in the future but I need some hints!

src/twisted/web/proxy.py
```diff
@@ -288,7 +288,7 @@ def render(self, request):
         if self.port == 80:
             host = self.host
         else:
-            host = self.host + u":" + str(self.port)
+            host = u"%s:%d" % (self.host, self.port)
```
wiml, line 291: If a hostname is given to us as a `str`, it should probably be encoded using `idna` rather than `ascii`. (Unless the HTTP RFC has some other requirement).

Maybe something along the lines of 

```
host = self.host.encode('idna')
if self.port != 80:
    host += b':' + str(self.port).encode('ascii')
```

or `host += b':%d' % (self.port,)` if you prefer format strings.

src/twisted/web/test/test_proxy.py
```diff
@@ -38,17 +38,21 @@ def _testRender(self, uri, expectedURI):
         # Clear the timeout if the tests failed
         self.addCleanup(channel.connectionLost, None)
 
-        channel.dataReceived(b"GET " +
-                             uri +
-                             b" HTTP/1.1\r\nAccept: text/html\r\n\r\n")
-
+        channel.dataReceived(
+            b'\r\n'.join([
+                b'GET %s HTTP/1.1' % uri,
```
wiml: If we don't want to use `+` here (do we need the implicit `str()` call?) then we should probably change the RHS of the `%` operator to an explicit tuple

docs/web/examples/reverse-proxy.py
```diff
@@ -13,6 +13,6 @@
 from twisted.internet import reactor
 from twisted.web import proxy, server
 
-site = server.Site(proxy.ReverseProxyResource('www.yahoo.com', 80, ''))
+site = server.Site(proxy.ReverseProxyResource('www.asdf.com', 80, b''))
```
wiml: Why the change to the hostname? Do we explicitly want `www.asdf.com` or should we change it to a [standard](https://tools.ietf.org/html/rfc2606) example.com hostname?

wiml (changes requested):
This looks like a good fix to me, just a few code review nits.

src/twisted/web/test/test_proxy.py
dmollerm: I don't get you. You mean to change this line to `b'GET %s HTTP/1.1' % (uri,),` ?

src/twisted/web/proxy.py
dmollerm: > If a hostname is given to us as a `str`, it should probably be encoded using `idna` rather than `ascii`. (Unless the HTTP RFC has some other requirement).

I wouldn't dare to mess with encodings. And AFAIU hostname is required to be unicode and the whole code would crash elsewhere otherwise. In any case I belive this should be handled at the webserver constructor or alike, not in the resource `render` method.

> or `host += b':%d' % (self.port,)` if you prefer format strings.

That would be elegant but beware that `host != self.host`

I don't really have a strong interest in changing these lines. I just felt the `+ u":" +` concatenation was somewhat creepy and a string format would look prettier. But this is a completely optional and opinionated change!

docs/web/examples/reverse-proxy.py
dmollerm: The point is that most popular websites simply redirect any request to their HTTPS service. asdf.com was the first non-local non-specific non-HTTPS dummy website I could come up with. But indeed example.com is a much better option, thanks!

src/twisted/web/proxy.py
dmollerm: > And AFAIU hostname is required to be unicode and the whole code would crash elsewhere otherwise.

My apologies, for it's been a while since I wrote this PR. Indeed hostname is not required to be unicode so we should mess with encodings at some point. Yet I think this should be done elsewhere and is beyond the purpose of this PR. Again, the change of this line is everything but important.

I'll check the bug tracker, may be I can find an issue that relates to this. Can you point me to any non-ascii standard website to test it?

wiml: > Yet I think this should be done elsewhere and is beyond the purpose of this PR

Looking at this further, the modern IDNA encoder is only a soft-depndency of Twisted, and this isn't the only place in this module where unicode hostnames are converted to a byte representation, so I agree this is a complicated enough change that it should be in its own PR. Sorry for the distraction!

> Can you point me to any non-ascii standard website to test it?

For curiosity's sake, here are a few: `http://santé.gouv.qc.ca/` (encodes as `xn--sant-epa.gouv.qc.ca`), `www.köln.de` (`www.xn--kln-sna.de`, but the site redirects to `www.koeln.de`).

src/twisted/web/test/test_proxy.py
wiml: That is what I was suggesting, but I noticed a deeper problem: Python 3.4 doesn't support `%`-interpolation for byte strings (that feature was [added in 3.5](https://docs.python.org/3/whatsnew/3.5.html)), and Twisted still supports Python 3.4. (See [this CI failure](https://travis-ci.org/twisted/twisted/jobs/453364412).)

dmollerm: Oh. Then I'll leave this code unmodified

dmollerm:
Should I squash the branch? 5 out of 9 commits are partial reverts and trunk merges. Which is twisted's policy?

wiml:
If you like. Twisted's core committers generally don't bother cleaning up the history of merges. I'll merge this PR either way.

(I personally like to make the commit history kind of tidy when making a PR but that's just a mild personal preference.)

If you do squash this PR I ask that you keep the commits with actual changes distinct from any merge-from-trunk commits, though, just to make it clear what's being done.

wiml (approved):

PR #1038, opened 2018-07-10
[9485] Skip CModuleSendmsgTests.test_shortsend

twm:
As discussed on the mailing list. I opted for a .misc newsfile as I don't think this is likely to interest anyone reading the changelog.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9485#comment:1

### DIFF
--- a/src/twisted/python/test/test_sendmsg.py
+++ b/src/twisted/python/test/test_sendmsg.py
@@ -408,6 +408,11 @@ class CModuleSendmsgTests(TestCase):
         received = recv1msg(self.output.fileno(), 0, len(message))
         self.assertEqual(len(received[0]), sent)
 
+    test_shortsend.skip = (
+        'Intermittently fails (due to kernel buffer sizing?).'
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Thanks for the cleanup.

.misc fragment is perfect :)

Rather than keeping code which is never run ... maybe we should just delete it.

We can copy the test in the ticket details for 9452 to make it easier to find it.

I don't see the point of keeping a test which is always skipped.
I feel that we are abusing the test skip functionality... and we turn it into a `.todo` thing.
Is ok to skip a test on Windows, as long as it runs on Linux.

At some point, I have cleaned/removed the `.todo` tests.
The .todo tests were code with time became code which is not maintained.

--------

We can merge it as is, but I prefer to just remove the test and have a clean code.

Thanks!

glyph:
Thank you @twm, thank you @adiroiban !

PR #1039, opened 2018-07-12
7927 unhandled error tb 3

jlitzingerdev:
This introduces a new API to address issues where tracebacks for unhandled errors are not printed.  There has been significant discussion in both the ticket and PRs, so for brevity I'm including links below.

The relevant tickets are:
* https://twistedmatrix.com/trac/ticket/7927
* https://twistedmatrix.com/trac/ticket/9456

This is the result of discussion on a review of an earlier candidate for 7927, located here: https://github.com/twisted/twisted/pull/1022#discussion_r200186589


jlitzingerdev:
I'm almost certain the flagged txchecker issue:
`329: W9401: (Used for checking comment format issues.), : Comments should begin with one whitespace` is a bug in twistedchecker.  It is interpreting the `#` in the system tag as a comment.

jlitzingerdev:
@glyph FYI, Initial PR for `eventAsText`.

twm:
It's definitely a txchecker bug: the same bit me on #1038. https://github.com/twisted/twistedchecker/issues/110

src/twisted/logger/_format.py
```diff
@@ -287,3 +238,179 @@ def formatWithCall(formatString, mapping):
...
+    @type includeSystem: L{bool}
+
+    @param formatTime: A time formatter
+    @type formatTime: L{callable} that takes an C{event} argument and returns
+        a L{unicode}
+
+    @return: A formatted string with specified options.
+    @rtype: L{unicode}
+
+    @since: Twisted NEXT
```
glyph, line 393: 🎉

src/twisted/logger/_format.py
```diff
@@ -287,3 +238,179 @@ def formatWithCall(formatString, mapping):
...
+            timeStamp=formatTime(event.get("log_time", None))
+        )
+
+    if includeSystem:
+        system = _formatSystem(event)
+        logText += u"[{system}] ".format(system=system)
+
+    logText += u"{eventText}".format(eventText=eventText)
+
+    return logText + u"\n"
```
glyph: The tacking on of the newline strikes me as a bit odd.  If one is using `eventAsText` directly, one might wish to separate log messages by something other than newlines.  (For example, they might be displayed in some kind of UI widget where the additional newline would space them out too much.)  Why include it?

src/twisted/logger/_format.py
```diff
@@ -287,3 +238,179 @@ def formatWithCall(formatString, mapping):
...
+    eventText = _formatEvent(event)
+    if includeTraceback and 'log_failure' in event:
+        f = event['log_failure']
+        traceback = _formatTraceback(f)
+        eventText = u"\n".join((eventText, traceback))
+
+    if not eventText:
+        return eventText
+
+    logText = u""
```
glyph: Normally I wouldn't be so concerned with a tiny micro-optimization, but given that this has historically been a *very* hot code path in Twisted, I think it might be worthwhile to take some trouble to avoid having to rebuild the format string in its entirety on every single call.

(Let's not block on this; it can easily be done in a follow-up PR.)

docs/core/howto/logger.rst
```diff
@@ -321,13 +321,13 @@ That interface simply describes a 1-argument callable that takes a ``dict`` , so
 .. code-block:: python
 
     from zope.interface import provider
-    from twisted.logger import ILogObserver, formatEvent
+    from twisted.logger import ILogObserver, eventAsText
 
     @provider(ILogObserver)
     def simpleObserver(event):
-        print(formatEvent(event))
+        print(eventAsText(event))
```
glyph: This example now produces double the newlines.

src/twisted/logger/_format.py
```diff
@@ -34,27 +34,12 @@ def formatEvent(event):
...
-        return formatWithCall(format, event)
-
-    except Exception as e:
-        return formatUnformattableEvent(event, e)
+    return eventAsText(
+        event,
+        includeTraceback=False,
+        includeTimestamp=False,
+        includeSystem=False,
+    )[:-1]
```
glyph: This `:-1` seems like a rather unfortunate consequence of the newline being included.  I realize that this is a "legacy" API now, ideally destined to be deprecated... but still, giving it intimate knowledge of the internals of another function seems weird.  (Why `:-1`?  What about `\r\n` newlines?  What about record separators? etc.)

glyph (commented):
I like pretty much everything about this except the dang newline.  It just sticks in my craw :).  Hopefully my explanation for why makes sense.  If you feel strongly otherwise please feel free to respond with your reasoning about why you think it should be included (but you probably also want to fix the docs and stuff in that case to not double up on newlines).

Regarding the txchecker failures:

- there are a number of actual lint errors, if you could fix those it would be great
- you can work around the bug by spelling it '\x23' in tests.  I realize this is a bit gross, but the standard thing to do is to work around issues like this and include a link to the txchecker bug so we can clean them up later, rather than littering the codebase with even more warnings that need to be ignored.

I think if you can address this stuff it's basically ready to land.  When you submit for re-review please feel free to be annoying on the mailing list and such so we can do so quickly :).

src/twisted/logger/_format.py
jlitzingerdev: I think this is actually more an evolutionary artifact of how I developed it.  I'm trying, but can't recall the rationale that led me to leave it in now.  I'll take it out and fix the appropriate places.

src/twisted/logger/_format.py
jlitzingerdev: I'll spend a bit on it, but I won't let it add much delay.

jlitzingerdev:
I'll fix up the style (sorry, that's an annoying miss on my part) and newline issues and resubmit, shouldn't be longer than the upcoming weekend.  I'll also try to get the optimization done.

src/twisted/logger/_format.py
jlitzingerdev: @glyph Since you mentioned this is a hot path, it leads me to another question.  There are helper functions `_formatEvent`, `_formatTraceback` used because (to me) they kept the code more readable.  However, the function call has overhead.  Should those be removed in favor of performance?

hawkowl:
Merged! Thank you.

PR #1043, opened 2018-07-24
[#9499] Better default tox config

hawkowl:
https://twistedmatrix.com/trac/ticket/9499

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,11 +44,6 @@ jobs:
           when: always
           command: |
             tox -r -e manifest-checker
-      - run:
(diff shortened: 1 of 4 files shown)


wsanchez:
This looks good to me.

There used to be two twisted checker runs.  The required one filters some things that are too buggy to fail the build and the other one is allow to fail but doesn't filter, and the latter is now gone.  I would use that in reviews, since not all of the issues it reports are false positives.  Is that removal intentional?

rodrigc (approved):
This is a good consolidation of the tox linting targets into one.

PR #1044, opened 2018-08-01
Default reactor to None, which makes class Server compatibile with old…

barry-scott-forcepoint:
…er version of twisted,

like 2.0.

https://twistedmatrix.com/trac/ticket/9503#ticket

### DIFF
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -801,9 +801,9 @@ class Server(_TLSServerMixin, Connection):
 
     _addressType = address.IPv4Address
 
-    def __init__(self, sock, protocol, client, server, sessionno, reactor):
+    def __init__(self, sock, protocol, client, server, sessionno, reactor=None):
         """
-        Server(sock, protocol, client, server, sessionno)
+        Server(sock, protocol, client, server, sessionno, reactor=None)
 
         Initialize it with a socket, a protocol, a descriptor for my peer (a
         tuple of host, port describing the other end of the connection), an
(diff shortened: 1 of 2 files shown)


wiml (commented):
This seems reasonable to me as far as it goes (though the same change should probably be made to `twisted.internet.unix.Server` as well; they both gained the extra argument in the same commit).

However, the comment at the top of `tcp.py` suggests that these interfaces should be thought of as kind of internal (which is presumably why an incompatible change was made). Why do you need to invoke `Server(...)` directly? Is there an enhancement that could be made to a higher level, more-public API instead?

barry-scott-forcepoint:
We are a MITM engine. The high level APIs do not allow us to get at the details of the HTTP requests that is necessary to build our application.

glyph:
> We are a MITM engine. The high level APIs do not allow us to get at the details of the HTTP requests that is necessary to build our application.

Hi @barry-scott-forcepoint,

In a general sense, I understand that you might need some deeper integration than might be available via the public API.  But, more specifically to `tcp.py`, what parameters do you need that aren't exposed publicly?

glyph:
(My personal feeling on this is that compatibility with Twisted 2.0, in APIs that specifically describe themselves as "internal", is an explicit non-goal.  We will fix stuff like this pre-release, since technically `tcp.Server` is in a "public" namespace with respect to our compatibility policy, but this change has been out there way too long to revert on its own.  I really hope that we have the bandwidth sometime soon to simply underscore-prefix `tcp.py`, `udp.py`, and friends.)

barry-scott-forcepoint:
I no longer need this patch merged. We refactored the code that needed this.

glyph:
> I no longer need this patch merged. We refactored the code that needed this.

Thanks for letting us know.  Hopefully using new public interfaces!  I am sorry that Twisted didn't provide a great experience in this regard, but thanks for your contribution nonetheless.

PR #1045, opened 2018-08-03
9504 barry scott forcepoint uds write failure hidden

barry-scott-forcepoint:
https://twistedmatrix.com/trac/ticket/9504#ticket

For class DatagramPort's write method a failure to write is suppressed.

Unix-domain-sockets are expected to reliable.

This means that it is not possible to arrange to retry a failed write and the datagram is lost silently.

### DIFF
--- a/src/twisted/internet/tcp.py
+++ b/src/twisted/internet/tcp.py
@@ -801,9 +801,9 @@ class Server(_TLSServerMixin, Connection):
 
     _addressType = address.IPv4Address
 
-    def __init__(self, sock, protocol, client, server, sessionno, reactor):
+    def __init__(self, sock, protocol, client, server, sessionno, reactor=None):
         """
-        Server(sock, protocol, client, server, sessionno)
+        Server(sock, protocol, client, server, sessionno, reactor=None)
 
         Initialize it with a socket, a protocol, a descriptor for my peer (a
         tuple of host, port describing the other end of the connection), an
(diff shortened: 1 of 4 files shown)


rodrigc:
It looks like in Twisted 8.1.0, `Server.__init__` did not have a reactor argument:

https://github.com/twisted/twisted/blob/twisted-8.1.0/twisted/internet/tcp.py#L632

but in 8.2.0, the reactor argument was introduced:

https://github.com/twisted/twisted/blob/twisted-8.2.0/twisted/internet/tcp.py#L736

It was introduced by @exarkun in https://github.com/twisted/twisted/commit/cb73a7fa2afdc4db99e55736e101cd258f31f635

rodrigc:
The `EAGAIN` stuff was introduced in https://github.com/twisted/twisted/commit/c754e6905b86e65740ded34bd459cca2de9aeec0

I understand the motivation for your change, since it looks like the intended buffering was not implemented.  Why did you only remove one of the cases of EAGAIN in this file?

rodrigc:
Both changes you made were relatively small, but I would have recommended that you separate them into two separate pull requests with two separate Trac tickets, so they could be reviewed separately.

barry-scott-forcepoint:
I see what I did wrong. I did git checkout -b when the branch was not trunk.
I'm going to close this PR and raise a new one for just the one fix.

barry-scott-forcepoint:
On Friday, 3 August 2018 11:12:06 BST Craig Rodrigues wrote:
> Both changes you made were relatively small, but I would have recommended
> that you separate them into two separate pull requests with two separate
> Trac tickets, so they could be reviewed separately.

I've been following the documented flow and hoped that I was indeed raising 
two indpendent PRs.

I think my error was to branch from the first branch and not from trunk...

I can redo the work and make a clean PR.

Barry

PR #1046, opened 2018-08-03
For class DatagramPort's write method a failure to write is suppressed.

### LINKED ISSUE #9504: internet/unix.py: failure to write UDS datagram is hidden
|[barry-scott-forcepoint's avatar](https://github.com/barry-scott-forcepoint)| @barry-scott-forcepoint reported|
|-|-|
|Trac ID|trac#9504|
|Type|defect|
|Created|2018-08-01 11:15:00Z|

For class DatagramPort's write method a failure to write
is suppressed.

Unix-domain-sockets are expected to reliable.

This means that it is not possible to arrange to retry a
failed write and the datagram is lost silently.

```
trac-id__9504 9504
type__defect defect
reporter__barry_scott_forcepoint barry-scott-forcepoint
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1533122100686444 1533122100686444
changetime__1548137350979202 1548137350979202
version__None None
owner__None None
cc__wiml
```
</details>

barry-scott-forcepoint:
Unix-domain-sockets are expected to be reliable.
    
This means that it is not possible to arrange to retry a failed write and
the datagram is lost silently.
    
Allow the EAGAIN to be handled by the caller to write().

Fixes #9504

### DIFF
--- a/src/twisted/internet/unix.py
+++ b/src/twisted/internet/unix.py
@@ -502,8 +502,13 @@ class DatagramPort(_UNIXPort, udp.Port):
 
     def write(self, datagram, address):
         """Write a datagram.
+
            @param datagram: the data to be sent
+           @type datagram: L{bytes}
+
            @param address: the address to send the datagram to
+           @type address: L{bytes}
+
            @raise socket.error: when socket is full with arg[0] == EAGAIN
         """
         try:
--- a/src/twisted/internet/unix.py
+++ b/src/twisted/internet/unix.py
@@ -502,7 +502,7 @@ class DatagramPort(_UNIXPort, udp.Port):
 
     def write(self, datagram, address):
         """Write a datagram.
-           raises socket.error EAGAIN
+           @raise socket.error: when socket is full with arg[0] == EAGAIN
         """
         try:
             return self.socket.sendto(datagram, address)
--- a/src/twisted/internet/unix.py
+++ b/src/twisted/internet/unix.py
@@ -501,7 +501,9 @@ class DatagramPort(_UNIXPort, udp.Port):
         self.fileno = self.socket.fileno
 
     def write(self, datagram, address):
-        """Write a datagram."""
+        """Write a datagram.
+           raises socket.error EAGAIN
+        """
         try:
             return self.socket.sendto(datagram, address)
         except socket.error as se:
@@ -510,11 +512,6 @@ class DatagramPort(_UNIXPort, udp.Port):
                 return self.write(datagram, address)
             elif no == EMSGSIZE:
                 raise error.MessageLengthError("message too long")
-            elif no == EAGAIN:
-                # oh, well, drop the data. The only difference from UDP
-                # is that UDP won't ever notice.
-                # TODO: add TCP-like buffering
-                pass
             else:
                 raise
 
--- a/src/twisted/internet/unix.py
+++ b/src/twisted/internet/unix.py
@@ -501,15 +501,16 @@ class DatagramPort(_UNIXPort, udp.Port):
         self.fileno = self.socket.fileno
 
     def write(self, datagram, address):
-        """Write a datagram.
+        """
+        Write a datagram.
 
-           @param datagram: the data to be sent
-           @type datagram: L{bytes}
+        @param datagram: the data to be sent
+        @type datagram: L{bytes}
 
-           @param address: the address to send the datagram to
-           @type address: L{bytes}
+        @param address: the address to send the datagram to
+        @type address: L{bytes}
 
-           @raise socket.error: when socket is full with arg[0] == EAGAIN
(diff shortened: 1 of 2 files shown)


barry-scott-forcepoint:
A comment on the V1 patch was why only remove the one EAGAIN?

I think there is only one place in the code that does a UDS sendto.

src/twisted/internet/unix.py
```diff
@@ -501,7 +501,9 @@ def _bindSocket(self):
         self.fileno = self.socket.fileno
 
     def write(self, datagram, address):
-        """Write a datagram."""
+        """Write a datagram.
+           raises socket.error EAGAIN
```
rodrigc: Twisted uses epydoc syntax for comments, so this should use `@raise` or `@raises`.  See: http://epydoc.sourceforge.net/manual-fields.html or other places in the Twisted code.

barry-scott-forcepoint: Done.

src/twisted/internet/unix.py
```diff
@@ -501,7 +501,17 @@ def _bindSocket(self):
...
+        """
+        Write a datagram.
+
+        @param datagram: the data to be sent
+        @type datagram: L{bytes}
+
+        @param address: the address to send the datagram to
+        @type address: L{bytes}
+
+        @raise socket.error: when socket is full with arg[0] == EAGAIN
```
glyph, line 513: FWIW: this is *definitely* the wrong fix.  `write` should never, ever raise EAGAIN.

Perhaps this could instead buffer and retry, even in the UDP case?

barry-scott-forcepoint: At the moment its wrong in that the messages are lost.

We put the retry logic into our code and handle retry and detecting that the received is broken based on knowledge of your application requirements.

Are you saying that what is needed here is the whole producer/comsumer interface?

glyph: Yes, that's the way Twisted deals with backpressure.

wiml: I've added a comment to [ticket 9504](https://twistedmatrix.com/trac/ticket/9504) with some thoughts.

@barry-scott-forcepoint  I wonder if you could comment there on what your application's needs are and whether they would be well-served by a tcp-like buffer and `pauseProducing` implementation, or whether you need finer control over what to do when a write returns `EAGAIN` / `ENOBUFS` / etc.?

barry-scott-forcepoint:
We use UDS (not UDP) because its reliable and message based, not steam like.

Our app is set of cooperating process that exchange data via UDS/

Until we found this bug we had hard to debug failures of our app.
Since implementing this fix and the retry logic we have not had any failures from our use of UDS.

The message boundaries are critical to our use of UDS.

An API that allow us to queue messages for transmission over UDS and handled the retry mechanism would work for us. We retry the write on a timer as being writeable is not the same as being able to accept the message. The socket may have space for 100 bytes and a 200 byte message will get EGAIN.

I'm not sure that back-pressure is the right solution here.

glyph:
> We use UDS (not UDP) because its reliable and message based, not steam like.

This is my mistake.  I just saw "datagram" and my eyes sort of glazed over.  You're absolutely correct that UNIX datagram sockets and UDP are completely different cases, and I'm sorry I wasn't paying close enough attention.

> Our app is set of cooperating process that exchange data via UDS/

(You do just mean "UNIX Datagram Sockets" here, right?)

> Until we found this bug we had hard to debug failures of our app.
> Since implementing this fix and the retry logic we have not had any failures from our use of UDS.

Good to know.

> The message boundaries are critical to our use of UDS.

Yes, dropping framing would be extremely bad, understandable :-).

> An API that allow us to queue messages for transmission over UDS and handled the retry mechanism would work for us. We retry the write on a timer as being writeable is not the same as being able to accept the message. The socket may have space for 100 bytes and a 200 byte message will get EGAIN.
> 
> I'm not sure that back-pressure is the right solution here.

Luckily I think we can both have what we want.

1. Backpressure is, unambiguously, the *correct* way to implement this.  To put it a different way: EAGAIN already *is* backpressure.  It's the operating system telling you that backpressure is being applied, that you're out of buffer space.  So we should propagate this in the same way that Twisted propagates it for other transports, which is to use the producer/consumer protocol.
2. Luckily, Twisted's implementation of backpressure is … incomplete, and happily lets you completely ignore it at the application level if you want.  In terms of system design this is highly unfortunate and is an invitation to write programs that just run out of memory and crash when they run on connections with asymmetric performance characteristics, but: (a) this turns out to be a practical problem waaaaay less often than one would guess, and (b) it seems to be what you want, so, congratulations, you can have it :-).

glyph:
I am leaving the ticket open, but given that this implementation strategy is a dead end for compatibility reasons, I'm closing this PR.

PR #1049, opened 2018-08-25
[#9512] Run Black

hawkowl:
https://twistedmatrix.com/trac/ticket/9512#ticket


hawkowl:
Will replace with a new one.

PR #1050, opened 2018-08-27
[#9501] Regenerate our Cython

hawkowl:
https://twistedmatrix.com/trac/ticket/9501


glyph (approved):
See trac.

src/twisted/test/test_shortcut.py
```diff
@@ -15,6 +15,9 @@
 except ImportError:
     skipReason = "Only runs on Windows with win32com"
 
+if sys.version_info[0:2] >= (3, 7):
```
altendky: Doesn't the full comparison (without `[0:2]`) work fine?  Also, I would have expected we would want to fix the test, not skip it?  Or is there some reason this shouldn't be supported.

altendky, line 18: Doesn't the full comparison (without `[0:2]`) work fine?  Also, I would have expected we would want to fix the test, not skip it?  Or is there some reason this shouldn't be supported.

hawkowl: shortcut was only relevant 10 years ago, so it's something we'll get rid of eventually.

PR #1055, opened 2018-08-30
Don't corrupt RSA private keys during loading.

agl:
Fifteen years ago, Twisted landed [a change](https://github.com/twisted/twisted/commit/3cf1ed6058067c1597ae54cb3a1cd95725a1e8d1) which enforced that the RSA primes were ordered p &les; q, but then serialised them backwards, so that actually p &ges; q, and it always recreated the CRT values. This was because of [this bug](https://twistedmatrix.com/trac/ticket/259), which is something to do with testing key serialisation round-tripping. 2003-era PyCrypto [did something](https://github.com/dlitz/pycrypto/blob/bb6a94896433587a7a819efde503337e4b9ab3a5/PublicKey/RSA.py#L74) very similar, so they were probably copying that in order to avoid PyCrypto changing their keys and triggering round-trip failures.

As time has gone by, much of the Twisted code has changed, but the initial condition to swap p & q if p > q has remained. However, the CRT values are no longer recomputed, so this just corrupts the key. (For example, `dmp1` stands for &ldquo;d mod p-1&rdquo; so clearly if p & q are swapped, that value has to be updated in order to be correct. The values `dmq1` and `iqmp` (inverse of q mod p) are likewise.)

OpenSSL has anti-glitch checking in its RSA implementation that is designed to stop power-glitching attacks. (An arithmetic error during CRT computation has a high chance of leaking the private key.) That anti-glitch code detects the error caused by swapping p and q and triggers a slow-path re-computation of signatures without the CRT. Thus OpenSSL is silently correcting this, but spending much more CPU to do so.

https://twistedmatrix.com/trac/ticket/9518

## Contributor Checklist:


alex:
There are failing tests that look relevant; I suspect they need to be un-swapped

glyph (dismissed):
See comment on Trac: https://twistedmatrix.com/trac/ticket/9518#comment:3

glyph:
(Also it looks like you're using an email address that isn't associated with your github account here - is that intentional?  If not maybe rebase...)

alex (approved):

PR #1056, opened 2018-09-05
[9445] Re-enable macOS on circleCI.

altendky:
Scope
=====
See https://twistedmatrix.com/trac/ticket/9445

This re-enabled macOS builds on circleCI after circelCI donated more build times for Twisted.

With macOS on CircleCI we can run macOS tests from branches which don't belong to Twisted repo and which are not run on Buildbot.

Changes
=======
I have now enabled the macOS builds only after static checkers and documentation is green.
I hope that this can reduce the time we use macOS on circleCI

How to test
=========
Check that this branch has macOS build https://circleci.com/gh/twisted/workflows/twisted

One build for this PR https://circleci.com/workflow-run/385ec102-6182-407f-9daa-ba3b4406377c


altendky:
I suppose we should make the py27 job `required`?  Also the `static_checkers` and `documentation`?  The Travis and AppVeyor py37 builds are marked as allowed failures for now (until they pass I figure, and we claim 3.7 support) so we could leave the Circle py37 as not required.

altendky:
Trunk is broken.  I guess I'll update this branch when trunk is ok again.

glyph:
@altendky trunk should be OK again now.  Travis should be good and busy for a while with all the PR rebuilds, though :)

glyph (approved):
OK. I think I understand what's going on here. Let's do it.

altendky:
Good timing. Just this week I got a project setup on Azure Pipelines...  All OSes, all Python versions, and PyPy.  Even provides artifact storage including between jobs (build a wheel first, use it in separate test jobs).  But sure, at this point that should probably be a later step even if you're interested.

https://dev.azure.com/altendky/exttr/_build/results?buildId=120

PR #1058, opened 2018-09-06
[9524] exc info holding error holder tests repr

altendky:
https://twistedmatrix.com/trac/ticket/9524

### DIFF
--- a/src/twisted/trial/test/test_runner.py
+++ b/src/twisted/trial/test/test_runner.py
@@ -844,16 +844,10 @@ class ErrorHolderTestsMixin(object):
         L{runner.ErrorHolder.__repr__} returns a string describing the error it
         holds.
         """
-        if sys.version_info < (3, 7):
-            expected = (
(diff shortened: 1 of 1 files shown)


rodrigc (changes requested):
Thanks for working on this.

Instead of checking for the Python version,
can you use [self.assertIn](https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug) instead?

`self.assertIn(.....,
                        repr(self.holder))
`

altendky:
Aren't we primarily wanting to assert the outer form?  I could alternatively do something like:

```python
try:
    1//0
except ZeroDivisionError as e:
    r = repr(e)
expected = (
    "<ErrorHolder description='description' error={}>".format(r)
)
```

Or get the actual error object directly and use its representation.  I just thought we'd like a pure constant string (well, two...).

rodrigc (approved):
Latest iteration looks fine.
Thanks!

PR #1066, opened 2018-09-09
[#9527] Deprecate functionality broken by PEP-479

hawkowl:
https://twistedmatrix.com/trac/ticket/9527

### DIFF
--- a/src/twisted/conch/newsfragments/9527.removal
+++ b/src/twisted/conch/newsfragments/9527.removal
@@ -1,1 +1,1 @@
-twisted.conch.ssh.filetransfer.ClientDirectory's use as a iterator has been deprecated.
+twisted.conch.ssh.filetransfer.ClientDirectory's use as an iterator has been deprecated.
--- a/src/twisted/conch/newsfragments/9527.removal
+++ b/src/twisted/conch/newsfragments/9527.removal
@@ -1,0 +1,1 @@
+twisted.conch.ssh.filetransfer.ClientDirectory's use as a iterator has been deprecated.
--- a/src/twisted/conch/ssh/filetransfer.py
+++ b/src/twisted/conch/ssh/filetransfer.py
@@ -7,6 +7,7 @@ from __future__ import division, absolute_import
 
 import errno
 import struct
+import warnings
 
 from zope.interface import implementer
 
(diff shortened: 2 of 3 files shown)


src/twisted/conch/newsfragments/9527.removal
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh.filetransfer.ClientDirectory's use as a iterator has been deprecated.
```
altendky: `a iterator` -> `an iterator`

altendky (changes requested):
Sorry to be pedantic but I figure I ought to be on my first review.

---

Spurious build errors?

https://twistedmatrix.com/trac/wiki/ReviewProcess#Reviewers:Howtoreviewachange
> Don't block a branch on a tool bug, but also, don't let any spurious failures go without filing an appropriate bug on the relevant tool first.

Are there already reports for this?

---

https://github.com/twisted/twisted/blob/a57877c133f18f590429a98c7f92ea63f9f30e0c/src/twisted/conch/test/test_filetransfer.py#L421

Locally this triggers the warning code to run but it isn't reported by trial.  Is this silence desired?  Is this use also intended to be deprecated or is this change effectively saying that you can no longer use `ClientDirectory` 'normally' in `@inlineCallbacks`?  I guess that would be a bit troublesome to me if we are just deprecating away usability with `@inlineCallbacks` (and async/await?).

---

https://github.com/twisted/twisted/blob/a57877c133f18f590429a98c7f92ea63f9f30e0c/src/twisted/conch/ssh/filetransfer.py#L935

This line seems to have no coverage on 2.7, 3.6, or 3.7.  Here's a test that seems to cover it, maybe even sensibly.

<details> <summary>test_openDirectoryAgain()</summary>

```python
def test_openDirectoryAgain(self):
    self._emptyBuffers()
    files = []
    second_pass = []

    def _getFiles(openDir):
        def append(f):
            if second_pass:
                files.append(f)
            return openDir
        d = defer.maybeDeferred(openDir.next)
        self._emptyBuffers()
        d.addCallback(append)
        d.addCallback(_getFiles)
        d.addErrback(_close, openDir)
        return d

    def _checkFiles(ignored):
        self.assertEqual(files, [])

    def _close(_, openDir):
        if not second_pass:
            d = openDir.close()

            d.addCallback(lambda _: _getFiles(openDir))
            second_pass.append(None)

            self._emptyBuffers()
        else:
            return _checkFiles(None)

        return d

    d.addCallback(_getFiles)
    return d
```
</details>

hawkowl:
@altendky I wouldn't mind that testcase on a separate PR, if you have time :) This is just a deprecation, so, we shouldn't expand the scope.

The deprecation of the interface I think is fine, since it's a) rarely used and b) abusing iterators.

WRT the linting errors, yeah, they're known and have bugs.

PR #1051, opened 2018-08-28
[9217] Build lots of wheels

altendky:
https://twistedmatrix.com/trac/ticket/9217

https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

#### Work to do:
- [ ] ~~Storage for Travis artifacts~~
  - Using Circle for now
- [x] Get `cibuildwheel` working on CircleCI for OSX wheels
- [ ] ~~Learn about how Twisted likes to deploy wheels~~
  - ~~Maybe automate deployment if interested)~~
    - I did do PyPi deployment from Travis for a project
    - [Glyph is interested](https://twistedmatrix.com/pipermail/twisted-python/2018-September/032050.html) but I'd suggest a separate PR
  - Automatic deployment will be addressed separately
- [x] Resolve test failure on `twisted.cred.test.test_strcred.SSHCheckerTests.test_isChecker` (https://github.com/twisted/twisted/pull/1051#issuecomment-416977723)
  - Passing now
- [x] Use a released version of `cibuildwheel` (probably within the week)
- [x] Build 3.7 wheels
  - Some way to not fail the whole build on their failure right now?

PR #1067, opened 2018-09-09
[9530] [WIP] test wheels not source

altendky:
Note that this is presently building on #1051 under the assumption that it will get accepted.  Presumably do not merge this first...

https://twistedmatrix.com/trac/ticket/9530

WIP for:
* [ ] #1051 getting accepted
* [ ] Fixup per 2019.01.01 review
* [ ] Hopefully eliminate `admin/x.py`, otherwise rename


altendky:
87 commits...  this is one embarassing mess of a pr

.circleci/prepare.sh
```diff
@@ -0,0 +1,8 @@
+$PYTHON --version
```
glyph: "$PYTHON" really needs to be quoted here.

(Also, Twisted shell-script house style generally involves semicolons at the end of each line, although the real Twisted house style is "don't write a shell script".)

.circleci/prepare.sh
```diff
@@ -0,0 +1,8 @@
...
+$PYTHON -m pip --version
+$PYTHON -m pip install -q --user --ignore-installed --upgrade virtualenv
+$PYTHON -m virtualenv -p $PYTHON venv
+venv/bin/pip install -U pip
```
glyph: Should this be `./venv/bin/pip` ?

.travis.yml
```diff
@@ -23,29 +23,48 @@ env:
 
 matrix:
   include:
+    - sudo: true
```
glyph: `sudo: true` is now redundant, I think? travis has deprecated / removed `sudo: false` builders.

.circleci/config.yml
```diff
@@ -16,34 +16,80 @@
...
+      - image: circleci/python:3.6
+    environment:
+      PYTHON: python3
+    steps:
+      # Get the source.
+      - checkout
+      - setup_remote_docker
+
+      - run:
+          name: Prepare the environment.
```
glyph: what is "the environment" here?

admin/x.py
```diff
@@ -0,0 +1,26 @@
+import argparse
```
glyph, line 1: Can this file have:

- a more expository name?
- a docstring indicating where and how it's used?

tox.ini
```diff
@@ -78,6 +83,14 @@ setenv =
    ; Skip twistedchecker warnings that are not yet required in the required run
    lint: TWISTEDCHECKER_SKIP_WARNINGS = W9208,C0302,C0103,C9302
 
+   CIBW_BUILD=cp27-* cp34-* cp35-* cp36-* cp37-*
```
glyph: this is starting to get pretty obscure / confusing. Can you add some comments as to what CIBW_SKIP and CIBW_BUILD are?  presumably they're something to do with the `cibuildwheel` tool, so just a link to the relevant docs would be enough, don't have to re-explain it all.

tox.ini
```diff
@@ -86,14 +99,31 @@ commands =
     python -c "import sys; print(sys.prefix)"
     python -c "import sys; print(sys.exec_prefix)"
     python -c "import sys; print(sys.executable)"
+    python -c "import sys; print(sys.version)"
```
glyph: This pattern seems a bit silly / slow.  Can we consolidate this all into one big `python -c` invocation?  Or maybe even put it into a python file, and then `python dump_all_version_info.py` or somesuch?

tox.ini
```diff
@@ -86,14 +99,31 @@ commands =
...
+    python -c "import sys; print(sys.platform)"
     python --version
 
+    {withcov,coverage,nocov,nomodules,tests}: cibuildwheel --output-dir {toxinidir}/wheelhouse {toxinidir}
+    {withcov,coverage,nocov,nomodules,tests}: ls -l {toxinidir}/wheelhouse
```
glyph: Is this just for debugging?

tox.ini
```diff
@@ -86,14 +99,31 @@ commands =
...
+    {withcov,coverage,nocov,nomodules,tests}: pip download --verbose --no-deps --no-index --dest {envtmpdir} --find-links {toxinidir}/wheelhouse twisted
+    {withcov,coverage,nocov,nomodules,tests}: ls -l {envtmpdir}
```
glyph: Similarly I'm wondering what the purpose of this `ls` is...

glyph (commented):
Not quite enough for a full code review here, but I figured this was useful enough feedback to pass along.  Can you generally just add more docs so that our ever-more-convoluted CI configuration doesn't become totally impenetrable? :)

.circleci/prepare.sh
altendky: As to the shell script, I've since learned how to reuse stuff both in yaml and specifically within circle config.  I'll plan to do that and drop the `.sh`.

.circleci/prepare.sh
altendky: Why would I `./` when I've already got a path specified?  `./` is just a way to get a path when you don't otherwise need one.  At least as far as I know...

.travis.yml
altendky: Huh, alrighty then.  And I guess Xenial is default?  I'll have to reevaluate my normal incantation for getting all the Python versions covered.  It always included sudo:true/dist:xenial to get 3.7 coverage.  (honestly not sure why i had sudo here, maybe it was required for docker stuff?)

https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments

tox.ini
altendky: I think it may have been but it seems a nice thing to have for the log?

altendky:
@glyph several of your comments were against changes made over in #1051 .  I updated those in https://github.com/twisted/twisted/pull/1051/commits/3b7aff382beca38f7edc37a3d78d6521436e6c2c.  If that builds well then I will update merge to here.  But, probably focus on #1051 first for more reviews, unless you are looking further ahead and want to review the test-wheels-instead switch at the same time.  Point was to make for multiple smaller things to review.

Thanks for the feedback.

admin/x.py
altendky: Added an action in the WIP list.  I'm going to leave it as is for now as it is really a foul thing that shouldn't exist.  Hopefully at some point I eliminate it.

.circleci/prepare.sh
glyph: Looking forward to that :)

.circleci/prepare.sh
glyph: I guess I picked up the habit somewhere and I’m not sure why.  It seems less ambiguous, because while slashes do disable $PATH-searching, it doesn’t *inherently* do that. Maybe it works differently in `zsh`?

tox.ini
glyph: Hmm… point taken, you can see all the platform specifiers in the filenames.  👍

glyph:
@altendky is this still worth doing, given the other changes that have happened?

PR #1068, opened 2018-09-23
[9481] Change DelayedCall repr to include details

twm:
This turns out to be quite simple, as there is already a `__str__` implementation that does the necessary: we can just alias `__repr__` to `__str__`.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. [#9481 t.i.b.DelayedCall repr should include useful information](https://twistedmatrix.com/trac/ticket/9481)

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -47,7 +47,7 @@ class DelayedCall:
     # enable .debug to record creator call stack, and it will be logged if
     # an exception occurs while the function is being run
     debug = False
-    _str = None
+    _repr = None
 
     def __init__(self, time, func, args, kw, cancel, reset,
                  seconds=runtimeSeconds):
@@ -101,7 +101,7 @@ class DelayedCall:
             self.canceller(self)
             self.cancelled = 1
             if self.debug:
-                self._str = str(self)
+                self._repr = repr(self)
             del self.func, self.args, self.kw
 
     def reset(self, secondsFromNow):
@@ -181,15 +181,15 @@ class DelayedCall:
         return self.time < other.time
 
 
-    def __str__(self):
+    def __repr__(self):
         """
-        Implement C{str()} and C{repr()} for L{DelayedCall} instances.
+        Implement C{repr()} for L{DelayedCall} instances.
 
         @rtype: C{str}
         @returns: String containing details of the L{DelayedCall}.
         """
-        if self._str is not None:
-            return self._str
+        if self._repr is not None:
+            return self._repr
         if hasattr(self, 'func'):
             # This code should be replaced by a utility function in reflect;
             # see ticket #6066:
@@ -223,16 +223,6 @@ class DelayedCall:
         L.append('>')
 
         return "".join(L)
-
-
-    def __repr__(self):
-        """
-        Implement C{repr()} in terms of C{str()} for backwards compatibility.
-
-        @rtype: C{str}
-        @returns: String containing details of the L{DelayedCall}.
-        """
-        return self.__str__()
 
 
 
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -225,7 +225,14 @@ class DelayedCall:
         return "".join(L)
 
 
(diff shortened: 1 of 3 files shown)


twm:
I moved the epytext syntax fix to https://github.com/twisted/twisted/pull/1069

altendky (commented):
`object.__str__()` already falls back on `__repr__()` so I would think it would be tidier to rename the method to `__repr__()` than to assign the method to both attribute names?

https://docs.python.org/3/reference/datamodel.html#object.__str__

twm:
@altendky I thought about doing it that way, but then the behavior would be different if `__str__` is overridden in a subclass of `DelayedCall` (which is unfortunately allowed as `DelayedCall` is public). This way isn't much more code but is more compatible. (This is hypothetical though &mdash; the only project I could think of which might need to subclass `DelayedCall` is [txlocal](https://bitbucket.org/hipchat/txlocal/overview) but it doesn't appear to.)

altendky:
@twm, in what scenario are you saying there would be a difference?

https://repl.it/@altendky/twisted-t9481-pr1068-001
<details> <summary>example</summary>

```python3
class ReprEqStr:
    def __str__(self):
        return "base"
    
    __repr__ = __str__

class ReprEqStrChild(ReprEqStr):
    def __str__(self):
        return "inherited"

print(str(ReprEqStr()))
print(repr(ReprEqStr()))
print(str(ReprEqStrChild()))
print(repr(ReprEqStrChild()))

print()

class Repr:
    def __repr__(self):
        return "base"
    
class ReprChild(Repr):
    def __str__(self):
        return "inherited"

print(str(Repr()))
print(repr(Repr()))
print(str(ReprChild()))
print(repr(ReprChild()))
```
```
base
base
inherited
base

base
base
inherited
base
```
</details>

twm:
Good point... I didn't think that through. If it were implemented like this:

```python
class DelayedCall:
    def __str__(self):
        return '...'
    def __repr__(self):
        return self.__str__()
```

Then it would be different. I'm torn... does being this fastidious about subclassability even matter here?

Thank you for your attention to detail!

altendky:
I guess I'm not understanding why we want this functionality (`repr` getting the `str`).  I wouldn't expect overriding `__str__()` in a subclass to change the `__repr__()`.  Given the `<object ...>`-like form It seems more like it was just an initial error that `__str__()` is what was chosen to be defined (or perhaps the semantics of `__str__` and `__repr__` were different in 2002 https://github.com/twisted/twisted/commit/9eaf5b1814af3876fba9d02d0a8942192177073d).  I'm probably missing something but I haven't thought of a (not-contrived) scenario where changing the `def __str__` to `def __repr__` would break existing code.  A grep for `self.__str__` turns up only one other use (below) which seems like it should just be `print(self)` or at least `print(str(self))`), but anyways.  So this doesn't seem to be a normal thing for Twisted to do.

https://github.com/twisted/twisted/blob/5492c16d5cb091efa5630113040e706d8a163fd2/src/twisted/python/usage.py#L198

twm:
The existing `__str__` method generates a `__repr__` style string (angle brackets, fully-qualified class name, etc.) so I am inclined to agree that declaring `__str__` instead of `__repr__` was an error or oversight. The question in my mind is whether the very-strictly-backward-compatible implementation here is worth the :confused: look it will elicit from future readers. Based on this conversation, I think not. If it is, in fact, backwards incompatible in practice we can hope the release process catches it. I'll change the implementation to just define `__repr__`.

altendky (approved):
I do generally look forward to better code but I know practically it's appropriate to maintain backwards compatibility.  I just don't think this extends to implementation details of `_`-starting things.  But, I won't be offended if you want someone with a bit more experience with Twisted policies to weigh in.  :]

Please merge.

altendky:
(https://twistedmatrix.com is down at the moment, appropriate comment in ticket pending availability)

twm:
Thank you for the review @altendky! I will merge when the merge forward goes green.

PR #1073, opened 2018-10-01
[9543] Expose Windows CreateProcess Creation Flags

guysv:
At the moment, every process created with reactor.spawnProcess gets
a hard-coded creation flag.
This commit exposes the creation flags to the user for fine-tunes.
The default creation flag is still the same (CREATE_NO_WINDOW).

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9543

## Regarding The Tests
As it is now, I exposed CreateProcess's creation flags parameter to any valid win32 creation flag. I guess some of them, or combinations of them, will break the code. The question is: should we whitelist-and-test the possible creation flags we want to expose, or expose all of them, and let the user be in charge of reading the docs and understanding how to make it work.


glyph:
Hi @guysv !

Is this intended to be in review?

guysv:
@glyph yes, sorry. I filed the ticket accordingly

glyph:
@guysv Sadly we have some more linter failures for you:

```
./src/twisted/internet/test/test_process.py:847:1: W293 blank line contains whitespace
./src/twisted/internet/test/test_process.py:854:80: E501 line too long (84 > 79 characters)
./src/twisted/internet/test/test_process.py:855:1: W293 blank line contains whitespace
./src/twisted/internet/test/test_process.py:859:1: W293 blank line contains whitespace
./src/twisted/internet/test/test_process.py:860:1: E305 expected 3 blank lines after class or function definition, found 1
```

src/twisted/internet/test/test_process.py
```diff
@@ -844,6 +844,19 @@ def spawnChild():
 
         reactor.callWhenRunning(spawnChild)
         self.runReactor(reactor)
+    
+    def test_creationFlagsThrows(self):
+        """
+        L{twisted.internet.process} cannot accept creation flags.
```
glyph, line 850: I think you mean `spawnProcess`, not `twisted.internet.process`.

glyph (changes requested):
Thanks for contributing this @guysv !

Sorry this took so long to get to.

A few things:

1. this really needs some tests for the *positive* path.  what happens when some flags are passed?
2. there needs to be documentation of some kind for this flag.  You might even want to have a separate interface for this, a subclass of `IReactorProcess`, like `IReactorProcessWindowsSpecific`.
3. the (new?) lint failures should be addressed.
4. nitpick: any reason `creationFlags` doesn't just default to `CREATE_NO_WINDOW`?  Why allow for `None`?

guysv:
Hi @glyph thats for the response!
Unfortunately, I lost my interest merging this PR in it's current state.
A better approach to customize win32 flags would be in my opinion to do it like they did it in [Asyncio](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec): wrap the standard library `subprocess.Popen` to which you can pass whatever flag you want. but that is not a quick fix.
Also, if anyone would ever need such functionality, they could monkey-patch the module to use the wanted Win-API flags.

glyph:
Thanks @guysv ! If you have lost interest in this PR and don't intend to take it to completion I will close it.  Thanks for contributing to Twisted.  Please feel free to open a new PR if you would like to implement a different strategy.  (However, there are a variety of reasons that stdlib `Popen` is maybe not the greatest implementation strategy, so I think that one would be a tough sell.)

PR #1076, opened 2018-10-07
[9522] ConnectionAborted message

twm:
Fix truncation of the ConnectionAborted message.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9522


src/twisted/internet/error.py
```diff
@@ -236,6 +236,17 @@ class ConnectionAborted(ConnectionLost):
...
     """
 
+    def __str__(self):
+        s = [(
+            "Connection was aborted locally using"
+            " ITCPTransport.abortConnection"
+        )]
+        if self.args:
+            s.append(': ')
+            s.append(' '.join(self.args))
```
glyph, line 246: This is potentially an incompatible change; `args` will now raise an exception if you put anything but native `str` into it.

glyph (approved):
If you can address the possibility of applications outside of Twisted instantiating `ConnectionAborted` with things other than `str`s, please go ahead and land.

src/twisted/internet/error.py
twm: I woulsn't have normally written it this way, certainly, but it is not incompatible. The superclass does the same: https://github.com/twisted/twisted/pull/1076/files#diff-0c6894f531d99b70536c09f2e433bd6cR225

PR #1077, opened 2018-10-10
Add __repr__ to BaseConnector

jan-matejka:
## Contributor Checklist:


src/twisted/internet/base.py
```diff
@@ -1168,6 +1168,10 @@ def getDestination(self):
             reflect.qual(self.__class__) + " did not implement "
             "getDestination")
```
twm, line 1170: You will need an additional blank like here.

src/twisted/test/test_internet.py
```diff
@@ -1392,3 +1393,10 @@ def testSSL(self, ssl=ssl):
 
         if _PY3:
             testSSL.skip = ("Re-enable once the Python 3 SSL port is done.")
+
+class ConnectorRepr(unittest.TestCase):
+    def test_tcp_repr(self):
+        c = Connector('localhost', 666, object(), 0, object())
+        expect = "<twisted.internet.tcp.Connector instance at %x %s" % (
+            id(c), str(c.getDestination()))
+        self.assertTrue(repr(c), expect)
```
twm: It looks like you meant `assertEqual` here, not `assertTrue`.

src/twisted/internet/base.py
```diff
@@ -1168,6 +1168,10 @@ def getDestination(self):
...
 
+    def __repr__(self):
+        return "<%s instance at %x %s>" % (
```
twm: I think that the `%x` part should be `0x%x` as object IDs are generally presented with that prefix but `%x` doesn't automatically add it:

```python
>>> object()
<object object at 0x7f5894500080>
>>> '%x' % id(object())
'7f58945000a0'
```

src/twisted/internet/base.py
```diff
@@ -1168,6 +1168,10 @@ def getDestination(self):
...
+            reflect.qual(self.__class__), id(self), str(self.getDestination()))
```
twm: The call to `str()` here is redundant &mdash; `%s` means `str()`.

src/twisted/test/test_internet.py
```diff
@@ -1392,3 +1393,10 @@ def testSSL(self, ssl=ssl):
 
         if _PY3:
             testSSL.skip = ("Re-enable once the Python 3 SSL port is done.")
+
+class ConnectorRepr(unittest.TestCase):
```
twm: Could you rename this `ConnectorReprTests` to match the naming convention we use elsewhere?

You will need two more blank lines above this to get the linting to pass. This is currently causing the Travis build to fail:

```
./src/twisted/test/test_internet.py:1397:1: E302 expected 3 blank lines, found 1
```

twm (changes requested):
Hi @yaccz, thanks for taking the time to submit [#9548](https://twistedmatrix.com/trac/ticket/9548) and this PR. Unfortunately there are a few issues (commented inline) which mean I can't merge just yet. I have one suggestion as well: consider including the `self.state` attribute in the repr as well as it is short and likely useful when debugging.

Once the Travis build is passing I'll take another look here. (Don't worry about the yellow buildbot status check &mdash; it won't go green until I push a branch for this PR, which I will do once Travis is green.)

Thanks!

src/twisted/internet/base.py
jan-matejka: Hi, thanks for the review. I fixed all the issues except this one as I am not clear on where to add the blank line. It currently seems consistent with the rest of the code to me.

jan-matejka:
I don't understand the lint failure `TypeError: modifiedBlankLines() takes 7 positional arguments but 9 were given`. It looks more like an internal error.

src/twisted/internet/base.py
twm: Hmm, this is odd. Normally spacing in Twisted is 3-2-1 (3 blank lines at module level, 2 at class level, 1 within functions and methods). I think that the linter is checking just the diff, so it isn't complaining here. If it doesn't care I don't, particularly in light of #1049. I just need a build which passes in order to merge.

twm (approved):
Yeah, I'm sorry about that confusing traceback. You are correct that it is an internal error. It is appearing on all builds &mdash; nothing to do with your changes. The actual errors are lower down:

```
./src/twisted/internet/base.py:1173:80: E501 line too long (86 > 79 characters)
./src/twisted/test/test_internet.py:1402:80: E501 line too long (88 > 79 characters)
```

So we've moved on to line length issues. If you can fix those up this looks good and I will merge.

Thanks!

src/twisted/internet/base.py
yaccz: So what can I do about it? I don't understand what the issue is from the CI output.

jan-matejka:
I pushed the line length fixes. Some buildbots are still running but travis look green now.

twm:
Thanks @yaccz!

PR #1078, opened 2018-10-18
[9546]  DeferredLock, DeferredSemaphore, etc as async context managers 

wiml:
* [X] There is [an associated ticket in Trac](https://twistedmatrix.com/trac/ticket/9546)

This is a pretty small change, most of the commit is the unit tests.

For the code which isn't parsable on older versions of Python, I copied the pattern used in other "`.3only`" tests.


src/twisted/test/test_defer.py.3only
```diff
@@ -0,0 +1,66 @@
...
+        return defer.ensureDeferred(result)
+    return wrapper
+
+
+
+class DeferredTestsAsync(unittest.TestCase):
+
+
+    @ensuringDeferred
+    async def test_asyncWithLock(self):
```
twm, line 33: There should be a docstring here.

src/twisted/test/test_defer.py.3only
```diff
@@ -0,0 +1,66 @@
...
+            d.addCallback(lambda _: lock.release())
+            self.assertTrue(lock.locked)
+            self.assertFalse(d.called)
+        self.assertTrue(d.called)
+        await d
+        self.assertFalse(lock.locked)
+
+
+    @ensuringDeferred
+    async def test_asyncWithSemaphore(self):
```
twm, line 50: And a docstring here too.

twm (changes requested):
I noted some missing docstrings (style issue), but functionally this looks to be in great shape.

I do think that there should be coverage of the exception case, something like this:

```python
    @ensuringDeferred
    async def test_asyncWithLockException(self):
        """
        C{Lock} does not handle exceptions when used as an async context
        manager.
        """
        lock = defer.DeferredLock()
        with self.assertRaises(IndentationError):
            async with lock:
                self.assertTrue(lock.locked)
                raise IndentationError()
        self.assertFalse(lock.locked)
```

This will be a useful addition to Twisted, so thank you! Since adding tests will involve code, I think there should be a re-review so I am marking this as "changes required". Feel free to ping me for a re-review — I will try to be prompt about it.

wiml:
Sounds good. Commit 55cab3ebb73ef3cff126474b7235f035aee610af should address your feedback.

twm (approved):
Looks great. Thanks @wiml! I'll go ahead and merge.

PR #1087, opened 2018-11-16
9125 mutually exclusive exception text

wiml:
## Contributor Checklist:

* [X] There is [an associated ticket in Trac](https://twistedmatrix.com/trac/ticket/9125)

### DIFF
--- a/src/twisted/newsfragments/9125.misc
+++ b/src/twisted/newsfragments/9125.misc
@@ -1,0 +1,1 @@
+Improved the exception text for mutually exclusive function arguments.
--- a/src/twisted/python/deprecate.py
+++ b/src/twisted/python/deprecate.py
@@ -788,7 +788,10 @@ def _mutuallyExclusiveArguments(argumentPairs):
             arguments = _passed(spec, args, kwargs)
             for this, that in argumentPairs:
                 if this in arguments and that in arguments:
-                    raise TypeError("nope")
+                    raise TypeError(
+                        ("The %r and %r arguments to %s "
+                         "are mutually exclusive.") %
+                        (this, that, _fullyQualifiedName(wrappee)))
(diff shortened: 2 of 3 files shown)


src/twisted/newsfragments/9125.misc
```diff
@@ -0,0 +1 @@
+Improved the exception text for mutually exclusive function arguments.
```
exarkun, line 1: misc fragments should be empty.  their text won't be included in a news file, it'll just be dropped.

if the text were to be kept and included, the fragment should follow the style guidelines from the twisted wiki.  roughly, "twisted.internet.ssl.CertificateOptions now raises an exception with better text when passed mutually exclusive arguments."

exarkun: And in this case, the change is actually an improvement to `twisted.python.deprecate`, so making the fragment text about the API there probably makes sense too.  (CertificateOptions could be given as an example beneficiary.)

exarkun (approved):
Thanks.  The implementation change looks good.  Just one minor comment about the news fragment (inline).

src/twisted/newsfragments/9125.misc
wiml: Ah, I see. Do you think it's worth replacing this newsfragment with an empty one, even though it'll make no difference in the long run (once the newsfragment is deleted after the next release)? Or change it to a non-`.misc` newsfragment? Or should I merge it as-is?

exarkun: I don't think it will hurt to merge it as-is and then you don't have to wait for another round of CI.

wiml: Good point. Merged. :)

PR #1090, opened 2018-12-02
Optimize @defer.inlineCallbacks by up to four times

p12tic:
Hi,

While looking into differences of performance between addCallback and inlineCallback approaches I've noticed several optimization opportunities in inlineCallbacks code path. After some experimentation I was able to achieve up to 4x speedup.

I did the following optimizations:

 - when the wrapped generator yields a `Deferred` to `inlineCallbacks` and we can prove that its result is going for us, we steal it instead of going through the `.addCallback`. `.addCallback` and `.callback` are relatively expensive and result in good performance improvement when they can be avoided.

 - when the wrapped generator returns its final result and we detect that the `Deferred` returned by the current `inlineCallbacks` decorator is immediately yielded to another `inlineCallbacks`, we inject the returned value directly to the parent `inlineCallbacks` loop instead of performing `.callback`. This situation is relatively frequent in code that uses `inlineCallbacks` extensively.

 - an fast-path for `Deferred.callback` method has been created for situation when we can guarantee that the `Deferred` does not have any callbacks yet.

 - we cache whether the result of a `Deferred` is a failure or not. This allows us to avoid a huge number of `isinstance` calls which turned out to have non-insignificant performance impact.

Additional unit tests have been added. Additionally, the unit tests of the buildbot project pass when using the optimized version of twisted. They are also around 0.5% faster.
 
The complete performance results and scripts used to create them are here: https://github.com/p12tic/gist-twisted-perf-tests. In particular, see the comparison on python 2.7 running on Ryzen 2 at 3.00GHz [here](https://github.com/p12tic/gist-twisted-perf-tests/blob/master/results-ryzen2-3.00ghz/comparison_python_27.txt) and the comparison on python 3.6 [here](https://github.com/p12tic/gist-twisted-perf-tests/blob/master/results-ryzen2-3.00ghz/comparison_python_36.txt).

The following performance changes can be observed:

 - `defer.succeed()` became more than 2x faster. See `test_X_Y_deferred` tests.

 - the overhead of `inlineCallbacks` in non-blocking, completely synchronous case has been reduced by up to 4.1 times. Most reasonable use cases show around 2-2.5x improvement. See `test_yield_X_deferred_return` and `test_yield_X_yield_Y_deferred_return` tests.

 - the overhead of `inlineCallbacks` in blocking, completely asynchronous case has been reduced by at 1.2-1.3 times when the generator of `inlineCallbacks`-decorated function is passed to yield in another `inlineCallbacks`-decorated function (See `test_yield_X_yield_Y_deferred_return_blocked` tests). Otherwise, the performance improvement is slight, only up to 5% (See `test_yield_X_deferred_return_blocked` tests).

 - the `callback()` on an empty Deferred became as much as 10% slower (the `test_X_Y_deferred_blocked` tests). That's caused by the overhead of storing the failure status to an instance variable. This change improves performance on any more complex use cases.

## Contributor Checklist:


p12tic:
I've updated the PR and fixed tests. Measured the performance: the tests, measurement scripts and results are located at https://github.com/p12tic/gist-twisted-perf-tests. I'll edit the description with the results.

p12tic:
When reviewing, please verify the correctness of the explanatory comments I've added. They show how I understood the current Twisted codebase and if I did an error there, then further assumptions when writing code could've introduced much harder to spot bugs :-) Thanks!

p12tic:
Looks like I confused myself about the actual amount of improvement. I've updated the format of comparisons in the linked tests repo to reduce the chance of this.

glyph:
@p12tic This change is very exciting. I'd like to land it, and it's in pretty good shape to be landed, I think.

However, the most performance-sensitive Twisted users will be using PyPy and/or PyPy3.

Can you:

1. add pypy and pypy3 verification to your performance suite to make sure that those runtimes aren't regressed by this,
2. add your performance suite as a PR on https://github.com/twisted-infra/twisted-benchmarks so that they'll be visible on https://speed.twistedmatrix.com (I apologize for the somewhat haphazard / neglected state of that site, but speed center doesn't let us report GH statuses for performance changes which is really what we want for changes like this)
3. verify that there is a positive-to-neutral impact on *existing* Deferred performance tests for all existing performance tests on all the interesting runtimes?

Thanks so much for your hard work on this performance improvement, I'm really hoping we can get this landed soon!

glyph:
@p12tic Just pinging in the idle hope you'll have some time to look at this :)

p12tic:
Hey @glyph, thanks for review. I didn't have time to go back to this PR so far, but expect to be able to do so in the following week. I agree with all three of your points - if pypy is an important configuration we better make sure we know how the PR affects it.

glyph:
@p12tic Do you think that you might be able to resurrect this? I think @itamarst is doing some optimization work here, and we may actually be able to have workable benchmarks now that we've got codspeed set up

p12tic:
I think I could try to resurrect this.

glyph:
Thanks, that would be great! I believe the codspeed PR will be merged soon, so we can have a benchmark with a baseline. Cc @itamarst in case any coordination is required with his performance work

itamarst:
Benchmarks have been merged. They do not however run on PyPy, which hadn't occurred to me, so I'll see about adding that.

p12tic:
Many ideas of the PR have already been merged. One idea was not:

 - when the wrapped generator yields a Deferred to inlineCallbacks and we can prove that its result is going for us, we steal it instead of going through the .addCallback. .addCallback and .callback are relatively expensive and result in good performance improvement when they can be avoided.

Better way forward is to switch away from inlineCallbacks to proper async/await code.

PR #1091, opened 2018-12-03
Make QueryFactory a public class and document how to use it

ktdreyer:
Make `QueryFactory` a public class.

Give an example of how to debug raw XML-RPC responses using `xmlrpc.Proxy`.

## Contributor Checklist:

* [x] There is an associated ticket in Trac: https://twistedmatrix.com/trac/ticket/9350


ktdreyer:
It's surprising that we require 3 blank lines, since that is not the default for my editor (vim with stock flake8 rules). It would be great for newcomers if we could drop that, since it looks like the existing code does not follow that convention.

adiroiban (approved):
Thanks. It looks good.

For `<i8></i8>` maybe we can extend the current Twisted XML-RPC code to support it... not sure how common/standard is that type... but if is useful for you and implemented in other projects, it might be worth having it in Twisted.

----------

The 3 lines is in most of the code.
I use 2 empty lines as this is how most tools are defined, but I prefer the 3 lines... as it makes more sense :)

1 empty line for splitting blocks inside a method / function.
2 empty lines for splitting methods
3 empty lines for splitting functions/classes 

I find it harder to read method code where a single empty lines is used for both delimiting methods and code inside a method.

Thanks for your contribution!

PR #1095, opened 2019-01-23
construct and assign portal and checkers consistently

graingert:
The example uses a mix of constructor form and the register form.

Use only the constructor form, and create all class properties in the classdef

## Contributor Checklist:

* [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )

### DIFF
--- a/docs/conch/examples/sshsimpleserver.py
+++ b/docs/conch/examples/sshsimpleserver.py
@@ -185,6 +185,9 @@ class ExampleSession(object):
 
 components.registerAdapter(ExampleSession, ExampleAvatar, session.ISession)
 
+passwdDB = InMemoryUsernamePasswordDatabaseDontUse(user='password')
+sshDB = SSHPublicKeyChecker(InMemorySSHKeyDB(
+    {b'user': [keys.Key.fromFile(CLIENT_RSA_PUBLIC)]}))
 
 
 class ExampleFactory(factory.SSHFactory):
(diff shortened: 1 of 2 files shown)


docs/conch/examples/sshsimpleserver.py
```diff
@@ -224,15 +229,6 @@ def getPrimes(self):
...
 
 
-portal = portal.Portal(ExampleRealm())
-passwdDB = InMemoryUsernamePasswordDatabaseDontUse()
-passwdDB.addUser(b'user', b'password')
-sshDB = SSHPublicKeyChecker(InMemorySSHKeyDB(
-    {b'user': [keys.Key.fromFile(CLIENT_RSA_PUBLIC)]}))
-portal.registerChecker(passwdDB)
-portal.registerChecker(sshDB)
-ExampleFactory.portal = portal
```
graingert, line 234: when reading the code I was very confused about how ExampleFactory knew about the ExampleRealm and the various checkers.

IMHOP it's very unusual to define class properties outside of the classdef, so I missed this line completely

cdunklau:
Good catch, this example is certainly confusing, and shows some probably-bad practice what with the assigning to class attributes and doing stuff at import time... but it looks like your changes basically just shuffle this around. I think you can do better :)

Instead of assigning to class attributes at all, you could change all of that factory tweaking to just tweak an *instance* of the factory instead of the class itself. This would best be done under the `if __name__ == '__main__':` branch, or better yet in a `def main():` that it calls.

graingert:
You're right. factory class instances should be independent

I think this should be in the constructor

cdunklau:
Oh yeah, the newsfile should probably be a .doc, not a .misc

graingert:
closed in favor of https://github.com/twisted/twisted/pull/1096 for better branch name

PR #1096, opened 2019-01-24
construct and assign portal and checkers consistently in ssh server example 9578

graingert:
https://twistedmatrix.com/trac/ticket/9578

## Contributor Checklist:


cdunklau:
Looks better. Only thing I'd change is to accept the dependencies as args to the factory's `__init__` and just set them as attributes instead of initializing them inside. Then just create those things in the runner bit and pass them into `ExampleFactory(...)`.

docs/conch/examples/sshsimpleserver.py
```diff
@@ -217,22 +217,20 @@ class ExampleFactory(factory.SSHFactory):
         b'ssh-connection': connection.SSHConnection
     }
 
+    def __init__(self):
```
markrwilliams, line 228: I agree with [@cdunklau's comment](https://github.com/twisted/twisted/pull/1096#issuecomment-457143556) - the constructor should accept these arguments.  That can be done in a separate PR, though.

markrwilliams (approved):
Thanks for the contribution! It's more idiomatic and it doesn't break the example, so :+1:

glyph:
@markrwilliams - @graingert isn't a project member and can't merge this :).  Did you intend to land this change?

markrwilliams:
@glyph Yes, but trunk was broken when I reviewed this!

graingert:
@markrwilliams @glyph can you re-review?

graingert:
@markrwilliams I can't see why this is failing CI - this is example code it's not actually run/tested?

glyph:
The lint job is noisy, but it's this line:

```
./docs/conch/examples/sshsimpleserver.py:234:1: E305 expected 3 blank lines after class or function definition, found 2
```

glyph:
The Windows failures on Azure *seem* spurious, but they make me suspicious since they're also Conch stuff.

PR #1098, opened 2019-01-30
Refactor Clock to use heapq instead of a sorted list

msbrogli:
Ticket: https://twistedmatrix.com/trac/ticket/9581#ticket


src/twisted/internet/task.py
```diff
@@ -780,24 +781,16 @@ def seconds(self):
...
-                               self.calls.remove,
-                               lambda c: None,
-                               self.seconds)
-        self.calls.append(dc)
-        self._sortCalls()
+                              what, a, kw,
+                              lambda c: None,
+                              lambda c: None,
+                              self.seconds)
+        heapq.heappush(self.calls, (dc.getTime(), dc))
```
markrwilliams: `Clock.calls` is part of `Clock`'s public API, so changing its internal structure breaks calling code including Twisted's own tests:
https://travis-ci.org/twisted/twisted/jobs/488068561#L4517

Fortunately, [reactors have used heaps to sort `DelayedCall`s](https://github.com/twisted/twisted/blob/411d08f688bd8cdf61650e0ea1b24e99e14330e0/src/twisted/internet/base.py#L823-L830) for [a long time](https://twistedmatrix.com/trac/ticket/707), so the class implements the [the methods necessary for ordering](https://github.com/twisted/twisted/blob/411d08f688bd8cdf61650e0ea1b24e99e14330e0/src/twisted/internet/base.py#L164-L181).

That means you can push the `DelayedCall` onto the heap directly:

```suggestion
        heapq.heappush(self.calls, dc)
```

src/twisted/internet/task.py
```diff
@@ -818,12 +811,14 @@ def advance(self, amount):
         time.
         """
         self.rightNow += amount
-        self._sortCalls()
-        while self.calls and self.calls[0].getTime() <= self.seconds():
-            call = self.calls.pop(0)
-            call.called = 1
-            call.func(*call.args, **call.kw)
-            self._sortCalls()
+        while self.calls:
```
markrwilliams: This can remain
```suggestion
        while self.calls and self.calls[0].getTime() <= self.seconds():
```
but...

src/twisted/internet/task.py
```diff
@@ -818,12 +811,14 @@ def advance(self, amount):
...
+            time, call = self.calls[0]  # smallest item
```
markrwilliams: ...you'll have to replace this...
```suggestion
            call = heapq.heappop(self.calls)
```

markrwilliams (changes requested):
We can't change the contents of `Clock.calls` but fortunately we don't have to - see inline comments.

msbrogli:
@markrwilliams Done. Let me know if you have any other comment, please.

glyph (changes requested):
Thanks for this contribution!  It would be great to make our `IReactorClock` implementations more efficient and maintain them to include less duplication overall.

However, the tests are failing, apparently because you believed `DelayedCall` instances were orderable; they're not, they implement only `__lt__` and `__le__`, so this needs changes.

It would be fine to apply e.g. `@functools.total_ordering` to make them sortable if you wanted to do that.

(When you've gotten the tests passing again, please be sure to re-attach the `review` keyword in Trac so that someone will notice the PR again.)

Thanks again!

glyph:
@msbrogli bump?

PR #1099, opened 2019-02-03
inlineCallbacks docstring: remove contradiction

tkluck:
One could argue that now that `return <value>` just works, we should document that as the preferred option, and make a side note about `returnValue()` for backwards compatibility.

That decision is not mine to make, however; I'll just contribute this patch to fix an obvious contradiction in the docstring.

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1560,11 +1560,11 @@ def inlineCallbacks(f):
 
     Your inlineCallbacks-enabled generator will return a L{Deferred} object, which
     will result in the return value of the generator (or will fail with a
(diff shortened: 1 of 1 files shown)


glyph:
Thanks for your interest in making Twisted better!

However, since this PR doesn't follow any of our contribution guidelines, it's unlikely to be merged, so I'm going to close it to clean up the PR list.  If anyone would like to shepherd it through the process, please feel free to do so!

tkluck:
Given the size and scope of this PR (tiny and tiny, respectively), unfortunately I cannot make time to find out about, or study your guidelines. I have no interest in pursuing this further, but you may wish to reconsider whether twisted wouldn't benefit from people making drive-by fixes like this one.

PR #1100, opened 2019-02-11
9580 cryptography 2.5 openssl 1.1.1 tls 1.3 fixes

markrwilliams:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9580

### DIFF
--- a/src/twisted/conch/client/knownhosts.py
+++ b/src/twisted/conch/client/knownhosts.py
@@ -489,7 +489,7 @@ class KnownHostsFile(object):
 
                 keytype = key.type()
 
-                if keytype is "EC":
+                if keytype == "EC":
                     keytype = "ECDSA"
 
                 prompt = (
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1764,6 +1764,12 @@ def _expandCipherString(cipherString, method, options):
     try:
         ctx.set_cipher_list(cipherString.encode('ascii'))
     except SSL.Error as e:
+        # OpenSSL 1.1.1 turns an invalid cipher list into TLS 1.3
+        # ciphers, so pyOpenSSL >= 19.0.0 raises an artificial Error
+        # that lacks a corresponding OpenSSL error if the cipher list
+        # consists only of these after a call to set_cipher_list.
+        if not e.args[0]:
+            return []
         if e.args[0][0][2] == 'no cipher match':
             return []
         else:
--- a/src/twisted/internet/_sslverify.py
+++ b/src/twisted/internet/_sslverify.py
@@ -1470,7 +1470,7 @@ class OpenSSLCertificateOptions(object):
         @type raiseMinimumTo: L{TLSVersion} constant
 
         @param insecurelyLowerMinimumTo: The minimum TLS version to use,
-            possibky lower than Twisted's default.  If not specified, it is a
+            possibly lower than Twisted's default.  If not specified, it is a
             generally considered safe default (TLSv1.0).  If you want to raise
             your minimum TLS version to above that of this default, use
             C{raiseMinimumTo}.  DO NOT use this argument unless you are
--- a/src/twisted/test/test_ssl.py
+++ b/src/twisted/test/test_ssl.py
@@ -32,6 +32,8 @@ except ImportError:
 
 from zope.interface import implementer
 
+
+
 class UnintelligentProtocol(basic.LineReceiver):
     """
     @ivar deferred: a deferred that will fire at connection lost.
--- a/src/twisted/test/test_ssl.py
+++ b/src/twisted/test/test_ssl.py
@@ -15,7 +15,8 @@ from twisted.python.runtime import platform
 from twisted.test.test_tcp import ProperlyCloseFilesMixin
 from twisted.test.proto_helpers import waitUntilAllDisconnected
 
-import os, errno
+import os
+import hamcrest
 
 try:
     from OpenSSL import SSL, crypto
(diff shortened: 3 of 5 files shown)


markrwilliams:
@alex Are you available to review this?

src/twisted/test/test_ssl.py
```diff
@@ -309,33 +305,26 @@ def getHandleExceptionType(self):
...
-            if platform.getType() == 'win32':
-                return errno.WSAENOTSOCK
-
-        # Otherwise, we expect an error about how we tried to write to a
-        # shutdown connection.  This is terribly implementation-specific.
-        return [('SSL routines', 'SSL_write', 'protocol is shutdown')]
+        # We expect an error about how we tried to write to a shutdown
+        # connection.  This is terribly implementation-specific.
+        return hamcrest.contains(
+            hamcrest.contains(
```
alex, line 319: Is it correct to nest two `contains`? I don't really know things about `hamcrest`

src/twisted/test/test_tcp.py
```diff
@@ -1190,10 +1194,13 @@ def clientDisconnected(result):
             if not server.lostConnectionReason.check(error.ConnectionClosed):
                 err(server.lostConnectionReason,
                     "Server lost connection for unexpected reason")
-            expectedErrorCode = self.getHandleErrorCode()
+            errorCodeMatcher = self.getHandleErrorCodeMatcher()
```
alex, line 1197: Why is this change necessary?

alex (commented):
I feel like the newsfragment should say that twisted now officially supports TLS1.3?

src/twisted/test/test_tcp.py
markrwilliams: The change in `test_ssl` that adds the nested `contains` matcher is in a test that inherits from _this_ test, which expects the `getHandleErrorCodeMatcher` "abstract method" to return something to compare against the raised exception's first argument.

Now the OpenSSL implementation detail could be one of two messages, and I happen to have introduced the PyHamcrest dependency, so I decided to address that ambiguity with a matcher, and changed the method name (which is in a test and not `proto_helpers`, so a private API) to reflect that.

markrwilliams:
> I feel like the newsfragment should say that twisted now officially supports TLS1.3?

I don't know that it actually does.  What I know is that it supports enough of TLS 1.3 to make our tests pass!

I guess the point of TLS 1.3 is that supporting its defaults is equivalent supporting TLS 1.3?

alex (commented):
This probably is fine, but to be honest I find hamcrest so difficult to read that I'm uncomfortable actually approving this and would suggest a different reviewer should have a look.

src/twisted/conch/client/knownhosts.py
```diff
@@ -489,7 +489,7 @@ def promptResponse(response):
 
                 keytype = key.type()
 
-                if keytype is "EC":
+                if keytype == "EC":
```
exarkun, line 492: Great.

src/twisted/internet/_sslverify.py
```diff
@@ -1470,7 +1470,7 @@ def __init__(self,
         @type raiseMinimumTo: L{TLSVersion} constant
 
         @param insecurelyLowerMinimumTo: The minimum TLS version to use,
-            possibky lower than Twisted's default.  If not specified, it is a
+            possibly lower than Twisted's default.  If not specified, it is a
```
exarkun, line 1473: Nice.

src/twisted/internet/_sslverify.py
```diff
@@ -1764,6 +1764,12 @@ def _expandCipherString(cipherString, method, options):
     try:
         ctx.set_cipher_list(cipherString.encode('ascii'))
     except SSL.Error as e:
+        # OpenSSL 1.1.1 turns an invalid cipher list into TLS 1.3
+        # ciphers, so pyOpenSSL >= 19.0.0 raises an artificial Error
+        # that lacks a corresponding OpenSSL error if the cipher list
+        # consists only of these after a call to set_cipher_list.
+        if not e.args[0]:
+            return []
```
exarkun, line 1772: Boy OpenSSL is baaaaaaaaaaaaaaaaaaaaaaad.  But, uh, I guess this one is at least half pyOpenSSL's fault.  Raising a bare `Error`?  pyOpenSSL goes out of its way to identify this case and then encodes that information in this extremely lossy form!  Lame!  I suggest this is worthy of a pyOpenSSL bug report.  The least pyOpenSSL could do is raise an exception specifically encoding the information it has uncovered here.

src/twisted/test/test_ssl.py
```diff
@@ -564,9 +553,6 @@ def checkResult(failure):
             [clientProtocol.done.addErrback(checkResult),
              serverProtocol.done.addErrback(checkResult)])
 
-    if newTLS is None:
-        test_bothSidesLoseConnection.skip = "Old SSL code doesn't always close cleanly."
-
```
exarkun, line 569: This seems unrelated?

src/twisted/test/test_tcp.py
```diff
@@ -1135,17 +1138,18 @@ def getHandleExceptionType(self):
...
+        socket handle.
         """
         # Windows and Python 3: returns WSAENOTSOCK
         # Windows and Python 2: returns EBADF
         # Linux, FreeBSD, macOS: returns EBADF
         if platform.isWindows() and _PY3:
-            return errno.WSAENOTSOCK
-        return errno.EBADF
+            return hamcrest.equal_to(errno.WSAENOTSOCK)
+        return hamcrest.equal_to(errno.EBADF)
```
exarkun, line 1152: I might say that it is more in the spirit of matchers for the conditional here to be _inside_ the matcher.  That does involve writing more code, though, and more code is worse than less code, so ... :man_shrugging:

exarkun (approved):
Thanks.  Looks pretty good.  Some inline comments.  Address as you like and then merge.

src/twisted/internet/_sslverify.py
glyph: Filed: https://github.com/pyca/pyopenssl/issues/819

jMyles:
> I feel like the newsfragment should say that twisted now officially supports TLS1.3?

@alex: Do I understand correctly, though, that there is no `TLSv1_2_METHOD` yet in pyOpenSSL?  Does that mean that, for the moment, even though there's support in Twisted and Cryptography 2.5, that there's still not quite a way to use it in eg `DefaultOpenSSLContextFactory`?  

Or is there another way to do TLS 1.3 that I'm missing?

glyph:
@jMyles Did you mean `TLSv1_3_METHOD`?  Generally `SSLv23_METHOD` is the only method you should use, because it doesn't mean "SSL 2 or SSL 3" it means "negotiate the version of TLS to use based on the handshake".  You can disable individual protocol versions with `SSL_OP_NO_*` as options, rather than methods; or you can use `SSL_CTX_set_min_proto_version`.  See https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_new.html

jMyles:
@glyph: Indeed I did mean `TLSv1_3_METHOD` - that was a typo.

And my use case involves both the client and the server running my code; how can I guarantee the use of TLS 1.3?

glyph:
Sigh. We need support in PyDoctor for our various enum-alikes (and real enums as well) since the answer *should* be on this page:

https://twistedmatrix.com/documents/19.2.1/api/twisted.internet.ssl.TLSVersion.html

glyph:
Here's the official place to find the answer:

https://twisted.readthedocs.io/en/twisted-19.2.0/core/howto/ssl.html#tls-protocol-options

glyph:
In brief: `options = CertificateOptions(..., raiseMinimumTo=TLSVersion.TLSv1_3)`

This is presuming that when TLSv1.4 comes out, you don't want the code to *preclude* negotiating TLS 1.4.

PR #1101, opened 2019-02-12
Don't use ctypes.util.find_library()

jwilk:
[ctypes.util.find_library()](https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library) is [buggy](https://github.com/docker-library/python/issues/111), and unnecessary for Twisted. Let's get rid of it.

This PR is provided as-is. It wasn't tested. I don't even use Twisted.


glyph:
Thanks for your interest in making Twisted better!

However, since this PR doesn't follow any of our contribution guidelines, it's unlikely to be merged, so I'm going to close it.  But this definitely looks like something that might be worth fixing, so if anyone would like to file a ticket in trac, make the tests pass - please feel free!

PR #1102, opened 2019-02-16
cryptography ec serialization deprecations

glyph:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9584#ticket

### DIFF
--- a/src/twisted/test/test_sslverify.py
+++ b/src/twisted/test/test_sslverify.py
@@ -206,7 +206,9 @@ def certificatesForAuthorityAndServer(serviceIdentity=u'example.com'):
     try:
         ipAddress = ipaddress.ip_address(serviceIdentity)
     except ValueError:
-        subjectAlternativeNames = [x509.DNSName(serviceIdentity)]
+        subjectAlternativeNames = [
+            x509.DNSName(serviceIdentity.encode("idna").decode("ascii"))
+        ]
     else:
         subjectAlternativeNames = [x509.IPAddress(ipAddress)]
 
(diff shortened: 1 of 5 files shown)


glyph:
@reaperhulk

src/twisted/test/test_sslverify.py
```diff
@@ -206,7 +206,9 @@ def certificatesForAuthorityAndServer(serviceIdentity=u'example.com'):
     try:
         ipAddress = ipaddress.ip_address(serviceIdentity)
     except ValueError:
-        subjectAlternativeNames = [x509.DNSName(serviceIdentity)]
+        subjectAlternativeNames = [
+            x509.DNSName(serviceIdentity.encode("idna").decode("ascii"))
```
reaperhulk, line 210: This is resolving a different deprecation. It looks right but I don’t know what your PR policies are on that.

reaperhulk (approved):
LGTM modulo one comment

reaperhulk:
You’ll need to bump your minimum cryptography version to 2.5 as well

glyph:
Re-spinning the build wheel on the failing builders just to make sure I’m not missing anything.

glyph:
Thanks @reaperhulk!

src/twisted/test/test_sslverify.py
glyph: Given that this would also be a .misc it’s nbd; if the change were something meriting is own news fragment it would need to be a different PR.

glyph:
We don’t really have good coverage of our version minimum pins (see for example the fact that no builders failed this PR) so I will defer the issue of updating dependency metadata. It’s illegal to have an old cryptography anyway, right?

PR #1105, opened 2019-02-26
[9135] typecheck url path of Resource.putChild()

wiml:
* [X] There is an associated ticket in Trac: [ticket 9135](https://twistedmatrix.com/trac/ticket/9135)

This is a minimal committable version of @lthibault (Louis Thibault)'s change from PR #779 : it warns instead of raising, and it has a test, docstring update, and newsfragment.

Although PR #779 has several good suggestions for additional behavior, I don't see a clear consensus or obviously-right path. But almost all suggestions start with this behavior as a base. We can eliminate 90% of the developer frustration just by emitting useful warnings here, and this change clearly improves Twisted. So I think we should incorporate this part of the behavior now and discuss refinements as future tickets.


rodrigc (approved):
This looks reasonable.  I would rather this error out, since that makes things easier to catch when things fail fast.  Not everyone runs with warnings turned on.

However, looks like people complained a lot by going down that path.

So, approving this.

wiml:
I agree with you about erroring out instead of warning, but I think @mithrandi has a good point in the other PR about it being a breaking change of sorts. And if there's one thing I like less than buggy software, it's software I'm afraid to upgrade because someone might have broken something I use. :)

At any rate, adding a warning paves the way towards making it an exception later.

rodrigc:
@wiml When I ported large chunks of Buildbot to Python 3, I had to hack Twisted to error out here because it was too hard to figure out how Buildbot was using str/bytes incorrectly with this stuff.
Otherwise on Python 3, things would silently "succeed", but then fail in weird places which were hard to debug.

But good you are pushing this forward, the way you did.  I'll push this along.

src/twisted/web/resource.py
```diff
@@ -214,8 +214,23 @@ def putChild(self, path, child):
         intended to have the root of a folder, e.g. /foo/, you want
         path to be ''.
 
+        @param path: A single path component.
+        @type path: C{bytes}
```
rodrigc: This should be L{bytes}

wiml: Are you sure? `C{}` and `L{}` are both used for rendering code / identifiers, it's just that `C{}` doesn't attempt to link, right? The other references to non-Twisted identifiers in that file, including `bytes`, use `C{}` (which makes sense to me; there's not much point in linking that identifier to the Python documentation for the `bytes` type).

rodrigc: Se: https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

glyph:
Can we not fix this to just work correctly?  i.e. keep the internal storage in bytes, but encode things according to hyperlink's rules for normalized path segments if we see unicode?

src/twisted/web/resource.py
wiml: Hm, that's pretty clear (even if it's not how I'd do it myself)! I'll make the change.

wiml:
> Can we not fix this to just work correctly?

I think the correct behavior *is* to reject Unicode strings, or at the very least, to reject strings not encodable in plain 7-bit ASCII. Path segments are semantically byte-strings, not character-strings. (Hyperlink deals with [IRIs](https://tools.ietf.org/html/rfc3987), I suppose; HTTP deals with URIs.) It is *very common* for path segments to be UTF-8-encoded text, but not necessary, and exceptions aren't all that rare — Latin-1 or other 8-bit script-specific character sets for example.

In my experience, papering over the distinction between bytes and characters at the wrong level leads to more errors over time. It's a real distinction, though an annoying one.

In the case of this particular API, it's a somewhat low level API, and strings passed to it are likely to be either literals in the source code, or arbitrary text passed in from elsewhere. In the first case, adding the `b''` is easy enough. In the second case, the programmer will need to figure out the applicable IRI/URI handling for their situation — arguably, resource traversal needs to know if the path segment should be interpreted as UTF-8 so that it can perform a normalization-aware child lookup. A `Resource` class that does that would undoubtedly be handy, but is 'way beyond the scope of this bug.

I think it would be reasonable to extend `putChild` to accept unicode strings if&f they're plain ASCII, but the documentation should probably be clear that it is merely a convenience for the 95% common case, and that if you truly expect to be handling IRIs, you'll need to supply some of your own batteries.

lthibault:
> I think the correct behavior is to reject Unicode strings, or at the very least, to reject strings not encodable in plain 7-bit ASCII. Path segments are semantically byte-strings, not character-strings. 

@wiml Perhaps I'm reading your comment incorrectly, but doesn't this argue for throwing a hard error?

As I said in https://github.com/twisted/twisted/pull/779, there's no sensible reason to pass anything other than ASCII strings (or unicode equivalents), and anything else is just a latent bug.

I firmly believe we're not doing anyone any favors by treating these as anything short of a `TypeError`.

wiml:
> I firmly believe we're not doing anyone any favors by treating these as anything short of a `TypeError`.

Honestly, I agree. I'm just trying to find the most agreeable compromise so we can get *something* committed and move forward a little bit.

rodrigc:
@lthibault and @wiml I agree with both of you.  https://github.com/twisted/twisted/pull/779 was definitely the way to go.  It was a simple change.

I've made that change locally in my Twisted tree when I ported large sections of Buildbot to Python 3, and having the TypeError allowed me to quickly find and fix incorrect uses of putChild().
Not everyone enables DeprecationWarnings or pays attention to them, so I don't find them super useful for tracking down bugs like this.

However, if @mithrandi [vetoed that change](https://github.com/twisted/twisted/pull/779#pullrequestreview-37986180) citing the Twisted compatibility policy, then I see no other way to move forward other than this patch.

mithrandi:
I definitely want a `TypeError` after the deprecation period; but I like glyph's suggestion too. https://url.spec.whatwg.org/#url-representation is pretty explicit about preferring UTF-8.

wiml:
I think that encoding the path segment into bytes is only half of what's needed to support IRIs; the resource must also perform a unicode-aware lookup of the child resource when handling a request. It seems to me that that's not quite compatible with strict URI behavior, so I think it should be opted-in to, either with a flag on Resource or a subclass (UnicodeResource or something).

I think the most conservative, least existing-code-breaking, least incorrect thing to do is this:

1. Open a distinct bug for Unicode-aware path segment handling, referencing these discussions
2. Merge this `DeprecationWarning` into Twisted
3. In a future release, convert the warning into a `TypeError` — unless a better idea has come along in the meantime

Unless someone objects, I'll do that?

rodrigc:
@lthibault @wiml I've gone ahead and merged this, and changed the newsfragment to a *removal* so that it shows up in the deprecated list in the release notes.  As per the [Twisted Compatibility Policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#incompatible-changes), after the next release of Twisted goes out with this change,
then we can change the behavior of putChild() to error out on non-bytes input either one year after that release, or 2 releases subsequent, whichever comes first.

PR #1107, opened 2019-03-05
two small fixes for python3

fchapoton:
and one tiny fix for speed

### DIFF
--- a/src/twisted/mail/relaymanager.py
+++ b/src/twisted/mail/relaymanager.py
@@ -353,7 +353,7 @@ class Queue:
         @rtype: L{list} of L{bytes}
         @return: The base filenames of messages waiting to be relayed.
         """
-        return self.waiting.keys()
+        return list(self.waiting.keys())
 
 
     def hasWaiting(self):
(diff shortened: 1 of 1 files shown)


src/twisted/mail/relaymanager.py
```diff
@@ -365,7 +365,7 @@ def hasWaiting(self):
         @return: C{True} if messages are waiting to be relayed.  C{False}
             otherwise.
         """
-        return len(self.waiting) > 0
+        return bool(len(self.waiting))
```
wiml, line 368: Out of curiosity, what is the motivation for this change? If the intent is to avoid extra operations, would `bool(self.waiting)` be better?

fchapoton:
Well, I have been told that computing the len only to see if it is not zero
is much slower than using bool, as bool has a smarter and faster way to see
if a list is empty or not.

fchapoton:
oops, it seems that my change is wrong indeed..

fchapoton:
Damn.. I have deleted my fork, and made a new one. But now I am no longer able to change anything to this pull request. Sorry.

PR #1109, opened 2019-03-06
[9593] Align IRequest/Request getHeader/setHeader types

twm:
Make the documentation match the implementation.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9593#comment:1
* [x] I have updated the automated tests. N/A


src/twisted/web/iweb.py
```diff
@@ -53,18 +53,26 @@ def getHeader(key):
         """
         Get an HTTP request header.
 
-        @type key: C{str}
+        @type key: C{bytes} or C{str}
```
wiml: The [Twisted coding standard page](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings) says that even references to Python's builtin types should use `L{}` instead of `C{}`.

src/twisted/web/http.py
```diff
@@ -974,12 +974,13 @@ def getHeader(self, key):
         """
         Get an HTTP request header.
 
-        @type key: C{bytes}
```
wiml, line 977: Is it necessary to duplicate interfaces' docstrings on their implementing classes when they're identical? Current practice is mixed, and it seems redundant, since the generated documentation will come out the same either way (the implementing class's methods will be documented with text taken from the interface).

wiml: (I should add that that comment isn't intended as an objection to this PR.)

src/twisted/web/iweb.py
twm: :+1: Good point, I was just mindlessly copying the existing style.

src/twisted/web/http.py
twm: I know that I've seen pydoctor inherit the docstrings before, so I removed them dupes from this PR, but it doesn't seem to be working properly in the case of `http.Request` or `server.Request`:

image

image

pydoctor is especially confused in the latter case &mdash; it lists the `Componentized` methods several times.

hawkowl: (note: pydoctor's confusion there is fixed on current pydoctor, as seen on 18.9s docs)

twm: Yes, the `Componentized` bit is fixed with pydoctor trunk.

It still doesn't inherit the documentation, though. `method`, `uri`, `path`, and `args` aren't shown, and the methods show up as undocumented:

image

twm: Maybe this is happening because of the mix of inheritance and interface annotation here?

```
http.Request      implements    IConsumer
     |            defines       setResponseCode
     |                          setHeader
     |
server.Request    implements    IRequest
```

In any case, since `http.Request` doesn't implement `IRequest` it needs to document these methods and attributes on its own. (Indeed, there are other subclasses of `http.Request` that don't implement `IRequest` either — `ProxyRequest` and `ReverseProxyRequest`).

glyph (approved):
My only quibble with this as-is is that it's such a big upgrade to the docs for this super core interface that I feel like it might deserve more than a `.misc` ;).  But by all means, let's land it soon!

twm:
Thank you for the review, @glyph! I've added a .doc newsfile and will merge when the build goes green.

twm:
Now all the buildbots are failing like this:

```
  File "c:\buildslave\windows7-64-py2_7-coverage\twisted\build\py27-alldeps-withcov-windows\lib\site-packages\twisted\python\test\test_url.py", line 486, in test_parseEqualSignInParamValue
    self.assertEqual(u.asText(), 'http://localhost/?=x%3Dx%3Dx')
  File "c:\buildslave\windows7-64-py2_7-coverage\twisted\build\py27-alldeps-withcov-windows\lib\site-packages\twisted\trial\_synctest.py", line 432, in assertEqual
  File "c:\python27\Lib\unittest\case.py", line 513, in assertEqual
  File "c:\python27\Lib\unittest\case.py", line 924, in assertMultiLineEqual
twisted.trial.unittest.FailTest: u'http://localhost/?=x=x=x' != u'http://localhost/?=x%3Dx%3Dx'
- http://localhost/?=x=x=x
?                     ^ ^
+ http://localhost/?=x%3Dx%3Dx
?                     ^^^ ^^^
```

The failure seems legitimate, so why weren't they before? I am so confused. -.-

twm:
image

New hyperlink?

twm:
Yes: https://twistedmatrix.com/trac/ticket/9619 :sob:

PR #1110, opened 2019-03-07
[9571] Errback any outstanding requests if the connection is lost

wiml:
* [X] There is [an associated ticket in Trac](https://twistedmatrix.com/trac/ticket/9571).


wiml:
I don't understand the two intermittent build failures. They're timeouts/hangs testing on py2.7, in conch, on Linux (fedora26 and rhel27), but using different reactors (poll and select), and failing on different tests each time. But I can't replicate the hangs locally, and I don't see how this change would cause them. I've [added a note to Trac #8879](https://twistedmatrix.com/trac/ticket/8879#comment:19).

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +531,21 @@ def connectionMade(self):
         self.sendPacket(FXP_INIT, data)
 
 
+    def connectionLost(self, reason):
+        """
+        Abort any pending requests.
+        """
+
+        # On Python2, we're an oldstyle class, so we can't use super().
+        FileTransferBase.connectionLost(self, reason)
```
adiroiban, line 550: Not sure if this should be called first or last.

I think that your implementation is right, and it should be called first.

Also, not sure if it make sense to call it at all. AFAIK there is no implementation of connectionLost in in super.

So, I think that instead of calling super() we should just set `self.connected = False` here.

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +531,21 @@ def connectionMade(self):
         self.sendPacket(FXP_INIT, data)
 
 
+    def connectionLost(self, reason):
+        """
+        Abort any pending requests.
```
adiroiban: Not sure what to say about this docstring.... my take was

```
Called when connection to the remote subsystem was lost.
```

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +531,21 @@ def connectionMade(self):
...
+        """
+
+        # On Python2, we're an oldstyle class, so we can't use super().
+        FileTransferBase.connectionLost(self, reason)
+
+        # If there are still requests waiting for responses when the
+        # connection is lost, fail them.
+        while self.openRequests:
+            _, deferred = self.openRequests.popitem()
+            deferred.errback(reason)
```
adiroiban: This is ok... but in my tests, `reason` was always `ConnectionDone`

Do you get a different Exception on errors?

My code is something like this... I call super last, but I set `connected = False` first.


```
    def connectionLost(self, reason=ConnectionDone):
        """
        Called when connection to the remote subsystem was lost.

        The way Twisted SSH is implemented,  this is never call with anything
        else that a ConnectionDone, even when the connection is lost.
        """
        self.connected = False

        while self.openRequests:
            # There are still requests waiting to be executed while the
            # client is disconnected.
            # We assume that the connection was lost, as otherwise, we should
            # have not seen open requests here.
            _, deferred = self.openRequests.popitem()
            deferred.errback(ConnectionLost('SFTP client request canceled'))

            # The request might have an associated file handler.
            self.wasAFile.pop(deferred, None)

        self._notifyConnectionLost(reason)   # this is my own extra code to allow notifications
        super(SFTPClientSubsystem, self).connectionLost(reason)
```

I have multiple places/things using a single SFTP client connection and I want them to be notify when connection is lost...
with your current implementation, only the places which have active calls will know that connection is down.

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -572,6 +576,46 @@ def test_openDirectoryIteratorDeprecated(self):
...
+        self.assertFalse(d.called)
+
+        # Lost connection should cause an errback
+        self.serverTransport.loseConnection()
+        self.serverTransport.clearBuffer()
+        self.clientTransport.clearBuffer()
+        self._emptyBuffers()
+
+        self.assertTrue(d.called)
+        self.assertFailure(d, ConnectionClosed)
```
adiroiban: I think that we should expect something more specific than a generic ConnectionClosed... which is called whether it was closed cleanly or non-cleanly.

I think ConnectionLost is more appropriate

adiroiban (changes requested):
Happy to see that you are sending this patch. Thanks!

I also hit this issue.

My suggestion is to also set `connected = False` here as there is no super implementation of connectionLost

So to code can look like

```
    def connectionLost(self, reason=ConnectionDone):
        """
        Called when connection to the remote subsystem was lost.

        The way Conch SSH is implemented, this is never call with anything
        else that a ConnectionDone, even when the connection is lost.
        """
        self.connected = False

        while self.openRequests:
            # There are still requests waiting to be executed while the
            # client is disconnected.
            # We assume that the connection was lost, as otherwise, we should
            # have not seen open requests here.
            _, deferred = self.openRequests.popitem()
            deferred.errback(ConnectionLost('SFTP client request canceled'))

            # The request might have an associated file handler.
            self.wasAFile.pop(deferred, None)
```

Let me know what do you think about my initial review, and later we can check the tests.

Happy to see that you wrote a functional/integration tests.

I only got a simple unit test for my version :)

adiroiban:
Hi @wiml  sorry for my review... I now see that I have already created the ticket in Trac  :)...and you already responded.

So... my comment about `connected = False` still stands...as super connectionLost does nothing for the client-side code.

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -498,6 +498,10 @@ def connectionLost(self, reason):
         """
         Clean all opened files and directories.
         """
+
+        # On Python2, we're an oldstyle class, so we can't use super().
+        FileTransferBase.connectionLost(self, reason)
```
adiroiban, line 512: What is the point of this?

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -572,6 +576,46 @@ def test_openDirectoryIteratorDeprecated(self):
...
+        self._emptyBuffers()
+        fh = yield d
+
+        # Intercept the handling of the read request on the server side
+        gotReadRequest = []
+
+        def _slowRead(offset, length):
+            self.assertEqual(gotReadRequest, [])
+            d = defer.Deferred()
+            gotReadRequest.append(d)
```
adiroiban: this is ok... but what I am doing in similar cases is just use something like `gotReadRequest.append(True)` 

With your implementation, when I read the code I was thinking that you are doing something with the new Deferred as placed in the list.

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -572,6 +576,46 @@ def test_openDirectoryIteratorDeprecated(self):
...
+            return d
+        [serverSideFh] = self.server.openFiles.values()
+        serverSideFh.readChunk = _slowRead
+        del serverSideFh
```
adiroiban, line 600: why `del` here?

adiroiban:
Here is my unit test in case it helps

```
    def getConnectedSession(self):
        """
        Return a SFTP client subsystem which is already connected and
        authenticated.
        """
        client = SFTPClientSubsystem()

        connection = SSHConnection()
        connection.transport = SSHClientTransport()
        session = SSHChannel(
            conn=connection,
            remoteMaxPacket=1000,
            remoteWindow=1000,
            )
        session.id = 0
        connection.channelsToRemoteChannel[session] = 0

        client.makeConnection(session)

        self.assertTrue(sut.connected)
        return client

    def test_connectionLost_pending_operations(self):
        """
        When the connection is lost all pending SFTP client operations are
        canceled with a ConnectionLost message.
        """
        sut = self.getConnectedSession()
        # Trigger a few client request which will not be answered before the
        # connection is lost.
        deferred_1 = sut.removeFile(b'some-path')
        deferred_2 = sut.openDirectory(b'some-folder')

        sut.connectionLost()

        self.assertIsFalse(sut.connected)
        self.assertIsEmpty(sut.openRequests)
        self.assertIsEmpty(sut.wasAFile)
        failure_1 = self.failureResultOf(deferred_1)
        failure_2 = self.failureResultOf(deferred_2)
        self.assertIsInstance(ConnectionLost, failure_1.value)
        self.assertEqual(
            'SFTP client request canceled', failure_1.value.args[0])
        self.assertIsInstance(ConnectionLost, failure_2.value)
```

src/twisted/conch/ssh/filetransfer.py
wiml: I don't get a different exception, but in my opinion that's a bug in Conch — it should provide a different exception if the connection fails in a different way. So I wanted to pass through the exception in case it someday becomes meaningful.

Perhaps this code should pass through the exception, *unless* it is a `ConnectionDone`, in which case it should be converted to `ConnectionLost` (since it cannot be a clean closure if there are unfinished requests)?

adiroiban: Well... I tracked this and from what I can see this is a low error in Twisted client... I am seeing similar errors in Twisted Web client.

But I am OK to accept this... I would still like to see a different exeption here, as I imagine that many other people using this will have to patch this code in their own code.

src/twisted/conch/test/test_filetransfer.py
wiml: Yes, I'm not sure why I did that. I copied that piece of test code from elsewhere in Twisted. I'll clean it up.

src/twisted/conch/ssh/filetransfer.py
wiml: Even though our `super` implementation is empty, it still exists … I just think it is good practice to invoke the super method when there is one, in case that method gains more functionality in the future. 

However you are right it should update `self.connected`. (I wonder why `Protocol.connectionLost()` doesn't reset `connected` anyway.)

src/twisted/conch/ssh/filetransfer.py
wiml: I updated the PR to convert a ConnectionDone into a ConnectionLost.

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -498,6 +498,13 @@ def connectionLost(self, reason):
...
+        # Not sure why self.connected isn't updated in
```
adiroiban: I know why... the IProtocol don't have `connected` as an attribute/member.
And across Twisted API self.connected is not well implemented.... we have connecting... disconnected ... disconnecting ... etc.

I think that this comment can be removed...but keep the implementation

but if you are calling `super` here,  you might want to implement connect = false in FileTransferBase

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +534,32 @@ def connectionMade(self):
...
+    def connectionLost(self, reason):
+        """
+        Abort any pending requests.
+        """
+
+        # On Python2, we're an oldstyle class, so we can't use super().
+        FileTransferBase.connectionLost(self, reason)
+        # Not sure why self.connected isn't updated in
+        # t.i.p.Protocol.connectionLost() instead
+        self.connected = False
```
adiroiban: see previous comment about IProtocol

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +534,32 @@ def connectionMade(self):
...
+                # Even if our transport was lost "cleanly", our
+                # requests were still not cancelled "cleanly".
+                reqreason = failure.Failure(
+                    error.ConnectionLost(str(reason.value)))
+            else:
+                reqreason = reason
+            while self.openRequests:
+                _, deferred = self.openRequests.popitem()
+                deferred.errback(reqreason)
+
```
adiroiban, line 564: this is ok, but why not also clean self.wasAFile ?

src/twisted/conch/ssh/filetransfer.py
wiml: Perhaps it should say both?

src/twisted/conch/ssh/filetransfer.py
wiml: I think in general, if the superclass has an implementation, even if it's empty right now, it's good practice to call it, in case it becomes non-empty in the future. This isn't a performance-critical section of code or anything.

src/twisted/conch/test/test_filetransfer.py
wiml: Ah, just habit to clean up a reference that won't be needed before "blocking" on something later in the method. It makes no difference here.

glyph:
@adiroiban bump

if you re-review this, https://twisted.reviews will be *empty* 😄

src/twisted/conch/ssh/filetransfer.py
adiroiban: ok. Thanks

src/twisted/conch/ssh/filetransfer.py
adiroiban: ok. this was a minor comment...so I am ok with anything :)

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +534,32 @@ def connectionMade(self):
...
+        # On Python2, we're an oldstyle class, so we can't use super().
+        FileTransferBase.connectionLost(self, reason)
+        # Not sure why self.connected isn't updated in
+        # t.i.p.Protocol.connectionLost() instead
+        self.connected = False
+
+        # If there are still requests waiting for responses when the
+        # connection is lost, fail them.
+        if self.openRequests:
+            if reason.check(error.ConnectionDone):
```
adiroiban: why do we have this condition here.
my understanding is that this will always be called with ConnectionDone... so `else` is newer called.

src/twisted/conch/ssh/filetransfer.py
```diff
@@ -527,6 +534,32 @@ def connectionMade(self):
...
+                # Even if our transport was lost "cleanly", our
+                # requests were still not cancelled "cleanly".
+                reqreason = failure.Failure(
```
adiroiban: I would prefer to have a better name than `reqreason`... maybe just overwrite/reuse `reason`

And why reuse the message from the `ConnectionDone` failure ...as it is a misleading error.

I think that we should put a custom message here that should better describe what happened and should make it easier for people to track the failure in the code.

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -560,7 +562,9 @@ def test_openDirectoryIteratorDeprecated(self):
         d = self.client.openDirectory(b'')
         self._emptyBuffers()
         openDir = yield d
-        openDir.next()
+        d_ = openDir.next()
```
adiroiban: can we have a better name for this variable, instead of just `d_`?

In the same time, I feel like the `_emptyBuffers` is abused here.

I think it should be made a public method, and maybe rename to "triggerCommunication" or something like that.

There is not point to make it private, as we are in testing code.

The only purpose of using a private method in a test is when you write something that should be use only by a single test

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -572,6 +576,46 @@ def test_openDirectoryIteratorDeprecated(self):
...
+        self.assertNoResult(d)
+
+        # Lost connection should cause an errback
+        self.serverTransport.loseConnection()
+        self.serverTransport.clearBuffer()
+        self.clientTransport.clearBuffer()
+        self._emptyBuffers()
+
+        self.assertFalse(self.client.connected)
+        self.failureResultOf(d, ConnectionLost)
```
adiroiban, line 616: I think that here we should have something like 
```
failure = self.failureResultOf(d, ConnectionLost)
self.assertEqual('Our expected message', failure.value.args[0])
```

adiroiban (changes requested):
Thanks for working on this and sorry for the delay.

This is better, but I am not 100%  happy with the patch :)

Thanks!

adiroiban:
@glyph Please doublecheck my comments and weight your opinion. 
I was not sure about what is reasonable to ask...so I have delayed the review.
Thanks!

src/twisted/conch/ssh/filetransfer.py
wiml: Ah, `wasAFile` doesn't exist any more (was removed for Twisted 18.4)

adiroiban: Uh. Thanks. Sorry for the noise.

src/twisted/conch/ssh/filetransfer.py
glyph: `IProtocol` is the interface that the *transport* requires of the protocol, so it's deliberately minimal.  The commonality of attributes like `connecting`, `disconnected`, et. al., are mostly an accident of history.

src/twisted/conch/ssh/filetransfer.py
glyph: I'm also curious about this.

src/twisted/conch/ssh/filetransfer.py
wiml: `IProtocol` doesn't, but `t.i.p.BaseProtocol` does have a `connected` attribute and updates it in some situations, just not others; that's why it seems a little odd to me that `t.i.p.Protocol` (the first ancestor class that implements `connectionLost()`) doesn't update `connected`. Or, perhaps, since `Protocol` has no method bodies, its method docstring should suggest that implementers update `connected`? Or `BaseProtocol` should lose the `connected` attribute and it should be implemented in each subclass that wants it? This seems like a question for a separate PR. :)

In either case, I think you're right I should move that code up into `FileTransferBase`.

src/twisted/conch/ssh/filetransfer.py
wiml: Because  it seems like a bug that the `reason` is always `ConnectionDone` even in the case of a non-clean closure. Anyway, I've rewritten this not to special-case `ConnectionDone`.

src/twisted/conch/ssh/filetransfer.py
wiml: I've changed this to always use the default message for `ConnectionLost()` but to chain to the original exception with `__cause__`, which I think provides the best of both worlds.

adiroiban (approved):
It looks good. Thanks!

Now I need to see how to merge this... as I have not merged something in trunk for a very long time.

wiml:
@adiroiban I should have permissions to merge this, I'll do it later today or tomorrow if no one raises any concerns. Thanks for the reviews!

glyph:
@wiml @adiroiban Merging to trunk is just clicking the "merge" button on Github, and pasting in the commit template here: https://twistedmatrix.com/trac/wiki/ReviewProcess#Authors:Howtomergethechangetotrunk

PR #1115, opened 2019-03-31
[#9603] Remove deprecated twisted.conch.insults modules

hawkowl:
https://twistedmatrix.com/trac/ticket/9603#ticket


src/twisted/newsfragments/9603.removal
```diff
@@ -0,0 +1 @@
+twisted.conch.insults.client and twisted.conch.insults.colors, deprecated since Twisted 10.1, have been removed.
```
markrwilliams, line 1: It might be nice to include what to replace these with.

markrwilliams (approved):
Thank you for removing deprecated code after the deprecation period has passed!

PR #1119, opened 2019-04-02
[2782] Notify channel if connection fails during open

wiml:
* [X] There are associated tickets in Trac: [2782](https://twistedmatrix.com/trac/ticket/2782),  [9610](https://twistedmatrix.com/trac/ticket/9610)
* [X] I have created a newsfragment in src/twisted/conch/newsfragments/

Noticed and fixed 9610 while fixing 2782.


src/twisted/conch/ssh/connection.py
```diff
@@ -65,8 +66,16 @@ def serviceStopped(self):
         """
         Called when the connection is stopped.
         """
-        for channel in list(self.channels.values()):
+        # Close any fully open channels
+        for channel in list(self.channelsToRemoteChannel.keys()):
```
markrwilliams, line 70: For clients, `SSHConnection.openChannel` puts the channel object into the instance's `channels` dict and moves it to `channelsToRemoteChannel` only in `SSHConnection.ssh_CHANNEL_OPEN_CONFIRMATION`, so this has the effect of closing only open client -> server channels.

src/twisted/conch/ssh/connection.py
```diff
@@ -65,8 +66,16 @@ def serviceStopped(self):
...
             self.channelClosed(channel)
+        # Indicate failure to any channels that were in the process of
+        # opening but not yet open.
+        while self.channels:
```
markrwilliams, line 74: If `channel.openFailed` somehow synchronously reopen a channel -- for example, to reconnect -- this will loop infinitely.  It's not clear that will happen here, but it's why many places in Twisted will copy a list like this first, like so:

```python
channels, self.channels = self.channels, []
while channels:
    ...
```

src/twisted/conch/ssh/connection.py
```diff
@@ -605,11 +614,12 @@ def channelClosed(self, channel):
             del self.localToRemoteChannel[channel.id]
             del self.channels[channel.id]
             del self.channelsToRemoteChannel[channel]
-            for d in self.deferreds.setdefault(channel.id, []):
```
markrwilliams, line 608: Oops -- this would hold a strong reference to all the Deferreds, wouldn't it?

src/twisted/conch/test/test_connection.py
```diff
@@ -52,6 +52,7 @@ class TestChannel(channel.SSHChannel):
     """
     name = b"TestChannel"
     gotOpen = False
+    gotClosed = False
```
markrwilliams, line 55: The makes the `assertFalse`s below work.

src/twisted/conch/test/test_connection.py
```diff
@@ -241,9 +242,16 @@ def test_serviceStopped(self):
         self.conn.openChannel(channel2)
         self.conn.ssh_CHANNEL_OPEN_CONFIRMATION(b'\x00\x00\x00\x00' * 4)
         self.assertTrue(channel1.gotOpen)
+        self.assertFalse(channel1.gotClosed)
```
markrwilliams, line 245: Did you add these to make sure your changes didn't incorrectly close a channel?

markrwilliams (approved):
These changes make sense!  It would be good to consider the same case from the perspective of server to client.

Please answer the questions, but they shouldn't block this.

src/twisted/conch/ssh/connection.py
wiml: My understanding of the implicit contract that `channelClosed` is upholding with its test is that a channel that hasn't had either `channelOpen` or `openFailed` shouldn't receive a `closed` call either — the channel should be guaranteed to received either `channelOpen` or `openFailed` first, and to receive `closed` only after a `channelOpen`.

So in this case, the code I added in the next couple of lines is there to handle the channels which `channelClosed` refused to handle.

I could add an explanatory comment if you think it would be a good idea.

src/twisted/conch/ssh/connection.py
wiml: Good point, and synchronously reopening a channel isn't an unlikely response, either.

I habitually use this pattern because it correctly handles *most* cases of synchronous modifications to the container, both additions and removals. In this case I think the only downside is the possibility of getting in a loop with synchronous re-opens, but IMHO an attempt to open a channel while the connection is closing should simply raise an exception (analogously to [bug 9572](https://twistedmatrix.com/trac/ticket/9572)). I'll look into that and either update this PR or open a separate bug for it.

src/twisted/conch/test/test_connection.py
wiml: Yes, but also just to make the tests more comprehensive. We know what that attribute should be at that point; might as well add the assertion.

wiml:
Thanks for the review! I'll look at the "synchronously re-opening" behavior more closely and once somebody fixes the circleci problems I'll merge this.

glyph:
@wiml Hopefully the Circle problem is addressed now, and once this has worked its way through the build queue it should be mergeable.

src/twisted/conch/ssh/connection.py
wiml: I filed a bug here: https://twistedmatrix.com/trac/ticket/9626 ; I think you're right that it's a possibility. @markrwilliams if you have time to weigh in on that ticket and 9572 I'd appreciate any thoughts you have.

PR #1126, opened 2019-04-09
[#9620] Prevent twisted.names.client.Resolver from infinite looping

hawkowl:
https://twistedmatrix.com/trac/ticket/9620#ticket

### DIFF
--- a/src/twisted/names/client.py
+++ b/src/twisted/names/client.py
@@ -217,13 +217,26 @@ class Resolver(common.ResolverBase):
         Return a new L{DNSDatagramProtocol} bound to a randomly selected port
         number.
         """
+        failures = 0
         proto = dns.DNSDatagramProtocol(self, reactor=self._reactor)
+
         while True:
             try:
                 self._reactor.listenUDP(dns.randomSource(), proto,
                                         interface=interface)
-            except error.CannotListenError:
-                pass
+            except error.CannotListenError as e:
+                failures += 1
+
+                if (hasattr(e.socketError, "errno") and
+                   e.socketError.errno == errno.EMFILE):
+                    # We've run out of file descriptors. Stop trying.
+                    raise
+
+                if failures >= 1000:
+                    # We've tried a thousand times and haven't found a port.
+                    # This is almost impossible, and likely means something
+                    # else weird is going on. Raise, as to not infinite loop.
+                    raise
             else:
                 return proto
 
--- a/src/twisted/names/newsfragments/9620.bugfix
+++ b/src/twisted/names/newsfragments/9620.bugfix
@@ -1,0 +1,1 @@
+twisted.names.client.Resolver will no longer infinite loop if it cannot bind a UDP port to use for resolving.
--- a/src/twisted/names/test/test_client.py
+++ b/src/twisted/names/test/test_client.py
@@ -7,6 +7,8 @@ Test cases for L{twisted.names.client}.
 
 from __future__ import division, absolute_import
 
+import errno
+
 from zope.interface.verify import verifyClass, verifyObject
 
 from twisted.python import failure
@@ -560,6 +562,48 @@ class ResolverTests(unittest.TestCase):
 
         resolver._connectedProtocol()
         self.assertEqual(len(set(ports)), 2)
+
+    def test_disallowedPortRepeatedly(self):
+        """
+        If port numbers that cannot be bound are repeatedly selected,
+        L{resolver._connectedProtocol} will give up eventually.
+        """
+        ports = []
+
+        class FakeReactor(object):
+            def listenUDP(self, port, *args, **kwargs):
+                ports.append(port)
+                raise CannotListenError(None, port, None)
+
+        resolver = client.Resolver(servers=[('example.com', 53)])
+        resolver._reactor = FakeReactor()
+
+        self.assertRaises(CannotListenError, resolver._connectedProtocol)
+        # 1000 is a good round number. I like it.
+        self.assertEqual(len(ports), 1000)
+
+
+    def test_runOutOfFiles(self):
+        """
+        If the process is out of files, L{Resolver._connectedProtocol}
+        will give up.
+        """
+        ports = []
+
+        class FakeReactor(object):
+            def listenUDP(self, port, *args, **kwargs):
+                ports.append(port)
+                err = OSError(errno.EMFILE, "Out of files :(")
+                raise CannotListenError(None, port, err)
+
+        resolver = client.Resolver(servers=[('example.com', 53)])
+        resolver._reactor = FakeReactor()
+
+        exc = self.assertRaises(CannotListenError, resolver._connectedProtocol)
+        # The EMFILE-containing exception was raised, and it did not try
+        # multiple times.
+        self.assertEqual(exc.socketError.errno, errno.EMFILE)
+        self.assertEqual(len(ports), 1)
 
 
     def test_differentProtocolAfterTimeout(self):


src/twisted/names/client.py
```diff
@@ -217,13 +217,26 @@ def _connectedProtocol(self, interface=''):
         Return a new L{DNSDatagramProtocol} bound to a randomly selected port
         number.
         """
+        failures = 0
         proto = dns.DNSDatagramProtocol(self, reactor=self._reactor)
+
         while True:
             try:
                 self._reactor.listenUDP(dns.randomSource(), proto,
```
glyph, line 225: This should really be binding port 0 rather than a random port.  I saw some IRC chatter where @exarkun was claiming that this doesn't work, but even if I believed him about this[1]:

1. we have lots of *other* places in Twisted where we bind to port 0 and expect it to work, so it's not clear why this one is special
2. port 0 binds the ephemeral port range, which is what clients should use; this binds some totally random port, which might be reserved or even privileged, which means it could get lucky and grab, say, 53, if this process has the appropriate cap, which might make for some hilarious failures.

so if it *is* going to bind a random port, it should be using `SystemRandom().randint(0xc000, 0xffff)` and not this range.

[1]: at least right now, I don't; the source cited was failed Tahoe-LAFS builds, I didn't see a link to any so I can't investigate their causes for myself, and I'm much more willing to believe there are bugs in Tahoe-LAFS than a failure of basic guarantees in modern Linux TCP stacks.  particularly because Twisted's test suite isn't particularly reliable these days, also depends on binding port 0, and I have not personally ever seen a failure that results from port 0 being unbindable.

src/twisted/names/client.py
```diff
@@ -217,13 +217,26 @@ def _connectedProtocol(self, interface=''):
...
-                pass
+            except error.CannotListenError as e:
+                failures += 1
+
+                if (hasattr(e.socketError, "errno") and
+                   e.socketError.errno == errno.EMFILE):
+                    # We've run out of file descriptors. Stop trying.
+                    raise
+
+                if failures >= 1000:
```
glyph, line 235: This seems like an odd idiom to repeat a thing.  Why not `for port in (randomPort() for each in six.range(1000)):` with a `return proto` in the `try` block?

src/twisted/names/client.py
```diff
@@ -217,13 +217,26 @@ def _connectedProtocol(self, interface=''):
...
         while True:
             try:
                 self._reactor.listenUDP(dns.randomSource(), proto,
                                         interface=interface)
-            except error.CannotListenError:
-                pass
+            except error.CannotListenError as e:
+                failures += 1
+
+                if (hasattr(e.socketError, "errno") and
```
glyph, line 230: I appreciate that retrying in a hot loop is unlikely to fix `EMFILE`, but this could be made much more reliable in FD-pressure circumstances by doing a *slow* retry with a `LoopingCall` and not violating the API contract of Resolver.queryUDP (i.e.: synchronously raising where it doesn't document exceptions and specifically says that it returns a `Deferred` that fails with `TimeoutError`.)

glyph (approved):
Thanks for fixing this.

This is a nasty bug and there's nothing wrong with the fix as such, so let's land it.

It would be great if you could address some of the feedback in a "yes, and" follow-up though!

src/twisted/names/client.py
exarkun: Here's a link to a failed Tahoe-LAFS build:

https://circleci.com/gh/tahoe-lafs/tahoe-lafs/6378#tests/containers/0

This is just one instance out of probably hundreds this year.  The frequency of failure is going down because I am gradually replacing all of the port 0 binding in Tahoe-LAFS with a different, more reliable port allocation strategy.

glyph: Wow, there is a *lot* going on here.

I will grant that Tahoe does not appear to be doing the *straightforwardly* incorrect thing that is often an issue (i.e.: bind port zero, ask the port what port it is actually, close it, then tell somebody else to bind that port assuming it's still available).

But `SameProcessStreamEndpointAssigner` isn't exactly a one-liner either.

My understanding of the desired process here is that the port assigner will bind an ephemeral FD, hold on to it for as long as possible, install its own endpoint plugin that allows the actual socket itself to be re-used, and try to convince Twisted to call `adoptStreamPort` / `_fromListeningDescriptor`.

However, the traceback indicates that the exception is coming up out of `Port.startListening` when `startListening` calls `skt.bind()`.

I started looking into the semantics of calling `bind` repeatedly on the same socket when the socket in question is potentially represented by multiple FDs, but then I noticed that all of this is in a suite prefaced with `if self._preexistingSocket is None:`.  And `self._preexistingSocket` is unconditionally set in `Port._fromListeningDescriptor` i.e. `adoptStreamSocket`.  Which makes sense: it's heavily implied that `adoptStreamSocket` is expecting a socket already bound and listening; messing with the passed socket's parameters would ruin the whole deal.

So investigating as closely as I can here, this still looks to me to be more likely a bug in Tahoe's test suite.  (Although I don't envy you the task of finding it - these shenanigans look quite complex and from my relatively superficial reading, totally correct, so I don't know how this traceback happens.)

exarkun: > I will grant that Tahoe does not appear to be doing the straightforwardly incorrect thing that is often an issue (i.e.: bind port zero, ask the port what port it is actually, close it, then tell somebody else to bind that port assuming it's still available).

I probably could have provided a bit more guidance along with this link.  I'll try to do so now.  I think much of your analysis is correct but I think you missed the critical point that Tahoe's test suite employs multiple strategies for port number assignment.  `SameProcessStreamEndpointAssigner` is the new mechanism I introduced and have been porting the test suite to use to fix the problem with bind() failing.  But it does also do the thing you said you think it doesn't - use bind(0) to assign a random number, close the socket, and then try to bind that port again.  It also does some straight-up bind(0)-and-use-the-resulting-socket.  *That* is the specific behavior that I think is an application bug because Linux doesn't actually guarantee that it will work.

Complicating the analysis is that fact that many tests in the Tahoe-LAFS test suite bind 3 or 5 or 50 or 60 (!!!) ports and two or three different mechanisms for port assignment across them.  I haven't yet seen a build fail because of the `SameProcessStreamEndpointAssigner` logic.  I've only observed the traceback you noted - `Port.startListening` when `startListening` calls `skt.bind()`.

> I will grant that Tahoe does not appear to be doing the straightforwardly incorrect thing that is often an issue (i.e.: bind port zero, ask the port what port it is actually, close it, then tell somebody else to bind that port assuming it's still available).

It used to do this a lot more than it does now.  I think that it _mostly_ only does this on Windows now (because it's difficult to implement the alternatives given lack of support in Twisted).  But last week I think I found a couple more places where this actually happens.  My guess is that any recent failures of this type are due to these remaining bugs.

> So investigating as closely as I can here, this still looks to me to be more likely a bug in Tahoe's test suite.

We agree on this but maybe not on what the bug is. :)  After reading your analysis, I'll renew my assertion that the bug (in addition to many others not relevant to this discussion) in Tahoe-LAFS is the code which does a single bind(0) without exception handling - assuming it must succeed (by way of `listenTCP(0, ...)` or some endpoint API on top of that) - and then proceeds to use the resulting listening port.  All my observations support the conclusion that this will randomly fail (probably if another process on the system tries to bind an ephemeral port at around the same time).

Since this is a transient error condition, the best solution I can see is to re-try the bind(0) until it succeeds (or until you've tried so many times you think you've proven it's not a transient error condition).

PR #1127, opened 2019-04-09
Add capturedLogs context manager for testing.

wsanchez:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. [#9617](https://twistedmatrix.com/trac/ticket/9617#ticket)
* [x] I have created a newsfragment in


wsanchez:
Appveyor failure is in iocp, appears unrelated

src/twisted/logger/__init__.py
```diff
@@ -85,6 +85,9 @@ def oops(self, data):
     # From ._json
     "eventAsJSON", "eventFromJSON",
     "jsonFileLogObserver", "eventsFromJSONLogFile",
+
+    # From _capture
```
glyph: Consistency nitpick: all the other "From"s are `From ._whatever` and this one skips the dot :)

docs/core/howto/logger.rst
```diff
@@ -277,10 +277,33 @@ System-provided event keys include:
 Avoid mutable event keys
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Emitting applications should be cautious about inserting objects into event which may be mutated later.
+Emitting applications should be cautious about inserting objects into events which may be mutated later.
```
glyph, line 280: Nice catch.

docs/core/howto/logger.rst
```diff
@@ -277,10 +277,33 @@ System-provided event keys include:
...
+
+.. code-block:: python
+
+    from twisted.logger import Logger, LogCapture, LogLevel
+    from twisted.trial.unittest import TestCase
+
+    class SomeTests(TestCase):
+
+        def test_capture(self):
+            with LogCapture() as capture:
```
glyph: 2 minor nitpicks here:

Does this really need to be its own whole object?  My inclination would be to just use a `list`; less API surface to care about supporting.  If we really want `asText`, then  `"".join(map(eventAsText, events))` could go in a public utility function.

In the various places I've had to implement this over the years, I've typically spelled it:

```
with capturedLogs() as captured:
   ...
# captured is a list of all the events now
```
and I feel like this reads a little better in english.

src/twisted/logger/_capture.py
```diff
@@ -0,0 +1,56 @@
...
+
+from io import StringIO
+
+from attr import attrib, attrs
+
+from twisted.logger import globalLogPublisher, textFileLogObserver
+
+
+
+@attrs(frozen=True, slots=True)
```
glyph: One other advantage of having this be a `@contextmanager` function is that we don't need to care about frozenness or slotsness :).  Some metaprogramming tools can get a little annoying to use with things that are slotsified and don't have a `__dict__`.

glyph (approved):
I appreciate the doc updates here, and the test coverage is good.

If you disagree with my feedback, I'm not a hard "no" on this API shape.

So I leave it up to you how to respond to my feedback - address, or don't, to your satisfaction, and then land :).

docs/core/howto/logger.rst
wsanchez: @adiroiban Any thoughts about keeping `asText()`?  I'm inclined to agree with Glyph that it's not needed.  It can also be a function we provide later if needed.

wsanchez: @glyph: [this](https://github.com/burningmantech/ranger-deploy/blob/master/src/deploy/ext/logger.py#L41) is what I had originally written in my app, which is I think is more like what you have in mind.

src/twisted/logger/__init__.py
wsanchez: GAWD I can't believe I did that

wsanchez:
New version is very little code, so it also has that going for it.

wsanchez:
Arg I didn't update the HOWTO

wsanchez:
Got another rando-looking failure in [`buildbot/ubuntu16.04-py3.5`](https://buildbot.twistedmatrix.com/builders/ubuntu16.04-py3.5/builds/1805/steps/epoll/logs/stdio).

docs/core/howto/logger.rst
adiroiban: The simple list looks good.

I guess that later we can return an instance which inherits from list and has an extra `.asText()` method :)

Thanks for the change :)

src/twisted/logger/test/test_capture.py
```diff
@@ -0,0 +1,39 @@
...
+# See LICENSE for details.
+
+"""
+Test cases for L{twisted.logger._capture}.
+"""
+
+from twisted.logger import Logger, LogLevel
+from twisted.trial.unittest import TestCase
+
+from .._capture import capturedLogs
```
adiroiban, line 11: maybe the canonical import should be used here, as a way to exemplify how we expect it to be publicly available. 

so this can look like

```
from twisted.logger import capturedLogs
```

is there any advantage of using the relative imports?

adiroiban (approved):
Just minor comments. Looks good to me. Thanks!

src/twisted/logger/test/test_capture.py
wsanchez: I used relative imports all over the logger package, so I think this should stay for consistency unless we wanna redo that usage in the package.

The advantage was that that module started in a different place and moving it into Twisted was a lot easier to manage with fewer import rewrites.

I do see you point, though, in that we are testing the private namespace and not the public namespace… I wouldn't object to a separate ticket for redoing that in test cases, I guess.

adiroiban: I am not a big fan of separate tickets as most of the time nothing will happen.

Instead, I prefer drive-by fixes based on "The Boy Scout Rule" http://www.informit.com/articles/article.aspx?p=1235624&seqNum=6

glyph: It's tempting to try to include random drive-by fixes in nearby code while you're maintaining something unrelated.  I've certainly done it.  But I think it's an antipattern.

Why is it tempting?

It's tempting because the overhead of an individual PR is too high, and so we're hoping to get more stuff through given each "cycle" of the code-review machinery, to reduce that overhead.

Why is the overhead of the PR review cycle high?

1. Fixed overhead, such as CI runtimes.  (We should work to decrease these, of course.)
2. Reviewer bandwidth; there's limited bandwidth for reviewers to review things, and so they can't spend that much time reviewing.
3. Reviewer latency; reviewers don't necessarily have time to come back every day.

Attempting to fix unrelated stuff allows you to reduce (1) (more changes per CI run) while *increasing* the overhead associated with 2 and 3.

By making bigger changes, you take up more of the time available to reviewers to evaluate your changes in each sitting, which (a) decreases the number of changes they can review, and (b) makes it less likely that they'll be able to come back again promptly to re-review the changes, both because you took up more of their time at a go, and you required them to think harder and generally use up more mental energy.

Make every change as small and self-contained and obvious as possible, and a reviewer with a few idle minutes can go in and stamp it, and possibly review multiple tickets in a sitting.

adiroiban: thanks for the explanation.
Still, I was adding a comment to a new code which was written... why write it in one way and then change it in a separate PR, when you can write it from the start

I am not talking about rewriting all the imports.

glyph: I do think there's something to be said for switching from one consistent style to another consistent style, but - point taken ;).

PR #1128, opened 2019-04-12
[#9622] Cache AcceptableCiphers

hawkowl:
https://twistedmatrix.com/trac/ticket/9622#ticket


glyph:
My first question is "why did neither codecov nor coveralls put a passing coverage status on this PR"

src/twisted/internet/_sslverify.py
```diff
@@ -1740,29 +1747,20 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
...
-        """
-        self.fullName = fullName
-
-
-    def __repr__(self):
-        """
-        A runnable representation of the cipher.
-        """
-        return 'OpenSSLCipher({0!r})'.format(self.fullName)
+    @param fullName: The full name of the cipher. For example
```
glyph: In this context it needs to be an `@ivar` not a `@param`.

hawkowl: Ah, yes. Fixed.

hawkowl:
Coveralls is happy, the two missing lines are importing the backport on 2.7: https://coveralls.io/builds/22936426

src/twisted/test/test_sslverify.py
```diff
@@ -2760,11 +2760,11 @@ def raiser(_):
 
     def test_returnsListOfICiphers(self):
```
evilham: Maybe this should be renamed as it's now dealing with tuples?

src/twisted/test/test_sslverify.py
```diff
@@ -2798,18 +2798,18 @@ def test_selectReturnsOnlyFromAvailable(self):
             sslverify.OpenSSLCipher('A'),
             sslverify.OpenSSLCipher('B'),
         ])
-        self.assertEqual([sslverify.OpenSSLCipher('B')],
+        self.assertEqual((sslverify.OpenSSLCipher('B'),),
                          ac.selectCiphers([sslverify.OpenSSLCipher('B'),
                                            sslverify.OpenSSLCipher('C')]))
 
 
     def test_fromOpenSSLCipherStringExpandsToListOfCiphers(self):
```
evilham: Same as above

src/twisted/internet/_sslverify.py
```diff
@@ -1779,7 +1777,7 @@ def _expandCipherString(cipherString, method, options):
 
     @return: The effective list of explicit ciphers that results from the
         arguments on the current platform.
-    @rtype: L{list} of L{ICipher}
+    @rtype: L{tuple} of L{ICipher}
```
evilham, line 1780: May be a silly question, but it's worth asking: is changing lists for tuples necessary?

If so, changing the docstrings is also needed.

evilham (commented):
It looks like we change an interface (list / tuple), unsure if that's critical.
In any case it should be made at least consistent in other places.

src/twisted/internet/_sslverify.py
hawkowl: Yes, because tuples are hashable and lists are not, and we need hashable types to make a low-overhead cache.

I'll update the docstrings.

twm:
I removed the backport that was bothering mypy, since we have dropped support for Python 2.7.

src/twisted/internet/_sslverify.py
```diff
@@ -1747,32 +1747,23 @@ def _verifyCallback(conn, cert, errno, depth, preverify_ok):
...
+    """
+    fullName = attr.ib()
 
 
 
+@lru_cache(maxsize=32)
 def _expandCipherString(cipherString, method, options):
     """
-    Expand C{cipherString} according to C{method} and C{options} to a list
+    Expand C{cipherString} according to C{method} and C{options} to a tuple
```
glyph, line 1766: As we annotate with types, this should (at some future point) probably be a `Sequence[ICipher]`, not a `Tuple`.

glyph (approved):
Very technically, this is an incompatible change, because of the type change on `availableCiphers`, but I'm going to make the judgement call that this is not something that you could realistically have depended on this in a meaningful way before.  In the future we will protect stuff like this with e.g. `Sequence`.

I can't wait to see how the perf numbers move when this lands!

src/twisted/internet/_sslverify.py
twm: `Sequence` doesn't imply `Hashable`, though. That means... mixing the two? Or a custom Protocol? I am not up to speed on this MyPy stuff.

PR #1129, opened 2019-04-13
[9625] improve cloud code coverage configuration

glyph:
- separate coverage preparation and publishing
- add coverage publishing to circle macOS builders
- add publishing to coveralls, which might actually be the good one now
- keep publishing to codecov so we can compare


glyph:
I'm going to merge this unilaterally because it is coupled to stateful configuration changes made in various web services, and I believe other PRs are going to start getting bogus build statuses that report incorrect things about coverage if I don't land it or roll those back.  Also, it's not a code change, it's a config change to how CI works.

I am happy to take feedback or roll this back if anyone disagrees with these changes.

PR #1130, opened 2019-04-25
Factor out pluggable resolver reactor implementation

exarkun:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9632

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -449,6 +449,7 @@ class _ThreePhaseEvent(object):
                     log.err()
 
 
+
 @implementer(IReactorPluggableNameResolver, IReactorPluggableResolver)
 class PluggableResolverMixin(object):
     """
@@ -626,7 +627,7 @@ class ReactorBase(PluggableResolverMixin):
 
 
     # IReactorCore
-    def resolve(self, name, timeout = (1, 3, 11, 45)):
+    def resolve(self, name, timeout=(1, 3, 11, 45)):
         """Return a Deferred that will resolve a hostname.
         """
         if not name:
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -476,17 +476,6 @@ class PluggableResolverMixin(object):
         return oldResolver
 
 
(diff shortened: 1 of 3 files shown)


src/twisted/internet/base.py
```diff
@@ -449,10 +449,60 @@ def _continueFiring(self, ignored):
                     log.err()
 
 
+@implementer(IReactorPluggableNameResolver, IReactorPluggableResolver)
+class PluggableResolverMixin(object):
```
hawkowl, line 454: Question: should this be public?

hawkowl: (I'm assuming the answer is "yes, since this can be used in third party reactors")

src/twisted/internet/test/test_resolver.py
```diff
@@ -538,7 +541,8 @@ def test_portNumber(self):
         self.assertEqual(receiver._ended, True)
 
 
-class JustEnoughReactor(ReactorBase, object):
+
+class JustEnoughReactor(ReactorBase):
     """
```
meejah, line 547: Not knowing why there is an `object` in the "old" inheritence explicitly, is this removed because `ReactorBase` now inherits from the new class, which inherits from `object` -- and thus there's at least one `object` in the tree, so python2 should be happy...?

meejah (approved):
LGTM :+1:

src/twisted/internet/base.py
exarkun: Yes.  Making it available for people to use was the point.

src/twisted/internet/test/test_resolver.py
exarkun: ReactorBase already inherited from object when JustEnoughReactor was first added so I'm not sure.  I think it was probably a mistake, but a harmless one (because having object in your hierarchy multiple times is the normal case).  I probably didn't need to remove it here but I was trying to remove `JustEnoughReactor` entirely and when that didn't work I put it back, but apparently not all of it. :)

PR #1133, opened 2019-05-04
Added tests for TLS1.3.

EdgeKing810:
This commit was done for the IETF 104 hackathon under CyberStorm.mu.

## Contributor Checklist:

* [x] There is an associated ticket in Trac here: https://twistedmatrix.com/trac/ticket/9635


src/twisted/test/test_sslverify.py
```diff
@@ -1229,43 +1229,61 @@ def test_tlsProtocolsTLSv1Point2Only(self):
         self.assertEqual(options, ctx._options & options)
 
 
+    def test_tlsProtocolsTLSv1Point3Only(self):
+        """
+        When calling L{sslverify.OpenSSLCertificateOptions} with
+        C{insecurelyLowerMinimumTo} and C{lowerMaximumSecurityTo} set to v1.3,
+        it will exclude all others.
```
glyph, line 1236: This test docstring seems wrong, in that it should *not* exclude all others simply because the minimum has been *lowered*.  There's no higher maximum security than TLSv1.3, so lowering the maximum shouldn't do anything either.  So this is interestingly demonstrating a deficit in the implementation!

The appropriate test here would be to use `raiseMinimumTo`, which will have the same effect - we should figure out how to fix this one so it behaves differently.

src/twisted/test/test_sslverify.py
```diff
@@ -1049,7 +1049,7 @@ def test_tlsProtocolsNoMethodWithAtLeast(self):
                 privateKey=self.sKey,
                 certificate=self.sCert,
                 method=SSL.SSLv23_METHOD,
-                raiseMinimumTo=sslverify.TLSVersion.TLSv1_2,
+                raiseMinimumTo=sslverify.TLSVersion.TLSv1_3,
```
glyph, line 1052: All these tests which change the flags but don't change any assertions probably should not be changing; their behavior should remain the same.

src/twisted/test/test_sslverify.py
```diff
@@ -1229,43 +1229,61 @@ def test_tlsProtocolsTLSv1Point2Only(self):
...
             certificate=self.sCert,
             insecurelyLowerMinimumTo=sslverify.TLSVersion.TLSv1_0,
-            lowerMaximumSecurityTo=sslverify.TLSVersion.TLSv1_2,
+            lowerMaximumSecurityTo=sslverify.TLSVersion.TLSv1_3,
         )
         opts._contextFactory = FakeContext
         ctx = opts.getContext()
         options = (SSL.OP_NO_SSLv2 | SSL.OP_NO_COMPRESSION |
-                   SSL.OP_CIPHER_SERVER_PREFERENCE | SSL.OP_NO_SSLv3 |
-                   opts._OP_NO_TLSv1_3)
```
glyph, line 1249: This actually *removes* test coverage; we want to ensure that if you intentionally disable TLSv1.3 by lowering the security *below* 1.3, that OP_NO_TLSv1_3 is indeed passed.

glyph (changes requested):
Hi @EdgeKing810 !  Thanks for your contribution to Twisted.

I don't think that these tests should be added as such, because they actually remove what little direct coverage of TLS 1.3 support we have; to wit, the use of OP_NO_TLSv1_3 when it is intentionally disabled.

Also, there are no uncovered lines of *code* in the system under test; looking at the coverage report over at https://coveralls.io/builds/25539285 I see a coverage increase only in the test module.

I'm definitely happy to have improved / better TLS v1.3 tests, but I don't think that this diff makes a noticeable improvement.  Please feel free to re-submit for review (reattaching the `review` keyword in trac) if you disagree though!

glyph:
As I've closed the ticket, I'm also going to close this PR.  Again, feel free to reopen if you disagree / think there's more to be done here.

PR #1134, opened 2019-05-05
[#9512] Initial configuration for black

hawkowl:
https://twistedmatrix.com/trac/ticket/9512


docs/core/development/policy/coding-standard.rst
```diff
@@ -76,18 +76,19 @@ Indentation is 4 spaces per indent.
 Tabs are not allowed.
 It is preferred that every block appears on a new line, so that control structure indentation is always visible.
 
-Lines are flowed at 79 columns.
```
glyph, line 79: 1. Can we just remove every aspect of this document covered by Black and just say "code must be formatted by Black"?
2. Given that this is one of the very few options Black supports, I'd really prefer to keep the 79-column line length limitation.  For one thing, it matches the docstring width, and for another, 79 columns is of a size where someone with average eyesight can comfortably fit 2 columns on a 15" laptop screen; 89 is too wide to practically make use of that space.  (If it weren't configurable I don't feel strongly enough about this to want to patch Black or anything, but it seems to be aware that 89 is a peculiar choice.)

docs/core/development/policy/coding-standard.rst
```diff
@@ -229,6 +230,8 @@ Docstrings should always be used to describe the purpose of methods, functions,
 Moreover, all methods, functions, classes, and modules must have docstrings.
 In addition to documenting the purpose of the object, the docstring must document all of parameters or attributes of the object.
 
+Docstrings must be reflowed at 79 characters.
```
glyph, line 215: If someone could turn https://github.com/glyph/python-docstring-mode/blob/master/docstring_wrap.py into a tool that ran over a codebase's docstrings rather than just an editor plugin shim helper gizmo, we could possibly Black-ify our epytext as well...

docs/core/development/policy/coding-standard.rst
glyph: (The 79-column thing is not a blocker, if this is something you disagree with I'm fine being overridden here.)

pyproject.toml
```diff
@@ -50,3 +50,6 @@
         directory = "misc"
         name = "Misc"
         showcontent = false
+
+[tool.black]
+skip-string-normalization = true
```
wsanchez: why?

tox.ini
```diff
@@ -142,9 +140,13 @@ basepython=python2.7
...
 
-# TwistedChecker needs to be 3.5
 [testenv:lint]
-basepython=python3.5
+skip_install=True
+basepython=python3.6
+
+[testenv:black]
+skip_install=True
+basepython=python3.6
```
wsanchez: FWIW I found it useful to add a test rnv `black-reformat` to that it's easy for a developer to reformat code to comply with `black`.  Example [here](https://github.com/burningmantech/ranger-deploy/blob/master/tox.ini#L71).

wsanchez:
Are we going to toss out the 3-newline-before-class thing when we start using `black` or is there a way to keep that?

pyproject.toml
wsanchez: If this is because `u""` is getting changed to `""`, I think that passing `--target-version=py27` might be a more appropriate way to get there.

docs/core/development/policy/coding-standard.rst
twm: Perhaps file a ticket against Black for this?

twm (approved):
All of the blocker comments here seem to have been addressed. Let's end the petty tyranny of twistedchecker forever. :tada:

docs/core/development/policy/coding-standard.rst
glyph: This docstring wrapper makes a number of assumptions about the format of the docstrings it's wrapping, most of which are going to be wrong for a lot of projects.  I don't think there's any way to realistically make it universal enough for Black to consider incorporating it.

glyph:
> Are we going to toss out the 3-newline-before-class thing when we start using `black` or is there a way to keep that?

We're tossing it; the appeal of adopting Black in the first place is that we will diverge from common practice in fewer situations.

docs/core/development/policy/coding-standard.rst
```diff
@@ -434,28 +419,6 @@ For example scripts you expect a Twisted user to run from the command-line, add
     #!/usr/bin/env python
 
 
-Standard Library Extension Modules
```
glyph, line 437: Can we do this doc update in a different PR?  It doesn't have anything to do with Black.

docs/core/development/policy/coding-standard.rst
```diff
@@ -477,15 +440,6 @@ For example, a Service subclass for Forums might be named ``twisted.forum.servic
 Since neither of these modules are volatile *(see above)* the classes may be imported directly into the user's namespace and not cause confusion.
 
 
-New-style Classes
```
glyph, line 480: This doesn't seem to have anything to do with Black either.

docs/core/development/policy/coding-standard.rst
```diff
@@ -618,17 +572,6 @@ An attribute (or function, method or class) should be considered private when on
 - The attribute is part of a known-to-be-sub-optimal interface and will certainly be removed in a future release.
 
 
-Python 3
```
glyph, line 621: This also doesn't seem to have anything to do with Black

docs/core/development/policy/coding-standard.rst
```diff
@@ -644,10 +587,10 @@ All SQL keywords should be in upper case.
 C Code
 ------
 
-C code must be optional, and work across multiple platforms (MSVC++9/10/14 for Pythons 2.7 and 3.5+ on Windows, as well as recent GCCs and Clangs for Linux, macOS, and FreeBSD).
+C code must be optional, and work across multiple platforms (MSVC++14 for Python 3.5+ on Windows, as well as recent GCCs and Clangs for Linux, macOS, and FreeBSD).
```
glyph, line 590: Not related to this change

docs/core/development/policy/coding-standard.rst
```diff
@@ -644,10 +587,10 @@ All SQL keywords should be in upper case.
...
 
 C code should be kept in external bindings packages which Twisted depends on.
-If creating new C extension modules, using `cffi <https://cffi.readthedocs.io/en/latest/>`_ is highly encouraged, as it will perform well on PyPy and CPython, and be easier to use on Python 2 and 3.
+If creating new C extension modules, using `cffi <https://cffi.readthedocs.io/en/latest/>`_ is highly encouraged, as it will perform well on PyPy and CPython.
```
glyph, line 593: Not related to this change

docs/core/development/policy/coding-standard.rst
```diff
@@ -178,17 +164,17 @@ While this applies to all packages within Twisted, one ``twisted.python`` deserv
 Strings
 -------
 
-All strings in Twisted which are not interfacing directly with Python (e.g. ``sys.path`` contents, module names, and anything which returns ``str`` on both Python 2 and 3)  should be marked explicitly as "bytestrings" or "text/Unicode strings".
-This is done by using the ``b`` (for bytestrings) or ``u`` (for Unicode strings) prefixes when using string literals.
-String literals not marked with this are "native/bare strings", and have a different meaning on Python 2 (where a bare string is a bytestring) and Python 3 (where a bare string is a Unicode string).
+All literal strings in Twisted which contain binary data must be prefixed with ``b`` (for bytestring).
```
glyph, line 167: This doesn't seem related to this change, but also, do we need to even say this any more?  This seems to be just a basic Python syntax fact now.

docs/core/development/policy/coding-standard.rst
mthuurne: Are the assumptions specific to Twisted or specific to particular docstring formats? In the latter case, being able to specify the docstring format to Black and/or supporting the `__docformat__` constant could be used to turn the docstring reformatting on/off.

glyph (changes requested):
I have three blockers here that I'd like to see resolved before landing this:

1. Let's please move out all the unrelated documentation changes so they're not blocked behind my *other* two blockers, and...
2. I actively *do not* want anyone running `black` over files in the Twisted codebase until we have a documented solution for what to do with existing PRs.  I want to make sure that when we run it for the first time, we [do not ruin Git Blame forever](https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame), which means we need a dedicated point in history where the formatting occurs, not a scattershot format-as-you-go approach.
3. We should also have a tooling to run over an existing PR to ensure that everyone with an open branch (of which there are presently **1853** on `origin` alone) doesn't need to manually reconstruct all of their work.  Over on Trac, Hawkowl says:

> I also believe that if you merge in this merge commit, blackify your PR, then merge trunk, git should know what to do.

If by "blackify your PR, then merge trunk" you mean "add a commit that runs `black`", when I try that I see big irrelevant conflicts, so at least this isn't entirely obvious.  But perhaps there's some difference in configuration?  Or maybe you meant something different about using history rewriting to make each change have `black` run over it on its own?  Can someone write down the specific commands to run, then verify that they work across some existing PRs?

> If we squash merge PRs, then we'll also get a nice history for those PRs.

We don't rewrite history currently and squash merges typically make this sort of problem worse, not better.

glyph:
(I should note that the *best* option here is not to document how someone can fix up their own work, but just to write a script that pushes a Black-fixup commit to every open PR; as long as we hit *most* of the open branches, it's fine; having to fix up one or two PRs or branches manually out of ~thousands is not too high a price to pay.)

mthuurne:
I'm attempting to add type annotations by automated replacements on the source tree, but the code style checker is a bottleneck there: even if the automated replacement doesn't break the rules, it can be rejected by the style checker if the replacement was close to preexisting non-compliant code. To correct these manually is too time consuming for some replacements, so I have to shelve those ideas until Black adoption has been merged.

Of course I'm a bit biased because of the above, but I don't think breaking PRs should be a blocker:
- there are currently 65 open PRs
- 40 open PRs are over a year old and will likely require rework even without considering Black
- the addition of type annotations will also cause PRs to require rework:
  - PRs will need to pass mypy checks before they can get merged
  - sometime soon we should start to require type annotations on new code, in my opinion

So all in all, there isn't a huge number of open PRs, but the ones that we do have will likely need manual updates anyway.

PR #1136, opened 2019-05-07
Fix wrong return value in addComponent docstring

Hnasar:
Since commit 0961531512 back in 2002, the docstring has incorrectly
stated that this function returns something.

Ticket https://twistedmatrix.com/trac/ticket/9637


glyph (approved):
Looks correct to me ✅

PR #1137, opened 2019-05-08
Move proto helpers 6435

eevelweezel:
Move contents of twisted.test.proto_helpers to twisted.protocols.utils.  Add deprecation warnings to objects in twisted.test.proto_helpers and unit tests for warnings, per the Compatibility Policy. 

Trac ticket: https://twistedmatrix.com/trac/ticket/6435


eevelweezel:
Please note I'll be travelling for the next 24 hours.  I'm looking into the test failure, but may not be able to respond immediately.

glyph:
Hey @eevelweezel ! Still haven't had time for a full review but I wanted to note my comment on https://twistedmatrix.com/trac/ticket/6435#comment:6 about module naming, in case Trac emails were not getting delivered for any of the usual reasons :)

src/twisted/test/proto_helpers.py
```diff
@@ -74,10 +81,20 @@ def connectionLost(self, reason):
             d.callback(None)
 
 
+
+deprecatedModuleAttribute(
+    Version("Twisted", 19, 3, 0),
```
glyph: As per our [coding standard](https://twisted.readthedocs.io/en/twisted-19.2.0/core/development/policy/coding-standard.html?highlight=coding%20standard#versioning) and [the docs for Incremental that it links to](https://github.com/twisted/incremental#updating), the right way to spell these versions is `Version("Twisted", "NEXT", 0, 0)`, so it'll be picked up in the next release cycle.  According to [our current versioning scheme](https://calver.org), 19.3 would mean "released in March of 2019".

src/twisted/test/proto_helpers.py
```diff
@@ -74,10 +81,20 @@ def connectionLost(self, reason):
...
+    "Please use twisted.internet.testing.AccumulatingProtocol instead",
```
glyph: I think it would make more sense to deprecate `proto_helpers` in its entirety, rather than every individual name?  You can put the `deprecatedModuleAttribute` at the package level (in `__init__.py`) instead, so you have fewer of these to spell out.

In either case, you should remove all the duplicated code from this module, and re-export the names via `__all__`, so that we don't have them all copied twice until their removal, and we don't get divergent behavior.

src/twisted/test/test_proto_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+
+from twisted.trial import unittest
+
+
+
+class DeprecationTests(unittest.TestCase):
+    """
+    Deprecations in L{twisted.test.proto_helpers}.
+    """
+    def test_AccumlatingProtocol(self):
```
glyph: nitpick: the coding-standard-compliant casing here would be `test_accumlatingProtocol`.

src/twisted/test/test_proto_helpers.py
```diff
@@ -0,0 +1,238 @@
+# Copyright (c) Twisted Matrix Laboratories.
```
glyph: nitpick on file naming: my personal preference here would be to move all the tests here into `src/twisted/internet/test/test_testing.py` so there's only one place to look.

src/twisted/test/test_proto_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    Deprecations in L{twisted.test.proto_helpers}.
+    """
+    def test_AccumlatingProtocol(self):
+        """
+        L{proto_helpers.AccumlatingProtocol} is deprecated and
+        generates a warning.
+        """
+        from twisted.test.proto_helpers import AccumulatingProtocol
+        AccumulatingProtocol  # shush
```
glyph: Love the comment 😉 .

src/twisted/test/test_proto_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+
+        warnings = self.flushWarnings(
```
glyph: These tests all seem substantially duplicated.  If you're going to stick with each object being its own deprecated attribute, can you refactor these to call a single helper so we don't have so much duplication?

src/twisted/test/test_proto_helpers.py
```diff
@@ -0,0 +1,238 @@
...
+            [self.test_AccumlatingProtocol])
+        self.assertEqual(DeprecationWarning, warnings[0]['category'])
+        self.assertEqual(
+            "twisted.test.proto_helpers.AccumulatingProtocol was "
+            "deprecated in Twisted 19.3.0: Please use "
+            "twisted.internet.testing.AccumulatingProtocol instead",
+            warnings[0]['message'])
+        self.assertEqual(1, len(warnings))
```
glyph: Also, if you fix the duplications, you could put an `assertIs` in to check that the deprecated alias points to the new, non-deprecated public name.

glyph (changes requested):
Thanks for this contribution!  Welcome to the Twisted community! :)

It's pretty close to mergeability but some of the changes discussed inline (such as the way versions are expressed, and eliminating all the duplicated code from the move) are important to address before landing.  If you can address that feedback and re-submit for review (by re-adding the 'review' keyword in Trac) we can probably get it landed without further ado.

Your thoroughness and attention to detail is really appreciated here!

src/twisted/test/proto_helpers.py
eevelweezel: OK, fixed.

src/twisted/test/proto_helpers.py
eevelweezel: Alright, moved the deprecation warning code to twisted.test.__init__.

src/twisted/test/test_proto_helpers.py
eevelweezel: Updated, as part of moving these tests to twisted.internet.test.test_testing.

src/twisted/test/test_proto_helpers.py
eevelweezel: Removed twisted.test.test_proto_helpers, and moved all deprecation tests to twisted.internet.test.test_testing.

src/twisted/test/test_proto_helpers.py
eevelweezel: Refactored these and moved them to twisted.internet.test.test_testing.DeprecationTests.

src/twisted/test/test_proto_helpers.py
eevelweezel: Done.

src/twisted/internet/test/test_testing.py
```diff
@@ -0,0 +1,502 @@
...
+    """
+    Deprecations in L{twisted.test.proto_helpers}.
+    """
+    def helper(self, test, obj):
+        new_path = 'twisted.internet.testing.{}'.format(obj.__name__)
+        warnings = self.flushWarnings(
+            [test])
+        self.assertEqual(DeprecationWarning, warnings[0]['category'])
+        self.assertEqual(1, len(warnings))
+        self.assertIn(new_path, warnings[0]['message'])
```
twm, line 429: This should also assert that the object and the two locations are identical. For example:

```python
from twisted.python.reflect import namedAny

self.assertIs(obj, namedAny(new_path))
```

src/twisted/test/proto_helpers.py
```diff
@@ -74,10 +93,12 @@ def connectionLost(self, reason):
             d.callback(None)
 
 
+
 class LineSendingProtocol(basic.LineReceiver):
```
twm: This class shouldn't still be defined here. It should be imported from `twisted.internet.testing`. The same goes for all the other stuff that was moved.

src/twisted/test/test_stringtransport.py
```diff
@@ -17,6 +17,7 @@
 from twisted.internet.protocol import ClientFactory, Factory
 
 
+
 class StringTransportTests(TestCase):
```
twm, line 34: Likewise, these tests were moved to `twisted.internet.test.test_testing`, no? So they aren't needed here anymore.

twm (changes requested):
Welcome to the community, @eevelweezel! 

I have a little more feedback, the gist of which is that everything should be _moved_ rather than _copied_. If there are two copies they can diverge, which isn't really what we want here &mdash; by promoting these helpers to a more prominent location we hope to make it more clear that they are generally useful (well, some of them are), which is likely to trigger improvements that should also feed back into Twisted's own test suite. Eventually we should update Twisted's own imports to point at `twisted.internet.testing` rather than the old location (if only to avoid deprecation warnings!).

Once this is in I'll file some follow-up tickets to get `twisted.internet.testing` into better shape, as I noticed some issues when looking through the PR, and there is [a lot of missing test coverage](https://coveralls.io/builds/23822744/source?filename=src/twisted/internet/testing.py#L355). You don't need to worry about that in this PR, though. Small bites.

(In case you're not familiar with Coveralls &mdash; you can find that report by clicking the coveralls/coveralls "Details" link in the build checks: 
image )

Thank you for this contribution!

eevelweezel:
I'll be traveling for the next week, but will address these issues when I return.

src/twisted/internet/test/test_testing.py
eevelweezel: Ah, makes sense.  Tests are updated.

src/twisted/test/proto_helpers.py
eevelweezel: twisted.test.proto_helpers is now importing the contents of twisted.internet.testing.

src/twisted/test/test_stringtransport.py
eevelweezel: Removed twisted.test.test_stringtransport.

PR #1139, opened 2019-05-12
isort + isort changes

hawkowl:
See https://github.com/twisted/twisted/pull/1132


hawkowl:
not needed, just a demonstration

PR #1141, opened 2019-05-17
[#9476] fix backoffPolicy OverflowError on attempts higher 1750

jandelgado:
This fixes https://twistedmatrix.com/trac/ticket/9476, where the backoffPolicy raised an OverflowError for connection attempts > 1750 when using the default settings.

## Contributor Checklist:


src/twisted/test/test_application.py
```diff
@@ -967,3 +968,27 @@ def getReactorTypes():
         self.assertRaises(SystemExit, options.opt_help_reactors)
         message = options.messageOutput.getvalue()
         self.assertNotIn("reactors not available", message)
+
+
+
+class BackoffPolicyTests(unittest.TestCase):
+    def test_calculates_correct_values(self):
```
twm, line 976: This test needs a docstring per the coding standard: https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html The same goes to the other two.

twm (changes requested):
First, thanks for this fix! I use `backoffPolicy()` pretty extensively and would have eventually been bitten by this bug.

The changes themselves look good. Thank you for writing tests for this function &mdash; I'm surprised that it slipped in without any. There is just one procedural nit blocking merge: all classes and methods [need docstrings](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html). The comments you left are pretty close to what's required.

Thanks again!

src/twisted/test/test_application.py
jandelgado: Added docstrings to the tests.

twm (approved):
Thank you @jandelgado! There is build failure but it is unrelated to your changes so I'm going to correct a nit in the docstring formatting to get it to run again (we just switched to Azure and I don't seem to be able to re-run it in the new system).

(The nit is that the """ should be on their own lines. I guess our linting tools don't catch this. :disappointed: I'd file a bug for this but there is some ongoing work to automate docstring formatting so the ticket would be obsoleted in short order.)

Thanks so much for contributing to Twisted and bearing with our process!

jandelgado:
Excellent! Thanks for the quick merge.

PR #1143, opened 2019-05-19
[9628] Fix trial3 not knowing the difference between ImportErrors and modules not existing

hawkowl:
https://twistedmatrix.com/trac/ticket/9628


src/twisted/trial/runner.py
```diff
@@ -724,8 +724,27 @@ def findByName(self, _name, recurse=False):
...
+                # Get the filename that the ImportError originated in.
+                filenameWhereHappened = tb.tb_frame.f_code.co_filename
+
+                # If it originated in the reflect file, then it's because it
+                # doesn't exist. If it originates elsewhere, it's because an
+                # ImportError happened in a module that does exist.
+                if filenameWhereHappened != reflect.__file__:
+                    raise
+
                 if remaining == "":
```
markrwilliams, line 744: this can never be true. `remaining` is assigned in two places: [to `None` in line 710](https://github.com/twisted/twisted/pull/1143/files#diff-c8eebbb513e06cd3494e0e38f6536198R710) and [to `remainingName` (the class, method, function etc. part of the fqpn) on line 723](https://github.com/twisted/twisted/pull/1143/files#diff-c8eebbb513e06cd3494e0e38f6536198R723).  in the second case it will be a string, but: a) the `break` prevents this block from running after that assignment, and b) `remaining` can only be `""` with an fqpn that ends in a dot, which is invalid.

the `ModuleNotFound` exception caught by the test comes from [`namedAny` on line 752](https://github.com/twisted/twisted/pull/1143/files#diff-c8eebbb513e06cd3494e0e38f6536198R752).  this should be changed to `if remaining is None:` for it to have the intended effect.

markrwilliams (approved):
@hawkowl explained why `Py3Loader` has to exist.  per this comment:

https://github.com/twisted/twisted/blob/e0b696f17260fc17cd5a9b3e7bc3ad1bc0ffdd54/src/twisted/trial/runner.py#L757-L759

Unbound methods are gone in Python 3, so the object returned by `namedAny('twisted.trial.test.test_loader.FinderPy3Tests.test_findNonPackage')` won't have a reference to its class.  `loadAnything` needs that to filter down to subclasses of `TestCase`.

That means `Py3TestLoader` can't just used `namedAny` the way `TestLoader` does, but instead has to reimplement that logic here.

Now, the stack walking logic here will be slow on PyPy, but trial is already slow on PyPy; we also have other instances of stack walking (see [_inlineCallbacks](https://github.com/twisted/twisted/blob/5b203b267f9869f2bd6e3ed17dcdaafa1fa227d3/src/twisted/internet/defer.py#L1423-L1462)). 

@hawkowl suggests reimplementing this with [`importlib.util.spec_from_file_location`](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location) which seems good, but enough work that it should land in a separate PR.  i'm approving and landing this because it fixes a bug ahead of a new release.

thanks @hawkowl !

PR #1144, opened 2019-05-20
[#9405] Deprecate twisted.news on Python 3.

hawkowl:
https://twistedmatrix.com/trac/ticket/9405


glyph (approved):
To comply with the [compatibility policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html) as written ("Cause imports or usage of the class/function/method to emit a DeprecationWarning"), we need the deprecation warning to be in place so that people who just *run* Twisted and don't read the changelog will at least have a *chance* of noticing the deprecation.

The best way to address the test failures which provoked its removal from the branch would be to say that `twisted.news` is not deprecated "on python 3", it's just deprecated.  If our python 2 support were to linger for some reason, it is still unmaintained and should still be removed.

The news fragment should also be updated to remove the reference to python 3.

Given that the tests are already written and just need to be resurrected and tweaked a tiny bit, feel free to land when you've addressed this minor feedback.

glyph:
@hawkowl  Where are we on this?

wsanchez:
Warning this steps on https://github.com/twisted/twisted/pull/915

PR #1148, opened 2019-05-30
Fix circular import in twisted.trial.reporter

cjwatson:
I didn't add a specific test for the circular import issue, partly because I couldn't see any existing ones for Twisted's existing circular import workarounds (do let me know if I missed something), and partly because it would be rather cumbersome to arrange that the test performs the import early enough to avoid running into the workaround described in https://twistedmatrix.com/trac/ticket/8267.


twm (approved):
Hi @cjwatson!

These changes look good. Ordinarily Twisted process calls for tests, but I can't find any prior art for how to test for circular imports, so I manually verified that this is a problem in Python 2.7 (albeit interestingly not in Python 3.7) and that importing succeeds on this branch. I think that will do for now, though it unfortunately leaves us vulnerable to future regressions.

Thanks for the contribution!

twm:
I am going to merge despite the Circle CI documentation failure because that build is failing on every PR I merge forward &mdash; it appears completely unconnected to these changes.

PR #1149, opened 2019-05-31
#9645 Implement FTP extensions for IPv6

### LINKED ISSUE #9645: twisted.protocols.ftp.FTP should support IPv6 extensions
|[cjwatson's avatar](https://github.com/cjwatson)| @cjwatson reported|
|-|-|
|Trac ID|trac#9645|
|Type|enhancement|
|Created|2019-05-31 13:52:32Z|

RFC 2428 defines FTP extensions needed to make it work over IPv6, which Twisted doesn't yet support.  This [blocks ppa.launchpad.net being available over IPv6](https://bugs.launchpad.net/txpkgupload/+bug/1807662) (well, we'll probably monkey-patch support rather than actually waiting for a Twisted release, but we still need to do most of the same work).

This is related to [#9085](https://github.com/twisted/twisted/issues/9085), but that's the client side whereas this is the server side.

```
trac-id__9645 9645
type__enhancement enhancement
reporter__cjwatson cjwatson
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__ftp ftp
keywords__None None
time__1559310752044062 1559310752044062
changetime__1570997590451628 1570997590451628
version__None None
owner__cjwatson cjwatson

```
</details>

cjwatson:
Fixes #9645

These are defined in RFC 2428.

The facilities in the RFC for requesting the use of a specific network
protocol don't fit very well into Twisted's view of the world, but I've
at least arranged for the command arguments to be validated
appropriately for the common cases of IPv4 and IPv6.

Tests for IPv6 support are only partial, because FTPClient doesn't yet
support this (see https://twistedmatrix.com/trac/ticket/9085).


cjwatson:
There are some remaining lint errors, but they're essentially all due to the justification of "=" characters in constant assignments near the top of the file, and mostly pre-existing except where I made my changes fit the surrounding code.  I think it would make more sense to waive these rather than clutter this PR with them, but if reviewers require it then I can do the reformatting.

src/twisted/protocols/ftp.py
```diff
@@ -737,6 +766,8 @@ class FTP(basic.LineReceiver, policies.TimeoutMixin, object):
     dtpPort = None
     dtpInstance = None
     binary = True
+    epsvAll = False
+    supportedNetworkProtocols = (_AFNUM_IP, _AFNUM_IP6)
```
adiroiban: Do we need these members as public members?

src/twisted/protocols/ftp.py
```diff
@@ -981,6 +1108,51 @@ def connFailed(err):
...
+        As described by U{RFC 2428 section
+        2<https://tools.ietf.org/html/rfc2428#section-2>}::
+
+            The EPRT command allows for the specification of an extended
+            address for the data connection.  The extended address MUST
+            consist of the network protocol as well as the network and
+            transport addresses.
+        """
+        if self.epsvAll:
+            return defer.fail(BadCmdSequenceError(
```
adiroiban: I think that is ok to return a failure here...

What I did, is create a standard reponse code like 

```
RESPONSE[EPSV_ALL_ACTIVE] = '503 EPSV ALL was enabled.'
```

and then just do

```
        if self._epsvAllActive:
            return (EPSV_ALL_ACTIVE,)
```

src/twisted/protocols/ftp.py
```diff
@@ -945,6 +978,29 @@ def ftp_PASV(self):
...
+                # they're encoded as IPv4 in the response.
+                host = str(address.ipv4_mapped)
+            else:
+                # There's no standard defining the behaviour of PASV with
+                # IPv6, so just claim it as unimplemented.  (Some servers
+                # return something like '0,0,0,0' in the host part of the
+                # response in order that at least clients that ignore the
+                # host part can work, and if it becomes necessary then we
+                # could do that too.)
+                return defer.fail(CmdNotImplementedError('PASV'))
```
adiroiban: This is ok.

Maybe the error can be:  `PASV available only for IPv4` ... or something more verbose for the client :)

Adding support for mapped IPV4 is nice, but I don't know if it is worth adding all this extra handling here.

What I have found for many of my deployments is that with FTP servers behind the load balances, people will want to hardcode the Load Balancer IP in the PASV response... so that old clients will continue to work.

src/twisted/protocols/ftp.py
```diff
@@ -706,6 +730,11 @@ class FTP(basic.LineReceiver, policies.TimeoutMixin, object):
     @ivar listenFactory: A callable with the signature of
         L{twisted.internet.interfaces.IReactorTCP.listenTCP} which will be used
         to create Ports for passive connections (mainly for testing).
+    @ivar epsvAll: If true, "EPSV ALL" was received from the client, requiring
```
glyph: Thanks so much for proactively adding docs for this!

src/twisted/protocols/ftp.py
glyph: Appreciate that you made this a tuple so that we wouldn't get cruft accumulating in it by accident :)

src/twisted/protocols/ftp.py
```diff
@@ -864,14 +896,15 @@ def call_ftp_command(command):
                 return BAD_CMD_SEQ, "RNTO required after RNFR"
 
 
-    def getDTPPort(self, factory):
+    def getDTPPort(self, factory, interface=''):
```
glyph: Can you add a `@param` explaining its usage, since this is technically modifying a public API?

Also, `''` is an ambiguous default value; is this supposed to be bytes or text?

src/twisted/protocols/ftp.py
```diff
@@ -945,6 +978,29 @@ def ftp_PASV(self):
...
         """
+        if self.epsvAll:
+            return defer.fail(BadCmdSequenceError(
+                'may not send PASV after EPSV ALL'))
+
+        host = self.transport.getHost().host
+        try:
+            address = ipaddress.IPv6Address(_coercedUnicode(host))
+        except ipaddress.AddressValueError:
+            pass
```
glyph, line 1010: Minor stylistic note: the best exception-handling form here would be to be *even more* constrained here; i.e.

```
coercedHost = _coercedUnicode(host)
try:
    address = IPv6Address(coercedHost)
except ...:
    pass
```

Mainly just a note for future reference.

src/twisted/test/test_ftp.py
```diff
@@ -36,8 +37,75 @@
 
 
 
+def _has_ipv6():
```
glyph: `test_conch` and `test_udp` both already define functions with *exactly* this name, and almost exactly this same implementation. 3 total repetitions of this code is too many :).  Please refactor it into a common (private, for now) location, so we're not making 3 different useless sockets during the test suite.

src/twisted/test/test_ftp.py
```diff
@@ -36,8 +37,75 @@
...
+
+
+def skipWithoutIPv6(f):
+    if not HAS_IPV6:
+        f.skip = "Does not work on systems without IPv6 support."
+    return f
+
+
+
+def _has_ipv4_mapped_addresses():
```
glyph: ahem, `_hasIPv4MappedAddresses` :)

src/twisted/test/test_ftp.py
```diff
@@ -11,6 +11,7 @@
 import getpass
 import string
 import random
+import socket
```
glyph, line 13: For what it's worth, these tests are in the old, bad style.  The new, good style would be to consistently use `twisted.internet.testing.MemoryReactor` and not require real, actual IPv6 local connectivity in order to cover this code.  I realize that this isn't really a new problem in this branch, so it doesn't need to be addressed by this branch, but I feel like I should really bring it up to get the word out.

src/twisted/test/test_ftp.py
```diff
@@ -1298,6 +1568,74 @@ def checkDownload(download):
 
 
 
+class FTPServerEprtDataConnectionTests(FTPServerPasvDataConnectionTests):
```
glyph, line 1573: Can you add a test docstring explaining what this covers more precisely?

src/twisted/test/test_ftp.py
```diff
@@ -1298,6 +1568,74 @@ def checkDownload(download):
...
+        d = defer.maybeDeferred(
+            FTPServerPasvDataConnectionTests.tearDown, self)
+        dl.append(d)
+        return defer.DeferredList(dl, fireOnOneErrback=True)
+
+
+    def test_EPRTCannotConnect(self):
+        """
+        Listen on a port, and immediately stop listening as a way to find a
+        port number that is definitely closed.
```
glyph: "definitely" is a strong word here :).

src/twisted/test/test_ftp.py
```diff
@@ -1298,6 +1568,74 @@ def checkDownload(download):
...
+        cmd = 'EPRT |1|127.0.0.1|%s|' % dataPort.getHost().port
+        self.client.queueStringCommand(cmd)
+        return deferred
+
+
+    def tearDown(self):
+        """
+        Tear down the connection.
+
+        @return: L{defer.DeferredList}
```
glyph: that fires when...?

src/twisted/test/test_ftp.py
```diff
@@ -1298,6 +1568,74 @@ def checkDownload(download):
...
+              for port in self.dataPorts]
+        d = defer.maybeDeferred(
+            FTPServerPasvDataConnectionTests.tearDown, self)
+        dl.append(d)
+        return defer.DeferredList(dl, fireOnOneErrback=True)
+
+
+    def test_EPRTCannotConnect(self):
+        """
+        Listen on a port, and immediately stop listening as a way to find a
```
glyph: Can you describe this in terms of what's being verified rather than what the setup does?  ( https://jml.io/pages/test-docstrings.html is the canonical guideline. )

src/twisted/python/_setup.py
```diff
@@ -288,6 +288,7 @@ class my_build_ext(build_ext_twisted):
         "hyperlink >= 17.1.1",
         "PyHamcrest >= 1.9.0",
         "attrs >= 17.4.0",
+        "ipaddress; python_version < '3.0'",
```
glyph: Thanks for correctly updating the package metadata for this requirement.  It strikes me as slightly unfortunate that this is our first `ipaddress` import anywhere, since we consistently use `twisted.internet.address` everywhere else.  Since I'm sending this back to you for feedback, maybe *consider* adding the IPv4 mapped test to our own class?  I wouldn't say it should be a hard requirement, since we probably transitively depend on this six different ways anyway now, and it is in the stdlib, but it does stick out just a little.

If you consider it and think it's the best choice, then by all means leave it :)

glyph (changes requested):
The changes requested are pretty minor; for the most part this is a great change.

Regarding the lint thing: when I'm looking for reviews to do, I tend to hit the ones with green checkmarks first, so I'd probably go ahead and make the style inconsistent to appease the mechanical checks.  It's not a dealbreaker, but it'll make it easier and quicker to complete the review.  Maybe we should do a lint-ratchet reduction PR to reformat the rest of the surrounding stuff but let's not bundle those together.

adiroiban (commented):
I didn't had time to finalize the review...but  I am sending my current comments.

cjwatson:
I filed #1190 to clean up lint, and will return to this once that's merged.

src/twisted/protocols/ftp.py
cjwatson: Effectively done by merging current Twisted trunk, which dropped Python 2 compatibility and thus the `_coercedUnicode` call just goes away.

src/twisted/protocols/ftp.py
cjwatson: I've added the `@param`.  The ambiguous default was an issue when you said this, but is no longer ambiguous now that Twisted only supports Python 3.

src/twisted/python/_setup.py
cjwatson: I agree this is novel, and I've considered the change you suggest.  I think if `twisted.internet.address` had any interrogation properties or methods rather than being pure data containers (with a few dunder methods), my answer might be different.  As it is, it seems really difficult to justify the effort and extra API surface of adding this to `twisted.internet.address` when there's already an adequate facility in the stdlib, especially now that we require Python 3 and so we no longer need an extra dependency for it.

So I prefer to leave this as it is, although I've dropped the setup change since `python_version < '3.0'` is now always true for Twisted.

src/twisted/protocols/ftp.py
cjwatson: I suppose not.  I've made them private now.

src/twisted/protocols/ftp.py
cjwatson: I suppose I could have done, but I was following the existing pattern set by `ftp_PASV` and `ftp_PORT`, and it didn't really seem worth lifting this up to the top level when each message is only used in one place.

src/twisted/protocols/ftp.py
cjwatson: I've made the error message clearer.

Not supporting IPv4-mapped addresses caused a [regression](https://bugs.launchpad.net/txpkgupload/+bug/1836059) in our environment, which is why I added this.  It's a relatively small amount of extra code compared to `EPSV` as a whole.

cjwatson:
@glyph @adiroiban Sorry for taking nearly two years to get round to dealing with your review comments!  I think I've dealt with pretty much everything now, apart from the `MemoryReactor` comment (which I agree with but I don't have the energy to refactor all the FTP tests at this point, and also as @glyph says it's not a new problem in this branch), and the `narrativedocs` test failures which are fixed in #1667, so this should be ready for another round of review now.

glyph:
No need to apologize — if anything, managing to finally come back after 2 years rather than just dropping it forever is praiseworthy :).

adiroiban (approved):
Looks good. Thanks.

I think that this can be merged as it's already a big improvement over trunk.

----------

 for my project I am using a highly modified version of twisted FTP protocol with support for py2 and Implicit/Explict FTPS, x509 based authentication, and machine listing and a few other extensions . client and server side.

if such functionality is needed in twisted and someone has time to review the changes, let me know and I can look at sending them for review

adiroiban:
I am enabled auto-merge. It looks like  the merge is blocked as @glyph  has rejected the changes.

Once the changes are approved by @glyph the PR should be automatically merged.

Thanks

adiroiban:
@glyph can  we merge this? Do you still think that this needs changes? Thanks!

src/twisted/python/_setup.py
glyph: > So I prefer to leave this as it is, although I've dropped the setup change since `python_version < '3.0'` is now always true for Twisted.

always False for Twisted, I hope :)

glyph (approved):

glyph:
In the future, you can always feel free to dismiss my reviews if you've reviewed the points that I raised and they've been addressed.

PR #1153, opened 2019-06-04
[9648] Pretty PyPI page

twm:
## Summary

* Adjust URLs to HTTPS instead of HTTP
* Add project URLs for documentation, source, and issues. They will display on PyPI like this: 
screenshot

* Generate the long description of the package from README.rst. A slightly hackish regex replacement is used to make relative URLs absolute so that they will work when it is displayed at https://pypi.org/project/Twisted/ . Refer to https://packaging.python.org/specifications/core-metadata/#description-content-type.
* Add a Tox environment that runs the `twine check` command. This validates that the reStructuredText of the long description is valid. (I explored adding a test in `test_setup.py` for this, but [readme_renderer](https://pypi.org/project/readme_renderer/) isn't really documented for API use and has a pretty hefty dependency tree, so I found it a bit much for the assurance it would offer.)
* Remove some badges that were rendered obsolete by the switch to Coveralls and Azure Piplelines.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9648#comment:1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,10 +41,10 @@ Additional instructions for installing this software are in `the installation in
 Documentation and Support
 -------------------------
 
-Twisted's documentation is available from the `Twisted Matrix website <http://twistedmatrix.com/documents/current/>`_.
+Twisted's documentation is available from the `Twisted Matrix website <https://twistedmatrix.com/documents/current/>`_.
 This documentation contains how-tos, code examples, and an API reference.
 
-Help is also available on the `Twisted mailing list <http://twistedmatrix.com/cgi-bin/mailman/listinfo/twisted-python>`_.
+Help is also available on the `Twisted mailing list <https://twistedmatrix.com/cgi-bin/mailman/listinfo/twisted-python>`_.
 
 There is also a pair of very lively IRC channels, ``#twisted`` (for general Twisted questions) and ``#twisted.web`` (for Twisted Web), on ``chat.freenode.net``.
 
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,6 @@ Twisted
 =======
 
 |pypi|_
-|coverage|_
 |travis|_
 |circleci|_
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -41,10 +39,10 @@ Additional instructions for installing this software are in `the installation in
 Documentation and Support
 -------------------------
 
-Twisted's documentation is available from the `Twisted Matrix website <http://twistedmatrix.com/documents/current/>`_.
+Twisted's documentation is available from the `Twisted Matrix website <https://twistedmatrix.com/documents/current/>`_.
```
glyph, line 42: Thanks for fixing this.

README.rst
```diff
@@ -2,9 +2,7 @@ Twisted
 =======
 
 |pypi|_
-|coverage|_
```
glyph, line 5: Not that I disagree, but why did you remove these, out of curiosity?

src/twisted/python/test/test_setup.py
```diff
@@ -69,7 +69,7 @@ def test_win32Definition(self):
 
 
 
-class OptionalDependenciesTests(TestCase):
+class OptionalDependenciesTests(SynchronousTestCase):
```
glyph, line 72: ❤️

src/twisted/python/test/test_setup.py
```diff
@@ -363,3 +363,39 @@ def test_find_package_modules(self):
             ]),
             sorted(result),
         )
+
+
+
+class LongDescriptionTests(SynchronousTestCase):
+    """
+    Tests for C{_getLongDescriptionArgs()}
```
glyph, line 371: Nitpick: direct references like this are generally an `L{}`.

glyph (approved):
I love everything about this.

README.rst
twm: This is the codecov badge. We switched to Coveralls, no?

src/twisted/python/test/test_setup.py
twm: :sob:

I seem to be unable to internalize epytext.

twm:
Thanks for the review, @glyph!

README.rst
glyph: We did, but they have a badge too :)

glyph: Oh, and I guess we're going to turn off Appveyor now that we have Azure?

PR #1156, opened 2019-06-13
[#9656] Disable traceback printing by default in Site and update twisted.web.tap to have an enable flag instead of a disable flag

hawkowl:
https://twistedmatrix.com/trac/ticket/9656#ticket


twm (approved):
It's great to see this hole plugged! The changes here look good. My only concern is that deprecation warnings are quite unlikely to be seen when running `twist` or `twistd`, since the Python interpreter suppresses them by default. We've already shipped warnings like this, though, so I don't think this need be addressed as part of this PR.

Merge away!

PR #1157, opened 2019-06-14
[9657] Update twisted.runner.procmon to use twisted.logger instead of twisted.python.log

jfhbrook:
This pull request updates the logging inside of twisted.runner.procmon to use twisted.logger. It also includes a new field that tracks whether the data came from stdout or stderr. Formatted output should be backwards compatible.

# Trac ticket: https://twistedmatrix.com/trac/ticket/9657#9657

# Style checks:

It seems like linting might be broken because the current version of pycodestyle [updated its list of arguments](https://github.com/PyCQA/pycodestyle/blob/3b258c375c6392b1475a487c0961cf7dbe03f838/pycodestyle.py#L316-L319) so that they [don't match the modified one in twistedchecker](https://github.com/twisted/twistedchecker/blob/master/twistedchecker/checkers/pycodestyleformat.py#L218). A naive patch in the tox environment made this pass but I haven't investigated deeply enough to know for sure. [That's now tracked here.](https://github.com/twisted/twistedchecker/issues/138)

# Newsfragments

I tried putting a newsfragment in `./src/twisted/runner/newsfragment`, if that's wrong I'm happy to move.

# tests:

Updated!


src/twisted/runner/procmon.py
```diff
@@ -71,14 +72,18 @@ class DummyTransport:
...
+    stream = None
     delimiter = b'\n'
 
     def lineReceived(self, line):
         try:
             line = line.decode('utf-8')
         except UnicodeDecodeError:
             line = repr(line)
-        log.msg(u'[%s] %s' % (self.tag, line))
+        ProcessMonitor.log.info(u'[{tag}] {line}',
```
jfhbrook: I'm pretty open to changing the names on these fields, since they didn't have semantic meaning with the old logger.

jfhbrook: Also had a shower thought: Calling this on ProcessMonitor means that it isn't possible to override the logger on an instance level. It's a bigger refactor, but we could also replace the dummy transport with a way to get the data back to the protocol and ultimately the monitor instance.

src/twisted/runner/procmon.py
```diff
@@ -90,19 +106,33 @@ class LoggingProtocol(protocol.ProcessProtocol):
...
         self.output.makeConnection(transport)
+        self.error.makeConnection(transport)
 
 
     def outReceived(self, data):
         self.output.dataReceived(data)
-        self.empty = data[-1] == b'\n'
-
-    errReceived = outReceived
+        self.outputEmpty = data[-1] == b'\n'
```
hawkowl: self.empty is technically public API here and so needs to be preserved (sorry). This is probably fine enough, since it only needs to be read only, because writing to it externally isn't supported anyway...

```
@property
def empty(self):
    return self.outputEmpty
```

src/twisted/runner/procmon.py
```diff
@@ -90,19 +106,33 @@ class LoggingProtocol(protocol.ProcessProtocol):
     def connectionMade(self):
         self.output = LineLogger()
         self.output.tag = self.name
+        self.output.stream = 'stdout'
```
hawkowl: These should probably be underscored, I don't see a reason for them to be public API. If you make them underscore prefixed, with a compat alias to `self.output` (and `self.empty` as mentioned below), that'd be great.

src/twisted/runner/procmon.py
```diff
@@ -145,11 +175,16 @@ class ProcessMonitor(service.Service):
     @ivar _reactor: A provider of L{IReactorProcess} and L{IReactorTime}
         which will be used to spawn processes and register delayed calls.
 
+    @type log: C{Logger}
```
hawkowl: This should be `L{Logger}` to produce an API link.

src/twisted/runner/procmon.py
```diff
@@ -145,11 +175,16 @@ class ProcessMonitor(service.Service):
...
+    @ivar log: An instance of twisted.logger.Logger used for propagating
+        log messages from spawned processes.
+
     """
     threshold = 1
     killTime = 5
     minRestartDelay = 1
     maxRestartDelay = 3600
+    log = Logger()
```
hawkowl, line 195: I was wondering if this should be `_log` to prevent making it a part of the public API as we would do in other cases, but I think this is correct for two reasons:

1. It's called by more than inside this class, since the public `LineLogger` calls it, and so making it private would require users who wish to use `LineLogger` (for whatever reason) have to implement a private API. If LineLogger were a private API, then this should be `_log`.

2. Having the logger defined here is important so that `ProcessMonitor` is given as the "source", not `LineLogger` (which is less useful). Therefore, we can't sidestep this problem by giving `LineLogger` their own loggers. This could be done by manually providing the source when loggers in `LineLogger`, but that feels fragile.

So, I think `log` here is the correct decision, even though it adds more public API surface. (Although, we should create an interface for Logger, so that we can say "an implementer of `ILogger`" in case we want to change the concrete implementation of Logger to something else while maintaining the same API.)

src/twisted/runner/procmon.py
```diff
@@ -145,11 +175,16 @@ class ProcessMonitor(service.Service):
     @ivar _reactor: A provider of L{IReactorProcess} and L{IReactorTime}
         which will be used to spawn processes and register delayed calls.
 
+    @type log: C{Logger}
+    @ivar log: An instance of twisted.logger.Logger used for propagating
```
hawkowl: Since that is given in the type (and the `L{}` syntax will create a link in the API docs), this should be "The logger used to propagate log messages from spawned processes" or so.

src/twisted/runner/newsfragments/9657.doc
```diff
@@ -0,0 +1 @@
+twisted.runner.procmon now logs to a twisted.logger.Logger instance defined on instances of ProcessMonitor instead of to the global legacy twisted.python.log instance
```
hawkowl: This needs a full stop at the end.

hawkowl (dismissed):
Hi @jfhbrook ! Thanks for the patch!

This is good overall and I think is an improvement, there's just some minor things regarding adding to the public API surface and docstrings that need adjustment. Even though this isn't the most used public API, our compat policy has to be consistent, grumble grumble :p 

Once you put it up for review again using the trac tag, feel free to ping me on IRC or so for a re-review. (I get so many github notifications that I won't see if you mention it here :( )

Thanks again!

src/twisted/runner/procmon.py
jfhbrook: I interpreted this as saying that creating the ILogger is something to be done in another PR, but let me know if that wasn't the expectation here.

glyph:
Should this be back in review?

jfhbrook:
I tried to address everything that I was intended to address but I wasn't sure what next steps were. I *think* those next steps are me signalling that I'm ready for a re-review.

glyph:
@jfhbrook Yep, that was it - thanks for resubmitting!

glyph:
Just catching up on some old Github emails - thank you for this contribution!

PR #1158, opened 2019-06-15
[#9653] Make HTTP/2 server connections not timeout after a minute

hawkowl:
Ticket: https://twistedmatrix.com/trac/ticket/9653


src/twisted/protocols/policies.py
```diff
@@ -712,6 +712,22 @@ def resetTimeout(self):
         if self.__timeoutCall is not None and self.timeOut is not None:
             self.__timeoutCall.reset(self.timeOut)
 
+
+    def cancelTimeout(self):
```
markrwilliams: On the one hand this is a new public API that isn't necessary; [passing `None` to `setTimeout` will cancel the timeout](https://twistedmatrix.com/documents/19.2.1/api/twisted.protocols.policies.TimeoutMixin.html#setTimeout).
On the other hand, `setTimeout`'s API is bad enough that a core committer with deep Twisted experience didn't see it, and I didn't either until I went hunting.  Maybe if we'd had this method in the first place we wouldn't have had this bug!
So I think I'm +0.75 for adding this method.
Should we deprecate the cancellation feature of `setTimeout`?

src/twisted/web/test/test_http.py
```diff
@@ -764,6 +768,60 @@ def _negotiatedProtocolForTransportInstance(self, t):
         return a._negotiatedProtocol
 
 
+    def test_h2CancelsH11Timeout(self):
+        """
+        When the transport is switched to H2, the HTTPChannel timeouts are
+        cancelled.
+        """
+        @attr.s
+        class FakeDelayedCall(object):
```
markrwilliams: Why not construct a real DelayedCall?

Also, it'd be nice if this implemented [`IDelayedCall`](https://twistedmatrix.com/documents/19.2.1/api/twisted.internet.interfaces.IDelayedCall.html) and there were a `verifyObject` call somewhere.

src/twisted/web/http.py
```diff
@@ -2947,6 +2947,9 @@ def dataReceived(self, data):
                 # itself if possible.
                 self._channel._networkProducer.unregisterProducer()
 
+                # Cancel the old channel's timeout.
+                self._channel.cancelTimeout()
+
                 transport = self._channel.transport
```
markrwilliams, line 2961: What do you think about removing the transport from the original channel as well, for a bit of defensive programming?

markrwilliams (approved):
Thank you for jumping on this! :100: tests and comments.

I've left some comments, but given the severity of this bug, please don't let addressing them prevent you from merging this fix!

glyph:
Bumping this because as @markrwilliams said, it is indeed fairly severe...

PR #1159, opened 2019-06-19
[#9660][twisted/internet/_glibbase] Fix Typeerror in simulate for py3

doadin:


src/twisted/newsfragments/9660.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet._glibbase.Simulate no longer raises TypeError with no timeout on Python 3.
```
twm: Please use the public name of the object in the changelog (`_glibbase` is a private module). Also, `Simulate` shouldn't be capitalized.

src/twisted/internet/_glibbase.py
```diff
@@ -382,9 +382,10 @@ def simulate(self):
         if self._simtag is not None:
             self._source_remove(self._simtag)
         self.iterate()
-        timeout = min(self.timeout(), 0.01)
+        timeout = self.timeout()
         if timeout is None:
             timeout = 0.01
+        timeout = min(self.timeout(), 0.01)
```
twm: Would this be cleaner as an `or`?

```py
if timeout is None or timeout > 0.01:
    timeout = 0.01
```

This would avoid the second call to `self.timeout()`.

twm (changes requested):
Hi @doadin, thank you for diving in and fixing this! However, we do need a test that exercises this code before we can merge &mdash; see the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html). I also noted a minor issue with the newsfile.

PR #922, opened 2017-10-31
Update pb.py

congzhangzh:
under python3, password will be a str but not bytes which need to be encode before call md5().update

downstream affect like: buildbot/buildbot#3726

PR #1160, opened 2019-06-30
fix issue https://twistedmatrix.com/trac/ticket/9667

congzhangzh:
https://twistedmatrix.com/trac/ticket/9667


congzhangzh:
downstream issue https://github.com/buildbot/buildbot/issues/3726

the respond function should work for both python2 and python3, as Buildbot work for python3 now.

Julian:
Hi @congzhangzh thanks for the contribution!

I'm not myself that familiar with the surrounding code in pb here, but from looking at the patch it's not completely obvious this is the right solution to me -- at very least, it needs a test showing either the issue or the proposed new behavior (which I can't tell which this falls into yet) :). I do see other pieces of surrounding code doing some implicit en/decoding in particular areas (e.g. it looks like `callRemote` is OK with being passed unicode), but that doesn't look like it's being done everywhere, and likely for good reason -- e.g. here it'd be surprising if in fact the function here for "respond" *did* take both unicode and bytes, rather than just asking the caller to provide the right type when doing the remote call. Again though, very possible I'm not familiar with the pb-specific norms. 

(Implicitly encoding to ascii also sounds suspicious -- if there were to be some implicit encoding, why ASCII and not UTF-8?)

Let me know your thoughts, and yeah at very least please show a test :)

congzhangzh:
Hi @Julian 

as the old https://github.com/twisted/twisted/pull/922, the other report the problem too.

>>e.g. here it'd be surprising if in fact the function here for "respond" did take both unicode and bytes, rather than just asking the caller to provide the right type when doing the remote call. Again though, very possible I'm not familiar with the pb-specific norms.

Password should be str as semantic, but MD5 requires bytes?

>>(Implicitly encoding to ASCII also sounds suspicious -- if there were to be some implicit encoding, why ASCII and not UTF-8?)

Password should be simple ASCII, right?

exarkun:
`twisted.spread.pb.respond` is called from `PBClientFactory._cbResponse` with its `password` argument.  `_cbResponse` gets its `password` argument from its caller, `PBClientFactory._cbSendUsername` which is just passing on its own `password` argument.  This comes from its caller, `PBClientFactory.login` which takes it out of the `credentials` value it receives.  The `credentials` value is an `IUsernamePassword` provider.  `IUsernamePassword.password` is documented as having type bytes.

Thus, `respond` expects to be called with a password of type `bytes`.  If it is called with the wrong type, indeed, it will do something wrong - such as raise an exception.

It sounds like Buildbot is constructing its `IUsernamePassword` credentials incorrectly, supplying `unicode` when it needs to be supplying `bytes`.  If Twisted were to change, the change which would seem to make sense is to add type checking to either `UsernamePassword` or to `PBClientFactory.login` so that the invalid value is detected and reporter earlier, where there's an easier chance of finding the true culprit.

congzhangzh:
@exarkun tks for your note.

@rodrigc how do you think about this?

twm:
Hi, I'm closing this PR as the changes it contains don't seem to be the right solution to the problem. The ticket remains open. Please feel free to reopen this PR or submit a fresh one. Thanks!

PR #1162, opened 2019-07-06
Fixes: ticket:9596

westfeld:
## Contributor Checklist:


glyph:
@westfeld was the intention to put this ticket into review? (We probably need to update the contribution checklist to make the need for this clearer...)

westfeld:
Well I am done implementing a fix for the ticket I cited, all tests passed, so I though a final review is needed to get it merged into trunk.

I would like to give something back to the twisted community but until now both attempts to fix something failed not because of the programming part, but because I did something wrong in the merging process.

I would appreciate a better documentation about how to support twisted and I would be happy to contribute.

glyph:
> I would appreciate a better documentation about how to support twisted and I would be happy to contribute.

Point taken.  I've got a PR up for review here: https://twistedmatrix.com/trac/ticket/9687 / https://github.com/twisted/twisted/pull/1176 which improves the PR template a little bit, including adding this instruction.  Want to review that one? :)

src/twisted/web/newsfragments/9596.bugfix
```diff
@@ -0,0 +1 @@
+Fixed return type of username and password to binary if no authorization header was found or an exception was thrown
```
glyph: Return type of what? The convention is always to include the fullly-qualified Python import path to the object being modified.  (In this case, `twisted.web.http.Request.getUser` / `.getPassword`.)

src/twisted/web/test/test_http.py
```diff
@@ -2144,7 +2144,7 @@ def process(self):
         self.patch(base64, 'decodestring', lambda x: [])
         self.runRequest(f, Request, 0)
         req = requests.pop()
-        self.assertEqual(('', ''), req.credentials)
+        self.assertEqual((b'', b''), req.credentials)
```
glyph, line 2147: Looks better :)

src/twisted/web/http.py
```diff
@@ -1535,18 +1535,18 @@ def _authorize(self):
...
             bas, upw = authh.split()
             if bas.lower() != b"basic":
                 raise ValueError()
             upw = base64.decodestring(upw)
             self.user, self.password = upw.split(b':', 1)
         except (binascii.Error, ValueError):
-            self.user = self.password = ""
-        except:
+            self.user = self.password = b''
+        except Exception:
```
glyph: This change is unnecessary and incorrect.  I assume you've got a linter yelling at you about `except:`, but to be compatible it should be `except Exception`.  Also, we don't use such a linter, so I'd rather we just skip this change entirely.

glyph (approved):
With a couple of minor tweaks I'd say that this is ready to land.

Would you mind making the two changes requested below:

- removing the `except Exception` change
- fixing the news file to include FPQNs? 

When you're done, submit for review again (add the keyword back to the Trac ticket) but note in your comment that it's pre-approved and you just need someone to commit it.

westfeld:
Now the CI pipeline is complaining about the E722 error I corrected in this commit 34172980213a761bb7d224f0b709f9265448b371

wsanchez:
@glyph Well, we did get a lint error:

```console
lint run-test: commands[3] | /bin/sh -c 'git diff trunk | python /home/travis/build/twisted/twisted/admin/pycodestyle-twisted.py --diff'
335./src/twisted/web/http.py:1547:9: E722 do not use bare 'except'
336
```

That said `pycodetyle` also shows a traceback in the log, so…

glyph:
Sorry for the incorrect analysis on the lint error; I didn't realize that this was leaking through from pycodestyle to our twistedchecker configuration; we should *probably* fix it, or we should give up and `except BaseException` everywhere we currently (legitimately) have a bareword.

Thanks @wsanchez & @hawkowl for addressing it.

PR #1166, opened 2019-07-23
Handle passwords with spaces in POP3 server.

evilham:
Closes trac #9100.
https://twistedmatrix.com/trac/ticket/9100


glyph (approved):
My only question, really, is "what about *two* spaces"?

evilham:
Oh, I missed adding such a test, but it will work because we split on `b' '` and then join with that and that trick behaves well without loss of information (read/writelines for files works like that).

```
>>> b'test  2   3'.split(b' ')
[b'test', b'', b'2', b'', b'', b'3']
```

glyph:
Thanks for checking on that!

PR #1171, opened 2019-07-31
Fix iqmp parameter in serialized RSA private keys

cjwatson:
Both OpenSSH and LSH expect the relevant parameter in their serialized
RSA private keys to be q^-1 mod p rather than p^-1 mod q.  Prior to
737db21084359e30e339d6cad38dcac487fb5051, p and q were sometimes
swapped, so this ended up correct anyway, but we should really write out
the correct value.

The "Secure Shell Key Agent v3" case may well need the same treatment,
but I haven't been able to work out which specific SSH implementation
that corresponds to in order to check.

## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9681#ticket

### DIFF
--- a/src/twisted/conch/newsfragments/9681.bugfix
+++ b/src/twisted/conch/newsfragments/9681.bugfix
@@ -1,0 +1,1 @@
+twisted.conch.ssh.keys now correctly writes the "iqmp" parameter in serialized RSA private keys as q^-1 mod p rather than p^-1 mod q.
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -1171,9 +1171,10 @@ class Key(object):
         type = self.type()
         data = self.data()
         if type == 'RSA':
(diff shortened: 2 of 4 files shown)


glyph:
This looks right to me, but I'd really love it if someone crypto-capable (@alex ? @reaperhulk ?) could glance at it and confirm before landing.

reaperhulk:
This looks correct in terms of switching to iqmp, yes.

Is `u` still useful elsewhere? Right now ipmq and iqmp are both being computed, which means this invmod func is being called twice:

https://github.com/pyca/cryptography/blob/c7681e80a68a97ba56453e7fbb960f0e59f4acad/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L183-L193

glyph:
Thanks @reaperhulk !  Let's follow up on (maybe?) getting rid of `u` if it's unnecessary in a follow-up.

PR #1173, opened 2019-08-01
Accept SSH environment variable passing

cjwatson:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9315


src/twisted/conch/interfaces.py
```diff
@@ -82,6 +82,26 @@ def getPty(term, windowSize, modes):
         fails, raise an exception.
         """
 
+    def setEnv(name, value):
```
twm: Unfortunately this isn't backwards compatible. [Interfaces are forever.](https://orbifold.xyz/interfaces-are-forever.html) You will have to create a new interface to host this method, e.g. `ISetEnv`, and check if the session object implements it in `request_env`.

src/twisted/conch/interfaces.py
```diff
@@ -82,6 +82,26 @@ def getPty(term, windowSize, modes):
...
+        Set an environment variable for the shell or command to be started.
+
+        From U{RFC 4254, section 6.4
+        <https://tools.ietf.org/html/rfc4254#section-6.4>}: "Uncontrolled
+        setting of environment variables in a privileged process can be a
+        security hazard.  It is recommended that implementations either
+        maintain a list of allowable variable names or only set environment
+        variables after the server process has dropped sufficient
+        privileges."
+
```
twm: It would be useful to mention what the OpenSSH-compatible behavior is here, since any alternative implementation will probably want to implement it.

src/twisted/conch/ssh/session.py
```diff
@@ -91,6 +91,25 @@ def request_pty_req(self, data):
         else:
             return 1
 
+    def request_env(self, data):
+        """
+        Process a request to pass an environment variable.
+
+        @type data: L{bytes}
```
twm, line 132: Please add `@param` and `@returns` as well, per the [coding standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html).

src/twisted/conch/test/test_session.py
```diff
@@ -705,10 +722,29 @@ def test_requestPty(self):
...
+        self.assertEqual(
+            self.session.session.ptyRequest, (b'good', (1, 2, 3, 4), []))
+
+
+    def test_setEnv(self):
+        """
+        When a client requests passing an environment variable, the
+        SSHSession object should make the request by getting an ISession
+        adapter for the avatar, then calling setEnv with the environment
+        variable name and value.
```
twm: This test doesn't seem to cover the `ValueError` code path that silently ignores a variable. We need that coverage.

twm (changes requested):
Hi Colin,

This is a great feature, and it's great to see it happening! I noted a few things in the review comments below. You'll need to fix the incompatible interface change and add the missing test coverage before we can land.  I'll add that I'm not too familiar with Conch myself, so I'm mostly looking at this from a general Twisted background and might miss Conch-specific stuff.

The ticket mentions implementing OpenSSH's environment filtering behavior. It seems like Conch's default server should do this, though if you prefer to file a follow-up ticket for that it's fine.

src/twisted/conch/interfaces.py
cjwatson: So, I understand that argument and I'm willing to do this, but I thought I'd first explain why I did it this way and see whether it changes your mind.

Generally speaking I try to fit into the style of existing code before almost any other consideration.  In this case, all of the `request_*` methods in `SSHSession` wrap their calls to the session object in very general exception handlers.  (I couldn't quite bring myself to imitate the bare `except`s that they use, but I imitated the rest of it.)  As such, it is effectively already true that `request_env` checks whether the session object implements the necessary method: if it doesn't, it results in an error being logged, but that's as far as it goes.  Of the current `ISession` implementers in Twisted itself, `docs/conch/examples/sshsimpleserver.py:ExampleSession`, `twisted.conch.manhole_ssh.TerminalSession`, and `twisted.conch.test.test_conch.RekeyAvatar` already fail to implement all of the methods in `ISession`, so for better or worse the current code can't be being strict about it.

Now, I can see that there might be an argument that `ISession` is a public interface and so external code is entitled to depend on it in other ways; but the above is why I took this approach in the first pass.

src/twisted/conch/interfaces.py
cjwatson: Done.

src/twisted/conch/ssh/session.py
cjwatson: Done.

src/twisted/conch/test/test_session.py
cjwatson: Done.

cjwatson:
Regarding the behaviour of Conch's default server: OpenSSH's default behaviour is to reject all environment variables, and the only default server that I know of that Conch ships is `twisted.conch.tap` which only has very minimal configuration.  My understanding is that the typical way to use Conch as a server is to use it as a library of components from which you can assemble a server with the custom behaviour you want (that's certainly how we use it in Launchpad).  But I could be wrong: perhaps @markrwilliams can elaborate on their requirements?

Note that the cited behaviour of OpenSSH in accepting `LANG` and `LC_*` is *not* the default behaviour upstream, but rather a [Debian-specific customisation](https://manpages.debian.org/stable/openssh-server/sshd_config.5.en.html).

src/twisted/conch/interfaces.py
twm: > Now, I can see that there might be an argument that `ISession` is a public interface and so external code is entitled to depend on it in other ways; but the above is why I took this approach in the first pass.

Yeah, that's pretty much it. I don't view it so much as an argument as Twisted policy. But maybe this should be an exception if none of the real-world implementations actually _implement_ the interface. This seems like a question for the mailing list &mdash; again, I'm not too familiar with Conch.

cjwatson: It's OK, I just wanted to get an idea of how firm the requirement was.  I've moved this to a new `ISessionSetEnv` interface (which necessitated some test rearrangements to be able to effectively test the case where the avatar doesn't implement that interface).

glyph:
FWIW I’d love it if Conch’s `tap` command line were a *bit* more robust in terms of supporting this sort of configuration.

src/twisted/conch/ssh/session.py
```diff
@@ -91,6 +92,33 @@ def request_pty_req(self, data):
...
+            as an SSH protocol string and concatenated.
+        @type data: L{bytes}
+        @return: A true value if the request to pass this environment
+            variable was accepted, otherwise a false value.
+        """
+        if not self.sessionSetEnv:
+            try:
+                self.sessionSetEnv = ISessionSetEnv(self.avatar)
+            except TypeError:
+                log.msg('unhandled request for env')
```
glyph: It makes me a bit sad to see a new old-style log message get added like this 😞 .  But I guess we don't have a great "list of migrations in progress" document that reviewers should look at, so it's not surprising it got missed.

To make a new-style log message, make a new `twisted.logger.Logger` instance at module scope, and call `log.<appropriate level method>`.  In this case, probably `log.info`.

Also: unhandled request for *what* env?  Of whom?  All log messages should include relevant information that will make them actionable by administrators, or at least by the implementor of the avatar class.  In this case I'd say something like `_log.info('SSH avatar {avatar} does not handle setting requested environment variables {env}', avatar=self.avatar, env=data)` or similar.

src/twisted/conch/ssh/session.py
```diff
@@ -91,6 +92,33 @@ def request_pty_req(self, data):
...
+
+        @param data: The environment variable name and value, each encoded
+            as an SSH protocol string and concatenated.
+        @type data: L{bytes}
+        @return: A true value if the request to pass this environment
+            variable was accepted, otherwise a false value.
+        """
+        if not self.sessionSetEnv:
+            try:
+                self.sessionSetEnv = ISessionSetEnv(self.avatar)
```
glyph: A better way to address the lack of type information than catching `TypeError` (which may be raised by a great many things in Python's runtime or standard library) is to do `ISessionSetEnv(self.avatar, None)`.

src/twisted/conch/ssh/session.py
```diff
@@ -33,9 +33,10 @@ def __init__(self, *args, **kw):
         self.buf = b''
         self.client = None
         self.session = None
+        self.sessionSetEnv = None
```
glyph: This interface is already somewhat over-broad; is it really necessary to make `sessionSetEnv` a public read/write variable here?  If so, we definitely need to document it for callers.  I mean, it'd be nice to have documentation of its type either way, but an underscore prefix means that documentation is only really maintainer-facing rather than user-facing ;)

src/twisted/conch/ssh/session.py
```diff
@@ -33,9 +33,10 @@ def __init__(self, *args, **kw):
...
 
     def request_subsystem(self, data):
-        subsystem, ignored= common.getNS(data)
+        subsystem, ignored = common.getNS(data)
```
glyph, line 68: (Thanks for the cleanup.)

src/twisted/conch/interfaces.py
```diff
@@ -113,6 +113,35 @@ def closed():
...
+        (OpenSSH refuses all environment variables by default, but has an
+        C{AcceptEnv} configuration option to select specific variables to
+        accept.)
+
+        @param name: The name of the environment variable to set.
+        @type name: L{bytes}
+        @param value: The value of the environment variable to set.
+        @type value: L{bytes}
+        @raise ValueError: if setting this environment variable is not
```
glyph: `ValueError` seems like a pretty broad exception to catch with no logging at all.  For example, if your application code is expecting an `int`-able value and your client transmitted `"1.0"`, you'll get a `ValueError` and it'll get swallowed.

In general, this is why using stdlib exceptions is generally a bad idea for specific failure modes.

Could we define a new exception here, `EnvironmentVariableNotPermitted`, so it's clear that it's being raised intentionally?  Alternately, if the only two possible behaviors here are "variable was set, OK" and "variable rejected", perhaps the return value ought to be an enum or something similar, and exceptions shouldn't be allowed in the interface?

src/twisted/conch/newsfragments/9315.feature
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh.session.SSHSession can now accept environment variables sent by the client, using a new ISessionSetEnv interface.
```
glyph: 👍 Love the description.  Could you qualify `ISessionSetEnv` and perhaps be a little more specific than "using" though?  Like "by your SSH avatar implementing"?

src/twisted/conch/test/test_session.py
```diff
@@ -112,6 +116,23 @@ def getPty(self, terminal, window, modes):
...
+    def setEnv(self, name, value):
+        """
+        If the requested environment variable is 'LD_PRELOAD', fail.  If it
+        is 'IGNORED', raise ValueError, which should cause it to be silently
+        ignored.  Otherwise, store the requested environment variable.
+
+        (Real applications should normally implement an allowed list rather
+        than a blocked list.)
+        """
+        if name == b'LD_PRELOAD':
```
glyph: This seems like a *suspiciously* specific value to choose for a stub test :).  Would you mind making it a less "set off security scanners" type value? :)

src/twisted/conch/ssh/session.py
```diff
@@ -91,6 +92,33 @@ def request_pty_req(self, data):
...
+        """
+        Process a request to pass an environment variable.
+
+        @param data: The environment variable name and value, each encoded
+            as an SSH protocol string and concatenated.
+        @type data: L{bytes}
+        @return: A true value if the request to pass this environment
+            variable was accepted, otherwise a false value.
+        """
+        if not self.sessionSetEnv:
```
glyph: This seems like a good candidate for the "active nothing" pattern, where it's never `None` but sometimes it's a specialized no-op implementation that refuses all variables?  This pushes some of the conditional logic below into the stub implementation, making it less complex.

src/twisted/conch/ssh/session.py
```diff
@@ -91,6 +92,33 @@ def request_pty_req(self, data):
...
+            except TypeError:
+                log.msg('unhandled request for env')
+                return 0
+        name, value, data = common.getNS(data, 2)
+        try:
+            self.sessionSetEnv.setEnv(name, value)
+        except ValueError:
+            return 0
+        except Exception:
+            log.err()
```
glyph: As above re: new vs. old style loggers: with a new style logger this would be `_log.failure()`

src/twisted/conch/test/test_session.py
```diff
@@ -78,6 +79,8 @@ class StubSessionForStubAvatar(object):
     @ivar avatar: the L{StubAvatar} we are adapting.
     @ivar ptyRequest: if present, the terminal, window size, and modes passed
         to the getPty method.
+    @ivar environ: a L{dict} of environment variables passed to the setEnv
```
glyph, line 83: Thanks for the docs!

glyph (changes requested):
Sorry to have a secondary review that piles on more feedback, but there are a couple of public-interface concerns here which I would really like to see addressed before landing.  Thanks again for doing this, this is a hugely useful feature for conch!

When you have addressed feedback, feel free to @ me here, and I'll do my best to re-review promptly to get this integrated.

src/twisted/conch/ssh/session.py
cjwatson: I'd prefer not to do this in this case, because that loses the distinction for logging purposes between "can't set environment variables because doing that isn't implemented" and "can't set environment variables because the implementation said no", which seems to me to be likely to be useful.  (Or, if I were to preserve the extra logging, as far as I can see it would make no difference to the complexity of the conditional logic below it: it still needs to cope with the three cases of OK, explicitly-forbidden, and error.)

cjwatson:
@glyph Thanks for the review.  I think I've addressed all your comments in one way or another; please re-review?

src/twisted/conch/ssh/session.py
ryban: Is there a particular reason to place the log at the module scope? It would be more helpful to me for the logger to be at a class attribute so the exact instance can be known from `log_source` in the event. In my case I want to associate logs from a certain SSH connection to the "owner" of that connection.

twm (approved):
Indeed, all the review feedback has been addressed. I will go ahead and merge (the test failure is a recurring Azure issue unrelated to these changes). Thank you very much!

src/twisted/conch/ssh/session.py
twm: @ryban I think we can address that as part of [#9773](https://twistedmatrix.com/trac/ticket/9773)

PR #1174, opened 2019-08-02
Use the undeprecated API in iosim

exarkun:
## Contributor Checklist:

* [x] There is an associated ticket in Trac. https://twistedmatrix.com/trac/ticket/9684

I have not updated the automated tests because none seem to cover deprecation warning.  I want to believe there used to be CI for deprecations but this might not really be true.  In any case, I see no such thing now.  If there is one, I'd be happy to be pointed at it.

### DIFF
--- a/src/twisted/test/iosim.py
+++ b/src/twisted/test/iosim.py
@@ -26,6 +26,7 @@ from twisted.internet import interfaces
 from twisted.internet.testing import MemoryReactorClock
 
 
+
 class TLSNegotiation:
     def __init__(self, obj, connectState):
         self.obj = obj
--- a/src/twisted/test/iosim.py
+++ b/src/twisted/test/iosim.py
@@ -23,8 +23,7 @@ from twisted.internet.error import ConnectionRefusedError
 from twisted.python.failure import Failure
 from twisted.internet import error
 from twisted.internet import interfaces
-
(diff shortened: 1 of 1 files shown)


Julian:
Looks like a flaky Windows test... Which seems unlikely to be the result of this...

twm (approved):
:+1: Looks like a reasonable change.

The failing test [is known to be flaky](https://twistedmatrix.com/trac/ticket/8879#comment:34).

glyph:
Thanks for doing the update, @exarkun !  Unfortunately I don't think we've ever made CI for warnings (including deprecations, both our own and upstream) work.  It would be great to have that.

exarkun:
FWIW Tahoe-LAFS has some CI for deprecations.  Unfortunately, it has an extremely high false positive rate (else why would I have submitted this PR ...).  Given the way the warnings module works I can't see a way to reliably deal with this.  Determining where the responsibility for correcting a warning falls seems to require human-level intelligence.

twm:
@exarkun Could you file a ticket requesting that Twisted test for deprecation warnings that describes what Tahoe-LAFS does? I would compare with what I'm doing at work, where this warning is going to cause some annoyance.

exarkun:
@twm I filed https://twistedmatrix.com/trac/ticket/9697

PR #1182, opened 2019-08-17
Have serial extra depend on pywin32 in Windows

altendky:
https://twistedmatrix.com/trac/ticket/9700

Other steps required to exercise this properly in testing:
* [ ] Add a `serial` (without `windows`) tox env job to Azure Pipelines CI

## Contributor Checklist:


src/twisted/internet/test/test_win32serialport.py
```diff
@@ -15,10 +15,18 @@
 from twisted.python.runtime import platform
 from twisted.internet.test.test_serialport import DoNothing
 
+
+testing_forced = os.environ.get('TWISTED_FORCE_SERIAL_TESTS', '')
+testing_forced = testing_forced.lower() in {'1', 'yes', 'true'}
```
exarkun: This diversity hardly seems necessary.  Set/not set is plenty of complexity to manage.

src/twisted/internet/test/test_win32serialport.py
```diff
@@ -15,10 +15,18 @@
...
+
+
 try:
     from twisted.internet import serialport
     import serial
 except ImportError:
+    if testing_forced:
```
exarkun: did Twisted switch from camelCase?

src/twisted/newsfragments/9700.bugfix
```diff
@@ -0,0 +1 @@
+Fixed installation of twisted[serial] being insufficient to import twisted.internet.serialport on Windows.
```
exarkun: This doesn't follow the style guide.

src/twisted/internet/test/test_win32serialport.py
altendky: Three choices does seem a bit much.

I do tend to like being able to explicitly both enable and disable though.  Is that part particularly a problem?  `#python` suggested `''` and `'0'` be false and all else be true.  Mostly I just think it would be good to have _some_ falsey value for boolean environment variables.

altendky:
@exarkun thanks for taking the time to discuss and help with this change.

src/twisted/internet/test/test_win32serialport.py
exarkun: I'd probably go for either (unset and) `""` is false and everything else is true or re-arrange the variable entirely.  I think having `"0"` be false is tricky because it suggests there are non-empty true values.  A less tricky alternative would be to have (unset and) `""` mean false, `"1"` mean true, and everything else be an explicit error (raise an exception, refuse to proceed).

An even more divergent alternative would be to make this something more like `TWISTED_FORCE_TESTS=serial[,...]`.  Read and validate it earlier (probably provide an API for reading and validating it) and reject any values that aren't explicitly supported.  By providing an API for interfacing with this, it becomes easier to make changes in the future and it gives an obvious point of centralization for controlling what kind of test forcing exists (which means you have something you can test, document, etc).

altendky: I should have said some _non-empty_ falsey value.  The point being to make it possible to explicitly describe the desired value with a value, not a lack of value.  I don't really follow why `"0"` and `"1"` would be tricky but it's probably more valuable to figure out how to define an environment variable as a private thing and not a part of the public API.

Yes, a whole system would be nice.  Then again we were talking about how the extras (or at least serial) should perhaps just be separate libraries.

I'll just go with set/not-set since that was the original preference.  So, how do we make this environment variable 'private'?

glyph: FWIW I'm not opposed to the "friendly" implementation (I do find it a bit weird when you do SOME_ENV_CONFIG=no and it turns the thing on rather than off), but I *am* opposed to doing it ad-hoc in the middle of a test module.  (It would be nice if we centralized this somehow more generally...)

glyph (approved):
LGTM!

PR #1184, opened 2019-08-21
[9547] Make FileBodyProducer cancellable

jchenggggg:
https://twistedmatrix.com/trac/ticket/9547

## Contributor Checklist:

`twisted.web._newclient.Request._writeToBodyProducerContentLength` returns a deferred that will be `callback`ed or `errback`ed depending on whether a Producer and Consumer pair are successful in writing the request body to the transport.  This deferred has a canceller, which calls the Producer's canceller.  However, the producer being used to reproduce the bug in 9547 (`FileBodyProducer`) has no canceller, so this just results in the cancel failure being passed to the producer's errback chain.  This errback assumes that since the producer is errbacking, it should already have stopped producing, so `stopProducing` is not called in this case.  However, the errback does stop the consumer.

This means that after the request has been cancelled, the producer will continue to produce as it is a scheduled task.  But the consumer in this case (`LengthEnforcingConsumer`) does not expect the producer to continue to write to it, and so when the producer tries to write, execution comes to a branch at `twisted/web/_newclient.py:992`, which has a comment stating that if you've reached it then you've written a bug.  In this branch, `stopProducing` is called on the producer (removing it from the task queue), and then an exception is raised (which causes the producer job to fail, and causes the task scheduler to try to remove the task from the queue again.) . This causes the exception reported in 9547.

This PR adds a canceller to `FileBodyProducer`.  The canceller calls `stopProducing`, which will prevent the producer from trying to produce after the request has already been cancelled.  As far as I can tell, `FileBodyProducer` is the only implementer of `IBodyProducer` for which it would make sense to have a canceller.  The other way to do this would have been to have `twisted.web._newclient.Request._writeToBodyProducerContentLength` call `stopProducing` directly instead of calling cancel on the producer's deferred, but to me it seems to make more sense that Producers should be the ones to know how to cancel themselves.

First time contributing, so appreciate any feedback :)


src/twisted/web/newsfragments/9547.bugfix
```diff
@@ -0,0 +1 @@
+FileBodyProducer will now stop producing when cancelled.  This behavior is relied on by twisted.web._newclient.Request.
```
glyph: - In news files, we should always qualify python identifiers that have changed - so `FileBodyProducer` should be `twisted.web.client.FileBodyProducer`
- The qualifications should always be by their *public* alias; so `twisted.web._newclient.Request` should be `twisted.web.client.Request`.  (I realize that it's annoying to figure out what the "public alias" of something defined in a private implementation module is, and that's why I wrote https://pypi.org/project/publication/ and I hope we can adopt it in Twisted one day :))
- `FileBodyProducer` doesn't itself have a `cancel` method, so it might be confusing to folks to say "when cancelled" here; "the `Deferred` returned by `FileBodyProducer.startProducing`" might be clearer.
- What does it mean that this is "relied upon" by `...client.Request`?

src/twisted/web/client.py
```diff
@@ -1170,14 +1170,24 @@ def startProducing(self, consumer):
         @param consumer: Any L{IConsumer} provider
         """
         self._task = self._cooperate(self._writeloop(consumer))
-        d = self._task.whenDone()
+
+        task_d = self._task.whenDone()
```
glyph: Twisted's variable naming convention would make this `taskD` not `task_d`.

src/twisted/web/client.py
```diff
@@ -1170,14 +1170,24 @@ def startProducing(self, consumer):
...
+
+        # Create another deferred here because cancellers can only be provided
+        # to the deferred constructor
+        def cancel_producer(ignored):
```
glyph: Same here with variable naming - `cancelProducer`

src/twisted/web/client.py
```diff
@@ -1170,14 +1170,24 @@ def startProducing(self, consumer):
...
+            self.stopProducing()
+            task_d.cancel()
+        producer_d = defer.Deferred(canceller=cancel_producer)
```
glyph: (and `producer_d` should be `producerD`)

glyph (approved):
Hi @jchenggggg !  Thanks for this contribution.  It's a great enhancement and I've put some minor feedback here.  One other thing that would be great to have would be to update `startProducing`'s docstring to explain what the `Deferred` does when you cancel it.

Would you mind making these tweaks, then re-attaching the 'review' keyword to the ticket?

Thanks!

twm (commented):
It would also be helpful to folks who implement  `IBodyProducer` in the future if the interface documented the expected cancellation behavior.

src/twisted/web/newsfragments/9547.bugfix
jchenggggg: Thanks!

> What does it mean that this is "relied upon" by ...client.Request?

I suppose it means that `...client.Request` has a method that seems to expect that calling `cancel()` on the `Deferred` which was returned from `FileBodyProducer.startProducing()` will actually do something?  Perhaps this isn't relevant here; I just removed the sentence!

glyph:
I made a few more tweaks to minimize the delta to the implementation while keeping the test that you supplied the same.  I'll merge if everything passes.

I think that the `stopProducing` causing a `Deferred` to *never fire* might be an error in the way that this is specified; it seems like it really ought to cause the `Deferred` to fail; every `Deferred` should eventually go one way or the other.  But that's not really this change's fault, so let's table the discussion for a future change. (cc @twm)

glyph:
Skipped the lint check because it was firing spuriously because of internal commits; the overall diff was leaving the line unchanged.

twm:
I agree, the current behavior is at best non-obvious. Regardless, thank you @jchenggggg and @glyph!

glyph:
@twm do you want to start a mailing list thread, or new trac ticket, for making the Deferred in question fail rather than never-returning?

twm:
I filed a ticket here: https://twistedmatrix.com/trac/ticket/9712 Given the compatibility implications I'm not sure it's a change worth making, though.

PR #1186, opened 2019-08-26
[9703][WIP] Add support for diffie-hellman-group14-sha256 in SSH transport.

adiroiban:
# Scope

This adds support for diffie-hellman-group14-sha256 in Conch SSH transport for both client and server side.

# Changes

Update _kex.py to know about SHA256 variant of Group14 and remove hardcoded SHA1 usage

Tests were done for server-side using OpenSSH client.

As a drive by, I have updated sshsimpleserver.py example to know about Group exchange greater than 4k..and remove the usage of public key on the server side as it can be obtained from the private key.

# To do

* Write OpenSSH integration tests for client side
* I don't know if unit test make sense... maybe write an unit test to make sure this is present in the list of default KEX

## Contributor Checklist:

* https://twistedmatrix.com/trac/ticket/9703#ticket


adiroiban:
still WIP as I need to write the tests for the client side

PR #1191, opened 2019-10-17
Fix serialization of date, time, and timedelta objects in python 3

mathieui:
Unicode strings cannot be sent as is in Banana, and need to be
byte-encoded. In python 3, this would fail for the date, time and
timedelta objects as the jellyfication would use a unicode string and
not a byte string.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9716
* [ ] I have updated the automated tests. (I do not know where to start writing such tests)


exarkun (changes requested):
Thanks.  It looks like this breaks the test suite.  Also, the fix needs to include a new test that demonstrates what was broken before and that the change makes it work.

glyph:
Thanks for your contribution @mathieui !

PR #1193, opened 2019-10-21
Support encoding to new OpenSSH private key format

cjwatson:
This is the other half of 4b6a2a2d85bc1493d8d0d1eb90a3896ac220da5c, and
is a prerequisite for adding support for Ed25519 keys which OpenSSH only
ever writes in the new format.

OpenSSH refuses to load keys in this format that lack a trailing
newline, so I adjusted the test data accordingly.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9683


src/twisted/conch/ssh/keys.py
```diff
@@ -1201,25 +1208,194 @@ def toString(self, type, extra=None):
...
+        @param passphrase: A passphrase to encrypt the key with.  Only
+            supported for private OpenSSH keys.
+        @type passphrase: L{bytes} or L{unicode} or L{None}
+
         @rtype: L{bytes}
         """
-        if isinstance(extra, unicode):
-            extra = extra.encode("utf-8")
+        if extra is not None:
+            # Compatibility with old parameter format.
```
glyph, line 1239: The docstring says this is deprecated, so please emit an actual deprecation warning.

Sadly, `twisted.python.deprecate` doesn't seem to have a "deprecated parameter" utility, so an ad-hoc `warnings.warn` might be our best bet here.  Given that `_mutuallyExclusiveArguments` lives there, arguably it should be souped up to indicate which arguments are deprecated, but you don't need to do anything that general to land this :)

src/twisted/conch/ssh/keys.py
```diff
@@ -1201,25 +1208,194 @@ def toString(self, type, extra=None):
...
+            if self.isPublic():
+                comment = extra
+            else:
+                passphrase = extra
+        if isinstance(comment, unicode):
+            comment = comment.encode("utf-8")
+        if isinstance(passphrase, unicode):
+            passphrase = passphrase.encode("utf-8")
```
glyph, line 1252: hmm I guess this isn't new but we should really be normalizing any passphrases that pass through here…

src/twisted/conch/ssh/keys.py
glyph: This should also update internal callers — particularly, `ckeygen` — so we don't trip over our own deprecation.

glyph (changes requested):
I have a small amount of requests for additional changes here, but it's mostly pretty good.  The main ones are:

- please actually emit a deprecation warning so at least people who are trying really hard to have their automated tests tell them about this actually have it tell them about this

- please update `ckeygen` to have an option that uses this format; mainly so it's easy to test interactively, but also, like, presumably this is the main way `toString` is gonna get invoked?

Other than that I think we're good to go!

src/twisted/conch/ssh/keys.py
cjwatson: Both done.

src/twisted/conch/ssh/keys.py
cjwatson: Can you explain what you mean by "normalize" in this context?

cjwatson:
Aside from one question, I've addressed your feedback.  `--private-key-subtype=v1` isn't the best option name I've ever come up with, but OpenSSH offers no useful inspiration (it calls its equivalent `-o`), and I don't expect this option will often need to be used aside from testing.

src/twisted/conch/ssh/keys.py
glyph: I mean like NFKC/NFKD normalization, as per https://pages.nist.gov/800-63-3/sp800-63b.html#memsecretver .

It's fine to leave this for a follow-up though, since as I noted it's not new.

glyph (approved):
L G T M.

Will merge shortly.

src/twisted/conch/ssh/keys.py
cjwatson: Thanks.  I've followed up on this in https://github.com/twisted/twisted/pull/1201.

PR #1199, opened 2019-11-18
Depend on attrs ≥19.2.0; use eq instead of cmp

twm:
Set the minimal dependency to attrs 19.2.0, which provides `eq` and deprecates `cmp`. 
See the [attrs changelog](https://www.attrs.org/en/stable/changelog.html#id2).

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9710
* [x] I have updated the automated tests. (N/A)


src/twisted/internet/address.py
```diff
@@ -98,7 +98,7 @@ class HostnameAddress(object):
 
 
 
-@attr.s(hash=False, repr=False, cmp=False)
+@attr.s(hash=False, repr=False, eq=False)
```
moshez, line 101: I was wondering what `order` was set to. I see the default is "same as eq", so it is doing the right thing, but I wonder if it's worthwhile calling it out explicitly.

moshez (approved):
Thank you for working on this! Always nice to be ahead of deprecations!

One minor non-blocking question, otherwise seems good to me.

src/twisted/internet/address.py
twm: The attrs defaults are perfectly sensible now that `cmp` has been split, so I don't think that's necessary.

twm:
Thank you for the review, @moshez! Landing now.

PR #1201, opened 2019-11-24
Normalize Unicode passphrases for SSH keys

cjwatson:
This is required by NIST 800-63B.  We choose NFKC normalization here.

It is possible, though unlikely, that some existing key files will have
passphrases that are the UTF-8 encoding of a Unicode string not in NFKC;
twisted.conch.ssh.keys.Key.fromString will fail to load those keys if
given such a passphrase as a Unicode string.  In that case the
workaround is to provide the passphrase as bytes instead.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9736


src/twisted/conch/ssh/keys.py
```diff
@@ -108,6 +109,46 @@ class FingerprintFormats(Names):
...
 
+class PassphraseNormalizationError(Exception):
+    """
+    Raised when a passphrase contains Unicode characters that cannot be
+    normalized using the available Unicode character database.
+    """
+
+
+
+def _normalizePassphrase(passphrase):
```
twm, line 119: This seems generally useful; perhaps it should be public?

twm (approved):
This looks reasonable to me. I had one tentative comment, but I am happy to merge without that if you prefer. Thanks for all these Conch contributions!

twm:
@cjwatson I'll go ahead and merge this. Thanks for contributing to Twisted!

PR #1204, opened 2019-11-27
fix exception from multiple calls to FileBodyProducer.stop

jlitzingerdev:
The current specification for IBodyProducer states that
implementations need to be written to allow simultaneous calls to
pauseProducing, resumeProducing, and stopProducing.  FileBodyProducer
currently only complies with that for pause and resumeProducing (where
comply is interpreted here as not raising an error).

Instead of allowing the TaskFinished exception to be raised on
multiple calls to stopProducing, catch and drop the exception.  The
use of "pass" is not ideal, but this avoids an internal book-keeping
variable to track the stop state and a branch for the normal flow.
The use of "pass" to handle stopped also appears within the inner
maybeStopped function of startProducing, so it isn't without precedent.

This has been discussed in several forums:
* https://github.com/twisted/treq/issues/185#issuecomment-555596421
* https://twistedmatrix.com/pipermail/twisted-python/2017-November/031730.html

Related ticket(s): 7457

Regression risk: Existing uses of FileBodyProducer that rely on
multiple calls to stopProducing to raise an exception will break.

## Contributor Checklist:

* [x] The associated ticket in Trac is here:https://twistedmatrix.com/trac/ticket/6528


jlitzingerdev:
@glyph thank you for merging trunk, apologies I didn't get around to it until now.

src/twisted/web/client.py
```diff
@@ -1158,7 +1158,10 @@ def stopProducing(self):
         stopping the underlying L{CooperativeTask}.
         """
         self._inputFile.close()
-        self._task.stop()
+        try:
+            self._task.stop()
+        except task.TaskFinished:
```
glyph, line 1165: My preference would be for a fix that properly tracked the state of `self._task` and didn't provoke this exception in the first place.

It seems to me that the right place to close `_inputFile` would be in `maybeStopped` (and also the currently no-op callback there), which is *already* checking for `TaskStopped`; it's just not doing anything with it.

Don't quite have time for a full review right now though, and I might just land this as-is to quash the error.  But if you'd like to do something that actually tracked the completed state of `self._task` so that `stopProducing` could do the right thing I know I'd be happier about it :)

jlitzingerdev: Thanks for the feedback!

I'm certainly not opposed to it, and to honest, one goal in proposing this solution (the simplest thing that I felt _could_ work) was to elicit opinions from those who know the code.  

It'll take me a bit (doesn't seem like there's a huge rush, and I'm not anxious to land it for landing its sake), but what I'm thinking is I'll leave this as-is, for reference, then do another PR with an alternative in line with your suggestion.  I'll reference the PRs to each other in the ticket as well as on Github.  Then at least there will be a couple options.

I'm not opposed to doing it to this PR, but I thought it might be helpful to show the genesis of the next solution.

glyph: Sounds good to me!  Hopefully one day I won't be exhausted and sick all the time and I can confidently stamp this :)

jlitzingerdev: No doubt, best wishes for health and energy!

PR #1207, opened 2019-12-16
[9678] Fix HTTP multipart parsing of chunked requests on Py3.7+

twm:
See a562242 for a detailed explanation. This is the simplest possible fix. I do think that we should probably ditch the `cgi` module for multipart parsing eventually because it is very `str`-oriented, but we want `bytes`.

This includes the tests from #1180. Thank you @msdemlei!

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9678


glyph (approved):
Thanks for updating this; the fix looks correct to me.

PR #1208, opened 2019-12-17
fix MemoryReactor import in test_runner to be from twisted.internet.t…

threexc:
https://twistedmatrix.com/trac/ticket/9746


twm (changes requested):
Hi @threexc,

Thanks for this! The code change looks good to me. However, I don't think that the [news file](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles) is quite right. This isn't a user-facing "removal" &mdash; we don't need to notify users [like this](https://github.com/twisted/twisted/blob/trunk/NEWS.rst#deprecations-and-removals). I think that it's a bugfix or, since it only appears to a developer running Twisted's own test suite, a "misc" change that doesn't need to appear in the changelog.

Please adjust the news file and resubmit for review.

threexc:
Thanks @twm . I've pushed it again to the working branch on my fork as a misc change.

twm (approved):
Thanks so much, @threexc! I'll try to get this in today.

PR #1212, opened 2019-12-30
[9496] Names SecondaryAuthority traceback

twm:
There is a lot of string type confusion in `twisted.names`. This is mostly about how domain names (and labels) are represented.

I'll be using the Python 3 names for types here: `bytes` and `str`.

The `twisted.internet.interfaces.IResolver` interface defines a bunch of methods, about one per DNS record type. These all take the same arguments, most importantly `name`, but the types aren't consistent — the `lookupAddress()` method defines `name` as `bytes`, but every other method (including `lookupIPV6Address()` documents it as `str`. Presumably this is a Python 3 porting oversight, but the confusion goes deeper.

Since `IResolver` has so many methods all of the implementers in Twisted subclass `twisted.names.common.ResolverBase`, which maps all the methods to a `_lookup()` bottleneck method that can be easily overridden. This includes `lookupAddress()`, so clearly its type signature is supposed to align with the rest.

Internally the query is usually translated into a `dns.Query` instance, which represents the name as a `dns.Name` instance. `dns.Name` coerces the name to `bytes`, encoding Unicode using the `idna` codec. So most implementations simply deal with bytes.

However, some implementers override a few methods directly. For example, `twisted.names.hosts.Resolver` only deals in IP addresses, so it overrides `lookupAddress()` and `lookupIPV6Address()` directly. This produces inconsistent behavior: the hosts implementation _only_ accepts `bytes`, not `str`.

All of this confusion is borne out in the tests: some pass `name` as `str`, a few as `bytes`. There is no systematic coverage of both types.

This is basically what's going on in `twisted.names.secondary.SecondaryAuthority` in Twisted #9496: it's not coercing the type of its `domain` parameter to `bytes`, so it doesn't match the query.

The design intent here seems to be that everything is `bytes`. However, the reality is that most things support both `bytes` and `str`, so code relies on that. I tried rejecting `str`, but stuff broke (and that wouldn't be backwards compatible anyway. I have therefore:

1. Updated `IResolver` to permit "`bytes` and `str`" for _name_ parameters. We should not mislead third-party implementers.
2. Added a public function, `twisted.names.dns.domainString()`, to do the coercion in a consistent way. It's public so third-party `IResolver` implementations can use it.
3. Added coercion to all of the `common.BaseResolver` wrapper methods, and document `_lookup()`'s type to be _only_ `bytes`. This makes it easier for third-party implementers.
4. Updated miscellaneous `IResolver` implementers to also do coercion using `domainString()`.
5. Use the `dns.Name.name` attribute instead of calling `str()` to coerce `dns.Name` to a string, the former being `bytes`.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9496


src/twisted/names/authority.py
```diff
@@ -26,6 +26,9 @@ def getSerial(filename='/tmp/twisted-names.serial'):
     State is stored in the given file.  If it does not exist, it is
     created with rw-/---/--- permissions.
 
+    This manipulates process-global state by calling C{os.umask()}, so it isn't
```
glyph, line 29: Thanks for adding this note.

src/twisted/names/authority.py
```diff
@@ -60,6 +63,9 @@ class FileAuthority(common.ResolverBase):
     """
     An Authority that is loaded from a file.
 
+    This is an abstract class that implements record search logic. To create
+    a functional resolver, subclass it and override the L{loadFile()} method.
```
glyph: (I don't think the parens are necessary for an epytext reference.)

src/twisted/names/authority.py
```diff
@@ -86,6 +95,18 @@ def __setstate__(self, state):
         self.__dict__ = state
 
 
+    def loadFile(self, filename):
+        """
+        Load DNS records from a file.
+
+        This method populates the I{soa} and I{records} attributes. It must be
+        overridden in a subclass. It is called once from the initializer.
```
glyph, line 103: Do you want to add a `raise NotImplementedError` just to make this clearer?

src/twisted/names/authority.py
```diff
@@ -86,6 +95,18 @@ def __setstate__(self, state):
...
+
+        @param filename: The I{filename} parameter that was passed to the
+        initilizer.
+        """
```
glyph, line 109: an explicit `@return` might also be nice.

src/twisted/names/common.py
```diff
@@ -192,8 +195,35 @@ def _cbRecords(self, records, name, effort):
...
 
 def extractRecord(resolver, name, answers, level=10):
+    """
+    Resolve a name to an IP address, following I{CNAME} records and I{NS}
+    referrals recursively.
+
+    This is an implementation detail of L{ResolverBase.getHostByName}.
+
+    @param resolver: The resolver to use for the next query (unless handling
+    an I{NS} referral).
```
glyph, line 206: `I{}` or `C{}` ?

src/twisted/names/common.py
```diff
@@ -192,8 +195,35 @@ def _cbRecords(self, records, name, effort):
...
+    recursion. The query returns L{None} when it reaches 0.
+    @type level: L{int}
+
+    @returns: The first IPv4 or IPv6 address (as a dotted quad or colon
+    quibbles), or L{None} when no result is found.
+    @rtype: native L{str} or L{None}
+    """
     if not level:
         return None
+    # FIXME: twisted.python.compat monkeypatches this if missing, so this
```
glyph, line 226: Can you file a ticket and link to it rather than just a FIXME comment?

src/twisted/names/common.py
```diff
@@ -83,103 +83,106 @@ def _lookup(self, name, cls, type, timeout):
...
 
 
     # IResolverSimple
     def getHostByName(self, name, timeout=None, effort=10):
+        name = dns.domainString(name)
         # XXX - respect timeout
-        return self.lookupAllRecords(name, timeout
-            ).addCallback(self._cbRecords, name, effort
-            )
+        # XXX - this should do A and AAAA lookups, not ANY (see RFC 8482).
```
glyph, line 182: again, file another ticket, link to it here?

src/twisted/names/common.py
```diff
@@ -210,20 +240,23 @@ def extractRecord(resolver, name, answers, level=10):
...
-            r = client.Resolver(servers=[(str(r.payload.name), dns.PORT)])
-            return r.lookupAddress(str(name)
-                ).addCallback(
-                    lambda records: extractRecord(
-                        r, name,
-                        records[_ANS] + records[_AUTH] + records[_ADD],
-                        level - 1))
+        if r.type != dns.NS:
+            continue
+        from twisted.names import client
```
glyph, line 251: I know the code isn't new, but I should probably point out that doing the local import *inside* a hot loop is probably not a great idea…

src/twisted/names/dns.py
```diff
@@ -246,6 +249,38 @@ def _nameToLabels(name):
...
+    the strings it returns.
+
+    @param domain: A domain name.  If passed as a text string it will be
+        C{idna} encoded.
+    @type domain: L{bytes} or L{str}
+
+    @returns: L{bytes} suitable for network transmission.
+    @rtype: L{bytes}
+
+    @since: Twisted NEXT
```
glyph, line 272: 👍 thank you for remembering the compat marker and the right syntax for it :)

src/twisted/names/dns.py
```diff
@@ -246,6 +249,38 @@ def _nameToLabels(name):
...
+    """
+    if isinstance(domain, unicode):
+        domain = domain.encode('idna')
```
glyph, line 275: You might want to use `twisted/internet/_idna.py` here rather than a straight `.encode`.  (The python stdlib codec is a much older version of the spec.)

src/twisted/names/newsfragments/9496.bugfix
```diff
@@ -0,0 +1 @@
+twisted.names.secondary.SecondaryAuthority now accepts str for its domain parameter, so twist dns --secondary now functions on Python 3.
```
glyph, line 1: Love this explanation.

src/twisted/names/secondary.py
```diff
@@ -128,20 +149,26 @@ def fromServerAddressAndDomain(cls, serverAddress, domain):
...
 
     def transfer(self):
-        if self.transferring:
+        """
+        Attempt a zone transfer.
+
+        @returns: A L{Deferred} that fires with L{None} when attempted zone
+            transfer has completed.
+        """
+        if self.transferring:  # <-- never true
```
glyph, line 171: File a bug for this, too?

src/twisted/names/secondary.py
```diff
@@ -128,20 +149,26 @@ def fromServerAddressAndDomain(cls, serverAddress, domain):
...
             return
```
glyph, line 172: This seems like it would make the docstring a lie, if it actually worked 😬

glyph (approved):
This looks great. I have some minor feedback inline that you can respond to, but please land when you have addressed to your satisfaction. Thanks for fixing it!

src/twisted/names/authority.py
twm: I considered that, but didn't think that would be backward-compatible. Right now a you can subclass `FileAuthority`, ignore `loadFile()`, and populate `soa` and `records` in your `__init__` instead.

Or something even weirder if you like!

https://github.com/twisted/twisted/blob/2d584b952bb6392a7002d1508b5146170cce1792/src/twisted/names/test/test_names.py#L38-L42

src/twisted/names/common.py
twm: `I{}` is used in most of the places record types are given any syntax, so I stuck with that.

src/twisted/names/dns.py
twm: Sure, will do.

src/twisted/names/secondary.py
twm: I'm afraid that :grimacing: summarizes this module too well. :(

twm (commented):
I've done what I can for the moment. Filing tickets will have to wait until dornkirk is back up.

twm:
Hmm, so switching to `_idnaBytes` results in test failures. This one in a test I added looks meritorious:

```
[ERROR]
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/test/test_dns.py", line 70, in test_idna
    self.assertEqual(b'xn--fwg.test', dns.domainString(u'\u203D.test'))
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/dns.py", line 280, in domainString
    domain = _idnaBytes(domain)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/internet/_idna.py", line 30, in _idnaBytes
    return idna.encode(text)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 358, in encode
    s = alabel(label)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 281, in alabel
    check_label(label)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 261, in check_label
    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
idna.core.InvalidCodepoint: Codepoint U+203D at position 1 of '‽' not allowed

twisted.names.test.test_dns.DomainStringTests.test_idna
```

However most of the failures are to do with SRV support:

```
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/test/test_srvconnect.py", line 118, in test_SRVPresent
    self.connector.connect()
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/srvconnect.py", line 109, in connect
    nativeString(self.domain)),
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/client.py", line 731, in lookupService
    return getResolver().lookupService(name, timeout)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/common.py", line 138, in lookupService
    return self._lookup(dns.domainString(name), dns.IN, dns.SRV, timeout)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/names/dns.py", line 280, in domainString
    domain = _idnaBytes(domain)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/twisted/internet/_idna.py", line 30, in _idnaBytes
    return idna.encode(text)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 358, in encode
    s = alabel(label)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 270, in alabel
    ulabel(label)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 304, in ulabel
    check_label(label)
  File "/home/twm/twisted/build/py36-alldeps-nocov/lib/python3.6/site-packages/idna/core.py", line 261, in check_label
    raise InvalidCodepoint('Codepoint {0} at position {1} of {2} not allowed'.format(_unot(cp_value), pos+1, repr(label)))
idna.core.InvalidCodepoint: Codepoint U+005F at position 1 of '_xmpp-server' not allowed

twisted.names.test.test_srvconnect.SRVConnectorTests.test_SRVPresent
```

This looks like https://github.com/kjd/idna/issues/17, which indicates that this is a little more complex than it seemed to me at first:

* IDNA encoding applies only to hostnames
* Hostnames _must not_ contain underscores
* DNS names _may_ contain underscores, e.g. [as a prefix](https://tools.ietf.org/html/rfc8553) like in SRV

Sadly, it looks like the name handling behavior we want actually varies by record type. We can't apply the `idna` package blindly (as `dns.Name` did with `str.encode('idna')`), since it actually validates its inputs as hostnames.

I think that I will back out this change and file a ticket once dornkirk returns.

twm:
Some quality time with [RFC 5890](https://tools.ietf.org/html/rfc5890) has convinced me that handling IDNA is entirely inappropriate at this layer. (Indeed, a more technically correct representation of DNS names would be `Tuple[bytes]`, as the dots are a convention of representation, though I doubt the practical utility of that.) I have backed out the use of `_idnaBytes`. We'll stick with the still incorrect &mdash; but compatible &mdash; use of IDNA2003. I'll go ahead and merge when the build is green.

twm:
The build is broken because of https://github.com/hamcrest/PyHamcrest/issues/131

twm:
Now the coverage service is failing. :angry:

glyph:
@twm I hopefully have addressed the thing that was breaking coverage and kicked the build via the trunk merge, so hopefully you can land when it completes.

twm:
Kicking build due to https://twistedmatrix.com/trac/ticket/8879#comment:36

twm:
Thank you @glyph!

PR #1215, opened 2020-01-08
[6647] IPv6 link-local address support for datagramProtocol

### LINKED ISSUE #6647: Accept embedded scope ids on IPv6 addresses when calling listenUDP
|[habnabit's avatar](https://github.com/habnabit)| @habnabit reported|
|-|-|
|Trac ID|trac#6647|
|Type|enhancement|
|Created|2013-07-27 22:21:21Z|

As a followup to [#5086](https://github.com/twisted/twisted/issues/5086), Twisted should provide support for listening on an [IPv6](https://twistedmatrix.com/trac/wiki/IPv6) address including a scope id. This allows listening on a link-local address for a particular interface.

Once this is resolved, Twisted servers should be able to bind UDP ports to IPv6 addresses such as `fe80::1%lo0`, as well as send datagrams to such IPv6 addresses. Where the platform requires the scope id, `IReactorUDP.listenUDP` and `IUDPTransport.write` will require the scope id as well.

```
trac-id__6647 6647
type__enhancement enhancement
reporter__habnabit habnabit
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__ipv6 ipv6
time__1374963681000000 1374963681000000
changetime__1581059522011957 1581059522011957
version__None None
owner__mhoeft mhoeft

```
</details>

mhoeft:
Fixes #6647


twm:
Hi @mhoeft,

I merged forward to clear some spurious test failures. I'm not able to do a proper review right now, but I will note that any review will ask for a test (or a justification as to why it's not possible to provide one).

mhoeft:
Hi @twm,

Thank you for your advices. Testing the proposed enhancement is not so easy - it is highly related to a particular platform specification (a link-local address generation method configured in OS and MAC address of a network interface).  I looked for detail specification of testing platforms (or a way to determine it) on the wiki but I can’t get it. 

Nevertheless, for Linux systems I would prepare the appropriate test using system tools (called via subprocess.Popen) adding a dummy network interface with determined MAC address. Is such approach consistent with your testing policy?

I also noticed errors in testing and travis logs, I will correct my code.

src/twisted/internet/udp.py
```diff
@@ -193,7 +193,11 @@ def _bindSocket(self):
             # Create a new socket and make it listen
             try:
                 skt = self.createInternetSocket()
-                skt.bind((self.interface, self.port))
+                if self.addressFamily == socket.AF_INET6:
+                    addr = socket.getaddrinfo(self.interface, self.port, 0, 0, 0, socket.AI_NUMERICHOST )[0][4]
```
glyph: To fully avoid any I/O, you want to use `AI_NUMERICSERV` as well… i.e. what `_resolveIPv6` does in `tcp.py`.  So I think you want to maybe refactor `_resolveIPv6` into something in maybe `twisted.internet.address` or something, and call it from here rather than calling `getaddrinfo` directly, since `[4]` is also a bit of a silly thing to re-encode here.

glyph (changes requested):
Hi @mhoeft !  Thanks so much for your contribution to Twisted!  It would be great to get this bug fixed.

As @twm suggested, any full review is going to ask for a unit test.  You can write a very simple / minimally scoped test here, one that fakes out the result of `createInternetSocket`, maybe even `getaddrinfo` too, and just asserts that it gets passed a thing to `.bind` that was returned from `getaddrinfo`; you don't need a full-stack test that verifies that link-local addresses "really work", since that's really tricky and local configuration dependent.

But some kind of test will be needed to land this.

PR #1217, opened 2020-01-09
run atexit handlers when cleanly shutting down twistd Fixes #9755 and #9945

### LINKED ISSUE #9755: cleanly shutting down a twistd application with SIGINT does not run atexit handlers
|[graingert's avatar](https://github.com/graingert)| @graingert reported|
|-|-|
|Trac ID|trac#9755|
|Type|enhancement|
|Created|2020-01-08 12:44:32Z|

https://github.com/twisted/twisted/blob/eecde4e931d77e035f2ceb8c669a563765afcc1a/docs/core/man/twistd.1#L107

I expect a clean shutdown to run atexit handlers, in my usecase pytest-cov's exit handler

https://github.com/twisted/twisted/pull/1217

```
trac-id__9755 9755
type__enhancement enhancement
reporter__graingert graingert
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1578487472185537 1578487472185537
changetime__1600356852114208 1600356852114208
version__None None
owner__None None

```
</details>

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here:  https://tm.tl/9755

### DIFF
--- a/src/twisted/application/app.py
+++ b/src/twisted/application/app.py
@@ -708,5 +708,7 @@ def _exitWithSignal(sig):
     @type sig:  C{int}
     """
     signal.signal(sig, signal.SIG_DFL)
-    atexit._run_exitfuncs()
-    os.kill(os.getpid(), sig)
+    try:
+        atexit._run_exitfuncs()
+    finally:
+        os.kill(os.getpid(), sig)
--- a/src/twisted/application/app.py
+++ b/src/twisted/application/app.py
@@ -3,6 +3,7 @@
 # See LICENSE for details.
 
 
+import atexit
 import sys
 import os
 import pdb
@@ -707,4 +708,5 @@ def _exitWithSignal(sig):
     @type sig:  C{int}
     """
     signal.signal(sig, signal.SIG_DFL)
+    atexit._run_exitfuncs()
     os.kill(os.getpid(), sig)
--- a/src/twisted/application/newsfragments/9755.bugfix
+++ b/src/twisted/application/newsfragments/9755.bugfix
@@ -1,0 +1,1 @@
+run atexit handlers when cleanly shutting down twistd
--- a/src/twisted/scripts/test/test_scripts.py
+++ b/src/twisted/scripts/test/test_scripts.py
@@ -7,20 +7,21 @@ Tests for the command-line scripts in the top-level I{bin/} directory.
 Tests for actual functionality belong elsewhere, written in a way that doesn't
 involve launching child processes.
 """
+import json
+import signal
+import subprocess
+import sys
+import textwrap
 
-from os import devnull, getcwd, chdir
-from sys import executable
-from subprocess import PIPE, Popen
-
-from twisted.trial.unittest import SkipTest, TestCase
-from twisted.python.modules import getModule
+from twisted.trial.unittest import TestCase
 from twisted.python.filepath import FilePath
 from twisted.python.test.test_shellcomp import ZshScriptTestMixin
+from twisted.scripts import htmlizer, trial, twistd
 
 
-def outputFromPythonScript(script, *args):
+def outputFromPythonModule(module, *args, check=True, **kwargs):
     """
-    Synchronously run a Python script, with the same Python interpreter that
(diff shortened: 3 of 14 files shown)


graingert:
why did this decrease the coverage so much?

glyph:
Probably an intermittent failure on coveralls' side :(

glyph:
(Force pushing a new commit to cause a new test run *should* fix the issue.)

src/twisted/application/newsfragments/9755.bugfix
```diff
@@ -0,0 +1 @@
+run atexit handlers when cleanly shutting down twistd
```
graingert, line 1:
```suggestion
Run atexit handlers when cleanly shutting down twistd
```

src/twisted/application/app.py
```diff
@@ -705,4 +706,5 @@ def _exitWithSignal(sig):
     @type sig:  C{int}
     """
     signal.signal(sig, signal.SIG_DFL)
+    atexit._run_exitfuncs()
```
glyph: This function is internal and undocumented at least as far as the official reference goes.  It seems like the better strategy might be to make your application register a shutdown function with the reactor separately from exitfuncs?

One thing I would be looking for from the documentation here: what are the exception-handling characteristics of this function? Anything we need to catch to make sure `_exitWithSignal` still equally reliably exits with a signal?

graingert: If an exit handler is registered that raises a new exception then python is expected to return a different exit code

glyph: Only in Python 2: https://bugs.python.org/issue27035

glyph (changes requested):
Can you add some test coverage for this?

src/twisted/scripts/test/test_scripts.py
```diff
@@ -66,75 +67,109 @@ def scriptTest(self, name):
...
+                import sys
+
+                print(json.dumps(sys.path))
+                """
+            ).encode()
+        )
+        output = json.loads(
+            outputFromPythonModule(twistd, '-ny', 'bar.tac', cwd=testDir.path, check=False)
+        )
+        self.assertIn(testDir.path, output)
```
graingert: previously this was happy if the testDir showed up in a stacktrace

src/twisted/scripts/test/test_scripts.py
```diff
@@ -7,134 +7,153 @@
...
-
-        @param name: A path fragment, relative to the I{bin} directory of a
-            Twisted source checkout, identifying a script to test.
-        @type name: C{str}
-
-        @raise SkipTest: if the script is not where it is expected to be.
-        """
-        script = self.bin.preauthChild(name)
-        if not script.exists():
-            raise SkipTest(
```
graingert: these tests were always skipped!

src/twisted/scripts/twistd.py
```diff
@@ -32,4 +33,14 @@ def run():
     app.run(runApp, ServerOptions)
 
 
+def set_path_then_run():
```
graingert, line 40: this is a new name because twisted.scripts.twistd.run never messsed with the sys.path

src/twisted/application/app.py
```diff
@@ -702,4 +703,7 @@ def _exitWithSignal(sig):
     @type sig:  C{int}
     """
```
graingert, line 709: Python 3.8 supports settings the exit code to -2 and running all the finalization normally with:

```suggestion
    """
    if sig == signal.SIGINT and sys.version_info > (3, 8):
        raise KeyboardInterrupt
```

but it's broken https://bugs.python.org/issue41602

graingert: @altendky https://bugs.python.org/issue41602 is fixed upstream now and waiting for backport

glyph:
@graingert I'm sorry that this has gotten to such a conflicted state in the interim :( can you address conflicts and then perhaps gesture at the robot ("please r e v i e w") when it merges clean again?

PR #1219, opened 2020-01-13
CramMD5ClientAuthenticator now specifies the digestmod argument to hmac.HMAC constructor explicitly.

mithodin:
The e-mail CramMD5ClientAuthenticator class now calls the hmac.HMAC constructor with the digestmod argument, as python 3.8 requires it to be specified explicitly.

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/9733

### DIFF
--- a/src/twisted/mail/newsfragments/9733.bugfix
+++ b/src/twisted/mail/newsfragments/9733.bugfix
@@ -1,1 +1,0 @@
-twisted.mail._cred.CramMD5ClientAuthenticator.challengeResponse(...) now calls hmac.HMAC.new(...) with explicit MD5 digestmod.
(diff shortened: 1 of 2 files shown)


src/twisted/mail/newsfragments/9733.bugfix
```diff
@@ -0,0 +1 @@
+twisted.mail._cred.CramMD5ClientAuthenticator.challengeResponse(...) now calls hmac.HMAC.new(...) with explicit MD5 digestmod.
```
glyph: 1. I would probably call this one a `misc` because it's about the internal implementation working on python 3.8, not something that's really interesting to users reviewing the release notes; the top-line item there is about a supported platform, not this specific API.
2. News fragments should always refer to public API names of things.  This one appears to be illustrating a bug in our process, because it *should* be showing up as public in `smtp` and `imap4` modules via `__all__` re-exports.  I filed this as a bug here: https://github.com/twisted/pydoctor/issues/184 to try to track it down.

glyph (approved):
Thank you for this contribution @mithodin !

I'll `misc-`ify the news fragment and land over the objections of the (broken) Azure Pipelines CI, which seems to be broken on every other branch as well — but curiously not `trunk`.

PR #1225, opened 2020-03-03
[9740] ESMTPSender: dont't force TLSv1.0 by default

mmilata:
Hello, I've also been bitten by matrix-org/synapse#6211 and would like to see if the fix is acceptable for inclusion in Twisted.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9740
  * There doesn't seem to be a test that tests for specific TLS version but it's possible I haven't been looking hard enough.


glyph (changes requested):
Hi Martin!

Thanks so much for your contribution to Twisted.

The correct fix, here, if you want proper TLS security on `twisted.mail` connections, would be to drop `ClientContextFactory` entirely, and use `optionsForClientTLS`.  Dropping TLS 1.0 as a protocol version is good, but security-wise this change is almost meaningless, as `ssl.ClientContextFactory` doesn't verify certificates, or hostnames, or really provide ''any'' proper security properties.  (Separate to this change, it should be deprecated and removed from Twisted entirely.)

Dropping in `optionsForClientTLS` here should not be much harder that just fixing the protocol version, and it would have the added benefit that it would use a whole slew of ''other'' better security attributes as well.  Would you mind modifying your patch to do that?

Thanks!

mmilata:
@glyph thanks for quick review. I tried to replace the `ClientContextFactory` ctor call with `optionsForClientTLS`, however it requires the server hostname which I'm not quite sure how to pass to it without breaking the public interface. PTAL

Anyway the TLS negotiation goes further now before rejecting my server's self-signed cert:laughing:. Oh well, probably a good time to replace it with LetsEncrypt one.

mmilata:
Installed LE certificate & now matrix-synapse can send emails through postfix configured to only use TLS1.1 or higher with this patch.

src/twisted/mail/smtp.py
```diff
@@ -2010,6 +2010,11 @@ def __init__(self, username, secret, contextFactory=None, *args, **kw):
         self.heloFallback = 0
         self.username = username
 
+        self._hostname = None
```
glyph: This block here can be abbreviated to `self._hostname = kw.pop('hostname', None)`, saving a few lines of code :)

src/twisted/mail/smtp.py
```diff
@@ -2028,14 +2033,15 @@ def _registerAuthenticators(self):
     def _getContextFactory(self):
         if self.context is not None:
             return self.context
+        if self._hostname is None:
+            return None
         try:
-            from twisted.internet import ssl
+            from twisted.internet._sslverify import optionsForClientTLS
```
glyph: This is a public name, exposed as  [`twisted.internet.ssl.optionsForClientTLS`](https://twistedmatrix.com/documents/19.10.0/api/twisted.internet.ssl.html#optionsForClientTLS), so please import it from there.  (Maybe Twisted should use [publication](https://pypi.org/project/publication/) to avoid this problem...)

src/twisted/mail/smtp.py
```diff
@@ -2179,9 +2189,22 @@ def cancel(d):
     if isinstance(password, unicode):
         password = password.encode("utf-8")
 
-    factory = ESMTPSenderFactory(username, password, from_addr, to_addrs, msg,
-        d, heloFallback=True, requireAuthentication=requireAuthentication,
-        requireTransportSecurity=requireTransportSecurity)
+    tlsHostname = smtphost
+    if not isinstance(tlsHostname, unicode):
+        tlsHostname = tlsHostname.decode("ascii")
```
glyph: This should be IDNA, not `"ascii"`.  See the internal `twisted.internet._idna` module for the shared utility for doing this.

src/twisted/mail/smtp.py
```diff
@@ -2028,14 +2033,15 @@ def _registerAuthenticators(self):
...
+            from twisted.internet._sslverify import optionsForClientTLS
         except ImportError:
             return None
         else:
             try:
-                context = ssl.ClientContextFactory()
-                context.method = ssl.SSL.TLSv1_METHOD
+                context = optionsForClientTLS(self._hostname)
                 return context
             except AttributeError:
```
glyph: I think this `except AttributeError` is now unnecessary.  Can you remove the whole `try` block around this code and see if any tests fail?

glyph:
Thanks again for your contribution!  I think the compatibility break is acceptable given the security needs here.  Please do re-add the `review` keyword so we notice it on twistedmatrix.com when you've addressed feedback, so we find it again.

rodrigc (approved):
I think @mmilata has addressed the review feedback, and this PR now looks OK.

mmilata:
Thanks @glyph & @rodrigc!

rodrigc:
@mmilata thank you for contributing to Twisted!

Neustradamus:
@rodrigc: Can you release a new version with this fix?

Several projects like Matrix wait you...
cc @clokep, @Half-Shot

@mmilata: Thanks!

Ticket: https://github.com/matrix-org/synapse/issues/6211

rodrigc:
@Neustradamus I am not the release manager for Twisted, so cannot release a new version.  @hawkowl is the release manager for Twisted.

Neustradamus:
@rodrigc: Thanks for your reply! :)

It is really important because a lot of people/servers have this problem...

@hawkowl: Thanks a lot in advance for a new release with this important fix.

Neustradamus:
People ask news about this security fix:
- https://twitter.com/CaptainKurtis/status/1292278623110660096

Neustradamus:
Dear @gjabell, @madpsy, @fadenb, @neilisfragile, @thereapman, @sbiberhofer, @henry-nicolas, @plamenh, @richvdh, @vmario89, @Chatcloud, @MisterAlainDev, @schwukas, @kaiyou, @babolivier, @mmilata, @clokep, @n3m3s1s, @Half-Shot, @exarkun, @anoadragon453: A good news, there are some "20.11.0" test-builds here:
- https://github.com/twisted/twisted/releases done by @rodrigc.

Time to test :)

cc: @glyph.

richvdh:
@Neustradamus you've already been asked not to mention the entire world over at https://github.com/matrix-org/synapse/issues/6211. If people are interested they will follow the issue themselves. Don't be obnoxious.

Neustradamus:
Good news the 21.02.0 RC1 is out!
- https://github.com/twisted/twisted/releases done by @rodrigc.

Neustradamus:
Good news, the 21.02.0 stable version is out!
- https://mail.python.org/archives/list/python-announce-list@python.org/thread/HHAY6XA5YRAYBA7IUL5UELX74H7W2X3V/
- https://github.com/twisted/twisted/releases/tag/twisted-21.2.0

Thanks a lot to @mmilata for this very important contribution and @rodrigc, @glyph and @twisted (Twisted Matrix Labs) team.

PR #1227, opened 2020-03-09
6014 ipv6 host header literal support

glyph:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/6014


twm (approved):
The implementation here looks fine to me. I have a few minor docstring suggestions, though:

Naively, I might be a bit surprised to find that a method named `getRequestHostname()` can return an IP address, since an IP address [is not a hostname](https://tools.ietf.org/html/rfc1123#page-13). It would be nice if the docstring clarified that.

The docstring also doesn't mention that it elides any port passed in the `Host` header. If a user were constructing a URL based on this information that might be surprising. Perhaps it should mention that the port is dropped?

Thanks glyph!

PR #1228, opened 2020-03-13
[#9772] Release 20.3.0

hawkowl:
Trac link: https://twistedmatrix.com/trac/ticket/9772


glyph (commented):
1. Can you fix the lint error in `copyright.py` so we won't get a linter failure for every release branch in the future?
2. `No newsfragments should be on the release branch.` seems like a legit error. Did something go wrong with the towncrier invocation?

hawkowl:
@glyph 

1. done
2. This doesn't use towncrier! It uses its own script to check for it, which I've now updated and fixed.

glyph (approved):
✅

PR #1229, opened 2020-03-15
[9774] Remove t.web.client global reactor import

twm:
Replace a global reactor import with local imports to fix https://github.com/twisted/treq/issues/107.

I tested the import side effect manually like so:

```
$ build/py37-alldeps-nocov-posix/bin/python
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0] on linux
>>> import sys, twisted.web.client
>>> 'twisted.internet.reactor' in sys.modules
False
```

As this is a style guide thing, I don't think that it requires a specific automated test. Ideally we'd have a linter enforce the absence of this kind of import.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9774#ticket
* [x] I have updated the automated tests. (N/A)


glyph (approved):
I love what you're doing here, and I don't want to block it, but what do you think of avoiding the inner-imports in multiple places by adding a `_reactor` attribute that's initialized lazily, say, via a property?  This would make future maintenance easier.

twm:
Thank you for the review!

Normally that's certainly what I'd do, but the only future maintenance of `getPage` and friends I anticipate is deletion, so I am inclined to merge as-is.

PR #1232, opened 2020-03-20
Fix AsyncioSelectorReactor.callLater().reset() and fix memory leaks

IlyaSkriblovsky:
`AsyncioSelectorReactor.callLater()` returns `DelayedCall`s whose `reset()` method doesn't work as expected.

1. `reset()` to later time simply doesn't work: timer gets fired at original time

2. `reset()` to earlier works, but original internal asyncio's timer isn't cancelled and this causes an exception in `AsyncioSelectorReactor`'s internal code

3. This is not neccessarily a bug, but worth to be fixed too: current implementation of `AsyncioSelectorReactor.callLater()` leaves many cyclically linked objects after timer is fired. For the app that uses timers heavily this leads to significally higher memory usage (and cpu too due to much more often GC invocations)

Here the gist with runnable demos of all three issues: ​https://gist.github.com/IlyaSkriblovsky/58876c9f97c617875e7b739fd0447807

Here is the RAM usage % graph from my servers that run timer-hungry twisted application. I've tried to migrate them from epollreactor to asyncioreactor for two times, but was forced to rollback and investigate the issue due to unusual RAM usage.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9780


IlyaSkriblovsky:
`lint` CI pipeline was failed, but error message seems to be unrelated with changes from this PR

rodrigc:
@IlyaSkriblovsky lint failed here: https://travis-ci.org/github/twisted/twisted/jobs/667882022#L339 because of line too long.  Can you fix that?  That should eliminate the lint error.

wiml:
This looks reasonable to me, and the new code definitely seems cleaner! Two comments:

* It looks like `_DCHandle` isn't used anywhere else, it should probably be removed as well.

* Aside from fixing the bug, this also moves the management of the timer queue from the asyncio implementation to Twisted's implementation. This is probably fine, but it might be good to check if there is anything especially good about asyncio's implementation which we might lose by doing this.

IlyaSkriblovsky:
Fixed line length and removed `_DCHandle`, thanks for noticing.

> it might be good to check if there is anything especially good about asyncio's implementation

Asyncio's timer handling in [base_events.py](https://github.com/python/cpython/blob/b54a99d6432de93de85be2b42a63774f8b4581a0/Lib/asyncio/base_events.py#L1815) looks almost identical with Twisted's [internet/base.py](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L872).

The same heapq-based timer list ([asyncio](https://github.com/python/cpython/blob/b54a99d6432de93de85be2b42a63774f8b4581a0/Lib/asyncio/base_events.py#L720), [twisted](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L832)), the same calculation of the time untill nearest timer and using it as a timeout argument for `select()` ([asyncio](https://github.com/python/cpython/blob/b54a99d6432de93de85be2b42a63774f8b4581a0/Lib/asyncio/base_events.py#L1854), [twisted](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L1293)). Even minor details like clearing cancelled timers if the number of cancellations is greater than 50% of total scheduled timers are done identically ([asyncio](https://github.com/python/cpython/blob/b54a99d6432de93de85be2b42a63774f8b4581a0/Lib/asyncio/base_events.py#L1824), [twisted](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/base.py#L924)).

So I think it is pretty safe to use Twisted's timer management implementation instead of Asyncio's.

rodrigc:
One more small lint problem: https://travis-ci.com/github/twisted/twisted/jobs/317595892#L341

IlyaSkriblovsky:
Thanks, fixed

rodrigc:
I submitted this patch: https://github.com/twisted/twisted/pull/1248
so hopefully in the future, the errors from lint will not be so confusing.

wiml (approved):
This looks good to me!

PR #1237, opened 2020-03-27
Make AsyncioSelectorReactor.seconds() return an epoch time

richvdh:
Fixes https://twistedmatrix.com/trac/ticket/9787.


richvdh:
I wasn't at all sure how to test this; ensuring that we're somewhere in the 21st century seemed like the best I could think of.

(maybe we should patch out time.time somehow?)

twm:
Doesn't this mean that [`IDelayedCall.getTime()`](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IDelayedCall.html#getTime) will return a time with a different base than that returned by `reactor.seconds()`?

(Fun observation, the `DelayedCall` docstring actually says "epoch time"!)

hawkowl:
@twm nah, this just makes it do what the other reactors do.

twm:
Ah okay, I had it backwards. It looks like this breaks `DelayedCall.reset()` because the `reset()` function doesn't convert the epoch time back to monotonic time when calling [`loop.call_at()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_at), which expects a value that uses the event loop's timebase:

https://github.com/richvdh/twisted/blob/c7ce3e219055a369e0e4c5200894a0527b9b8830/src/twisted/internet/asyncioreactor.py#L297-L298

```py
        def reset(dc):
            dchandle.handle = self._asyncioEventloop.call_at(dc.time, run)
```

However reset is apparently pretty broken anyway: https://github.com/twisted/twisted/pull/1232

PR #1238, opened 2020-03-28
[#9784] Remove Python 2 from CI

hawkowl:
https://twistedmatrix.com/trac/ticket/9784

Since the config for Azure Pipelines is done in the web UI, I've gone ahead and done that already.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -6,7 +6,7 @@ Installation Requirements
 
 To install Twisted, you need:
 
-- Python 2.7 (full functionality) or 3.5/3.6/3.7 (subset of functionality).
+- Python 3.5/3.6/3.7.
 
 - `setuptools <https://pypi.python.org/pypi/setuptools>`_
   (installed automatically if you use pip).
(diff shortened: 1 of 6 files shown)


INSTALL.rst
```diff
@@ -6,7 +6,7 @@ Installation Requirements
 
 To install Twisted, you need:
 
-- Python 2.7 (full functionality) or 3.5/3.6/3.7 (subset of functionality).
```
rodrigc, line 9: I would just simplify this and put: Python 3.5 or higher
If we come across a Python version where Twisted doesn't work, we can update this later.

rodrigc (approved):
Minor tweak to the README but looks OK.

src/twisted/__init__.py
```diff
@@ -21,3 +22,9 @@
     "twisted",
     "news"
 )
+
+if sys.version_info < (3, 5):
```
twm, line 26: Please update the package metadata to match as well:

https://github.com/twisted/twisted/blob/34fd0bdd369ba238d7927a9d7356e14dee9ed01a/src/twisted/python/_setup.py#L75

I think that also having the runtime check is wise, since old Pips may not pick up on the metadata, but having the static check should reduce the number of folks bitten by this removal.

INSTALL.rst
hawkowl: Currently a whole bunch of tests fail on 3.8, so as per our policy we don't say we support it quite yet.

twm (commented):
Oh no, I forgot to submit this...

PR #1240, opened 2020-03-29
[9778] unregister processes before calling back to protocol's processEnded

altendky:
https://twistedmatrix.com/trac/ticket/9778


altendky:
Kick CI

altendky:
Kick it some more...

altendky:
Fun, Travis seems to be not reporting back today.

src/twisted/newsfragments/9778.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet.process.BaseProcess will be removed from twisted.internet.process.reapProcessHandlers before calling ProcessProtocol.processEnded()
```
twm: I don't think that this will make sense to most users &mdash; `BaseProcess` is really an implementation detail (I'd guess it would be private if this code had been written recently).

Perhaps:

> On POSIX systems a reaped process's transport is removed from twisted.internet.process.reapProcessHandlers before the associated twisted.internet.interfaces.IProcessProtocol is notified.

I am tempted to say it should be a `.misc`, even. It's basically a cleanup.

src/twisted/internet/test/test_process.py
```diff
@@ -844,6 +845,59 @@ def spawnChild():
...
+
+    @onlyOnPOSIX
+    def test_process_unregistered_before_protocol_ended_callback(self):
+        """
+        Process is removed from reapProcessHandler dict before running
+        ProcessProtocol.processEnded() callback.
+        """
+        results = []
+
+        class TestProcessProtocol(ProcessProtocol):
```
twm, line 858: All of these inner classes and functions [need docstrings](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings).

twm (approved):
The meat of this looks good! Feel free to land with some minor stylistic fixes.

Thanks!

twm (approved):

PR #1241, opened 2020-03-30
[#9790] Remove support for Python 2 in _setup.py

hawkowl:
https://twistedmatrix.com/trac/ticket/9790#ticket

This also removes some old OSes (solaris 10???) and bumps a few dependencies to be simpler (idna mostly). It also has the py2 removal changelog entry.


src/twisted/python/_setup.py
```diff
@@ -81,34 +75,25 @@
...
-        'idna >= 0.6, != 2.3',
+        'idna >= 2.4',
     ],
     conch=[
         'pyasn1',
         'cryptography >= 2.6',
         'appdirs >= 1.4.0',
         'bcrypt >= 3.0.0',
     ],
-    soap=['soappy'],
```
glyph, line 111: lol `twisted.web.soap` still exists

src/twisted/python/_setup.py
```diff
@@ -262,21 +240,16 @@ class my_build_ext(build_ext_twisted):
...
         "zope.interface >= 4.4.2",
         "constantly >= 15.1",
         "incremental >= 16.10.1",
         "Automat >= 0.3.0",
         "hyperlink >= 17.1.1",
-        # PyHamcrest 1.10.0 is Python 3 only, but lacks package metadata that
-        # says so. This condition can be dropped when Twisted drops support for
-        # Python 2.7.
-        "PyHamcrest >= 1.9.0, != 1.10.0",
+        "PyHamcrest >= 1.9.0",
```
glyph, line 252: Shouldn't this be a test-only dependency?

src/twisted/python/_setup.py
glyph: (Given its apparent total lack of test coverage I'm suggesting we delete it; anyone who wanted this functionality on python 3 would have noticed it was missing some time ago.)

src/twisted/python/_setup.py
```diff
@@ -190,11 +173,6 @@ def __init__(self, *args, **kwargs):
             ],
         libraries=["ws2_32"],
         condition=lambda _: _isCPython and sys.platform == "win32"),
-
-    ConditionalExtension(
-        "twisted.python._sendmsg",
```
glyph, line 195: and good riddance!!!!!

rodrigc (approved):
This looks OK

PR #1242, opened 2020-03-30
[#9789] Remove _oldstyle and things that use it.

hawkowl:
https://twistedmatrix.com/trac/ticket/9789


hawkowl:
Note: I don't think any of the lint errors are caused by this, just by line numbers shifting about.

rodrigc (approved):
This is a straight removal, looks OK

rodrigc:
The lint is failing because of:

```
TypeError: modifiedBlankLines() takes 7 positional arguments but 9 were given
```

twistedchecker implements modifiedBlankLines() here: https://github.com/twisted/twistedchecker/blob/80060e1c07cf5d67d747dbec8ec0e5ee913e8929/twistedchecker/checkers/pycodestyleformat.py#L218

which overrides the modifiedBlankLines() in pycodestyle.

I remember seeing this in the past.  You need to lock pycodestyle and twistedchecker to very specific versions, otherwise twistedchecker fails like this.

I guess long term, twisted should get ride of twistedchecker.

PR #1243, opened 2020-03-30
[#9791] Remove .3only.py files and reintegrate them into the regular test files

hawkowl:
Also clean up some PY3 checks in those files while I'm there.

Trac: https://twistedmatrix.com/trac/ticket/9791#ticket


glyph (approved):
make it so!

PR #1244, opened 2020-04-02
[4964] Object `trans` inconsistent after exception

tonal:
In method `_runInteraction` object `trans` inconsistent after raise exception in interaction.
See https://twistedmatrix.com/trac/ticket/4964 for example.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/4964

### DIFF
--- a/src/twisted/enterprise/adbapi.py
+++ b/src/twisted/enterprise/adbapi.py
@@ -47,7 +47,7 @@ class Connection(object):
         pass
 
 
-    def rollback(self, exValue):
+    def rollback(self, exValue=None):
         if not self._pool.reconnect:
             self._connection.rollback()
             return
--- a/src/twisted/enterprise/adbapi.py
+++ b/src/twisted/enterprise/adbapi.py
@@ -47,7 +47,7 @@ class Connection(object):
         pass
 
 
-    def rollback(self):
+    def rollback(self, exValue):
         if not self._pool.reconnect:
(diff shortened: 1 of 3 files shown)


glyph (changes requested):
Thank you for your contribution to Twisted!

Would you mind adding some unit tests for this?  It looks like a good fix but I'd generally like to look at the test cases to make sure I'm reviewing it properly.

src/twisted/enterprise/adbapi.py
```diff
@@ -458,16 +458,16 @@ def _close(self, conn):
...
+            trans = self.transactionFactory(self, conn)
             result = interaction(trans, *args, **kw)
             trans.close()
             conn.commit()
             return result
         except:
             excType, excValue, excTraceback = sys.exc_info()
             try:
-                conn.rollback()
+                conn.rollback(excValue)
```
glyph, line 470: `conn` is a DB-API object, whose interface is defined in PEP 249.  https://www.python.org/dev/peps/pep-0249/#id4  So we can't add an exception to the `rollback` method, even if some drivers (which ones?) might support it.

glyph:
Leaving #4964 open in case someone wants to propose an alternate fix, but it seems like this code has been abandoned for quite some time.  Please feel free to reopen with resolved conflicts and passing tests though!

PR #1245, opened 2020-04-02
[#9793] Remove some unported modules and now unused code from twisted.python

hawkowl:
https://twistedmatrix.com/trac/ticket/9793#ticket


src/twisted/test/test_compat.py
```diff
@@ -585,11 +444,7 @@ def test_unicode(self):
         """
         C{compat.unicode} is C{str} on Python 3, C{unicode} on Python 2.
         """
-        if _PY3:
-            expected = str
-        else:
-            expected = unicode
-        self.assertIs(unicodeCompat, expected)
+        self.assertIs(unicodeCompat, str)
```
rodrigc, line 447: This looks fine, but in the docstring above, you might want to remove mention of Python 2, since
you are removing the test for Python 2.

rodrigc (approved):
Minor comment, but looks OK, since this is a straight removal of code.

rodrigc (changes requested):
I think this PR is OK except for the one minor docstring suggestion that I had.

I notice some of the tests were failing due to:

```
packages/twisted/python/compat.py", line 413, in networkString
    raise TypeError("Can only convert text to bytes on Python 3")
builtins.TypeError: Can only convert text to bytes on Python 3
twisted.test.test_compat.NetworkStringTests.test_bytes
--------------------------------------------------------
```

Depending on what you want to do, My recommendation is to convert `networkString()` from:

```
    def networkString(s):
        if not isinstance(s, unicode):
            raise TypeError("Can only convert text to bytes on Python 3")
        return s.encode('ascii')
```

to:

```
    def networkString(s):
        if instance(s, bytes):
            return s
        elif isinstance(s, unicode):
            return s.encode('ascii')
        else:
            raise TypeError("Can only convert text to bytes on Python 3")
```

rodrigc (approved):
This looks OK now.

rodrigc:
lint is failing due to: https://github.com/twisted/twisted/pull/1242#issuecomment-611192960

PR #1247, opened 2020-04-08
Remove PY3 checks in trial

rodrigc:
## Contributor Checklist:

* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9795


src/twisted/trial/runner.py
```diff
@@ -62,12 +61,8 @@ def isPackageDirectory(dirname):
     if dirname is a package directory.  Otherwise, returns False
     """
     def _getSuffixes():
-        if _PY3:
-            import importlib
-            return importlib.machinery.all_suffixes()
-        else:
-            import imp
-            return list(zip(*imp.get_suffixes()))[0]
+        import importlib
```
hawkowl: these two importlib imports can move to the top (1)

src/twisted/trial/runner.py
```diff
@@ -126,20 +121,14 @@ def _importFromFile(fn, moduleName=None):
...
-            raise SyntaxError(fn)
-        module = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(module)
-        sys.modules[moduleName] = module
-    else:
-        import imp
-
-        with open(fn, 'r') as fd:
-            module = imp.load_source(moduleName, fn, fd)
+    import importlib
```
hawkowl: these two importlib imports can move to the top (2)

hawkowl (changes requested):
Since there's no more reason for Py3TestLoader, the subclass and the overridden functions could be integrated back in and delete a bunch of old code there. We could do this in another ticket, but it feels like it might make sense here.

src/twisted/trial/runner.py
rodrigc: done

src/twisted/trial/runner.py
rodrigc: done

rodrigc:
> Since there's no more reason for Py3TestLoader, the subclass and the overridden functions could be integrated back in and delete a bunch of old code there. We could do this in another ticket, but it feels like it might make sense here.

DONE

PR #1250, opened 2020-04-12
Add CPython 3.8 to CI

altendky:
https://twistedmatrix.com/trac/ticket/9798

PR #1251, opened 2020-04-12
Add digestmod parameter to HMAC.__init__() to fix Python 3.8+

rodrigc:
https://twistedmatrix.com/trac/ticket/9800

In Python 3.4+ if you did not specify a `digestmod` parameter to `HMAC.__init__()`, the digestmod would default to md5.

In Python 3.8+, if you don't specify a digestmod, you get an error:

​https://github.com/python/cpython/commit/8bb0b5b03cffa2a2e74f248ef479a9e7fbe95cf4 ​https://github.com/python/cpython/commit/f33c57d5c780da1500619f548585792bb5b750ee

This is similar to the fix that was done in https://github.com/twisted/twisted/pull/1219

### DIFF
--- a/src/twisted/cred/credentials.py
+++ b/src/twisted/cred/credentials.py
@@ -441,7 +441,8 @@ class CramMD5Credentials(object):
 
 
     def checkPassword(self, password):
-        verify = hexlify(hmac.HMAC(password, self.challenge).digest())
+        verify = hexlify(hmac.HMAC(password, self.challenge,
+                         digestmod=md5).digest())
         return verify == self.response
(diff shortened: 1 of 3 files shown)


altendky (approved):
It reads ok to me.  Fair warning that I am not well versed on the crypto side of things so I'm going purely off the deprecation notice https://docs.python.org/3/library/hmac.html#hmac.new and diff.  I believe it's your responsibility to decide if my review is sufficient.  `:]`  If we get #1250 in we can also verify in CI that all these errors are gone.

PR #1252, opened 2020-04-13
Change import of cgi.parse_qs to urllib.parse.parse_qs

rodrigc:
https://twistedmatrix.com/trac/ticket/9801


rodrigc:
@altendky Can you look at this?  With this all the tests pass on Python 3.8

PR #1255, opened 2020-04-15
Remove __future__ imports

rodrigc:
https://twistedmatrix.com/trac/ticket/9803

Now that Python 3.5 is the minimum version of Python supported by Twisted,
we can get rid of some `__future__`  imports that were used for Python 2 compatibility.


adiroiban (approved):
All good. Thanks for the cleanup.

PR #1256, opened 2020-04-21
Require CPython 3.8 passes on Windows and macOS

altendky:
https://twistedmatrix.com/trac/ticket/9799


altendky:
Waiting for CPython 3.8 to pass in CI.  No updates should be needed to this PR beyond catch up merges from trunk.  When it passes it can be merged so we retain 3.8 compatibility going forward.

altendky:
This should be more granular.  One for macOS and another for Windows.  See https://twistedmatrix.com/trac/ticket/9799#comment:9 for discussion.

PR #1257, opened 2020-05-05
Revert "Do not allow travis failures on Python 3.8"

glyph:
Reopens: ticket:9798


glyph:
Apparently I misunderstood the sequence of events here. oops!

PR #1258, opened 2020-05-05
Require CPython 3.8 passes on macOS

altendky:
https://twistedmatrix.com/trac/ticket/9808

confusion...  discussion at https://twistedmatrix.com/trac/ticket/9799#comment:9


rodrigc (approved):
OK

PR #1259, opened 2020-05-05
[WIP] Require CPython 3.8 passes on Windows

altendky:
https://twistedmatrix.com/trac/ticket/9809

confusion...  discussion at https://twistedmatrix.com/trac/ticket/9799#comment:9


rodrigc:
@altendky This fails on Windows due to this: https://dev.azure.com/twistedmatrix/twisted/_build/results?buildId=1711&view=logs&j=c822da56-1c3f-5527-8882-a570698b0a88&t=e0a40107-1633-5cd6-a3ae-f8d8a91ebac8&l=16013

This tries to instantiate the AsyncioSelectorReactor.  It looks like the AsyncioSelectorReactor doesn't work on Windows

In this test:
https://github.com/twisted/twisted/blob/21692573637727252fdc21c872968d413e02cd13/src/twisted/internet/test/test_asyncioreactor.py#L15

It looks like on Python 3.8, `NotImplementedError` is raised instead of `ImportError`.

So probably this test needs to be changed so that it can handle `NotImplementedError` in addition to `ImportError`

altendky:
This ticket and PR are only about CI.  I wasn't specifically taking on the task of achieving 3.8 compatibility at this point.  The default loop change is the only thing I'm offhand familiar with for Windows.

https://twistedmatrix.com/trac/ticket/9766

altendky:
Test changes do not belong here. This is not about fixing tests.

altendky:
The Twisted `AsyncioSelectorReactor` just calls `asyncio.get_event_loop()` which for Windows Python 3.8 does not actually get a selector event loop but rather the proactor event loop which doesn't implement `.add_reader()` and thus we get a `NotImplementedError`.  I don't think we want to just skip all the asyncio tests because of that.  I know that any change for this does not belong in this PR.  See 9766.

https://twistedmatrix.com/trac/ticket/9766

https://github.com/twisted/twisted/blob/21692573637727252fdc21c872968d413e02cd13/src/twisted/internet/asyncioreactor.py#L33-L43

rodrigc:
It's OK to fix tests as a side-effect of enabling CI.  I've done it before.
Getting the `AsyncioSelectorReactor` to work on Windows Python 3.8 looks like it will involve more work, so if you want to submit a separate PR to get that to work, that is OK.

altendky:
I would rather keep these changes clean for clarity.  Are the changes you added still relevant or were they just exploratory and now we know the problem is real rather than just a matter of disabling tests?

rodrigc:
@altendky I fixed the Windows asyncio reactor on Python 3.8+ in https://github.com/twisted/twisted/pull/1338 .
Can you review that?

altendky:
@rodrigc, I just got started today, but I'm working on it.

rodrigc:
Fixed in https://github.com/twisted/twisted/pull/1338

PR #1260, opened 2020-05-05
Add mypy to tox

rodrigc:
https://twistedmatrix.com/trac/ticket/9810

I based this loosely on the mypy CI in klein: https://github.com/twisted/klein/blob/master/tox.ini#L284

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ env:
     - PYENV_ROOT="$HOME/.pyenv"
 
 
-matrix:
+jobs:
   include:
     - services:
         - docker
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -31,7 +31,7 @@ prune .travis
 prune .github
 
 # Include test-running utilities for downstream packagers
-include tox.ini .coveragerc
+include mypy.ini tox.ini .coveragerc
 
 # Include our docs templates
 recursive-include src/twisted/python/_pydoctortemplates *.html
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,20 @@
+[mypy]
+namespace_packages=True
+plugins=mypy_zope:plugin
+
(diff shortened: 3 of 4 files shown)


rodrigc:
mypy run is here: https://travis-ci.com/github/twisted/twisted/jobs/327861205

adiroiban:
Hi and thanks for working on this.
I think that the tox changes are ok.

Not sure about the value of adding it now to travis with ignore-errors.
My experience is that few people will check those builds and they will just use resources on the travis infrastructure.

Maybe just leave the tox info, and enable travis once the run has exit code 0.

But if you think that they are necessary, I am ok with this PR.

adiroiban:
I am looking at the docs from here https://mypy.readthedocs.io/en/latest/existing_code.html

Maybe have a separate `mypy-progress` tox environment which will only check some files and which developers will constantly update.

I couldn't find any diff tool for mypy.

rodrigc:
@adiroiban yes, I agree that few people will check the builds if ignore-errors is turned on.

@glyph had some good ideas for how to add mypy to Twisted: https://twistedmatrix.com/pipermail/twisted-python/2020-April/065050.html

I am looking to see how hard it would be to fix the Twisted code to reduce the mypy errors.
I took a first step by integrating the mypy-zope plugin.

adiroiban (approved):
I think that this should be merged ASAP to make it easy to work on mypy changes in separate PRs.

Feel free to merge is at is is.

I would prefer to keep tox, but don't run travis jobs just yet.

But not a blocker. So if travis job is useful for your, please merge as it is.

Thanks!

rodrigc:
OK, I changed this PR to remove mypy from travis.

For now, I can test out mypy by doing this from the command-line:

```
tox -e mypy
```

I can add mypy back to travis later, after fixing all the mypy errors.

PR #1261, opened 2020-05-05
Fix mypy error: Too few arguments to "makeTestCaseClasses"

rodrigc:
https://twistedmatrix.com/trac/ticket/9811

### DIFF
--- a/src/twisted/internet/test/reactormixins.py
+++ b/src/twisted/internet/test/reactormixins.py
@@ -342,7 +342,7 @@ class ReactorBuilder:
             name = (cls.__name__ + "." + shortReactorName +
                     "Tests").replace(".", "_")
 
-            class testcase(cls, SynchronousTestCase):  # type: ignore
+            class testcase(cls, SynchronousTestCase):  # type: ignore[valid-type,misc]   # noqa
                 __module__ = cls.__module__
                 if reactor in cls.skippedReactors:
                     skip = cls.skippedReactors[reactor]
--- a/src/twisted/internet/test/reactormixins.py
+++ b/src/twisted/internet/test/reactormixins.py
@@ -17,7 +17,10 @@ __metaclass__ = type
 
 __all__ = ['TestTimeoutError', 'ReactorBuilder', 'needsRunningReactor']
 
-import os, signal, time
+import os
+import signal
+import time
+from typing import Dict, Type, Union
 
 from twisted.trial.unittest import SynchronousTestCase, SkipTest
 from twisted.trial.util import DEFAULT_TIMEOUT_DURATION, acquireAttribute
@@ -27,14 +30,14 @@ from twisted.python.deprecate import _fullyQualifiedName as fullyQualifiedName
 
 from twisted.python import log
 from twisted.python.failure import Failure
-from twisted.python.compat import _PY3
 
 
 # Access private APIs.
-if platform.isWindows():
-    process = None
-else:
+try:
     from twisted.internet import process
+except ImportError:
+    process = None  # type: ignore
+
 
 
 
@@ -147,8 +150,7 @@ class ReactorBuilder:
                 "twisted.internet.gireactor.GIReactor",
                 "twisted.internet.gtk3reactor.Gtk3Reactor"])
 
-        if _PY3:
-            _reactors.append(
+        _reactors.append(
                 "twisted.internet.asyncioreactor.AsyncioSelectorReactor")
 
         if platform.isMacOSX():
@@ -168,7 +170,7 @@ class ReactorBuilder:
     reactorFactory = None
     originalHandler = None
     requiredInterfaces = None
-    skippedReactors = {}
+    skippedReactors = {}  # type: Dict[str, str]
 
     def setUp(self):
         """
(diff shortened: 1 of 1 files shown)


rodrigc:
This helps address some of the mypy errors found here: https://travis-ci.com/github/twisted/twisted/jobs/327861205#L4888

such as:

```
src/twisted/internet/test/reactormixins.py:350:27: error: Incompatible types in
assignment (expression has type "classmethod", variable has type
"Callable[[ReactorBuilder], Any]")  [assignment]
        makeTestCaseClasses = classmethod(makeTestCaseClasses)
```

```
src/twisted/internet/test/test_win32events.py:199:18: error: Too few arguments
for "makeTestCaseClasses"  [call-arg]
    globals().update(Win32EventsTestsBuilder.makeTestCaseClasses())
```

I followed the suggestions here: https://stackoverflow.com/questions/44640479/mypy-annotation-for-classmethod-returning-instance
for using `TypeVar`

src/twisted/internet/test/reactormixins.py
```diff
@@ -27,14 +30,14 @@
...
 
 
 # Access private APIs.
-if platform.isWindows():
-    process = None
-else:
+try:
     from twisted.internet import process
+except ImportError:
+    process = None  # type: ignore
```
glyph: Could you file a ticket for a more mypy-friendly way of doing this, and link it here before landing?  I am OK with adding `type: ignore` for an import hack like this (particularly in a test helper) but before the pattern replicates ad infinitum I would like to record the fact that we should clean it up somehow later.

src/twisted/internet/test/reactormixins.py
```diff
@@ -324,26 +326,30 @@ def stop():
             timedOutCall.cancel()
 
 
-    def makeTestCaseClasses(cls):
+    @classmethod
+    def makeTestCaseClasses(
+        cls: Type['ReactorBuilder']
```
glyph, line 333: Why are these all quoted?  The names are imported, so they should be in scope?

src/twisted/internet/test/reactormixins.py
```diff
@@ -324,26 +326,30 @@ def stop():
...
+    ) -> Dict[str, Union[Type['ReactorBuilder'], Type['SynchronousTestCase']]]:
         """
         Create a L{SynchronousTestCase} subclass which mixes in C{cls} for each
         known reactor and return a dict mapping their names to them.
         """
-        classes = {}
+        classes = {}  # type: Dict[str, Union[Type['ReactorBuilder'], Type['SynchronousTestCase']]]   # noqa
```
glyph: Quoting in type *comments* seems doubly pointless, even if it works?

src/twisted/internet/test/reactormixins.py
```diff
@@ -324,26 +326,30 @@ def stop():
...
         for reactor in cls._reactors:
             shortReactorName = reactor.split(".")[-1]
-            name = (cls.__name__ + "." + shortReactorName + "Tests").replace(".", "_")
-            class testcase(cls, SynchronousTestCase):
+            name = (cls.__name__ + "." + shortReactorName +
```
glyph, line 342: I guess twistedchecker complained at you?

src/twisted/internet/test/reactormixins.py
```diff
@@ -324,26 +326,30 @@ def stop():
...
+                    "Tests").replace(".", "_")
+
+            class testcase(cls, SynchronousTestCase):  # type: ignore
```
glyph: What doesn't mypy like about this?  Again, I would like type: ignores to be used sparingly and we should generally have a plan for how to get rid of them, at least file an upstream bug against mypy.

src/twisted/internet/test/reactormixins.py
```diff
@@ -324,26 +326,30 @@ def stop():
...
                 __module__ = cls.__module__
                 if reactor in cls.skippedReactors:
                     skip = cls.skippedReactors[reactor]
                 try:
                     reactorFactory = namedAny(reactor)
-                except:
+                except Exception:
```
glyph: This should be `BaseException` to avoid a change to its semantics.

glyph (approved):
Aside from needing to annotating the `type: ignore` comments before landing, this looks great, thank you!

src/twisted/internet/test/reactormixins.py
rodrigc: There already is an open issue https://github.com/python/mypy/issues/1297#issuecomment-568663999 where this use-case was mentioned.  I'll follow up there, and see what I can find out.

src/twisted/internet/test/reactormixins.py
rodrigc: `ReactorBuilder` is defined in the same file, so I had to use the syntax `Class name forward references`, as described here: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#class-name-forward-references

otherwise mypy would complain.

src/twisted/internet/test/reactormixins.py
rodrigc: I'm not sure exactly how mypy parses the file and comments to do all the type checks.  I just kept the use the same to be consistent.

src/twisted/internet/test/reactormixins.py
rodrigc: Yes, I made a lot of small changes like this to appease twistedchecker in terms of line spacing, line length, etc.

src/twisted/internet/test/reactormixins.py
rodrigc: Actually mypy did not complain.  If I didn't put ReactorBuilder in quotes, pyflakes complained with:

```
src/twisted/internet/test/reactormixins.py:331:19 undefined name 'ReactorBuilder'
src/twisted/internet/test/reactormixins.py:332:31 undefined name 'ReactorBuilder'
```

rodrigc: I've changed this so SynchronousTestCase is not in quotes, but I still need 'ReactorBuilder` to be in quotes for the annotation.

src/twisted/internet/test/reactormixins.py
rodrigc: If I didn't put that `#type: ignore` there, mypy would give:

```
src/twisted/internet/test/reactormixins.py:343:28: error: Variable "cls" is not valid as a type  [valid-type]
                class testcase(cls, SynchronousTestCase):
                               ^
src/twisted/internet/test/reactormixins.py:343:28: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
src/twisted/internet/test/reactormixins.py:343:28: error: Invalid base class "cls"  [misc]
                class testcase(cls, SynchronousTestCase):
```

The reason for this is that `cls` is a parameter, and the type of the parameter is only known at runtime.  So mypy doesn't know what type it is, and can't check it.

src/twisted/internet/test/reactormixins.py
mthuurne: For mypy's sake you never need to quote, but the Python VM does require it when using a type name before it is defined, unless you use `from __future__ import annotations`, but that's a Python 3.7 feature.

Of course the Python VM doesn't process the contents of comments, so no quoting is necessary there.

src/twisted/internet/test/reactormixins.py
mthuurne: Note that if you have to use `type: ignore`, you can specify the error category to ignore, which reduces the chance of accidentally silencing unexpected errors. In this case it would be `# type: ignore[valid-type]`.

src/twisted/internet/test/reactormixins.py
rodrigc: So based on this: https://github.com/python/mypy/issues/1297#issuecomment-508593494
I think we can remove the `#type: ignore`, and do this:

```
try
    from twisted.internet import process as _process
except ImportError:
    process = None
else:
    process = _process
```

src/twisted/internet/test/reactormixins.py
rodrigc: @mthuurne Yes, for now I need to quote `'ReactorBuilder'`.  Since Twisted still supports Python 3.5 and 3.6, I cannot use `from __future__ import annotations`

src/twisted/internet/test/reactormixins.py
rodrigc: @mthuurne thanks for the feedback.  I've changed this line to specifically turn off two mypy errors on this line:

```
class testcase(cls, SynchronousTestCase):  # type: ignore[valid-type,misc]
```

PR #1262, opened 2020-05-06
[#9807] contextvars support for Deferred

hawkowl:
https://twistedmatrix.com/trac/ticket/9807


src/twisted/internet/defer.py
```diff
@@ -295,13 +299,55 @@ def __init__(self, canceller=None):
...
+        """
+        if not _contextvarsSupport:
+            raise RuntimeError(
+                "contextvars support is not available on this platform."
+            )
+
+        return self._context
+
+
+    def setContext(self, context: _Context) -> NoReturn:
```
ShadowJonathan, line 336: Why does this never return? The `self._context = context` bit runs at the end of the function, after which it implicitly returns (`None`).

Should this maybe be `Optional[NoReturn]`? If that even affects anything.

PR #1264, opened 2020-05-07
Use unittest.skipIf decorator for skipping certain test functions.

rodrigc:
https://twistedmatrix.com/trac/ticket/9812

This reduces errors reported by mypy of the form:

```
src/twisted/words/test/test_domish.py:193:9: error: "Callable[[ElementTests], Any]" has no attribute "skip"  [attr-defined]
            test_addContentBytes.skip = (
```

### DIFF
--- a/src/twisted/conch/test/test_default.py
+++ b/src/twisted/conch/test/test_default.py
@@ -9,17 +9,7 @@ Tests for L{twisted.conch.client.default}.
 import sys
 
 from twisted.python.reflect import requireModule
-
-if requireModule('cryptography') and requireModule('pyasn1'):
-    from twisted.conch.client.agent import SSHAgentClient
-    from twisted.conch.client.default import SSHUserAuthClient
-    from twisted.conch.client.options import ConchOptions
-    from twisted.conch.client import default
-    from twisted.conch.ssh.keys import Key
-    skip = None
-else:
-    skip = "cryptography and PyASN1 required for twisted.conch.client.default."
-
+from unittest import skipIf
 from twisted.trial.unittest import TestCase
 from twisted.python.filepath import FilePath
 from twisted.conch.error import ConchError
@@ -28,18 +18,33 @@ from twisted.test.proto_helpers import StringTransport
 from twisted.python.compat import nativeString
 from twisted.python.runtime import platform
 
+doSkip = False
+skipReason = ""
+
+if requireModule('cryptography') and requireModule('pyasn1'):
+    from twisted.conch.client.agent import SSHAgentClient
+    from twisted.conch.client.default import SSHUserAuthClient
+    from twisted.conch.client.options import ConchOptions
+    from twisted.conch.client import default
+    from twisted.conch.ssh.keys import Key
+else:
+    doSkip = True
+    skipReason = (
+        "cryptography and PyASN1 required for twisted.conch.client.default.")
+    skip = skipReason  # no SSL available, skip the entire module
+
 if platform.isWindows():
-    windowsSkip = (
+    doSkip = True
+    skipReason = (
         "genericAnswers and getPassword does not work on Windows."
         " Should be fixed as part of fixing bug 6409 and 6410")
-else:
-    windowsSkip = skip
 
-ttySkip = None
 if not sys.stdin.isatty():
-    ttySkip = "sys.stdin is not an interactive tty"
+    doSkip = True
+    skipReason = "sys.stdin is not an interactive tty"
 if not sys.stdout.isatty():
-    ttySkip = "sys.stdout is not an interactive tty"
+    doSkip = True
+    skipReason = "sys.stdout is not an interactive tty"
 
 
 
@@ -192,6 +197,7 @@ class SSHUserAuthClientTests(TestCase):
         return client.getPrivateKey().addCallback(_cbGetPrivateKey)
 
 
+    @skipIf(doSkip, skipReason)
     def test_getPassword(self):
         """
         Get the password using
@@ -217,9 +223,8 @@ class SSHUserAuthClientTests(TestCase):
         d.addCallback(self.assertEqual, b'bad password')
         return d
 
-    test_getPassword.skip = windowsSkip or ttySkip
 
-
+    @skipIf(doSkip, skipReason)
     def test_getPasswordPrompt(self):
         """
         Get the password using
@@ -239,9 +244,8 @@ class SSHUserAuthClientTests(TestCase):
         d.addCallback(self.assertEqual, b'bad password')
         return d
 
-    test_getPasswordPrompt.skip = windowsSkip or ttySkip
 
-
+    @skipIf(doSkip, skipReason)
     def test_getPasswordConchError(self):
         """
         Get the password using
@@ -264,9 +268,8 @@ class SSHUserAuthClientTests(TestCase):
             return fail
         self.assertFailure(d, ConchError)
 
-    test_getPasswordConchError.skip = windowsSkip or ttySkip
 
-
+    @skipIf(doSkip, skipReason)
     def test_getGenericAnswers(self):
         """
         L{twisted.conch.client.default.SSHUserAuthClient.getGenericAnswers}
@@ -291,8 +294,6 @@ class SSHUserAuthClientTests(TestCase):
         d.addCallback(
             self.assertListEqual, ["getpass", "raw_input"])
         return d
-
-    test_getGenericAnswers.skip = windowsSkip or ttySkip
 
 
 
--- a/src/twisted/conch/test/test_scripts.py
+++ b/src/twisted/conch/test/test_scripts.py
@@ -4,56 +4,52 @@
 """
 Tests for the command-line interfaces to conch.
 """
+from unittest import skipIf
 from twisted.python.reflect import requireModule
+from twisted.python.test.test_shellcomp import ZshScriptTestMixin
+from twisted.scripts.test.test_scripts import ScriptTestsMixin
+from twisted.trial.unittest import TestCase
 
-if requireModule('pyasn1'):
-    pyasn1Skip = None
-else:
-    pyasn1Skip =  "Cannot run without PyASN1"
+doSkip = False
+skipReason = ""
 
-if requireModule('cryptography'):
-    cryptoSkip = None
-else:
+if not requireModule('pyasn1'):
+    doSkip = True
+    skipReason = "Cannot run without PyASN1"
+
+if not requireModule('cryptography'):
+    doSkip = True
     cryptoSkip = "can't run w/o cryptography"
 
-if requireModule('tty'):
-    ttySkip = None
-else:
+if not requireModule('tty'):
+    doSkip = True
     ttySkip = "can't run w/o tty"
 
 try:
-    import Tkinter
+    import tkinter
 except ImportError:
-    tkskip = "can't run w/o Tkinter"
+    doSkip = True
+    skipReason = "can't run w/o tkinter"
 else:
     try:
-        Tkinter.Tk().destroy()
-    except Tkinter.TclError as e:
-        tkskip = "Can't test Tkinter: " + str(e)
-    else:
-        tkskip = None
-
-from twisted.trial.unittest import TestCase
-from twisted.scripts.test.test_scripts import ScriptTestsMixin
-from twisted.python.test.test_shellcomp import ZshScriptTestMixin
+        tkinter.Tk().destroy()
+    except tkinter.TclError as e:
+        doSkip = True
+        skipReason = "Can't test Tkinter: " + str(e)
 
 
 
+@skipIf(doSkip, skipReason)
 class ScriptTests(TestCase, ScriptTestsMixin):
     """
     Tests for the Conch scripts.
     """
-    skip = pyasn1Skip or cryptoSkip
-
-
     def test_conch(self):
         self.scriptTest("conch/conch")
-    test_conch.skip = ttySkip or skip
 
 
     def test_cftp(self):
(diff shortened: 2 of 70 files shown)


rodrigc:
@wiml Could you take a look at this?

src/twisted/conch/test/test_default.py
```diff
@@ -9,37 +9,42 @@
...
+from twisted.trial.unittest import TestCase
+from twisted.python.filepath import FilePath
+from twisted.conch.error import ConchError
+from twisted.conch.test import keydata
+from twisted.test.proto_helpers import StringTransport
+from twisted.python.compat import nativeString
+from twisted.python.runtime import platform
+
+doSkip = False
+skipReason = ""
```
wiml, line 22: I don't especially like this pattern (having a pair of generically-named `skip` variables that are always set and used in sync). I think it'd be clearer to either use informatively-named flags (like you did with `HAS_IPV6`) or store the text in the skip flag (like the way this file was written before, `type: Optional[str]`).

src/twisted/conch/test/test_session.py
```diff
@@ -1096,6 +1097,8 @@ def test_connectionMade(self):
         self.assertEqual(self.transport.buf, b'buffer')
 
 
+    @skipIf(getattr(signal, 'SIGALRM', None) is None,
```
wiml: I'm sure it's not worth changing for this PR but this could be written more clearly as `not hasattr()`

src/twisted/internet/test/test_unix.py
```diff
@@ -292,6 +290,8 @@ def makeConnection(self, transport):
         self.assertEqual(server.addresses['peer'], client.addresses['host'])
 
 
+    @skipIf(sendmsgSkip, "sendmsg extension unavailable, "
```
wiml: Storing the message in `sendmsgSkip` would prevent having to duplicate the message unnecessarily

src/twisted/test/test_internet.py
```diff
@@ -1363,38 +1365,42 @@ def test_reentrantStreamingProducerUsingWriteSequence(self):
...
-            portNo = p.getHost().port
-            self.assertNotEqual(str(p).find(str(portNo)), -1,
-                                "%d not found in %s" % (portNo, p))
-            return p.stopListening()
 
-        if _PY3:
-            testSSL.skip = ("Re-enable once the Python 3 SSL port is done.")
+    @skipIf(not interfaces.IReactorSSL(reactor, None),
+            "IReactorSSL is needed")
+    @skipIf(not ssl, "SSL support is missing")
```
wiml, line 1391: I'm curious, is there a situation where we have `IReactorSSL` but not the `ssl` module?

src/twisted/test/test_paths.py
```diff
@@ -639,7 +636,7 @@ def symlink(self, target, name):
             host platform.
         """
         if symlinkSkip:
-            raise SkipTest(symlinkSkip)
+            raise SkipTest("Platform does not support symlinks")
```
wiml: Should this test even be here? It seems to me that all the tests that call this method should have `@skipIf` already.

src/twisted/test/test_plugin.py
```diff
@@ -536,47 +527,6 @@ def test_developmentPluginAvailability(self):
             self.assertEqual(names, ['app', 'dev'])
 
 
-    def test_freshPyReplacesStalePyc(self):
```
wiml, line 539: Did you intend to remove this entire test?

src/twisted/test/test_ssl.py
```diff
@@ -271,11 +271,13 @@ def __init__(self, *args, **kw):
 
 
 
-class StolenTCPTests(ProperlyCloseFilesMixin, unittest.TestCase):
+class StolenTCPTests(ProperlyCloseFilesMixin, TestCase):
     """
     For SSL transports, test many of the same things which are tested for
     TCP transports.
     """
+    if interfaces.IReactorSSL(reactor, None) is None:
```
wiml, line 279: Should this (and the later instances in this file) be using `@skipIf` ?  (If not, I think the old version, where it's clear that a single conditional is causing multiple test classes to be skipped in the same way, is clearer)

src/twisted/test/test_stdio.py
```diff
@@ -88,9 +86,11 @@ def processEnded(self, reason):
 
 
 
-class StandardInputOutputTests(unittest.TestCase):
+class StandardInputOutputTests(TestCase):
 
-    skip = skipWindowsNopywin32
+    if platform.isWindows() and requireModule('win32process') is None:
```
wiml, line 91: Should this be using `@skipIf` ?

src/twisted/test/test_unix.py
```diff
@@ -312,6 +290,10 @@ class DatagramUnixSocketTests(unittest.TestCase):
     """
     Test datagram UNIX sockets.
     """
+
+    if not interfaces.IReactorUNIXDatagram(reactor, None):
```
wiml: Should this be using `@skipIf` ?

src/twisted/trial/test/test_assertions.py
```diff
@@ -749,7 +749,7 @@ def test_assertDictEqual(self):
         """
         self.assertDictEqual({'a': 1}, {'a': 1})
     if getattr(unittest.SynchronousTestCase, 'assertDictEqual', None) is None:
-        test_assertDictEqual.skip = (
+        test_assertDictEqual.skip = (  # type: ignore[attr-defined]
```
wiml, line 752: Should this be using `@skipIf` ?

src/twisted/web/test/test_agent.py
```diff
@@ -62,8 +62,8 @@
     from twisted.internet import ssl as _ssl
 except ImportError:
     ssl = None
-    skipWhenNoSSL = "SSL not present, cannot run SSL tests."
```
wiml, line 65: If we're moving the message text out to the individual tests anyway, I think it'd be clearer to have a single flag here named "sslPresent" or something, and use `@skipIf(not sslPresent, ...)` and `@skipIf(sslPresent, ...)` as appropriate

src/twisted/web/test/test_agent.py
```diff
@@ -1421,7 +1419,8 @@ class AgentHTTPSTests(TestCase, FakeReactorAndConnectMixin,
     """
     Tests for the new HTTP client API that depends on SSL.
     """
-    skip = skipWhenNoSSL
+    if skipWhenNoSSL:
```
wiml: Should this be using `@skipIf` ?

src/twisted/conch/test/test_scripts.py
```diff
@@ -4,56 +4,52 @@
...
-if requireModule('tty'):
-    ttySkip = None
-else:
+if not requireModule('tty'):
+    doSkip = True
     ttySkip = "can't run w/o tty"
 
 try:
-    import Tkinter
+    import tkinter
```
wiml, line 29: This seems like it either breaks or fixes the tests on case-sensitive filesystems (probably fixes them). It might be worth opening another trac ticket for this (both tickets can be listed on the merge commit's `fixes:` line) and also fixing the other two capital-T `Tkinter` imports in `examples/tkinterdemo.py` and `howto/choosing-reactor.rst`.

wiml (changes requested):
Whew, that's a lot of files! (Might have been easier to split it into one PR for all of the import-style changes and one for the changes that affect tests/skipping). Looks generally good to me, I have left a few comments on possible oversights and stylistic opinions in-line.

src/twisted/test/test_paths.py
rodrigc: OK, I've taken this out.

src/twisted/test/test_unix.py
rodrigc: Yes, that's right.  I've taken it out.

src/twisted/trial/test/test_assertions.py
rodrigc: I'm keeping that usage of `.skip` in that one place, because although I changed to use `@skipIf` in many places,  I want to make sure that the old behavior in trial which looks for a `.skip` attribute is not broken.

src/twisted/web/test/test_agent.py
rodrigc: that makes sense to me.  I've changed it to `@skipIf(not sslPresent`

src/twisted/web/test/test_agent.py
rodrigc: yes, I've changed it

src/twisted/conch/test/test_scripts.py
rodrigc: No, this is a Python 2 vs. Python 3 thing.  On Python 2, the module was named `Tkinter`, but on Python 3, it was changed to `tkinter`.  

I can open a separate PR for the example and howto.

src/twisted/conch/test/test_session.py
rodrigc: OK, I've changed this to use `not hasattr()`

src/twisted/test/test_internet.py
rodrigc: I'm not sure, the tests were written like that before.

src/twisted/internet/test/test_unix.py
rodrigc: OK, I have replaced this with:
`@skipIf(not sendmsg , sendmsgSkipReason)`

src/twisted/conch/test/test_default.py
rodrigc: In this file, the existing pattern of skipping is a bit complicated, because there are multiple reasons for skipping tests, i.e. Windows, invalid tty, pyasn1 unavailable.

I tried to keep the pattern the same while introducing the use of `skipIf`

src/twisted/test/test_internet.py
adiroiban: From twisted/python/_setup.py

```
    requirements = [
        "zope.interface >= 4.4.2",
        "constantly >= 15.1",
        "incremental >= 16.10.1",
        "Automat >= 0.3.0",
        "hyperlink >= 17.1.1",
        "PyHamcrest >= 1.9.0",
        "attrs >= 17.4.0",
    ]
```

pyopenssl/cryptography is not a hard requirement.

You need to do the extra `[tls]` install.

```
_EXTRA_OPTIONS = dict(
    dev=_dev,
    tls=[
        'pyopenssl >= 16.0.0',
        # service_identity 18.1.0 added support for validating IP addresses in
        # certificate subjectAltNames
        'service_identity >= 18.1.0',
        # idna 2.3 introduced some changes that break a few things.  Avoid it.
        # The problems were fixed in 2.4.
        'idna >= 0.6, != 2.3',
    ],
```
But I agree that it might be practicle to always have pyopenssl as a dependency.

src/twisted/words/test/test_jabberxmlstream.py
```diff
@@ -21,12 +21,13 @@
...
-    from twisted.internet import ssl
+    from twisted.internet import ssl as _ssl
 except ImportError:
     ssl = None
-    skipWhenNoSSL = "SSL not available"
+    skipWhenNoSSL = True
 else:
-    skipWhenNoSSL = None
+    ssl = _ssl
+    skipWhenNoSSL = False
```
adiroiban: To reduce code duplication, maybe have something like this:

```
    ssl = None
    skipWhenNoSSL = (True, "SSL not available")
else:
    ssl = _ssl
    skipWhenNoSSL = (False, "")

@skipIf(*skipWhenNoSSL)
def test_wantedSupported(self):
```

Now I know that is a lot of work, so maybe just leave the duplicate code as it is.

adiroiban: And since there are a lot of SSL skip conditions, maybe it would make sense to have a shared code like this

```
from twisted.trial import conditionals


@conditionals.skipWhenNoTLS
def test_wantedSupported(self):
```

In this way, all the SSL skipping code is controller from a single place.

And there can be separate conditionals for each type of extra requirements, as defined in _setup.py

src/twisted/web/test/test_webclient.py
```diff
@@ -866,6 +866,13 @@ def test_setURLRelativePath(self):
 
 
 class WebClientSSLTests(WebClientTests):
+
+    if ssl is None or not hasattr(ssl, 'DefaultOpenSSLContextFactory'):
+        skip = "OpenSSL not present"
```
adiroiban, line 871: Why not use the skipIf decorator here?

src/twisted/web/test/test_distrib.py
```diff
@@ -517,13 +518,11 @@ def cbRendered(ignored):
         return result
 
 
+    @skipIf(not pwd, "pwd module required")
```
adiroiban, line 521: Maybe this can be replaced by a shared skip condition for Unix... most Unix systems have pwd. 
It is missing only on Win

adiroiban (approved):
Thanks for the changes.

It would be nice to use this opportunity to organize the skipping conditions used in Twisted tests.

Like shared conditions for skipping on WIndows or on Unix, or when OpenSSL is missing

---- 

There is significant code duplication in some skip call, but we have the same duplication in `trunk`.

```diff
- @skipIf(not _PY35PLUS, "coroutines not available before Python 3.5")
+ onlyOnPy35Plus = not _PY35PLUS, "coroutines not available before Python 3.5"
+ @skipIf(*onlyOnPy35Plus)

```

----

As a drive by comment, I think that SOAP support can be removed

--------

But this change is an overall improvement over `trunk` so I think that it can be merged.

PR #1268, opened 2020-05-11
Always use sys.stdout.buffer which is always available on Python 3+

rodrigc:
https://twistedmatrix.com/trac/ticket/9823

Fix mypy errors like:

```
src/twisted/test/process_stdinreader.py:20:14: error: Incompatible types in assignment (expression has type "TextIO", variable has type "BinaryIO")  [assignment]
        stdout = sys.stdout
```

### DIFF
--- a/src/twisted/internet/test/process_cli.py
+++ b/src/twisted/internet/test/process_cli.py
@@ -1,4 +1,3 @@
-
 import sys
 import os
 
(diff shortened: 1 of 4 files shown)


adiroiban (commented):
Thanks.

stdout.buffer changes are ok.

I know nothing about python type system...I am still on python 2.7.

I don't understand  the type:ignore changes

`msvcrt.setmode(sys.stdout.fileno(), os.O_BINARY)  # type:ignore[attr-defined]  # noqa`

What is the point of using type hints, when you start ignoring them ? :)

I think that this needs to be reviewed by someone else, which is familiar with Python type system

rodrigc:
mypy complains on that line because `msvcrt` can only be imported on Windows.
So in that case mypy doesn't have enough information to know that `setmode` is part of the msvcrt module, and thus emits an error.  There is no other clean way to silence this error, other than putting this comment to turn off this particular error on this line.

meejah (approved):
:+1: The actual code changes here look good to me, given "python3 only on trunk".

The type-annotation comments would be the first in Twisted, but there does seem to be support for moving to more mypy usage, so I think this is fine.

PR #1269, opened 2020-05-11
[#9825] Deferred.fromCoroutine

hawkowl:
https://tm.tl/#9825


docs/core/howto/defer-intro.rst
```diff
@@ -315,6 +315,10 @@ Otherwise, ``g`` will be passed the asynchronous equivalent of the return value
 Inline callbacks - using 'yield'
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. note::
+
+    Unless you require Python 2 compatibility, writing coroutines with the functionality described in "Coroutines with async/await" is preferred over ``inlineCallbacks``.
```
glyph: I’d say “compatibility with python 2 and hence older versions of Twisted” since as written this suggests the version of Twisted this is documenting might work on python 2. It might also be useful to briefly describe the reasons for preferring it? In brief:

- Dedicated syntax
- compatibility with asyncio
- higher performance (optimized C code in the interpreter takes care of the trampoline so twisted doesn’t have to)

src/twisted/internet/defer.py
```diff
@@ -855,6 +856,53 @@ def uncancel(result):
...
+        @param coro: The coroutine object to schedule.
+        @type coro: A Python 3.5+ C{async def} coroutine or a Python 3.4+
+            C{yield from} coroutine.
+
+        @raise ValueError: If C{coro} is not a coroutine or generator.
+
+        @rtype: L{Deferred}
+        """
+        if not iscoroutine(coro) and not isinstance(coro, types.GeneratorType):
+            raise ValueError("%r is not a coroutine" % (coro,))
```
glyph: If you’re going to use a built-in exception here, it should really be TypeError since this is an error about types not values. But it should be a specific subclass of TypeError particularly given that it’s caught below; NotACoroutineError or something; maybe inherit from TypeError if you want to be able to classify it that way.

src/twisted/internet/defer.py
```diff
@@ -886,49 +934,23 @@ def ensureDeferred(coro):
...
-        raise ValueError("%r is not a coroutine or a Deferred" % (coro,))
-
-    # Must be a Deferred
-    return coro
+    if isinstance(coro, Deferred):
+        return coro
+    else:
+        try:
+            return Deferred.fromCoroutine(coro)
+        except ValueError:
```
glyph: Definitely don’t just catch a builtin here, make that more specific exception type.

src/twisted/internet/defer.py
```diff
@@ -886,49 +934,23 @@ def ensureDeferred(coro):
...
+            # It's not a coroutine. Raise an exception, but say that it's also
+            # not a Deferred so the error makes sense.
+            raise ValueError("%r is not a coroutine or a Deferred" % (coro,))
```
glyph: I think you want raise ... from (the other exception just raised) for clarity when debugging? Just a suggestion though, feel free to override. Also should be a new exception type though, derived from TypeError not ValueError

docs/core/howto/defer-intro.rst
```diff
@@ -397,24 +401,21 @@ Coroutines with async/await
...
-
     .. versionadded:: 16.4
 
-On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`.
-It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``, and is a function rather than a decorator.
-
-Calling a coroutine (that is, the result of a function defined by ``async def funcname():``) with :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>` will allow you to "await" on Deferreds inside it, and will return a standard Deferred.
-You can mix and match code which uses regular Deferreds, ``inlineCallbacks``, and ``ensureDeferred`` freely.
+Python 3.5 introduced :pep:`492` ("Coroutines with async and await syntax") and native coroutines.
+:api:`twisted.internet.defer.Deferred.fromCoroutine <Deferred.fromCoroutine>` allows you to write coroutines with the ``async def`` syntax and ``await`` on Deferreds, similar to ``inlineCallbacks``.
```
glyph: If someone is reading this document in order to *understand* some code that's calling `ensureDeferred`, the lack of any mention here might be a little confusing.  I think it might be good to put in an aside here, documenting `ensureDeferred` and saying "earlier versions of Twisted only had the more type-ambiguous `ensureDeferred`..." or something, at least until we deprecate it.

glyph (approved):
The changes requested are pretty trivial (mainly, just "change the exception type to something more specific derived from `TypeError` and not `ValueError`", and "don't *totally* remove `ensureDeferred` from the docs") so if you can address them, go ahead and land!

docs/core/howto/defer-intro.rst
```diff
@@ -315,6 +315,10 @@ Otherwise, ``g`` will be passed the asynchronous equivalent of the return value
 Inline callbacks - using 'yield'
```
twm, line 377: If we want folks to use `async def` instead of `@inlineCallbacks` the sections should be swapped so that the suggested option is first.

src/twisted/internet/defer.py
```diff
@@ -855,6 +856,53 @@ def uncancel(result):
...
+        Schedule the execution of a coroutine that awaits/yields from
+        L{Deferred}s, wrapping it in a L{Deferred} that will fire on
+        success/failure of the coroutine.
+
+        Coroutine functions return a coroutine object, similar to how
+        generators work. This function turns that coroutine into a Deferred,
+        meaning that it can be used in regular Twisted code. For example::
+
+            import treq
+            from twisted.internet.defer import ensureDeferred
```
wsanchez: `ensureDeferred` ➜ `fromCoroutine` here, yeah?

src/twisted/internet/defer.py
wsanchez: +1 to subclass of `TypeError`.

src/twisted/test/test_defer.py
```diff
@@ -1435,6 +1436,30 @@ def ic(d):
...
 
 
+    def test_fromCoroutineRequiresCoroutine(self):
+        """
+        L{Deferred.fromCoroutine} requires a coroutine object or a generator,
+        and will reject things that are not that.
+        """
+        thingsThatAreNotCoroutines = [
+            # Lambda
+            lambda x: x,
```
wsanchez, line 1438: Lambda doesn't differ from function, does it?

wsanchez (approved):
This is great.

src/twisted/internet/defer.py
```diff
@@ -856,6 +857,53 @@ def uncancel(result):
         return self
 
 
+    @classmethod
+    def fromCoroutine(cls, coro):
+        """
+        Schedule the execution of a coroutine that awaits/yields from
```
```suggestion
        Schedule the execution of a coroutine that awaits
```

`yield from` coros are deprecated

src/twisted/internet/defer.py
```diff
@@ -856,6 +857,53 @@ def uncancel(result):
...
+                results = {}
+                for page in pages:
+                    results[page] = await treq.content(await treq.get(page))
+                return results
+
+            def main(reactor):
+                pages = [
+                    "http://localhost:8080"
+                ]
+                d = ensureDeferred(crawl(pages))
```
graingert: `fromCoroutine`?

docs/core/howto/defer-intro.rst
graingert: I think it's worth directly deprecating inlineCallbacks: https://github.com/twisted/twisted/pull/1374

wsanchez: @graingert I agree, but let's do that in the other PR.

wsanchez: @glyph see [updated text](https://github.com/twisted/twisted/pull/1269/files#diff-f6e9f4618d9eb1551a84a53bd2700661R376)

docs/core/howto/defer-intro.rst
wsanchez: done

src/twisted/internet/defer.py
wsanchez: done

src/twisted/internet/defer.py
wsanchez: done

src/twisted/internet/defer.py
wsanchez: done

docs/core/howto/defer-intro.rst
glyph: Love it. Thank you!

docs/core/howto/defer-intro.rst
wsanchez: @glyph how's this: https://github.com/twisted/twisted/pull/1269/files#diff-f6e9f4618d9eb1551a84a53bd2700661R329

src/twisted/internet/defer.py
wsanchez: NotACoroutineError added, subclass of TypeError

src/twisted/internet/defer.py
wsanchez: NotACoroutineError added, subclass of TypeError

src/twisted/internet/defer.py
wsanchez: NotACoroutineError added, subclass of TypeError

src/twisted/internet/defer.py
wsanchez: @glyph: Not sure if you meant more than [fixing the type](https://github.com/twisted/twisted/pull/1269/files#diff-0a695d3d59d1e9452e21eb981b3bb687R958)

twm (approved):
Thanks, everyone! Let's get this in.

PR #1270, opened 2020-05-12
contextvars backport compatibility

altendky:
https://twistedmatrix.com/trac/ticket/9826

WIP for:
- [x] any attempt at all to fix the issue
- [x] newsfragment
- [x] a good and clear solution with some understanding
- [x] update any relevant comments in proximity to the change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ jobs:
     - python: 3.5
       env: TOXENV=py35-alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push
     - python: 3.6
+      env: TOXENV=py36-nodeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push
+    - python: 3.6
       env: TOXENV=py36-alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push
     - python: 3.7
       env: TOXENV=py37-alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push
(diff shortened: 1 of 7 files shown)


altendky:
Circle failed...  kick CI

altendky:
Travis choked, kick CI

altendky:
Windows Py3.5 iocp passed locally so...  kick CI.

hawkowl:
@altendky (terse because I spotted this just as I was heading to bed) I believe there's also documentation that needs to be added in `docs/core/` or so which details the optional extras

altendky:
@hawkowl, thanks.  It's `docs/installation/howto/optional.rst`.  Also, I forgot the `setup.py` tests...  which did find an issue when I added them.  Fixes incoming.

altendky:
I wasn't sure if the `contextvars` requirements should be specified as `; python_version < "3.7"` or not.  The backport claims to do nothing when installed in 3.7+ so it ought to be optional.  Might still be good though.

hawkowl:
@altendky it would be nice if it was optional -- then it wouldn't result in an unneeded package when you're doing `[all_non_platform]` (which will cause pip to complain if you force install without it)

altendky:
@hawkowl, sorry, I'm not quite following.  Optional as in not in `all_non_platform`?  Or conditional with the `python_version` condition?  I'm not clear what situation pip would complain about.

hawkowl:
With the python_version condition -- then it doesn't add a dependency. Even if the backport does nothing on 3.7, pip and similar tools will complain, and binaries installed by pip (which check for dependencies being installed before running) will not work if someone uninstalls the contextvars backport after installing all_non_platform (I think -- I might be wrong about this).

I think it's worth not downloading if it's not needed, anyway, since it's still a package to download and install.

altendky:
Sure, I'll add the condition. Thanks for the explanation.

PR #1271, opened 2020-05-12
Add more type annotations

rodrigc:
https://twistedmatrix.com/trac/ticket/9827

Add some type annotations which eliminate mypy errors such as:

```
src/twisted/words/protocols/jabber/jid.py:108:1: error: Need type annotation for '__internJIDs' (hint: "__internJIDs: Dict[<type>, <type>] = ...")  [var-annotated]
    __internJIDs = {}
```

### DIFF
--- a/src/twisted/internet/fdesc.py
+++ b/src/twisted/internet/fdesc.py
@@ -10,9 +10,11 @@ Utility functions for dealing with POSIX file descriptors.
 import os
 import errno
 try:
-    import fcntl
+    import fcntl as _fcntl
 except ImportError:
     fcntl = None
+else:
+    fcntl = _fcntl
 
 # twisted imports
 from twisted.internet.main import CONNECTION_LOST, CONNECTION_DONE
--- a/src/twisted/internet/process.py
+++ b/src/twisted/internet/process.py
@@ -10,13 +10,6 @@ Do NOT use this module directly - use reactor.spawnProcess() instead.
 Maintainer: Itamar Shtull-Trauring
 """
 
-
-from twisted.python.runtime import platform
-
-if platform.isWindows():
-    raise ImportError(("twisted.internet.process does not work on Windows. "
-                       "Use the reactor.spawnProcess() API instead."))
-
 import errno
 import gc
 import os
@@ -26,15 +19,8 @@ import stat
 import sys
 import traceback
 
-try:
-    import pty
-except ImportError:
-    pty = None
-
-try:
-    import fcntl, termios
-except ImportError:
-    fcntl = None
+from typing import Callable, Dict, Optional
+from twisted.python.runtime import platform
 
 from zope.interface import implementer
 
@@ -46,11 +32,32 @@ from twisted.internet.main import CONNECTION_LOST, CONNECTION_DONE
 from twisted.internet._baseprocess import BaseProcess
 from twisted.internet.interfaces import IProcessTransport
 
+if platform.isWindows():
+    raise ImportError(("twisted.internet.process does not work on Windows. "
+                       "Use the reactor.spawnProcess() API instead."))
+
+try:
+    import pty as _pty
+except ImportError:
+    pty = None
+else:
+    pty = _pty
+
+try:
+    import fcntl as _fcntl
+    import termios
+except ImportError:
+    fcntl = None
+else:
+    fcntl = _fcntl
+
 # Some people were importing this, which is incorrect, just keeping it
 # here for backwards compatibility:
 ProcessExitedAlready = error.ProcessExitedAlready
 
-reapProcessHandlers = {}
+reapProcessHandlers = {}  # type: Dict[int, Callable]
+
+
 
 def reapAllProcesses():
     """
@@ -281,7 +288,7 @@ class _BaseProcess(BaseProcess, object):
     """
     Base class for Process and PTYProcess.
     """
-    status = None
+    status = None  # type: Optional[int]
     pid = None
 
     def reapProcess(self):
--- a/src/twisted/internet/protocol.py
+++ b/src/twisted/internet/protocol.py
@@ -11,6 +11,7 @@ Twisted.  The Protocol class contains some introductory material.
 
 
 import random
+from typing import Optional, Type
 from zope.interface import implementer
 
 from twisted.python import log, failure, components
@@ -29,7 +30,7 @@ class Factory:
     """
 
     # Put a subclass of Protocol here:
-    protocol = None
+    protocol = None  # type: Optional[Type['Protocol']]
 
     numPorts = 0
     noisy = True
--- a/src/twisted/internet/test/connectionmixins.py
+++ b/src/twisted/internet/test/connectionmixins.py
@@ -10,6 +10,7 @@ Various helpers for tests for connection-oriented transports.
 import socket
 
 from gc import collect
+from typing import Optional
 from weakref import ref
 
 from zope.interface.verify import verifyObject
@@ -291,8 +292,7 @@ class ConnectionTestsMixin(object):
     implementations.
     """
 
-    # This should be a reactormixins.EndpointCreator instance.
-    endpoints = None
+    endpoints = None  # type: Optional[EndpointCreator]
 
 
     def test_logPrefix(self):
--- a/src/twisted/internet/test/reactormixins.py
+++ b/src/twisted/internet/test/reactormixins.py
@@ -20,7 +20,9 @@ __all__ = ['TestTimeoutError', 'ReactorBuilder', 'needsRunningReactor']
 import os
 import signal
 import time
-from typing import Dict, Type, Union
+from typing import Dict, Sequence, Optional, Type, Union
+
+from zope.interface import Interface
 
 from twisted.trial.unittest import SynchronousTestCase, SkipTest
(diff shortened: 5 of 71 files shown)


rodrigc:
This PR eliminates about 200 errors from mypy.
There is still a ways to go before turning mypy on by default as part of CI for Twisted, but we are getting closer...

src/twisted/internet/fdesc.py
```diff
@@ -10,9 +10,11 @@
 import os
 import errno
 try:
-    import fcntl
+    import fcntl as _fcntl
 except ImportError:
     fcntl = None
+else:
+    fcntl = _fcntl
```
wsanchez, line 17: What's going on with this pattern? It's in a few places

src/twisted/internet/process.py
```diff
@@ -281,7 +288,7 @@ class _BaseProcess(BaseProcess, object):
     """
     Base class for Process and PTYProcess.
     """
-    status = None
+    status = None  # type: int
```
wsanchez: If None is valid, this should be `Optional[int]`

src/twisted/internet/protocol.py
```diff
@@ -29,7 +30,7 @@ class Factory:
     """
 
     # Put a subclass of Protocol here:
-    protocol = None
+    protocol = None  # type: Type['Protocol']
```
wsanchez: `Optional[…]`

src/twisted/internet/test/connectionmixins.py
```diff
@@ -291,8 +291,7 @@ class ConnectionTestsMixin(object):
     implementations.
     """
 
-    # This should be a reactormixins.EndpointCreator instance.
-    endpoints = None
+    endpoints = None  # type: EndpointCreator
```
wsanchez: `Optional[EndpointCreator]`

src/twisted/internet/test/reactormixins.py
```diff
@@ -171,7 +173,7 @@ class ReactorBuilder:
 
     reactorFactory = None
     originalHandler = None
-    requiredInterfaces = None
+    requiredInterfaces = None  # type: Sequence[Type[Interface]]
```
wsanchez: `Optional[Sequence[Type[Interface]]]`

src/twisted/internet/test/test_tcp.py
```diff
@@ -2832,7 +2833,7 @@ class AbortConnectionMixin(object):
     Unit tests for L{ITransport.abortConnection}.
     """
     # Override in subclasses, should be an EndpointCreator instance:
-    endpoints = None
+    endpoints = None  # type: EndpointCreator
```
wsanchez: `Optional[EndpointCreator]`

src/twisted/internet/unix.py
```diff
@@ -67,7 +71,7 @@ class _SendmsgMixin(object):
         registered producer, if there is one.
     """
 
-    _writeSomeDataBase = None
+    _writeSomeDataBase = None  # type: Type[FileDescriptor]
```
wsanchez: `Optional[Type[FileDescriptor]]`

src/twisted/mail/pop3client.py
```diff
@@ -1261,4 +1262,6 @@ def quit(self):
         """
         return self.sendShort(b'QUIT', None)
 
-__all__ = []
+
+
+__all__ = []  # type: List[str]
```
wsanchez, line 1267: Seems like a `mypy` bug if this is needed…

src/twisted/mail/test/test_imap.py
```diff
@@ -49,7 +50,8 @@
 try:
     from twisted.test.ssl_helpers import ClientTLSContext, ServerTLSContext
 except ImportError:
-    ClientTLSContext = ServerTLSContext = None
+    ClientTLSContext = None  # type: ignore[assignment,misc]
+    ServerTLSContext = None  # type: ignore[assignment,misc]
```
wsanchez, line 54: `Optional[…]`

src/twisted/protocols/htb.py
```diff
@@ -41,8 +42,8 @@ class Bucket:
     @type rate: C{int}
     """
 
-    maxburst = None
-    rate = None
+    maxburst = None  # type: int
+    rate = None  # type: int
```
wsanchez: `Optional[int]`

src/twisted/python/formmethod.py
```diff
@@ -162,9 +174,10 @@ def coerce(self, val):
         return result
 
 
+
 class Float(Argument):
 
-    defaultDefault = None
+    defaultDefault = None  # type: float
```
wsanchez: `Optional[float]`

src/twisted/python/formmethod.py
```diff
@@ -271,15 +291,20 @@ def coerce(self, file):
...
+        raise ValueError
     return x
 
+
+
 class Date(Argument):
     """A date -- (year, month, day) tuple."""
 
-    defaultDefault = None
+    defaultDefault = None  # type: Tuple[int, int, int]
```
wsanchez: `Optional[…]`

src/twisted/python/usage.py
```diff
@@ -559,7 +560,8 @@ class Completer(object):
     This class produces no completion matches itself - see the various
     subclasses for specific completion functionality.
     """
-    _descr = None
+    _descr = None  # type: str
```
wsanchez: `Optional[str]`

src/twisted/runner/inetdconf.py
```diff
@@ -46,7 +46,7 @@ class SimpleConfFile:
     """
 
     commentChar = '#'
-    defaultFilename = None
+    defaultFilename = None  # type: str
```
wsanchez: `Optional[str]`

wsanchez (approved):
Looks good.

I think optionals should be annotated as such (I'd want to set `strict_optional=True` for `mypy`), but if that's not how Twisted wants to roll, I guess that's not a blocker.

src/twisted/internet/fdesc.py
rodrigc: I took this pattern from here: https://github.com/python/mypy/issues/1297#issuecomment-508593494

I couldn't figure out a cleaner way to eliminate the mypy error, other than doing:

```
try:
    import fcntl
except ImportError:
    fcntl = None    # type: ignore
```

Maybe it would be better to do this?

```
from twisted.python.reflect import requireModule

fcntl = requireModule("fcntl")
```

mthuurne: In the `requireModule` approach, would mypy still be able to use type hints/stubs for the imported module? My guess is it wouldn't, in which case this approach would reduce the reach of the type checking.

rodrigc: I don't know how to specify type hints for a module.  I've looked for examples and asked on https://gitter.im/python/typing but haven't found an example.

mthuurne: I don't know if it's even possible to attach a package name to a dynamically imported module in a way that mypy understands. I tried adding `assert fcntl.__name__ == 'fcntl'`, but that doesn't change the way mypy sees the module.

The `fcntl = _fcntl` pattern you used might be the best option, for now.

rodrigc: Well if I do this from the repl:

```
>>> import fcntl
>>> type(fcntl)
<class 'module'>
```

I couldn't figure out a way to add an annotation for `class 'module'`

mthuurne: The annotation for "a module" is `types.ModuleType`. However, that tells mypy that it's dealing with an arbitrary module, not one specific module for which it might know the type signatures of its contents.

mthuurne: To clarify things with an example:
```py
from types import ModuleType
from twisted.python.reflect import requireModule

import fcntl as direct
indirect: ModuleType = requireModule('fcntl')

reveal_type(direct)
reveal_type(indirect)

direct.flock()
indirect.flock()
```
For this code, mypy outputs:
```
testcase.py:7:13: note: Revealed type is '_importlib_modulespec.ModuleType'
testcase.py:8:13: note: Revealed type is '_importlib_modulespec.ModuleType'
testcase.py:10:1: error: Too few arguments for "flock"  [call-arg]
testcase.py:11:1: error: Module has no attribute "flock"  [attr-defined]
```
So while the revealed type is equal, a direct import results in mypy knowing that `flock` exists and what its signature is, while it does not have that knowledge if the import is done via `requireModule`.

rodrigc: How would you add type annotations for the return type of `requireModule` here: https://github.com/twisted/twisted/blob/4e42bf4928266b5135be6d66f1532d52317d8731/src/twisted/python/reflect.py#L179  ?

mthuurne: The best I can think of is:
```py
from types import ModuleType
from typing import TypeVar, Union
T = TypeVar('T')
def requireModule(name: str, default: T = None) -> Union[ModuleType, T]:
```
That will not make it easy to annotate any code using `requireModule`, but a better solution might not exist.

rodrigc:
@wsanchez OK, I added `Optional` to all the places that you suggested.

PR #1272, opened 2020-05-14
Allow @unittest.skipIf decorator to specify that a test class should be skipped

rodrigc:
https://twistedmatrix.com/trac/ticket/9829

### DIFF
--- a/src/twisted/conch/test/test_cftp.py
+++ b/src/twisted/conch/test/test_cftp.py
@@ -7,10 +7,17 @@ Tests for L{twisted.conch.scripts.cftp}.
 """
 
 import locale
-import time, sys, os, operator, getpass, struct
+import getpass
+import operator
+import os
+import struct
+import sys
+import time
+from unittest import skipIf
 from io import BytesIO
 
 from twisted.python.filepath import FilePath
+from twisted.python.procutils import which
 from twisted.python.reflect import requireModule
 from zope.interface import implementer
 
@@ -18,7 +25,6 @@ pyasn1 = requireModule('pyasn1')
 cryptography = requireModule('cryptography')
 unix = requireModule('twisted.conch.unix')
 
-_reason = None
 if cryptography and pyasn1:
     try:
         from twisted.conch.scripts import cftp
@@ -46,7 +52,14 @@ from twisted.trial.unittest import TestCase
 
 
 
+skipTests = False
+if None in (unix, cryptography, pyasn1,
+            interfaces.IReactorProcess(reactor, None)):
+    skipTests = True
 
+
+
+@skipIf(skipTests, "don't run w/o spawnProcess or cryptography or pyasn1")
 class SSHSessionTests(TestCase):
     """
     Tests for L{twisted.conch.scripts.cftp.SSHSession}.
@@ -346,6 +359,7 @@ class InMemoryRemoteFile(BytesIO):
 
 
 
+@skipIf(skipTests, "don't run w/o spawnProcess or cryptography or pyasn1")
 class StdioClientTests(TestCase):
     """
     Tests for L{cftp.StdioClient}.
@@ -945,6 +959,7 @@ class CFTPClientTestBase(SFTPTestBase):
 
 
 
+@skipIf(skipTests, "don't run w/o spawnProcess or cryptography or pyasn1")
 class OurServerCmdLineClientTests(CFTPClientTestBase):
     """
     Functional tests which launch a SFTP server over TCP on localhost and check
@@ -1320,6 +1335,7 @@ class OurServerCmdLineClientTests(CFTPClientTestBase):
 
 
 
+@skipIf(skipTests, "don't run w/o spawnProcess or cryptography or pyasn1")
 class OurServerBatchFileTests(CFTPClientTestBase):
     """
     Functional tests which launch a SFTP server over localhost and checks csftp
@@ -1424,6 +1440,8 @@ exit
 
 
 
+@skipIf(skipTests, "don't run w/o spawnProcess or cryptography or pyasn1")
+@skipIf(not which('sftp'), "no sftp command-line client available")
 class OurServerSftpClientTests(CFTPClientTestBase):
     """
     Test the sftp server against sftp command line client.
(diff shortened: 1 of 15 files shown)


src/twisted/conch/test/test_cftp.py
```diff
@@ -46,7 +53,9 @@
 
 
 
-
+@skipIf(None in (unix, cryptography, pyasn1,
+                 interfaces.IReactorProcess(reactor, None)),
+        "don't run w/o spawnProcess or cryptography or pyasn1")
```
adiroiban: I see that the condition is repeated... maybe keep the similar condition in a variable

```suggestion
cftp_skip_conditions = (
    None in (unix, cryptography, pyasn1, interfaces.IReactorProcess(reactor, None)),
    "don't run w/o spawnProcess or cryptography or pyasn1",
    )
@skipIf(*cftp_skip_conditions)
```

src/twisted/test/test_amp.py
```diff
@@ -2008,8 +2012,6 @@ def test_negotiationFailedByClosing(self):
         # reasonable.
         self.assertFailure(d, error.PeerVerifyError)
 
-    skip = skipSSL
```
adiroiban, line 2011: I cant add a comment for skipSSL definition... but I thin that line 35 to line 38 can be removed as skipSSL is no longer used ... or should no longer be used.

src/twisted/trial/_synctest.py
```diff
@@ -1015,10 +1015,17 @@ def getSkip(self):
...
+        and C{__unittest_skip_why__} attributes which are set by the standard
+        library L{unittest.skip} function.
         Returns L{None} if it cannot find anything. See L{TestCase} docstring
         for more details.
         """
-        return util.acquireAttribute(self._parents, 'skip', None)
+        skipReason = util.acquireAttribute(self._parents, 'skip', None)
+        if skipReason is None:
+            if getattr(self, "__unittest_skip__", None):
+                skipReason = getattr(self, "__unittest_skip_why__", None)
```
adiroiban: Is this correct?

It looks like if it has `__unittest_skip__ == True and __unittest_skip_why__ == "" `  it will not skip

```suggestion
           # See if we have a stdlib skip condition.
            skipReason = getattr(self, "__unittest_skip__", None)
            if skipReason:
                # See if we have more details about the skip.
                skipReason = getattr(self, "__unittest_skip_why__", True) or True
```

adiroiban (changes requested):
Changes looks good. Thanks. I am very happy with the decorators. Much easier to read.

Where multiple tests are skipped with the same condition, it would be nice to have the condition stored in a variable and shared between all the tests.
This is kind of a regression from trunk

I see that python stdlib has "reason" as a mandatory argument... but it will allow you to set it as empty.. and I think that trial should still skip when reason is empty.

```
>>> import unittest
>>> unittest.skipIf(True, '')
```

src/twisted/test/test_amp.py
rodrigc: skipSSL is still used in this file.  I didn't remove it here, because I have a separate PR where I used @skipIf to convert functions: https://github.com/twisted/twisted/pull/1264

src/twisted/trial/_synctest.py
rodrigc: I implemented it like that to be consistent with the behavior in
* unittest: https://github.com/python/cpython/blob/515fce4fc4bb0d2db97b17df275cf90640017f56/Lib/unittest/case.py#L569
* pytest: https://github.com/pytest-dev/pytest/blob/952762207a2de76bf2eb254dee16f3b53d79a77b/src/_pytest/unittest.py#L92

src/twisted/test/test_amp.py
adiroiban: ok.

src/twisted/trial/_synctest.py
adiroiban: What I can read in cpython, is that if you have `__unittest_skip__` , `self._addSkip` is always called... with or without a reason.

But from what I can see in your implementation, `__unittest_skip_why__` will overwrite `__unittest_skip__`

So if you have `__unittest_skip__ == True ` and `__unittest_skip_why__ == False` the implementation will end up with   skipReason = False and the test is not skipped.

rodrigc: You are misreading unittest.  `__unittest_skip__` is a boolean value, and `__unittest_skip_why__` is a string value.  Look at the pytest implementation as well.

src/twisted/conch/test/test_cftp.py
rodrigc: OK I changed this condition by using a variable.

src/twisted/trial/_synctest.py
rodrigc: I changed the implementation here slightly to match what was done in pytest.

adiroiban: My expectation is that if I have `@unittest.skipIf(True, '')` the test is skipped.
From what I can read in cpython code, this test is skipped.
From what I can read from the code in this branch, the test is not skipped.

rodrigc: You are reading the unittest code wrong.  That's not how it behaves.
For this example:

```
from unittest import TestCase
from unittest import skipIf, main

class TestStringMethods(TestCase):

    @skipIf(True, '')
    def test_upper(self):
        print("RAN test_upper")
        self.assertEqual('foo'.upper(), 'FOO')

    @skipIf(False, '')
    def test_isupper(self):
        print("RAN test_isupper")
        self.assertTrue('FOO'.isupper())
        self.assertFalse('Foo'.isupper())

    @skipIf(True, '')
    def test_split(self):
        print("RAN test_split")
        s = 'hello world'
        self.assertEqual(s.split(), ['hello', 'world'])
        # check that s.split fails when the separator is not a string
        with self.assertRaises(TypeError):
            s.split(2)


if __name__ == '__main__':
    main()
```

Only the first argument to `skipIf` is used for skipping.  The second argument, even if it is an empty string, does not control whether skipping occurs or not.

rodrigc: I updated this PR, and added more testcases with:

```
@skipIf(False, "")

@skipIf(True, "")
```

on both classes and functions.

src/twisted/trial/_synctest.py
```diff
@@ -1022,15 +1023,25 @@ def shortDescription(self):
...
+        Returns (L{False}, L{None}) if it cannot find anything.
+        See L{TestCase} docstring for more details.
+        """
+        skipReason = util.acquireAttribute(self._parents, 'skip', None)
+        doSkip = skipReason is not None
+        if skipReason is None:
+            doSkip = getattr(self, "__unittest_skip__", False)
+            if doSkip:
+                skipReason = getattr(self, "__unittest_skip_why__", "")
+        return (doSkip, skipReason)
```
adiroiban, line 1044: Can we have something like this and not change the return value for getSkip  ?

It will no longer return `None` ...but `False` instead, but I think that this can reduce downstream changes. 

```suggestion
        return skipReason or doSkip
```

Now, this is only testing infrastructure public dode... so if something breaks, it will break at the testing phase and not in production.

So feel free to merge as it is.

i am very happy with the new automated tests :)

adiroiban (approved):
Looks  good. Thank!
Just a minor comment.

Feel free to merge as you wish :)

PR #1274, opened 2020-05-23
SMB initial version 

ihaywood3:
This is the minimal proof-of-concept SMB component . All it can do is negotiate a session and authenticate, no actual file access yet but its already a large commit for review (sorry)

https://twistedmatrix.com/trac/ticket/9818


src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,209 @@
...
+    @param epoch: seconds since 1/1/1970
+    @type epoch: L{float}
+
+    @rtype: L{int}
+    """
+    return int(epoch * 10000000.0) + 116444736000000000
+
+
+
+class GeneralStruct:
```
glyph: Can we use `attrs` metadata for this to avoid inheritance?

src/twisted/protocols/_smb/ntlm.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def ntlm_challenge(self, data):
+        raise base.SMBError("invalid to send NTLM challenge to a server")
+
+    def ntlm_negotiate(self, data):
+        neg = NegType(data)
+        flags = flags2set(neg.flags)
+        log.debug("NTLM NEGOTIATE")
+        log.debug("--------------")
+        log.debug("Flags           %r" % list(flags))
```
glyph: This is not how you use Twisted's logger. The string being passed is a *format* string, not a message.  So this should be:

`log.debug("Flags {flags}", flags=list(flags))`

src/twisted/protocols/_smb/security_blob.py
```diff
@@ -0,0 +1,285 @@
...
+"""Implement the "security buffer" an intricate sub-protocol
+used for authentication
+
+Blobs implemented as DER-encoded ASN.1 objects
+which in turn encapsulate NTLMv2 challenge-response authentication
+
+
+much of code adapted from Mike Teo's pysmb """
+
+from __future__ import absolute_import, division
```
glyph: It's python 3 only now, no need for this :).

glyph:
One other thing that I am noticing as I'm looking over these commits: they're not properly associated with an email address verified on your github account, which I think you probably want to fix.  (You can add & verify additional addresses on your github account at https://github.com/settings/emails )

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+    def __init__(self, msg, ntstatus=0xC0000001):
+        self.msg = msg
+        self.ntstatus = ntstatus
+
+    def __str__(self):
+        return "%s 0x%08x" % (self.msg, self.ntstatus)
+
+
+
+def u2nt_time(epoch):
```
glyph: The twisted coding convention would have this named as `unixToNTTime`

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+    Convert UNIX epoch time to NT filestamp
+    quoting from spec: The FILETIME structure is a 64-bit value
+    that represents the number of 100-nanosecond intervals that
+    have elapsed since January 1, 1601, Coordinated Universal Time.
+
+    @param epoch: seconds since 1/1/1970
```
glyph: Nitpick: can you phrase this as an ISO timestamp? `1970-01-01`, specifically.

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def __len__(self):
+        return self._size
+
+    def __bytes__(self):
+        return self.pack()
+
+
+
+def nstruct(fields):
```
glyph: This idiom is going to be absolute hell to get mypy to typecheck; it's going to complain about dynamic subclasses.  Could we possibly do this with annotations and attrs fields instead of inventing a new thing here?

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+    dynamically subclass GeneralStruct with a speeific binary structure
```
glyph: "specific"

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+    while n == 0 or n in d:
+        n, = struct.unpack("<I", secureRandom(4, True))
+    d[n] = val
+    return n
+
+
+
+class SMBPacketReceiver(protocol.Protocol):
+    """
+    basic SMB 2.0 packets over TCP have a 4-byte header: null byte
```
glyph: Twisted docstring style would be `"An L{SMBPacketReceiver} is a wire protocol parser for the SMB framing mechanism, which is...."`

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
+# Copyright (c) Twisted Matrix Laboratories.
```
glyph: I appreciate the need for this binary protocol-parsing toolkit and I have a bunch of notes about stuff like mypy and attrs and whatnot, but I think the main important change is to ensure that it's an implementation detail, and that we think carefully about what the shape of the public interface for this stuff is; i.e. let's not include this byte-unpacking framework in the public API for SMB initially.

I realize that we're already in `_smb` since this is experimental, but I would suggest putting this in `_smb/_base.py` to clarify that it's not *intended* to be public API surface even after the package is more complete.

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+def int32key(d, val):
+    """
+    generate a new random key for a dictionary
+    @param d: dictionary with 32-bit int keys
+    @type d: L{dict}
+    @param val: new dictionary value
+    @rtype: L{int}
+    """
+    assert len(d) < 0xc0000000  # otherwiae dict so big hard to find keys
+    n, = struct.unpack("<I", secureRandom(4, True))
```
glyph: `4` is a magic value; I think this would be more cleanly expressed like this:

```python3
_leint = struct.Struct("<I")

def int32key(d, val):
    n = 0
    while n == 0 or n in d:
        [n] = _leint.unpack(secureRandom(_leint.size, True))
    d[n] = val
    return n
```

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def dataReceived(self, data):
+        self.data += data
+        self._processData()
+
+    def _processData(self):
+        if len(self.data) < 5:
+            return
+        x, y = struct.unpack("!xBH", self.data[:4])
+        size = (x << 16) + y
```
glyph: once again "4" and "16" could be made into `.size` attributes of `Struct` objects

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def packetReceived(self, packet):
+        """
+        called for each complete packet received over network
+        override in descendants
+
+        @param packet: raw packet data
+        @type packet: L{bytes}
+        """
+        pass
```
glyph: `pass` is unnecessary if you've got a docstring

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+        @param dara: packet to send
+        @type data: L{bytes}
+        """
+        size = len(data)
+        assert size < 0xffffff
+        x = (size & 0xff0000) >> 16
+        y = size & 0xffff
+        self.transport.write(struct.pack("!BBH", 0, x, y) + data)
+
+    def packetReceived(self, packet):
```
glyph: I know this stack-of-superclasses is a pattern throughout twisted but it's really bad 😬.  We should have a `self._packetReceiver` object and this should be `self._packetReceiver.packetReceived` inside `dataReceived`, then delegate on down for each level of transformation.  Avoiding subclassing as an interface avoids lots of namespace-collision problems down the road (and allows for a super narrow top-level interface that doesn't expose any of these internals.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+"""Implement Microsoft's Server Message Block protocol"""
+
+import struct
+import binascii
+import uuid
+import time
+from collections import namedtuple
+
+from twisted.protocols._smb import base, security_blob
+from twisted.protocols._smb.interfaces import (ISMBServer, IFilesystem, IPipe,
```
glyph: To avoid having a bunch of `interfaces.py` files and the potential import confusion that might provoke, we have generally gone with `<whatever>.i<whatever>` for interface module names; so in this case, `from twisted.protocols._smb import ismb`

src/twisted/protocols/_smb/base.py
```diff
@@ -0,0 +1,207 @@
...
+        unpack data according to the pattern
+        exceas data goes in buffer
+
+        @param data: the data
+        @type data: L{bytes}
+        """
+        ret = struct.unpack(self._pattern, data[:self._size])
+        for i in range(len(ret)):
+            self._values[self._fields_names[i]] = ret[i]
+        self.buffer = data[self._size:]
```
glyph: It really looks like you want to be using a `struct.Struct` here rather than the function versions

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+STATUS_LOCK_NOT_GRANTED = 0xC0000055  # generated when requesting lock
+STATUS_LOGON_FAILURE = 0xC000006D
+STATUS_DISK_FULL = 0xC000007F
+STATUS_ACCOUNT_RESTRICTION = 0xC000006E
+STATUS_PASSWORD_EXPIRED = 0xC0000071
+STATUS_ACCOUNT_DISABLED = 0xC0000072
+STATUS_FILE_INVALID = 0xC0000098
+STATUS_DEVICE_DATA_ERROR = 0xC000009C
+STATUS_BAD_NETWORK_NAME = 0xC00000CC  # = "share not found"
+
```
glyph: Do we want to group these into an `Enum`, maybe?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        self.message_id = 0
+        self.tree_id = 0
+        self.session_id = 0
+        self.async_id = 0
+        self.first_session_setup = True
+        self.is_async = False
+        self.is_related = False
+        self.is_signed = False
+        self.blob_manager = security_blob.BlobManager(factory.domain)
+        self.trees = {}
```
glyph: Surely we would want all of these attributes to be private?

glyph: Also, Twisted naming convention would be `isAsync`, `isRelated`, `isSigned`, etc

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+ACCESS_SYSTEM_SECURITY = 0x01000000
+MAXIMUM_ALLOWED = 0x02000000
+GENERIC_ALL = 0x10000000
+GENERIC_EXECUTE = 0x20000000
+GENERIC_WRITE = 0x40000000
+GENERIC_READ = 0x80000000
+
+
+
+class SMBConnection(base.SMBPacketReceiver):
```
glyph: Yeah this subclassing pattern just makes me so sad to see in new code. I'm sorry to have set such a bad example over the years 😞 .  If my earlier suggestion was not clear and you'd like to discuss alternate implementation approaches I'd be happy to do that.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        log.debug("flags           0x{flags:x} {s}", flags=self.hdr_flags, s=s)
+        log.debug("next command    0x{nc:x}", nc=self.next_command)
+        log.debug("message ID      0x{mid:x}", mid=self.message_id)
+        log.debug("session ID      0x{sid:x}", sid=self.session_id)
+        if self.is_async:
+            log.debug("async ID        0x{aid:x}", aid=self.async_id)
+        else:
+            log.debug("tree ID         0x{tid:x}", tid=self.tree_id)
+        log.debug("signature       {sig}",
+                  sig=binascii.hexlify(self.signature))
```
glyph: This should really be one large log message, rather than tons of tiny individual ones.  Thinking from the perspective of a log *consumer*, this is a single event, and you wouldn't want to write code to try to heuristically figure out where it begins and ends.  We have had much bitter experience with the annoying split-up message in `defer.py` that does this, that should be a single message.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        try:
+            name, req_type, resp_type = COMMANDS[self.hdr_command]
+        except IndexError:
```
glyph: A better idiom for this type of check is `if self.hdr_command in COMMANDS:`; that way you don't involve exception handling in normal operation.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+                if hasattr(self, func) and req_type:
+                    req = req_type(packet[64:])
+                    resp = resp_type()
+                    getattr(self, func)(req, resp)
+                else:
+                    log.error("command '{cmd}' not implemented",
+                              cmd=COMMANDS[self.hdr_command][0])
+                    self.error_response(STATUS_NOT_IMPLEMENTED)
+            except base.SMBError as e:
+                log.error(str(e))
```
glyph: Again, the arg to `log.error` is a format string, not just a string, and so you have to do this more like `log.error("SMB error: {e}", e=e)`

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+                self.error_response(e.ntstatus)
+            except BaseException:
+                log.failure("in {cmd}", cmd=COMMANDS[self.hdr_command][0])
+                self.error_response(STATUS_UNSUCCESSFUL)
+        if self.is_related and self.next_command > 0:
+            self.packetReceived(packet[self.next_command:])
```
glyph: Tail recursion is risky if these packets might get big; you can end up blowing the recursion limit. Could this be unwound to be iterable instead?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        @param command: command name or id, defaults to same as received packet
+        @type command: L{str} or L{int}
+
+        @param payload: the after-header data
+        @type payload: L{bytes}
+
+        @param status: packet status, an NTSTATUS code
+        @type status: L{int}
+        """
+        payload = bytes(payload)
```
glyph: If the payload is required to be bytes why is this necessary?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        log.debug("dialect count   {dc}", dc=req.dialect_count)
+        s = ""
+        if self.signing_enabled:
+            s += "ENABLED "
+        if self.signing_required:
+            s += "REQUIRED"
+        log.debug("signing         0x{sm:02x} {s}", sm=req.security_mode, s=s)
+        log.debug("client UUID     {uuid}", uuid=self.client_uuid)
+        log.debug("dialects        {dlt!r}",
+                  dlt=["%04x" % x for x in dialects])
```
glyph: Again, one big log message :)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        self.negotiate_response(dialects)
+
+    def error_response(self, ntstatus):
+        self.send_with_header(b'\x09\0\0\0\0\0\0\0', status=ntstatus)
```
glyph: Can this magic value be pulled out to a constant?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+    def smb_session_setup(self, req, resp):
+        blob = req.buffer[req.offset - len(req) - 64:req.offset - len(req) -
+                          64 + req.buflen]
+        log.debug("SESSION SETUP")
+        log.debug("-------------")
+        log.debug("Size             {sz}", sz=req.size)
+        log.debug("Security mode    0x{sm:08x}", sm=req.security_mode)
+        log.debug("Capabilities     0x{cap:08x}", cap=req.capabilities)
+        log.debug("Channel          0x{chl:08x}", chl=req.channel)
+        log.debug("Prev. session ID 0x{pid:016x}", pid=req.prev_session_id)
```
glyph: (big log message)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        _, self.avatar, self.logout_thunk = t
+        resp.buffer = self.blob_manager.generateAuthResponseBlob(True)
+        log.debug("successful login")
+        self.session_setup_response(resp, STATUS_SUCCESS)
+
+    def _eb_login(self, failure, resp):
+        log.debug(failure.getTraceback())
+        resp.buffer = self.blob_manager.generateAuthResponseBlob(False)
+        self.session_setup_response(resp, STATUS_LOGON_FAILURE)
+
```
glyph: I think it might be better to define these inline rather than as private methods; private methods that are only supposed to be called as specific callbacks in specific places just pollute the class namespace and add confusion for future maintainers about when & where they're OK to call.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+    def session_setup_response(self, resp, ntstatus):
+        log.debug("session_setup_response")
+        resp.flags = 0
+        if self.blob_manager.credential == ANONYMOUS:
+            resp.flags |= SESSION_FLAG_IS_NULL
+        resp.size = 9
+        resp.offset = 72
```
glyph: `72` is one of the more magic values I've seen so far here :)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        assert req.size == 9
+        path = req.buffer[req.offset - len(req) - 64:req.offset - len(req) -
+                          64 + req.buflen]
+        path = path.decode("utf-16le").split("\\")[-1]
+        try:
+            share = self.avatar.getShare(path)
+        except KeyError:
+            self.error_response(STATUS_BAD_NETWORK_NAME)
+            return
+        if IFilesystem.providedBy(share):
```
glyph, line 639: Love the use of ZI inspection here!

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+            # FUTURE: select these values from share object
+            resp.flags = SHAREFLAG_MANUAL_CACHING
+            resp.capabilities = 0
+            resp.max_perms = (FILE_READ_DATA | FILE_WRITE_DATA
+                              | FILE_APPEND_DATA | FILE_READ_EA | FILE_WRITE_EA
+                              | FILE_DELETE_CHILD | FILE_EXECUTE
+                              | FILE_READ_ATTRIBUTES | FILE_WRITE_ATTRIBUTES
+                              | DELETE | READ_CONTROL | WRITE_DAC | WRITE_OWNER
+                              | SYNCHRONIZE)
+        elif IPipe.providedBy(share):
```
glyph: Can it only be one? What happens if it provides more than one? Do we have any error reporting to the user?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+    def smb_tree_connect(self, req, resp):
+        if self.avatar is None:
+            self.error_response(STATUS_ACCESS_DENIED)
+            return
+        assert req.size == 9
+        path = req.buffer[req.offset - len(req) - 64:req.offset - len(req) -
+                          64 + req.buflen]
+        path = path.decode("utf-16le").split("\\")[-1]
+        try:
+            share = self.avatar.getShare(path)
```
glyph: LOVE to see some idiomatic cred usage :)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,519 @@
...
+        except KeyError:
```
glyph: Mmm… `KeyError` isn't a great error here, you probably want a custom exception. Also should this really be raised synchronously? This seems like the sort of thing that might want an async result...

src/twisted/protocols/_smb/__init__.py
```diff
@@ -0,0 +1,10 @@
+# -*- test-case-name: twisted.protocols._smb.tests -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+"""
+Server Message Block protocol implementation
+"""
+
+from twisted.protocols._smb.core import SMBFactory
+
+__all__ = ['SMBFactory']
```
glyph: Why is just this one name promoted up to the package namespace?

src/twisted/protocols/_smb/ntlm.py
```diff
@@ -0,0 +1,380 @@
...
+
+    @rtype: L{int}
+
+    @type s: L{set} of L{str}
+    """
+    flags = 0
+    for i in s:
+        flags |= FLAGS[i]
+    return flags
+
```
glyph: I think you might get a lot of what you want here out of `IntFlag`

glyph (commented):
There's a lot here, so I'm just providing another partial review as I had time to look through some more of the files :)

src/twisted/protocols/_smb/base.py
ihaywood3: please look at latest commit which uses attrs based approach to binary packing /unpacking 

if it's ok I'll move the code across in ntlm.py and core.py

src/twisted/protocols/_smb/__init__.py
ihaywood3: simply so the user can access as twisted.protocols.smb.SMBFactory as opposed to twisted.protocols.smb.core.SMBFactory  Over time other classes required for public API (mainly interfaces) would be added.

src/twisted/protocols/_smb/core.py
ihaywood3: it makes no sense for a share to be more than one type

src/twisted/protocols/_smb/base.py
ihaywood3: new implementation does use Struct

src/twisted/protocols/_smb/base.py
ihaywood3: how does the contained object (PacketReceiver) call out to the container with new incoming packets? Deferred doesn't work because it's for multiple use. Does the container supply a callback ?

src/twisted/protocols/_smb/base.py
ihaywood3: pretty much all the modules are for internal use (heads up, there's more coming, even just to get to browsing a directory) My hope is only `SMBFactory` and a few interfaces form the public API and these are promoted through `__init__.py`

src/twisted/protocols/_smb/base.py
ihaywood3: will be removed if new implementation is acceptable

src/twisted/protocols/_smb/core.py
ihaywood3: `SMBConnection` is now gone most of these are now header fields see `HeaderSync` and are not exposed as instance variables

src/twisted/protocols/_smb/core.py
ihaywood3: it could also be a `GeneralStruct` which had a `__bytes__` method but as this is going away I will remove it

src/twisted/protocols/_smb/core.py
ihaywood3: Each packet type has two "magic" numbers, a request size and a response size, some have a third, the "offset". For reasons I can't fathom they're true magic: nearly but not quite the same as the size of the packet on the wire. If we move to `attrs` based pack/unpack these can be set as defaults in the class definition.

All other instances of "magic" values should be factored out to named constants now.

src/twisted/protocols/_smb/ntlm.py
ihaywood3: Correct. But  `IntFlag` was introduced in 3.6 so it is not yet available, I will shift this code over when 3.5 support requirement is dropped.

src/twisted/protocols/_smb/base.py
ihaywood3: This is now done

src/twisted/protocols/_smb/core.py
ihaywood3: The code now throws `AssertionError` if a share tries to implement more than one type.

src/twisted/protocols/_smb/base.py
ihaywood3: `core.py` has been moved to a composition based style and I've removed the `SMBConnection`

src/twisted/protocols/_smb/core.py
ihaywood3: As above I have removed `SMBConnection` entirely

glyph (commented):
@ihaywood3 Can you resolve the merge conflicts, and get the commit status to be green? I'm sorry this is lingering so long, but it's a lot of new code to get through

src/twisted/protocols/_smb/__init__.py
```diff
@@ -0,0 +1,10 @@
+# -*- test-case-name: twisted.protocols._smb.tests -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+"""
+Server Message Block protocol implementation
+"""
+
+from twisted.protocols._smb.core import SMBFactory
+
+__all__ = ["SMBFactory"]
```
glyph, line 10: I do love a carefully-minimized public interface that we'll commit to, but it seems likely that maybe a *bit* more stuff will need to be exported in order to be useful? (This is more of a question, not really a request for changes.)

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+# -*- test-case-name: twisted.protocols._smb.tests -*-
+"""
+base classes for SMB networking
```
glyph, line 5: Minor organizational note: the pattern of "base" modules with base classes that other stuff should inherit from is an unfortunate anti-pattern from the bad old days of treating inheritance as our primary extensibility mechanism.

I will also note that this module of "base classes" appears to contain a bunch of functions and only 2 classes, only one of which appears to be plausibly extensible. So a different name/docstring would seem to be in order.

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+def unpack(cls: Any, data: bytes, offset: int = 0, remainder: int = IGNORE) -> Any:
+    """
+    unpack binary data into an object.
+
+    @param cls: the class, must be decorated with L{attr.s} and have
+    members with the appropriate metadata using the helpers from this
+    module.
+    @type cls: L{type}
```
glyph, line 208: `@type` epydoc field are now redundant with type annotations; pydoctor has supported these since 2020! https://github.com/twisted/pydoctor/issues/136

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    return val
+
+
+IGNORE = 0
+ERROR = 1
+OFFSET = 2
+DATA = 3
+
+
+def unpack(cls: Any, data: bytes, offset: int = 0, remainder: int = IGNORE) -> Any:
```
glyph, line 201: There is a way to properly express "takes a type, returns an instance of that type" with mypy; you don't have to use `Any` here:

```suggestion
T = TypeVar("T")
def unpack(cls: Type[T], data: bytes, offset: int = 0, remainder: int = IGNORE) -> T:
```

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    obj = cls(**kwargs)
+    if remainder <= ERROR:
+        return obj
+    elif remainder == OFFSET:
+        return (obj, offset + strct.size)
+    else:
+        return (obj, data[offset + strct.size :])
+
+
+def smb_fields(cls: Any) -> List[Any]:
```
glyph, line 249: coding standard: we don't use underscores like PEP8 in Twisted
```suggestion
def smbFields(cls: Type[object]) -> List[Any]:
```

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    return [i for i in attr.fields(cls) if SMB_METADATA in i.metadata]
+
+
+def _get_struct(cls: Any) -> struct.Struct:
```
glyph, line 253:
```suggestion
def _getStruct(cls: Any) -> struct.Struct:
```

glyph: Also, minor naming thing: `getX` is a bit vague, I usually like to title these something like `structFromType` to be more descriptive.

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    try:
+        # we use classes to hold cache of Structs as precompiling is more
+        # efficient
+        strct = cast(struct.Struct, cls._struct)
+    except AttributeError:
```
glyph, line 258: this `try` block hits 2 attribute accesses. What I prefer to do in these cases is something like:

```
STRUCT = '__twisted_smb_struct__'
...
existingStruct = getattr(cls, STRUCT, None)
if existingStruct is not None:
    return existingStruct
newStruct = ...
setattr(cls, STRUCT, newStruct)
return newStruct
```

1. Less opportunity for weird edge cases since the error is very narrowly caught.
2. Also since the type checker has no visibility into the `_struct` attribute, using a constant to describe its name ensures that it's correctly the same in both places.
3. Using a constant also makes it less annoying to use a more properly namespaced name. Possibly not super important in this case, since it seems like 3rd party definition of SMB structures is probably not a goal, but it's still a good habit in case someone copies from this example.

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+        strct = struct.Struct(
+            "<" + "".join(i.metadata[SMB_METADATA] for i in smb_fields(cls))
+        )
+        cls._struct = strct
+    return strct
+
+
+def calcsize(cls: Any) -> int:
```
glyph, line 266:
```suggestion
def calcsize(cls: Type[object]) -> int:
```

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+
+def int32key(d: Dict[int, Any], val: Any) -> int:
+    """
+    generate a new random key for a dictionary
+    @param d: dictionary with 32-bit int keys
+    @type d: L{dict}
+    @param val: new dictionary value
+    @rtype: L{int}
+    """
+    assert len(d) < 0xC0000000  # otherwiae dict so big hard to find keys
```
glyph, line 287: In the rare case that this assertion is actually tripped, let's give the poor sysadmin an error message to look at ;)

```suggestion
    assert len(d) < 0xC0000000, "dictionary too large to find non-conflicting keys"
```

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    n = 0
+    while n == 0 or n in d:
+        [n] = _leint.unpack(secureRandom(_leint.size, True))
+    d[n] = val
+    return n
```
glyph, line 292: This seems like a peculiar and unnecessarily expensive way to get a random integer in a particular range.   Maybe something more like this?

```
from random import SystemRandom
UINT32_MAX = 0xff_ff_ff_ff
randint = SystemRandom().randint
MAX_ATTEMPTS = 32

```suggestion
def int32key(d: Dict[int, Any], val: Any) -> int:
    """
    generate a new random key for a dictionary
    @param d: dictionary with 32-bit int keys
    @type d: L{dict}
    @param val: new dictionary value
    @rtype: L{int}
    """
    for ignored in range(MAX_ATTEMPTS):
        n = randint(1, UINT32_MAX)
        if n not in d:
            return n
    assert False, "dictionary is too full to find a free key"
```

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+                transmission.
+    @type data: L{bytes}
+
+    @ivar hdr: the parsed header
+    @ivar body: the parsed body
+    """
+
+    data: bytes = attr.ib()
+    _proto: "SMBPacketReceiver" = attr.ib()
+    hdr: Any = attr.ib(default=None)
```
glyph, line 310: Could you express this as a union of something or a `Protocol` type? It seems like there should be *some* expectation of `hdr`'s type.

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+
+class SMBPacketReceiver(protocol.Protocol):
+    """
+    A L{SMBPacketReceiver} is a wire protocol parser for the SMB framing
+    mechanism, which consist of a 4-byte header: single null
+    and a 24-bit length field.
+    """
+
+    def __init__(
+        self, packetReceived: Callable[[SMBPacket], None], ctx: Dict[str, Any]
```
glyph, line 354: Love having `packetReceived` available via composition rather than needing to literally inherit :)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
+# -*- test-case-name: twisted.protocols._smb.tests -*-
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+"""Implement Microsoft's Server Message Block protocol"""
```
glyph, line 4: Codecov is extremely mad about this file, I think because `tests.py` is in the wrong place: it should be in a `test` directory somewhere (`twisted/protocols/_smb/test/` would be fine), and named `test_smb.py` rather than just `tests`.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+    server_uuid = uuid()
+    capabilities = medium()
+    max_transact = medium(MAX_TRANSACT_SIZE)
+    max_read = medium(MAX_READ_SIZE)
+    max_write = medium(MAX_WRITE_SIZE)
+    time = long()
+    boot_time = long()
+    offset = short(128, locked=True)
+    buflen = short()
+    reserved2 = medium()
```
glyph, line 72: Given that the respective functions here all return `Any`, they should have type annotations so that they get type checked.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+    ("negotiate", NegReq, NegResp),
+    ("session_setup", SessionReq, SessionResp),
+    ("logoff", BasicPacket, BasicPacket),
+    ("tree_connect", TreeReq, TreeResp),
+    ("tree_disconnect", BasicPacket, BasicPacket),
+]
+
+
+# the complete list of NT statuses is very large, so just
+# add those actually used
```
glyph, line 139: Can you include a link to the source of these so that if more need to be added there's a canonical place to go for them?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+    BAD_NETWORK_NAME = 0xC00000CC  # = "share not found"
+
+
+FLAG_SERVER = 0x01
+FLAG_ASYNC = 0x02
+FLAG_RELATED = 0x04
+FLAG_SIGNED = 0x08
+FLAG_PRIORITY_MASK = 0x70
+FLAG_DFS_OPERATION = 0x10000000
+FLAG_REPLAY_OPERATION = 0x20000000
```
glyph, line 169: It seems like this should be an `enum.Flag` and not just a bunch of ints

glyph: Can you remove the enums / flags to a separate module for ease of readability? This file is quite large as it is :)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+status          {status}
+command         {cmd!r} {cmdn:02x}
+credit request  {cr}
+flags           0x{flags:04x}{flags_desc}
+next command    0x{nc:x}
+message ID      0x{mid:x}
+session ID      0x{sid:x}
+async ID        0x{aid:x}
+tree ID         0x{tid:x}
+signature       {sig}""",
```
glyph, line 351: FYI, you can avoid this type of ugly outdent by using `textwrap.dedent` on an indented string.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+            sid=packet.hdr.session_id,
+            aid=packet.hdr.async_id,
+            tid=packet.hdr.tree_id,
+            sig=binascii.hexlify(packet.hdr.signature),
+        )
+        if packet.hdr.command < len(COMMANDS):
+            name, req_type, resp_type = COMMANDS[packet.hdr.command]
+            func = "smb_" + name
+            try:
+                if func in globals() and req_type:
```
glyph, line 372: Dispatch out of `globals()` is potentially quite slow (`globals()`, `locals()`, `currentframe()` and the like are debugging and metaprogramming features, not the sort of thing you should be calling in a tight loop like this) and error prone; it's hard to debug when something goes wrong with this.  Not to mention that the type checker has no visibility here.

Luckily, here you already control all the relevant namespaces, so you don't need this level of dynamic metaprogramming! Define `COMMANDS` at the bottom of the file, and just add the functions to it directly, then do `name, reqType, respType, handler = COMMANDS[packet.hdr.command]`.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+                    log.error(
+                        "command '{cmd}' not implemented",
+                        cmd=COMMANDS[packet.hdr.command][0],
+                    )
+                    errorResponse(packet, NTStatus.NOT_IMPLEMENTED)
+            except _base.SMBError as e:
+                log.error("SMB error: {e}", e=str(e))
+                errorResponse(packet, e.ntstatus)
+            except BaseException:
+                log.failure("in {cmd}", cmd=COMMANDS[packet.hdr.command][0])
```
glyph, line 388: You already pulled this value out above, no need to go doing the dispatch lookup again.

```suggestion
                log.failure("in {cmd}", cmd=name)
```

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+
+
+def sendHeader(
+    packet: _base.SMBPacket,
+    command: Optional[Union[int, str]] = None,
+    status: Union[NTStatus, int] = NTStatus.SUCCESS,
+) -> None:
+    """
+    prepare and transmit a SMB header and payload
+    so actually a full packet but focus of function on header construction
```
glyph, line 404: This line feels superfluous, more like an implementation note-to-self than any guidance on how to use the API?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+
+    @param command: command name or id, defaults to same as received packet
+    @type command: L{str} or L{int}
```
glyph, line 407: Once again, don't need these `@type`s any more.
```suggestion
```

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+        blob_manager.receiveResp(blob)
+        if blob_manager.credential:
+            log.debug("got credential: %r" % blob_manager.credential)
+            d = packet.ctx["factory"].portal.login(
+                blob_manager.credential,
+                SMBMind(
+                    packet.body.prev_session_id,
+                    blob_manager.credential.domain,
+                    packet.ctx["addr"],
+                ),
```
glyph, line 552: Thank you for taking the time to integrate this with cred!

src/twisted/protocols/_smb/_base.py
```diff
@@ -0,0 +1,396 @@
...
+    body: Any = attr.ib(default=None)
+
+    @property
+    def ctx(self) -> Dict[str, Any]:
+        """
+        the connection context: objects that need to persist scross
+        packets
+        @rtype: L{dict}
+        """
+        return self._proto.ctx
```
glyph, line 320: Clients of this class appear to have very strong expectations on the types of the values that get added to this dictionary. Should this be a user-supplied class, or collection of classes? `Dict[str, Any]` feels like way too broad of a type, particularly given that this seems like something that will be exposed to applications via `packetReceived`.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+) -> None:
+    """
+    send negotiate response
+
+    @type packet: L{_base.SMBPacket}
+
+    @param dialects: dialects offered by client, if C{None}, 2.02 used
+    @type dialects: L{list} of L{int}
+    """
+    log.debug("negotiateResponse")
```
glyph, line 491: messages that don't supply any parameters seem like a bit of a code smell to me, is there anything about this request that a debug log might include?

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+    blob_manager = packet.ctx["blob_manager"]
```
glyph, line 492: I really want mypy to have some visibility into *why* we expect `packet.ctx` to have a `blob_manager` key and why we expect it to be a `BlobManager`.

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+                    | FILE_WRITE_ATTRIBUTES
+                    | DELETE
+                    | READ_CONTROL
+                    | WRITE_DAC
+                    | WRITE_OWNER
+                    | SYNCHRONIZE
+                ),
+            )
+        if IPipe.providedBy(share):
+            assert resp is None, "share can only be one type"
```
glyph, line 663: `assert` seems to be the wrong way to handle this.  I'd prefer to see something like:

```
providedInterfaces = [each for each in IFilesystem, IPipe, IPrinter if each.providedBy(share)]
if len(providedInterfaces) != 1:
    raise ImplementerError(f"share type {type{share)} must implement exactly one of IFilesystem, IPipe, or IPrinter but it provides {providedInterfaces}")
responseBuilder = responseBuilders[providedInterfaces[0]]
resp = responseBuilder()
```

This would make it easier to see which thing corresponds to which type, keep this function shorter, and give the programmer a much better error message when they mess something up (as well as still working with `-OO` which may start mattering more in the world of more-optimized future Python)

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+
+def smb_tree_disconnect(packet: _base.SMBPacket, resp_type: type) -> None:
+    del packet.ctx["trees"][packet.hdr.tree_id]
+    packet.data = _base.pack(resp_type())
+    sendHeader(packet)
+
+
+class SMBFactory(protocol.Factory):
+    """
+    Factory for SMB servers
```
glyph, line 733: Given that this is the main entrypoint to the whole system, a slightly more expansive docstring about how to use it would be helpful

src/twisted/protocols/_smb/core.py
```diff
@@ -0,0 +1,760 @@
...
+    """
+
+    def __init__(self, portal: Portal, domain: str = "WORKGROUP") -> None:
+        """
+        @param portal: the configured portal
+        @type portal: L{twisted.cred.portal.Portal}
```
glyph, line 739: Maybe say something about the expectation of the portal's configured checkers & realm?

src/twisted/protocols/_smb/ismb.py
```diff
@@ -0,0 +1,61 @@
...
+Various interfaces for realms, avatars and related objects
+"""
+
+from typing import List, Union
+
+from zope.interface import Attribute, Interface
+
+
+class NoSuchShare(Exception):
+    pass
```
glyph, line 14: this should have a docstring

src/twisted/protocols/_smb/ismb.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class ISMBServer(Interface):
+    """
+    A SMB server avatar, contains a number of "shares" (filesystems/printers/
+    IPC pipes)
+    """
+
+    session_id = Attribute("the assigned int64 session ID")
```
glyph, line 23: We can clean up some of the coding-standard issues in the internals after the fact, but public interfaces really need to have names we can live with long-term :)

```suggestion
    sessionID = Attribute("the assigned int64 session ID")
```

src/twisted/protocols/_smb/ismb.py
```diff
@@ -0,0 +1,61 @@
...
+class IPrinter(Interface):
+    """
+    A share representing a printer
+    """
+
+
+class IPipe(Interface):
+    """
+    A share representing a interprocess communication (IPC) pipe
+    """
```
glyph, line 61: Empty interfaces are a bit of a code smell; if we have some responsibilities we expect to delegate to things, those should be expressed in the form of methods.  But I can see that there's a lot of code here already and I can imagine you have more implementation that might show up to build upon this. Is there a reason that these are empty for now, and do you plan to add things before this gets pushed as a public API?

src/twisted/protocols/_smb/security_blob.py
```diff
@@ -0,0 +1,329 @@
...
+from twisted.logger import Logger
+from twisted.protocols._smb import _base, ntlm
+
+log = Logger()
+
+
+class BlobManager:
+    """
+    encapsulates the authentication negotiation state
+    callers just send blobs in and out and get a credential
```
glyph, line 27: can you describe this briefly in terms of the order of the methods callers are expected to use? I don't know a ton about NTLM nd it's a little hard to imagine the expectations here.

src/twisted/protocols/_smb/security_blob.py
```diff
@@ -0,0 +1,329 @@
...
+    may be added over time
+    """
+
+    def __init__(self, domain: str) -> None:
+        """
+        @param domain: the server NetBIOS domain
+        @type domain: L{str}
+        """
+        self.domain = domain
+
```
glyph, line 40: This should be an attrs class, no need for a custom `__init__`

src/twisted/protocols/_smb/security_blob.py
```diff
@@ -0,0 +1,329 @@
...
+Blobs implemented as DER-encoded ASN.1 objects
+which in turn encapsulate NTLMv2 challenge-response authentication
+
+
+much of code adapted from Mike Teo's pysmb """
+
+from typing import Optional, cast
+
+from pyasn1.codec.der import decoder, encoder  # type: ignore
+from pyasn1.type import char, constraint, namedtype, namedval, tag, univ  # type: ignore
```
glyph, line 16: This should be depending on https://pypi.org/project/types-pyasn1/, not `type: ignore`ing the imports

glyph: Also, we *finally* eliminated our dependency on pyasn1 for everything else, so you'll need to add a new extra to `tool.hatch.metadata.hooks.custom.optional-dependencies` in `pyproject.toml`.

src/twisted/protocols/_smb/tests.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/python3
```
glyph, line 1: This module needs to be renamed as described in the comments above in order to actually run. Codecov's not complaining about this module but I think that's because it emitted so many other warnings that it exceeded its maximum for annotations on a single PR.

src/twisted/protocols/_smb/tests.py
```diff
@@ -0,0 +1,363 @@
...
+
+@attr.s
+class FakeStruct2:
+    i = _base.short()
+    b = _base.byte()
+    s = _base.octets(3)
+
+
+class TestBase(unittest.TestCase):
+    def test_base_pack(self) -> None:
```
glyph, line 69: Tests need docstrings so that maintainers understand what properties they're testing. They can (and should!) be very short, per https://jml.io/test-docstrings/ making a short assertion about the specific property they're testing.

src/twisted/protocols/_smb/tests.py
```diff
@@ -0,0 +1,363 @@
...
+        data = struct.pack("<HBf3sQ", 525, 42, 4.2, b"bob", 424242)
+        r = FakeStruct(one=525)  # type: ignore
+        r.two = 42
+        r.four = b"bob"
+        self.assertEqual(_base.pack(r), data)
+        with self.assertRaises(AssertionError):
+            r = FakeStruct(five=424243)  # type: ignore
```
glyph, line 76: If you need `type:ignore` they should be as narrow as possible. But why do you need it at all here? This is a very simple expression which should typecheck just fine

src/twisted/protocols/_smb/tests.py
```diff
@@ -0,0 +1,363 @@
...
+    cast(IReactorProcess, reactor).spawnProcess(pro, args[0], args, usePTY=usePTY)  # type: ignore
+    return pro.d
+
+
+TESTPORT = 5445
+SMBCLIENT = "/usr/bin/smbclient"
+
+
+@python_unittest.skipUnless(os.access(SMBCLIENT, os.X_OK), "smbclient unavailable")
+class SambaClientTests(unittest.TestCase):
```
glyph, line 325: Including the word "integration" here should help future maintainers know that this is going to be slow and potentially flaky, which is useful to understand in advance.  It's great to have this test, just want to make sure that it is appropriately categorized and most narrow unit-tests of behavior should be tested elsewhere as well.

```suggestion
class SambaClientIntegrationTests(unittest.TestCase):
```

.git-blame-ignore-revs
```diff
@@ -10,4 +11,5 @@
 # Note that these features require Git 2.23 (released August 2019).
 03cabb7f55d54e2b09eefd0e03df2d440f962bc4
 bc96c774be6e307e8e3e4d39780d37b045d0973a
+2249eebd81b195c72c9293a1a2224fa1134d4ca3
```
glyph, line 14: The idea here is that we didn't want to misattribute the authorship of lines when running the *initial* big syntax-cleanup pass; I don't think that you need to add this here; you actually authored all the lines you're reformatting.  (It's fine if you think it's helpful for blame legibility, this PR is big and quite old, I just really don't want to establish the precedent that everyone who runs `black` starts appending to this file.)

glyph (changes requested):
To get a more useful review on the branch as a whole, the tests really need to run, so that's the highest priority to fix.  New code like this should have 100% coverage.

Also a little bit of documentation of how to set up SMBClient manually as per the integration test so that I can poke at it and see what happens would be helpful.  (Is it possible to point a Mac or Windows machine at this yet?)

I see `_smb._base` is double-private, indicating that it's not going to be a public interface when this gets promoted to a public package, but if core, ismb, ntlm, and security_blob are all going to be public, the naming convention issues all need to be fixed, i.e. `thing_like_this` needs to be `thingLikeThis`, `smb_handler_name` needs to be `smb_handlerName`.

There should be some introductory `.rst` documentation as well, at some point, although that doesn't need to be in this PR.

Thanks again for keeping up with this contribution over the years.  2023 is going to be the year of SMB in Twisted, for sure :)

PR #1275, opened 2020-05-27
Replace base64.*string() functions to fix py3.9 support

mgorny:
Replace base64.decodestring() and .encodestring() functions as they
were deprecated since Python 3.1 in favor of (equivalent) .decodebytes()
and .encodebytes(), and were eventually removed in Python 3.9.

While at it, replace most of their uses with base64.b64encode()
and .b64decode() that are preferable to the former wrt ticket #6446,
and they do not introduce line breaks that the twisted code usually
discarded.

Use .decodebytes() and .encodebytes() in DirDBM as it seems to rely
on the exact presence of newlines, and changing that would break
backwards compatibility.

Fixes: ticket:6446
Fixes: ticket:9831

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9831
  * Note: the lint test failed for me but on existing lines unrelated to my change

### DIFF
--- a/src/twisted/conch/scripts/tkconch.py
+++ b/src/twisted/conch/scripts/tkconch.py
@@ -414,11 +414,12 @@ class SSHClientTransport(transport.SSHClientTransport):
             frame.write(
                 "Warning: Permanently added '%s' (%s) to the list of "
                 "known hosts.\r\n" %
-                (khHost, {b'ssh-dss':'DSA', b'ssh-rsa':'RSA'}[keyType]))
-            with open(os.path.expanduser('~/.ssh/known_hosts'), 'a') as known_hosts:
-                encodedKey = base64.encodestring(pubKey).replace(b'\n', b'')
+                (khHost, {b'ssh-dss': 'DSA', b'ssh-rsa': 'RSA'}[keyType]))
+            with open(os.path.expanduser(
+                      '~/.ssh/known_hosts'), 'a') as known_hosts:
+                encodedKey = base64.b64encode(pubKey)
                 known_hosts.write('\n%s %s %s' % (khHost, keyType, encodedKey))
-        except:
+        except BaseException:
             log.deferr()
             raise error.ConchError
 
--- a/src/twisted/conch/test/test_keys.py
+++ b/src/twisted/conch/test/test_keys.py
@@ -404,7 +404,7 @@ SUrCyZXsNh6VXwjs3gKQ
 
         self.assertRaises(
             keys.BadKeyError,
-            keys.Key.fromString, data=b'{' + base64.encodestring(sexp) + b'}',
+            keys.Key.fromString, data=b'{' + base64.b64encode(sexp) + b'}',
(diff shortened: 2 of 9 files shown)


mgorny:
I don't think these CI failures are related to my changes.

wiml (approved):
Also nice that this cleans up a bunch of `strip` calls that aren't needed with the newer api. Looks good to me!

wiml:
A question for @hawkowl or someone more familiar with towncrier than I am: is it a problem that the same bug number is mentioned in multiple newsfragments in subdirectories? Should they be moved up to the top level?

hawkowl:
(the CI failures are related to Azure Pipelines giving Unicode but the Python being configured to output cp1252 by default on Windows :/ -- been trying to debug it and come up with a solution)

Generally, unless there's a specific feature change (e.g. multiple external APIs in different sections), I collapse them into a single one in the main folder. I kind of want to get rid of the subpackages newsfragments, since it doesn't feel terribly useful now? Not sure.

glyph:
> (the CI failures are related to Azure Pipelines giving Unicode but the Python being configured to output cp1252 by default on Windows :/ -- been trying to debug it and come up with a solution)

What do you mean by "giving" unicode?  Giving unicode to what?  It seems like there's one failure:

```pytb
[ERROR]
  File "d:\a\1\s\build\alldeps-withcov-windows\lib\site-packages\twisted\test\test_process.py", line 2412, in gotEnvironment
    environ[self.goodKey.encode('ascii')],
builtins.KeyError: b'UNICODE'

twisted.test.test_process.Win32UnicodeEnvironmentTests.test_encodableUnicodeEnvironment
```

`environ` is always bytes-keys, right?

hawkowl:
> environ is always bytes-keys, right?

It fails because it raises an encoding error which gets printed to stdout because sys.stdout is cp1252 🙃

rodrigc:
I think that `Win32UnicodeEnvironmentTests.test_encodableUnicodeEnvironment`
is bogus and should be deleted.  By default `os.environ` is Unicode on all platforms in Python 3. 

There is not guarantee that all the environment variables in the environment will be ASCII.

Specifically what seems to be triggering this bug is that  **Michał Górny**'s name has non-ASCII characters in it.  Someone, when Azure Pipelines picks up this PR to build, something from Michał's name enters the environment, and triggers this one test failure on Windows.

I worked around this problem in: https://github.com/twisted/twisted/pull/1276
by amending the commit on Michał 's branch to have my name.

rodrigc:
> ```
> 
> `environ` is always bytes-keys, right?

No, `os.environ` is always Unicode in Python 3.

rodrigc (approved):

mgorny:
FTR, you can change my name to 'Michal Gorny'.

rodrigc:
Fixed in https://github.com/twisted/twisted/pull/1304

PR #1277, opened 2020-05-27
Fix verifyCryptedPassword() for crypt.crypt() throwing in py3.9

mgorny:
In Python 3.9, the crypt.crypt() function may throw an exception
if the underlying crypt() function fails.  Update
verifyCryptedPassword() to account for that, and preserve the existing
behavior of returning False in that case.

Fixes: ticket:9833

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9833

### DIFF
--- a/src/twisted/plugins/cred_unix.py
+++ b/src/twisted/plugins/cred_unix.py
@@ -43,7 +43,10 @@ def verifyCryptedPassword(crypted, pw):
         pw = pw.decode('utf-8')
     if not isinstance(crypted, StringType):
         crypted = crypted.decode('utf-8')
-    return crypt.crypt(pw, crypted) == crypted
+    try:
+        return crypt.crypt(pw, crypted) == crypted
+    except OSError:
+        return False
 
 
 
--- a/src/twisted/plugins/cred_unix.py
+++ b/src/twisted/plugins/cred_unix.py
@@ -16,7 +16,6 @@ from twisted.cred.checkers import ICredentialsChecker
 from twisted.cred.credentials import IUsernamePassword
 from twisted.cred.error import UnauthorizedLogin
 from twisted.internet import defer
-from twisted.python.compat import StringType
 
 
 
@@ -39,9 +38,9 @@ def verifyCryptedPassword(crypted, pw):
 
     if crypt is None:
         raise NotImplementedError("cred_unix not supported on this platform")
-    if not isinstance(pw, StringType):
+    if isinstance(pw, bytes):
         pw = pw.decode('utf-8')
-    if not isinstance(crypted, StringType):
+    if isinstance(crypted, bytes):
         crypted = crypted.decode('utf-8')
     try:
         return crypt.crypt(pw, crypted) == crypted
@@ -79,7 +78,7 @@ class UNIXChecker:
         @type username: L{unicode}/L{str} or L{bytes}
         """
         try:
-            if not isinstance(username, StringType):
+            if isinstance(username, bytes):
                 username = username.decode('utf-8')
             cryptedPass = pwd.getpwnam(username).pw_passwd
         except KeyError:
@@ -107,7 +106,7 @@ class UNIXChecker:
         @type username: L{unicode}/L{str} or L{bytes}
         """
         try:
-            if not isinstance(username, StringType):
+            if isinstance(username, bytes):
                 username = username.decode('utf-8')
             if getattr(spwd.struct_spwd, "sp_pwdp", None):
                 # Python 3
--- a/src/twisted/conch/checkers.py
+++ b/src/twisted/conch/checkers.py
@@ -16,8 +16,6 @@ try:
     import pwd
 except ImportError:
     pwd = None  # type: ignore[assignment]
-else:
-    import crypt
 
 try:
     import spwd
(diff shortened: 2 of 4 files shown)


src/twisted/plugins/cred_unix.py
```diff
@@ -43,7 +43,10 @@ def verifyCryptedPassword(crypted, pw):
         pw = pw.decode('utf-8')
     if not isinstance(crypted, StringType):
         crypted = crypted.decode('utf-8')
-    return crypt.crypt(pw, crypted) == crypted
+    try:
+        return crypt.crypt(pw, crypted) == crypted
+    except OSError:
+        return False
```
glyph, line 48: Maybe conch should… just be importing this function, given that we've implemented the exact same thing in both places here?

glyph (changes requested):
Thanks so much for contributing to Twisted!  This is a great bugfix.

Could you write a test that forces `crypt.crypt` to fail with a test mock?  It looks like this is a decrease to coverage because the tests that mock `crypt` assume it behaves like it used to.

src/twisted/plugins/cred_unix.py
mgorny: Will do.

mgorny:
> Could you write a test that forces `crypt.crypt` to fail with a test mock? It looks like this is a decrease to coverage because the tests that mock `crypt` assume it behaves like it used to.

I don't really understand what tests are you referring to. My grep skills fail to find anything mocking `crypt`. Are you saying that we should emulate the behavior for older Python versions?

src/twisted/plugins/cred_unix.py
rodrigc: @mgorny fixed this by adding `from twisted.plugins.cred_unix import verifyCryptedPassword` in `src/twisted/conch/checkers.py`

src/twisted/plugins/cred_unix.py
```diff
@@ -39,11 +38,14 @@ def verifyCryptedPassword(crypted, pw):
...
         raise NotImplementedError("cred_unix not supported on this platform")
-    if not isinstance(pw, StringType):
+    if isinstance(pw, bytes):
         pw = pw.decode('utf-8')
-    if not isinstance(crypted, StringType):
+    if isinstance(crypted, bytes):
         crypted = crypted.decode('utf-8')
-    return crypt.crypt(pw, crypted) == crypted
+    try:
+        return crypt.crypt(pw, crypted) == crypted
```
rodrigc, line 46: @mgorny In what case does `crypt.crypt()` throw an `OSError`?

In this PR: https://github.com/ansible/ansible/pull/70246

it mentions that `crypt.crypt` in Python 3.9 will throw an OSError if the algorithm is not supported.  Is there a test that we can write to trigger this error, so that there is test coverage of the `except OSError:` branch?

See:
https://coveralls.io/builds/32570311/source?filename=src%2Ftwisted%2Fplugins%2Fcred_unix.py#L46

rodrigc: Looks like `crypt.crypt()` throws an OSError if the underlying `crypt()` C function fails due to this change in Python 3.9:

https://github.com/python/cpython/commit/0d3fe8ae4961bf551e7d5e42559e2ede1a08fd7c

It might not be so easy to write a test for this using the existing Python `crypt.crypt` function.
Using a mock of crypt which always throws an `OSError` as @glyph suggested might be the easier way to go.

mgorny: That's weird. The current set of tests was sufficient to get it to throw for me. IIRC it was throwing on one of the invalid password cases.

rodrigc: Ah, maybe it's because `crypt.crypt` only throws `OSError` on Python 3.9, but none of the CI for Twisted runs on Python 3.9.
In that case it would be good to write a test which mocks out `crypt.crypt` and always throws OSError.

mgorny:
```python
In [1]: import crypt                                                                                                                               

In [2]: crypt.crypt('test', '')                                                                                                                    
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-7185f9de0941> in <module>
----> 1 crypt.crypt('test', '')

/usr/lib/python3.9/crypt.py in crypt(word, salt)
     80     if salt is None or isinstance(salt, _Method):
     81         salt = mksalt(salt)
---> 82     return _crypt.crypt(word, salt)
     83 
     84 

OSError: [Errno 22] Invalid argument
```

rodrigc: Yes, but that only fails like that on Python 3.9, and Twisted's CI does not yet run on Python 3.9

mgorny: I'm sorry, I've written my comment before reading yours ;-). I'm already working on a mocked test case, just need a few minutes to run the suite.

rodrigc: No worries. :)
It looks like Travis also has a **3.9-dev** target, see: https://github.com/jazzband/pip-tools/pull/1126/files
.

If you want, we can also add **3.9-dev** to .travis.yml .
That will be another way to get the coverage numbers up.

mgorny: I've pushed a test that should cover the coverage problem. Ofc, enabling tests on 3.9 would also be nice.

rodrigc: Thanks, this looks ok now, and coverage is better: https://coveralls.io/builds/32572286

rodrigc (approved):
looks ok now

PR #1278, opened 2020-05-27
Use xml.etree.ElementTree instead of deprecated cElementTree

mgorny:
The xml.etree.cElementTree is deprecated, and has been removed in Python
3.9.  At the same time, xml.etree.ElementTree has already been using
cElementTree implicitly since Python 3.3.  Update test_flatten to use
the latter to provide compatibility with newer Python versions.

Fixes: ticket:9834

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9834

### DIFF
--- a/src/twisted/test/test_process.py
+++ b/src/twisted/test/test_process.py
@@ -2390,18 +2390,18 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
     def test_encodableUnicodeEnvironment(self):
         """
         Test C{os.environ} (inherited by every subprocess on Windows) that
-        contains an ascii-encodable Unicode string. This is different from
-        passing Unicode environment explicitly to spawnProcess (which is not
-        supported on Python 2).
+        contains an ascii-encodable Unicode string.
         """
         os.environ[self.goodKey] = self.goodValue
         self.addCleanup(operator.delitem, os.environ, self.goodKey)
 
-        p = GetEnvironmentDictionary.run(reactor, [], properEnv)
+        p = GetEnvironmentDictionary.run(reactor, [], os.environ)
+
         def gotEnvironment(environ):
             self.assertEqual(
                 environ[self.goodKey.encode('ascii')],
                 self.goodValue.encode('ascii'))
+
         return p.getResult().addCallback(gotEnvironment)
 
 
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -10,7 +10,7 @@ import sys
 import traceback
 from unittest import skipIf
 
-from xml.etree.cElementTree import XML
+from xml.etree.ElementTree import XML
 
 from collections import OrderedDict
 


rodrigc:
@mgorny As you are working on your PR's for Twisted, can you run:

```
tox -e lint
```

in the top-level directory of your tree before pushing to GitHub?
This will identify some small style/whitespace errors in your PR which show up as errors
in the CI for Twisted.

rodrigc:
@mgorny can you run `tox -e lint` in the top-level of your tree before pushing to GitHub, and take a pass at fixing any whitespace/style errors reported by the linter?

In your other PR, I fixed a few simple ones here: https://github.com/twisted/twisted/pull/1275/commits/686f20fd070b523a425c7a05e62c7577687ca050

src/twisted/internet/base.py
```diff
@@ -971,7 +971,7 @@ def argChecker(arg):
             """
             if isinstance(arg, unicode):
                 try:
-                    arg = arg.encode(defaultEncoding)
```
rodrigc: I guess you put this in here for debugging some of the CI failures on Windows?

rodrigc (changes requested):

src/twisted/web/test/test_flatten.py
```diff
@@ -9,7 +9,7 @@
 import sys
 import traceback
 
-from xml.etree.cElementTree import XML
```
rodrigc, line 13: This part looks OK.

mgorny:
> @mgorny As you are working on your PR's for Twisted, can you run:
> 
> ```
> tox -e lint
> ```
> 
> in the top-level directory of your tree before pushing to GitHub?
> This will identify some small style/whitespace errors in your PR which show up as errors
> in the CI for Twisted.

I did but I don't think it found anything in my changes. I wasn't sure if I'm supposed to make irrelevant style fixes in existing code.

mgorny:
Actually, I don't think there were any reports for files I've modified.

rodrigc:
@mgorny The linter failed on your branch here: https://travis-ci.com/github/twisted/twisted/jobs/341031233

You can confirm by running `tox -e lint`

mgorny:
> @mgorny The linter failed on your branch here: https://travis-ci.com/github/twisted/twisted/jobs/341031233
> 
> You can confirm by running `tox -e lint`

My commit does not touch this file. @hawkowl 's does.

hawkowl:
Yeah, sorry. That was me trying to debug Azure Pipelines breaking on your name so then it goes green :/ feel free to force push it back up to remove the commits.

mgorny:
> Yeah, sorry. That was me trying to debug Azure Pipelines breaking on your name so then it goes green :/ feel free to force push it back up to remove the commits.

No luck, then?

hawkowl:
I know that Azure Pipelines is passing in Unicode fine, it's just that older Python 3s (before the UTF-8 on Windows PEP) choke on outputting it because the default encoding goes to cp1252 :(

So, that means that we might just have to run the Twisted tests on Windows with -X utf8 or so, but then we're running the tests with a "non-standard" version/configuration of Python. But, since the default behaviour breaks for dumb reasons (accepting Unicode environment variables that it can't then output), perhaps running without -X utf8 or equiv on Windows should be deprecated/warned about as defective by design and unsupported (especially since Windows isn't DOS anymore, and all Linux/MacOS tests rely on proper Unicode support).

I'll raise this on the ML, since this may have other solutions (such as manually configuring the output in Python space, perhaps) that I haven't thought of.

As far as this/other PRs: I think that due to the issue being that Windows is extremely dumb, we can safely ignore the tests that fail on Windows on all your PRs as unrelated. I hope to get to reviewing these soon, just been caught up with starting a new job. 

Sorry for the hassle, and thanks again for your contributions. 💜

src/twisted/test/test_process.py
```diff
@@ -483,7 +483,7 @@ class GetEnvironmentDictionary(UtilityProcessProtocol):
     and expose it to interested parties.
     """
     programName = b"twisted.test.process_getenv"
-
+
```
glyph: This trailing whitespace is probably not necessary :).

azure-pipelines/run_test_steps.yml
```diff
@@ -53,6 +53,8 @@ steps:
     continueOnError: ${{ parameters.allowFailure }}
     env:
       TWISTED_REACTOR: ${{ parameters.windowsReactor }}
+      PYTHONUTF8: "1"
+      PYTHONIOENCODING: "utf8:surrogateescape"
```
glyph: It looks like this was an attempt to fix the failing windows tests… which are still failing.

In general I am concerned about adding env vars like this to the test suite, since we might inadvertently introduce features into Twisted where it *can't* work without custom configuration on Windows, so we should remove this.

src/twisted/web/test/test_flatten.py
glyph: This does look fine, and I wonder why it would provoke these Windows failures.

glyph (changes requested):
Sadly we're still failing CI on this, and I don't think the tweak to fix this is a good idea.  But also, are we just seeing these CI failures on trunk too?

If you could please back out the unrelated stuff, we'll separately try to address the build failure on trunk.

mgorny:
Ok, reverted to my original change.

rodrigc (approved):
The original change for using xml.etree.ElementTree is OK.

The CI errors have nothing to do with this change.

rodrigc:
@mgorny it looks like the CI errors have to do with the fact that your GitHub ID has non-ASCII characters in its name, because I cannot reproduce the problem on Azure when I submit the same code change from my branch under my GitHub ID.

rodrigc:
@mgorny This problem is actually kind of interesting. :)

I will get your Python 3.9 changes into the tree today, but if you are cool with it, I'd like to collaborate with you a bit and try to fix this Unicode thing on the Twisted side.
We should not block people from submitting fixes to Twisted if they have non-ASCII characters in their name!!  This is 2020!

mgorny:
Sure. I presume you'd need me to submit some boilerplate pull request to get my ID on it, correct?

rodrigc:
Fixed in

rodrigc:
@mgorny OK, no rush on this, but whenever you have time, can you do the following:

1.  Take my branch `9858-rodrigc-mgorny-unicode`, make a new branch from that.
2.  Make a small but trivial change to some .py file in the Twisted tree.  Try to do the minimal change, so as to not break any of the linters (we don't want to have linter failures adding to the confusion).
3.  Make a new PR with your new branch using your mgorny GitHub ID.

I bet you that the CI will fail on Windows for that....

@hawkowl did a really good write-up about the problem of UTF-8 vs. CP1252 here: https://twistedmatrix.com/pipermail/twisted-python/2020-June/065140.html

It took me a while to connect the dots that your GitHub ID was triggering these failures!

PR #1280, opened 2020-05-29
Interface methods should not have 'self' argument

rodrigc:
https://twistedmatrix.com/trac/ticket/9837

This fixes several warnings that I found by running:

```
tox -e mypy
```

such as:

```
src/twisted/protocols/haproxy/_interfaces.py:36:5: error: Interface methods should not have 'self' argument  [misc]
        def feed(self, data):
        ^
src/twisted/protocols/haproxy/_interfaces.py:52:5: error: Interface methods should not have 'self' argument  [misc]
        def parse(self, line):
        ^
```


adiroiban (approved):
All good. thanks! This can be merged.

PR #1284, opened 2020-05-31
[#9841] Clean up type checking/hinting for twisted.web.http_headers

hawkowl:
https://twistedmatrix.com/trac/ticket/9841

### DIFF
--- a/src/twisted/web/http_headers.py
+++ b/src/twisted/web/http_headers.py
@@ -6,11 +6,13 @@
 An API for storing HTTP header names and values.
 """
 
+from typing import List, Iterator, Tuple, Optional, Dict, Union, Any, cast
 
 from twisted.python.compat import comparable, cmp, unicode
+from twisted.python._typing import String
 
 
-def _dashCapitalize(name):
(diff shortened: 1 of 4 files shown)


src/twisted/web/http_headers.py
```diff
@@ -208,20 +209,35 @@ def setRawHeaders(self, name, values):
         @param values: A list of strings each one being a header value of
             the given name.
 
+        @raise TypeError: Raised if C{values} is not a L{list} of L{bytes}
+            or L{unicode} strings, or if C{name} is not a L{bytes} or
+            L{unicode} string.
```
twm, line 214: Should we still be documenting things as `unicode` in this brave new Python 3 world? I was tempted to change these in my last PR.

wsanchez: We should switch to `str`.  There is no ambiguity left there.

rodrigc:
@wsanchez @twm I continued this PR at https://github.com/twisted/twisted/pull/1389

rodrigc:
Fixed in https://github.com/twisted/twisted/pull/1389

PR #1285, opened 2020-06-01
Remove some _PY3 checks and eliminate Python2 only code

rodrigc:
https://twistedmatrix.com/trac/ticket/9842

This eliminates more errors from mypy

### DIFF
--- a/src/twisted/conch/client/default.py
+++ b/src/twisted/conch/client/default.py
@@ -14,7 +14,7 @@ interact with a known_hosts database, use L{twisted.conch.client.knownhosts}.
 
 from twisted.python import log
 from twisted.python.compat import (
-    nativeString, raw_input, _PY3, _b64decodebytes as decodebytes)
+    nativeString, raw_input, _b64decodebytes as decodebytes)
 from twisted.python.filepath import FilePath
 
 from twisted.conch.error import ConchError
@@ -25,10 +25,11 @@ from twisted.conch.client.knownhosts import KnownHostsFile, ConsoleUI
 
 from twisted.conch.client import agent
 
-import os, sys, getpass, contextlib
-
-if _PY3:
-    import io
+import contextlib
+import getpass
+import io
+import os
+import sys
 
 # The default location of the known hosts file (probably should be parsed out
 # of an ssh config file someday).
@@ -321,13 +322,9 @@ class SSHUserAuthClient(userauth.SSHUserAuthClient):
         @return: File objects for reading and writing to /dev/tty,
                  corresponding to standard input and standard output.
         @rtype: A L{tuple} of L{io.TextIOWrapper} on Python 3.
-                A L{tuple} of binary files on Python 2.
         """
-        stdin = open("/dev/tty", "rb")
-        stdout = open("/dev/tty", "wb")
-        if _PY3:
-            stdin = io.TextIOWrapper(stdin)
-            stdout = io.TextIOWrapper(stdout)
+        stdin = io.TextIOWrapper(open("/dev/tty", "rb"))
+        stdout = io.TextIOWrapper(open("/dev/tty", "wb"))
         return stdin, stdout
 
 
--- a/src/twisted/conch/ls.py
+++ b/src/twisted/conch/ls.py
@@ -6,7 +6,6 @@ import array
 import stat
 
 from time import time, strftime, localtime
-from twisted.python.compat import _PY3
 
 # Locale-independent month names to use instead of strftime's
 _MONTH_NAMES = dict(list(zip(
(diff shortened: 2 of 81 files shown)


src/twisted/conch/client/default.py
```diff
@@ -325,9 +326,8 @@ def _openTty(cls):
         """
         stdin = open("/dev/tty", "rb")
         stdout = open("/dev/tty", "wb")
-        if _PY3:
-            stdin = io.TextIOWrapper(stdin)
-            stdout = io.TextIOWrapper(stdout)
+        stdin = io.TextIOWrapper(stdin)
```
adiroiban:
```suggestion
        stdin = io.TextIOWrapper(open("/dev/tty", "rb"))
```

rodrigc: @adiroiban OK, I made the change

src/twisted/conch/ls.py
```diff
@@ -22,41 +21,59 @@ def lsLine(name, s):
...
+        perms[0] = ord('l')
+    elif stat.S_ISSOCK(ft):
+        perms[0] = ord('s')
+    else:
+        perms[0] = ord('!')
     # User
-    if mode&stat.S_IRUSR:perms[1] = ord('r')
-    if mode&stat.S_IWUSR:perms[2] = ord('w')
-    if mode&stat.S_IXUSR:perms[3] = ord('x')
+    if mode & stat.S_IRUSR:
```
rodrigc, line 41: I had to make some changes like this to appease twistedchecker and pycodestyle

src/twisted/conch/scripts/cftp.py
```diff
@@ -769,31 +769,31 @@ def _ebNotADirectory(self, reason, path, glob):
...
         else:
             reason = files
             reason.trap(EOFError)
             directory.close()
-            return l
+            return matchedFiles
 
     def _abbrevSize(self, size):
-        # from http://mail.python.org/pipermail/python-list/1999-December/018395.html
+        # from http://mail.python.org/pipermail/python-list/1999-December/018395.html  # noqa
```
adiroiban, line 790: Why `noqa`? This looks like normal code which is called during the process output.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -283,10 +285,9 @@ def displayPublicKey(options):
         if not options.get('pass'):
             options['pass'] = getpass.getpass('Enter passphrase: ')
         key = keys.Key.fromFile(
-            options['filename'], passphrase = options['pass'])
+            options['filename'], passphrase=options['pass'])
     displayKey = key.public().toString('openssh')
-    if _PY3:
-        displayKey = displayKey.decode("ascii")
+    displayKey = displayKey.decode("ascii")
```
adiroiban: Minor comment 
```suggestion
    displayKey = key.public().toString('openssh').decode("ascii")
```

src/twisted/conch/test/test_cftp.py
```diff
@@ -469,10 +470,7 @@ def test_printProgressBarReporting(self):
 
         self.client._printProgressBar(wrapper, startTime)
 
-        if _PY3:
-            result = b"\rb'sample' 40% 4.0kB 2.0kBps 00:03 "
-        else:
-            result = "\rsample 40% 4.0kB 2.0kBps 00:03 "
+        result = b"\rb'sample' 40% 4.0kB 2.0kBps 00:03 "
```
adiroiban, line 473: Why show the bytes representation in the progress bar?

I guess that the progress bar is for human consumption so here it should show the human readable name in UTF8 encoding.

So this might be a bug in cftp.
I think that is ok to leave the tests as it is, but the test should have a comment (and maybe a link to a ticket) explaining that this is not quite right.

So line 846 from scripts/cftp.py should be `front = f.name.decode('utf-8)`
https://github.com/twisted/twisted/pull/1285/files#diff-937ab7dc02c3bea8d0a3359822655bdeR846

src/twisted/internet/process.py
```diff
@@ -428,25 +428,23 @@ def _fork(self, path, uid, gid, executable, args, environment, **kwargs):
...
+                        # getfilesystemencoding() or sys.stderr.encoding,
+                        # because we want an encoding that will be able to
+                        # encode the full range of code points.  We are
+                        # (most likely) talking to the parent process on
+                        # the other end of this pipe and not the filesystem
+                        # or the original sys.stderr, so there's no point
+                        # in trying to match the encoding of one of those
+                        # objects.
+
+                        stderr = io.TextIOWrapper(stderr, encoding="utf-8")
```
adiroiban: Maybe is better to move the stderr definition together
```suggestion
                        stderr = os.fdopen(2, 'wb')
                        stderr = io.TextIOWrapper(stderr, encoding="utf-8")
```

or

```suggestion
                        stderr = io.TextIOWrapper(os.fdopen(2, 'wb'), encoding="utf-8")
```

src/twisted/logger/test/test_stdlib.py
```diff
@@ -275,8 +275,7 @@ def handlerAndBytesIO():
     output = BytesIO()
     stream = output
```
adiroiban, line 276: The above line can be removed

src/twisted/python/reflect.py
```diff
@@ -627,6 +627,5 @@ def objgrep(start, goal, eq=isLike, path='', paths=None, seen=None,
     'fullyQualifiedName']
 
 
-if _PY3:
-    # This is to be removed when fixing #6986
-    __all__.remove('objgrep')
+# This is to be removed when fixing #6986
+__all__.remove('objgrep')
```
adiroiban, line 631: I don't understand why not delete all the code depending on objgrep. For example `findInstances` .

Once  Trac#6986 is fixed the code can be added bac,

src/twisted/spread/jelly.py
```diff
@@ -489,25 +486,21 @@ def jelly(self, obj):
...
+            elif isinstance(obj, types.MethodType):
+                aSelf = obj.__self__
+                aFunc = obj.__func__
+                aClass = aSelf.__class__
                 return [b"method", aFunc.__name__, self.jelly(aSelf),
                         self.jelly(aClass)]
             elif objType is unicode:
                 return [b'unicode', obj.encode('UTF-8')]
-            elif objType is type(None):
+            elif isinstance(obj, type(None)):
```
adiroiban, line 499: This looks like a complicated way of expressing
```suggestion
            elif obj is None:
```

src/twisted/test/test_plugin.py
```diff
@@ -527,6 +528,39 @@ def test_developmentPluginAvailability(self):
             self.assertEqual(names, ['app', 'dev'])
 
 
+    def test_freshPyReplacesStalePyc(self):
```
adiroiban, line 531: is this change in the scope of this PR?

src/twisted/web/test/test_http_headers.py
```diff
@@ -611,14 +610,14 @@ def test_subclassRepr(self):
...
+        foo = "foo\u00E1"
+        bar = "bar\u2603"
+        baz = "baz"
         fooEncoded = "'foo\\xe1'"
         barEncoded = "'bar\\xe2\\x98\\x83'"
-        if _PY3:
-            fooEncoded = "b" + fooEncoded
-            barEncoded = "b" + barEncoded
+        fooEncoded = "b" + fooEncoded
+        barEncoded = "b" + barEncoded
```
adiroiban: Remove the 2 lines above and define them directly as
```suggestion
        fooEncoded = "b'foo\\xe1'"
        barEncoded = "b'bar\\xe2\\x98\\x83'"
```

adiroiban (approved):
Thanks for the changes.

Assuming that all the tests are green I think that this can be merged.

I only have minor comments

I don't know what to day about the docstring `L{unicode}` to `L{str}`.
I think that it would still help to have explicit unicode, to differentiate from a possible case in which the the old `L{str}` from python2 could mean bytes. 

If I see unicode in docs, then I know is unicode, if I see {str} I run more tests and read to code to make sure if this is bytes or unicode.

adiroiban (changes requested):
This branch was forced pushed while I was reviewing it ... so not sure what to say.
I guess that my previous review is no longer valid.

I would prefer to merge with squash and then be able to push without force in the branch.

src/twisted/conch/scripts/cftp.py
rodrigc: `noqa` is needed to quieten pyflakes and pycodestyle if a line is too long.  That URL is long, so I couldn't shorten it.

src/twisted/conch/scripts/ckeygen.py
rodrigc: OK, fixed.

src/twisted/conch/test/test_cftp.py
rodrigc: This is OK.  The context of this is:

```
        result = b"\rb'sample' 40% 4.0kB 2.0kBps 00:03 "
        self.assertEqual(self.client.transport.value(), result)
```

`self.client.transport` is of type `twisted.conch.test.test_cftp.InMemorySSHChannel` which inherits from `twisted.internet.testing.StringTransport`.  The `value()` method of this returns bytes.  It is unfortunate that the name of this class is `StringTransport`, but this class was created a long time ago, in the Python 2 days.

src/twisted/internet/process.py
rodrigc: OK, fixed.

src/twisted/python/reflect.py
rodrigc: This can be done in a separate PR as part of #6986

src/twisted/web/test/test_http_headers.py
rodrigc: OK, fixed.

PR #1286, opened 2020-06-02
New decorator t.p.deprecate.deprecatedKeywordParameter().

wiml:
* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9844

### DIFF
--- a/src/twisted/python/test/test_deprecate.py
+++ b/src/twisted/python/test/test_deprecate.py
@@ -738,7 +738,6 @@ def functionWithDeprecatedParameter(a, b, c=1, foo=2, bar=3):
     """
     Function with a deprecated keyword parameter.
     """
-    pass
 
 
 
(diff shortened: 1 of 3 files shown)


src/twisted/python/test/test_deprecate.py
```diff
@@ -732,6 +733,15 @@ def someProperty(self, value):
 
 
 
+@deprecatedKeywordParameter(Version('Twisted', 19, 2, 0), 'foo')
+def functionWithDeprecatedParameter(a, b, c=1, foo=2, bar=3):
+    """
+    Function with a deprecated keyword parameter.
+    """
+    pass
```
rodrigc: Not a big deal, but if you have a valid doc string, the function is still valid
if you omit the `pass`.

rodrigc (approved):
Other than one small comment, I think this looks ok.

rodrigc:
@wiml I have added a mypy check to Twisted, which can be run in the top-level Twisted directory with:

```
tox -e mypy
```

I see one mypy error in the code introduced in this commit:

```
src/twisted/python/deprecate.py:856:16: error: Incompatible return value type (got "Callable[..., Any]", expected "_Tc")  [return-value]
```

Could you take a look at that?  That is a more advanced usage of type annotations.   I am not savvy enough to know with why mypy is complaining about this usage. :)

PR #1287, opened 2020-06-02
xmlrpc: verify HTTPS certificates by default

ktdreyer:
Use `ssl.optionsForClientTLS()` when opening the HTTPS connection with `connectSSL()`. With this change, Twisted will use `ssl.platformTrust()` to verify the SSL connection by default.

Add a new `trustRoot` argument to the `Proxy` class constructor. This allows users to control how to trust the HTTPS certificate.

Raise `ssl.SSL.Error` in `QueryProtocol` instead of silently skipping it.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9836

### DIFF
--- a/src/twisted/web/newsfragments/9836.bugfix
+++ b/src/twisted/web/newsfragments/9836.bugfix
@@ -1,0 +1,1 @@
+xmlrpc's Proxy class now verifies HTTPS certificates by default.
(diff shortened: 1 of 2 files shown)


ktdreyer:
After digging further into `twisted.web.xmlrpc`, I think we should rewrite the module to use a `twisted.web.client.Agent` instead. The current Twisted xmlrpc client still uses HTTP/1.0. The `Agent` class would give us proper HTTPS support, and the ability to use `HTTPConnectionPool`, etc.

PR #1289, opened 2020-06-09
Eliminate various mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9847

### DIFF
--- a/src/twisted/__main__.py
+++ b/src/twisted/__main__.py
@@ -11,5 +11,5 @@ from pkg_resources import load_entry_point
 
 if __name__ == '__main__':
     sys.exit(
-        load_entry_point('Twisted', 'console_scripts', 'twist')()
+        load_entry_point('Twisted', 'console_scripts', 'twist')()  # type: ignore[func-returns-value]  # noqa
     )
--- a/src/twisted/conch/test/test_conch.py
+++ b/src/twisted/conch/test/test_conch.py
@@ -32,28 +32,27 @@ try:
 except ImportError:
     pass
 
-try:
-    import pyasn1
-except ImportError:
-    pyasn1 = None
-
+pyasn1 = requireModule('pyasn1')
 cryptography = requireModule("cryptography")
+
 if cryptography:
     from twisted.conch.avatar import ConchUser
     from twisted.conch.ssh.session import ISession, SSHSession, wrapProtocol
 else:
     from twisted.conch.interfaces import ISession
 
-    class ConchUser:
+    class ConchUser:  # type: ignore[no-redef]
         pass
 try:
     from twisted.conch.scripts.conch import (
-        SSHSession as StdioInteractingSession
+        SSHSession as _StdioInteractingSession
     )
 except ImportError as e:
     StdioInteractingSession = None
     _reason = str(e)
     del e
+else:
+    StdioInteractingSession = _StdioInteractingSession
 
 
 
--- a/src/twisted/internet/test/test_unix.py
+++ b/src/twisted/internet/test/test_unix.py
@@ -17,9 +17,11 @@ from typing import Optional, Sequence, Type
 from unittest import skipIf
 
 try:
-    from socket import AF_UNIX
+    from socket import AF_UNIX as _AF_UNIX
 except ImportError:
     AF_UNIX = None
+else:
+    AF_UNIX = _AF_UNIX
 
 from zope.interface import Interface, implementer
 
--- a/src/twisted/trial/test/detests.py
+++ b/src/twisted/trial/test/detests.py
@@ -150,7 +150,8 @@ class DeferredTests(unittest.TestCase):
         d = defer.succeed('todo')
         d.addCallback(self._cb_error)
         return d
-    test_expectedFailure.todo = "Expected failure"
+    test_expectedFailure.todo = "Expected failure"  # type: ignore[attr-defined]  # noqa
+
 
 
 class TimeoutTests(unittest.TestCase):
@@ -160,7 +161,7 @@ class TimeoutTests(unittest.TestCase):
         d = defer.Deferred()
         reactor.callLater(0, d.callback, 'hoorj!')
         return d
-    test_pass.timeout = 2
+    test_pass.timeout = 2  # type: ignore[attr-defined]
 
     def test_passDefault(self):
         # test default timeout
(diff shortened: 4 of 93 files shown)


src/twisted/__main__.py
```diff
@@ -11,5 +11,5 @@
 
 if __name__ == '__main__':
     sys.exit(
-        load_entry_point('Twisted', 'console_scripts', 'twist')()
+        load_entry_point('Twisted', 'console_scripts', 'twist')()  # type: ignore[func-returns-value]  # noqa
```
adiroiban, line 14: load_entry_point() is a common API, why it needs to have the type ignored?

adiroiban (changes requested):
It looks like there are 2 "elimination" methods here:

1. Fix the code
2. Add a mypy ignore comment.

I would prefer to have separate PRs for each method.
I would be able to review the changes not related to the ignore comments, like for example `    cryptSkip = ''`

adiroiban:
@rodrigc thanks for working on this. I can imagine that it is frustrating to not have a PR reviewed.... but is hard for me to do any quality review.

I also that this PR has some `@skipIf` take-2 changes. Is better to have them in a separate PR and I can review them right away

src/twisted/__main__.py
rodrigc: Without this change, mypy reports this error:

```
src/twisted/__main__.py:14:9: error: "load_entry_point" does not return a value  [func-returns-value]
            load_entry_point('Twisted', 'console_scripts', 'twist')()
            ^
src/twisted/trial/__main__.py:9:9: error: "load_entry_point" does not return a value  [func-returns-value]
            load_entry_point('Twisted', 'console_scripts', 'trial')()
```

This comes from setuptools: https://github.com/pypa/setuptools/blob/11051abd79b04782d759720a4d9dac2d2bbbf49c/pkg_resources/__init__.py#L488

I thought it was safer to add the comment than to change the code.

rodrigc:
No it is not frustrating, don't worry :).  I think it is OK, to break this up into smaller PR's.  It was not clear how to break things up, until actually trying to squelch a bunch of errors.
It's useful to get review feedback even on this big PR just to get things rolling.

src/twisted/__main__.py
wsanchez: It looks like `load_entry_point` returns some object or `None` and this is relying on the funky handling by `sys.exit` for non-`int` values…

The comment is OK… if `load_entry_point` eventually gets a type hint, we'll get an error from `mypy` that the comment is unnecessary.

Why do we also need the `noqa` comment?

src/twisted/conch/test/test_conch.py
```diff
@@ -32,28 +32,27 @@
...
 else:
     from twisted.conch.interfaces import ISession
 
-    class ConchUser:
+    class ConchUser:  # type: ignore[no-redef]
         pass
 try:
     from twisted.conch.scripts.conch import (
-        SSHSession as StdioInteractingSession
+        SSHSession as _StdioInteractingSession
```
wsanchez, line 48: Can't use `no-redef`? This is less desirable than a comment, since a fix to `mypy` to handle these import cases. won't notice that this change isn't needed any more.

src/twisted/trial/test/detests.py
```diff
@@ -150,7 +150,8 @@ def test_expectedFailure(self):
         d = defer.succeed('todo')
         d.addCallback(self._cb_error)
         return d
-    test_expectedFailure.todo = "Expected failure"
+    test_expectedFailure.todo = "Expected failure"  # type: ignore[attr-defined]  # noqa
```
wsanchez, line 153: I wonder if some `@todo`, `@skip`, etc. decorators would make this ness stupid looking.

I've been doing the same, but there are a *lot* of these in Twisted.

src/twisted/__main__.py
wsanchez: OK `noqa` is line length… I wish `mypy` had a way to comment on the next line or something.

src/twisted/internet/test/test_unix.py
```diff
@@ -17,9 +17,11 @@
 from unittest import skipIf
 
 try:
-    from socket import AF_UNIX
+    from socket import AF_UNIX as _AF_UNIX
 except ImportError:
     AF_UNIX = None
+else:
+    AF_UNIX = _AF_UNIX
```
wsanchez, line 24: I would suggest ` # type: ignore[assignment]` on the line where we assign `None` instead.

src/twisted/conch/test/test_conch.py
wsanchez: OK I think this is here for the line where assign `None`.

I would suggest ` # type: ignore[assignment]` on the line where we assign `None` instead.

wsanchez (approved):
I made a couple of comments but nothing that is blocking.  This is good progress.

src/twisted/trial/test/detests.py
rodrigc: Well in other PR's I moved all the tests to use the @skipIf decorator from the Python unittest core library.
Implementing a @todo decorator in Twisted may not be a bad idea, and would be a good way to clean these up without adding comments specific to mypy

src/twisted/__main__.py
rodrigc: Yes, `noqa` is for line length.  I couldn't figure out a better way to do this when adding the mypy comment would cause the line length to exceed the length tolerated by pycodestyle.

mthuurne: An alternative could be to teach pycodestyle to (optionally) strip comments for tooling when computing line length. Perhaps look for a pattern of the comment starting with a single word followed by a colon, or just ignore all comments that follow code.

src/twisted/internet/test/test_unix.py
rodrigc: @wsanchez I implemented this pattern:

```
try:
      from blah import foo as _foo
except ImportError
     foo = None
else:
     foo = _foo
```

Based on the discussion in this mypy issue:
https://github.com/python/mypy/issues/1297#issuecomment-508593494

I tried to avoid using mypy comments as much as possible, and go with restructuring the Python code, unless changing the code made things worse/complicated.

Instead of following that approach, are you recommending that it is better to go with the mypy comment so that:

```
try:
    from blah import foo
except ImportError:
    foo = None  # type: ignore[assignment]
```

??

PR #1292, opened 2020-06-16
Remove _connDone parameter from FileDescriptor.loseConnection()

rodrigc:
https://twistedmatrix.com/trac/ticket/9849

The signature of FileDescriptor.loseConnection() was modified here:
https://github.com/twisted/twisted/commit/081d393ab03da92d744d8fb2b5d77705662a0caa

I think extending that signature was wrong, since it doesn't match the base class signature of that method.


adiroiban (approved):
Looks good. Thanks!

PR #1293, opened 2020-06-16
Add some function annotations in twisted.internet

rodrigc:
https://twistedmatrix.com/trac/ticket/9850

Based on looking at https://github.com/twisted/twisted/pull/1265

I think if more of the functions in twisted.internet add function annotations,
it would help us to catch more subtle bugs by using mypy.

### DIFF
--- a/src/twisted/internet/abstract.py
+++ b/src/twisted/internet/abstract.py
@@ -8,6 +8,7 @@ Support for generic select()able objects.
 
 
 from socket import AF_INET, AF_INET6, inet_pton, error
+from typing import Sequence
 
 from zope.interface import implementer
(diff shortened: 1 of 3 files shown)


adiroiban (approved):
This looks awesome.

I see that python 3.5 is te minimum version and I think this can be merged.

Thanks!

rodrigc:
@clokep Can you take a look at this, and see if it would have helped find the problem in https://github.com/twisted/twisted/pull/1265 using mypy?  You fixed https://github.com/twisted/twisted/pull/1265 separately and that is good, but it is unfortunate you had to find out the problem at run time!

PR #1294, opened 2020-06-17
Eliminate more mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9851

### DIFF
--- a/src/twisted/conch/test/test_insults.py
+++ b/src/twisted/conch/test/test_insults.py
@@ -2,6 +2,8 @@
 # Copyright (c) Twisted Matrix Laboratories.
 # See LICENSE for details.
 
+from typing import Optional, Type
+
 from twisted.trial import unittest
 from twisted.test.proto_helpers import StringTransport
 
@@ -12,6 +14,7 @@ from twisted.conch.insults.insults import (CS_UK, CS_US, CS_DRAWING,
                                            BLINK, UNDERLINE)
 from twisted.conch.insults.insults import G0, G1
 from twisted.conch.insults.insults import modes, privateModes
+from twisted.internet.protocol import Protocol
 from twisted.python.compat import intToBytes, iterbytes
 from twisted.python.constants import ValueConstant, Values
 
@@ -201,18 +204,27 @@ def testByte%(groupName)s(self):
 
     self.verifyResults(transport, proto, parser)
 """
-class ByteGroupingsMixin(MockMixin):
-    protocolFactory = None
 
-    for word, n in [('Pairs', 2), ('Triples', 3), ('Quads', 4), ('Quints', 5), ('Sexes', 6)]:
+
+
+class ByteGroupingsMixin(MockMixin):
+    protocolFactory = None  # type: Optional[Type[Protocol]]
+
+    for word, n in [('Pairs', 2), ('Triples', 3), ('Quads', 4), ('Quints', 5),
+                    ('Sexes', 6)]:
         exec(_byteGroupingTestTemplate % {'groupName': word, 'bytesPer': n})
     del word, n
 
     def verifyResults(self, transport, proto, parser):
-        result = self.assertCall(occurrences(proto).pop(0), "makeConnection", (parser,))
+        result = self.assertCall(occurrences(proto).pop(0), "makeConnection",
+                                 (parser,))
         self.assertEqual(occurrences(result), [])
 
+
+
 del _byteGroupingTestTemplate
+
+
 
 class ServerArrowKeysTests(ByteGroupingsMixin, unittest.TestCase):
     protocolFactory = ServerProtocol
--- a/src/twisted/conch/test/test_transport.py
+++ b/src/twisted/conch/test/test_transport.py
@@ -11,6 +11,7 @@ import re
 import string
 import struct
 import types
+from typing import Optional, Type
 
 from hashlib import md5, sha1, sha256, sha384, sha512
 from twisted import __version__ as twisted_version
@@ -395,7 +396,7 @@ class TransportTestCase(TestCase):
     """
     Base class for transport test cases.
     """
-    klass = None
+    klass = None  # type: Optional[Type[transport.SSHTransportBase]]
 
     if dependencySkip:
         skip = dependencySkip
@@ -498,7 +499,7 @@ class BaseSSHTransportBaseCase:
     Base case for TransportBase tests.
     """
 
-    klass = MockTransportBase
+    klass = MockTransportBase  # type: Optional[Type[transport.SSHTransportBase]]  # noqa
 
 
 
@@ -1411,7 +1412,7 @@ class ServerSSHTransportBaseCase(ServerAndClientSSHTransportBaseCase):
     Base case for SSHServerTransport tests.
     """
 
-    klass = transport.SSHServerTransport
+    klass = transport.SSHServerTransport  # type: Optional[Type[transport.SSHTransportBase]] # noqa
 
 
     def setUp(self):
@@ -1929,7 +1930,7 @@ class ClientSSHTransportBaseCase(ServerAndClientSSHTransportBaseCase):
     Base case for SSHClientTransport tests.
     """
 
-    klass = transport.SSHClientTransport
+    klass = transport.SSHClientTransport  # type: Optional[Type[transport.SSHTransportBase]]  # noqa
 
 
     def verifyHostKey(self, pubKey, fingerprint):
--- a/src/twisted/conch/checkers.py
+++ b/src/twisted/conch/checkers.py
@@ -14,14 +14,14 @@ import errno
 try:
     import pwd
 except ImportError:
-    pwd = None
+    pwd = None  # type: ignore[assignment]
 else:
     import crypt
 
 try:
     import spwd
 except ImportError:
-    spwd = None
+    spwd = None  # type: ignore[assignment]
 
 from zope.interface import providedBy, implementer, Interface
 
--- a/src/twisted/conch/test/test_agent.py
+++ b/src/twisted/conch/test/test_agent.py
@@ -13,7 +13,7 @@ from twisted.test import iosim
 try:
     import cryptography
 except ImportError:
-    cryptography = None
+    cryptography = None  # type: ignore[assignment]
 
 try:
     import pyasn1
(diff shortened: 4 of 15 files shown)


src/twisted/conch/test/test_channel.py
```diff
@@ -93,13 +94,12 @@ def connectSSHTransport(service, hostAddress=None, peerAddress=None):
 
 
 
-class ChannelTests(unittest.TestCase):
+@skipIf(skipTest, skipTest)
```
adiroiban: Since we are looking at strict typing... should this be something like this?
With an explicit bool type and not with str.
```suggestion
@skipIf(bool(skipTest), skipTest)
```

src/twisted/conch/test/test_insults.py
```diff
@@ -201,19 +204,28 @@ def testByte%(groupName)s(self):
 
     self.verifyResults(transport, proto, parser)
 """
+
+
+
 class ByteGroupingsMixin(MockMixin):
-    protocolFactory = None
+    protocolFactory = None  # type: Optional[Type[Protocol]]
```
adiroiban, line 211: Maybe the mixin can be initialzied with a generic abstract protocol which fails at any call or access  ...so that we don't need to have it as "optional".

the protocolFactory is not really optional... it is defined there just as a way to document the dependencies/parameters of this mixin ...and to have the code fail if the mixin is used without defining the protocolFactory


but I guess that since this is testing code is ok to have it  as `None`

src/twisted/conch/test/test_transport.py
```diff
@@ -1411,7 +1412,7 @@ class ServerSSHTransportBaseCase(ServerAndClientSSHTransportBaseCase):
     Base case for SSHServerTransport tests.
     """
 
-    klass = transport.SSHServerTransport
+    klass = transport.SSHServerTransport  # type: Optional[Type[transport.SSHTransportBase]] # noqa
```
adiroiban, line 1415: maybe python code style checker should be updated and ignore long lines if they are type definition. 

There is no much value in adding `# noqa` to each typing line.

src/twisted/internet/cfreactor.py
```diff
@@ -44,7 +44,8 @@
 _preserveSOError = 1 << 6
 
 
-class _WakerPlus(_Waker):
+
+class _WakerPlus(_Waker):  # type: ignore[valid-type,misc]
```
adiroiban: I don't understand this type definition.

src/twisted/words/protocols/irc.py
```diff
@@ -170,7 +171,7 @@ class _CommandDispatcherMixin(object):
     @type prefix: C{str}
     @ivar prefix: Command handler prefix, used to locate handler attributes
     """
-    prefix = None
+    prefix = None  # type: Optional[str]
```
adiroiban: why not initiate with an empty string ?

src/twisted/words/protocols/irc.py
```diff
@@ -262,7 +263,7 @@ class IRC(protocol.Protocol):
     buffer = ""
     hostname = None
 
-    encoding = None
+    encoding = None  # type: Optional[str]
```
adiroiban: maybe initiate with the default `ascii` encoding.

src/twisted/words/protocols/jabber/xmlstream.py
```diff
@@ -163,7 +164,7 @@ class ConnectAuthenticator(Authenticator):
     Authenticator for initiating entities.
     """
 
-    namespace = None
+    namespace = None  # type: Optional[str]
```
adiroiban: I would say that this can be initialized to the empty string and then the xmlstream code can handle an empty string as if no namespace is required.

adiroiban (commented):
Thanks for the changes.

I think that using `Optional` inside the test code is ok... but for the main code, is best if we have a single type

It would be easier for review if the type annotation changes are made in a separate PR to the normal changes which don't require type annotation... or from the changes in test code

src/twisted/conch/test/test_insults.py
mthuurne: In Python 3.6, you could annotate the variable without assigning to it:
```py
class ByteGroupingsMixin(MockMixin):
    protocolFactory: Type[Protocol]
```
I don't know if there is an alternative syntax for older Python versions, but if there isn't, this could be a motivation to move to 3.6 now.

src/twisted/conch/test/test_channel.py
rodrigc: The first argument to skipIf takes a condition which is internally evaluated:
https://github.com/python/cpython/blob/3.8/Lib/unittest/case.py#L132

so I think it is OK to leave it as-is without an explicit bool type.

src/twisted/conch/test/test_insults.py
rodrigc: @adiroiban I wasn't sure whether to change these usages of `None` in base classes or not.  For these particular cases, I took the conservative approach and just used `Optional` for the annotation.

@mthuurne I brought up the topic of dropping Python 3.5 support on the mailing list:
https://twistedmatrix.com/pipermail/twisted-python/2020-May/065066.html
No decision on that has been made yet, but @glyph and @hawkowl seem OK with the idea.  If Python 3.6 can become the minimal version of Python for Twisted, then Twisted can definitely use PEP-0526 variable annotations, but until Python 3.6 support is dropped, we can only add annotations which are Python 3.5 compatible.

wsanchez (commented):
At 102 files, with different categories of change, this is far too much change for a single PR.

I strongly suggest that you make separate PRs for `# type: ignore` comments versus code changes versus additions of type hints, and that for the latter two, that you make separate PRs for each submodule, unless the total diff is a lot smaller than this.

src/twisted/conch/test/test_channel.py
wsanchez: I don't think wrapping in `bool` adds any value here if it's not causing a typing error.
If `skipIf` is not type hinted yet, when it gets a type hint, this will raise an error if a bool is necessary, and will not if it isn't.

src/twisted/conch/test/test_insults.py
wsanchez: This conversation is a strong endorsement for why this change shouldn't be mixed in with a bunch of totally dissimilar changes, by the way.

rodrigc:
@wsanchez OK, I will break this patch up into separate PR's

Now that I'm getting the hang of things, I have started creating smaller PR's such as:

https://github.com/twisted/twisted/pull/1295
https://github.com/twisted/twisted/pull/1296
https://github.com/twisted/twisted/pull/1297
https://github.com/twisted/twisted/pull/1298

src/twisted/conch/test/test_insults.py
wsanchez: @mthuurne Good point. You can't do that in 3.5 with a type comment?

One problem is that not assigning to `None` is arguably a compatibility issue? Some code might check for `None` to see if it's been set…

An option that doesn't change the code is to set the type to `Type[Protocol]` and also use a `# type: ignore` comment to suppress the assignment error.  That correctly acknowledges the status quo and leave the question of whether to change the behavior to a separate PR.

I think that when add typing to a large codebase like Twisted, that may be the better strategy in a lot of cases; we can seek and destroy all the `# type: ignore` comments *after* documenting the current code behavior with accurate type hints.

src/twisted/conch/test/test_transport.py
wsanchez: +1

glyph:
> At 102 files, with different categories of change, this is far too much change for a single PR.

I really have to echo this - the maximum reviewable size for a PR is about 400 lines.

> OK, I will break this patch up into separate PR's

Thanks for doing that!

Are you finding that you have a lot of types that depend on each other?  One thing you can do is you can have a set of "stacked" PRs, that target each others' branches.

So for example you start with PR `A`, with branch `branch-for-a`; that gets a merge-target of `trunk` as usual.  While you're waiting for review on that, you can start PR `B`, with branch `branch-for-b`, but set `B`'s merge target to point at `branch-for-a` rather than `trunk`.  When `A` is merged, Github will *automatically* set the merge-base for `B` to `trunk`, which makes it possible to stage all your PRs in sequence and not do any manual work when each one gets landed.

rodrigc:
@glyph  I tend to try to do very isolated commits with few interdependencies, so I think I can split things up into separate PR's.  I already have 6 separate PR's and have more on the way.

src/twisted/conch/test/test_transport.py
adiroiban: Or remove support for Python 3.5 and use standard type annotation and get rid of all these silly comments :)

src/twisted/conch/test/test_channel.py
adiroiban: ok. I am still not sure what is the purpose of all this type checking in Python :)

I was thinking that if the direction of Twisted is to remove dynamic typing, then any change to the existing code should be done with strict types and for example don't mix an empty string with boot or 0 .

src/twisted/conch/test/test_insults.py
adiroiban: > One problem is that not assigning to None is arguably a compatibility issue? Some code might check for None to see if it's been set…

Thanks for your input.

In this particular usage, this is just testing code.
But yes, for compatibility reasons, I think that for the general code None should be accepted in many places.

I was thinking that we add strict typing in Twisted so that when an user of the library will run a type checker on the code the errors will help write a better code.

This is all very subjecting:  for a better code, an implementation should have dedicated `isStarted` attributes and don't use the value of protocolFactory to decide if something is started or not.

src/twisted/internet/cfreactor.py
rodrigc: @adiroiban This is needed because the class `_Waker` is defined in two different ways in branches of an if statement, so mypy will complain.  I have moved this patch to another PR: https://github.com/twisted/twisted/pull/1306/files#diff-7968560e2c5c7ff374a8a890a3d87bd7L47

rodrigc:
@adiroiban @mthuurne @wsanchez In this PR, I have reduce the number of changed files to 16.

I have also split out the changes that were originally here into these PR's:

https://github.com/twisted/twisted/pull/1313
https://github.com/twisted/twisted/pull/1314
https://github.com/twisted/twisted/pull/1315
https://github.com/twisted/twisted/pull/1316


When you have time, please re-review this.

adiroiban (approved):
Look good. Thanks.

PR #1296, opened 2020-06-19
[mypy] Remove duplicate definition of workers in src/twisted/python/threadpool.py

rodrigc:
https://twistedmatrix.com/trac/ticket/9854

This removes this mypy error:

```
src/twisted/python/threadpool.py:84:6: error: Name 'workers' already defined on line 44  [no-redef]
```


adiroiban (approved):
Thanks. All good!

PR #1297, opened 2020-06-19
[mypy] Fix error: "All conditional function variants must have identical signatures"

rodrigc:
https://twistedmatrix.com/trac/ticket/9855

Eliminate these mypy errors:

```
src/twisted/python/lockfile.py:53:9: error: All conditional function variants must have identical signatures  [misc]
src/twisted/python/lockfile.py:69:5: error: All conditional function variants must have identical signatures  [misc]
src/twisted/python/lockfile.py:95:5: error: All conditional function variants must have identical signatures  [misc]
src/twisted/python/lockfile.py:112:5: error: All conditional function variants must have identical signatures  [misc]
src/twisted/internet/defer.py:46:5: error: All conditional function variants must have identical signatures  [misc]
src/twisted/protocols/ident.py:160:9: error: All conditional function variants must have identical signatures  [misc]
src/twisted/persisted/sob.py:79:13: error: All conditional function variants must have identical signatures  [misc]
src/twisted/web/_flatten.py:20:5: error: All conditional function variants must have identical signatures  [misc]
```

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -42,8 +42,9 @@ except ImportError:
         def run(f, *args, **kwargs):
             return f(*args, **kwargs)
 
-
-    def _copy_context():
+    # typing ignored due to:
+    # https://github.com/python/typeshed/issues/4249
+    def _copy_context():  # type: ignore[misc]
         return _NoContext
 
 log = Logger()
--- a/src/twisted/python/lockfile.py
+++ b/src/twisted/python/lockfile.py
@@ -50,7 +50,9 @@ else:
         ERROR_ACCESS_DENIED = 5
         ERROR_INVALID_PARAMETER = 87
 
-        def kill(pid, signal):
+        # typing ignored due to:
+        # https://github.com/python/typeshed/issues/4249
+        def kill(pid, signal):  # type: ignore[misc]
             try:
                 OpenProcess(0, 0, pid)
             except pywintypes.error as e:
@@ -66,7 +68,9 @@ else:
     _open = open
 
 
-    def symlink(value, filename):
+    # typing ignored due to:
+    # https://github.com/python/typeshed/issues/4249
+    def symlink(value, filename):  # type: ignore[misc]
         """
         Write a file at C{filename} with the contents of C{value}. See the
         above comment block as to why this is needed.
@@ -92,7 +96,9 @@ else:
             raise
 
 
-    def readlink(filename):
+    # typing ignored due to:
+    # https://github.com/python/typeshed/issues/4249
+    def readlink(filename):  # type: ignore[misc]
         """
         Read the contents of C{filename}. See the above comment block as to why
         this is needed.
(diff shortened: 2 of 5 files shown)


src/twisted/python/lockfile.py
```diff
@@ -50,7 +50,7 @@ def unique():
         ERROR_ACCESS_DENIED = 5
         ERROR_INVALID_PARAMETER = 87
 
-        def kill(pid, signal):
+        def kill(pid, signal):  # type: ignore[misc]
```
adiroiban, line 55: why should this be ignored? I think that will kill always have `ini, int`

rodrigc: @adiroiban I explained this a bit in the commit log for https://github.com/twisted/twisted/pull/1297/commits/c6f85c065ef91ccdd5848cab4e30c67b36cfcbe4

```
Tell mypy to ignore functions which try to replace functions implemented

in C.

For example, 'os.kill' is implemented in C,
and as far as mypy is concerned, has a different signature than
def kill(pid, signal)
```

If you do:

```
from os import kill
help(kill)
```

you will see that the signature of the function is:

```
kill(pid, signal, /)
```

The only way to implement a `/` parameter outside of C, is to use Python 3.8 or higher which implements PEP 570.  So in this case it was easier to put the comment for mypy to ignore this line.

adiroiban: Thanks for the explanation. TIL :)

I think that this information should not be put in a commit message, or a review response, but rather as a code comment.

At the same time, I think that mypy should have some special annotation for positional only arguments, and to not just use an ignore annotation.
I see that mypy only got this for python3.8 https://github.com/python/mypy/pull/6900 without any backward compatibility.

So I think that this needs a comment, at least to indicate that this can be removed once Twisted drops support for 3.5, 3.6 and 3.7.

Thanks

src/twisted/internet/defer.py
```diff
@@ -43,7 +43,7 @@ def run(f, *args, **kwargs):
             return f(*args, **kwargs)
 
 
-    def _copy_context():
+    def _copy_context():  # type: ignore[misc]
```
adiroiban, line 47: I don't understand this change.
Is this from the same category as " Tell mypy to ignore functions which try to replace functions implemented " ?

rodrigc: If you do not put that comment there, mypy will give you an error such as:

```
src/twisted/internet/defer.py:46:5: error: All conditional function variants must have identical signatures  [misc]
```

From this line:

```
from contextvars import copy_context as _copy_context
```

The `copy_context` function is coming from a C library, not Python code.  So looks like mypy considers that `from contextvars import copy_context` is pulling in a function which
has a signature which is different from the Python function:

```
def _copy_context():
    return _NoContext
```

adiroiban: Thanks for the info.

I guess that ignoring it here will not do much harm.

But as a general note, this looks to me like a bug in mypy, and I don't think that we should update twisted code to ignore types just because of a mypy bug.

If we ignore something due to a bug, I think that an extra comment should be added and a link to the mypy bug.
If the bug is fixed we can track it and remove the ignore statement.

As a general rule, I think that each mypy ignore comment should have extra information inside the code to explain why it was ignored.
```suggestion
    # The typing is ignored as mypy is confused by
    # from contextvars import copy_context as _copy_context
    # SOME LINK TO MYPY BUG REPORT
    def _copy_context():  # type: ignore[misc]
```

adiroiban (approved):
All good,

Please add extra comment for typing which is ignored due to mypy bugs.

PR #1298, opened 2020-06-19
[mypy] Implement some functions defined in zope interfaces

rodrigc:
https://twistedmatrix.com/trac/ticket/9852

When running mypy over the twisted codebase with:

```
tox -e mypy
```

There are many errors that look like this:

```
src/twisted/words/protocols/jabber/sasl_mechanisms.py:47:1: error: 'Anonymous' is missing following 'ISASLMechanism' interface members: getResponse.  [misc]
src/twisted/words/protocols/jabber/sasl_mechanisms.py:61:1: error: 'Plain' is missing following 'ISASLMechanism' interface members: getResponse.  [misc]
src/twisted/internet/_dumbwin32proc.py:110:1: error: 'Process' is missing following 'twisted.internet.interfaces.ITransport' interface members: getHost, getPeer.
src/twisted/internet/abstract.py:151:1: error: 'FileDescriptor' is missing following 'ITransport' interface members: getHost, getPeer.  [misc]
src/twisted/internet/process.py:623:1: error: 'Process' is missing following 'twisted.internet.interfaces.ITransport' interface members: getHost, getPeer.  [misc]
src/twisted/internet/process.py:959:1: error: 'PTYProcess' is missing following 'twisted.internet.interfaces.ITransport' interface members: getHost, getPeer.  [misc]
src/twisted/internet/process.py:959:1: error: 'PTYProcess' is missing following 'IProcessTransport' interface members: closeChildFD, writeToChild.  [misc]
src/twisted/internet/_resolver.py:31:1: error: 'HostResolution' is missing following 'IHostResolution' interface members: cancel.  [misc]
src/twisted/internet/base.py:504:1: error: 'ReactorBase' is missing following 'IReactorCore' interface members: run.  [misc]
src/twisted/internet/_newtls.py:162:1: error: 'ConnectionMixin' is missing following 'twisted.internet.interfaces.ITCPTransport' interface members: getHost, getPeer,
```

In an effort to quieten these errors, this
PR adds comments to turn off these mypy warnings in a few places.

In a few other places, I implemented no-op methods in some classes, to appease mypy.
add comments

### DIFF
--- a/src/twisted/application/runner/test/test_pidfile.py
+++ b/src/twisted/application/runner/test/test_pidfile.py
@@ -474,3 +474,70 @@ class DummyFilePath(object):
 
     def exists(self):
         return self._exists
+
+
+    def sep(self):
+        # IFilePath.sep
+        raise NotImplementedError("Unimplemented: DummyFilePath.sep")
+
+
+    def child(self):
+        # IFilePath.child
+        raise NotImplementedError("Unimplemented: DummyFilePath.child")
+
+
+    def changed(self):
+        # IFilePath.changed
+        raise NotImplementedError("Unimplemented: DummyFilePath.changed")
+
+
+    def getsize(self):
+        # IFilePath.getsize
+        raise NotImplementedError("Unimplemented: DummyFilePath.getsize")
+
+
+    def getModificationTime(self):
+        # IFilePath.getModificationTime
+        raise NotImplementedError(
+            "Unimplemented: DummyFilePath.getModificationTime")
+
+
+    def getStatusChangeTime(self):
+        # IFilePath.getStatusChangeTime
+        raise NotImplementedError(
+            "Unimplemented: DummyFilePath.getStatusChangeTime")
+
+
+    def getAccessTime(self):
+        # IFilePath.getAccessTime
+        raise NotImplementedError("Unimplemented: DummyFilePath.getAccessTime")
+
+
+    def isdir(self):
+        # IFilePath.isdir
+        raise NotImplementedError("Unimplemented: DummyFilePath.isdir")
+
+
+    def isfile(self):
+        # IFilePath.isfile
+        raise NotImplementedError("Unimplemented: DummyFilePath.isfile")
+
+
+    def children(self):
+        # IFilePath.children
+        raise NotImplementedError("Unimplemented: DummyFilePath.children")
+
+
+    def basename(self):
+        # IFilePath.basename
+        raise NotImplementedError("Unimplemented: DummyFilePath.basename")
+
+
+    def parent(self):
+        # IFilePath.parent
+        raise NotImplementedError("Unimplemented: DummyFilePath.parent")
+
+
+    def sibling(self):
+        # IFilePath.sibling
+        raise NotImplementedError("Unimplemented: DummyFilePath.sibling")
--- a/src/twisted/conch/insults/helper.py
+++ b/src/twisted/conch/insults/helper.py
@@ -456,6 +456,64 @@ class TerminalBuffer(protocol.Protocol):
         return b'\n'.join(lines)
 
 
+    def getHost(self):
+        # ITransport.getHost
+        raise NotImplementedError("Unimplemented: TerminalBuffer.getHost")
+
+
+    def getPeer(self):
+        # ITransport.getPeer
+        raise NotImplementedError("Unimplemented: TerminalBuffer.getPeer")
+
+
+    def loseConnection(self):
+        # ITransport.loseConnection
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.loseConnection")
+
+
+    def writeSequence(self, data):
+        # ITransport.writeSequence
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.writeSequence")
+
+
+    def horizontalTabulationSet(self):
+        # ITerminalTransport.horizontalTabulationSet
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.horizontalTabulationSet")
+
+
+    def tabulationClear(self):
+        # TerminalTransport.tabulationClear
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.tabulationClear")
+
+
+    def tabulationClearAll(self):
+        # TerminalTransport.tabulationClearAll
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.tabulationClearAll")
+
+
+    def doubleHeightLine(self, top=True):
+        # ITerminalTransport.doubleHeightLine
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.doubleHeightLine")
+
+
+    def singleWidthLine(self):
+        # ITerminalTransport.singleWidthLine
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.singleWidthLine")
+
+
+    def doubleWidthLine(self):
+        # ITerminalTransport.doubleWidthLine
+        raise NotImplementedError(
+            "Unimplemented: TerminalBuffer.doubleWidthLine")
+
+
 
 class ExpectationTimeout(Exception):
     pass
--- a/src/twisted/mail/imap4.py
+++ b/src/twisted/mail/imap4.py
@@ -5194,6 +5194,11 @@ class MemoryAccount(MemoryAccountWithoutNamespaces):
         return None
 
 
+    def getUserNamespaces(self):
+        # INamespacePresenter.getUserNamespaces
+        return None
+
+
 
 _statusRequestDict = {
     'MESSAGES': 'getMessageCount',
--- a/src/twisted/python/test/test_components.py
+++ b/src/twisted/python/test/test_components.py
@@ -587,8 +587,11 @@ class IProxiedSubInterface(IProxiedInterface):
         """
 
 
+
 @implementer(IProxiedInterface)
-class Yayable(object):
+class Yayable(object):  # type: ignore[misc]
+    # class does not implement Attribute ifaceAttribute
+    # so we need to turn off mypy warning
     """
     A provider of L{IProxiedInterface} which increments a counter for
     every call to C{yay}.
@@ -599,6 +602,7 @@ class Yayable(object):
     def __init__(self):
         self.yays = 0
         self.yayArgs = []
+
 
     def yay(self, *a, **kw):
         """
@@ -611,12 +615,17 @@ class Yayable(object):
 
 
 @implementer(IProxiedSubInterface)
-class Booable(object):
+class Booable(object):  # type: ignore[misc]
+    # class does not implement Attribute ifaceAttribute
+    # so we need to turn off mypy warning
     """
     An implementation of IProxiedSubInterface
     """
+
     yayed = False
     booed = False
+
+
     def yay(self):
         """
         Mark the fact that 'yay' has been called.
--- a/src/twisted/application/test/test_internet.py
+++ b/src/twisted/application/test/test_internet.py
@@ -129,6 +129,18 @@ class FakePort(object):
         self.deferred = Deferred()
         return self.deferred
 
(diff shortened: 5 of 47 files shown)


src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -474,3 +474,107 @@ def remove(self):
 
     def exists(self):
         return self._exists
+
+
+    def sep(self):
+        """
+        @see: L{IFilePath.sep}
+
+        Not implemented
```
adiroiban: shouldn't this raise NotImplemented error?
In the case in which some new test is using it, it will raise an explicit error, rather than a generic error about the None type missing some string attributes.

src/twisted/conch/insults/helper.py
```diff
@@ -456,6 +456,76 @@ def __bytes__(self):
         return b'\n'.join(lines)
 
 
+    def getHost(self):
+        """
+        @see: L{ITransport.getHost}
+        """
+        raise NotImplementedError()
```
adiroiban:
```suggestion
        raise NotImplementedError('getHost not implemented')
```
When raising a NotImplementedError I think is best to provide the name... in this way, when the error is raise and you don't have the traceback, you can get a clue from the text of the error

```
>>> str(NotImplementedError())
''
>>> str(NotImplementedError('some_method'))
'some_method'
>>> str(NotImplementedError('some_method not implemented'))
'some_method not implemented' 
```

src/twisted/internet/_newtls.py
```diff
@@ -159,7 +159,13 @@ def startTLS(transport, contextFactory, normal, bypass):
 
 
 @implementer(ITLSTransport)
-class ConnectionMixin(object):
+class ConnectionMixin(object):  # type: ignore[misc]
```
adiroiban: I see that this is only used in a single place

src/twisted/internet/tcp.py
```
@implementer(interfaces.ITCPTransport, interfaces.ISystemHandle)
class Connection(_TLSConnectionMixin, abstract.FileDescriptor, _SocketCloser,
                 _AbortingMixin):
```

So maybe this should not be a mixin, but rather a full transport

```suggestion
class ConnectionMixin(abstract.FileDescriptor):
```

but I see that abstract.FileDescriptor is not really fully implementing the ITransport interface

I guess is best to leave this as a separate ticket.

So instead of `# XXX` create a ticket and add the link to that ticket in the comment to mark is as a "TODO" or "FIXME"

src/twisted/internet/unix.py
```diff
@@ -52,7 +52,11 @@ def _ancillaryDescriptor(fd):
 
 
 @implementer(interfaces.IUNIXTransport)
-class _SendmsgMixin(object):
+class _SendmsgMixin(object):  # type: ignore[misc]
```
adiroiban: maybe this mixing should not define that it implement an interface... if it doesn't implement it.

Instead, we can have

```
@implementer(interfaces.IUNIXTransport)
class Server(_SendmsgMixin, tcp.Server):

@implementer(interfaces.IUNIXTransport)
class Client(_SendmsgMixin, tcp.BaseClient):
```

Is a bit of a code duplication, but I think that this is the correct way of writing this code.

src/twisted/mail/imap4.py
```diff
@@ -5196,6 +5196,13 @@ def getOtherNamespaces(self):
         return None
 
 
+    def getUserNamespaces(self):
+        """
+        @see: L{INamespacePresenter.getUserNamespaces}
+        """
+        return None
```
adiroiban, line 5199: I know that in the docstring we have 
```
If no namespaces of this type exist, None should be returned.
```

But what would happen if we return an empty tuple here?

It will be a sequence and the method will always return the same type.

src/twisted/python/test/test_components.py
```diff
@@ -587,8 +587,11 @@ def boo():
         """
 
 
+
 @implementer(IProxiedInterface)
-class Yayable(object):
+class Yayable(object):  # type: ignore[misc]
+    # class does not implement Attribute ifaceAttribute
+    # so we need to turn off mypy warning
```
adiroiban, line 594:
```suggestion
class Yayable(object):
   # This is replaced by each test. Here we add a default value. 
    ifaceAttribute = object()
```

adiroiban (commented):
Thanks for the changes.

This is an easy to review chunk of changes :)

I am not sure what are the rules for returning None or raising a NotImplementedError from the stub methods.

I don't see the value of `# type: ignore[misc]`.
We should either work at defining the proper types ,  or not using types at all.

src/twisted/application/runner/test/test_pidfile.py
mthuurne: Also I think it's better to have no docstring rather than one containing only `@see`, since if there is no docstring, pydoctor will inherit the docstring from the superclass/interface.

src/twisted/python/test/test_components.py
rodrigc: @adiroiban I made the change you suggested, but then the test failed with:

```
[FAIL]
  File "/Users/crodrigues/Twisted/src/twisted/python/test/test_components.py", line 907, in test_attributeCustomization
    self.assertFalse(hasattr(yayable, 'ifaceAttribute'))
  File "/Users/crodrigues/Twisted/src/twisted/trial/_synctest.py", line 390, in assertFalse
    super(_Assertions, self).assertFalse(condition, msg)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 699, in assertFalse
twisted.trial.unittest.FailTest: True is not false
```

So the test explicitly checks that ifaceAttribute is not there, before adding that attribute.

I'm not sure if it is OK to change the test?

src/twisted/mail/imap4.py
rodrigc: I'm not sure.  Maybe this interface should have been implemented to return an empty tuple instead of None, but I didn't want to change the semantics of the function, as defined by the interface.

src/twisted/internet/unix.py
rodrigc: I agree with your fix, and have implemented it.
`_SendmsgMixin` should not be marked as implementing `IUNIXTransport` if it does not fully implement it.

In twisted/internet/interfaces.py , maybe `IUNIXTransport` should have been a standalone interface, and not have inherited from `ITransport`.

Ideally, interfaces should be as minimal as possible, and avoiding inheritance hierarchies in interfaces is a good idea IMHO.

Since `_SendmsgMixin` is a private class, I think it is OK to change it as you suggest.

However, as much as possible, I want to avoid changing existing interfaces in Twisted.

src/twisted/conch/insults/helper.py
rodrigc: OK, I changed things to `raise NotImplementedError("some message")`

src/twisted/internet/_newtls.py
rodrigc: Since `ConnectionMixin` is defined in `_newtls.py`, maybe we can assume that entire module is private?
If so, would it be better to remove @implementer(ITLSTransport) from the ConnectionMixin, and add it to `twisted.internet.tcp.Connection`?

I really don't like how some of these interfaces have been defined and inherit from each other....

rodrigc: I've created a new PR for this: https://github.com/twisted/twisted/pull/1299

and removed the changes to ConnectionMixin from here.

src/twisted/internet/unix.py
rodrigc: I shifted that change over to https://github.com/twisted/twisted/pull/1299

src/twisted/application/runner/test/test_pidfile.py
rodrigc: @mthuurne I put the @see docstring in there, because when looking directly at the code, it is sometimes difficult to figure out what method comes from what interface.  I can take out the docstring though if people don't want it.

mthuurne: I guess for test code it doesn't matter much, but for modules that end users are going to be reading the documentation of, I think it's better to have the inherited documentation inlined by pydoctor, since that means it is accessible without having to navigate to a different page when reading the docs as HTML.

src/twisted/mail/imap4.py
adiroiban: True. Thanks. This should have the optional type for sure. Sorry for the noise.

src/twisted/python/test/test_components.py
adiroiban, line 594: It looks like ifaceAttribute was left out on purpose... but I think that we should have a comment to make it clear why the interface implementation is broken. 

```suggestion
class Yayable(object):  # type: ignore[misc]
    """
    This class does not defines the required ifaceAttribute so that it can be used
    to test that proxied attributes can also be removed.
    """
```

So it should have an ignore anyway, but I think that the test can be extended to be  explicit and also demonstrate that no wizardry is done and that only the instance attribute is removed. 

something like this

```diff
diff --git a/src/twisted/python/test/test_components.py b/src/twisted/python/test/test_components.py
index 17a571c47..bef7e31d8 100644
--- a/src/twisted/python/test/test_components.py
+++ b/src/twisted/python/test/test_components.py
@@ -571,6 +571,12 @@ class IProxiedInterface(Interface):
     ifaceAttribute = Attribute("""
         An example declared attribute, which should be proxied.""")
 
+
+    deleteAttribute = Attribute("""
+        Proxied attribute used in the test checking the attribute
+        delete operation.""")
+
+
     def yay(*a, **kw):
         """
         A sample method which should be proxied.
@@ -593,10 +599,16 @@ class Yayable(object):
     A provider of L{IProxiedInterface} which increments a counter for
     every call to C{yay}.
 
+    `deleteAttribute` is not implemented on purpose to support the delete
+    tests.
+
     @ivar yays: The number of times C{yay} has been called.
     """
+    ifaceAttribute = 'class-value'
+
 
     def __init__(self):
+        self.ifaceAttribute = 'instance-value'
         self.yays = 0
         self.yayArgs = []
 
@@ -743,12 +755,17 @@ class ProxyForInterfaceTests(unittest.SynchronousTestCase):
         """
         The attributes that proxy objects proxy should be deletable and affect
         the original object.
+
+        It only removes the instance attribute.
         """
         yayable = Yayable()
-        yayable.ifaceAttribute = None
+        yayable.ifaceAttribute = 'instance'
+        yayable.deleteAttribute = 'something'
         proxy = proxyForInterface(IProxiedInterface)(yayable)
         del proxy.ifaceAttribute
-        self.assertFalse(hasattr(yayable, 'ifaceAttribute'))
+        del proxy.deleteAttribute
+        self.assertEqual('class-value', yayable.ifaceAttribute)
+        self.assertFalse(hasattr(yayable, 'deleteAttribute'))
 
 
     def test_multipleMethods(self):
@@ -894,4 +911,4 @@ class ProxyForInterfaceTests(unittest.SynchronousTestCase):
         proxy.ifaceAttribute = thingy
         self.assertIs(yayable.ifaceAttribute, thingy)
         del proxy.ifaceAttribute
-        self.assertFalse(hasattr(yayable, 'ifaceAttribute'))
+        self.assertEqual('class-value', proxy.ifaceAttribute)
```

adiroiban (approved):
all good. thanks.
only minor comments.

For public API I also prefer not to have any docstring at all and let pydoctor do its magic.

For test code, rather than "@see: L{IListeningPort.getHost}" I prefer a comment explaining that this is just a stub implementation created to satisfy the interface.

With an IDE you already get the `@see` functionality.

src/twisted/python/test/test_components.py
adiroiban: I think that we can change the tests as the 'hasAttr' check is not 100% correct.

As we have seen here, even if the instance member is removed, hasAttr will still  return True...as the class has the attrribute.

I have proposed a patch to improve this test and make it explicit that only the instance attribute is removed.

PR #1300, opened 2020-06-20
[mypy] Use @property, @staticmethod, @skipIf decorators

rodrigc:
https://twistedmatrix.com/trac/ticket/9857

Use @property, @staticmethod, @skipIf decorators to reduce mypy errors.


adiroiban (approved):
This looks good, Thanks!

Hope mypy https://github.com/python/mypy/issues/8083 will be fixed soon so that we can use the property assignation syntax.

But the decorator is easier to read...so this change is an improvement.

Thanks again!

PR #1302, opened 2020-06-20
Fix Win32UnicodeEnvironmentTests.test_encodableUnicodeEnvironment for Azure pipelines

rodrigc:
https://twistedmatrix.com/trac/ticket/9858

@mgorny helped me debug this under this PR: https://github.com/twisted/twisted/pull/1305

When mgorny does a pull request from his GitHub ID, they following environment variable is set on Azure pipelines:

```
BUILD_SOURCEVERSIONAUTHOR Michał Górny
```

That triggered a codepath which caused the CI to fail on Windows, even if he did trivial changes like only changing whitespace.

This patch fixes the problem.  I also added another test, putting an environment variable similar to Michał's name.


rodrigc:
@mgorny can you code review this?  I am pretty sure that this test bug on Windows has been breaking the CI on your PR's.  I have cherry picked this fix to your PR's: https://github.com/twisted/twisted/pull/1278
https://github.com/twisted/twisted/pull/1275

to try it out.

I want to get your Python 3.9 PR's in (with your full non-ASCII name in the checkin comments!!)

mgorny:
It seems that something is still failing but I don't seem to grasp the error.

rodrigc:
The travis build failed, due to lint: https://travis-ci.com/github/twisted/twisted/jobs/351767847
I've pushed another fix for that.

rodrigc:
This post from @hawkowl is a very good description of UTF-8 vs. CP1252 encoding issues on Windows: https://twistedmatrix.com/pipermail/twisted-python/2020-June/065140.html

src/twisted/test/test_process.py
```diff
@@ -2383,25 +2337,46 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
...
+
+
+    def test_UTF8StringInEnvironment(self):
+        """
+        L{os.environ} (inherited by every subprocess on Windows) can
+        contain a UTF-8 string.
+        """
+        envKey = 'TWISTED_BUILD_SOURCEVERSIONAUTHOR'
+        envKeyBytes = b'TWISTED_BUILD_SOURCEVERSIONAUTHOR'
+        envVal = "Speciał Committór"
```
mgorny, line 2428: :+1:

src/twisted/test/test_process.py
```diff
@@ -2383,25 +2337,46 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
...
 
 
     def test_encodableUnicodeEnvironment(self):
         """
-        Test C{os.environ} (inherited by every subprocess on Windows) that
-        contains an ascii-encodable Unicode string. This is different from
-        passing Unicode environment explicitly to spawnProcess (which is not
-        supported on Python 2).
+        Test L{os.environ} (inherited by every subprocess on Windows) that
+        contains an ascii-encodable Unicode string.
```
adiroiban: This test is hard to understand.

I think that the docstring should be:

```
GetEnvironmentDictionary automatically encodes all environment variables using SOME encoding.
```

From the other tests, it looks like that `SOME` encoding is `UTF-8`

src/twisted/test/test_process.py
```diff
@@ -2383,25 +2337,46 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
     """
     Tests for Unicode environment on Windows
     """
-    goodKey = u'UNICODE'
-    goodValue = u'UNICODE'
+    goodKey = 'UNICODE'
+    goodKeyBytes = b'UNICODE'
+    goodValue = 'UNICODE'
+    goodValueBytes = b'UNICODE'
```
adiroiban:
```suggestion
    goodKey = 'key-\N{snoman}'
    goodValue = 'value-\N{sun}'
```

but I prefer to not have them defined here... but rather defined the only inside the test.

src/twisted/test/test_process.py
```diff
@@ -2383,25 +2337,46 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
...
-        def gotEnvironment(environ):
-            self.assertEqual(
-                environ[self.goodKey.encode('ascii')],
-                self.goodValue.encode('ascii'))
+        p = GetEnvironmentDictionary.run(reactor, [], os.environ)
+
+        def gotEnvironment(environb):
+            self.assertEqual(environb[self.goodKeyBytes], self.goodValueBytes)
+
+        return p.getResult().addCallback(gotEnvironment)
```
adiroiban: I prefer to have all the test data inside the test, in this way it is easier to read... similar to what you did in test_UTF8StringInEnvironment

and maybe have a single test.

```suggestion
        os.environ['key-\N{snowman}'] = 'value-\N{sun}'
        os.environ['key-ascii''] = 'value-ascii'
        self.addCleanup(operator.delitem, os.environ, 'key-\N{snowman}')
        self.addCleanup(operator.delitem, os.environ, 'key-ascii')
     
        p = GetEnvironmentDictionary.run(reactor, [], os.environ)

        def gotEnvironment(environb):
           self.assertEqual(environb[b'key-ascii'], b'value-ascii')
            self.assertEqual(environb[b'key-\xe2\x98\x83'], b'value-\xe2\x98\x89')

        return p.getResult().addCallback(gotEnvironment)
```

src/twisted/test/test_process.py
```diff
@@ -2383,25 +2337,46 @@ class Win32UnicodeEnvironmentTests(unittest.TestCase):
...
+
+
+    def test_UTF8StringInEnvironment(self):
+        """
+        L{os.environ} (inherited by every subprocess on Windows) can
+        contain a UTF-8 string.
```
adiroiban: I am not sure I understand this docstring. ... will it handle Unicode key or value.

It looks like in this test, the environment contain Unicode (not UTF-8) and once  GetEnvironmentDictionary is called the value is automatically converted/encoded to UTF-8.

Does it mean that only the value can be Unicode?

I looks like both tests will have the same unicode, so my suggestion is to have a single test.

azure-pipelines/windows_test_jobs.yml
```diff
@@ -11,13 +11,15 @@ parameters:
...
-
 jobs:
 - ${{ each pyver in parameters.pythonVersions }}:
   - job: ${{ format('windows_{0}_{1}reactor', pyver.key, parameters.reactor) }}
     displayName: ${{ format('Windows Py{0} {1}reactor', pyver.value, parameters.reactor) }}
     pool:
       vmImage: ${{ parameters.imageName }}
+    variables:
+      PYTHONUTF8: 1
+      PYTHONIOENCODING: "utf-8:surrogateescape"
```
adiroiban, line 28: I think that these variables need a comment.
Why do we need to set them and why not use the default values.

if these variables are required to run twisted under Windows, I think that this should be reflected in the public documentation.

adiroiban (changes requested):
Thanks for the change.

I would like to have the docstring of the tests updated, to make it clear what is the expected behaviour.

At the same time, I think that this should be a bugfix.
I guess that this does not affect only our CI, but also anyone trying to use spawn process on WIndows with unicode EVN values.

So the newsfragment can be something like this:

twisted.internet.reactor.spawnProcess can now be used with Unicode environment variables.

It would also have to have a full functional/integrated test to check if the unicode environment variables are available in side the new process as unicode or as UTF-8 bytes.

That is, extend Win32ProcessTests to have an test_osEnviron_bytes and test_osEnviron_unicode , similar to test_stdinReader_bytesArgs/test_stdinReader_unicodeArgs

Let me know if you need help with the tests and I suggest a path for such a test.

In the unit tests I see that the environ is converted to UTF-8, but is not clear how the environmental is actually available/visible inside the spawned process.

And I think that this should be a generic test, for all reactors and not only for Windows / iocp.

I see that you made a change to posix_reactor, but I don't see any updated tests for Unicode env vars.
Does it mean that posix Unicode env vars were already working, and we already have a test for that.
In that case, I think that those existing tests should also be enabled for Windows.

hawkowl:
I think changing the Reactor portion is wrong -- Windows environment variables are stored internally as Unicode, and the APIs take Unicode. Therefore, we should ask users to provide the native (Unicode) representation in the API, decoding it ourselves from bytes if need be, since if we pass bytes to the Python APIs, it may do implicit decoding based on the character map. There is no real bytes encoding of a Windows environment variable, unless we (or Python) encodes it.

The solution is basically not to change any of the code, but to ensure that these tests run under an accurate code page. The Unicode tests fail because of the Latin-1 code page, and so them failing is "correct". The issue is that we're dealing with Unicode, so we want a working Unicode input/output system.

azure-pipelines/windows_test_jobs.yml
rodrigc: @adiroiban Where in the public documentation?

src/twisted/test/test_process.py
rodrigc: This is documented in the Python 3 docs: https://docs.python.org/3/library/os.html#os.environ  `os.environ` contains `str` keys and values.  `str` is Unicode on Python 3.

azure-pipelines/windows_test_jobs.yml
adiroiban: I would say that it should be somewhere in the installation part 
 https://twistedmatrix.com/documents/current/installation/index.html

Maybe create a new page to document the Windows runtime environment  requirements... as well as the async io event loop change that is required on Windows.

src/twisted/test/test_process.py
rodrigc: OK I did this

azure-pipelines/windows_test_jobs.yml
rodrigc: OK, I added a comment.

src/twisted/test/test_process.py
rodrigc: OK I added this doc string

src/twisted/test/test_process.py
rodrigc: Done

rodrigc:
> I see that you made a change to posix_reactor, but I don't see any updated tests for Unicode env vars.
> Does it mean that posix Unicode env vars were already working, and we already have a test for that.
> In that case, I think that those existing tests should also be enabled for Windows.

On Python 3, `os.environ` is `str` on all platforms, including POSIX.  In Python 3, `str` is Unicode, including on POSIX.

rodrigc:
> That is, extend Win32ProcessTests to have an test_osEnviron_bytes and test_osEnviron_unicode , similar to test_stdinReader_bytesArgs/test_stdinReader_unicodeArgs
> 
> Let me know if you need help with the tests and I suggest a path for such a test.


@adiroiban I don't understand what you are asking for here.  On Windows, `os.environ` is Unicode.
On Windows, there is no `os.environb`because `os.supports_bytes_environ` is always False.
See:
https://docs.python.org/3/library/os.html#os.environ
https://docs.python.org/3/library/os.html#os.environb
https://docs.python.org/3/library/os.html#os.supports_bytes_environ

rodrigc:
> I think changing the Reactor portion is wrong -- Windows environment variables are stored internally as Unicode, and the APIs take Unicode. Therefore, we should ask users to provide the native (Unicode) representation in the API, decoding it ourselves from bytes if need be, since if we pass bytes to the Python APIs, it may do implicit decoding based on the character map. There is no real bytes encoding of a Windows environment variable, unless we (or Python) encodes it.
> 
> The solution is basically not to change any of the code, but to ensure that these tests run under an accurate code page. The Unicode tests fail because of the Latin-1 code page, and so them failing is "correct". The issue is that we're dealing with Unicode, so we want a working Unicode input/output system.

@hawkowl OK, I took out the change which eliminated `argChecker()` from the reactor.
I made one slight change to `twisted/internet/base.py` to put the actual bad key or value in the TypeError message,
to make it a bit easier to debug when something goes wrong.

src/twisted/test/test_process.py
```diff
@@ -496,18 +496,18 @@ def parseChunks(self, chunks):
         strings giving key value pairs of the environment from that process.
         Return this as a dictionary.
         """
-        environString = b''.join(chunks)
-        if not environString:
+        environBytes = b''.join(chunks)
```
glyph, line 499: 👍 love the more descriptive names

glyph (approved):
The narrower change here LGTM. Let’s land this so we can get non-ASCII contributors in here ASAP :)

rodrigc:
@mgorny Thanks a lot to Michał Górny who was patient in helping me track down the problem and come up with a fix.  I found it very weird that if I did a PR from the exact same branch as @mgorny, my branch would pass CI, but if @mgorny did another PR from the same branch, the CI on his PR would fail.

It took me a long time to track it down to an environment variable:

```
BUILD_SOURCEVERSIONAUTHOR Michał Górny
```

which had non-ASCII characters in the value, and thus triggered CI failures on Windows. :)

PR #1306, opened 2020-06-21
[mypy] [9861] Tell mypy to ignore classes which derive from dynamic base classes

rodrigc:
https://twistedmatrix.com/trac/ticket/9861

mypy complains if it cannot figure what a base class is.  For example:

```
src/twisted/_threads/test/test_team.py:17:24: error: Unsupported dynamic base class "proxyForInterface"  [misc]
```

This PR quietens those errors.

### DIFF
--- a/src/twisted/_threads/test/test_team.py
+++ b/src/twisted/_threads/test/test_team.py
@@ -14,7 +14,9 @@ from twisted.python.components import proxyForInterface
 from twisted.python.failure import Failure
 from .. import IWorker, Team, createMemoryWorker, AlreadyQuit
 
-class ContextualWorker(proxyForInterface(IWorker, "_realWorker")):
+
+
+class ContextualWorker(proxyForInterface(IWorker, "_realWorker")):  # type: ignore[misc]  # noqa
     """
     A worker implementation that supplies a context.
     """
--- a/src/twisted/internet/cfreactor.py
+++ b/src/twisted/internet/cfreactor.py
@@ -44,6 +44,7 @@ _WRITE = 1
 _preserveSOError = 1 << 6
 
 
+
 class _WakerPlus(_Waker):
     """
     The normal Twisted waker will simply wake up the main loop, which causes an
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -315,7 +315,7 @@ class _IProcessTransportWithConsumerAndProducer(interfaces.IProcessTransport,
 
 
 class _ProcessEndpointTransport(
-        proxyForInterface(_IProcessTransportWithConsumerAndProducer,
+        proxyForInterface(_IProcessTransportWithConsumerAndProducer,  # type: ignore[misc]  # noqa
                           '_process')):
     """
     An L{ITransport}, L{IProcessTransport}, L{IConsumer}, and L{IPushProducer}
--- a/src/twisted/web/client.py
+++ b/src/twisted/web/client.py
@@ -1959,7 +1959,7 @@ class CookieAgent(object):
 
 
 
-class GzipDecoder(proxyForInterface(IResponse)):
+class GzipDecoder(proxyForInterface(IResponse)):  # type: ignore[misc]
     """
     A wrapper for a L{Response} instance which handles gzip'ed body.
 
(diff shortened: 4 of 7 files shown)


src/twisted/internet/cfreactor.py
```diff
@@ -44,7 +44,8 @@
 _preserveSOError = 1 << 6
 
 
-class _WakerPlus(_Waker):
+
+class _WakerPlus(_Waker):  # type: ignore[valid-type,misc]
```
adiroiban: I don't understand this.

This looks like a normal _Waker type to me.

src/twisted/_threads/test/test_team.py
```diff
@@ -14,7 +14,9 @@
 from twisted.python.failure import Failure
 from .. import IWorker, Team, createMemoryWorker, AlreadyQuit
 
-class ContextualWorker(proxyForInterface(IWorker, "_realWorker")):
+
+
+class ContextualWorker(proxyForInterface(IWorker, "_realWorker")):  # type: ignore[misc]  # noqa
```
adiroiban, line 19: I don't know what to say.

Can't this be defined as the IWorker interface type ?

It is a proxy for that interface, so it should implement that interface.

adiroiban (commented):
Thanks for the update.

For now, just a comment.

I was expecting that proxyForInterface will generate an instance which provides that interface

src/twisted/internet/cfreactor.py
rodrigc: @adiroiban `_Waker` can be defined one of two ways here: https://github.com/twisted/twisted/blob/04ece218a9f7bbeaa1faa37346317abbf37c5bd7/src/twisted/internet/posixbase.py#L192
as `_UnixWaker` or `_SocketWaker`.

Since these are two different types, this confuses the static type checking in mypy, so we need to add the comment to tell mypy to ignore this line.

src/twisted/_threads/test/test_team.py
rodrigc: @adiroiban This code was added by @glyph in https://github.com/twisted/twisted/commit/b572e7e1ffa299f9751721405d5a5536995df85e#diff-cd7281783ad732cac003024589861371R18

Without this comment, mypy complains with:

```
src/twisted/_threads/test/test_team.py:17:24: error: Unsupported dynamic base class "proxyForInterface"  [misc]
    class ContextualWorker(proxyForInterface(IWorker, "_realWorker")):
```

Since `proxyForInterface` dynamically creates a class, this comment needs to be added to tell mypy to ignore this line, since mypy can only statically typecheck the class.

glyph:
We should, at some point, write and then ship a mypy plug-in that deals with this and eliminates the need for the `type: ignore`. These often turn out to be tractably small. But obviously let’s unblock for now.

src/twisted/_threads/test/test_team.py
adiroiban: Ok. In this case is ok to ignore all the proxyForInterface usage.

I guess that instead of creating a twisted pluging, is best if zope.interface could implement an `@delegator(ISomeThing, '_conextName')` decorator which will do the same thing as proxyForInterface.

I was using https://launchpad.net/lazr.delegates to get something similar but I think that this is something that has more to do with zope interface than zope components.

src/twisted/internet/cfreactor.py
adiroiban: ok... I still don't think that this is the right way to fix it

I think that rather than using an Union, there should be a _IWaker interface implemented by all wakers and then just use the IWaker interface to refer to the type of any waker.

I think that Union should be used for the case when there are implementation with different interfaces. But all wakers have the same interface.

@glyph  can you please check if the IWaker interface from below is correct?

```
class _IWaker(Inteface):
    """
    Interface to wake up the event loop based on the self-pipe trick.

    The I{self-pipe trick<http://cr.yp.to/docs/selfpipe.html>}, used to wake
    up the main loop from another thread or a signal handler.
    This is why we have wakeUp together with doRead

    This is used by threads or signals to wake up the event loop.
    """
    disconnected = Attribute('')

    def wakeUp():
        """Called when the event should be wake up.
        """

    def doRead():
        """Read some data from my connection and discard it.
        """

    def connectionLost(reason):
        """Called when connection was closed and the pipes.
        """
```

adiroiban (changes requested):
thanks. Please consider using a common IWaker interface instead of the Union type definition.

src/twisted/internet/cfreactor.py
rodrigc: I've moved `_Waker` changes to https://github.com/twisted/twisted/pull/1329

adiroiban (approved):
All good. Many thanks for all your work!

PR #1307, opened 2020-06-21
[mypy] suppress errors after ImportError

rodrigc:
https://twistedmatrix.com/trac/ticket/9862

In cases where an import can fail, such as:

```
try:
    from twisted.internet.stdio import StandardIO, PipeAddress
except ImportError:
    # fallback if pywin32 is not installed
    StandardIO = None
    PipeAddress = None
```

mypy will fail with:

```
src/twisted/internet/endpoints.py:45:5: error: Cannot assign multiple types to name "StandardIO" without an explicit "Type[...]" annotation  [misc]
src/twisted/internet/endpoints.py:45:18: error: Incompatible types in assignment (expression has type "None", variable has type "Type[StandardIO]")  [assignment]
```

This PR cleans up those errors.

### DIFF
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -42,8 +42,8 @@ try:
     from twisted.internet.stdio import StandardIO, PipeAddress
 except ImportError:
     # fallback if pywin32 is not installed
-    StandardIO = None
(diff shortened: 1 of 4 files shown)


adiroiban (approved):
All good. thanks!

PR #1308, opened 2020-06-21
[mypy] Rename some variables which collide with other variables

rodrigc:
https://twistedmatrix.com/trac/ticket/9863

In src/twisted/conch/scripts/conch.py , this line:

```
# Rest of code in "run"
options = None
```

causes mypy to complain with:

```
src/twisted/conch/scripts/conch.py:105:11: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
```

This PR cleans up a few errors similar to that.

### DIFF
--- a/src/twisted/test/stdio_test_halfclose.py
+++ b/src/twisted/test/stdio_test_halfclose.py
@@ -65,7 +65,7 @@ if __name__ == '__main__':
     reflect.namedAny(sys.argv[1]).install()
     log.startLogging(open(sys.argv[2], 'wb'))
     from twisted.internet import reactor
-    protocol = HalfCloseProtocol()
-    stdio.StandardIO(protocol)
+    halfCloseProtocol = HalfCloseProtocol()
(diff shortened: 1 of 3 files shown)


src/twisted/test/stdio_test_halfclose.py
```diff
@@ -65,7 +65,7 @@ def writeConnectionLost(self, reason):
     reflect.namedAny(sys.argv[1]).install()
     log.startLogging(open(sys.argv[2], 'wb'))
     from twisted.internet import reactor
-    protocol = HalfCloseProtocol()
-    stdio.StandardIO(protocol)
+    protocolHalfCloseProtocol = HalfCloseProtocol()
```
adiroiban: maybe just 
```suggestion
    halfCloseProtocol = HalfCloseProtocol()
```

adiroiban (approved):
looks good. thanks!

PR #1309, opened 2020-06-21
[mypy] Change type of variable to bool to match base class

rodrigc:
https://twistedmatrix.com/trac/ticket/9864

In a few places there are errors like this:

```
src/twisted/conch/scripts/tkconch.py:353:13: error: Incompatible types in assignment (expression has type "int", base class "Factory" defined the type as "bool")  [assignment]
```

Where in a subclass, a variable assignment does not match the type of the variable specified in the base class.

This PR cleans some of those up.


adiroiban (approved):
All good.Thanks!

PR #1311, opened 2020-06-21
[mypy] Change delimiter to bytes in irc.py

rodrigc:
https://twistedmatrix.com/trac/ticket/9866

This PR cleans up this error:

```
src/twisted/words/protocols/irc.py:2920:17: error: Incompatible types in assignment (expression has type "str", base class "LineReceiver" defined the type as "bytes")  [assignment]
```


adiroiban (approved):
Looks good. Thanks.

PR #1313, opened 2020-06-22

rodrigc:
https://twistedmatrix.com/trac/ticket/9867

Add some variable annotations to reduce mypy warnings

### DIFF
--- a/src/twisted/mail/protocols.py
+++ b/src/twisted/mail/protocols.py
@@ -37,7 +37,7 @@ class DomainDeliveryBase:
         Sub-classes should set this appropriately.
     """
     service = None
-    protocolName = None
+    protocolName = b'not-implemented-protocol'  # type: bytes
 
     def __init__(self, service, user, host=smtp.DNSNAME):
         """
--- a/src/twisted/mail/test/test_smtp.py
+++ b/src/twisted/mail/test/test_smtp.py
@@ -11,6 +11,7 @@ import base64
 
 from io import BytesIO
 from unittest import skipIf
+from typing import Optional, Type
 
 from zope.interface import implementer, directlyProvides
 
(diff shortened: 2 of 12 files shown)


src/twisted/mail/test/test_smtp.py
```diff
@@ -384,8 +385,8 @@ class DummyESMTP(DummyProto, smtp.ESMTP):
 
 
 class AnotherTestCase:
-    serverClass = None
-    clientClass = None
+    serverClass = None  # type: Optional[Type[DummyProto]]
```
adiroiban: I think that this should be a generic protocol. and not the specific DummyProto

```suggestion
    serverClass = None  # type: Optional[Type[protocol.Protocol]]
```

src/twisted/mail/test/test_smtp.py
```diff
@@ -384,8 +385,8 @@ class DummyESMTP(DummyProto, smtp.ESMTP):
...
+    clientClass = None  # type: Optional[Type[MyClient]]
```
adiroiban: Maybe 
```suggestion
    clientClass = None  # type: Optional[Type[smtp.SMTPClient]]
```

src/twisted/mail/protocols.py
```diff
@@ -37,7 +38,7 @@ class DomainDeliveryBase:
         Sub-classes should set this appropriately.
     """
     service = None
-    protocolName = None
+    protocolName = None  # type: Optional[bytes]
```
adiroiban: I don't think that this is optional. protocolName should always be defined, as described by the docstring.

Maybe have something like this.
```suggestion
    protocolName = b'not-implemented-protocol`
```

adiroiban (approved):
Only minor comments. Thanks!

src/twisted/mail/test/test_smtp.py
rodrigc: Done

src/twisted/mail/protocols.py
rodrigc: Done

PR #1314, opened 2020-06-22

rodrigc:
https://twistedmatrix.com/trac/ticket/9869

This adds type annotations in a few places to reduce mypy errors.


adiroiban (approved):
It looks good. thanks.

Ii am not sure what is the scope / or method for making these changes.

It would help to have for example all `except ImportError` handling in a single PR

PR #1315, opened 2020-06-22

rodrigc:
https://twistedmatrix.com/trac/ticket/9870

This adds some annotations to reduce mypy errors.

### DIFF
--- a/src/twisted/test/test_process.py
+++ b/src/twisted/test/test_process.py
@@ -33,7 +33,7 @@ from unittest import skipIf
 try:
     import fcntl
 except ImportError:
-    fcntl = None
+    fcntl = None  # type: ignore[assignment]
     process = None
 else:
     from twisted.internet import process
@@ -389,7 +389,7 @@ class UtilityProcessProtocol(protocol.ProcessProtocol):
 
     @ivar programName: The name of the program to run.
     """
-    programName = None
+    programName = b""  # type: bytes
 
     @classmethod
     def run(cls, reactor, argv, env):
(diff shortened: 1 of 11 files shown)


src/twisted/test/test_process.py
```diff
@@ -389,7 +390,7 @@ class UtilityProcessProtocol(protocol.ProcessProtocol):
 
     @ivar programName: The name of the program to run.
     """
-    programName = None
+    programName = None  # type: Optional[bytes]
```
adiroiban: since this is testing code, it can be initialized with a default value.

I don't think that having programName as None is a valid value as the code should always fail in that case.

```suggestion
    programName = b'define-your-own-name'
```

src/twisted/web/http.py
```diff
@@ -733,8 +733,8 @@ class Request:
     finished = 0
     code = OK
     code_message = RESPONSES[OK]
-    method = "(no method yet)"
-    clientproto = b"(no clientproto yet)"
+    method = b"(no method yet)"  # type: bytes
```
adiroiban, line 736: why is this required? should this be auto detected from the normal assignment ?
Is this a mypy bug?

adiroiban (approved):
Looks good. Only minor comments. Thanks

src/twisted/web/http.py
rodrigc: All subclasses define method as bytes, which is correct.  It is incorrectly defined as str here.

For example, elsewhere in this file:

```
        if self.method == b"POST" and ctype and clength:
            mfd = b'multipart/form-data'
            key, pdict = _parseHeader(ctype)
```

src/twisted/test/test_process.py
rodrigc: Done

PR #1316, opened 2020-06-22
[mypy] [9871] Add some variable annotations

rodrigc:
https://twistedmatrix.com/trac/ticket/9871

This adds annotations to some variables to reduce mypy errors.

### DIFF
--- a/src/twisted/web/test/test_http.py
+++ b/src/twisted/web/test/test_http.py
@@ -14,6 +14,7 @@ import hamcrest
 
 from urllib.parse import urlparse, urlunsplit, clear_cache, parse_qs
 from unittest import skipIf
+from typing import Sequence, Union
 
 from io import BytesIO
 from itertools import cycle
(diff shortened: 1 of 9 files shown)


src/twisted/web/test/test_http.py
```diff
@@ -270,7 +271,7 @@ class HTTP1_0Tests(unittest.TestCase, ResponseTestMixin):
          b"Command: GET",
          b"Version: HTTP/1.0",
          b"Content-Length: 13",
-         b"'''\nNone\n'''\n")]
+         b"'''\nNone\n'''\n")]  # type: Union[Sequence[Sequence[bytes]], bytes]
```
adiroiban, line 274: maybe the tests should be updated to always have Sequence[Sequence[bytes]]

adiroiban (approved):
Looks good. thanks. Only minor comments.

PR #1317, opened 2020-06-22
[mypy] Add lint and mypy instructions to README.rst

rodrigc:
https://twistedmatrix.com/trac/ticket/9872

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ For information on changes in this release, see the `NEWS <NEWS.rst>`_ file.
 What is this?
 -------------
 
-Twisted is an event-based framework for internet applications, supporting Python 2.7 and Python 3.5+.
+Twisted is an event-based framework for internet applications, supporting Python 3.5+.
 It includes modules for many different purposes, including the following:
 
 - ``twisted.web``: HTTP clients and servers, HTML templating, and a WSGI server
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -52,15 +52,15 @@ Unit Tests
 
 Twisted has a comprehensive test suite, which can be run by ``tox``::
 
-  $ tox -l            # to view all test environments
-  $ tox -e py27-tests # to run the tests for Python 2.7
-  $ tox -e py35-tests # to run the tests for Python 3.5
+  $ tox -l                       # to view all test environments
+  $ tox -e py35-tests            # to run the tests for Python 3.5
+  $ tox -e py36-tests            # to run the tests for Python 3.6
```
adiroiban: I think that a single py3 is enough ... and maybe suggest python3.8 as default.

adiroiban (approved):
Thanks.

I think that adding this will not hurt.... but my suggestion is to move this info into the CONTRIBUTING file.

The CONTRIBUTING is dedicated to twisted developers, while README is more for users and a general introduction to the repo. 

I am confused by all the Twisted DEV documentation.

Beside CONTRIBUTING from the repo we have this official documentation page

https://twistedmatrix.com/documents/current/core/development/policy/code-dev.html#running-tests

We also have a wiki page https://twistedmatrix.com/trac/wiki/TwistedDevelopment

--------

I prefer having development documentation inside the repo like CONTRIBUTING and .github/PULL_REQUEST_TEMPLATE as in this way you can update the documentation at the same time with the code changes.

For example, when support for Python2.7 was removed you can run a search for all the repo for py2.7 or `2.7` and this search should also hit the dev documentation.

README.rst
rodrigc: Actually, it looks like `tests` was removed from tox in https://github.com/twisted/twisted/commit/fcd6ea08f4043996f5d886625da7a2996dda3372

I changed to:

```
+  $ tox -l                       # to view all test environments
+  $ tox -e nocov                 # to run all tests without coverage
+  $ tox -e withcov               # to run all tests for with coverage
+  $ tox -e alldeps-withcov-posix # install all dependencies, run tests on POSIX platform
```

PR #1319, opened 2020-06-26
9874 mthuurne missing docstring field names

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9874
* [x] ~~I have updated the automated tests.~~ Only docstrings were changed.

### DIFF
--- a/src/twisted/names/secondary.py
+++ b/src/twisted/names/secondary.py
@@ -107,10 +107,10 @@ class SecondaryAuthority(FileAuthority):
 
     @ivar _port: The port number of the server from which zone transfers will
     be attempted.
-    @type: C{int}
+    @type _port: C{int}
 
     @ivar domain: The domain for which this is the secondary authority.
-    @type: C{bytes}
+    @type domain: C{bytes}
 
     @ivar _reactor: The reactor to use to perform the zone transfers, or
     L{None} to use the global reactor.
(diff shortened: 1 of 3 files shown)


src/twisted/names/secondary.py
```diff
@@ -107,10 +107,10 @@ class SecondaryAuthority(FileAuthority):
 
     @ivar _port: The port number of the server from which zone transfers will
     be attempted.
-    @type: C{int}
+    @type _port: C{int}
 
     @ivar domain: The domain for which this is the secondary authority.
-    @type: C{bytes}
```
rodrigc, line 113: From the https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html#docstrings

```
When you are referring to a type, you should use L{}, whether it’s in the stdlib , in Twisted or somewhere else.
```

So this needs to be `L{bytes}`.

rodrigc (approved):
This looks OK, just make that one change.

In future, please consider adding type annotations for the functions in which you are fixing the
doc strings.  This will help the effort in adding type annotations to Twisted.

src/twisted/names/secondary.py
mthuurne: I'll fix it.

There are hundreds of other cases of `C{typename}` in Twisted's sources though, so maybe it's worth doing a mechanical replace over the whole source tree at some point. Or maybe we can skip that and instead do an automated conversion from `@type` to type annotation/comment.

PR #1322, opened 2020-06-29
[mypy] [9877] Reduce mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9877

### DIFF
--- a/src/twisted/pair/ethernet.py
+++ b/src/twisted/pair/ethernet.py
@@ -15,13 +15,17 @@ from twisted.pair import raw
 from zope.interface import implementer, Interface
 
 
+
 class IEthernetProtocol(Interface):
     """An interface for protocols that handle Ethernet frames"""
-    def addProto():
+    def addProto(num, proto):
         """Add an IRawPacketProtocol protocol"""
 
-    def datagramReceived():
+
+    def datagramReceived(data, partial):
         """An Ethernet frame has been received"""
+
+
 
 class EthernetHeader:
     def __init__(self, data):
@@ -36,6 +40,7 @@ class EthernetProtocol(protocol.AbstractDatagramProtocol):
     def __init__(self):
         self.etherProtos = {}
 
+
     def addProto(self, num, proto):
         proto = raw.IRawPacketProtocol(proto)
         if num < 0:
@@ -45,6 +50,7 @@ class EthernetProtocol(protocol.AbstractDatagramProtocol):
         if num not in self.etherProtos:
             self.etherProtos[num] = []
         self.etherProtos[num].append(proto)
+
 
     def datagramReceived(self, data, partial=0):
         header = EthernetHeader(data[:14])
--- a/src/twisted/conch/test/test_telnet.py
+++ b/src/twisted/conch/test/test_telnet.py
@@ -84,12 +84,12 @@ class TestProtocol:
         pass
 
 
-    def unhandledCommand(self):
+    def unhandledCommand(self, command, argument):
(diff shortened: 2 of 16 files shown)


src/twisted/internet/base.py
```diff
@@ -719,7 +719,9 @@ def fireSystemEvent(self, eventType):
             event.fireEvent()
 
 
-    def addSystemEventTrigger(self, _phase, _eventType, _f, *args, **kw):
+    # We need to ignore mypy errors here because the '_f' parameter is
+    # not in IReactorCore.addSystemEventTrigger
+    def addSystemEventTrigger(self, _phase, _eventType, _f, *args, **kw):  # type: ignore[override]  # noqa
```
adiroiban: I think that this should follow the signature from the interface: `def addSystemEventTrigger(phase, eventType, callable, *args, **kw):` and rename `_phase` -> `phase` and `_f` -> `callable`

src/twisted/internet/base.py
```diff
@@ -785,7 +787,9 @@ def run(self):
 
     seconds = staticmethod(runtimeSeconds)
 
-    def callLater(self, _seconds, _f, *args, **kw):
+    # We need to ignore mypy errors here because the '_f' parameter is
+    # not in IReactorTime.callLater
+    def callLater(self, _seconds, _f, *args, **kw):  # type: ignore[override]
```
adiroiban: just use the signature from the interface.

src/twisted/pair/ethernet.py
```diff
@@ -15,14 +15,18 @@
 from zope.interface import implementer, Interface
 
 
+
 class IEthernetProtocol(Interface):
     """An interface for protocols that handle Ethernet frames"""
-    def addProto():
+    def addProto(num, proto):
```
adiroiban, line 21: this interface change looks good can can be done here... but maybe use a `.remove` release note to let other know that the IEthernetProtocol interface was updated to match all the existing implementations.

adiroiban (changes requested):
It looks good.

For addSystemEventTrigger and callLater I think that the implementation should just use the same argument names as defined in the interface.
I have no idea why *private* arguments were used.

Thanks again.

src/twisted/internet/base.py
rodrigc: How?  The signature from the interface takes 4 args, but this function signature takes 5 args.

adiroiban: Which is the 5th argument?

This change would look valid to me:

```suggestion
    def callLater(self, delay, callable, *args, **kw):]
```

This is based on

```
src/twisted/internet/interfaces.py:    def callLater(delay, callable, *args, **kw):
```

src/twisted/internet/base.py
rodrigc: I've moved this to: https://github.com/twisted/twisted/pull/1325

src/twisted/internet/base.py
rodrigc: I've moved this to https://github.com/twisted/twisted/pull/1325

rodrigc: I've moved this to https://github.com/twisted/twisted/pull/1325

src/twisted/pair/ethernet.py
rodrigc: I added newsfragment

adiroiban (approved):
Looks good. Thanks!

I think that  src/twisted/newsfragments/9877.misc file can be removed.

PR #1324, opened 2020-06-29
[mypy] [9879] Reduce mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9879


adiroiban (approved):
Thanks for the cleanup. All good.

PR #1325, opened 2020-06-29
[mypy] [9880] Reduce mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9880

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -8,7 +8,7 @@ Very basic functionality for a Reactor implementation.
 
 
 import socket  # needed only for sync-dns
-from typing import List
+from typing import Any, Callable, List
 from zope.interface import implementer, classImplements
 
 import sys
@@ -720,7 +720,8 @@ class ReactorBase(PluggableResolverMixin):
             event.fireEvent()
 
 
-    def addSystemEventTrigger(self, phase, eventType, callable, *args, **kw):
+    def addSystemEventTrigger(self, phase: str, eventType: str,
+                              callable: Callable[..., Any], *args, **kw):
         """See twisted.internet.interfaces.IReactorCore.addSystemEventTrigger.
         """
         assert builtins.callable(callable), \
@@ -738,7 +739,7 @@ class ReactorBase(PluggableResolverMixin):
         self._eventTriggers[eventType].removeTrigger(handle)
 
 
-    def callWhenRunning(self, callable, *args, **kw):
+    def callWhenRunning(self, callable: Callable[..., Any], *args, **kw):
         """See twisted.internet.interfaces.IReactorCore.callWhenRunning.
         """
         if self.running:
@@ -746,6 +747,7 @@ class ReactorBase(PluggableResolverMixin):
         else:
             return self.addSystemEventTrigger('after', 'startup',
                                               callable, *args, **kw)
+
 
     def startRunning(self):
         """
@@ -787,7 +789,7 @@ class ReactorBase(PluggableResolverMixin):
 
     seconds = staticmethod(runtimeSeconds)
 
-    def callLater(self, delay, callable, *args, **kw):
+    def callLater(self, delay, callable: Callable[..., Any], *args, **kw):
         """See twisted.internet.interfaces.IReactorTime.callLater.
         """
         assert builtins.callable(callable), \
@@ -800,6 +802,7 @@ class ReactorBase(PluggableResolverMixin):
                            seconds=self.seconds)
         self._newTimedCalls.append(tple)
         return tple
+
 
     def _moveCallLaterSooner(self, tple):
         # Linear time find: slow.
--- a/src/twisted/internet/interfaces.py
+++ b/src/twisted/internet/interfaces.py
@@ -8,8 +8,10 @@ Maintainer: Itamar Shtull-Trauring
 """
 
 
+from typing import Any, Callable, Sequence, Union
 from twisted.python.failure import Failure
 from zope.interface import Interface, Attribute
+
 
 
 class IAddress(Interface):
@@ -56,7 +58,7 @@ class IConnector(Interface):
 
 
 class IResolverSimple(Interface):
-    def getHostByName(name, timeout):
+    def getHostByName(name: Union[bytes, str], timeout: Sequence[int]):
         """
         Resolve the domain name C{name} into an IP address.
 
@@ -150,8 +152,9 @@ class IHostnameResolver(Interface):
     @since: Twisted 17.1.0
     """
 
-    def resolveHostName(resolutionReceiver, hostName, portNumber=0,
-                        addressTypes=None, transportSemantics='TCP'):
+    def resolveHostName(resolutionReceiver: IResolutionReceiver, hostName: str,
+                        portNumber: int, addressTypes: Sequence[IAddress],
+                        transportSemantics: str):
         """
         Initiate a hostname resolution.
 
@@ -189,7 +192,7 @@ class IHostnameResolver(Interface):
 
 
 class IResolver(IResolverSimple):
-    def query(query, timeout=None):
+    def query(query, timeout):
         """
         Dispatch C{query} to the method which can handle its type.
 
@@ -212,7 +215,7 @@ class IResolver(IResolverSimple):
         """
 
 
-    def lookupAddress(name, timeout=None):
+    def lookupAddress(name, timeout):
         """
         Perform an A record lookup.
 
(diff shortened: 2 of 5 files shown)


src/twisted/internet/interfaces.py
```diff
@@ -1243,14 +1250,14 @@ def seconds():
         """
 
 
-    def callLater(delay, callable, *args, **kw):
+    def callLater(delay, f: Callable[..., Any], *args, **kw):
```
adiroiban: I don't know what to say about this interface change.

@glyph  do you have any suggestion?

I would way that this change should be made in a separate PR.

I don't know if `f` is the best name ... maybe use `func`

```
$ grep -r "def callLater(" src/
src/twisted/protocols/policies.py:    def callLater(self, period, func):
src/twisted/protocols/policies.py:    def callLater(self, period, func):
src/twisted/protocols/policies.py:    def callLater(self, period, func):
src/twisted/test/test_policies.py:    def callLater(self, period, func):
src/twisted/test/test_policies.py:    def callLater(self, period, func):
src/twisted/test/test_policies.py:    def callLater(self, timeout, func, *args, **kwargs):
src/twisted/names/dns.py:    def callLater(self, period, func, *args):
src/twisted/conch/test/test_helper.py:    def callLater(self, timeout, f, *a, **kw):
src/twisted/web/http.py:    def callLater(self):
src/twisted/web/http.py:    def callLater(self, value):
src/twisted/internet/_threadedselect.py:    def callLater(self, *args, **kw):
src/twisted/internet/interfaces.py:    def callLater(delay, callable, *args, **kw):
src/twisted/internet/asyncioreactor.py:    def callLater(self, seconds, f, *args, **kwargs):
src/twisted/internet/cfreactor.py:    def callLater(self, _seconds, _f, *args, **kw):
src/twisted/internet/task.py:    def callLater(self, when, what, *a, **kw):
src/twisted/internet/base.py:    def callLater(self, _seconds, _f, *args, **kw):
src/twisted/internet/_glibbase.py:    def callLater(self, *args, **kwargs):
```

but is best to keep the scope of this PR to only type hinting and make the backward incompatible changes in a separate PR.

adiroiban (changes requested):
Thanks. This looks good.

My request it to remove the backward incompatible changes from this PR and have them in a separate PR dedicated to changing the signature of the interfaces.

glyph:
> My request it to remove the backward incompatible changes from this PR and have them in a separate PR dedicated to changing the signature of the interfaces.

FWIW: very much agreed on this point :)

rodrigc:
> > My request it to remove the backward incompatible changes from this PR and have them in a separate PR dedicated to changing the signature of the interfaces.
> 
> FWIW: very much agreed on this point :)


@glyph @adiroiban OK I submitted a separate PR for this: https://github.com/twisted/twisted/pull/1341

src/twisted/internet/interfaces.py
```diff
@@ -1340,7 +1347,7 @@ class IReactorFromThreads(Interface):
     @since: 15.4
     """
 
-    def callFromThread(callable, *args, **kw):
+    def callFromThread(f: Callable[..., Any], *args, **kw):
```
adiroiban: This is still an interface change callable -> f

src/twisted/internet/interfaces.py
```diff
@@ -1364,7 +1371,7 @@ class IReactorInThreads(Interface):
     @since: 15.4
     """
 
-    def callInThread(callable, *args, **kwargs):
+    def callInThread(f: Callable[..., Any], *args, **kwargs):
```
adiroiban: This is still an interface change callable -> f

src/twisted/internet/interfaces.py
```diff
@@ -1508,15 +1517,15 @@ def removeSystemEventTrigger(triggerID):
             C{triggerID}.
         """
 
-    def callWhenRunning(callable, *args, **kw):
+    def callWhenRunning(f: Callable[..., Any], *args, **kw):
```
adiroiban: This is still an interface change callable -> f

src/twisted/internet/task.py
```diff
@@ -786,15 +787,15 @@ def _sortCalls(self):
         self.calls.sort(key=lambda a: a.getTime())
 
 
-    def callLater(self, when, what, *a, **kw):
+    def callLater(self, delay, f: Callable[..., Any], *args, **kw):
```
adiroiban: should keep `when` and `what` as they are.

The code from below is valid code, but will break after this change

```python
reactor.callLater(when=0.3, what=recover_response)
```

src/twisted/internet/interfaces.py
adiroiban: This is still an interface change callable -> f

src/twisted/internet/base.py
```diff
@@ -719,14 +719,15 @@ def fireSystemEvent(self, eventType):
             event.fireEvent()
 
 
-    def addSystemEventTrigger(self, _phase, _eventType, _f, *args, **kw):
+    def addSystemEventTrigger(self, phase: str, eventType: str,
+                              f: Callable[..., Any], *args, **kw):
```
adiroiban: This is still a signature change _f -> f 

But since _f is a "private" we can change it to public... but if it should be change, it should be change to `callable` to match the interface definition.

The interface is defined as
```
src/twisted/internet/interfaces.py:    def addSystemEventTrigger(phase, eventType, callable, *args, **kw):
```

src/twisted/internet/base.py
```diff
@@ -736,14 +737,14 @@ def removeSystemEventTrigger(self, triggerID):
         self._eventTriggers[eventType].removeTrigger(handle)
 
 
-    def callWhenRunning(self, _callable, *args, **kw):
+    def callWhenRunning(self, f: Callable[..., Any], *args, **kw):
```
adiroiban: this is a signature change same as addSystemEventTrigger ... this should be named `callable`

src/twisted/internet/test/test_base.py
```diff
@@ -71,7 +72,7 @@ def seconds(self):
         pass
 
 
-    def callInThread(self, callable, *args, **kwargs):
+    def callInThread(self, f: Callable[..., Any], *args, **kwargs):
```
adiroiban: this is a signature change.

src/twisted/internet/testing.py
```diff
@@ -577,15 +578,16 @@ def fireSystemEvent(self, eventType):
         raise NotImplementedError()
 
 
-    def addSystemEventTrigger(self, phase, eventType, callable, *args, **kw):
+    def addSystemEventTrigger(self, phase: str, eventType: str,
+                              f: Callable[..., Any], *args, **kw):
```
adiroiban: I guess that for the testing code this is ok, but is better to have this in the separate PR.

adiroiban (changes requested):
Thanks for the update.
I still see changes to the interface and method signature.

src/twisted/internet/test/test_base.py
rodrigc: I took this signature change out

src/twisted/internet/base.py
rodrigc: I took this signature change out

src/twisted/internet/base.py
rodrigc: I took this signature change out

src/twisted/internet/task.py
rodrigc: I took this out

src/twisted/internet/interfaces.py
rodrigc: I took this out

src/twisted/internet/interfaces.py
rodrigc: I took this out

src/twisted/internet/interfaces.py
rodrigc: I took this out

src/twisted/internet/testing.py
rodrigc: I took this out

rodrigc:
@adiroiban I changed this PR to just focus on removing default arguments from interfaces, and adding type annotations to some method arguments

adiroiban (approved):
Wow. Great work. Thanks.

I see that mypy is not complaining about the callable argument.

I have create the following tests and mypy is ok with the file

```python
from zope.interface import Interface, Attribute, implementer

class ITest(Interface):
    test1 = Attribute('test1')
    def test2(callable, two):
        "Order or arguments."

@implementer(ITest)
class Test:
    def __init__(self):
        self.test1 = 1

    def test2(self, callable, one):
        """
        Mixed order.
        """
```

PR #1326, opened 2020-06-30
[mypy] [9881] Remove twisted.python.filepath.FilePath.statinfo

rodrigc:
https://twistedmatrix.com/trac/ticket/9881

Remove `twisted.python.filepath.FilePath.statinfo` which has been deprecated since Twisted 15.0.0

This eliminates some mypy errors as well.


adiroiban (approved):
All good. Thanks!

PR #1329, opened 2020-07-01
[mypy] [9884] Introduce an _IWaker interface, reduce mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9884

This cleans up these mypy errors:

```
src/twisted/internet/posixbase.py:195:5: error: Cannot assign multiple types to name "_Waker" without an explicit "Type[...]" annotation  [misc]
src/twisted/internet/posixbase.py:195:14: error: Incompatible types in assignment (expression has type "Type[_SocketWaker]", variable has type "Type[_UnixWaker]")  [assignment]

src/twisted/internet/cfreactor.py:47:18: error: Variable "twisted.internet.posixbase._Waker" is not valid as a type  [valid-type]
src/twisted/internet/cfreactor.py:47:18: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
src/twisted/internet/cfreactor.py:47:18: error: Invalid base class "_Waker"  [misc]
```

### DIFF
--- a/src/twisted/internet/posixbase.py
+++ b/src/twisted/internet/posixbase.py
@@ -12,9 +12,9 @@ import errno
 import os
 import sys
 
-from typing import Union, Sequence, Type
+from typing import Sequence
 
-from zope.interface import implementer, classImplements
+from zope.interface import Attribute, Interface, implementer, classImplements
 
 from twisted.internet import error, udp, tcp
 from twisted.internet.base import ReactorBase, _SignalReactorMixin
@@ -63,6 +63,37 @@ if platform.isWindows():
         win32process = None
 
 
+
+class _IWaker(Interface):
+    """
+    Interface to wake up the event loop based on the self-pipe trick.
+
+    The U{I{self-pipe trick}<http://cr.yp.to/docs/selfpipe.html>}, used to wake
+    up the main loop from another thread or a signal handler.
+    This is why we have wakeUp together with doRead
+
+    This is used by threads or signals to wake up the event loop.
+    """
+    disconnected = Attribute('')
+
+    def wakeUp():
+        """
+        Called when the event should be wake up.
+        """
+
+    def doRead():
+        """
+        Read some data from my connection and discard it.
+        """
+
+    def connectionLost(reason: failure.Failure):
+        """
+        Called when connection was closed and the pipes.
+        """
+
+
+
+@implementer(_IWaker)
 class _SocketWaker(log.Logger):
     """
     The I{self-pipe trick<http://cr.yp.to/docs/selfpipe.html>}, implemented
@@ -99,13 +130,16 @@ class _SocketWaker(log.Logger):
             if e.args[0] != errno.WSAEWOULDBLOCK:
                 raise
 
+
     def doRead(self):
-        """Read some data from my connection.
+        """
+        Read some data from my connection.
         """
         try:
             self.r.recv(8192)
         except socket.error:
             pass
+
 
     def connectionLost(self, reason):
         self.r.close()
@@ -165,6 +199,7 @@ class _FDWaker(log.Logger, object):
 
 
 
+@implementer(_IWaker)
 class _UnixWaker(_FDWaker):
     """
     This class provides a simple interface to wake up the event loop.
@@ -189,10 +224,11 @@ class _UnixWaker(_FDWaker):
 
 
 if platformType == 'posix':
-    _Waker = _UnixWaker  # type: Union[Type[_FDWaker], Type[_SocketWaker]]
+    _Waker = _UnixWaker
 else:
     # Primarily Windows and Jython.
-    _Waker = _SocketWaker
+    _Waker = _SocketWaker  # type: ignore[misc,assignment]
+
 
 
 class _SIGCHLDWaker(_FDWaker):


src/twisted/internet/cfreactor.py
```diff
@@ -44,7 +44,8 @@
 _preserveSOError = 1 << 6
 
 
-class _WakerPlus(_Waker):
+
+class _WakerPlus(_Waker):  # type: ignore[valid-type,misc]
```
adiroiban: I don't understand why the extra typing comment is required. _Waker should now by any _IWaker and _WakerPlus implements that interface.

adiroiban (commented):
Looks good. Thanks for creating this PR.

I think that this should be reviewed by someone else as the current interface is the one proposed by myself... so I can't review my own interface.

The `_FDWaker` is missing the wakeUp.  Maybe rename _FDWaker to _FDSelfReader


_SIGCHLDWaker also has no wakeUp since it is triggered implicitly by the signal handling.

I guess that this is still a better than the current code, at least for _UnixWaker and _SocketWaker

src/twisted/internet/cfreactor.py
rodrigc: Without that comment, mypy will give this error in this branch:

```
src/twisted/internet/cfreactor.py:48:18: error: Variable "twisted.internet.posixbase._Waker" is not valid as a type  [valid-type]
src/twisted/internet/cfreactor.py:48:18: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
src/twisted/internet/cfreactor.py:48:18: error: Invalid base class "_Waker"  [misc]
```

`_Waker` is not a concrete type according to the static checker, because here: https://github.com/twisted/twisted/pull/1329/files#diff-efe69815061dceb870b109751c0bd433L192
_Waker can be a `_UnixWaker` or a `_SocketWaker`.  Even though with this patch both those classes `@implementer _IWaker`, from the static checker's perspective, those are two different classes.  So `_WakerPlus` either inherits from one of those classes or another, depending on the platform.  So `_Waker` is a dynamic base class, which mypy does not support.

adiroiban: what a mess... not sure if this is a mypy or mypy-zope bug... or I am not using mypy-zope the right way

to me, it looks like a mypy-zope bug

Reading the linked docs, it looks like _Waker need an explicit type definition.

My understanding is that the following should do the job and is the-right-way(tm) to have the type defined

```diff
diff --git a/src/twisted/internet/posixbase.py b/src/twisted/internet/posixbase.py
index 9bda9c9a4..49390637f 100644
--- a/src/twisted/internet/posixbase.py
+++ b/src/twisted/internet/posixbase.py
@@ -223,8 +223,9 @@ class _UnixWaker(_FDWaker):
 
 
 
+_Waker: _IWaker
 if platformType == 'posix':
-    _Waker = _UnixWaker  # type: Type[_IWaker]
+    _Waker = _UnixWaker
 else:
     # Primarily Windows and Jython.
     _Waker = _SocketWaker
```

adiroiban: I have created this report https://github.com/Shoobx/mypy-zope/issues/19

adiroiban: ok... so it looks like mypy is ignored  when we have a variable type alias :)  ... just variable types are supported... and just alias types are supported... but when you combine them things get messy.

My take is to consider _Waker a type alias and to ignore the variable assignment error.


the diff would be as below.

In this way _Waker is a valid type

```diff
diff --git a/src/twisted/internet/cfreactor.py b/src/twisted/internet/cfreactor.py
index 4c9d0b60a..2de8ccb22 100644
--- a/src/twisted/internet/cfreactor.py
+++ b/src/twisted/internet/cfreactor.py
@@ -45,7 +45,7 @@ _preserveSOError = 1 << 6
 
 
 
-class _WakerPlus(_Waker):  # type: ignore[valid-type,misc]
+class _WakerPlus(_Waker):
     """
     The normal Twisted waker will simply wake up the main loop, which causes an
     iteration to run, which in turn causes L{ReactorBase.runUntilCurrent}
diff --git a/src/twisted/internet/posixbase.py b/src/twisted/internet/posixbase.py
index 9bda9c9a4..fc1c83a64 100644
--- a/src/twisted/internet/posixbase.py
+++ b/src/twisted/internet/posixbase.py
@@ -224,10 +224,10 @@ class _UnixWaker(_FDWaker):
 
 
 if platformType == 'posix':
-    _Waker = _UnixWaker  # type: Type[_IWaker]
+    _Waker = _UnixWaker
 else:
     # Primarily Windows and Jython.
-    _Waker = _SocketWaker
+    _Waker = _SocketWaker  # type: ignore[misc,assignment]
 
 
```

rodrigc: OK, I incorporated this patch.

adiroiban: with this path, all _Waker will be _UnixWaker ... not really what we want... but I have no idea how to define it in mypy

adiroiban (approved):
Thanks! All good.

PR #1332, opened 2020-07-02
9887 - twisted.web.twcgi can now handle url parameters in python 3

ndg63276:
##  Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9887

### DIFF
--- a/src/twisted/web/test/test_cgi.py
+++ b/src/twisted/web/test/test_cgi.py
@@ -101,15 +101,7 @@ class PythonScript(twcgi.FilteredScript):
 
 
 
-class CGITests(unittest.TestCase):
-    """
-    Tests for L{twcgi.FilteredScript}.
-    """
-
-    if not interfaces.IReactorProcess.providedBy(reactor):
-        skip = "CGI tests require a functional reactor.spawnProcess()"
-
-
+class _StartServerAndTearDownMixin:
     def startServer(self, cgi):
         root = resource.Resource()
         cgipath = util.sibpath(__file__, cgi)
@@ -129,6 +121,16 @@ class CGITests(unittest.TestCase):
         with open(cgiFilename, 'wt') as cgiFile:
             cgiFile.write(source)
         return cgiFilename
+
+
+
+class CGITests(_StartServerAndTearDownMixin, unittest.TestCase):
+    """
+    Tests for L{twcgi.FilteredScript}.
+    """
+
+    if not interfaces.IReactorProcess.providedBy(reactor):
+        skip = "CGI tests require a functional reactor.spawnProcess()"
 
 
     def test_CGI(self):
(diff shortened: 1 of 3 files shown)


rodrigc (approved):
This looks OK

rodrigc:
@ndg63276 Do you think you could add a test to `src/twisted/web/test/test_cgi.py` for this?

exarkun:
What does "Approved" and also "Can you add a test" mean in one review?  It seems kind of ambiguous.

src/twisted/web/test/test_cgi.py
```diff
@@ -363,6 +374,44 @@ class CGIScriptTests(unittest.TestCase):
...
+
+
+    def test_urlParameters(self):
+        """
+        If the CGI script is passed URL parameters, dont fall over,
+        as per ticket 9887.
+        """
+        cgiFilename = os.path.abspath(self.mktemp())
+        with open(cgiFilename, 'wt') as cgiFile:
+            cgiFile.write(URL_PARAMETER_CGI)
```
exarkun: This file could also just be a sibling of `test_cgi.py` and then it doesn't need to be written out to a temporary file every time the test runs.

src/twisted/web/test/test_cgi.py
```diff
@@ -363,6 +374,44 @@ class CGIScriptTests(unittest.TestCase):
...
+            return self.p.stopListening()
+
+
+    def startServer(self, cgi):
+        root = resource.Resource()
+        cgipath = util.sibpath(__file__, cgi)
+        root.putChild(b"cgi", PythonScript(cgipath))
+        site = server.Site(root)
+        self.p = reactor.listenTCP(0, site)
+        return self.p.getHost().port
```
exarkun: This style of testing is, unfortunately, prone to spurious failures.  `listenTCP(0, ...)` will fail in some cases.  Also, the platform TCP/IP stack is necessary for testing the application changes here.  Instead, you can exercise all of the logic by moving bytes around in-memory.

I'm not sure what the current best practices in Twisted are for this kind of testing but I know that there is `twisted.test.iosim` and `twisted.internet.testing` with both offer tools that might be of use.

src/twisted/web/test/test_cgi.py
```diff
@@ -363,6 +372,45 @@ class CGIScriptTests(unittest.TestCase):
     """
     Tests for L{twcgi.CGIScript}.
     """
+    def tearDown(self):
```
rodrigc: Thanks for implementing this test.  The `teardown()` and `startServer` methods have been copied from `class CGITests`, which is fine to get this test going.
Can you refactor these two methods into a new mixin class, and reuse the mixin in `CGITests` and `CGIScriptTests`.

Also, as you are finishing this off, can you run the tox command here: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted

just to double check you are addressing any style issues before pushing to GitHub?

src/twisted/web/test/test_cgi.py
ndg63276: Sorry, I'm new to this kind of testing. This construction is used repeatedly in test_cgi.py.

src/twisted/web/test/test_cgi.py
ndg63276: This was just copied from CGITests.startServer in the same file.

ndg63276:
```
  lint: commands succeeded
  newsfragment: commands succeeded
  manifest-checker: commands succeeded
  congratulations :)

rodrigc (approved):
I think this looks OK. 

It looks the the use case you discovered on Python 3 was missed by the existing tests.  Thanks for writing a new test to catch this use case.

The test you wrote follows the convention of the existing CGI tests, so that is OK.

It would be possible to refactor the CGI tests to not actually start a server which listens on a socket as pointed out by exarkun.  If you are motivated enough to dig into the Twisted test code and make that change, please file another PR to make that change.
However that might be a bit intimidating for a new contributor to tackle, so don't feel pressured to do that.

What you contributed in this PR is a valid fix with a nice test to get things going.

PR #1333, opened 2020-07-02
[mypy] [9889] Reduce mypy errors

rodrigc:
https://twistedmatrix.com/trac/ticket/9889

### DIFF
--- a/src/twisted/internet/test/test_epollreactor.py
+++ b/src/twisted/internet/test/test_epollreactor.py
@@ -5,14 +5,18 @@
 Tests for L{twisted.internet.epollreactor}.
 """
 
+from unittest import skipIf
 
 from twisted.trial.unittest import TestCase
-try:
-    from twisted.internet.epollreactor import _ContinuousPolling
-except ImportError:
-    _ContinuousPolling = None
+from twisted.internet.posixbase import _ContinuousPolling
 from twisted.internet.task import Clock
 from twisted.internet.error import ConnectionDone
+
+
+try:
+    from twisted.internet import epollreactor
+except ImportError:
+    epollreactor = None  # type: ignore[assignment,misc]
 
 
 
@@ -43,6 +47,7 @@ class Descriptor(object):
 
 
 
+@skipIf(not epollreactor, "epoll not supported in this environment.")
 class ContinuousPollingTests(TestCase):
     """
     L{_ContinuousPolling} can be used to read and write from C{FileDescriptor}
@@ -242,6 +247,3 @@ class ContinuousPollingTests(TestCase):
         writer = object()
         poller.addWriter(writer)
         self.assertIn(writer, poller.getWriters())
-
-    if _ContinuousPolling is None:
-        skip = "epoll not supported in this environment."
--- a/src/twisted/words/test/test_jabberjstrports.py
+++ b/src/twisted/words/test/test_jabberjstrports.py
@@ -12,6 +12,7 @@ from twisted.words.protocols.jabber import jstrports
 from twisted.application.internet import TCPClient
 
 
+
 class JabberStrPortsPlaceHolderTests(unittest.TestCase):
     """
     Tests for L{jstrports}
--- a/src/twisted/test/test_failure.py
+++ b/src/twisted/test/test_failure.py
@@ -23,7 +23,7 @@ from twisted.trial.unittest import SynchronousTestCase
 
 
 try:
-    from twisted.test import raiser
+    from twisted.test import raiser  # type: ignore[attr-defined]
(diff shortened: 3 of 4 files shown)


src/twisted/internet/test/test_epollreactor.py
```diff
@@ -8,9 +8,9 @@
 
 from twisted.trial.unittest import TestCase
 try:
-    from twisted.internet.epollreactor import _ContinuousPolling
+    from twisted.internet.posixbase import _ContinuousPolling
```
adiroiban: I think that we should have epollreactor ... this is a test for epollreactor so I think that we should have an explicit import here.

src/twisted/_threads/test/test_team.py
```diff
@@ -8,7 +8,7 @@
 
 from twisted.trial.unittest import SynchronousTestCase
 
-from twisted.python.context import call, get
+from twisted.python.context import call, get  # type: ignore[attr-defined]
```
adiroiban: I don't know to say about t.p.context.py

I have no idea why installContextTracker is a public API.

@glyph do you see any issue with the following diff  which removes installContextTracker and the use of global

```diff
diff --git a/src/twisted/python/context.py b/src/twisted/python/context.py
index 1348822e1..e208e3dc2 100644
--- a/src/twisted/python/context.py
+++ b/src/twisted/python/context.py
@@ -22,7 +22,6 @@ defaultContextDict = {}  # type: Dict[Type[object], Dict[str, str]]
 setDefault = defaultContextDict.__setitem__
 
 
-
 class ContextTracker:
     """
     A L{ContextTracker} provides a way to pass arbitrary key/value data up and
@@ -124,13 +123,7 @@ class ThreadedContextTracker(object):
         return self.currentContext().getContext(key, default)
 
 
-def installContextTracker(ctr):
-    global theContextTracker
-    global call
-    global get
-
-    theContextTracker = ctr
-    call = theContextTracker.callWithContext
-    get = theContextTracker.getContext
 
-installContextTracker(ThreadedContextTracker())
+theContextTracker = ThreadedContextTracker()
+call = theContextTracker.callWithContext
+get = theContextTracker.getContext
```

src/twisted/words/test/test_jabberjstrports.py
```diff
@@ -9,7 +9,8 @@
 from twisted.trial import unittest
 
 from twisted.words.protocols.jabber import jstrports
-from twisted.application.internet import TCPClient
+from twisted.application.internet import TCPClient  # type: ignore[attr-defined]  # noqa
```
adiroiban: this is ok... the code from t.application.internet is ingenious  and we can keep it... I would vote for refactoring it to make it easier to read... for multicast and UDP we only have server-side applicitons.

below is the code from application/internet.py
```
_doc={
'Client':
"""Connect to %(tran)s

Call reactor.connect%(tran)s when the service starts, with the
arguments given to the constructor.
""",
'Server':
"""Serve %(tran)s clients

Call reactor.listen%(tran)s when the service starts, with the
arguments given to the constructor. When the service stops,
stop listening. See twisted.internet.interfaces for documentation
on arguments to the reactor method.
""",
}

for tran in 'TCP UNIX SSL UDP UNIXDatagram Multicast'.split():
    for side in 'Server Client'.split():
        if tran == "Multicast" and side == "Client":
            continue
        if tran == "UDP" and side == "Client":
            continue
        base = globals()['_Abstract'+side]
        doc = _doc[side] % vars()

        klass = type(tran+side, (base,), {'method': tran, '__doc__': doc})
        globals()[tran+side] = klass
```

adiroiban (commented):
just a comment for now... but I would say that this require a change and have an explicit import for epollreactor

thanks

src/twisted/internet/test/test_epollreactor.py
rodrigc: OK, I added an explicit import of epollreactor to test_epollreactor.py

src/twisted/words/test/test_jabberjstrports.py
rodrigc: OK, I removed the change for this PR, and refactored the code here: https://github.com/twisted/twisted/pull/1339

src/twisted/_threads/test/test_team.py
rodrigc: OK, I removed changes from this PR, and created a new PR to address this: https://github.com/twisted/twisted/pull/1340

rodrigc:
@adiroiban Can I merge this?  I incorporated your feedback.

adiroiban (approved):
All good. Thanks for the update!

PR #1338, opened 2020-07-05
Fix AsyncioSelectorReactor on Windows Python 3.8+

rodrigc:
https://twistedmatrix.com/trac/ticket/9766
https://twistedmatrix.com/trac/ticket/9809

twisted.internet.asyncioreactor.AsyncioSelectorReactor has been changed to use asyncio.SelectorEventLoop

if asyncio.get_event_loop() does not return a SelectorEventLoop.
This fixes AsyncioSelectorReactor on Python 3.8+, where
asyncio.get_event_loop() was changed to return a ProactorEventLoop
in [bpo-34687](https://bugs.python.org/issue34687).  See: https://github.com/python/cpython/commit/37aae9dcf18753b8ffda99d1a5758a90af852464.


azure-pipelines/windows_test_jobs.yml
```diff
@@ -24,4 +24,4 @@ jobs:
         platform: windows
         pythonVersion: ${{ pyver.value }}
         windowsReactor: ${{ parameters.reactor }}
-        allowFailure: ${{ eq(pyver.value, '3.8') }}
+        allowFailure: false
```
adiroiban: I think that we can remove this whole line as false is the default value.

src/twisted/internet/test/test_asyncioreactor.py
```diff
@@ -25,6 +25,8 @@ class AsyncioSelectorReactorTests(ReactorBuilder, SynchronousTestCase):
     """
     L{AsyncioSelectorReactor} tests.
     """
+    @skipIf(not isinstance(get_event_loop(), SelectorEventLoop),
+            "SelectorEventLoop is not default event loop")
```
adiroiban: I don't understand this test.

I think that this test should not be skipped but instead set it up in such a way that both twisted and asyncio are using the same event loop.

I guess that the documentaiton for AsyncioSelectorReactor should be updated to explain that if someone want to mix Twisted with asyncio they must do the required setup  for having twisted and asynio on the same event loop.

With the comment that everything is ok on Linux/macOS and that for Windows extra steps are required.

I have never used asyncio so I have no idea how other libraries are setting up their event loop.

maybe have something like https://github.com/tornadoweb/tornado/pull/2686/files

src/twisted/internet/asyncioreactor.py
```diff
@@ -37,14 +37,18 @@ class AsyncioSelectorReactor(PosixReactorBase):
...
 
-    def __init__(self, eventloop=None):
+    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
 
         if eventloop is None:
             eventloop = get_event_loop()
+            # On Python 3.8+, asyncio.get_event_loop() on
+            # Windows was changed to return a ProactorEventLoop
+            if not isinstance(eventloop, SelectorEventLoop):
+                eventloop = SelectorEventLoop()
```
adiroiban: I see that in Tornodo,  the recommended way to use the selector is by calling 

```
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
https://docs.python.org/3/library/asyncio-policy.html#asyncio-policies

--------

If in Twisted we create a new eventloop, Twisted will run in its own loop and will not share events with other async libs

adiroiban (commented):
I have never used asyncio but just by reading the documentation this doesn't look right.

I think that Twisted should use the global asyncio event loop instead of creating a separate event loop.
I guess that this is the purpose of test_defaultEventLoopFromGlobalPolicy which now fails.
This PR is green as the test was skipped.
But I think that the test should not be skipped and instead the code should be updated to have it pass.

Now I don't know if Twisted should auto set the selector policy.

Maybe is best to follow the Tornodo example and have each user set its own policy.

Just my feedback... feel free to ignore it.

Thanks!

azure-pipelines/windows_test_jobs.yml
rodrigc: Removed.

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: I think following Tornado's example is the way to go.

src/twisted/internet/asyncioreactor.py
rodrigc: I've updated the patch, and implemented this.

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: I've implemented this

src/twisted/internet/test/test_asyncioreactor.py
```diff
@@ -5,35 +5,43 @@
...
-        """
-        reactor = AsyncioSelectorReactor()
-        future = asyncio.Future()
+
+    _defaultEventLoop = DefaultEventLoopPolicy().new_event_loop()
+    _defaultEventLoopIsSelector = isinstance(_defaultEventLoop,
+                                             SelectorEventLoop)
+
+    def _test_reactor(self, reactor):
+        future = Future()
```
adiroiban, line 53: I think that this needs a docstring.
It can be renamed to something like `def assertSharedEventLoop` .
No need to make it private as this is testing code.
To me, this looks like specialized assertion code.

src/twisted/internet/test/test_asyncioreactor.py
```diff
@@ -48,8 +56,107 @@ def completed(future):
...
+        """
+        set_event_loop_policy(WindowsSelectorEventLoopPolicy())
+        reactor = AsyncioSelectorReactor()
+        self._test_reactor(reactor)
+        set_event_loop_policy(None)
+
+
     def test_seconds(self):
         """L{seconds} should return a plausible epoch time."""
+        if hasWindowsSelectorEventLoopPolicy:
```
adiroiban, line 186: I think that all the tests using _test_reactor should e moved into a separate test case.
And the `AsyncioSelectorReactorTests` test case can have a setup method in which it will set WindowsSelectorEventLoopPolicy and reset on teardown.

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,61 @@
...
+                raise ValueError(
+                    "get_event_loop() returns: {} which is not a "
+                    "SelectorEventLoop".format(_eventloop))
+
+            eventloop = _eventloop
+        else:
+            if not isinstance(eventloop, SelectorEventLoop):
+                raise ValueError(
+                    "eventloop is: {} which is not a subclass of "
+                    "SelectorEventLoop".format(eventloop))
```
adiroiban: maybe simplify the sanity checks 
 
```suggestion
            eventloop = _eventloop
 
        # On Python 3.8+, asyncio.get_event_loop() on
        # Windows was changed to return a ProactorEventLoop
        if not isinstance(eventloop, SelectorEventLoop):
            raise ValueError(
                "eventloop is: {} which is not a subclass of "
                "SelectorEventLoop".format(eventloop))
```

adiroiban (commented):
It looks good. Thanks!
I left some minor comments.
The final review should be done by someone using asyncio

src/twisted/internet/asyncioreactor.py
rodrigc: Done

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: Done

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: Done, renamed to `assertReactorWorksWithAsyncioFuture`

rodrigc:
@adiroiban can you have another look?  I don't think that many people are using Twisted with Python's asyncio, because:
1.  Python's asyncio works on Python-3 only, and many Twisted users were on Python 2 for a long time
2. Python's asyncio has been changing in every Python release

The main intent of this PR is to just keep up with all the changes in Python's asyncio, so that for people who want to use it in Twisted, they will have a better time, especially on Windows.

adiroiban:
It looks good to me. But I am not using async io. I can't check things in a real production environment.

If there is no other twisted core developer interested in asyncio I think that is best to move this code outside of twisted/twisted.

rodrigc:
This PR affects a reactor implementation.  Reactor implementations typically have been implemented in the core of Twisted, not outside.

src/twisted/conch/test/test_knownhosts.py
```diff
@@ -1186,6 +1188,8 @@ class FakeObject(object):
     """
 
 
+
+@skipIf(not FilePath("/dev/tty").exists(), "Platform lacks /dev/tty")
```
altendky, line 1159: Did this PR trigger this to be needed?  It's not clear to me why it would.

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,56 @@
...
+    with Python 3.8+ must call
+    C{asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())}
+    before instantiating and running L{AsyncioSelectorReactor}.
     """
     _asyncClosed = False
     _log = Logger()
 
-    def __init__(self, eventloop=None):
+    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
+        if not eventloop:
```
altendky:
```suggestion
        if eventloop is None:
```
Is there a reason this needed to change?

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,56 @@
...
+            _eventloop = get_event_loop()  # type: AbstractEventLoop
+            eventloop = cast(Optional[SelectorEventLoop], _eventloop)
```
altendky: If the cast is retained, it doesn't need the `Optional[]` does it?

```suggestion
            eventloop = cast(SelectorEventLoop, _eventloop)
```

Though I don't think the cast should be needed given the `isinstance()` check below.  Casts strike me as mostly just bypassing the type hinting and as such should be used sparingly.

```python
    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
        if not eventloop:
            selected_eventloop = get_event_loop()  # type: AbstractEventLoop
        else:
            selected_eventloop = eventloop

        # On Python 3.8+, asyncio.get_event_loop() on
        # Windows was changed to return a ProactorEventLoop
        if not isinstance(selected_eventloop, SelectorEventLoop):
            raise ValueError(
                "eventloop is: {} which is not a subclass of "
                "SelectorEventLoop".format(eventloop))

        self._asyncioEventloop = selected_eventloop  # type: SelectorEventLoop
```

(sorry this overlaps the other suggestions...)

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,56 @@
...
 
-        if eventloop is None:
-            eventloop = get_event_loop()
+        # On Python 3.8+, asyncio.get_event_loop() on
+        # Windows was changed to return a ProactorEventLoop
```
altendky, line 57:
```suggestion
        # On Python 3.8+, asyncio.get_event_loop() on
        # Windows was changed to return a ProactorEventLoop
        # unless the loop policy has been changed.
```

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,56 @@
...
+        if not isinstance(eventloop, SelectorEventLoop):
+            raise ValueError(
+                "eventloop is: {} which is not a subclass of "
+                "SelectorEventLoop".format(eventloop))
 
         self._asyncioEventloop = eventloop
```
altendky:
```suggestion
        self._asyncioEventloop = selected_eventloop  # type: SelectorEventLoop
```
Just to be explicit and overconstrain against the above logic to make sure it does what we want.  Maybe?

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,56 @@
...
+            eventloop = cast(Optional[SelectorEventLoop], _eventloop)
 
-        if eventloop is None:
-            eventloop = get_event_loop()
+        # On Python 3.8+, asyncio.get_event_loop() on
+        # Windows was changed to return a ProactorEventLoop
+        if not isinstance(eventloop, SelectorEventLoop):
+            raise ValueError(
+                "eventloop is: {} which is not a subclass of "
+                "SelectorEventLoop".format(eventloop))
```
altendky:
```suggestion
            raise ValueError(
                "SelectorEventLoop required, instead got:"
                + " {}".format(eventloop))
```
It's not obvious to me that the variable name should be referenced, especially since the value might come from inside this function as opposed to the parameter.

It doesn't seem that there is any Twisted policy on this, but for the record, my preference is to have specific exceptions from the library rather than general stdlib exceptions that are harder to achieve uniqueness of.  Picking apart exceptions based on their message is a pain when trying to catch only specific things.

src/twisted/internet/test/test_asyncioreactor.py
```diff
@@ -48,8 +60,107 @@ def completed(future):
         self.assertEqual(result, [True])
 
 
+    @skipIf(not _defaultEventLoopIsSelector,
+            "default event loop: {}\nis not is of type SelectorEventLoop "
```
altendky:
```suggestion
            "default event loop: {}\nis not of type SelectorEventLoop "
```

altendky:
@rodrigc, what did you change?  Given it was a force repush of all the commits, is there some way for me to see the modifications?

altendky (changes requested):
Since things are still changing, here's what I've got for now.  I can follow up after these points are gone through and other in-work changes are done.

azure-pipelines/windows_test_jobs.yml
```diff
@@ -32,4 +32,3 @@ jobs:
         platform: windows
         pythonVersion: ${{ pyver.value }}
         windowsReactor: ${{ parameters.reactor }}
-        allowFailure: ${{ eq(pyver.value, '3.8') }}
```
altendky, line 35: I don't see any benefit in crossing this with PR #1259 and ticket [9809](https://twistedmatrix.com/trac/ticket/9809).

src/twisted/internet/asyncioreactor.py
rodrigc: OK, incorporated.

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: OK, incorporated

src/twisted/internet/asyncioreactor.py
rodrigc: The cast and optional is still needed because since the`eventloop` parameter can be `None`, it needs to be `Optional` when specifying the type.

src/twisted/internet/asyncioreactor.py
rodrigc: OK, incorporated.

src/twisted/internet/asyncioreactor.py
rodrigc: OK, incorporated.

src/twisted/internet/asyncioreactor.py
rodrigc: I don't understand the sentence you wrote.  Can you clarify?

src/twisted/conch/test/test_knownhosts.py
rodrigc: For some reason, this test started failing, so I added it to make further progress on this PR.

There have been hundreds of lines of changes in trunk recently, so I have been updating this PR to stay current.

azure-pipelines/windows_test_jobs.yml
rodrigc: This is a minor one line change which simply removes ignoring failurs on Python 3.8
during Windows CI.  It's OK to leave this in here and address ticket 9809 at the same time.

rodrigc:
@altendky I incorporated your review feedback

azure-pipelines/windows_test_jobs.yml
altendky: If there weren't an existing PR addressing the existing ticket, sure.  I don't get why it needs to be confused by mixing it in here.

src/twisted/conch/test/test_knownhosts.py
altendky: We should probably know why it is needed if it is going to be included?

src/twisted/internet/asyncioreactor.py
altendky: Just casting to `SelectorEventLoop` seems to be fine given that that will satisfy an optional one.  Note that `eventloop` is not considered optional by mypy after the if.

https://mypy-play.net/?mypy=latest&python=3.8&gist=cb0b71f99dedf0745d49ce76a0f6dea3
```python
from asyncio import AbstractEventLoop, SelectorEventLoop, get_event_loop
from typing import Optional, cast

reveal_type(get_event_loop)

class C:
    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
        if eventloop is None:
            _eventloop = get_event_loop()  # type: AbstractEventLoop
            eventloop = cast(SelectorEventLoop, _eventloop)
        
        reveal_type(eventloop)
```
```
main.py:4: note: Revealed type is 'def () -> asyncio.events.AbstractEventLoop'
main.py:12: note: Revealed type is 'asyncio.unix_events._UnixSelectorEventLoop'
```

But given that we can comfortably skip the casting entirely and let mypy really check things properly, isn't that a better option?

src/twisted/internet/asyncioreactor.py
altendky: It seems it might be good to explicitly state the type we want the `_asyncioEventLoop` attribute to have.  This avoids refactors of the logic preceding this assignment accidentally creating something else and unintentionally changing the type of the attribute.  It also causes mypy to check and confirm the result of the logic above really achieves the type we want.

src/twisted/internet/asyncioreactor.py
altendky: Here it is all together with both forms satisfying (at least a basic config of) mypy.

https://mypy-play.net/?mypy=latest&python=3.8&gist=112efdb08e33c25b8dbb6458261c574a
```python
from asyncio import AbstractEventLoop, SelectorEventLoop, get_event_loop
from typing import Optional, cast

reveal_type(get_event_loop)

class C:
    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
        if eventloop is None:
            _eventloop = get_event_loop()
            eventloop = cast(SelectorEventLoop, _eventloop)
        
        reveal_type(eventloop)

        if not isinstance(eventloop, SelectorEventLoop):
            raise ValueError(
                "eventloop is: {} which is not a subclass of "
                "SelectorEventLoop".format(eventloop))

        reveal_type(eventloop)
        self._asyncioEventloop = eventloop  # type: SelectorEventLoop


class D:
    def __init__(self, eventloop: Optional[SelectorEventLoop] = None):
        if eventloop is None:
            selected_eventloop = get_event_loop()
        else:
            selected_eventloop = eventloop

        reveal_type(selected_eventloop)

        if not isinstance(selected_eventloop, SelectorEventLoop):
            raise ValueError(
                "eventloop is: {} which is not a subclass of "
                "SelectorEventLoop".format(selected_eventloop))

        reveal_type(selected_eventloop)
        self._asyncioEventloop = selected_eventloop  # type: SelectorEventLoop
```

```
main.py:4: note: Revealed type is 'def () -> asyncio.events.AbstractEventLoop'
main.py:12: note: Revealed type is 'asyncio.unix_events._UnixSelectorEventLoop'
main.py:19: note: Revealed type is 'asyncio.unix_events._UnixSelectorEventLoop'
main.py:30: note: Revealed type is 'asyncio.events.AbstractEventLoop'
main.py:37: note: Revealed type is 'asyncio.unix_events._UnixSelectorEventLoop'
```

graingert: Yes, mypy allows type refinement of a name in a scope

src/twisted/internet/asyncioreactor.py
```diff
@@ -9,42 +9,57 @@
...
+            _eventloop = get_event_loop()  # type: AbstractEventLoop
+            eventloop = cast(Optional[SelectorEventLoop], _eventloop)
+
+        # On Python 3.8+, asyncio.get_event_loop() on
+        # Windows was changed to return a ProactorEventLoop
+        # unless the loop policy has been changed.
+        if not isinstance(eventloop, SelectorEventLoop):
+            raise ValueError(
+                "SelectorEventLoop required, instead got: {}".format(
+                eventloop))
```
mthuurne, line 58: If the event loop can be rejected because of its type, maybe raising `TypeError` is more appropriate?

src/twisted/internet/asyncioreactor.py
rodrigc: Thanks for the clarification.  I have incorporated your suggestion.

src/twisted/internet/asyncioreactor.py
rodrigc: @altendky I'm sorry, but I don't understand what you are asking for in your review feedback.  Can you clarify things more clearly?

altendky: Do not use `cast()`.  It is basically disabling mypy.  Instead of letting mypy detect what the types are, you are telling mypy to just believe you that the types are what you want them to be.  There are presumably various corners cases where you find yourself between a bug in mypy and a not-yet-implemented feature in mypy and `cast()` is the 'easy' solution.  This isn't such a case.

The `cast()` is also fairly redundant with the `isinstance()` check which is another indicator that it isn't a good approach.

I provided alternative code without any casting that passes mypy and should end up doing the same thing as the code in the PR.  It also isn't particularly more complicated, if at all.

rodrigc: OK, I incorporated your feedback and removed the `cast()`

src/twisted/internet/test/test_asyncioreactor.py
```diff
@@ -47,8 +64,122 @@ def completed(future):
...
+        not _defaultEventLoopIsSelector,
+        "default event loop: {}\nis not of type SelectorEventLoop "
+        "on Python {}.{} ({})".format(
+            type(_defaultEventLoop),
+            sys.version_info.major,
+            sys.version_info.minor,
+            platform.getType(),
+        ),
+    )
+    def test_newSelectorEventLoopFromDefaultEventLoopPolicy(self):
```
twm, line 97: There should be a docstring here.

src/twisted/conch/test/test_knownhosts.py
twm: This change looks odd to me. Those tests mock the open function, so how is it failing? If you have a reference to the failure you saw please file a ticket; there may be something wrong with the suite.

Do the tests still fail without this change?

src/twisted/internet/test/test_asyncioreactor.py
twm: I agree, but I don't think it's necessary to hold up the PR for this.

twm (changes requested):
Please fill in the missing docstring. Otherwise this looks okay to me, with the caveat that I am assuming the problem description to be correct as I know nothing of Windowsland.

src/twisted/internet/test/test_asyncioreactor.py
rodrigc: OK, I have added the docstring

twm (approved):
Thanks!

src/twisted/internet/test/test_asyncioreactor.py
altendky: It should at least have been a context manager or `try`/`finally` since presently it will just not reset the policy for any failing test...  plus I'm already going to go copy/paste this stuff over into #1263 just to catch up.

src/twisted/conch/test/test_knownhosts.py
altendky: Follow up ticket: https://twistedmatrix.com/trac/ticket/9966

PR #1339, opened 2020-07-07
[mypy] [9895] Define Server and Client classes explicitly

rodrigc:
https://twistedmatrix.com/trac/ticket/9895

d562a59
Instead, just define these classes explicitly:

- TCPServer
- TCPClient
- UNIXServer
- UNIXClient
- SSLServer
- SSLClient
- UNIXDatagramServer
- UNIXDatagramClient
- UDPServer
- MulticastServer

This eliminates mypy errors such as:

```
src/twisted/words/test/test_jabberjstrports.py:12:1: error: Module 'twisted.application.internet' has no attribute 'TCPClient'  [attr-defined]
    from twisted.application.internet import TCPClient
```

### DIFF
--- a/src/twisted/application/internet.py
+++ b/src/twisted/application/internet.py
@@ -97,7 +97,7 @@ class _AbstractServer(_VolatileDataService):
     """
 
     volatile = ['_port']
-    method = None
+    method = ""  # type: str
     reactor = None
 
     _port = None
@@ -138,8 +138,8 @@ class _AbstractServer(_VolatileDataService):
         @rtype: an object providing
             L{twisted.internet.interfaces.IListeningPort}.
         """
-        return getattr(_maybeGlobalReactor(self.reactor),
-                       'listen%s' % (self.method,))(*self.args, **self.kwargs)
+        return getattr(_maybeGlobalReactor(self.reactor), 'listen{}'.format(
+                       self.method,))(*self.args, **self.kwargs)
 
 
 
@@ -161,7 +161,7 @@ class _AbstractClient(_VolatileDataService):
     """
 
     volatile = ['_connection']
-    method = None
+    method = ""  # type: str
     reactor = None
 
     _connection = None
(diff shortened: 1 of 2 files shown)


src/twisted/application/internet.py
```diff
@@ -97,7 +97,7 @@ class _AbstractServer(_VolatileDataService):
     """
 
     volatile = ['_port']
-    method = None
+    method = None  # type: Optional[str]
```
mthuurne: I don't think `method` is actually optional: every concrete class provides a string. In Python 3.6 you could simply write `method: str` without assigning a value, but I don't know about any elegant 3.5 equivalent. Casting `None` to `str` would work and even though it looks ugly, it would simplify code that wants to use the `method` attribute.

src/twisted/application/internet.py
```diff
@@ -197,34 +197,82 @@ def _getConnection(self):
...
-        if tran == "UDP" and side == "Client":
-            continue
-        base = globals()['_Abstract'+side]
-        doc = _doc[side] % vars()
 
-        klass = type(tran+side, (base,), {'method': tran, '__doc__': doc})
-        globals()[tran+side] = klass
+
+class TCPServer(_AbstractServer):
+    __doc__ = _doc['Server'].format(tran='TCP')
```
mthuurne: While this is a lot simpler than before, this expression is still too complex for pydoctor to be able to compute the docstring using static analysis.

If you could split `_doc` into `_serverDoc` and `_clientDoc`, that would simplify the expression. I checked and found no dynamic lookups in `_doc` in the module.

src/twisted/application/internet.py
```diff
@@ -1149,9 +1197,9 @@ def stopService(self):
...
+
 __all__ = (['TimerService', 'CooperatorService', 'MulticastServer',
             'StreamServerEndpointService', 'UDPServer',
-            'ClientService'] +
-           [tran + side
-            for tran in 'TCP UNIX SSL UNIXDatagram'.split()
-            for side in 'Server Client'.split()])
+            'ClientService', 'TCPServer', 'TCPClient',
+            'UNIXServer', 'UNIXClient', 'SSLServer', 'SSLClient',
+            'UNIXDatagramServer', 'UNIXDatagramClient'])
```
mthuurne: The outer parentheses no longer serve a purpose now that everything is in a single list.

src/twisted/application/internet.py
rodrigc: `method: str` doesn't work in Python 3.5

src/twisted/application/internet.py
rodrigc: Fixed

src/twisted/application/internet.py
rodrigc: Fixed

src/twisted/application/internet.py
rodrigc: I think we can define method as the empty string.

mthuurne (approved):
Looks good. Thanks!

PR #1340, opened 2020-07-07
[mypy] [9896] Reduce mypy errors: explicitly define global variables in context module

rodrigc:
https://twistedmatrix.com/trac/ticket/9896

Explicitly define global variables to eliminate mypy errors:
* theContextTracker
* call
* get

This eliminates mypy errors:

```
src/twisted/_threads/test/test_team.py:11:1: error: Module 'twisted.python.context' has no attribute 'call'  [attr-defined]
    from twisted.python.context import call, get
    ^
src/twisted/_threads/test/test_team.py:11:1: error: Module 'twisted.python.context' has no attribute 'get'  [attr-defined]
    from twisted.python.context import call, get
```

### DIFF
--- a/src/twisted/python/context.py
+++ b/src/twisted/python/context.py
@@ -124,6 +124,13 @@ class ThreadedContextTracker(object):
         return self.currentContext().getContext(key, default)
 
 
+
+theContextTracker = ThreadedContextTracker()
+call = theContextTracker.callWithContext
+get = theContextTracker.getContext
+
+
+
 def installContextTracker(ctr):
     global theContextTracker
     global call
@@ -132,5 +139,3 @@ def installContextTracker(ctr):
     theContextTracker = ctr
     call = theContextTracker.callWithContext
     get = theContextTracker.getContext
-
-installContextTracker(ThreadedContextTracker())


src/twisted/python/context.py
```diff
@@ -124,6 +124,13 @@ def getContext(self, key, default=None):
...
 
 
+
+theContextTracker = ThreadedContextTracker()
+call = theContextTracker.callWithContext
+get = theContextTracker.getContext
+
+
+
 def installContextTracker(ctr):
```
adiroiban, line 134: In theory this is public API. I think that this should have a .removal release note to inform that this method was removed.
I don't think that this needs the deprecation process but please also send an email to the mailing list to let anyone know that this will be removed and see if there are any objections.
After 2 weeks this can be merged.

Maybe this API is important..so better to check with the others.

Thanks

adiroiban (commented):
Looks good but the API removal should be part of the release notes. Thanks

src/twisted/python/context.py
rodrigc: I did not remove `installContextTracker` in this PR.  So I don't think I need a `.removal` newsfragment.

adiroiban: True... Sorry. I was testing the GitHub mobile app for this review. My bad.
In this case, i think that installContextTracker needs to be deprecated here.
I don't see it used anywhere else and I think that it should be removed.
Thanks!

rodrigc: @adiroiban Can I merge this as-is and deprecate installContextTracker() in a separate PR?  @hawkowl wrote this.  I'm not sure if there are other use cases for using this that are outside the Twisted codebase.

adiroiban (approved):
All good. Thanks. Please check with Amber and see if the code can be deprecated.
Thanks.

PR #1342, opened 2020-07-09
[mypy] [9898] Change type of  twisted.internet.protocol.Factory.factory to allow AbstractDatagramProtocol.

rodrigc:
https://twistedmatrix.com/trac/ticket/9898

This fixes these mypy errors:

```
src/twisted/pair/test/test_tuntap.py:1359:24: error: Incompatible types in assignment (expression has type "Type[IPRecordingProtocol]", variable has type
"Optional[Type[Protocol]]")  [assignment]
        factory.protocol = IPRecordingProtocol
                           ^
src/twisted/pair/test/test_tuntap.py:1382:24: error: Incompatible types in assignment (expression has type "Type[EthernetRecordingProtocol]", variable has type
"Optional[Type[Protocol]]")  [assignment]
        factory.protocol = EthernetRecordingProtocol
```

### DIFF
--- a/src/twisted/internet/protocol.py
+++ b/src/twisted/internet/protocol.py
@@ -11,7 +11,7 @@ Twisted.  The Protocol class contains some introductory material.
 
 
 import random
-from typing import Optional, Type
+from typing import Optional, Type, Union
 from zope.interface import implementer
 
 from twisted.python import log, failure, components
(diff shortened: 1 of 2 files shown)


adiroiban (approved):
Looks good. Thanks. Can be merged.

Not sure if we should have AbstractProtocol which is inherited by both Protocol and AbstractDatagramProtocol

PR #1343, opened 2020-07-09
[mypy] [9899] Eliminate mypy error in src/twisted/python/test/test_components.py

rodrigc:
https://twistedmatrix.com/trac/ticket/9899

Eliminate this mypy error:

```
src/twisted/python/test/test_components.py:629:5: error: Signature of "Booable" incompatible with "yay" of supertype
"twisted.python.test.test_components.IProxiedInterface"  [override]
        def yay(self):
        ^
```


adiroiban (approved):
All good. Thanks!

PR #1345, opened 2020-07-09
[mypy] [9901] Eliminate mypy error in src/twisted/words/im/basesupport.py

rodrigc:
https://twistedmatrix.com/trac/ticket/9901

This fixes this mypy error:

```
src/twisted/words/im/basesupport.py:130:16: error: "AbstractClientMixin" has no attribute "_protoBase"  [attr-defined]
            return self._protoBase.connectionLost(self, reason)
```

`_protoBase` is defined in `AbstractClientMixin.__init__()`:
`https://github.com/twisted/twisted/blob/4bef9c70a4cd0e9f186272908c135024c7641143/src/twisted/words/im/basesupport.py#L113

### DIFF
--- a/src/twisted/words/im/basesupport.py
+++ b/src/twisted/words/im/basesupport.py
@@ -7,8 +7,7 @@
 
 You will find these useful if you're adding a new protocol to IM.
 """
-
-# Abstract representation of chat "model" classes
+from typing import Type
 
 from twisted.words.im.locals import OFFLINE, OfflineError
 
@@ -20,6 +19,8 @@ from twisted.persisted import styles
 from twisted.internet import error
 
 
+
+# Abstract representation of chat "model" classes
 
 class AbstractGroup:
     def __init__(self, name, account):
@@ -108,6 +109,8 @@ class AbstractClientMixin:
 
     @ivar _logonDeferred: Fired when I am done logging in.
     """
+    _protoBase = None  # type: Type[Protocol]
+
     def __init__(self, account, chatui, logonDeferred):
         for base in self.__class__.__bases__:
             if issubclass(base, Protocol):
@@ -127,7 +130,7 @@ class AbstractClientMixin:
     def connectionLost(self, reason: Failure = connectionDone):
         self.account._clientLost(self, reason)
         self.unregisterAsAccountClient()
-        return self._protoBase.connectionLost(self, reason)
+        return self._protoBase.connectionLost(self, reason)  # type: ignore[arg-type]  # noqa
 
 
     def unregisterAsAccountClient(self):


src/twisted/words/im/basesupport.py
```diff
@@ -121,13 +121,17 @@ def __init__(self, account, chatui, logonDeferred):
 
 
     def connectionMade(self):
-        self._protoBase.connectionMade(self)
+        _protoBase = getattr(self, '_protoBase', None)
```
adiroiban: I don't know what to day.
Instead of this hack to work around mypy reporting, I think that is better to define `_protoBase` in AbstractClientMixin

Maybe
```python
class AbstractClientMixin:
    _protoBase = None
```

or 

```python
class AbstractClientMixin:
    _protoBase : Type[Protocol]
```

mthuurne: The latter (`_protoBase : Type[Protocol]`) is by far the cleanest solution. The problem with working around it on access (`getattr()` etc.) is that every single access needs a workaround, which includes code outside of Twisted itself.

Also the extracted documentation will be incomplete (if the attribute is not declared) or wrong (if the attribute is declared with the wrong type, including `None` as an inferred type). It is possible to override this using `@type` in the docstring, but if this can be avoided the code will be easier to read and maintain.

To make it work in Python 3.5 you could cast `None` to `Type[Protocol]` or add a `type: ignore` comment. But maybe type annotations are a compelling enough reason to switch to Python 3.6 now rather than later.

mthuurne:
I don't understand why `_protoBase` goes from mandatory to optional in this PR. Is there a reason that new interface is superior? Because this change is not necessary to eliminate the mypy error.

src/twisted/words/im/basesupport.py
rodrigc: @adiroiban This won't work on Python 3.5:

```
class AbstractClientMixin:
    _protoBase : Type[Protocol]
```

rodrigc: OK, I changed to this:

```
class AbstractClientMixin:
    _protoBase = None  # type: Type[Protocol]
```

src/twisted/words/im/basesupport.py
```diff
@@ -21,6 +20,8 @@
 
 
 
+# Abstract representation of chat "model" classes
```
adiroiban, line 23: why a comment and not a docstring?

src/twisted/words/im/basesupport.py
```diff
@@ -127,7 +130,7 @@ def connectionMade(self):
     def connectionLost(self, reason: Failure = connectionDone):
         self.account._clientLost(self, reason)
         self.unregisterAsAccountClient()
-        return self._protoBase.connectionLost(self, reason)
+        return self._protoBase.connectionLost(self, reason)  # type: ignore[arg-type]  # noqa
```
adiroiban, line 133: why is this required?

adiroiban (commented):
Thanks for the update.
Just a comment for now, as I don't understand the last type hint ignore.

src/twisted/words/im/basesupport.py
rodrigc: Without that `ignore`, this error is returned:

```
src/twisted/words/im/basesupport.py:133:47: error: Argument 1 to "connectionLost" of "Protocol" has incompatible type "AbstractClientMixin"; expected "Protocol"  [arg-type]
```

adiroiban: Thanks. I think that is ok to comment it.

AbstractClientMixin is not really a mixing ... it has an init is part of multiple inheritance.
I am expecting from a mixing to not have any `__init__` and to always inherit it last

adiroiban (approved):
All good. Thanks. Please move the comment to a docstring.

src/twisted/words/im/basesupport.py
rodrigc: This comment is for multiple classes in this file, not for one specific class.
This comment was in the file before, I just moved it to after the imports.

PR #1346, opened 2020-07-09
[mypy] [9902] Fix mypy error in decorator.py

rodrigc:
https://twistedmatrix.com/trac/ticket/9902

This fixes this mypy error:

```
src/twisted/python/deprecate.py:856:16: error: Incompatible return value type (got "Callable[..., Any]", expected "_Tc")  [return-value]
```

and follows the example at: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators


wsanchez (approved):
LGTM

I'd suggest moving the cast up two lines to the assignment. It doesn't matter in this case, but I find it's a good practice to get the type correct ASAP, which enables `mypy` to check, eg. whether the first argument to `_appendToDosctring()` is the right type.

rodrigc:
> 
> I'd suggest moving the cast up two lines to the assignment. It doesn't matter in this case, but I find it's a good practice to get the type correct ASAP, which enables `mypy` to check, eg. whether the first argument to `_appendToDosctring()` is the right type.


Done.

PR #1348, opened 2020-07-10
[mypy] If mypy CI fails, then report that error

rodrigc:
https://twistedmatrix.com/trac/ticket/9903

This can be merged after https://github.com/twisted/twisted/pull/1346 is merged

### DIFF
--- a/azure-pipelines/mypy_jobs.yml
+++ b/azure-pipelines/mypy_jobs.yml
@@ -22,4 +22,3 @@ jobs:
 
   - script: 'python -m tox -e mypy'
     displayName: 'Run mypy'
-    continueOnError: true


adiroiban:
Thanks for the update.

I think that mypy should be moved to GitHub actions.
In this way you can check the status without leaving GitHub action.
It will also produce and independent commit status, not linked to the twisted.twisted Azure pipeline.

I think that "twisted.twisted" Azure pipeline should only be used for the `stable` trial tests.
I think that a separate Azure Pipeline should be created for `experimental` / `unstable` trial tests and other checks.

My suggestion for fixing this is at https://github.com/twisted/twisted/pull/1349

rodrigc:
Closing in favor of https://github.com/twisted/twisted/pull/1349

PR #1349, opened 2020-07-11
Run mypy on GitHub actions.

adiroiban:
https://twistedmatrix.com/trac/ticket/9903

Scope
=====

This moves the mypy check from Azure Pipelines to GitHub Actions.

GitHub actions provides better integration with PR and at the same time can be managed without extra access to Azure DevOps admin portal.

Changes
=======

Remove mypy from Azure Pipelines and moves it to GitHub Actions

It tries to run all the setup code in separate steps so that mypy step will have minimal noise and we can see how long it takes to run only the mypy test, without installing deps and creating the virtualenv

How to test
=========

I have no idea how to test this.
The problem is that GitHub actions will run an action only if it is first merged in master.


.github/workflows/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+#
+# Available VMs
+# https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+#
+name: GHA-Python
+
+# Only run on pull requests and only for Python files.
+on:
+  push:
+    branches: [ master ]
```
rodrigc: Why specify `master` branch here?  Twisted doesn't use a `master` branch.

There is a fuller example here: https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#starting-with-the-python-workflow-template

rodrigc (changes requested):

.github/workflows/main.yaml
```diff
@@ -0,0 +1,31 @@
+#
+# GitHub actions Python based test runs.
+#
+# Available VMs
+# https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+#
+name: GHA-Python
```
rodrigc: Since this is using Github Actions, you don't need to have `GHA-` in the name.

adiroiban:
Yes. I am stupid :) Thanks for your reaview and your part2 PR :)

adiroiban:
Ah.. you PR is on top of this one. I think that you can just commit in this PR.
I will reopen this PR.

rodrigc:
@adiroiban can you look at https://github.com/twisted/twisted/pull/1346 ?  With that one-liner, all the mypy errors are gone

rodrigc (approved):
Now that https://github.com/twisted/twisted/pull/1346 has been merged, the mypy job produces 0 errors.
So I can think it is on to enable a mypy run on GitHub Actions and delete the mypy job from Azure Pipelines

rodrigc:
@adiroiban does this look ok to you now?  mypy run now passes on GitHub Actions

adiroiban:
Thanks @rodrigc for all your work. Much appreciated.

I went for installing the tox deps in a separate steps (so that we have a log with all the version) and execute the tests in tox quiet mode in another test.

In this way, the output of the test is as focused as possible,  without info about the installed dependencies.

Free free to revert to your version if you like.

I have also added `pull_request` as if you just push a commit, and then you create a PR, it will not re-run the checks.

rodrigc:
This looks fine.

README.rst
```diff
@@ -111,3 +112,6 @@ Again, see the included `LICENSE <LICENSE>`_ file for specific legal details.
 
 .. |circleci| image:: https://circleci.com/gh/twisted/twisted.svg?style=svg
 .. _circleci: https://circleci.com/gh/twisted/twisted
+
+.. |mypy| image:: https://github.com/twisted/twisted/workflows/mypy.yaml/badge.svg
```
adiroiban, line 116: @rodrigc 
The URL should be `https://github.com/twisted/twisted/workflows/mypy/badge.svg`

This can be changed as drive by with the next PR to be merged.

Thanks!

adiroiban:
Great work @rodrigc

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,48 @@
...
+name: mypy
+
+on:
+  push:
+    branches: [ trunk ]
+  pull_request:
+    branches: [ trunk ]
+
+jobs:
+  test_run:
```
adiroiban, line 18: this should be renamed `mypy_test_run` as otherwise the configuration inside GitHub protected branches is confusing.

This might be a bug in GitHub, as instead of `mypy/test_run` the branch protection is visible as just `test_run`.

adiroiban:
@glyph FYI  I have updated  the GitHub Protected Branch sections to require a green mypy test before a merge in trunk.

glyph:
> @glyph FYI I have updated the GitHub Protected Branch sections to require a green mypy test before a merge in trunk.

would you mind shooting a message to the mailing list for this?

PR #1350, opened 2020-07-11
#1191 take 2

glyph:
Supersedes #1191 .

## Commit Message

```
Author: mathieui, glyph

Reviewer: 

Fixes: ticket:9716

Fixed serialization of timedelta, date, and time objects in twisted.spread.
```

Unicode strings cannot be sent as is in Banana, and need to be
byte-encoded. In python 3, this would fail for the date, time and
timedelta objects as the jellyfication would use a unicode string and
not a byte string.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9716


rodrigc (approved):
Thanks to @mathieui for the initial work, and to @glyph for picking this up and continuing it.
It looks like this combination slipped through the cracks, so thanks for adding a testcase.

At some point, I think removing `twisted.python.compat.unicode` from this file and replacing it with `str` would be OK, but that doesn't need to be done now.
In a Python 3-only world for Twisted we don't need to do the brain flips between unicode/str/bytes representation on Python 2 vs. 3 any more. 😀

PR #741, opened 2017-03-11
[9076] Remove CE chunked header from t.w.http CONNECT response

izzy:
According to [RFC 7231 Section 4.3.6](https://tools.ietf.org/html/rfc7231#page-31) "A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response to CONNECT"

Fix for [Trac ticket 9076](https://twistedmatrix.com/trac/ticket/9076#ticket)

PR #1358, opened 2020-07-19
#9076 HTTP CONNECT chunked Content-Length suppression

### LINKED ISSUE #9076: t.w.http must not set the Transfer-Encoding header for CONNECT requests
|[mkzero's avatar](https://github.com/mkzero)| @mkzero reported|
|-|-|
|Trac ID|trac#9076|
|Type|defect|
|Created|2017-03-11 10:35:04Z|
|Branch|https://github.com/twisted/twisted/tree/9076-mkzero-t-w-web-connect-compatibility|

According to [RFC 7231 Section 4.3.6](https://tools.ietf.org/html/rfc7231#page-31) "A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response to CONNECT".

Using t.w.http and thus t.w.proxy with connect requests does not adhere to this, so projects inheriting t.w.proxy for CONNECT proxies will break compatibility with this RFC as t.w.http does send these headers.

```
trac-id__9076 9076
type__defect defect
reporter__mkzero mkzero
priority__normal normal
milestone__None None
branch__9076_mkzero_t_w_web_connect_compatibility 9076-mkzero-t-w-web-connect-compatibility
branch_author__ 
status__new new
resolution__None None
component__web web
keywords__None None
time__1489228504317618 1489228504317618
changetime__1602406619350621 1602406619350621
version__None None
owner__twm twm

```
</details>

twm:
This is #741 merged forward. I couldn't resolve the conflicts on that PR directly, perhaps because the original author has deleted their fork.

Fixes #9076.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9076


rodrigc:
This looks OK, but lint failures need to be fixed: https://travis-ci.com/github/twisted/twisted/jobs/362488290

You can use
`tox -e lint` 
locally before pushing.

twm:
@rodrigc It wasn't actually possible to make this pass lint by only changing formatting before #1365 because pycodestyle-twisted had both of these warnings enabled:

* W504 line break after binary operator
* W503 line break before binary operator

I did _try_, but I wasn't looking to make semantic changes to this PR, just shepherd it through the process.

src/twisted/web/http.py
```diff
@@ -1151,6 +1154,11 @@ def write(self, data):
 
             for name, values in self.responseHeaders.getAllRawHeaders():
                 for value in values:
+                    if (self.method == b"CONNECT" and self.code // 100 == 2 and
+                        name == b"Content-Length"):
+                        self._log.info("Warning: Removing Content-Length"
```
glyph: 1. Can we include some indication of *which* request as a parameter to this message?  Include the URL, or a reference to which render method was invoked, or both?  This log message should be actionable; which method set this header, so I can go fix it?
2. If this is a warning, don't say `Warning:`, just use `log.warn` :)  (This will also help make it more actionable, since if we're filtering out `info`-level messages that don't need to be acted upon, we'll still see it.)
3. can you test the emission of this log message?
4. can you include a reference to the relevant RFC here?

src/twisted/web/http.py
```diff
@@ -1129,9 +1129,12 @@ def write(self, data):
             # if we don't have a content length, we send data in
             # chunked mode, so that we can support pipelining in
             # persistent connections.
+            # For no-body responses and CONNECT requests this
+            # does not apply and the header must not be sent.
             if ((version == b"HTTP/1.1") and
                 (self.responseHeaders.getRawHeaders(b'content-length') is None) and
-                self.method != b"HEAD" and self.code not in NO_BODY_CODES):
+                self.method != b"HEAD" and self.code not in NO_BODY_CODES and
+                (self.method != b"CONNECT" and self.code // 100 == 2)):
```
glyph: Hmm... is there any other way that `transfer-encoding` headers could get set?  Possibly beyond the scope of this PR, but it seems like we should validate that somehow as well.

(Just aesthetically, I'd rather `200 <= self.code < 300` than `// 100 == 2`)

glyph: (Also, should this just be `== 200`? What would it even mean to respond to `CONNECT` with 201 Created or 204 No Content?)

src/twisted/web/http.py
```diff
@@ -1185,6 +1185,7 @@ def write(self, data):
                 and (self.responseHeaders.getRawHeaders(b"content-length") is None)
                 and self.method != b"HEAD"
                 and self.code not in NO_BODY_CODES
+                and (self.method != b"CONNECT" and self.code // 100 == 2)
```
glyph: "successful connect" is expressed twice here; it would be nice to unify that into a helper function or private method.

src/twisted/web/http.py
```diff
@@ -1205,6 +1206,16 @@ def write(self, data):
 
             for name, values in self.responseHeaders.getAllRawHeaders():
                 for value in values:
+                    if (
+                        self.method == b"CONNECT"
+                        and self.code // 100 == 2
+                        and name == b"Content-Length"
```
glyph: The `name ==` comparison should be executed first so that the slightly more expensive retrieve-an-attribute / compare a string, whereas `name==` is just a local variable comparison which is slightly cheaper and we only care about doing this for one header.

src/twisted/web/http.py
```diff
@@ -1205,6 +1206,16 @@ def write(self, data):
...
+                    ):
+                        self._log.info(
+                            "Warning: Removing Content-Length"
+                            " header in response to CONNECT request"
```
glyph: Should we log the request here, or its URL?  Anything useful to include in the format string here?

glyph (approved):
If you could do the minor cleanups inline, go ahead and land :)

graingert:
> W504 line break after binary operator
> W503 line break before binary operator

These should be handled by black and ignored in pycodestyle because we use it via flake8 extend-ignore

src/twisted/web/http.py
```diff
@@ -1236,6 +1239,13 @@ def write(self, data):
 
             for name, values in self.responseHeaders.getAllRawHeaders():
                 for value in values:
+                    if name == b"Content-Length" and isOkay and isCONNECT:
```
glyph, line 1230: Is there any case-sensitivity issue here?

glyph: It occurs to me that filtering these headers might be better-accomplished by doing a pre-processing pass here along the lines of:

1. putting `b"Content-Length"` into a constant of some kind, it is wild that we still use literals everywhere
2. doing this:
```
self.responseHeaders.removeHeader(CONTENT_LENGTH)
```

If we want to still emit a warning rather than silently filter it, we could include a `hasHeader` too.

glyph (commented):
I'm still +1 here, but I think we should confirm that case variations don't cause issues.

src/twisted/web/http.py
adiroiban: This should be ok.

```python
    def getAllRawHeaders(self) -> Iterator[Tuple[bytes, Sequence[bytes]]]:
        """
        Return an iterator of key, value pairs of all headers contained in this
        object, as L{bytes}.  The keys are capitalized in canonical
        capitalization.
        """
        for k, v in self._rawHeaders.items():
            yield self._canonicalNameCaps(k), v
```

src/twisted/web/http.py
```diff
@@ -1236,6 +1239,13 @@ def write(self, data):
...
+                        self._log.info(
+                            "Warning: Content-Length header was"
+                            " sent in response to CONNECT request."
+                            " Dropping it."
```
adiroiban: Maybe I don't understand what is going in here :)

My understanding is that the header was received by us / the Twisted based app.

It was sent by the remote HTTP proxy server.  

```suggestion
                            "Warning: Content-Length header was"
                            " received in response to CONNECT request."
                            " Dropping it."
```

adiroiban:
Is the new test good enough?

If I revert this change and then run the test `$ trial twisted.web.test.test_http.ConnectRequestTests`, the test still passes

```diff
diff --git a/src/twisted/web/http.py b/src/twisted/web/http.py
index 1cc8ce038f..91a3f13790 100644
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1218,7 +1218,6 @@ class Request:
                 and (self.responseHeaders.getRawHeaders(b"content-length") is None)
                 and self.method != b"HEAD"
                 and self.code not in NO_BODY_CODES
-                and (not isCONNECT and isOkay)
             ):
                 headers.append((b"Transfer-Encoding", b"chunked"))
                 self.chunked = 1
```

src/twisted/web/http.py
glyph: Great, thank you for tracking that down. Comment withdrawn :)

src/twisted/web/http.py
glyph: Just corrected the tense in the suggestion, but I don't think it's necessary inaccurate - somebody's always sending, someone's receiving.

twm:
Circling back here, I've taken another look at #9076. Here's my understanding:

- The reporter implemented an HTTP proxy that handles the CONNECT method (I don't see any implementation of this in Twisted).
- The [CONNECT method](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6) MUST NOT send a Transfer-Encoding or Content-Length header, because they are nonsense in the context of hijacking the connection.
- `twisted.web.http.Request.write` will chunk the response unless *Content-Length* is set (setting the forbidden *Transfer-Encoding* header in the process).
- As chunking would mangle the CONNECT bytestream, presumably the reporter sets `Content-Length` header to avoid chunking. This should work ("A client MUST ignore any Content-Length or Transfer-Encoding header fields received in a successful response to CONNECT.") but violates a prohibition ("A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response to CONNECT.").
- The reporter would prefer a standards-compliant response without the *Content-Length* header.

Thus this PR attempts to implement:

1. Twisted should not add *Content-Length* nor *Transfer-Encoding* automatically in the CONNECT 2xx case (nor should it do the actual chunking).
2. Twisted should suppress the *Content-Length* header, if present, and log a warning.

The first item seems straightforwardly correct to me. However, suppressing *Content-Length* seems wrong because we're masking a bug in user code (and adding log chatter). Ideally we'd raise an exception &mdash; it's a bug!

Unfortunately that would break working code, so I think the correct course of action in this PR is:

1. Add support for the CONNECT 2xx special case.
2. Deprecate passing through *Content-Length* and *Transfer-Encoding* headers in that special case.
3. File a follow-up to change the deprecation warning to an exception.

I don't think that the gross violation of specifications exception applies here because compliant clients will ignore these headers.

src/twisted/web/http.py
```diff
@@ -1224,6 +1227,13 @@ def write(self, data):
 
             for name, values in self.responseHeaders.getAllRawHeaders():
                 for value in values:
+                    if name == b"Content-Length" and isOkay and isCONNECT:
+                        self._log.info(
```
adiroiban: as commented by Tom,

I think is best to have this for a separate PR/ticket and raise a deprecation warning rather than a log warning.

and in the future raise an exception.

Or maybe just raise an exception right away as this can be considered a bug :)

adiroiban (changes requested):
I think that is best to focus this PR on the fix of not returning `ransfer-Encoding` for the CONNECT response.

We can look at chunk encoding + content type fix in a separate ticket.

@twm if you prefer, I can take over this PR and implement the suggested changes

adiroiban:
I am trying to get this merged before doing a release...

I will continue working on this... I was interrupted by the flaky tests and I tried  to fix them here https://github.com/twisted/twisted/pull/12019

adiroiban:
I have removed the release blocker as I have only marked it for myself as a reminder.
but we can proceed with the next release without this merged.

PR #1360, opened 2020-07-19
Add type annotations to __repr__() and __str__()

mthuurne:
The first two commits are scripted replacements that annotate all `__repr__()` and `__str__()` methods in the source tree. Since an annotated method has its body checked by mypy, this triggers a few error messages for those bodies that need to be fixed manually, which is what the remainder of the commits are for. Please read the individual commit comments for details.

There are two significant behavioral changes:
- `twisted.names.dns.RRHeader` now raises `ValueError` instead of `AssertionError` if the payload type does not match
- descriptive error messages from `twisted.internet.error` are now present when running with `python -OO`

I added a news fragment for each of them.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9918

### DIFF
--- a/src/twisted/python/usage.py
+++ b/src/twisted/python/usage.py
@@ -547,7 +547,7 @@ class Options(dict):
 
         return s + longdesc + commands
 
-    #def __repr__(self):
+    #def __repr__(self) -> str:
     #    XXX: It'd be cool if we could return a succinct representation
     #        of which flags and options are set here.
 
--- a/src/twisted/python/usage.py
+++ b/src/twisted/python/usage.py
@@ -547,9 +547,6 @@ class Options(dict):
 
         return s + longdesc + commands
 
-    #def __repr__(self) -> str:
-    #    XXX: It'd be cool if we could return a succinct representation
-    #        of which flags and options are set here.
 
 
 _ZSH = 'zsh'
--- a/src/twisted/python/usage.py
+++ b/src/twisted/python/usage.py
@@ -446,7 +446,7 @@ class Options(dict):
         return longOpt, shortOpt, docs, settings, synonyms, dispatch
 
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.getSynopsis() + '\n' + self.getUsage(width=None)
 
     def getSynopsis(self):
--- a/src/twisted/runner/procmon.py
+++ b/src/twisted/runner/procmon.py
@@ -409,7 +409,7 @@ class ProcessMonitor(service.Service):
             self.stopProcess(name)
 
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         l = []
         for name, proc in self._processes.items():
             uidgid = ''
--- a/src/twisted/runner/procmon.py
+++ b/src/twisted/runner/procmon.py
@@ -410,7 +410,7 @@ class ProcessMonitor(service.Service):
 
 
     def __repr__(self) -> str:
-        l = []
+        lst = []
         for name, proc in self._processes.items():
             uidgid = ''
             if proc.uid is not None:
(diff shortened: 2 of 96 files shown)


src/twisted/python/usage.py
```diff
@@ -547,9 +547,10 @@ def getUsage(self, width=None):
 
         return s + longdesc + commands
 
-    #def __repr__(self):
```
rodrigc, line 550: Not a big deal but maybe better just to delete this commented out block

src/twisted/runner/procmon.py
```diff
@@ -420,7 +420,7 @@ def __repr__(self):
 
             if uidgid:
                 uidgid = '(' + uidgid + ')'
-            l.append('%r%s: %r' % (name, uidgid, proc.args))
```
rodrigc, line 423: Not a big deal, but consider using format strings

src/twisted/web/microdom.py
```diff
@@ -123,10 +123,11 @@ def __init__(self, filename, expect, got, endLine, endCol, begLine, begCol):
          self.endCol) = filename, expect, got, begLine, begCol, endLine, endCol
 
 
-    def __str__(self):
-        return ("expected </%s>, got </%s> line: %s col: %s, began line: %s col: %s"
```
rodrigc, line 127: consider using format strings

src/twisted/web/sux.py
```diff
@@ -71,9 +71,11 @@ def __init__(self, filename, line, col, message):
         self.col = col
         self.message = message
 
-    def __str__(self):
-       return "%s:%s:%s: %s" % (self.filename, self.line, self.col,
```
rodrigc, line 75: Consider using format strings

src/twisted/names/newsfragments/9918.misc
```diff
@@ -0,0 +1 @@
+New interface IEncodableRecord combines IEncodable and IRecord, which is useful when using type annotations.
```
rodrigc: Newsfragments which are misc need to be empty.  See:

https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles

Can you separate out the changes where you introduce a IEncodableRecord interface into a new PR, and leave this .misc file as empty?

Everything else in this PR is very simple in terms of adding return types to `__repr__()` and are easy to merge in.

rodrigc (changes requested):

src/twisted/names/newsfragments/9918.misc
mthuurne: Done: `IEncodableRecord` and the other `dns` changes were split off to #1362

src/twisted/python/usage.py
mthuurne: Will do.

src/twisted/runner/procmon.py
mthuurne: I'd rather do that when we drop Python 3.5 support, so we can go straight to using f-strings.

mthuurne:
The Travis run failed because the coveralls.io API was over capacity. It has a "rerun" link, but it doesn't actually rerun the test. Given that it passed on an earlier version of this PR and all tests passed on Windows and macOS, I don't expect the Linux tests would be any different.

rodrigc:
If coveralls fails, don't worry about it.  Sometimes coveralls fails randomly, and there is nothing we can do about it.
I have enough access to re-trigger builds on Travis, so I can do that if need be.

mthuurne:
I don't mind the failed check if it doesn't stand in the way of the review.

src/twisted/test/test_failure.py
```diff
@@ -535,10 +535,10 @@ def test_repr(self):
 
 class BrokenStr(Exception):
     """
-    An exception class the instances of which cannot be presented as strings via
-    C{str}.
+    An exception class the instances of which cannot be presented as strings
+    via C{str}.
```
rodrigc: Should be L{str} in doc-string

rodrigc (approved):
One minor comment for docstring, but otherwise looks OK.

src/twisted/test/test_failure.py
mthuurne: I added a commit to fix this, but maybe we should do a dedicated PR to fix links to builtins in the entire source tree, because I think this is a problem that happens a lot.

PR #1361, opened 2020-07-20
 Add type annotations to __eq__() et al

mthuurne:
Some of the existing code was incorrect or outdated when passed an uncomparable value: they would return `False` or raise `TypeError`. I updated those to return `NotImplemented` instead.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9919

### DIFF
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -946,7 +946,7 @@ class Key(object):
         """
         Return True if other represents an object with the same key.
         """
-        if type(self) == type(other):
+        if isinstance(other, Key):
             return self.type() == other.type() and self.data() == other.data()
         else:
             return NotImplemented
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -942,7 +942,7 @@ class Key(object):
         """
         self._keyObject = keyObject
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         """
         Return True if other represents an object with the same key.
         """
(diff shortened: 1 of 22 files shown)


src/twisted/conch/ssh/keys.py
```diff
@@ -942,24 +942,15 @@ def __init__(self, keyObject):
         """
         self._keyObject = keyObject
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
```
rodrigc, line 945: The return type is specified as `bool`, but this function can also return `NotImplemented`.
Is it correct to specify `bool` as the return type?

mthuurne: I think this is handled as a special case by mypy. In typeshed, which is probably the closest thing we have to an authority on this, `bool` is also used as the return type for `__eq__`.

rodrigc: Yes you are correct.  In mypy, `__eq__()` is part of `BINARY_MAGIC_METHODS`: https://github.com/python/mypy/blob/23dba19098857a72e0ad2284b164e2fceaee44ac/mypy/sharedparse.py#L55

 which is checked here: https://github.com/python/mypy/blob/259e0cfb65b656ce0847be7c41a3a12f7c8b00a0/mypy/checker.py#L3201

and with a specific exception for `builtins.NotImplemented` here:  https://github.com/python/mypy/blob/259e0cfb65b656ce0847be7c41a3a12f7c8b00a0/mypy/checker.py#L5053

rodrigc (approved):
This looks OK

PR #1363, opened 2020-07-22
 Drop obsolete private definitions from twisted.python.compat

mthuurne:
Many of the definitions in `twisted.python.compat` have become obsolete now that Python 2.7 support has been dropped. For private definitions, that means we can drop them as soon as Twisted itself no longer needs them.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9921
* [x] The changes pass minimal style checks (see: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted )
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )
* [x] I have updated the automated tests.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.

PR #1364, opened 2020-07-23
Deprecate obsolete public definitions from twisted.python.compat

mthuurne:
This is the public counterpart to #1363. There are a lot of definitions that were useful in the transition to Python 3 but no longer serve a purpose.

Twisted can stop using these internally right now. In case there are external users, we should deprecate them, so we can remove the obsolete definitions at some point in the future.

For another PR:
- clean up uses of `str` and `unicode` in docstrings: there are still a lot of docstrings that use `str` where they mean `bytes`
- `networkString()` and `nativeString()` are probably obsolete, but having them in the code could help when adding type annotations, so I'd rather remove them at a later stage

# Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9922


rodrigc:
If you are going to merge things into Twisted, and ignore review feedback from me, please don't ask me to review your PR's.

mthuurne:
I asked for your review of this PR because it addresses the removal of uses of `long`, `unicode`, `iteritems()` etc. that you requested in #1363. I'm sorry if I hit the "merge" button on that PR too soon, but I assure you that I'm not ignoring your feedback.

src/twisted/test/test_process.py
```diff
@@ -2266,7 +2266,7 @@ def test_stdinReader_bytesArgs(self):
         if os.supports_bytes_environ:
             env = dict(os.environb)
         else:
-            env = bytesEnviron()
+            env = dict(os.environ)
```
mthuurne: Since `os.supports_bytes_environ` is always false on Windows and this test is skipped on non-Windows platforms, does it make sense to have an `if` statement here? It suggests the test is more flexible than it actually is.

It also obfuscates the fact that all OS-filled entries of the `env` dictionary will be of type `str`; only the `PYTHONPATH` entry that the test adds will be in `bytes`.

src/twisted/internet/test/test_process.py
```diff
@@ -582,7 +582,7 @@ def processExited(self, reason):
         # Only required if the test fails on systems that support
         # the process module.
         if process is not None:
-            for pid, handler in items(process.reapProcessHandlers):
+            for pid, handler in list(process.reapProcessHandlers.items()):
```
wsanchez, line 585: Why `list()`?

mthuurne: `unregisterReapProcessHandler()` executes `del reapProcessHandlers[pid]`, which modifies the dictionary we're iterating over and `dict` does not support that.

_Edit:_ Although in this case, `unregisterReapProcessHandler()` is followed by `self.fail()`, so the iteration doesn't continue. So `list()` is indeed not necessary here. Should I remove it or leave it in case the loop body ever changes?

src/twisted/test/test_process.py
wsanchez: I agree with nixing the `if` statement

wsanchez (approved):
OK.

This is obviously a large diff, and it would have been be a lot easier to review if you did separate PRs for each type of change (`unicode` ➜ `str`, `long` ➜ `int`, etc.).  Please consider that next time.

src/twisted/internet/test/test_process.py
wsanchez: Since it's a test… meh

mthuurne: I'll leave in the `list()` then: in a test it is more important to be obviously correct than to be efficient.

src/twisted/test/test_process.py
mthuurne: Done.

mthuurne:
> This is obviously a large diff, and it would have been be a lot easier to review if you did separate PRs for each type of change (`unicode` ➜ `str`, `long` ➜ `int`, etc.). Please consider that next time.

I had considered that, but that would have introduced another problem: because the imports are close together, I would either have to include all previous PRs in each PR, which would complicate reviewing, or base all PRs on trunk, in which case there would be a lot of merge conflicts. Also updating 38 PRs when trunk changes would be a lot of work and trigger a lot of redundant CI activity. So I think it would have been messy either way.

What I perhaps should have done is suggest a per-commit review, since that simplifies the review process while only requiring a single merge at the end.

In any case, thanks for reviewing this PR despite its size.

PR #1367, opened 2020-07-25
Drop obsolete private definitions from twisted.python.compat

mthuurne:
Many of the definitions in `twisted.python.compat` have become obsolete now that Python 2.7 support has been dropped. For private definitions, that means we can drop them as soon as Twisted itself no longer needs them.

I merged #1363 when I thought all review comments had been addressed in #1364, but later learned that the reviewer didn't agree. I am sorry for not asking explicitly whether it was OK to merge.

I do still think that this PR in its current form contains all the changes needed for its scope, which is to drop obsolete private definitions. Removing the uses of public definitions from `compat` belongs in #1364 instead.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9921

### DIFF
--- a/src/twisted/conch/ssh/connection.py
+++ b/src/twisted/conch/ssh/connection.py
@@ -17,8 +17,7 @@ from twisted.conch.ssh import service, common
 from twisted.conch import error
 from twisted.internet import defer
 from twisted.python import log
-from twisted.python.compat import (
-    nativeString, networkString, long, _bytesChr as chr)
+from twisted.python.compat import nativeString, networkString, long
 
 
 
@@ -444,10 +443,12 @@ class SSHConnection(service.SSHService):
         if channel.localClosed:
             return
         log.msg('sending request %r' % (requestType))
-        self.transport.sendPacket(MSG_CHANNEL_REQUEST, struct.pack('>L',
-                                    self.channelsToRemoteChannel[channel])
-                                  + common.NS(requestType)+chr(wantReply)
-                                  + data)
+        self.transport.sendPacket(
+                MSG_CHANNEL_REQUEST,
+                struct.pack('>L', self.channelsToRemoteChannel[channel])
+                + common.NS(requestType)
+                + (b'\1' if wantReply else b'\0')
+                + data)
         if wantReply:
             d = defer.Deferred()
             self.deferreds.setdefault(channel.id, []).append(d)
@@ -647,6 +648,5 @@ for name, value in locals().copy().items():
         messages[value] = name  # Doesn't handle doubles
 
 alphanums = networkString(string.ascii_letters + string.digits)
-TRANSLATE_TABLE = b''.join([chr(i) in alphanums and chr(i) or b'_'
-                            for i in range(256)])
+TRANSLATE_TABLE = bytes(i if i in alphanums else ord('_') for i in range(256))
 SSHConnection.protocolMessages = messages
--- a/src/twisted/conch/telnet.py
+++ b/src/twisted/conch/telnet.py
@@ -15,9 +15,24 @@ from zope.interface import implementer
 
 from twisted.internet import protocol, interfaces as iinternet, defer
 from twisted.python import log
-from twisted.python.compat import _bytesChr as chr, iterbytes
+from twisted.python.compat import iterbytes
 
-MODE = chr(1)
+
+
+def _chr(i: int) -> bytes:
+    """Create a byte sequence of length 1.
+
+    U{RFC 854<https://tools.ietf.org/html/rfc854>} specifies codes in decimal,
+    but Python can only handle L{bytes} literals in octal or hexadecimal.
+    This helper function bridges that gap.
+
+    @param i: The value of the only byte in the sequence.
+    """
+    return bytes((i,))
+
+
+
+MODE = _chr(1)
 EDIT = 1
 TRAPSIG = 2
 MODE_ACK = 4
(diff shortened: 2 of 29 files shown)


src/twisted/conch/ssh/connection.py
```diff
@@ -444,10 +443,12 @@ def sendRequest(self, channel, requestType, data, wantReply=0):
...
         log.msg('sending request %r' % (requestType))
-        self.transport.sendPacket(MSG_CHANNEL_REQUEST, struct.pack('>L',
-                                    self.channelsToRemoteChannel[channel])
-                                  + common.NS(requestType)+chr(wantReply)
-                                  + data)
+        self.transport.sendPacket(
+                MSG_CHANNEL_REQUEST,
+                struct.pack('>L', self.channelsToRemoteChannel[channel])
+                + common.NS(requestType)
+                + bytes([wantReply])
```
twm: It seems like we could avoid the list allocation and function call here pretty easily.

```suggestion
                + b'\x01' if wantReply else b'\x00'
```

src/twisted/conch/telnet.py
```diff
@@ -15,7 +15,14 @@
 
 from twisted.internet import protocol, interfaces as iinternet, defer
 from twisted.python import log
-from twisted.python.compat import _bytesChr as chr, iterbytes
+from twisted.python.compat import iterbytes
+
+
+
+def chr(i):
```
twm: :warning: This is new public API... I don't think we want this?

twm (commented):
I don't have time for a full review, but I'm not sure that the `bytes([chr])` idiom is the right one. It looks like a tuple is ~4x as fast:

```
In [1]: %timeit bytes([123])                                                                                                                                                 
135 ns ± 2.19 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [8]: %timeit bytes((123,))                                                                                                                                                
112 ns ± 0.411 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I also tried [`int.to_bytes()`](https://docs.python.org/3/library/stdtypes.html#int.to_bytes) but it's not any faster than `byte([chr]`.

mthuurne:
> I don't have time for a full review, but I'm not sure that the `bytes([chr])` idiom is the right one.

`bytes([chr])` is what the original implementation of `_bytesChr()` used. I'm not particularly attached to it.

> It looks like a tuple is ~4x as fast:
> 
> ```
> In [1]: %timeit bytes([123])                                                                                                                                                 
> 135 ns ± 2.19 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> 
> In [8]: %timeit bytes((123,))                                                                                                                                                
> 112 ns ± 0.411 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> ```

Eh, the mean is about 15% lower (17% in your measurement, 15% on my PC), it's the deviation that is over 4x lower.

However, since the code is equally readable, I do think it's worth using tuples instead. I'll update the PR.

src/twisted/conch/telnet.py
mthuurne: Indeed, we don't want this.

The original code did this:
```py
from twisted.python.compat import _bytesChr as chr
```
So I thought defining `chr()` locally would be the minimal change. However, an imported public name that is not in `__all__` isn't considered API while a local definition is. I'll rename it to `_chr()`.

In most other modules I replaced constant values by byte literals, but in the case of telnet [the RFC](https://tools.ietf.org/html/rfc854) specifies codes in decimal and there doesn't seem to be a way to do byte literals in decimal. Using byte literals in hexadecimal (`BRK = b'\xF3'`) would be more efficient at runtime, but would also make maintenance of the code harder.

src/twisted/conch/ssh/connection.py
mthuurne: That needs some parentheses, otherwise the operator precedence breaks things.

If we care about performance, the use of `+` over `bytes.join()` might be a bigger issue here. Also structs could be precompiled, so we wouldn't have to gamble that the built-in struct cache is large enough. But unlike those, what you suggested is a pretty small change, so I could make it if you prefer.

mthuurne:
I updated the `Remove _bytesChr() from compat` commit to not use `bytes([...])` and made the helper function in `telnet.py` private.

twm:
>> It looks like a tuple is ~4x as fast:
>>
>> ...
>
> Eh, the mean is about 15% lower (17% in your measurement, 15% on my PC), it's the deviation that is over 4x lower.

😳 I am embarrassed to have misread that so badly. Thank you for humoring me.

src/twisted/conch/telnet.py
```diff
@@ -15,9 +15,16 @@
 
 from twisted.internet import protocol, interfaces as iinternet, defer
 from twisted.python import log
-from twisted.python.compat import _bytesChr as chr, iterbytes
+from twisted.python.compat import iterbytes
 
-MODE = chr(1)
+
+
+def _chr(i: int) -> bytes:
```
twm, line 22: A docstring is required here. The explanation about the spec using decimal in your earlier comment would be good content!

src/twisted/words/test/test_domish.py
```diff
@@ -181,14 +181,6 @@ def test_addContentNativeStringASCII(self):
         self.assertEqual("native", str(element))
 
 
-    def test_addContentBytes(self):
```
twm: Are we generally planning to replace runtime type checks with type annotations? That's a behavioral change that will make debugging more difficult for anyone not using typing/MyPy. I don't think that this sort of change should occur in a .misc PR: it at least merits a release note. I'd prefer that `addElement` grow an `if not isinstance(text, str): raise TypeError` check.

src/twisted/words/xish/domish.py
```diff
@@ -511,9 +510,8 @@ def addChild(self, node):
         self.children.append(node)
         return node
 
-    def addContent(self, text):
+    def addContent(self, text: str) -> str:
```
twm, line 513: :+1:

src/twisted/words/protocols/jabber/client.py
```diff
@@ -115,7 +115,7 @@ def initialize(self):
 
     def _cbAuthQuery(self, iq):
         jid = self.xmlstream.authenticator.jid
-        password = _coercedUnicode(self.xmlstream.authenticator.password)
```
twm, line 118: Reviewer note: assuming that the type assertion in `addElement` is restored this is not a behavioral change. This value is either passed to `addElement` or `hashPassword()`, which raises `TypeError` for non-`str` passwords.

twm (changes requested):
This looks good, thank you! I noted some minor changes to make before it can be landed.

src/twisted/conch/telnet.py
mthuurne: I'll add one.

src/twisted/words/test/test_domish.py
mthuurne: Typing is considered optional for Python as a language, but I don't think it should be considered optional for Twisted as a project. Since Twisted has mypy as one of the checks in CI, developers working on Twisted cannot ignore mypy.

That said, there are two scenarios where mypy cannot detect type errors:
- when the calling code inside Twisted isn't annotated yet
- when a project using Twisted isn't using mypy

Over time, the first category should shrink to zero, but that may take a while. I think the latter category is also likely to shrink over time, but that is a theory that is unproven at this time. Right now the runtime type check still has some value, so I'll re-add it.

mthuurne:
- The commit `Remove test for domish.Element.addContent() raising TypeError` was dropped.
- `Element.addContent()` once again raises `TypeError` when called with a `bytes` argument.
- A docstring was added to `_chr()` in `src/twisted/conch/telnet.py`.

twm (approved):
Looks great, thank you!

PR #1368, opened 2020-08-03
Drop explicit inheritance from 'object' class

mthuurne:
Inheriting from `object` was necessary in Python 2 to force new-style classes. In Python 3, it's just noise.

Substitutions were done automatically with `sed`, except:
- `twisted/words/xish/xpathparser.g` was done by hand and `xpathparser.py` was re-generated from that
- `twisted/test/test_plugin.py` was done by hand

I also had to update the definitions for `check-manifest` in `tox.ini`, since the old ignore pattern no longer works with the latest release of `check-manifest`.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9927

### DIFF
--- a/src/twisted/words/xish/xpathparser.py
+++ b/src/twisted/words/xish/xpathparser.py
@@ -28,7 +28,7 @@
 # 3.) Edit the output to depend on the embedded runtime, and remove extraneous
 #     imports:
 #
-#         sed -e '/^# Begin/,${/^[^ ].*mport/d}' -e '/^[^#]/s/runtime\.//g' \
+#         sed -e '/^# Begin/,${/^[^ ].*mport/d}' -e 's/runtime\.//g' \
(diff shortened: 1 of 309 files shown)


src/twisted/words/xish/xpathparser.py
```diff
@@ -28,7 +28,7 @@
 # 3.) Edit the output to depend on the embedded runtime, and remove extraneous
 #     imports:
 #
-#         sed -e '/^# Begin/,${/^[^ ].*mport/d}' -e '/^[^#]/s/runtime\.//g' \
```
wsanchez, line 31: Is this diff intentional?

wsanchez (changes requested):
Change to sed script in comment looks unrelated… and maybe incorrect?
`tox.ini` changes seem unrelated.  Since this is a big diff, it's probably best not to have riders.

wsanchez (approved):
Ah, never mind, I see the comments now about why

PR #1371, opened 2020-08-04
[9932] rename amp.String to amp.Binary with alias

graingert:
## Remove this section

Please have a look at [our developer documentation](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch) before submitting your Pull Request.

Please note that the trac ticket, news fragment, and review submission portions of this process apply to *all* pull requests, no matter how small; if you don't do them, it's likely that nobody will even notice your PR needs a review.

## Contributor Checklist:

* [ ] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/
* [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles )

### DIFF
--- a/src/twisted/protocols/amp.py
+++ b/src/twisted/protocols/amp.py
@@ -1465,9 +1465,9 @@ class Integer(Argument):
 
 
 
-class String(Argument):
+class Binary(Argument):
     """
-    Don't do any conversion at all; just pass through 'str'.
+    Don't do any conversion at all; just pass through 'bytes'.
     """
     def toString(self, inObject):
         return inObject
@@ -1475,6 +1475,8 @@ class String(Argument):
     def fromString(self, inString):
         return inString
 
+
+String = Binary
 
 
 class Float(Argument):


src/twisted/protocols/amp.py
```diff
@@ -1465,9 +1465,9 @@ def toString(self, inObject):
 
 
 
-class String(Argument):
+class Binary(Argument):
     """
-    Don't do any conversion at all; just pass through 'str'.
+    Don't do any conversion at all; just pass through 'bytes'.
```
twm, line 1470: I think that `'bytes'` should be a reference, like `C{bytes}`.

src/twisted/protocols/amp.py
```diff
@@ -1476,6 +1476,8 @@ def fromString(self, inString):
         return inString
 
 
+String = Binary
```
twm, line 1479: The alias needs documentation, if only `Compatibility alias of C{Binary}.`

twm (changes requested):
This needs some sort of test coverage. It would be best to update all the tests that reference `String` to use the new name of `Binary`. Then add a test with an assertion like ``self.assertIs(amp.String, amp.Binary)`` for the alias.

wiml:
In keeping with Python naming conventions, would it be better to name it `Bytes` instead of `Binary`?

twm:
We had a discussion about that on IRC. I am okay with either one, but I got the impression that @graingert prefers `Bytes` and will rename it.

PR #1377, opened 2020-08-22
[9509] Move PyHamcrest dependency to [test] extra

twm:
I'm not 100% sure of this approach, since the `dev` extra contains quite a few packages that are now installed into the `nodeps` Tox environments. Consider this a proposal for discussion. Other possible solutions:

1. Drop the linting and build tools from the ``dev`` extra. They are redundant with specifications in `tox.ini` anyway.

2. Add a new extra, e.g. ``test``, and put PyHamcrest there instead of `dev`.

3. Rewrite the tests that use PyHamcrest without it; drop the dependency entirely.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9509


glyph (dismissed):
> Add a new extra, e.g. test, and put PyHamcrest there instead of dev.

Let's go with this one. I don't want `nodeps` to include e.g. `sphinx` and then have us discover that Twisted doesn't actually work in a venv without `sphinx` installed.

I kinda hate test-only deps (since it's not possible to test the configuration where the test-deps are absent as they ostensibly would be in prod), but, I think that rewriting the tests to avoid pyhamcrest would make them noticeably worse, so let's stick with this for now.

twm:
I have merged forward. I also added the cython-test-exception-raiser to the `[test]` extra.

adiroiban (approved):
Thanks for this well written PR :) (with just one exception)

I don't have any better idea, so I think that we should approve it for now.

-------

I find `dev` and `test` confusing.

My understanding is that `test` is introduced to help with `nodeps` testing.

My understanding is that since we have nodeps testing as a "supported" test mode, all the tests were written in such a way that if the main deps is not found, the test is skipped.

With that in mind, the tests depending on cython exception raised and pyhamcrest should be skipped. 

PyHamcrest looks like a clean package, but in the future we might add another testing package with might depend on sphinx and then we will end up with sphinx in `nodeps` run.  

--------

I don't understand the first option: `Drop the linting and build tools from the dev extra. They are redundant with specifications in tox.ini anyway.`

-----------

I am +1 for "Rewrite the tests that use PyHamcrest without it; drop the dependency entirely." .... but I don't think there will be a person willing to spend time on it.
If I hit such a test, I just write a new one without using pyhamcrest... so maybe this can be a long term goal.

twm:
@adiroiban Thank you for the review!

> I don't understand the first option: `Drop the linting and build tools from the dev extra. They are redundant with specifications in tox.ini anyway.`

I mean that there are dependencies specified directly in tox.ini:

https://github.com/twisted/twisted/blob/f6980e302bf084acdabe0eaf03f43053b0e73f94/tox.ini#L70-L81

These are what's actually used in the various linting envs. The `dev` extra is only used to build the narrative documentation in `tox.ini`:

https://github.com/twisted/twisted/blob/f6980e302bf084acdabe0eaf03f43053b0e73f94/tox.ini#L56-L57

The `dev` extra seems to be intended for the [Without Tox](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#WithoutTox) workflow, and is also used by the Read the Docs build. We should probably extract a `docs` extra for the latter purpose, or even better generate a pinned requirements file with the full transitive dependencies.

That's all outside the scope of this PR, though. So thanks again, and I will merge.

adiroiban:
@twm thanks for the explanation. It makes sense.

My take is to remove dependency definition from tox and move everything into setup.cfg ... the main reason being that in this way you can develop twisted without tox.
When I do tdd for twisted I don't use tox, as tox is slow.

The second reason, is that in this wall all the dependencies are defined in a single place.

I think that a 'docs' or 'read the docs' dependency list will follow soon... I remember reviewing something related in a PR by @graingert ... but I don't remember which one.

I remember there was an ugly side effect of read the docs theme, trying to install PIL and making the Windows build complicated

PR #1380, opened 2020-09-12
[9512] Initial configuration for black, take two

twm:
This is a rebase of, and closes #1134. I removed the documentation changes that proved so contentious, but added a few necessary to reconcile with [The _Black_ Code Style](https://github.com/psf/black/blob/master/docs/the_black_code_style.md).

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9512


tox.ini
```diff
@@ -151,6 +148,25 @@ deps=https://github.com/twisted/pydoctor/archive/3f9c64829dfa040b334c9ae27c332c7
 [testenv:manifest-checker]
 skip_install = true
 
+[testenv:lint]
+skip_install=True
+basepython=python3.6
+
+[testenv:black]
+skip_install=True
+basepython=python3.6
```
rodrigc: In this PR, the basepython for running black is Python 3.6.
However, Python 3.5 support has not yet been dropped from Twisted.

Do you need to specify `--target` to black, to prevent style changes incompatible with Python 3.5 from going in?

I'm not in favor of pinning the tox environment for black to a specific Python version,
because a Twisted contributor may have a different version of Python installed on their system,
and `tox -e black` will not work for them.

tox.ini
```diff
@@ -151,6 +148,25 @@ deps=https://github.com/twisted/pydoctor/archive/3f9c64829dfa040b334c9ae27c332c7
 [testenv:manifest-checker]
 skip_install = true
 
+[testenv:lint]
+skip_install=True
+basepython=python3.6
```
rodrigc: I'm not in favor of pinning the lint target to Python 3.6,
because `tox -e lint` will not work for contributors who do not have Python 3.6 installed.

tox.ini
```diff
@@ -65,9 +65,8 @@ deps =
     twine: twine
 
     lint: pyflakes
-    lint: twistedchecker>=0.7.4
-    lint: diff-cover==0.9.12
-    lint: pycodestyle
+
+    black: black>=19.3b0
```
rodrigc: Further down in the `deps` section, you have: `black==20.8b1`.
Might be good to keep this consistent

rodrigc (commented):
The old pycodestyle target would fail in CI if someone submitted a PR,
but it did not pass pycodestyle changes.

If someone submits a PR, and it is not formatted with black,
will there be a CI failure, so that the user will know that they have to reformat their patch with black,
with `tox -e black`?

rodrigc (changes requested):
Reference to `pycodestyle` should be removed from `README.rst`.

Might want to add reference to `tox -e black` there.

tox.ini
twm: I believe that the `target-version` in `pyproject.toml` ensures that 3.5-compatible syntax is used.

tox.ini
twm: I don't really understand the motivation for this change myself. Perhaps the goal was to prevent linting tools from running under Python 2.7 if Tox was installed with that Python version? I'll remove it.

tox.ini
twm: Thanks, I'll clean this up.

tox.ini
rodrigc: I think you are correct.  That's good enough then.

rodrigc:
If a user files a PR, and their PR does not match black's coding style,
how will they, or a code reviewer know that they need to run `tox -e black` to fix things up?

twm:
@rodrigc Thank you for the review! I believe that I have addressed all your comments.

# Changes

I consolidated the Tox configuration to have a factor-based style throughout. This seemed to be most similar to what was already there. I'd contrast the factor-based style with the [explicit `[testenv:foo]` style used in Klein](https://github.com/twisted/klein/blob/master/tox.ini). I did lift the `[default]` section technique for variables from Klein, though.

I removed some detritus for a `pyflakes` Tox environment. There is also some stuff for a `wheel` factor (not the same as the `wheels` factor!), but I'm not sure enough about that to remove it, and that's really out of scope for this PR, anyway.

I added a GitHub Actions workflow that runs the linting tools. I like GHA for this because it is lower-latency than Travis, though the log UI is more difficult to use. I also filed [#9953](https://twistedmatrix.com/trac/ticket/9953) to do more of this.

# Merge Procedure

There is a bit of instability in Black's output on one file that'll need to be worked around when running it ([filed upstream here](https://github.com/psf/black/issues/1629#issuecomment-691552345)), so the procedure on merge will be:

    tox -e black-reformat -- --fast src/twisted/words/test/test_irc.py
    tox -e black-reformat -- --fast src/twisted/words/test/test_irc.py
    tox -e black-reformat

Running the tool twice with `--fast` is the workaround suggested in the Black ticket I linked. I have verified that it works for this file.

Overall:

1. Run Black (as above)
2. Commit the result; push to `trunk`. No review is required because it is the output of a tool.
3. Create a PR that adds the above commit to `.git-blame-ignore-revs` and makes the lint GHA build required.

README.rst
```diff
@@ -70,13 +70,22 @@ Some of these tests may fail if you:
...
 
-  $ tox -e lint   # run pyflakes and pycodestyle to check code against coding standards
-  $ tox -e mypy   # run mypy static type checker to check for type errors
+  $ tox -e lint   # run pyflakes to check syntax
+  $ tox -e black  # run Black to check against coding standard
+  $ tox -e mypy   # run MyPy static type checker to check for type errors
+
+To automatically format code according to the coding standards::
+
+  $ tox -e black-reformat
```
rodrigc, line 84: Nice, this gives clear instructions on how to reformat the code.

docs/core/development/policy/coding-standard.rst
```diff
@@ -76,18 +76,19 @@ Indentation is 4 spaces per indent.
 Tabs are not allowed.
 It is preferred that every block appears on a new line, so that control structure indentation is always visible.
 
-Lines are flowed at 79 columns.
+Lines are flowed at 88 columns per `The Black Code Style <https://github.com/psf/black/blob/master/docs/the_black_code_style.md>`_.
```
rodrigc, line 79: This is OK.  In future, we may want to remove all mention of number of columns, etc.
from this document, and just say that Twisted follows the style guidelines as followed by Black, and just give a pointer Black's guidelines.  In the end, very few people will read this anyways.

As long as they are told during review time to run `tox -e black-reformat` to make their
PR acceptable, that is probably good enough.

rodrigc:
Might want to add the required invocation of black to the wiki here: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [ trunk ]
+  pull_request:
+    branches: [ trunk ]
+
+jobs:
+  lint:
+
+    runs-on: ubuntu-20.04
+    env:
+      TOXENV: "lint,black,manifest-checker,twine"
```
rodrigc, line 14: Adding this GitHub action which invokes black is good.

tox.ini
```diff
@@ -148,20 +165,10 @@ commands =
 [testenv:apidocs]
 deps=https://github.com/twisted/pydoctor/archive/3f9c64829dfa040b334c9ae27c332c7078356e79.zip
 
-[testenv:manifest-checker]
```
rodrigc, line 151: OK, looks like this block which got deleted moved to: https://github.com/twisted/twisted/pull/1380/files#diff-b91f3d5bd63fcd17221b267e851608e8R108

rodrigc (approved):
I think this looks ok.
1.  You have a GitHub Action which invokes black, so we can get feedback from CI
2.  `tox -e black` runs the check, and `tox -e black-reformat` can be run by a contributor to reformat their code

So I would say merge this, and let's get this going.

In follow-up items, I would say:

1. Make a follow-up PR which reformats the tree.  I can quickly review/approve
2. Make a follow-up PR which adds a value to `.git-blame-ignore-revs`.  I can quickly review/approve
3. Send a e-mail to the mailing list, informing that coding style has changed to black, and give pointers to what a user should do to make their existing PR's acceptable to the new coding style.
4.  The docs regarding coding style should be cut to the bare minimum, and just have pointers to the Black guidelines.  Also, just mention in the docs that `tox -e black-reformat` must be run to make any contribution acceptable.
5.  Take a pass through the wiki, and add references to black as appropriate, such as: https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted
6.  Do we need to add Black to the **Contributor Checklist:** that is displayed for ever PR?

PR #1385, opened 2020-09-14
Fix some of the Twisted web examples on Python 3

rodrigc:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9959


docs/web/examples/fortune.rpy
```diff
@@ -38,12 +39,13 @@ def render_GET(self, request):
...
         Deferred which fires with its output.
         """
-        request.write("<pre>\n")
+        request.write(b"<pre>\n")
         deferred = utils.getProcessOutput(self.fortune)
-        deferred.addCallback(lambda s:
-                             (request.write(s+"\n"), request.finish()))
-        deferred.addErrback(lambda s:
-                     (request.write(str(s)), request.finish()))
+        deferred.addCallback(lambda s: (request.write(s + b"\n"), request.finish()))
```
twm, line 44: Man, this is a gross hack. Not that you should fix it now, I just couldn't let it pass.

docs/web/examples/rootscript.py
```diff
@@ -19,21 +19,21 @@
 from twisted.web import vhost, static, script, server
 from twisted.application import internet, service
 
-default = static.Data('text/html', '')
+default = static.Data("text/html", "")
```
twm: Shouldn't the second string here be `bytes`?

twm (approved):
This looks good! I think that there is still a type error with `static.Data` that should be addressed before merge. Additionally, please uncomment this line:

https://github.com/twisted/twisted/blob/e1c54d7e72efa10e578fe50211a4cdebc5d0ee3c/tox.ini#L46

Thanks!

docs/web/examples/rootscript.py
rodrigc: The first parameter to twisted.web.static.Data is documented as `bytes`, while the
the second parameter to twisted.web.static.Data is documented as `str`: https://github.com/twisted/twisted/blob/trunk/src/twisted/web/static.py#L57

I will fix this up.


At some point it will be good to get type annotations for more of twisted.web into the tree, such as this PR: https://github.com/twisted/twisted/pull/1284/ because tracking these types via hand is a pain.

rodrigc: Fixed by changing this to:

```
default = static.Data(b"", "text/html")
```

PR #1386, opened 2020-09-15
Drop support for Python 3.5

rodrigc:
https://twistedmatrix.com/trac/ticket/9958

This has been discussed on the mailing list: ​https://twistedmatrix.com/pipermail/twisted-python/2020-May/065066.html

This PR cannot be merged until  *AFTER* one more major Twisted release which will be announced as the last one with Python 3.5 support.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,7 +15,7 @@ jobs:
       TOXENV: "alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push"
     strategy:
       matrix:
-          python-version: [3.5, 3.6, 3.7, 3.8, 3.9, pypy3]
+          python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
 
     steps:
     - uses: actions/checkout@v2
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -6,7 +6,7 @@ Installation Requirements
 
 To install Twisted, you need:
 
-- Python 3.5/3.6/3.7.
+- Python 3.6/3.7/3.8.
 
 - `setuptools <https://pypi.python.org/pypi/setuptools>`_
   (installed automatically if you use pip).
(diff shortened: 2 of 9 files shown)


wsanchez (approved):
Approved pending a Twisted release announcing the change.

rodrigc:
@graingert why did you close and delete this?

graingert (changes requested):
There's still a few more references to python 3.5 remaining, I think we've caught them all in #1528

rodrigc:
@graingert thanks for providing feedback on this PR.  I originally worked on it in September 2020, and got approval
from @wsanchez .  My intention was not to merge it until *after* a Twisted release was done where that release was the last one support Twisted 3.5.
I released Twisted 21.2.0, which is the last one to support Python 3.5.

As you have noticed, there are some things that have changed in trunk since September that need to get updated in this
PR.

Right now, I want to focus on a few post-release tasks with the 21.2.0 release.  

When the 21.2.0 post-release tasks are completed, I then want to refocus on this PR.

At that point I will:

1.  Let the dust settle on the 21.2.0 release, for maybe a week or so.

2.  bring this PR up to date with the latest trunk (with your feedback)

3.  send out another HEADSUP style e-mail, similar to what I sent out here:  https://twistedmatrix.com/pipermail/twisted-python/2020-May/065066.html

4.   Get review approval on this PR (again)

5.   Merge after approval.

@graingert does that sound OK to you?

adiroiban:
If Craig is busy with other tasks and not actively working on the py3.5 right now,
then why block someone else to work on py3.5 removal ?

This doesn't sound like free software collaboration to me.

We already agree that python3.5 is gone... so I don't see any technical reason to delay any merge that removes 3.5 ... partially or completely.

wsanchez:
All changes here have been included in #1528, which has some additional related work, so let's move there.

@rodrigc I appreciate the thorough approach, but we don't need to let the last release to settle in order to move forward, I think.  We've already sent a heads-up and announced the change in a release.

PR #1388, opened 2020-09-15
Remove _runSequentially

graingert:
in favour of a for loop

## Remove this section

Please have a look at [our developer documentation](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch) before submitting your Pull Request.

Please note that the trac ticket, news fragment, and review submission portions of this process apply to *all* pull requests, no matter how small; if you don't do them, it's likely that nobody will even notice your PR needs a review.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9962


src/twisted/trial/_asynctest.py
```diff
@@ -190,22 +190,24 @@ def _ebDeferTearDown(self, failure, result):
...
-    def _cbDeferRunCleanups(self, cleanupResults, result):
-        for flag, testFailure in cleanupResults:
-            if flag == defer.FAILURE:
-                result.addError(self, testFailure)
-                if testFailure.check(KeyboardInterrupt):
-                    result.stop()
-                self._passed = False
+        for failure_ in failures:
+            result.addError(self, failure)
+            if failure_.check(KeyboardInterrupt):
```
graingert: should KeyboardInterrupt abort cleanups?

mthuurne: It should not abort the cleanups, but it should abort _after_ running the cleanups, which is what happens here since this code runs after collecting all the cleanups results.

graingert: I think it should be more like this:

https://github.com/graingert/twisted/blob/1bfb3b5d6966afe43e46cf38cd10f41f46096eb9/src/twisted/trial/_synctest.py#L1412-L1419

src/twisted/trial/_asynctest.py
```diff
@@ -190,22 +190,24 @@ def _ebDeferTearDown(self, failure, result):
             result.stop()
         self._passed = False
 
+    @defer.inlineCallbacks
     def deferRunCleanups(self, ignored, result):
         """
         Run any scheduled cleanups and report errors (if any to the result
```
mthuurne:
```suggestion
        Run any scheduled cleanups and report errors (if any) to the result.
```

src/twisted/trial/newsfragments/9962.misc
```diff
@@ -0,0 +1 @@
+remove _twisted.trial.util:runSequentially
```
mthuurne, line 1:
```suggestion
remove twisted.trial.util._runSequentially
```

mthuurne: However, content of `.misc` news fragments is ignored, so an empty file might be better.

src/twisted/trial/_asynctest.py
```diff
@@ -190,22 +190,24 @@ def _ebDeferTearDown(self, failure, result):
...
         object.
         """
-        d = self._runCleanups()
-        d.addCallback(self._cbDeferRunCleanups, result)
-        return d
+        failures = []
+        for func, args, kwargs in self._cleanups[::-1]:
```
mthuurne, line 200:
```suggestion
        for func, args, kwargs in reversed(self._cleanups):
```

src/twisted/trial/newsfragments/9962.misc
exarkun: The follow the style guideline, this would be "twisted.trial.util._runSequentially has been removed."

However, it's a private API so it doesn't seem worth text in the news file anyway.   Also worthy of note may be that .misc files have their contents discarded anyway so there's no reason to put any text here anyway.

src/twisted/trial/_asynctest.py
mthuurne: Note that `reversed()` does not make a copy, but since there is no `removeCleanup()` method, the `_cleanups` list shouldn't be modified during iteration.

mthuurne: Does it make sense to update `_runCleanups()` in `SynchronousTestCase` (superclass to the async `TestCase`) to use the same iteration style? Currently it still uses `while` + `pop()`.

src/twisted/trial/_asynctest.py
```diff
@@ -190,22 +190,24 @@ def _ebDeferTearDown(self, failure, result):
...
-        for flag, testFailure in cleanupResults:
-            if flag == defer.FAILURE:
-                result.addError(self, testFailure)
-                if testFailure.check(KeyboardInterrupt):
-                    result.stop()
-                self._passed = False
+        for f in failures:
+            result.addError(self, f)
+            if f.check(KeyboardInterrupt):
+                result.stop()
```
graingert:
```suggestion
```

I think this is unreachable because KeyboardInterrupt won't be appended to failures

mthuurne (approved):
I suggested some minor changes, please have a look and include them if you agree. But none of them are blocking issues.

src/twisted/trial/_asynctest.py
graingert: that allows re-entrant cleanups

src/twisted/trial/newsfragments/9962.misc
graingert: > Please note that the trac ticket, news fragment, and review submission portions of this process apply to all pull requests, no matter how small; if you don't do them, it's likely that nobody will even notice your PR needs a review.

is there a spec of when the trac ticket/news fragment is needed?

src/twisted/trial/_asynctest.py
graingert: would it be better to allow re-entrant cleanup in `_asynctest.py`?

mthuurne: I'm not sure if re-entrant cleanups are a feature worth having, but if it's a feature that Trial has right now (either by design or by accident), the safest thing is to keep it.

src/twisted/trial/newsfragments/9962.misc
mthuurne: > is there a spec of when the trac ticket/news fragment is needed?

Yes, it's linked from the checklist template.

mthuurne: > is there a spec of when the trac ticket/news fragment is needed?

Yes, it's linked from the checklist template.

src/twisted/trial/_asynctest.py
mthuurne: Indeed as you remarked on IRC, `KeyboardInterrupt` does not inherit from `Exception`.

src/twisted/trial/_asynctest.py
graingert: what happens if you addCleanup while iterating over reversed(list(...))?

src/twisted/trial/newsfragments/9962.misc
graingert:
```suggestion
```

src/twisted/trial/_asynctest.py
mthuurne: With just `reversed(self._cleanups)` it's undefined, with `reversed(list(self._cleanups))` the added cleanup would be ignored.

src/twisted/trial/_asynctest.py
```diff
@@ -205,8 +205,6 @@ def deferRunCleanups(self, ignored, result):
 
         for f in failures:
             result.addError(self, f)
-            if f.check(KeyboardInterrupt):
-                result.stop()
```
mthuurne: Even though this condition is never true, is removing the check an improvement or does it takes us even further from the desired behavior?

graingert: removing unreachable code is an improvement, swapping this for AsyncExitStack and SyncExitStack could be done in a followup?

src/twisted/trial/_asynctest.py
graingert: I think sticking with the `self._cleanups[::-1]:` is best as that's idiomatic for reversing and copying a list

src/twisted/trial/test/test_tests.py
```diff
@@ -965,6 +965,28 @@ def test_multipleErrorsReported(self):
...
+    def test_cleanupsAreReentrant(self):
+        @self.test.addCleanup
+        def reenter():
+            self.test.append("foo")
+            self.test.addCleanup(self.test.append, "bar")
+
+        self.test.run(self.result)
+        self.assertEqual(["setUp", "runTest", "foo", "bar", "tearDown"], self.test.log)
+
+    def test_cleanupsIgnoreKeyboardInterrupt(self):
```
mthuurne:
```suggestion
    def test_cleanupsHandleKeyboardInterrupt(self):
```

src/twisted/trial/test/test_tests.py
```diff
@@ -965,6 +965,28 @@ def test_multipleErrorsReported(self):
...
+
+        @self.test.addCleanup
+        def ctrlc():
+            raise KeyboardInterrupt
+
+        self.test.addCleanup(self.test.append, "bar")
+
+        self.test.run(self.result)
+        self.assertEqual(["setUp", "runTest", "bar", "foo", "tearDown"], self.test.log)
+
```
mthuurne:
```suggestion
        self.assertTrue(self.result.shouldStop)

```

src/twisted/trial/test/test_tests.py
```diff
@@ -965,6 +965,28 @@ def test_multipleErrorsReported(self):
         self.assertEqual(error2.getErrorMessage(), "foo")
 
 
+    def test_cleanupsAreReentrant(self):
```
mthuurne: Is "reentrant" the correct term here? It suggests both callbacks are on the call stack simultaneously, which is not the case.

src/twisted/trial/test/test_tests.py
mthuurne: I think it would be better to split off the `KeyboardInterrupt` changes into a new PR, since this is a bug fix which should be documented in the release notes.

graingert: Agreed, shall I reset to https://github.com/twisted/twisted/pull/1388/commits/450e0e118e5539ebf46127d5b98ece58e04aa066 ?

src/twisted/trial/test/test_tests.py
```diff
@@ -965,6 +965,28 @@ def test_multipleErrorsReported(self):
...
+            self.test.append("foo")
+            self.test.addCleanup(self.test.append, "bar")
+
+        self.test.run(self.result)
+        self.assertEqual(["setUp", "runTest", "foo", "bar", "tearDown"], self.test.log)
+
+    def test_cleanupsIgnoreKeyboardInterrupt(self):
+        self.test.addCleanup(self.test.append, "foo")
+
+        @self.test.addCleanup
```
mthuurne: For consistency, maybe just make a plain call instead of using the decoration syntax.

graingert: I don't know what you mean, can you use `` ```suggest ``

src/twisted/trial/_asynctest.py
mthuurne: OK.

src/twisted/trial/_asynctest.py
mthuurne: Any reactor that inherits `_SignalReactorMixin` replaces the default `SIGINT` handler that raises `KeyboardInterrupt` with one that doesn't, so in practice cleanup handlers cannot be aborted with CTRL-C. Addressing this issue is going to require more invasive changes and then indeed it doesn't help to keep this code around.

mthuurne (approved):
I think it's ready to land now.

Note that [inconsistencies between the sync and async `TestCase` implementations](https://twistedmatrix.com/trac/ticket/9963) will be addressed separately; this PR preserves the current behavior even though it is inconsistent.

mthuurne:
@exarkun indicated on IRC that he doesn't intend to re-review this PR.

The [Travis check](https://travis-ci.com/github/twisted/twisted/builds/186462713) passed, but somehow that result didn't reach GitHub.

With that, I'll merge this PR now.

PR #1389, opened 2020-09-15
[#9841] Clean up type checking/hinting for twisted.web.http_headers (take 2)

rodrigc:
https://twistedmatrix.com/trac/ticket/9841

This is a continuation of, and closes, #1284


src/twisted/python/_typing.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+
+"""
+Short-hands for L{typing}.
+"""
+
+from typing import Union
+
+String = Union[bytes, str]
```
mthuurne: We could use [`typing.AnyStr`](https://docs.python.org/3/library/typing.html?highlight=typing#typing.AnyStr) instead.

src/twisted/web/http_headers.py
```diff
@@ -235,12 +267,14 @@ def addRawHeader(self, name, value):
 
         self.setRawHeaders(name, values)
 
-    def getRawHeaders(self, name, default=None):
+    def getRawHeaders(
+        self, name: String, default: Optional[Any] = None
+    ) -> Optional[Union[List[String], Any]]:
```
mthuurne: [`typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload) could be used to express that the returned string type matches `name`. If this were used in combination with a `TypeVar` for the type of `default`, the return type would be exact.

src/twisted/web/http_headers.py
```diff
@@ -71,8 +72,8 @@ class Headers:
         b"x-xss-protection": b"X-XSS-Protection",
     }
 
-    def __init__(self, rawHeaders=None):
-        self._rawHeaders = {}
+    def __init__(self, rawHeaders: Optional[Dict[String, List[String]]] = None):
```
glyph: Do we really want `Dict` here, or is this more of a `Mapping`?  Similarly; `List` or `Sequence`?  (We should be careful as we add types not to get too overly specific and make everybody's unit tests need `type: ignore` all over the place; use the least specific type that works for inputs)

src/twisted/python/_typing.py
glyph: Yeah, let's not do this.

glyph: (I'd much rather just inline `Union[bytes, str]` if we're using it in a context where a typevar doesn't work.  Having this as an alias makes it too convenient to reach for.)

src/twisted/web/http_headers.py
```diff
@@ -97,13 +98,13 @@ def __cmp__(self, other):
             )
         return NotImplemented
 
-    def _encodeName(self, name):
+    def _encodeName(self, name: String) -> bytes:
```
glyph: yeah AnyStr here would definitely work better

src/twisted/web/http_headers.py
```diff
@@ -209,23 +214,50 @@ def setRawHeaders(self, name, values):
...
         name = _sanitizeLinearWhitespace(self._encodeName(name))
         encodedValues = [
             _sanitizeLinearWhitespace(v) for v in self._encodeValues(values)
         ]
 
-        self._rawHeaders[name] = self._encodeValues(encodedValues)
+        self._rawHeaders[name] = encodedValues
 
-    def addRawHeader(self, name, value):
+    def addRawHeader(self, name: String, value: String) -> None:
```
glyph: Here's a place where `AnyStr` would be awkward; `name` and `value` *don't* have to be the same type.  But I'd still rather just inline `Union[bytes, str]` here twice separately.

glyph (changes requested):
Thanks for fixing this up!  I *really* don't want a type called `String`.  If you really want to define an alias (I don't think we should, I think `Union[str, bytes]` is clearer and easier to read, and these type annotations are largely for people to read) it should really be `SighItCouldBeBytesOrTextIDontKnow`

src/twisted/web/http_headers.py
rodrigc: OK, I replaced the private alias `String` with `Union[bytes, str]` everywhere.
I also deleted the `_typing.py` file.

src/twisted/web/http_headers.py
rodrigc: I went with  `Union[bytes, str]` here.  Do you want that, or `AnyStr`?

src/twisted/web/http_headers.py
rodrigc: The docstring for `_rawHeaders` is:

```
   @ivar _rawHeaders: A L{dict} mapping header names as L{bytes} to L{list}s of
        header values as L{bytes}.
```

and the initial variable annotation is:

```
       self._rawHeaders = {}  # type: Dict[bytes, List[bytes]]
```


Is it better to stick with this, or to change the rawHeaders argument to:

```
    def __init__(
        self,
        rawHeaders: Optional[Mapping[Union[bytes, str], Sequence[Union[bytes, str]]]] = None,
    ):
```

??

twm (approved):
Awesome, you beat me to this. :-) This looks great, and Glyph's concern has been addressed, so merge away. Thank you!

src/twisted/python/_typing.py
rodrigc: I deleted this file.

src/twisted/web/http_headers.py
rodrigc: @wsanchez @twm What is the best way to use `typing.overload` here?  Will this still retain compatibility with things like klein and treq?

src/twisted/web/http_headers.py
```diff
@@ -71,8 +71,11 @@ class Headers:
         b"x-xss-protection": b"X-XSS-Protection",
     }
 
-    def __init__(self, rawHeaders=None):
-        self._rawHeaders = {}
+    def __init__(
+        self,
+        rawHeaders: Optional[Dict[Union[bytes, str], List[Union[bytes, str]]]] = None,
```
wsanchez: I think that you want `AnyStr` here instead of `Union[bytes, str]`, which would constrain the dictionary to use either `bytes` or `str` but not a mishmash of both.

src/twisted/web/http_headers.py
```diff
@@ -97,13 +100,13 @@ def __cmp__(self, other):
...
 
-    def _encodeName(self, name):
+    def _encodeName(self, name: Union[bytes, str]) -> bytes:
         """
         Encode the name of a header (eg 'Content-Type') to an ISO-8859-1 encoded
         bytestring if required.
 
         @param name: A HTTP header name
-        @type name: L{unicode} or L{bytes}
+        @type name: L{str} or L{bytes}
```
wsanchez: Not really part of this review, I guess, but shouldn't we be getting rid of `@type` and `@rtype` in the epytext now? It's redundant when you have typing info in the actual code, no?

src/twisted/web/http_headers.py
```diff
@@ -167,40 +170,46 @@ def copy(self):
...
+        @type name: L{bytes} or L{str}
         @param name: The name of the HTTP header to remove.
 
         @return: L{None}
         """
         self._rawHeaders.pop(self._encodeName(name), None)
 
-    def setRawHeaders(self, name, values):
+    def setRawHeaders(
+        self, name: Union[bytes, str], values: List[Union[bytes, str]]
```
wsanchez: `Union[bytes, str]` ➜ `AnyStr`, I think

src/twisted/web/http_headers.py
```diff
@@ -209,23 +218,50 @@ def setRawHeaders(self, name, values):
...
+            if not isinstance(value, (bytes, str)):
+                raise TypeError(
+                    "Header value at position %s is an instance of %r, not "
+                    "bytes or str"
+                    % (
+                        count,
+                        type(value),
+                    )
+                )
+
```
wsanchez: Why is this included in this PR? It's not adding type info; it's doing something else.

src/twisted/web/http_headers.py
```diff
@@ -209,23 +218,50 @@ def setRawHeaders(self, name, values):
...
         name = _sanitizeLinearWhitespace(self._encodeName(name))
         encodedValues = [
             _sanitizeLinearWhitespace(v) for v in self._encodeValues(values)
         ]
 
-        self._rawHeaders[name] = self._encodeValues(encodedValues)
+        self._rawHeaders[name] = encodedValues
 
-    def addRawHeader(self, name, value):
+    def addRawHeader(self, name: Union[bytes, str], value: Union[bytes, str]) -> None:
```
wsanchez: `Union[bytes, str]` ➜ `AnyStr`

src/twisted/web/http_headers.py
```diff
@@ -209,23 +218,50 @@ def setRawHeaders(self, name, values):
...
+            raise TypeError(
+                "Header name is an instance of %r, " "not bytes or str" % (type(name),)
+            )
+
+        if not isinstance(value, (bytes, str)):
+            raise TypeError(
+                "Header value is an instance of %r, not "
+                "bytes or str" % (type(value),)
+            )
+
```
wsanchez, line 217: Same as above, why in this PR?

src/twisted/web/http_headers.py
```diff
@@ -235,12 +271,14 @@ def addRawHeader(self, name, value):
 
         self.setRawHeaders(name, values)
 
-    def getRawHeaders(self, name, default=None):
+    def getRawHeaders(
+        self, name: Union[bytes, str], default: Optional[Any] = None
+    ) -> Optional[Union[List[Union[bytes, str]], Any]]:
```
wsanchez: `Union[bytes, str]` ➜ `AnyStr`

Isn't `Any` under-constraining the values here?  Aren't the values also strings?

src/twisted/web/http_headers.py
```diff
@@ -254,11 +292,11 @@ def getRawHeaders(self, name, default=None):
         encodedName = self._encodeName(name)
         values = self._rawHeaders.get(encodedName, default)
 
-        if isinstance(name, str) and values is not default:
+        if isinstance(name, str) and values is not default and values is not None:
```
wsanchez: I imagine this was generating a `mypy` error because `values` is optional… and it looks like `_decodeValues(None)` would have raised an exception…?  So this changes behavior in that case… does that just never happen in practice?

src/twisted/web/http_headers.py
```diff
@@ -43,8 +43,8 @@ class Headers:
     """
     Stores HTTP headers in a key and multiple value format.
 
-    Most methods accept L{bytes} and L{unicode}, with an internal L{bytes}
-    representation. When passed L{unicode}, header names (e.g. 'Content-Type')
+    Most methods accept L{bytes} and L{str}, with an internal L{bytes}
+    representation. When passed L{str}, header names (e.g. 'Content-Type')
```
wsanchez: It's not really a requirement for this PR, but I feel like we should remove the phrase about the internal representation, because that's… internal and really none of your business as a client of the API.

src/twisted/web/http_headers.py
wsanchez: I guess from a backward compat POV, this isn't really a problem, but it's not clear that calling isinstance on each item is worth the work; I mean it wasn't before, and now with the type checking, you should be catching such things with `mypy`, no?

src/twisted/web/http_headers.py
rodrigc: That's a good idea.

src/twisted/web/http_headers.py
rodrigc: Yes, without this change, mypy errors out with:

```
src/twisted/web/http_headers.py:296:39: error: Argument 1 to "_decodeValues" of "Headers" has incompatible type "Optional[List[bytes]]"; expected "List[bytes]"  [arg-type]
                return self._decodeValues(values)
```

src/twisted/web/http_headers.py
twm: The default is `Any`. Would a type parameter be the correct way to model this?

src/twisted/web/http_headers.py
rodrigc: Removed

src/twisted/web/http_headers.py
rodrigc: Fixed

src/twisted/web/http_headers.py
rodrigc: Fixed

src/twisted/web/http_headers.py
rodrigc: Fixed

src/twisted/web/http_headers.py
twm: When `_sanitizeLinearWhitespace` was introduced it added an implicit type assertion (technically this was a backwards-incompatible change; it caught a number of places where non-string values were passed in some codebases I'm aware of, fortunately all in tests). When this happens it produces a mildly cryptic exception, so I'm in favor of the much-better exceptions added here.

src/twisted/web/http_headers.py
rodrigc: I went with AnyStr as suggested by @wsanchez

src/twisted/web/http_headers.py
rodrigc: Fixed

twm (commented):
@wsanchez w.r.t. your `AnyStr` suggestions, don't the unions represent the true behavior of the implementation? I expect that I have code somewhere that passes mixed types. It works. Do we consider constraining the types further a compatible change?

src/twisted/web/http_headers.py
rodrigc: In `test_getRawHeadersDefaultValue(`, https://github.com/twisted/twisted/blob/bc96c774be6e307e8e3e4d39780d37b045d0973a/src/twisted/web/test/test_http_headers.py#L115 

`default` of type `object()` is passed, so I think this needs to by `Any`.

src/twisted/web/http_headers.py
rodrigc: This came from the original PR done by @hawkowl here: https://github.com/twisted/twisted/pull/1284/files#diff-effe5888987605608f990d71470cc65dR229

src/twisted/web/http_headers.py
rodrigc: This came from @hawkowl 's original PR here: https://github.com/twisted/twisted/pull/1284/files#diff-effe5888987605608f990d71470cc65dR223

src/twisted/web/http_headers.py
wsanchez: We should use `Mapping` and `Sequence` here unless the implementation actually expects to be given those specific type (which would be unfortunate).

The type of `_rawHeaders` is internal, so it can (and should be, in this case), more specific than the inputs to `__init__`, so I would not argue that it dictates the answer here.

FWIW I think that Twisted's docstrings a very prone to saying `list` where `Iterable` or `Sequence` is more appropriate, so we should be cleaning that up as we add typing, not preserving it.  In docstrings, the mistake produces no errors. With type checkers, we start to need the accuracy in a stricter sense.

src/twisted/web/http_headers.py
wsanchez: In this case, it doesn't need to be `AnyStr`, since the return value and `name` don't have to match.  That said, I think it's probably OK to use here… you just need to be careful that `AnyStr` != `Union[bytes, str] ` in that the types must be the same.

src/twisted/web/http_headers.py
wsanchez: ok

src/twisted/web/http_headers.py
wsanchez: No, that's a bad reason.

I mean the test is doing it wrong.  I think it's fine that the test is testing what happens when you do it wrong, or cheating to make sure `default` is pass along as-is.  The solution is to ignore or cast the type in the test (`default = cast(bytes, object())` or something).

Don't pass on the wrong to the client, though.

src/twisted/web/http_headers.py
wsanchez: I would argue that it would be more correct to put `assert values is not None`  into the `if` block. It's not a valid state.

Thinking… ok… what's actually going on here is that the return type for `self._rawHeaders.get` should be `List[AnyStr]`. The exception would be if `default` is not of that type (eg. `None`), but really that's not a correct type for `default`, which should also be `List[AnyStr]`, not `Any`.  And the return type for this method should be `List[AnyStr]`; the only reason it isn't is that we set `default`'s type to `Any`.

(Note the `Optional` in the return type is a bit nuts: `Optional` is a `Union`, which here contains another `Union`, which contains `Any`… all of that text just means: `Any`.)

I think this method's type is way too loose.  Other people should weigh in here, but I do think the correct thing to do here is to start by fixing the method type to:

```python
def getRawHeaders(self, name: AnyStr, default: List[AnyStr] = None) -> List[AnyStr]:
```

Certainly, an existing client might do something (arguably wrong) like: `rhs = h.getRawHeaders("foo", None)` but I think that this code would be better off with `rhs = h.getRawHeaders("foo", [])`, so it would be good if `mypy` actually pointed out to them that they could clean that up.

wsanchez:
> @wsanchez w.r.t. your `AnyStr` suggestions, don't the unions represent the true behavior of the implementation? I expect that I have code somewhere that passes mixed types. It works. Do we consider constraining the types further a compatible change?

Well, I'm not proposing that we change the implementation, so I don't think that being more strict in the typing is a compatibility issue.  If typing caused runtime exceptions, that would be different.

I think the point of typing is to encourage clients to use the code as intended, reducing the odds of bugs. It's true that there might be code providing mixed inputs.  The relevant input points for that are… `__init__` and `setRawHeaders`…

You could have code that sends in `bytes` for a name and `str` for a value… that was probably considerable more likely with Python 2, but fortunately we don't support that now… I guess I'd argue that if I were doing that in code, I'd appreciate something telling me and fixing it would either be easy or I bet something is work looking at with this mismatch in mind.

My feeling on this is: let someone file a bug if it's too strict a constraint and we can discuss it. It's not going to break at runtime, so we gain the opportunity to look at examples.  Too loose means things don't get noticed that should.

src/twisted/web/http_headers.py
wsanchez: Actually, I think `Iterable`, not `Sequence`.  Which necessarily means same in `setRawHeaders`.  Which then also means `_encodeValues` and friends.

wsanchez: Actually, I think `Iterable`, not `Sequence`.  Note this necessarily means the same in `setRawHeaders`.  Which then also means `_encodeValues` and friends.

Also, the type check for values in `setRawHeaders` is wrong for this reason… it shouldn't have to be a list; we should be checking that it's an iterable.  That might mean `hasattr(values, "__iter__")` instead.

wsanchez: FWIW I'm very prone to using tuples instead of lists, because I don't like mutable types when you don't need mutability, and a class like this should not take in and modify a mutable object without being very clear that it's doing so.  (Which is what using `List` implies.)

src/twisted/web/http_headers.py
rodrigc: OK, I put `assert values is not None` in the if block

src/twisted/web/http_headers.py
rodrigc: OK, I changed to `Iterable`.

src/twisted/web/http_headers.py
rodrigc: I tried change to:

```
def getRawHeaders(self, name: AnyStr, default: List[AnyStr] = None) -> List[AnyStr]:
```

but got:

```
src/twisted/web/http_headers.py:283:39: error: Argument 1 to "_decodeValues" of "Headers" has incompatible type "Union[List[bytes], List[str]]"; expected "Iterable[bytes]"
[arg-type]
                return self._decodeValues(values)
                                          ^
src/twisted/web/http_headers.py:284:16: error: Incompatible return value type (got "Union[List[bytes], List[str], None]", expected "List[str]")  [return-value]
            return values
                   ^
src/twisted/web/http_headers.py:284:16: error: Incompatible return value type (got "Optional[List[bytes]]", expected "List[bytes]")  [return-value]
            return values
                   ^
Found 3 errors in 1 file (checked 839 source files)
```

twm:
@wsanchez That seems reasonable to me.  I filed [#9964](https://twistedmatrix.com/trac/ticket/9964) to incorporate that into the compatibility policy.

src/twisted/web/http_headers.py
rodrigc: Actually I had to change this back to `Sequence`, because some of the test cases which passed in a `dict` were not raising a `TypeError`

src/twisted/web/http_headers.py
rodrigc: OK, I got this to work.  I had to refactor the code a little bit but now  `getRawHeaders()` is:

```
 def getRawHeaders(
        self, name: AnyStr, default: List[AnyStr] = None
    ) -> Optional[List[AnyStr]]:
```

All the tests and mypy pass.

rodrigc: @wsanchez does this look OK now?

wsanchez: OK so this still has default as `Optional` (`mypy` will make it optional because it's defaulting to `None` because we haven't turned on `no_implicit_optional`, which I think we should…) and so we return an `Optional`.

I really think we shouldn't be returning `None` here and `self._rawHeaders.get(encodedName, None)` should instead read `self._rawHeaders.get(encodedName, [])`.  That would be a change in behavior.

@glyph I feel like you've expressed similar things but I'm not sure what compatibility requires or if I'm remembering you incorrectly.

wsanchez: Here is @glyph's suggestion:

```python
def getRawHeaders(self, name: AnyStr, default: T = None) -> Union[List[AnyStr], T]:
```

That gives you a return type of `List[AnyStr]` (which makes me happy) unless you pass in a default value that isn't an `AnyStr` (which I think makes you a weirdo but whatever).

rodrigc: I tried that and got:

```
src/twisted/web/http_headers.py:227:38: error: Name 'T' is not defined  [name-defined]
            self, name: AnyStr, default: T = None
                                         ^
src/twisted/web/http_headers.py:228:30: error: Name 'T' is not defined  [name-defined]
        ) -> Union[List[AnyStr], T]:
```

rodrigc: Doesn't this take us in a circle back to:

```
def getRawHeaders(self, name: AnyStr, default: Any = None) -> Union[List[AnyStr], Any]:
```

mthuurne: The typical definition for a non-specific type variable is:
```py
from typing import TypeVar
T = TypeVar('T')
```
While `T` and `Any` will both accept any type for the default argument, the return type will be different. For example if the method is called with `bytes` and `None` arguments, the signature with `T` will produce a return type of `Optional[List[bytes]]` while the signature with `Any` will produce a return type of `Any`.

rodrigc: @wsanchez @glyph Can you be a bit more clear about what you want to be implemented for `getRawHeaders`?

Based on @wsanchez's feedback, I tried something like this:

```
    _T = TypeVar('_T')

    def getRawHeaders(
        self, name: AnyStr, default: Optional[_T] = None
    ) -> Union[List[AnyStr], Optional[_T]]:
        """
        Returns a list of headers matching the given name as the raw string
        given.

        @param name: The name of the HTTP header to get the values of.

        @param default: The value to return if no header with the given C{name}
            exists.

        @return: If the named header is present, a L{list} of its
            values.  Otherwise, C{default}.
        """
        encodedName = self._encodeName(name)
        values = self._rawHeaders.get(encodedName, [])
        if not values:
            return default

        if isinstance(name, str):
            return [v if isinstance(v, str) else v.decode("utf8") for v in values]
        return values
```

Would that work for you?

wsanchez: Yeah sorry, you do need to define `T`, and `_T` is a better choice since the module is public.

You shouldn't need to make it `Optional`, though, since `_T` is unbounded; it's basically `Any`.

rodrigc: The reason I made it `Optional` is because if I do this:

```
    _T = TypeVar('_T')

    def getRawHeaders(
        self, name: AnyStr, default: _T = None
    ) -> Union[List[AnyStr], _T]:
        """
        Returns a list of headers matching the given name as the raw string
        given.

        @param name: The name of the HTTP header to get the values of.

        @param default: The value to return if no header with the given C{name}
            exists.

        @return: If the named header is present, a L{list} of its
            values.  Otherwise, C{default}.
        """
        encodedName = self._encodeName(name)
        values = self._rawHeaders.get(encodedName, [])
        if not values:
            return default

        if isinstance(name, str):
            return [v if isinstance(v, str) else v.decode("utf8") for v in values]
        return values
```

Then mypy complains with:

```
src/twisted/web/http_headers.py:248:20: error: Incompatible return value type (got "Optional[_T]", expected "Union[List[str], _T]")  [return-value]
                return default
                       ^
src/twisted/web/http_headers.py:248:20: error: Incompatible return value type (got "Optional[_T]", expected "Union[List[bytes], _T]")  [return-value]
                return default
                       ^
Found 2 errors in 1 file (checked 839 source files)
```

glyph: Default parameters and type vars interact in odd ways, and `None` weirder than most. I've filed https://github.com/python/mypy/issues/9468 because I couldn't find the existing bug about this (I swear there is one…)

For the time being probably just dealing with the `Optional` like this is the best policy.

glyph:
> I expect that I have code somewhere that passes mixed types. It works.

What do you mean?  How do you pass a "mixed type" to a single invocation of `getRawHeaders`?

glyph:
> My feeling on this is: let someone file a bug if it's too strict a constraint and we can discuss it. It's not going to break at runtime, so we gain the opportunity to look at examples.

This is my feeling as well.

More broadly, going forward with types: the main reason we have a compat policy constrain runtime behavior so strictly is so that, if you, in a production-ish context, do `pip install --upgrade twisted` to upgrade by one version, your system will not instantly break horribly in an impossible to diagnose way; it'll start emitting warnings, which hopefully you can see and deal with.

Making it easier to see those warnings and deal with them promptly is a constant challenge; we have a lot of work to do with logging in general to make this more transparent.  (This will continue to be a problem as there are plenty of state-dependent APIs in Twisted where type-checking won't save you.)

However, with types, you only run type-checking in a controlled development context, not as part of making your software work, so I am in favor of loudly and instantly blowing up at the type-checking phase, because this makes it *easier* for developers to update their code before it gets to production; seeing a bunch of type-check errors as part of CI does not break a running system.

src/twisted/web/http_headers.py
rodrigc: Looks like the mypy issue you filed is a dupe of https://github.com/python/mypy/issues/3737, https://github.com/python/mypy/issues/8708, https://github.com/python/mypy/issues/8739, https://github.com/python/mypy/issues/9179, https://github.com/python/mypy/issues/9432 :)

rodrigc: OK, I have changed the signature of `getRawHeaders()` to:

```
    def getRawHeaders(
        self, name: AnyStr, default: Optional[_T] = None
    ) -> Union[List[AnyStr], Optional[_T]]:
```

Does this look OK now?

glyph: > Looks like the mypy issue you filed is a dupe of [python/mypy#3737](https://github.com/python/mypy/issues/3737), [python/mypy#8708](https://github.com/python/mypy/issues/8708), [python/mypy#8739](https://github.com/python/mypy/issues/8739), [python/mypy#9179](https://github.com/python/mypy/issues/9179), [python/mypy#9432](https://github.com/python/mypy/issues/9432) :)

Yeah and I'm already following like 3 of these :)

wsanchez: OK I see.  It's harmless, if superfluous.  Let's call it good for now.

wsanchez (approved):
Great progress here, thanks.

src/twisted/web/http_headers.py
rodrigc: @wsanchez @glyph @twm thanks for the review!

PR #1395, opened 2020-09-23
9970 mypy unreachable

wsanchez:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9970


mypy.ini
```diff
@@ -17,4 +22,41 @@ check_untyped_defs       = False
 disallow_incomplete_defs = False
 no_implicit_optional     = False
 warn_return_any          = False
-warn_unreachable         = False
+
+# Disable warn_unreachable for some modules until we clean them up
```
wsanchez, line 26: I'll follow this up with separate PRs for the remaining cases, as they require more input and I'm trying to keep this PR to the low-hanging fruit.

src/twisted/conch/checkers.py
```diff
@@ -13,14 +13,18 @@
 from base64 import decodebytes
 
 try:
-    import pwd
+    import pwd as _pwd
 except ImportError:
-    pwd = None  # type: ignore[assignment]
+    pwd = None
+else:
+    pwd = _pwd
```
wsanchez, line 20: Using @cjerdonek's [workaround](https://github.com/python/mypy/issues/1297#issuecomment-508593494) to "Incompatible types in assignment" errors throughout this PR.

I found that `# type: ignore` as a work-around brings along other fallout, including that it confuses Mypy when checking usage of these symbols later in the file.

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -21,24 +21,34 @@
...
+    cryptography = _cryptography
 
 try:
-    from twisted.conch.avatar import ConchUser
+    from twisted.conch.avatar import ConchUser as _ConchUser
 except ImportError:
-    ConchUser = object  # type: ignore[assignment,misc]
+    ConchUser = object
+else:
+    ConchUser = _ConchUser  # type: ignore[misc]
```
wsanchez, line 51: Error here is

```console
src/twisted/conch/test/test_filetransfer.py:51:5: error: Cannot assign multiple types to name "ConchUser" without an explicit "Type[...]" annotation  [misc]
```

src/twisted/conch/test/test_filetransfer.py
```diff
@@ -93,7 +103,7 @@ def __init__(self, avatar):
     )
 else:
 
-    class FileTransferForTestAvatar(SFTPServerForUnixConchUser):
+    class FileTransferForTestAvatar(SFTPServerForUnixConchUser):  # type: ignore[misc,valid-type]
```
wsanchez, line 106: Errors here are:

```console
src/twisted/conch/test/test_filetransfer.py:106:37: error: Variable "twisted.conch.test.test_filetransfer.SFTPServerForUnixConchUser" is not valid as a type  [valid-type]
        class FileTransferForTestAvatar(SFTPServerForUnixConchUser):
                                        ^
src/twisted/conch/test/test_filetransfer.py:106:37: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
src/twisted/conch/test/test_filetransfer.py:106:37: error: Invalid base class "SFTPServerForUnixConchUser"  [misc]
        class FileTransferForTestAvatar(SFTPServerForUnixConchUser):
                                        ^
src/twisted/conch/test/test_cftp.py:751:1: error: Cannot determine consistent method resolution order (MRO) for "SFTPTestProcess"  [misc]
    class SFTPTestProcess(protocol.ProcessProtocol):
    ^
```

src/twisted/conch/test/test_manhole.py
```diff
@@ -9,6 +9,9 @@
 """
 
 import traceback
+from typing import Optional
+
+ssh = None  # type: Optional[bool]
```
wsanchez, line 14: Without this, the error is:

```console
src/twisted/conch/test/test_manhole.py:19:1: error: Incompatible import of "ssh" (imported name has type "bool", local name has type "object")  [misc]
    from twisted.conch.test.test_recvline import (
    ^
```

There are a few similar others.

src/twisted/internet/abstract.py
```diff
@@ -18,6 +18,11 @@
 from twisted.internet import interfaces, main
 
 
+def _dataMustBeBytes(obj):
+    if not isinstance(obj, bytes):  # no, really, I mean it
+        raise TypeError("Data must be bytes")
```
wsanchez, line 23: Abstracted out so that he error message raised is consistent.

This used to check for string (in the various places this code existed), which caused Mypy to complain that the `raise` line is unreachable, since the tested object was typed as `bytes`.

Changing it to test for `bytes` instead got Mypy not to complain (presumably because type-checking-and-raising is common), and it has the advantage of raising for other invalid types, not just `str`; if we expect `bytes`, that's what we should be testing for.

src/twisted/internet/test/test_tcp.py
```diff
@@ -3010,7 +3010,7 @@ class TestException(Exception):
                 self.assertFalse(self.events)
                 raise TestException()
 
-        self.assertEqual(1, len(self.events))
+        self.assertEqual(1, len(self.events))  # type: ignore[unreachable]
```
wsanchez, line 3013: There are a few tests that capture exceptions using a contest manager; Mypy doesn't know that the exceptions are captured, so it things the following code is unreachable.

src/twisted/python/urlpath.py
```diff
@@ -139,12 +139,7 @@ def fromString(klass, url):
         @rtype: L{URLPath}
         """
         if not isinstance(url, str):
-            raise ValueError("'url' must be a str or unicode")
-        if isinstance(url, bytes):
-            # On Python 2, accepting 'str' (for compatibility) means we might
-            # get 'bytes'.  On py3, this will not work with bytes due to the
-            # check above.
-            return klass.fromBytes(url)
```
wsanchez, line 147: No more Py2

src/twisted/python/urlpath.py
```diff
@@ -163,10 +158,6 @@ def fromBytes(klass, url):
         if not isinstance(url, bytes):
             raise ValueError("'url' must be bytes")
         quoted = urlquote(url, safe=_allascii)
-        if isinstance(quoted, bytes):
-            # This will only be bytes on python 2, where we can transform it
-            # into unicode.  On python 3, urlquote always returns str.
-            quoted = quoted.decode("ascii")
```
wsanchez, line 169: No more Py2

src/twisted/spread/jelly.py
```diff
@@ -721,11 +721,8 @@ def _unjelly_decimal(self, exp):
         return decimal.Decimal((sign, guts, exponent))
 
     def _unjelly_boolean(self, exp):
-        if bool:
-            assert exp[0] in (b"true", b"false")
-            return exp[0] == b"true"
-        else:
-            return Unpersistable("Could not unpersist boolean: %s" % (exp[0],))
```
wsanchez, line 728: `if bool` is always `True`.

wsanchez:
@rodrigc I suspect you'll find this helpful.

src/twisted/conch/checkers.py
rodrigc: Ah ok, in my first go around with mypy, I chose to go with adding `#type: ignore[assignment]` instead of going with the [workaround](https://github.com/python/mypy/issues/1297#issuecomment-508593494).

Your change looks fine.

In future, I hope that mypy can handle this kind of conditional import, but I don't think that will happen soon.

rodrigc (approved):
This looks fine.  Good to see we are increasing the number of mypy checks in Twisted.

PR #1397, opened 2020-09-24
add twisted.internet.defer.coroFnToDeferredFn decorator for convenient use of async def functions

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9972

### DIFF
--- a/docs/core/howto/defer-intro.rst
+++ b/docs/core/howto/defer-intro.rst
@@ -401,8 +401,10 @@ Coroutines with async/await
 
     .. versionadded:: 16.4
 
-On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`.
-It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``, and is a function rather than a decorator.
+On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.coroFnToDeferredFn <coroFnToDeferredFn>`.
+
+It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``.
+"await" functionality can also be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`. It is similar to ``coroFnToDeferredFn``, except that it is a function rather than a decorator.
 
 Calling a coroutine (that is, the result of a function defined by ``async def funcname():``) with :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>` will allow you to "await" on Deferreds inside it, and will return a standard Deferred.
 You can mix and match code which uses regular Deferreds, ``inlineCallbacks``, and ``ensureDeferred`` freely.
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -18,6 +18,7 @@ Maintainer: Glyph Lefkowitz
 
 
 import attr
+import inspect
 import traceback
 import types
 import warnings
@@ -925,6 +926,48 @@ def ensureDeferred(coro):
 
     # Must be a Deferred
     return coro
+
+
+def coroFnToDeferredFn(corofn):
+    """
+    L{coroFnToDeferredFn} helps you define coroutinefunctions where previously
+    there were functions that returned Deferreds
+
+    Coroutine functions return a coroutine object, similar to how generators
+    work. This function turns that coroutinefunction into a Deferred function,
+    meaning that it can be used in legacy Twisted code. For example::
+
+        import treq
+        from twisted.internet import defer, task
+
+        async def crawl(pages):
+            results = {}
+            for page in pages:
+                results[page] = await treq.content(await treq.get(page))
+            return results
+
+        @defer.coroFnToDeferredFn
+        async def main(reactor):
+            pages = [
+                "http://localhost:8080"
+            ]
+            print(await crawl(pages))
+
+        task.react(main)
+
+    @param corofn: The coroutinefunction to wrap
+    @type coroutinefunction: An C{async def}
+
+    @rtype: typing.Callable
+    """
+
+    assert inspect.iscoroutinefunction(corofn)
+
+    @wraps(corofn)
+    def wrapper(*args, **kwargs):
+        return ensureDeferred(corofn(*args, **kwargs))
+
+    return wrapper
 
 
 class DebugInfo:
--- a/src/twisted/internet/test/test_defer_await.py
+++ b/src/twisted/internet/test/test_defer_await.py
@@ -9,7 +9,13 @@ Tests for C{await} support in Deferreds.
 import types
 
 from twisted.python.failure import Failure
-from twisted.internet.defer import Deferred, maybeDeferred, ensureDeferred, fail
+from twisted.internet.defer import (
+    Deferred,
+    maybeDeferred,
+    ensureDeferred,
+    fail,
+    coroFnToDeferredFn,
+)
 from twisted.trial.unittest import TestCase
 from twisted.internet.task import Clock
 
@@ -57,6 +63,34 @@ class AwaitTests(TestCase):
 
         # Now it's a Deferred.
         d = ensureDeferred(r)
+        self.assertIsInstance(d, Deferred)
+
+        # The Deferred has the result we want.
+        res = self.successResultOf(d)
+        self.assertEqual(res, "foo")
+
+    def test_coroFnToDeferredFn(self):
+        """
+        L{coroFnToDeferredFn} will wrap a coroutinefunction with a
+        function that returns a L{Deferred}.
+        """
+
+        @coroFnToDeferredFn
+        async def run():
+            d = Deferred()
+            d.callback("bar")
+            await d
+            res = await run2()
+            return res
+
+        async def run2():
+            d = Deferred()
+            d.callback("foo")
+            res = await d
+            return res
+
+        d = run()
+        # it's a Deferred.
         self.assertIsInstance(d, Deferred)
 
         # The Deferred has the result we want.
--- a/src/twisted/newsfragments/9972.feature
+++ b/src/twisted/newsfragments/9972.feature
@@ -1,0 +1,1 @@
+add twisted.internet.defer.coroFnToDeferredFn decorator for convenient use of async def functions


graingert:
I want some way to be able to be able progressively remove @coroFnToDeferredFn, eg

I think something like this
```python
class Deferred:
    _awaited = False
    def __init__(self, canceller=None, mustAwait=False):
         self._mustAwait = mustAwait
         ...

    def __await__(self):
        self._awaited = True
        return self

    def addCallbacks(self, callback, ...):
        if self.mustAwait and not self._awaited:
            raise DeprecationWarning(...)
        ...
```
        
```python
@coroFnToDeferredFn(mustAwait=True)  # would pass the flag to ensureDeferred and then to Deferred
async def foo():
    ...
```

graingert (commented):
.

docs/core/howto/defer-intro.rst
```diff
@@ -401,8 +401,10 @@ Coroutines with async/await
 
     .. versionadded:: 16.4
 
-On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`.
+On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.coroFnToDeferredFn <coroFnToDeferredFn>`.
+
 It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``, and is a function rather than a decorator.
```
graingert:
```suggestion
It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``.
```

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
...
+
+        task.react(main)
+
+    @param corofn: The coroutinefunction to wrap
+    @type coroutinefunction: A Python 3.5+ C{async def} C{coroutinefunction}
+
+    @rtype: typing.Callable
+    """
+
+    assert inspect.iscoroutinefunction(corofn)
```
graingert, line 964: intentionally not using asyncio.iscoroutinefunction to avoid supporting deprecated `@asyncio.coroutine`

src/twisted/internet/test/test_defer_await.py
```diff
@@ -63,6 +69,29 @@ async def run2():
         res = self.successResultOf(d)
         self.assertEqual(res, "foo")
 
+    def test_coroFnToDeferredFn(self):
```
adiroiban, line 72: This needs a docstring.

For the test, I was expecting to see an explicit triggering of "bar" callback.

So that we can have something like this

```
d = run()
# Check that d is a deferrred, but without a final result
bar_deferred.callback("bar")
# Check that `d` has now a result.
```

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
...
+def coroFnToDeferredFn(corofn):
+    """
+    L{coroFnToDeferredFn} helps you define coroutinefunctions where previously
+    there were functions that returned Deferreds
+
+    Coroutine functions return a coroutine object, similar to how generators
+    work. This function turns that coroutinefunction into a Deferred function,
+    meaning that it can be used in legacy Twisted code. For example::
+
+        import treq
```
adiroiban, line 940: can we have an example that is using only twisted and don't rely on an external library

```suggestion
        from twisted.web.client import getPage
```

src/twisted/internet/test/test_defer_await.py
graingert: > For the test, I was expecting to see an explicit triggering of "bar" callback.

I copied the previous test. I'd like to refactor all these tests to use `@coroFnToDeferred`

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
...
+        async def main(reactor):
+            pages = [
+                "http://localhost:8080"
+            ]
+            print(await crawl(pages))
+
+        task.react(main)
+
+    @param corofn: The coroutinefunction to wrap
+    @type coroutinefunction: A Python 3.5+ C{async def} C{coroutinefunction}
```
adiroiban: Since Twisted is supported only on Python 3.5 and up, I don't think that we need to mention Python3.5 here

src/twisted/internet/test/test_defer_await.py
graingert:
```suggestion
    def test_coroFnToDeferredFn(self):
        """
        L{coroFnToDeferredFn} will wrap a coroutinefunction with a
        function that returns a L{Deferred}.
        """
```

src/twisted/internet/defer.py
graingert: This is copied from the ensureDeferred docstring, and there's many many legacy mentions of old pythons. I think they should all be handled in one go
```suggestion
    @type coroutinefunction: An C{async def}
```

src/twisted/internet/defer.py
graingert: this is copied from ensureDeferred:

https://github.com/twisted/twisted/blob/a71a78145a59a4dea61a84d37c22bb738c9a4224/src/twisted/internet/defer.py#L883-L919

docs/core/howto/defer-intro.rst
```diff
@@ -401,8 +401,10 @@ Coroutines with async/await
 
     .. versionadded:: 16.4
 
-On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`.
-It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``, and is a function rather than a decorator.
+On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.coroFnToDeferredFn <coroFnToDeferredFn>`.
+
+It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``.
+"await" functionality can also be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`. It is similar to ``coroFnToDeferredFn``, except that it is a function rather than a decorator.
```
mthuurne, line 407: I think the decorator is important enough to come with an example, rather than being listed only as an aside.

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
     return coro
 
 
+def coroFnToDeferredFn(corofn):
```
mthuurne, line 931: I'm not very fond of the name `coroFnToDeferredFn`. It is accurate, but awkward to remember and type. I don't have a better name though; I'll give it some thought.

I think it's worth adding type annotations: even though mypy's support for decorators is rather weak at the moment, it would at least force the body to be type-checked.

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
...
+
+        task.react(main)
+
+    @param corofn: The coroutinefunction to wrap
+    @type coroutinefunction: An C{async def}
+
+    @rtype: typing.Callable
+    """
+
+    assert inspect.iscoroutinefunction(corofn)
```
mthuurne, line 964: Why not raise `TypeError` instead?

src/twisted/internet/defer.py
```diff
@@ -927,6 +928,48 @@ def main(reactor):
...
+
+    @wraps(corofn)
+    def wrapper(*args, **kwargs):
+        return ensureDeferred(corofn(*args, **kwargs))
```
mthuurne, line 968: You could call `_cancellableInlineCallbacks()` directly and skip over `ensureDeferred()` re-checking the type.

src/twisted/internet/defer.py
mthuurne: It would be worth putting that in a comment in the code, since someone maintaining this code is unlikely to stumble upon this note here on GitHub.

src/twisted/newsfragments/9972.feature
```diff
@@ -0,0 +1 @@
+add twisted.internet.defer.coroFnToDeferredFn decorator for convenient use of async def functions
```
mthuurne, line 1: I think it is customary to make this a full sentence, so starting with a capital and ending with a period.

graingert:
```suggestion
Add twisted.internet.defer.coroFnToDeferredFn decorator for convenient use of async def functions.
```

docs/core/howto/defer-intro.rst
```diff
@@ -401,8 +401,10 @@ Coroutines with async/await
 
     .. versionadded:: 16.4
 
-On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.ensureDeferred <ensureDeferred>`.
-It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``, and is a function rather than a decorator.
+On Python 3.5 and higher, the :pep:`492` ("Coroutines with async and await syntax") "await" functionality can be used with Deferreds by the use of :api:`twisted.internet.defer.coroFnToDeferredFn <coroFnToDeferredFn>`.
```
altendky, line 404: Maybe this should be ruled out of scope but we could drop the 3.5 caveat given that no Twisted with this documentation will run on anything less.

docs/core/howto/defer-intro.rst
```diff
@@ -401,8 +401,10 @@ Coroutines with async/await
...
+
+It is similar to ``inlineCallbacks``, except that it uses the ``await`` keyword instead of ``yield``, the ``return`` keyword instead of ``returnValue``.
```
altendky, line 406: Also maybe out of scope but `returnValue` isn't required.

docs/core/howto/defer-intro.rst
adiroiban: I have never used Twisted with py3 and I have never used ensureDeferred... so maybe for someone else this is obvious.

But to me it is not clear why ensureDeferred was removed from this documentation and why we need the new `coroFnToDeferredFn`

src/twisted/internet/defer.py
adiroiban: Maybe since this will be always applied to a corotine function, there is no need to add "coroFo' to the name.

Maybe name it:

* `@defer.coroutineWraper`
* `@defer.coroutineWrap`
* `@defer.deferCoroutine`

graingert: * `@defer.red`
* `@defer.redFn`

To mark the function as a red function

https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/

mthuurne: Other options suggested on IRC:
- `@defer.deferred`
- `@defer.returnDeferred`
- `@defer.applyEnsureDeferred`

I think `returnDeferred` has a good balance between being short & simple and describing what the decorator does.

graingert:
```
@(lambda f: lambda *args, **kwargs: ensureDeferred(f(*args, **kwargs)))
```
for Python 3.9

src/twisted/internet/defer.py
mthuurne: Another option would be using `Deferred.fromCoroutine()` once #1269 lands. That would still perform redundant type checks though, but one less than `ensureDeferred()` does.

src/twisted/internet/defer.py
wsanchez: `@coroutineFunctionToDeferredFunction` is fine by me, but I'm not afraid of long names.

Looks like [in my code](https://github.com/burningmantech/ranger-ims-server/blob/master/src/ims/ext/trial.py#L36) I've called it `asyncAsDeferred`.

src/twisted/internet/defer.py
wsanchez: Since you are in `defer.py` I would go with `_cancellableInlineCallbacks`, I think.

twm:
Why not teach `@inlineCallbacks` to handle coroutines? Then nobody has to change their decorators.

src/twisted/internet/defer.py
twm: `test_deferred.py` calls it `ensuringDeferred`.

I think it should be called `inlineCallbacks` though

adiroiban (changes requested):
What is the status of this PR?

is it possible to implement it via the suggestion from twm ?

> Why not teach @inlineCallbacks to handle coroutines? Then nobody has to change their decorators.

twm:
FWIW, I'd be happy to implement my suggestion if folks like it.

adiroiban:
I am +1 for having a smart `inlineCallbacks` that can handle non-generator functions.... but at the same time is not hard to use `inlineCorotine` or something like that.

I am not using  coroutines so I don't want to influence this decision.

adiroiban:
I think that we have already sent enough time deciding a name so if @graingert want to continue with this, pick any name and call it done :)

Also, if implementing this via a separate decorator is easier, I am fine with that :)

adiroiban:
@graingert think that this is no longer needed... so I am going to  close this as won't fix

https://freenode.logbot.info/twisted-dev/20210201#c6750701

We can re-open it at any time

graingert:
> @graingert think that this is no longer needed... so I am going to close this as won't fix
> 
> [freenode.logbot.info/twisted-dev/20210201#c6750701](https://freenode.logbot.info/twisted-dev/20210201#c6750701)
> 
> We can re-open it at any time

most of my usecase for this decorator would be resolved by https://github.com/twisted/twisted/pull/1399/files

I'm just getting stuck with the runWithWarningsSuppressed and stack manipulation

PR #1399, opened 2020-09-25
8977 Support async def coroutine functions in trial.

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/8977#comment:10
* [ ] I ran `pipx run black src/twisted` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)


src/twisted/trial/_asynctest.py
```diff
@@ -25,7 +26,13 @@
 from twisted.trial import itrial, util
 from twisted.trial._synctest import FailTest, SkipTest, SynchronousTestCase
 
-_wait_is_running = []  # type: List[None]
+_wait_is_running = []  # t_ype: List[None]
```
graingert:
```suggestion
_wait_is_running = []  # type: List[None]
```

src/twisted/trial/_asynctest.py
```diff
@@ -119,9 +138,8 @@ def onTimeout(d):
                 )
             )
             return defer.fail(exc)
-        d = defer.maybeDeferred(
-            utils.runWithWarningsSuppressed, self._getSuppress(), method
-        )
+
+        d = defer.ensureDeferred(self._runCorofnWithWarningsSuppressed)
```
graingert:
```suggestion
        d = defer.Deferred.fromCoroutine(self._runCorofnWithWarningsSuppressed(method))
```

src/twisted/internet/utils.py
```diff
@@ -191,6 +192,17 @@ def _resetWarningFilters(passthrough, addedFilters):
     return passthrough
 
 
+@contextlib.contextmanager
+def suppressWarningsCM(supressedWarnings):
+    for wargs, kwargs in supressedWarnings:
+        warnings.filterwarnings(*args, **kwargs)
```
graingert:
```suggestion
    for args, kwargs in supressedWarnings:
        warnings.filterwarnings(*args, **kwargs)
```

src/twisted/trial/_asynctest.py
```diff
@@ -81,36 +94,14 @@ def _eb(failure):
...
 
-        timeout = self.getTimeout()
-
-        def onTimeout(d):
-            e = defer.TimeoutError(
-                f"{self!r} ({methodName}) still running at {timeout} secs"
-            )
-            f = failure.Failure(e)
-            # try to errback the deferred that the test returns (for no gorram
-            # reason) (see issue1005 and test_errorPropagation in
```
graingert, line 93: https://twistedmatrix.com/trac/ticket/1005#comment:4

> It shouldn't errback the deferred, it should call .cancel on it.

src/twisted/trial/_asynctest.py
```diff
@@ -28,6 +29,18 @@
 _wait_is_running: List[None] = []
 
 
+async def _maybeCoroutine(*args, **kw):
+    try:  # simulate `async def _maybeCoroutine(func, /, *args, **kwargs):`
+        func, *args = args
+    except ValueError:
```
graingert:
```suggestion
    except ValueError:  # pragma: no cover
```

src/twisted/trial/_asynctest.py
```diff
@@ -119,12 +110,22 @@ def onTimeout(d):
...
-        d = defer.maybeDeferred(
-            utils.runWithWarningsSuppressed, self._getSuppress(), method
+
+        def _cancelledToTimedOutError(value, timeout):
+            if isinstance(value, failure.Failure):
+                value.trap(defer.CancelledError)
+                raise defer.TimeoutError(
+                    f"{self!r} ({methodName}) still running at {timeout} secs"
+                )
+            return value
```
graingert: ah this is important we should fail the test if the cancel is supressed

src/twisted/trial/test/detests.py
```diff
@@ -126,8 +126,15 @@ def test_passInlineCallbacks(self):
...
-        self._touchClass(None)
         yield None
+        self._touchClass(None)
+
+    async def test_passCoroutineFunction(self):
+        """
+        Test case that is a coroutine function
+        """
+        await defer.succeed(None)
+        self._touchClass(None)
```
graingert, line 137: this should definitely be covered

src/twisted/trial/_asynctest.py
```diff
@@ -81,36 +94,14 @@ def _eb(failure):
...
-            e = defer.TimeoutError(
-                f"{self!r} ({methodName}) still running at {timeout} secs"
-            )
-            f = failure.Failure(e)
-            # try to errback the deferred that the test returns (for no gorram
-            # reason) (see issue1005 and test_errorPropagation in
-            # test_deferred)
-            try:
-                d.errback(f)
-            except defer.AlreadyCalledError:
```
graingert, line 97: Instead of just crashing here we keep passing the cancellation down the chain of Deferred https://github.com/twisted/twisted/blob/release-21.2.0-10091/src/twisted/internet/defer.py#L737

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
...
+            """
+            A fake SynchronousTestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")
```
graingert:
```suggestion
                self.fail("this should never be reached")  # pragma: no cover
```

src/twisted/trial/test/test_tests.py
```diff
@@ -1417,6 +1417,43 @@ def test_generator(self):
...
+            """
+            A fake TestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")
```
graingert:
```suggestion
                self.fail("this should never be reached")  # pragma: no cover
```

src/twisted/trial/test/test_deferred.py
```diff
@@ -220,7 +229,29 @@ def test_errorPropagation(self):
         result = self.runTest("test_errorPropagation")
         self.assertFalse(result.wasSuccessful())
         self.assertEqual(result.testsRun, 1)
-        self._wasTimeout(detests.TimeoutTests.timedOut)
+        error = detests.TimeoutTests.timedOut
+        self.assertEqual(error.check(defer.CancelledError), defer.CancelledError)
```
graingert, line 232: This is *technically* a breaking change. Rather than throwing a TimeoutError into the test code, trial now uses the standard deferred timeout mechanism which throws a CancelledError instead.

However this only occurs at the point of failing a trial user's test anyway, and the old errback approach means this exception couldn't be caught without `.crash`ing the reactor which puts this change well outside the realm of supported behaviour

src/twisted/trial/reporter.py
```diff
@@ -497,35 +517,105 @@ def cleanupErrors(self, errs):
             BrokenTestCaseWarning,
         )
 
+    def _trimRunnerFrames(self, frames):
```
glyph: This literally never uses `self` once, so should probably be a free function.

src/twisted/trial/reporter.py
```diff
@@ -497,35 +517,105 @@ def cleanupErrors(self, errs):
...
+            ("_runCallbacks", "defer"),
+        ]
+
+        frame4 = frameInfo(frames[4])
+        if [frame0, frame1, frame2, frame3, frame4] == asyncFailureShimCvarCase:
+            return frames[5:]
+
+        return frames
+
     def _trimFrames(self, frames):
```
glyph: Gosh this function is getting enormously complicated, and it didn't start simple. I kind of want it to live in its own module and:

1. abstract this stuff a bit so you can arbitrarily say what "irrelevant" frames you want to throw away
2. shows you a `... X frames elided ...` message by default so you know *something* is going on
3. make a cmdline option for debugging weird cases that *doesn't* do any of this traceback stripping

src/twisted/trial/test/detests.py
```diff
@@ -208,6 +215,32 @@ def timedOut(err):
 
     test_errorPropagation.timeout = 0.1  # type: ignore[attr-defined]
 
+    async def test_cancelSuppression(self):
```
glyph, line 218: needs a docstring

src/twisted/trial/test/detests.py
```diff
@@ -208,6 +215,32 @@ def timedOut(err):
...
+            await sleep(0.2)
+        except defer.CancelledError:
+            await sleep(0.1)
+
+    test_cancelSuppression.timeout = 0.1  # type: ignore[attr-defined]
+
+    class MyError(Exception):
+        pass
+
+    async def test_cancelReplacement(self):
```
glyph, line 234: also needs a docstring

src/twisted/trial/test/detests.py
```diff
@@ -208,6 +215,32 @@ def timedOut(err):
...
+            await task.deferLater(clock=reactor, delay=delay)
+
+        try:
+            await sleep(0.2)
+        except defer.CancelledError:
+            await sleep(0.1)
+
+    test_cancelSuppression.timeout = 0.1  # type: ignore[attr-defined]
+
+    class MyError(Exception):
```
glyph, line 231: This seems like a weird place to put this - any particular reason it's here?

src/twisted/trial/test/detests.py
```diff
@@ -208,6 +215,32 @@ def timedOut(err):
...
+
+    async def test_cancelReplacement(self):
+        async def sleep(delay):
+            await task.deferLater(clock=reactor, delay=delay)
+
+        try:
+            await sleep(0.2)
+        except defer.CancelledError:
+            await sleep(0.1)
+            raise self.MyError("replaced error")
```
glyph, line 244: (Again why do we want to get this from `self`?)

src/twisted/trial/test/test_deferred.py
```diff
@@ -133,6 +133,15 @@ def test_passInlineCallbacks(self):
         self.assertEqual(result.testsRun, 1)
         self.assertTrue(detests.DeferredTests.touched)
 
+    def test_passCoroutineFunction(self):
+        """
+        The body of an async def test gets run.
```
glyph, line 137: something about the reactor here, maybe? completes when the coroutine runs to completion, runs with `Deferred.fromCoroutine` or something, i.e. it's not expecting a `Future`

src/twisted/trial/test/test_deferred.py
glyph: I don't think that's true. It's a change, but canceling a `Deferred` is never a breaking change - your caller could always do it if the `Deferred` is async.

src/twisted/trial/test/test_tests.py
```diff
@@ -1417,6 +1417,43 @@ def test_generator(self):
...
+            A fake TestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")  # pragma: no cover
+                yield
```
glyph: pragma: no cover needed here too, or maybe on the suite-starting colon, I think coverage reflects that

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
             "is a generator function and therefore will never run",
             result.errors[0][1].value.args[0],
         )
+
+    def test_synchronousTestCaseErrorOnAsyncGeneratorFunction(self):
+        """
+        In a SynchronousTestCase, a test method which is a generator function
+        is reported as an error, as such a method will never run assertions.
```
glyph, line 1495: Do we do this for other return value types (i.e. `Deferred`?)  Should we?

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
...
+            A fake SynchronousTestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")  # pragma: no cover
+                yield
```
glyph: need another no cover here too

src/twisted/trial/reporter.py
graingert: Pytest and weukzeug hide frames containing a `__tracebackhide__` or `__traceback_hide__` local

src/twisted/trial/test/detests.py
graingert: So I can check issubclass in the corresponding test case

src/twisted/trial/test/test_tests.py
graingert: Yeah Generators

src/twisted/trial/test/test_deferred.py
graingert: The test_passCoroutineFunction traps with a completed Deferred and so never interacts with the reactor

src/twisted/trial/test/detests.py
graingert:
```suggestion
    async def test_cancelSuppression(self):
        """Test case that ignores its timeout"""
```

src/twisted/trial/test/detests.py
graingert:
```suggestion
    async def test_cancelReplacement(self):
        """Test case that replaces its CancelledError"""
```

src/twisted/trial/test/detests.py
graingert: here: https://github.com/twisted/twisted/pull/1399/files#diff-5696cfe643e8b3e6e895831e510f76a90262ecac11a8f5e38e0d68c173d72ecdR253

src/twisted/trial/test/detests.py
graingert: Where else would I get it from? It's idiomatic to use self, but I can pull it from the class directly instead:

```suggestion
            raise __class__.MyError("replaced error")
```

src/twisted/trial/reporter.py
graingert: how about the `@_extraneous` decorator?

src/twisted/trial/test/test_tests.py
```diff
@@ -1417,6 +1417,43 @@ def test_generator(self):
...
+        In a TestCase, a test method which is an async generator function is reported
+        as an error, as such a method will never run assertions.
+        """
+
+        class AsyncGeneratorTestCase(unittest.TestCase):
+            """
+            A fake TestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
```
graingert:
```suggestion
            async def test_async_generator(self):  # pragma: no cover
```

src/twisted/trial/test/test_tests.py
```diff
@@ -1417,6 +1417,43 @@ def test_generator(self):
...
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")  # pragma: no cover
```
graingert, line 1434:
```suggestion
                self.fail("this should never be reached")
```

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
...
+            """
+            A fake SynchronousTestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")  # pragma: no cover
```
graingert, line 1508:
```suggestion
                self.fail("this should never be reached")
```

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
...
+        In a SynchronousTestCase, a test method which is a generator function
+        is reported as an error, as such a method will never run assertions.
+        """
+
+        class AsyncGeneratorSynchronousTestCase(unittest.SynchronousTestCase):
+            """
+            A fake SynchronousTestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):
```
graingert:
```suggestion
            async def test_async_generator(self):  # pragma: no cover
```

src/twisted/trial/test/test_tests.py
graingert: Trial doesn't check return types, only the function's flags

graingert: pytest does this: https://github.com/pytest-dev/pytest/blob/1ba5b485650d6f4ef8ef1ab6e2f782e2cf315ca8/src/_pytest/python.py#L181-L187

src/twisted/trial/newsfragments/8977.feature
```diff
@@ -0,0 +1 @@
+Support async def coroutine functions in trial.
```
graingert, line 1:
```suggestion
Support async def coroutine functions in trial.
``twisted.trial.unittest.TestCase`` now times out Deferreds using the built in cancel mechanism.
```

src/twisted/trial/reporter.py
```diff
@@ -350,6 +370,143 @@ def stopTest(self, test):
...
+
+    return newFrames
+
+
+def _formatFailureTraceback(*, detail, fail):
+    if isinstance(fail, str):
+        return fail.rstrip() + "\n"
+    fail.frames, frames = _trimFrames(fail.frames), fail.frames
+    result = fail.getTraceback(detail=detail, elideFrameworkCode=True)
+    fail.frames = frames
```
graingert:
```suggestion
    fail.frames, frames = _trimFrames(fail.frames), fail.frames
    try:
        result = fail.getTraceback(detail=detail, elideFrameworkCode=True)
    finally:
        fail.frames = frames
```

src/twisted/trial/reporter.py
graingert: @glyph I'm inclined to agree, eg. I'd like those 3 features also. However this frame sniffing is at my limit of understanding of Twisted internals and so I don't know how to proceed here.

src/twisted/trial/reporter.py
```diff
@@ -497,97 +654,8 @@ def cleanupErrors(self, errs):
...
-        asyncCase = (("maybeDeferred", "defer"), ("runWithWarningsSuppressed", "utils"))
-
-        twoFrames = ((firstMethod, firstFile), (secondMethod, secondFile))
-
-        # On PY3, we have an extra frame which is reraising the exception
-        for frame in newFrames:
-            frameFile = os.path.splitext(os.path.basename(frame[1]))[0]
-            if frameFile == "compat" and frame[0] == "reraise":
-                # If it's in the compat module and is reraise, BLAM IT
-                newFrames.pop(newFrames.index(frame))
```
graingert, line 560: compat.reraise never happens anymore

https://github.com/twisted/twisted/pull/1620

src/twisted/trial/reporter.py
glyph: For now let's just stick this into a separate file so that it's easier to come back to as an isolated bit of functionality in a follow-up.

src/twisted/trial/_frames.py
```diff
@@ -0,0 +1,167 @@
...
+    frame4 = frameInfo(frames[4])
+    if [frame0, frame1, frame2, frame3, frame4] == asyncFailureShimCvarCase:
+        return frames[5:]
+
+    return frames
+
+
+def _trimFrames(frames):
+    """
+    Trim frames to remove internal paths.
```
graingert, line 106:
```suggestion
    Trim frames to remove internal paths.
    
    TODO: https://twistedmatrix.com/trac/ticket/10220
```

src/twisted/trial/_asynctest.py
```diff
@@ -28,6 +29,18 @@
 _wait_is_running: List[None] = []
 
 
+async def _maybeCoroutine(*args, **kw):
```
graingert: this _maybeCoroutine is only used once and adds an extra frame, it's better to inline it

src/twisted/trial/reporter.py
graingert: I'm not going to mess with this right now

src/twisted/trial/test/test_tests.py
```diff
@@ -1417,6 +1417,43 @@ def test_generator(self):
...
+            A fake TestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):  # pragma: no cover
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")
+                yield
```
graingert:
```suggestion
                self.fail("this should never be reached")  # pragma: no cover
                yield  # pragma: no cover
```

src/twisted/trial/test/test_tests.py
```diff
@@ -1453,3 +1490,40 @@ def test_generator(self):
...
+            A fake SynchronousTestCase for testing purposes.
+            """
+
+            async def test_async_generator(self):  # pragma: no cover
+                """
+                A method which is also a generator function, for testing
+                purposes.
+                """
+                self.fail("this should never be reached")
+                yield
```
graingert:
```suggestion
                self.fail("this should never be reached")  # pragma: no cover
                yield  # pragma: no cover
```

src/twisted/internet/utils.py
```diff
@@ -191,6 +192,17 @@ def _resetWarningFilters(passthrough, addedFilters):
     return passthrough
 
 
+@contextlib.contextmanager
+def suppressWarningsCM(supressedWarnings):
```
adiroiban: Maybe name it
```suggestion
def suppressedWarnings(supressedWarnings):
```
then you can use it like

```python
with suppressedWarnings():
    some_code()
```

graingert: done

src/twisted/internet/utils.py
```diff
@@ -191,6 +192,17 @@ def _resetWarningFilters(passthrough, addedFilters):
...
 
+@contextlib.contextmanager
+def suppressedWarnings(supressedWarnings):
+    for args, kwargs in supressedWarnings:
+        warnings.filterwarnings(*args, **kwargs)
+    addedFilters = warnings.filters[: len(supressedWarnings)]
+    try:
+        yield
+    finally:
+        _resetWarningFilters(None, addedFilters)
```
graingert:
```suggestion
@contextlib.contextmanager
def suppressedWarnings(suppressedWarnings):
    for args, kwargs in suppressedWarnings:
        warnings.filterwarnings(*args, **kwargs)
    addedFilters = warnings.filters[: len(suppressedWarnings)]
    try:
        yield
    finally:
        _resetWarningFilters(None, addedFilters)
```

glyph (changes requested):
Needs conflicts resolved so I'm punting this out of the review queue for now :(.

exarkun:
Note this feature works (at least in some basic form) in trunk already.  The change to `maybeDeferred` which made it aware of coroutines caused trial to treat `async def` tests approximately the same as it treats tests that return a Deferred.

There are surely extra tests and documentation that should be added as well - just noting that the headline functionality exists in Twisted already.

PR #1401, opened 2020-09-25
Enable warn_unreachable for remaining files in twisted.internet.

wsanchez:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9976

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -17,19 +17,21 @@ Maintainer: Glyph Lefkowitz
 """
 
 
-import attr
+from asyncio import iscoroutine
+from functools import wraps
+from sys import exc_info, version_info
 import traceback
 import types
+from typing import Optional
 import warnings
-from asyncio import iscoroutine
-from sys import exc_info, version_info
-from functools import wraps
+
+import attr
 from incremental import Version
 
 # Twisted imports
-from twisted.python.compat import cmp, comparable
-from twisted.python import lockfile, failure
 from twisted.logger import Logger
+from twisted.python import lockfile, failure
+from twisted.python.compat import cmp, comparable
 from twisted.python.deprecate import warnAboutFunction, deprecated
 
 try:
@@ -264,7 +266,7 @@ class Deferred:
     # sets it directly.
     debug = False
 
-    _chainedTo = None
+    _chainedTo = None  # type: Optional[Deferred]
 
     def __init__(self, canceller=None):
         """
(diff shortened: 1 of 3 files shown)


src/twisted/internet/test/test_endpoints.py
```diff
@@ -1324,7 +1324,7 @@ def createServerEndpoint(self, reactor, factory, **listenArgs):
         address = IPv4Address("TCP", "0.0.0.0", 0)
 
         if listenArgs is None:
-            listenArgs = {}
```
wsanchez, line 1327: In order for `listenArgs` to be `None` here, it would have to be passed in as a keyword argument itself:

```python
>>> def createServerEndpoint(self, reactor, factory, **listenArgs):
...     pass
... 
>>> createServerEndpoint(None, None, None, listenArgs=None)
>>> 
```
 
It's a weird thing to do, and I don't see any uses of that in Twisted itself, but uh… ok.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1324,7 +1324,7 @@ def createServerEndpoint(self, reactor, factory, **listenArgs):
...
+            listenArgs = {}  # type: ignore[unreachable]
```
wsanchez: In order for `listenArgs` to be `None` here, it would have to be passed in as a keyword argument itself:

```python
>>> def createServerEndpoint(self, reactor, factory, **listenArgs):
...     pass
... 
>>> createServerEndpoint(None, None, None, listenArgs=None)
>>> 
```
 
It's a weird thing to do, and I don't see any uses of that in Twisted itself, but uh… ok.

src/twisted/internet/defer.py
```diff
@@ -17,19 +17,21 @@
...
+
+import attr
 from incremental import Version
 
 # Twisted imports
-from twisted.python.compat import cmp, comparable
-from twisted.python import lockfile, failure
 from twisted.logger import Logger
+from twisted.python import lockfile, failure
+from twisted.python.compat import cmp, comparable
```
wsanchez, line 34: `import Optional` is new. Rest is sorting.

src/twisted/internet/test/test_endpoints.py
mthuurne: I checked in Python 3.6 and 3.8 (and even 2.7, out of curiosity), and when called with `listenArgs=None`, the value of `listenArgs` will be `{'listenArgs': None}` instead of `None`. So I think mypy is correct and the code is actually unreachable.

rodrigc (approved):
This change is pretty straightforward.  Looks OK.

src/twisted/internet/test/test_endpoints.py
wsanchez: Oh, I didn't think to look at what actually got handed in!  OK, I'll remove these lines.

PR #1405, opened 2020-09-26
[9813] Fix maybeDeferred's handling of coroutines

twm:
When `maybeDeferred` executes a coroutine function it should run the resulting coroutine, rather than returning a coroutine object wrapped in a `Deferred`.

This implementation only changes `maybeDeferred`. Arguably `Deferred.callback()` and `Deferred.errback()` should guard against being called with coroutines in the say way they currently reject `Deferred`s. I'm interested in what reviewers think.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9813

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -145,7 +145,8 @@ def maybeDeferred(f, *args, **kw):
     Invoke a function that may or may not return a L{Deferred}.
 
     Call the given function with the given arguments.  If the returned
-    object is a L{Deferred}, return it.  If the returned object is a L{Failure},
+    object is a L{Deferred}, return it.  If the returned object is a coroutine,
+    execute it per L{ensureDeferred}.  If the returned object is a L{Failure},
     wrap it with L{fail} and return it.  Otherwise, wrap it in L{succeed} and
     return it.  If an exception is raised, convert it to a L{Failure}, wrap it
     in L{fail}, and then return it.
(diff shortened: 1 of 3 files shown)


graingert:
This will have strange behaviour in _asynctest and elsewhere because runWithWarningsSupressed checks if something is deferreds rather than Awaitable https://github.com/twisted/twisted/blob/twisted-20.3.0/src/twisted/internet/utils.py#L206

src/twisted/internet/defer.py
```diff
@@ -167,6 +168,8 @@ def maybeDeferred(f, *args, **kw):
 
     if isinstance(result, Deferred):
         return result
+    elif iscoroutine(result):
```
graingert, line 171: I think this is too dangerous in general, in a mixed asyncio/twisted application this could be a legitimate return value

Perhaps a warning would be better?

graingert (changes requested):

graingert:
This needs a test for trial _asynctest

src/twisted/internet/defer.py
```diff
@@ -145,7 +145,8 @@ def maybeDeferred(f, *args, **kw):
     Invoke a function that may or may not return a L{Deferred}.
 
     Call the given function with the given arguments.  If the returned
-    object is a L{Deferred}, return it.  If the returned object is a L{Failure},
+    object is a L{Deferred}, return it.  If the returned object is a coroutine,
+    execute it per L{ensureDeferred}.  If the returned object is a L{Failure},
```
wsanchez, line 149: per ➜ using

wsanchez, line 149: ensureDeferred ➜ fromCoroutine when we get #1269

PR #1410, opened 2020-09-30
Add type hints to twisted.internet.application.runner and twisted.internet.application.twist.

wsanchez:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9983


src/twisted/application/runner/_exit.py
```diff
@@ -6,35 +6,9 @@
 System exit support.
 """
 
+from enum import IntEnum
 from sys import stdout, stderr, exit as sysexit
-
-from constantly import Values, ValueConstant
-
-
-def exit(status, message=None):
```
wsanchez, line 14: Moved down after definition of `ExitStatus`, so we don't need a forward reference to it.

src/twisted/application/runner/_exit.py
```diff
@@ -66,72 +40,60 @@ class Status:  # type: ignore[no-redef]
         EX_CONFIG = EX__BASE + 14
 
 
-class ExitStatus(Values):
```
wsanchez, line 69: Used an `IntEnum` here instead of constantly because type support is far cleaner and we don't need anything from `constantly` here.

wsanchez: Note: compatibility should be OK since this is not yet public API.

src/twisted/application/runner/test/test_exit.py
```diff
@@ -18,43 +19,28 @@ class ExitTests(twisted.trial.unittest.TestCase):
...
-        """
-        self.assertRaises(ValueError, exit, "foo")
-
-    def test_exitStatusStringInt(self):
-        """
-        L{exit} given a L{str} status code that is a string integer passes the
-        corresponding L{int} to L{sys.exit}.
-        """
-        exit("1234")
-        self.assertEqual(self.exit.arg, 1234)
```
wsanchez, line 46: I can't remember why I allowed `str` values in the first place, but let's not.

mypy.ini
```diff
@@ -26,12 +26,45 @@ warn_unused_ignores      = True
...
+[mypy-twisted.application.twist._twist]
+allow_untyped_defs = True
+
+[mypy-twisted.application.twist.test.test_options]
+allow_untyped_defs = True
+check_untyped_defs = False
+
+[mypy-twisted.application.twist.test.test_twist]
+allow_untyped_defs = True
+check_untyped_defs = False
```
wsanchez: Longer list because instead of whole mode, we're excluding specific files, since we fixed some in the module.

src/twisted/application/runner/_pidfile.py
```diff
@@ -8,59 +8,60 @@
...
+from types import TracebackType
+from typing import Optional, Type
 
 from zope.interface import Interface, implementer
 
 from twisted.logger import Logger
+from twisted.python.filepath import FilePath
 
 
 class IPIDFile(Interface):
```
wsanchez, line 20: Since this isn't public API yet… question: should this be a [Protocol](https://mypy.readthedocs.io/en/stable/protocols.html?highlight=__exit__) or stay as a Zope Interface?

@glyph IIRC you said Zope last time I asked, but want to make sure before we publish this API.

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -417,83 +421,3 @@ def test_contextManager(self):
 
         with pidFile:
             pass
-
-
-@implementer(IFilePath)
-class DummyFilePath:
```
wsanchez, line 423: This was janky; we use a temp file now instead.

src/twisted/application/runner/test/test_runner.py
```diff
@@ -53,8 +62,8 @@ def setUp(self):
         # Patch globalLogBeginner so that we aren't trying to install multiple
         # global log observers.
 
-        self.stdout = BytesIO()
-        self.stderr = BytesIO()
+        self.stdout = StringIO()
+        self.stderr = StringIO()
```
wsanchez, line 66: In the `sys` module, mypy says these are `TextIO` not `BytesIO`.

src/twisted/application/runner/_pidfile.py
glyph: I still have a slight preference for Interface given the fact that many of these methods have quite generic names.  (I mean, it'd be hard to implement `writeRunningPID` by accident, but not impossible.)  But if anyone else has a strong opinion here I would be OK with either.  I don't see any `.providedBy` checks so either kind of works.

Thanks for checking in :-).

src/twisted/application/runner/test/test_runner.py
glyph: The situation was real ugly when you had to take py2 into account but as of py3 it's simple and straightforward: stdio streams are text, if you want bytes, you have to access their `.buffer` attributes.  Even separate form mypy, In the current docs it's quite clearly laid out: https://docs.python.org/3.8/library/sys.html#sys.stdin

src/twisted/application/runner/_exit.py
```diff
@@ -6,35 +6,9 @@
...
-    if message:
-        if code == 0:
-            out = stdout
-        else:
-            out = stderr
-        out.write(message)
-        out.write("\n")
-
-    sysexit(code)
+from typing import NoReturn, Optional, Union
```
mthuurne: [`typing.NoReturn`](https://docs.python.org/3/library/typing.html#typing.NoReturn) was added in Python 3.5.4 and 3.6.2. Some options to deal with this:
- bump Twisted's version requirements slightly
- import `NoReturn` from `typing_extensions` instead and add that package as a runtime dependency

Or if it would be acceptable to require 3.5.4/3.6.2+ only during type checking:
- catch `ImportError` and fall back to `None` if it's not found
- `import typing` and use `"typing.NoReturn"` (annotation as string)

src/twisted/application/runner/_exit.py
```diff
@@ -66,72 +40,60 @@ class Status:  # type: ignore[no-redef]
...
+def exit(status: Union[int, ExitStatus], message: Optional[str] = None) -> NoReturn:
+    """
+    Exit the python interpreter with the given status and an optional message.
+
+    @param status: An exit status. An appropriate value from L{ExitStatus} is
+        recommended.
+    @param message: An options message to print.
+    """
+    if message:
+        if status == 0:
```
mthuurne: Maybe this is outside the scope of this PR, but wouldn't it be clearer to compare to `EX_OK` instead of 0?

src/twisted/application/runner/_exit.py
```diff
@@ -66,72 +40,60 @@ class Status:  # type: ignore[no-redef]
...
+    EX_CONFIG = Status.EX_CONFIG
+
+
+def exit(status: Union[int, ExitStatus], message: Optional[str] = None) -> NoReturn:
+    """
+    Exit the python interpreter with the given status and an optional message.
+
+    @param status: An exit status. An appropriate value from L{ExitStatus} is
+        recommended.
+    @param message: An options message to print.
```
mthuurne:
```suggestion
    @param message: An optional message to print.
```

src/twisted/application/runner/_exit.py
```diff
@@ -66,72 +40,60 @@ class Status:  # type: ignore[no-redef]
...
+    EX_OSFILE = Status.EX_OSFILE
+    EX_CANTCREAT = Status.EX_CANTCREAT
+    EX_IOERR = Status.EX_IOERR
+    EX_TEMPFAIL = Status.EX_TEMPFAIL
+    EX_PROTOCOL = Status.EX_PROTOCOL
+    EX_NOPERM = Status.EX_NOPERM
+    EX_CONFIG = Status.EX_CONFIG
+
+
+def exit(status: Union[int, ExitStatus], message: Optional[str] = None) -> NoReturn:
```
mthuurne: If there is no backwards compatibility concern, perhaps you could perhaps drop `Optional`, so use `message: str = ""` instead. That might make it more clear that empty strings aren't printed.

src/twisted/application/runner/_pidfile.py
wsanchez: The main argument for `Protocol` is using the stdlib, but it's not like that's gong to change much for Twisted.  I'll leave it alone.

src/twisted/application/runner/_pidfile.py
```diff
@@ -81,7 +86,7 @@ def __exit__(excType, excValue, traceback):
 @implementer(IPIDFile)
 class PIDFile:
     """
-    Concrete implementation of L{IPIDFile} based on C{IFilePath}.
+    Concrete implementation of L{IPIDFile} based on C{FilePath}.
```
mthuurne: There is probably a good reason behind this, but it's not immediately obvious to me why there is a need here to use the concrete class rather than the interface.

src/twisted/application/runner/test/test_runner.py
wsanchez: Ah, I hadn't noticed that documentation, thanks

src/twisted/application/runner/_exit.py
wsanchez: Yes, thanks

src/twisted/application/runner/_exit.py
wsanchez: thanks

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -31,7 +30,7 @@
 from twisted.trial.unittest import SkipTest
 
 
-def ifPlatformSupported(f):
+def ifPlatformSupported(f: Callable[..., Any]) -> Callable[..., Any]:
```
mthuurne, line 33: You could make the return type of the `Callable` stricter using something like this:
```py
TestMethodRT = TypeVar("TestMethodRT", None, defer.Deferred)
```
I'm not sure it's worth the effort though.

src/twisted/application/runner/_exit.py
wsanchez: Nice, yes

src/twisted/application/runner/test/test_pidfile.py
```diff
@@ -75,182 +73,188 @@ class PIDFileTests(twisted.trial.unittest.TestCase):
...
         L{PIDFile.isRunning} raises L{StalePIDFileError} for a process that
         does not exist (errno=ESRCH).
         """
-        pidFile = PIDFile(DummyFilePath())
+        pidFile = PIDFile(self.filePath())
         pidFile._write(1337)
 
-        def kill(pid, signal):
+        def kill(pid: int, signal: int) -> None:
             raise OSError(errno.ESRCH, "No such process")
```
mthuurne, line 234: Maybe use `NoReturn` instead?

src/twisted/application/runner/_pidfile.py
wsanchez: It's not "based on" so much as "using".  I'll remove the "based on" clause there, since that's implementation detail.

src/twisted/application/runner/test/test_pidfile.py
wsanchez: This is a test module, no it's not public, and this is just to select the tests to run so… seems not worth it.  My pedantic self is tempted, but they it's mad that `Deferred` can't yet be stated as `Deferred[Any]` and so cannot be satisfied.

src/twisted/application/runner/test/test_pidfile.py
wsanchez: The thing it is emulating returns, though

src/twisted/application/runner/test/test_runner.py
```diff
@@ -273,56 +293,56 @@ def test_startReactorWithReactor(self):
...
-        @type callerName: L{str}
         """
         if callerName is None:
             callerName = methodName
 
         arguments = dict(a=object(), b=object(), c=object())
         argumentsSeen = []
 
-        def hook(**arguments):
+        def hook(**arguments: Any) -> None:
```
mthuurne: If you're not making any assumptions at all about the type, you could use `object` instead of `Any`.

wsanchez:
@mthuurne Thanks for the feedback… good stuff.  Sorry I didn't realize you were reviewing and I piled in some more files. :-/

src/twisted/application/runner/test/test_runner.py
```diff
@@ -398,19 +425,19 @@ class DummyKill:
...
 
 
 class DummyStandardIO:
     """
-    Stub for L{sys} which provides L{BytesIO} streams as stdout and stderr.
+    Stub for L{sys} which provides L{StringIO} streams as stdout and stderr.
     """
 
-    def __init__(self, stdout, stderr):
+    def __init__(self, stdout: TextIO, stderr: TextIO) -> None:
```
mthuurne, line 440: Should this be updated to `StringIO` as well?

wsanchez: `StringIO` is the concrete class; `TextIO` is the type.  (This is unlike `BytesIO` where the names match…)

src/twisted/application/runner/test/test_runner.py
wsanchez: Hrm.  I tend to use `Any` in those cases… I feel like `object` is being specific when I mean to be non-specific, no?

mthuurne:
> @mthuurne Thanks for the feedback… good stuff. Sorry I didn't realize you were reviewing and I piled in some more files. :-/

I didn't plan to do a full review, but I spotted one thing and then another and by then I thought I might as well skim the rest of the code as well. I went over the test code pretty quickly though.

src/twisted/application/runner/test/test_runner.py
mthuurne: You're right. I always use `IO[str]` myself, so I forgot about `TextIO`.

src/twisted/application/runner/_exit.py
mthuurne: If we want to use `Protocol` or `Literal` in the near future, `typing_extensions` would be the best solution.

src/twisted/application/runner/test/test_pidfile.py
mthuurne: As far as I know, `NoReturn` is compatible with any return type, so it shouldn't be an issue for mypy. But perhaps sticking to the emulated signature is indeed clearer.

src/twisted/application/runner/test/test_runner.py
mthuurne: The way I use them, `object` means "any object and I don't care what operations it implements", while `Any` means "I know something about this object that is too complex to express in Python's static type system".

The advantage of using `object` is that mypy will warn if you do make assumptions about supported operations later, while with `Any` mypy will assume that whatever you do is correct.

src/twisted/application/runner/_exit.py
glyph: Don't bother with `ImportError` - do an `if TYPE_CHECKING:` and only do the import in there. I'm totally fine with requiring more recent pythons for type checking.  I guess we have to live with quoted identifiers because 3.5 doesn't have `from __future__ import annotations` though?

mthuurne: Even [`TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) doesn't exist in 3.5.0 though; I learned today that it was added in 3.5.2.

A few weeks ago we had a 3.5.0 machine in the CI pool, but I don't know if that is still the case after the recent CI consolidation efforts.

`from __future__ import annotations` is available since 3.7, so it will be a while until we can use it, unless we decide to skip 3.6 and jump from 3.5 to 3.7 after the next release.

glyph: 3.5 is basically unused as I understand it, so I have no strong feelings about dropping it, but 3.6 is the most recent available pypy3 language version so that one's a bit of a non-starter :-(

glyph:
Just looking at the amount of comment churn this is generating, could we possibly split this into multiple PRs?  Maybe get rid of the 'and' in the subject and do one for each module?

src/twisted/application/runner/test/test_runner.py
wsanchez: Ah.  OK one reason I tend to use `Any` for `**kwargs` in particular is that I often pass those on to some other callable in that form, and if you use `object`, `mypy` might complain that said callable wants something else there.  I prefer to be consistent about that than to start with `object` expect if I actually use it then switch to `Any`.

Looking at this specific example, this really is a "whatever you gave me is fine" scenario; I'm just storing your crap in a list for you to look at later.  I have no opinion about that crap.

But then, the only expected operation on these things for the purposes of the test is… equality, which `object` has.  So… ok.

This reasoning is a bit subtle, so we should think about how to write it down.

wsanchez: It remains unclear to me what the best choice is [here](https://github.com/twisted/twisted/pull/1410/files#diff-6e519d82a6efe6b3cc078b899bbc320fL423), for example.  Still `object`, though I'm YOLO-emulating another API for testing?

wsanchez:
> Just looking at the amount of comment churn this is generating, could we possibly split this into multiple PRs? Maybe get rid of the 'and' in the subject and do one for each module?

Well the modules are pretty related, the changes are all the same sort, and the comment churn is churning away.

I'm not sure that having that churn in two places is an improvement…

src/twisted/application/runner/_exit.py
wsanchez: FWIW, all of our CI passed with this import in place…

wsanchez: …which means that whatever change we make to accommodate this will not be confirmed to work in our CI.

Which to me implies that our support for Python versions that break in this way is not actually a thing. Right?

glyph:
> Well the modules are pretty related, the changes are all the same sort, and the comment churn is churning away.
> 
> I'm not sure that having that churn in two places is an improvement…

If we can't land 'em independently, then please, carry on :).

src/twisted/application/runner/_exit.py
wsanchez: Ah, no `pypy3` is seeing this, OK.  I'll go ahead and fix.

wsanchez: Catching `ImportError` and falling back to `None` is problematic because `mypy` doesn't like `None`-or-`NoReturn` as a type. The forward reference works.

wsanchez:
OK all feedback integrated and CI is green

src/twisted/application/runner/_pidfile.py
```diff
@@ -91,26 +96,23 @@ class PIDFile:
     _log = Logger()
 
     @staticmethod
-    def _format(pid):
+    def _format(pid: int) -> bytes:
         """
         Format a PID file's content.
 
         @param pid: A process ID.
-        @type pid: int
```
adiroiban, line 99: Is it ok to remove pydoctor type annotation now?

I see in pydoctor that `Type annotations on function definitions are not supported yet`

mthuurne: They are supported in the `master` branch, but not yet in a release. I'll try to get a pydoctor release out soon.

src/twisted/application/twist/test/test_options.py
```diff
@@ -196,38 +203,43 @@ def test_logFileCantOpen(self):
...
 
         @param format: The format of the file.
-        @param expectedObserver: The expected observer.
+        @param expectedObserverFactory: The expected observer factory.
         """
         options = TwistOptions()
         options.opt_log_format(format)
 
-        self.assertIdentical(options["fileLogObserverFactory"], expectedObserver)
+        self.assertIdentical(options["fileLogObserverFactory"], expectedObserverFactory)
```
rodrigc, line 225: ok

rodrigc (approved):
This looks OK

PR #1415, opened 2020-10-01
Convert CircleCI build for wheels to GitHub Actions

rodrigc:
## Contributor Checklist:

* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9988 https://twistedmatrix.com/trac/newticket and replace this comment with the ticket number. -->


adiroiban (changes requested):
It looks good. Thanks.

But I think that as part of this PR we/you should also save the resulting wheels as artifacts.

In this way, if a branch is created for a release, the PR will create pre-release wheels which can be download from github and tested manually.

This should replace the step in which pre-release builds are manually created and uploaded to Twisted website.

-------

You can see "artifacts" download link/list as an example here: https://github.com/twisted/ldaptor/pull/198/checks    -> Artifacts 13

 This can be done with https://github.com/actions/upload-artifact

this can be done as a simple final step:

```
- uses: actions/upload-artifact@v2
  with:
    path: path/where/wheels/are/created
```

rodrigc:
@adiroiban At this link:

https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts#about-workflow-artifacts

It says that the retention policy for artifacts generated from a PR is 90 days.

However, do you know what the storage limits are for the Twisted project on GitHub?  I tried looking for the info at:

https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/managing-your-spending-limit-for-github-actions#about-spending-limits-for-github-actions

but could not find it.

adiroiban:
I would not worry about the file size.

I could not find info about file size limite for GitHub Actions...but I expect to be similar to Azure - https://docs.microsoft.com/en-us/azure/devops/artifacts/reference/limits?view=azure-devops

And to implement only the generic 1TB archive.

--------

90 days should be more enough than. 

The artifacts will be used only for manual pre-release/PR alpha-stage testing.

------

The official pre-released will be published to PyPi anyway and will be there forever.

The (pre)release process will be something like this:

* when a tag is created build using parallel jobs all the wheels and upload them as artifacts
* Have a final job which will wait for all the wheel jobs, download the artifacts and will publish them to Pypi

adiroiban:
The final job for publishing to pypi can be like this.

```
  all-successful:
    runs-on: ubuntu-latest
    needs: [build_wheel]
    steps:
      - name: Download dists for PyPI
        uses: actions/download-artifact@v2
        with:
          path: dist

      - name: Display structure of downloaded files
        run: ls -R

      - name: Publish to PyPI for a new tag
        if: startsWith(github.ref, 'refs/tags')
        uses: pypa/gh-action-pypi-publish@v1.3.1
        with:
          password: ${{ secrets.PYPI_GITHUB_PACKAGE_UPLOAD }}
```

The publish should only be done on a tag and we should ask @glyph or someone else to create a PyPI token and save it to GItHub Twisted repo secrets as PYPI_GITHUB_PACKAGE_UPLOAD

rodrigc:
OK, I got artifact uploading to work.  If you go to https://github.com/twisted/twisted/pull/1415/checks?check_run_id=1198769232 , and click on the **Artifacts** dropdown on the right, you can see them.

adiroiban:
Why not remove the while circle-ci config?
Why do we still need circle-ci ?

Would you consider adding the `all-successful:` job to this PR.
In this way, we can check that we can also download all the wheels in a single place and have them ready for a future push to Pypi.

Thanks!

adiroiban:
This PR can be merged as it is... but I have a question.
Does it make any sense to distribute all the twisted test files inside the wheel package?

My understanding is that wheels are to be used for production.
At least on Windows, I see that installing/updating a python package with many files takes considerable time.

1600 non test files and 400 test files... so 20% of the files are test files.

```sh
$ ls -Ral src  | grep -v test | wc -l
1678

$ ls -Ral src  | grep test | wc -l
441
```

rodrigc:
I removed the CircleCI files, and the build failed with this: https://app.circleci.com/pipelines/github/twisted/twisted/1240/workflows/b382b01c-6dd3-437c-a4b1-7fdbe5043d70/jobs/11273

How do we shut off CircleCI?

rodrigc:
To address your question, I think it is OK to leave the wheel as-is, and address any packaging issues
in a separate PR/discussion.  The Twisted package already packages a lot of stuff as-is,
and cutting things down will definitely involve some thought and discussion, because you never know what workflows out there people have built up around the existing packaging.

adiroiban (approved):
Looks good. Thanks!

Looking forward to having the wheels automatically pushed to PyPI.

I went to Circle-CI and clicked "Stop building"

image

rodrigc:
@adiroiban can you recommend how to restructure the dependencies in the GitHub Actions so that they run something like this:

```
+---------------+
|  manifest     +------------>+                               +-------->+--------+
|  checker      |             |                               |         |  tox test
+---------------+             |                               |         |        |
                              |                               |         +--------+
                              |                               |
                              |                               |
+---------------+             |                               |
|   mypy        +------------^+       +-------------+         |         +--------+
|               |            +------->+ Build wheels|         +-------->+ tox    |
+---------------+             |       |             +-------->+         | test   |
                              |       +-------------+         |         +--------+
+---------------+             |                               |
|    lint       |             |                               |
|               +------------>+                               |
|               |                                             |         +--------+
+---------------+                                             +-------->+tox     |
                                                                        |test    |
                                                                        +--------+
```

I think [jobs.<job_id>.needs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) is the feature that can do it.

rodrigc:
Also, what do you think of using the [releases](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release) event in GitHub Actions for doing a release related workflow?  That workflow would include uploading artifacts to PyPI...

glyph:
> Also, what do you think of using the [releases](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release) event in GitHub Actions for doing a release related workflow? That workflow would include uploading artifacts to PyPI...

Personally I love this idea (dedicated UI and discrete events for release versions? yes please!) but I have no experience with it.

adiroiban:
I can help with GitHub Actions integration and automated or manual triggering the automated release process.

I thing that tox is already or can automatically build a wheel and run tests based on that wheel.

so I think the process should look like

```
mypy and other static checkers  ->
tox code tests based on wheel  ->  on success -> Use the wheels created by tox as artifacts -> publish to pypi 
```

but, instead of doing all this with twisted/twisted , maybe we can create a pet project inside  the twisted to experiment with this release process.

In this way, we will not trigger a huge number of tests and we can get a quick feedback.

GitHub Actions has a nice UI to manually trigger a worklow and allows you to create custom inputs. for example:

image

But I think that putting all the release process together should be done in a separate repo that we can use for testing and learning how to do it.

adiroiban:
But yes, rather than doing custom triggers, we can use the standard "New release"  page at https://github.com/twisted/twisted/releases/new

So yes +1 for using the release GUI

rodrigc:
OK, I will try and get a **release** workflow going.
I just learned that GitHub Actions implemented something in August called [Composite Run Steps](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/).
Here is an example: https://github.com/Catrobat/Catroweb/pull/824/files

So I think I can refactor some of the actions to have better dependencies, and also have a release workflow.

adiroiban:
Great. I am working at https://twistedmatrix.com/trac/ticket/9994 which I hope can help with having the docs available for a release.

I think that the main tasks for a release are:
* make sure all tests are green
* publish wheels to PyPI
* publish source distribution to PyPI
* publish narrative documentation
* publish API documentation

Right now, for download we have the following pages:
* WIKI - https://twistedmatrix.com/trac/wiki/Downloads
* Twisted Web static file - https://twistedmatrix.com/Releases/

I guess that the wiki page can be updated to inform the release is now only on PyPI.

For the static page... maybe upload a README to inform that latest versions are on PYPI.

Later we might setup a cron on TwistedMatrix VM to sync/pull the files into the archive. In this way we can have a backup/mirror.

glyph:
> On Oct 3, 2020, at 3:43 PM, Adi Roiban <notifications@github.com> wrote:
> 
> 
> Great. I am working at https://twistedmatrix.com/trac/ticket/9994 <https://twistedmatrix.com/trac/ticket/9994> which I hope can help with having the docs available for a release.
> 
Amazing. So hype for this!
> I think that the main tasks for a release are:
> 
> make sure all tests are green
> publish wheels to PyPI
> publish source distribution to PyPI
> publish narrative documentation
> publish API documentation
I think so!  I would put "publish source distribution" first if the order is significant here, since that's the "lower level" artifact from which wheels can be derived.
> Right now, for download we have the following pages:
> 
> WIKI - https://twistedmatrix.com/trac/wiki/Downloads <https://twistedmatrix.com/trac/wiki/Downloads>
> Twisted Web static file - https://twistedmatrix.com/Releases/ <https://twistedmatrix.com/Releases/>
> I guess that the wiki page can be updated to inform the release is now only on PyPI.
> 
> For the static page... maybe upload a README to inform that latest versions are on PYPI.
> 
For the record, I'm also in favor of this.

In the interest of avoiding Chesterton's fence, here's some brief historical context:

When this release infrastructure was set up, PyPI did not exist; then, for a long time, it was unreliable.  During this period, Twisted could be installed with nothing but Python.  Nowadays, even a minimally-functional Twisted requires half a dozen dependencies, and you probably need a lot more than that to build anything useful.  As a result, practically speaking, you can't currently install Twisted without reliable access to the rest of PyPI.  So there's no point in wasting effort maintaining these archives to pretend that the project is self-sufficient.
> Later we might setup a cron on TwistedMatrix VM to sync/pull the files into the archive. In this way we can have a backup/mirror.
> 
In the interests of simplicity, I'd say let's not do this until we need it for some specific reason.  I don't object, but I also don't think it's going to solve much of a real problem.

PR #1416, opened 2020-10-01
Update dependencies in tox.ini

wsanchez:
Mostly looking to update `mypy`, but `mypy-zope` and `cibuildwheel` also need bumps.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9989


wsanchez:
Note that in CI, we are getting this error from Mypy:

```console
src/twisted/internet/kqreactor.py:23:1: error: Module 'select' has no attribute 'KQ_FILTER_READ'  [attr-defined]
8
src/twisted/internet/kqreactor.py:23:1: error: Module 'select' has no attribute 'KQ_FILTER_WRITE'  [attr-defined]
9
src/twisted/internet/kqreactor.py:24:1: error: Module 'select' has no attribute 'KQ_EV_DELETE'  [attr-defined]
10
src/twisted/internet/kqreactor.py:24:1: error: Module 'select' has no attribute 'KQ_EV_ADD'  [attr-defined]
11
src/twisted/internet/kqreactor.py:24:1: error: Module 'select' has no attribute 'KQ_EV_EOF'  [attr-defined]
```

I have added an annotation to ignore this, but then on macOS, this adds an error:

```console
src/twisted/internet/kqreactor.py:23: error: unused 'type: ignore' comment
```

Conversely, on macOS, we see this error:

```console
src/twisted/internet/epollreactor.py:15:1: error: Module 'select' has no attribute 'epoll'; maybe "poll"?  [attr-defined]
src/twisted/internet/epollreactor.py:15:1: error: Module 'select' has no attribute 'EPOLLHUP'; maybe "POLLHUP"?  [attr-defined]
src/twisted/internet/epollreactor.py:15:1: error: Module 'select' has no attribute 'EPOLLERR'; maybe "POLLERR"?  [attr-defined]
src/twisted/internet/epollreactor.py:15:1: error: Module 'select' has no attribute 'EPOLLIN'; maybe "POLLIN"?  [attr-defined]
src/twisted/internet/epollreactor.py:15:1: error: Module 'select' has no attribute 'EPOLLOUT'; maybe "POLLOUT"?  [attr-defined]
```

And I can't ignore that error without creating a complaint about the unused ignore in CI.

[I made mention of this in Mypy's issue tracker.](https://github.com/python/mypy/issues/8823#issuecomment-702422255)

wsanchez:
OK Mypy [will notice platform checks](https://mypy.readthedocs.io/en/stable/common_issues.html?highlight=platform#python-version-and-system-platform-checks) using conditionals on `sys.platform`, so using that.

src/twisted/internet/kqreactor.py
```diff
@@ -16,12 +16,20 @@
    kqreactor.install()
 """
 
+import sys
+
+# For mypy's benefit
+assert not sys.platform.startswith("linux")
```
wsanchez: I can't say I really like this, since it assumes that no Linux version supports kqueue, and I don't know if that's right…

I tried this for the epollreactor.py on macOS, but it doesn't suppress the error:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    import sys
    assert sys.platform != "darwin"
```

So I can't figure out how to only make the assertion for mypy.

rodrigc: I think that's a pretty safe bet.  kqueue() started on FreeBSD ( https://people.freebsd.org/~jlemon/papers/kqueue.pdf ), and was later ported to other BSD operating systems including Darwin.

However, I don't like to start putting the equivalent of the old C way of doing platform specific #ifdef's if we don't have to, to avoid breaking some oddball platform that may use Twisted.

Is it possible to appease mypy by doing something like:

```
import select
assert 'epoll' in dir(select)
```

or:

```
import select
assert 'kqueue' in dir(select)
```

On Linux, `dir(select)` has `epoll`, but not `kqueue`,
while on MacOS, `dir(select)` has `kqueue` but not `epoll`

Nayakaditya:
again and again it show that wheater i've tried lots of time this pip install twisted
even after it getting show error.

ImportError : no modules named twisted.web 

please help

plz plz plz plz plz plz plz plz plz plz

src/twisted/internet/kqreactor.py
wsanchez: @rodrigc: nah, didn't work :-(

I guess we can leave this `sys.platform` hack in place for now.  If we were hypothetically to end up with `kqueue` on Linux, this assertion would trip in CI and maybe by then we'll have better options.

wsanchez: Turns out even that wasn't working, and I was about to give up, but @graingert to the rescue with a suggestion to use `getattr` (though he was talking about #1427, but I'll take it!).

wsanchez:
OK all set

rodrigc (approved):
OK, this is better than the platform-specific checks that were in the original PR.

PR #1417, opened 2020-10-02
Implement AMP version 2

jothan:
Implement AMP v2 as described here: https://amp-protocol.net/conversations_v2.html

## Contributor Checklist:

* [x] The associated tickets in Trac are here:
 * https://twistedmatrix.com/trac/ticket/2529
 * https://twistedmatrix.com/trac/ticket/9488


src/twisted/protocols/amp.py
```diff
@@ -657,7 +658,7 @@ def copy(self):
         newBox.update(self)
         return newBox
 
-    def serialize(self):
+    def serialize(self, version2=False):
```
rodrigc: Can you add type annotations for any new functions which you have added or modified?  The Twisted tree is slowly moving to add type annotations

src/twisted/protocols/amp.py
```diff
@@ -2613,6 +2634,35 @@ def connectionLost(self, reason):
         self.transport = None
 
 
+class AMPv2(AMP):
+    def _write_box(self, box):
```
rodrigc: Annotate?

src/twisted/protocols/amp.py
```diff
@@ -2613,6 +2634,35 @@ def connectionLost(self, reason):
...
+        self.transport.write(box.serialize(version2=True))
+
+    def proto_value(self, string):
```
rodrigc: Annotate?

src/twisted/protocols/amp.py
```diff
@@ -2613,6 +2634,35 @@ def connectionLost(self, reason):
...
+        if len(string) < MAX_VALUE_LENGTH:
+            self._currentBox[self._currentKey] = string
+            self._currentKey = None
+            self.MAX_LENGTH = self._MAX_KEY_LENGTH
+            return "key"
+
+        self._currentValue = [string]
+        return "valuecont"
+
+    def proto_valuecont(self, string):
```
rodrigc: Annotate?

src/twisted/test/test_amp.py
```diff
@@ -95,6 +95,14 @@ def amp_HELLO(self, box):
         return amp.Box(hello=box[b"hello"])
 
 
+class SimpleSymmetricProtocol2(amp.AMPv2):
+    def sendHello(self, text):
```
rodrigc: Annotate?

src/twisted/test/test_amp.py
```diff
@@ -95,6 +95,14 @@ def amp_HELLO(self, box):
...
+        return self.callRemoteString(b"hello", hello=text)
+
+    def amp_HELLO(self, box):
```
rodrigc: Annotate?

rodrigc (changes requested):
Thanks for the contribution

jothan:
@rodrigc I have added the type annotations as you requested. May I resolve the comments you added ?

src/twisted/protocols/amp.py
```diff
@@ -2613,6 +2634,35 @@ def connectionLost(self, reason):
         self.transport = None
 
 
+class AMPv2(AMP):
```
glyph, line 2646: This class and all added methods here need docstrings.

rodrigc:
@jothan At the top-level Twisted directory, you can run `tox -e mypy` to check things before mypy before pushing to GitHub.  That will save a few cycles.

jothan:
@rodrigc Thanks for the tip. However, I'm a bit stumped by the remaining errors, do you have any tips ?

src/twisted/protocols/amp.py
```diff
@@ -657,15 +658,15 @@ def copy(self):
...
+    def serialize(self, version2: bool = False) -> bytes:
         """
         Convert me into a wire-encoded string.
 
         @return: a C{bytes} encoded according to the rules described in the
             module docstring.
         """
         i = sorted(self.items())
-        L = []
+        L: List[bytes] = []
```
rodrigc: This needs to be:

```
L = []  # type: List[bytes]
```

because Twisted still supports Python 3.5 which does not support the new syntax for type annotations on variables.  Hopefully by the end of the year, Twisted will be able to drop Python 3.5 support.

jothan:
Alright, I figured out which type annotations I needed, mypy is now happy.

rodrigc (changes requested):

rodrigc:
Thanks for being patient and submitting the annotations.  This is slowly making the Twisted codebase better, and easier to understand.

src/twisted/protocols/amp.py
jothan: Ok for this change, but how do I fix this for Python 3.5 ?

```python
class AmpBox(Dict[bytes, bytes]):
```

jothan:
@rodrigc I totally understand the need for those annotations, especially in a large code base.

My AMP implementation is in Rust with types everywhere ;P https://github.com/jothan/amp-async

src/twisted/protocols/amp.py
jothan: Oh, Python 3.5 seems perfectly happy with that class declaration.

rodrigc (approved):
I think this looks OK, but I'll leave open for another reviewer to give it a once over.

Nice to see you have experience writing AMP protocol parsers in Rust.  I need to learn Rust one of these days. :)

src/twisted/newsfragments/9488.feature
```diff
@@ -0,0 +1 @@
+Implement the AMP version 2 protocol to allow values greater than 64k.
```
glyph, line 1: We shouldn't have duplicate NEWS entries like this, as it produces useless duplicative noise in the changelog.  The correct thing to do here is to close the second ticket as a duplicate in trac, then have there be one NEWS file.

src/twisted/protocols/amp.py
```diff
@@ -239,6 +239,7 @@ def trapZero(result):
 
 __all__ = [
     "AMP",
+    "AMPv2",
```
glyph, line 242: This should not be exposed as a separate class, but rather a protocol-version parameter.

src/twisted/protocols/amp.py
```diff
@@ -608,7 +609,7 @@ class IncompatibleVersions(AmpError):
 PROTOCOL_ERRORS = {UNHANDLED_ERROR_CODE: UnhandledCommand}
 
 
-class AmpBox(dict):
+class AmpBox(Dict[bytes, bytes]):
```
glyph, line 612: We really should have landed type annotations as a separate, prior change. They're not related to "AMPv2".

src/twisted/protocols/amp.py
```diff
@@ -657,15 +658,15 @@ def copy(self):
         newBox.update(self)
         return newBox
 
-    def serialize(self):
+    def serialize(self, version2: bool = False) -> bytes:
```
glyph, line 661: This "version2" parameter needs a `@param` in the docstring.

src/twisted/protocols/amp.py
```diff
@@ -674,11 +675,29 @@ def serialize(self):
...
-                w(kv)
+
+            w(pack("!H", len(k)))
+            w(k)
+
+            if version2:
+                vio = BytesIO(v)
+                del v
+
+                # If the value is an exact multiple of 65535, the last
```
glyph, line 688: how is the value "an exact multiple" of 65535 if it is not 65535 itself?

src/twisted/protocols/amp.py
```diff
@@ -674,11 +675,29 @@ def serialize(self):
...
+                # chunk containing data will be followed by a zero-length chunk
+                # (i.e. when read() returns EOF).
+                chunk = vio.read(MAX_VALUE_LENGTH)
+                while True:
+                    w(pack("!H", len(chunk)))
```
glyph, line 693: I realize that this is coming from the "spec" on amp-protocol.net, but there were never any discussions I saw about how to implement a V2.  Given that this is neither backwards compatible nor has any wire-level negotiation it seems like a pretty bad feature for a version bump.

glyph: (A better way to have done this would be to normalize the meaning of repeated keys, but I digress.)

src/twisted/protocols/amp.py
```diff
@@ -868,13 +889,17 @@ def _sendBoxCommand(self, command, box, requiresAnswer=True):
...
             result = self._outstandingRequests[tag] = Deferred()
-        else:
-            result = None
+
         return result
 
-    def callRemoteString(self, command, requiresAnswer=True, **kw):
+    def callRemoteString(
+        self, command: bytes, requiresAnswer: bool = True, **kw: bytes
+    ) -> Optional[Deferred]:
```
glyph, line 902: `Optional[Deferred]` is the wrong way to specify this signature; what you want is an `@overload` to correlate the value of `requiresAnswer` to the type of the response.

src/twisted/protocols/amp.py
```diff
@@ -2613,6 +2643,52 @@ def connectionLost(self, reason):
...
+    """
+
+    _currentValue = None  # type: Optional[List[bytes]]
+
+    def _write_box(self, box: AmpBox) -> None:
+        """Send an AmpBox using the AMPv2 format."""
+        assert self.transport is not None
+        self.transport.write(box.serialize(version2=True))
+
+    def proto_value(self, string: bytes) -> str:
```
glyph, line 2658: These methods do not belong on an `AMP` subclass, if they belong on a subclass at all, but a `BinaryBoxProtocol` subclass.

There's a mea-culpa here too; the inheritance-for-convenience design of AMP was a mistake, and should not have created this dichotomy; but, this is also why this should not be a new class, though, and instead just a parameter to `BinaryBoxProtocol` that `AMP` can propagate.

glyph (commented):
I'm sorry that I didn't have time to get in a code review before this landed, but my attention was brought to it and I have some post-hoc feedback.

src/twisted/newsfragments/9488.feature
jothan: You might want to update the guidelines in this case, from the news fragment instructions: " You should replace <ticket number> with the ticket number which is being resolved by the change (if multiple tickets are resolved, multiple files with the same contents should be added). "

src/twisted/protocols/amp.py
jothan: I know, sorry for the noise, I wanted to follow trunk closely with tests passing.

src/twisted/protocols/amp.py
jothan: Well, it is _mostly_ backwards compatible as long as values are smaller than 65534 bytes. Considering it is a breaking change, it should be signaled via a compatibility version change. I would have called AMP v1.1, but AMP v2 already existed conceptually.

src/twisted/protocols/amp.py
jothan: Guess I should remove the word "exact" here.

src/twisted/protocols/amp.py
jothan: Yeah, I got hit by that. My initial unpublished patch was version parameters everywhere.

glyph:
@jothan in case you are not on the Twisted mailing list, I began a discussion about this here: https://twistedmatrix.com/pipermail/twisted-python/2020-October/065288.html

I'm leaning towards doing a revert based on that.

Thanks for your contribution to Twisted, sorry to burden it with more work!

src/twisted/newsfragments/9488.feature
glyph: @jothan Hmm. Apparently I'm wrong about this.  I thought that this was an idiom that predated our usage of towncrier, but apparently towncrier adopted this same idiom of hashing the contents of the files; I just created a sample project to test it out and it does the right thing in all the edge cases that I thought would not be handled.  Sorry for the mistake!  I am still probably going to be rolling this back out for the other reasons discussed below, but at least we can forget about this...

glyph:
@jothan Would you mind re-submitting just the type-annotation bits of this PR first, since they're not really related to the other stuff happening here, and would be useful to integrate on their own?

src/twisted/protocols/amp.py
glyph: If you have a key which is just a bit over 64k bytes, you end up screwing up the alternation of keys and values, which means that the failure mode involves the peer hanging forever; that's a pretty bad failure mode.

There's a way to do this which is *entirely* backwards compatible, though; just send the same key repeatedly, or send a key with `.1`, `.2`, etc appended to it.  This ends up adding 0.02% or so worth of overhead to the data (and if you have really large data, you want what's described on https://twistedmatrix.com/trac/ticket/2529 anyway, not just a way to circumvent the length-prefix limit).

You can also do this as a new parameter-type declaration, which also allows you to interrogate your peer's capabilities by checking for an `UNHANDLED` response if you have something you want to upgrade.

If we wanted to do something dramatically better, [netstrings](https://en.wikipedia.org/wiki/Netstring) are the way to go; arguably AMP should have used these in the first place.

PR #1420, opened 2020-10-03
[9947] smb 4: vfs layer

ihaywood3:
https://twistedmatrix.com/trac/ticket/9947


ihaywood3:
will be replaced with simpler PR

PR #1426, opened 2020-10-06
Add type hints for twisted.logger

wsanchez:
Add type hints for `twisted.logger`

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9995

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -93,10 +93,6 @@ allow_incomplete_defs = True
 [mypy-twisted.internet.testing]
 allow_incomplete_defs = True
 
-[mypy-twisted.logger.*]
-allow_untyped_defs = True
-check_untyped_defs = False
-
 [mypy-twisted.mail.*]
 allow_untyped_defs = True
 check_untyped_defs = False
--- a/src/twisted/logger/_filter.py
+++ b/src/twisted/logger/_filter.py
@@ -7,13 +7,15 @@ Filtering log observer.
 """
 
 from functools import partial
+from typing import Dict, Iterable
 
 from zope.interface import Interface, implementer
 
 from constantly import NamedConstant, Names
 
+from ._interfaces import ILogObserver, LogEvent
 from ._levels import InvalidLogLevelError, LogLevel
-from ._observer import ILogObserver
+from ._observer import bitbucketLogObserver
 
 
 class PredicateResult(Names):
@@ -46,7 +48,7 @@ class ILogFilterPredicate(Interface):
     A predicate that determined whether an event should be logged.
     """
 
-    def __call__(event):
+    def __call__(event: LogEvent) -> NamedConstant:
         """
         Determine whether an event should be logged.
 
@@ -54,7 +56,7 @@ class ILogFilterPredicate(Interface):
         """
 
 
-def shouldLogEvent(predicates, event):
+def shouldLogEvent(predicates: Iterable[ILogFilterPredicate], event: LogEvent) -> bool:
     """
     Determine whether an event should be logged, based on the result of
     C{predicates}.
@@ -70,13 +72,9 @@ def shouldLogEvent(predicates, event):
     run out, at which point we return C{True}.
 
     @param predicates: The predicates to use.
-    @type predicates: iterable of L{ILogFilterPredicate}
-
     @param event: An event
-    @type event: L{dict}
 
     @return: True if the message should be forwarded on, C{False} if not.
-    @rtype: L{bool}
     """
     for predicate in predicates:
         result = predicate(event)
@@ -97,31 +95,31 @@ class FilteringLogObserver:
     based on applying a series of L{ILogFilterPredicate}s.
     """
 
-    def __init__(self, observer, predicates, negativeObserver=lambda event: None):
+    def __init__(
+        self,
+        observer: ILogObserver,
+        predicates: Iterable[ILogFilterPredicate],
+        negativeObserver: ILogObserver = bitbucketLogObserver,
+    ) -> None:
         """
         @param observer: An observer to which this observer will forward
             events when C{predictates} yield a positive result.
-        @type observer: L{ILogObserver}
-
         @param predicates: Predicates to apply to events before forwarding to
             the wrapped observer.
-        @type predicates: ordered iterable of predicates
-
         @param negativeObserver: An observer to which this observer will
             forward events when C{predictates} yield a negative result.
-        @type negativeObserver: L{ILogObserver}
         """
         self._observer = observer
         self._shouldLogEvent = partial(shouldLogEvent, list(predicates))
         self._negativeObserver = negativeObserver
 
-    def __call__(self, event):
+    def __call__(self, event: LogEvent) -> None:
         """
         Forward to next observer if predicate allows it.
         """
         if self._shouldLogEvent(event):
             if "log_trace" in event:
-                event["log_trace"].append((self, self.observer))
+                event["log_trace"].append((self, self._observer))
             self._observer(event)
         else:
             self._negativeObserver(event)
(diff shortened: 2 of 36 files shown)


wsanchez:
Breakage in CI appears unrelated.

mypy.ini
```diff
@@ -93,10 +93,6 @@ allow_incomplete_defs = True
 [mypy-twisted.internet.testing]
 allow_incomplete_defs = True
 
-[mypy-twisted.logger.*]
-allow_untyped_defs = True
-check_untyped_defs = False
-
```
wsanchez, line 99: When you say it like this, it seems easy.  ;-)

wsanchez:
One thing I notice that didn't like as I went through this is typing constants from `constantly` as, for example `NamedConstant` instead of `LogLevel`, the latter of which is how one would type an `Enum` value.

I wonder if it's worth migrating (in another PR, I think) to `Enum` for something like `LogLevel`, since the usage of those values *shouldn't* cause a compatibility problem…

src/twisted/logger/_filter.py
```diff
@@ -154,15 +151,13 @@ def logLevelForNamespace(self, namespace):
         C{logLevelForNamespace("mypackage.subpackage")} will return
         C{LogLevel.debug}.
 
-        @param namespace: A logging namespace, or L{None} for the default
+        @param namespace: A logging namespace.  Use C{""} for the default
```
wsanchez, line 154: Stop using `None` as the default namespace, as it's distinction from `""` is subtle and unintentional and I found a bug here that resulted from that distinction.

src/twisted/logger/_filter.py
```diff
@@ -154,15 +151,13 @@ def logLevelForNamespace(self, namespace):
...
             namespace.
-        @type namespace: L{str} (native string)
 
         @return: The log level for the specified namespace.
-        @rtype: L{LogLevel}
         """
         if not namespace:
-            return self._logLevelsByNamespace[None]
+            return self._logLevelsByNamespace[""]
```
wsanchez, line 160: Notice here that we treat `None` and `""` as the same thing, and always return the same/shared log level for either namespace.

src/twisted/logger/_filter.py
```diff
@@ -176,44 +171,41 @@ def logLevelForNamespace(self, namespace):
...
+    def __call__(self, event: LogEvent) -> NamedConstant:
         eventLevel = event.get("log_level", None)
-        namespace = event.get("log_namespace", None)
-        namespaceLevel = self.logLevelForNamespace(namespace)
+        if eventLevel is None:
+            return PredicateResult.no
 
-        if (
-            eventLevel is None
-            or namespace is None
```
wsanchez, line 213: Unlike `logLevelForNamespace` (which we just called to get a log level to compare against), where we treated `None` and `""` the same, we distinguish between them here (`namespace is None` vs. `not namespace`).  So our logic here is inconsistent.

The solution in this PR is to make the default `""` instead of `None`, but to continue to handle `None` (which we get via `not namespace`) for compatibility to existing usage of `None`.

src/twisted/logger/_filter.py
```diff
@@ -176,44 +171,41 @@ def logLevelForNamespace(self, namespace):
...
-            or LogLevel._priorityForLevel(eventLevel)
-            < LogLevel._priorityForLevel(namespaceLevel)
```
wsanchez, line 215: We get rid of `_priorityForLevel` in this PR. It's no longer needed because [Trac ticket #6523](https://twistedmatrix.com/trac/ticket/6523) is fixed, so comparing the levels directly works.

src/twisted/logger/_json.py
```diff
@@ -40,42 +43,16 @@ def failureAsJSON(failure):
     )
 
 
-def asBytes(obj):
```
wsanchez, line 43: This was not in use.

src/twisted/logger/_levels.py
```diff
@@ -82,25 +79,3 @@ def levelWithName(cls, name):
...
-
-        @return: A numeric index indicating priority (lower is higher level).
-        @rtype: L{int}
-        """
-        return cls._levelPriorities[level]
-
-
-LogLevel._levelPriorities = dict(
-    (level, index) for (index, level) in (enumerate(LogLevel.iterconstants()))
-)
```
wsanchez, line 106: https://twistedmatrix.com/trac/ticket/6523 was fixed

src/twisted/logger/_logger.py
```diff
@@ -67,11 +64,11 @@ def __init__(self, namespace=None, source=None, observer=None):
         if observer is None:
             from ._global import globalLogPublisher
 
-            self.observer = globalLogPublisher
+            self.observer = globalLogPublisher  # type: ILogObserver
         else:
             self.observer = observer
 
-    def __get__(self, oself, type=None):
+    def __get__(self, instance: object, owner: Optional[type] = None) -> object:
```
wsanchez, line 71: Changed argument names to match `object.__get__`

src/twisted/logger/_observer.py
```diff
@@ -17,53 +20,6 @@
 )
 
 
-class ILogObserver(Interface):
```
wsanchez, line 20: Moved to `_interfaces.py` to prevent circular import problems.

src/twisted/logger/_observer.py
```diff
@@ -99,24 +55,23 @@ def removeObserver(self, observer):
...
             pass
 
-    def __call__(self, event):
+    def __call__(self, event: LogEvent) -> None:
         """
         Forward events to contained observers.
         """
-        if "log_trace" in event:
+        if "log_trace" not in event:
+            trace = None  # type: Optional[Callable[[ILogObserver], None]]
```
wsanchez, line 63: Swapped the conditional branches because it's easier to apply the type hint that way.

src/twisted/logger/_observer.py
```diff
@@ -136,17 +91,23 @@ def trace(observer):
...
+        @return: A L{Logger} without the given observer.
         """
         errorPublisher = LogPublisher(
             *[obs for obs in self._observers if obs is not observer]
         )
         return Logger(observer=errorPublisher)
+
+
+@implementer(ILogObserver)
+def bitbucketLogObserver(event: LogEvent) -> None:
```
wsanchez, line 110: Used where `lambda e: None` was used, but (a) it's a single object and (b) it's property typed.

src/twisted/logger/test/test_flatten.py
```diff
@@ -158,13 +162,28 @@ def keyFromFormat(format):
...
+            ),
+        ) = aFormatter.parse("{x}")
+        assert fieldName is not None
+
+        self.assertEqual(
+            sameFlattener.flatKey(fieldName, formatSpec, conversion), "x!:"
+        )
+        self.assertEqual(
+            sameFlattener.flatKey(fieldName, formatSpec, conversion), "x!:/2"
+        )
```
wsanchez, line 182: Don't build a `*key` thing; spell it all out

src/twisted/logger/test/test_stdlib.py
```diff
@@ -37,12 +38,48 @@ def nextLine():
     )
 
 
+class StdlibLoggingContainer:
```
wsanchez, line 41: Moved up form bottom of file to avoid the need for forward references to it.

src/twisted/logger/_filter.py
```diff
@@ -136,16 +134,15 @@ class LogLevelFilterPredicate:
     Events that not not have a log level or namespace are also dropped.
     """
 
-    def __init__(self, defaultLogLevel=LogLevel.info):
+    def __init__(self, defaultLogLevel: NamedConstant = LogLevel.info) -> None:
         """
         @param defaultLogLevel: The default minimum log level.
-        @type defaultLogLevel: L{LogLevel}
```
rodrigc, line 142: In a separate commit, consider updating the coding standard at:

https://github.com/twisted/twisted/blob/trunk/docs/core/development/policy/coding-standard.rst

to remove the use of `@type`, and encouraging the use of type annotations.

rodrigc (approved):
This is a fairly large patch, but the changes are straightforward and clean.

PR #1427, opened 2020-10-07
[5356] Reduce the failure rate for low-level IOCP port listen test.

adiroiban:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/5356


src/twisted/internet/test/test_iocp.py
```diff
@@ -71,9 +73,36 @@ def _acceptAddressTest(self, family, localhost):
...
+                break
+            except error as socketError:
+                if socketError.errno == WSAENOTCONN:
+                    # Without a sleep here even retrying 20 times will fail.
+                    # This should allow other threads to execute.
+                    time.sleep(0.1)
+                    pass
+                else:
+                    # Not the excepted error so we re-raise the error without
+                    # retying.
```
mthuurne: Typos: "excepted" -> "expected" and "retying" -> "retrying".

src/twisted/internet/test/test_iocp.py
```diff
@@ -71,9 +73,36 @@ def _acceptAddressTest(self, family, localhost):
...
+                server.setsockopt(
+                    SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
+                )
+                break
+            except error as socketError:
+                if socketError.errno == WSAENOTCONN:
+                    # Without a sleep here even retrying 20 times will fail.
+                    # This should allow other threads to execute.
+                    time.sleep(0.1)
+                    pass
```
mthuurne: The `pass` is redundant.

src/twisted/internet/test/test_iocp.py
```diff
@@ -71,9 +73,36 @@ def _acceptAddressTest(self, family, localhost):
...
+                else:
+                    # Not the excepted error so we re-raise the error without
+                    # retying.
+                    raise
+            # Second try also failed so we just re-raise the error.
+            raise
```
mthuurne: I don't understand; what is that "second try"?

adiroiban:
This is ready to review...

So... to minimize the test run time... it will try to call setsockopt right away and then wait only on errors.

src/twisted/internet/iocpreactor/tcp.py
```diff
@@ -560,6 +560,11 @@ def handleAccept(self, rc, evt):
             )
             return False
         else:
+            # Inherit the properties from the listening port socket as
+            # documented in the `Remarks` section of AcceptEx.
+            # https://msdn.microsoft.com/en-us/library/ms737524(v=VS.85).aspx
```
mthuurne: While that URL still works, it now redirects to:
https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nf-mswsock-acceptex

src/twisted/internet/test/test_iocp.py
```diff
@@ -71,9 +72,37 @@ def _acceptAddressTest(self, family, localhost):
...
+            # Calling setsockopt after _iocp.accept might fail for both IPv4
+            # and IPV6 with [Errno 10057] A request to send or receive ...
+            # This is when ERROR_IO_PENDING is returned and means that the
+            # socket is not yet ready and accept will be handled via the
+            # callback event.
+            # For this test, with the purpose of keeping the test simple,
+            # we don't implement the event callback.
+            # The event callback functionality is tested via the high level
+            # tests for general reactor API.
+            # We retry twice as retrying once is not enough.
```
graingert:
```suggestion
            # We make three attempts because two are not enough.
```

Is this from experiment? or just a guess?

src/twisted/internet/test/test_iocp.py
```diff
@@ -71,9 +72,37 @@ def _acceptAddressTest(self, family, localhost):
...
+            try:
+                server.setsockopt(
+                    SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
+                )
+                break
+            except error as socketError:
```
graingert: socket.error is a deprecated alias for `OSError`

adiroiban:
Many thanks @graingert and @mthuurne for your review.

I am not convince that this will fix the error :( ...as I saw another error on ipv4. Might need to retry more than 2 times.

But I will try to trigger more tests and see how it goes

adiroiban:
@graingert and @mthuurne please give it another try.

I hope that this will reduce the failure rate for this test.
I expect the test to still fail from time to time, and when it happened the number of retries can be increased.

Thanks!

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +69,47 @@ def _acceptAddressTest(self, family, localhost):
...
+            # tests for general reactor API.
+            # We retry multiple times to cover.
+            try:
+                server.setsockopt(
+                    SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
+                )
+                # Reset any previous error and stop retrying.
+                socketError = None
+                break
+            except OSError as socketError:
```
graingert, line 111: won't this socketError get deleted?

```python
def demo():
    e = None
    try:
        raise Exception
    except Exception as e:
        pass

    return e


demo()
```

```pytb
  File "demo.py", line 11, in <module>
    demo()
  File demo.py", line 8, in demo
    return e
UnboundLocalError: local variable 'e' referenced before assignment
```

adiroiban: OMG. Python 3 is crazy :) True

Sorry, I am still working only on Python2.

But why was this error not picked up by the linters.

it looks like a common error to be make when migrating from py2 to py3.

thanks @graingert . Great feedback :)

graingert:
```python
for attempts_remaining in range(4, -1, -1):
    try:
        server.setsockopt(SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno()))
    except OSError as e:
        if attempts_remaining == 0 or e.errno != getattr(errno, "WSAENOTCONN"):
            raise
        time.sleep(0.2)
```

adiroiban: will fix this.

More info https://stackoverflow.com/a/44258937/539264 and https://www.python.org/dev/peps/pep-3110/

src/twisted/internet/test/test_iocp.py
```diff
@@ -94,21 +94,22 @@ def _acceptAddressTest(self, family, localhost):
...
+                if lastError.errno == getattr(errno, "WSAENOTCONN"):
                     # Without a sleep here even retrying 20 times will fail.
                     # This should allow other threads to execute.
                     time.sleep(0.2)
                 else:
                     # This is not the expected error so re-raise the error without retrying.
                     raise
 
-        if socketError:
+        if lastError:
```
graingert: I think it's neater to count backwards, than assign and re-assign the error:

```python
for attempts_remaining in reversed(range(5)):
    try:
        server.setsockopt(SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno()))
        break
    except OSError as e:
        if attempts_remaining == 0 or e.errno != getattr(errno, "WSAENOTCONN"):
            raise
        time.sleep(0.2)
```

adiroiban: yep. Will do it. thanks :)

adiroiban:
I think that this is ready for a new round of reviews.

5 years ago when I tested the initial code I think that I have temporary update the SupportTests to call the ipv4 and ipv6 accept tests 100 times in a row in a single test run.

I think that I can also do it now to see that if the tests is executed 100 or 1000 times in the Azure or GitHub VMs they will not fail.

src/twisted/internet/test/test_iocp.py
```diff
@@ -53,6 +54,10 @@ def _acceptAddressTest(self, family, localhost):
         address family of C{family} and assert that the result of
         L{iocpsupport.get_accept_addrs} is consistent with the result of
         C{socket.getsockname} and C{socket.getpeername}.
+
+        This is a flaky tests.
```
mthuurne:
```suggestion
        These are a flaky tests.
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +69,47 @@ def _acceptAddressTest(self, family, localhost):
...
         self.addCleanup(server.close)
         buff = array("B", b"\0" * 256)
         self.assertEqual(0, _iocp.accept(port.fileno(), server.fileno(), buff, None))
-        server.setsockopt(
-            SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
-        )
+
+        for attemptsRemaining in reversed(range(5)):
+            # Calling setsockopt after _iocp.accept might fail for both IPv4
+            # and IPV6 with [Errno 10057] A request to send or receive ...
```
mthuurne:
```suggestion
            # and IPV6 with "[Errno 10057] A request to send or receive ..."
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +69,47 @@ def _acceptAddressTest(self, family, localhost):
...
+                # The socket is not yet ready to accept connections,
+                # setsockopt fails.
+                if attemptsRemaining < 0:
+                    # We rant out of retries.
+                    raise
+
+                # getattr is used below to make mypy happy.
+                if socketError.errno != getattr(errno, "WSAENOTCONN"):
+                    # This is not the expected error so re-raise the error without retrying.
+                    raise
```
mthuurne, line 115: I think it would be clearer to check `errno` before `attemptsRemaining`. The `errno` check is like a subtype check: we don't want to catch all `OSError` instances, but only one particular kind. So it would make sense to have it immediately following the `except OSError` line.

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +69,47 @@ def _acceptAddressTest(self, family, localhost):
...
+            # We retry multiple times to cover.
+            try:
+                server.setsockopt(
+                    SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
+                )
+                break
+            except OSError as socketError:
+                # The socket is not yet ready to accept connections,
+                # setsockopt fails.
+                if attemptsRemaining < 0:
```
mthuurne: `reversed(range(5))` produces the sequence (4, 3, 2, 1, 0), so there is no way `attemptsRemaining` is ever below zero.

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +69,47 @@ def _acceptAddressTest(self, family, localhost):
...
 
         server = socket(family, SOCK_STREAM)
         self.addCleanup(server.close)
         buff = array("B", b"\0" * 256)
         self.assertEqual(0, _iocp.accept(port.fileno(), server.fileno(), buff, None))
-        server.setsockopt(
-            SOL_SOCKET, SO_UPDATE_ACCEPT_CONTEXT, pack("P", port.fileno())
-        )
+
+        for attemptsRemaining in reversed(range(5)):
```
mthuurne, line 95: Since the attempts + delay regulate the number of unintended failures, why not set the values a bit higher? For example a 10 second or even 30 second timeout is fine, since that worst case is only hit when the test fails and even then only when it fails in a very particular way (`accept()` succeeding but the socket never reaching connected state).

mthuurne (changes requested):
I added some more comments. Most are minor, but the `attemptsRemaining < 0` one looks like a bug.

src/twisted/internet/test/test_iocp.py
adiroiban: I think that this test should not take more than 5 seconds to executes.

If it takes 10 seconds for an IOCP socket to accept an already triggered connection, then I think that something is very wrong.

Most of the time this is an instant operation.... so anything about 5 second should raise a warning.

I wanted to keep lower values to see how often it fails with a timeout below 1 second.

src/twisted/internet/test/test_iocp.py
adiroiban: True. Stupid me :) Thanks

src/twisted/internet/test/test_iocp.py
mthuurne: A test case taking unusually long before passing can indeed be an indication of a deeper problem, but if we are interested in measuring that, it would be better to make that part of the test framework rather than individual tests, in my opinion.

I have no data to support my gut feeling that a 10 seconds timeout is better than 1 second, so let's leave it as it is and see what happens.

mthuurne (approved):
Looks good now. Let's hope our theories are correct and this indeed fixes the flaky tests.

adiroiban:
Thanks @mthuurne  for the review.

I pushed another commit just to test that all is ok.

There are many comments, but I tried to summarize all the information for the PR conversation.

Let me know if it make sense.

Thanks!

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
         L{iocpsupport.get_accept_addrs} is consistent with the result of
         C{socket.getsockname} and C{socket.getpeername}.
 
+        A port starts listening (is binded) at the low-level socket without
```
mthuurne:
```suggestion
        A port starts listening (is bound) at the low-level socket without
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
...
+        calling accept yet.
```
mthuurne:
```suggestion
        calling accept() yet.
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
...
+        A client is then connected.
+        After the client is connected IOCP accept is called, which is the
```
mthuurne:
```suggestion
        After the client is connected IOCP accept() is called, which is the
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
...
+        targeted of these tests.
```
mthuurne:
```suggestion
        target of these tests.
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
...
+
+        Most of the time, the socket is ready instantly, but sometimes
+        the socket is not ready right away after calling IOCP accept.
```
mthuurne:
```suggestion
        the socket is not ready right away after calling IOCP accept().
```

src/twisted/internet/test/test_iocp.py
```diff
@@ -55,9 +55,24 @@ def _acceptAddressTest(self, family, localhost):
...
+        It should not make more than 5 seconds for a socket to be ready, as
```
mthuurne:
```suggestion
        It should not take more than 5 seconds for a socket to be ready, as
```

adiroiban:
We have another failure on IOCP...but is not related to the scope of this PR.

https://dev.azure.com/twistedmatrix/twisted/_build/results?buildId=2813&view=logs&j=103d3058-bd81-5adc-9d09-335c8b1168a8&t=c2351c64-3c18-5e69-f4d7-151e9583a5ee&l=15866

```
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
Selectables:
<ThrottlingProtocol #1 on 52211>

twisted.test.test_policies.ThrottlingTests.test_limit
```

mthuurne:
Is that change in `policies.ThrottlingFactory` related to the IOCP accept test? Because at a first glance it seems unrelated, in which case it would be better to address it in a separate PR.

adiroiban:
Hi.  Thanks for the review. You are right.

I will revert it.

 I just wanted to test the code and trigger another set of tests for IOCP. 

With tests on Azure I can't retry a failed test so I had to commit something.

I will revert the change and if it will pass I will merge it and open another ticket for the other test.

src/twisted/internet/test/test_iocp.py
```diff
@@ -64,16 +84,47 @@ def _acceptAddressTest(self, family, localhost):
...
+            except OSError as socketError:
+                # getattr is used below to make mypy happy.
+                if socketError.errno != getattr(errno, "WSAENOTCONN"):
+                    # This is not the expected error so re-raise the error without retrying.
+                    raise
+
+                # The socket is not yet ready to accept connections,
+                # setsockopt fails.
+                if attemptsRemaining == 0:
+                    # We rant out of retries.
```
graingert:
```suggestion
                    # We ran out of retries.
```

```suggestion
                    # We ran out of attempts.
```

graingert (approved):

mthuurne:
> With tests on Azure I can't retry a failed test so I had to commit something.

That's strange, because I do get the option to re-run failed tests with Azure. You have to click through to Azure's web front-end though; you can't trigger it from GitHub. And while the re-run link is shown as soon as a test fails, it only seems to actually start re-running things if clicked when all tasks within a run have completed.

_Edit:_ Actually, I misremembered: the re-run link is in the sidebar on GitHub and **not** on the Azure site.

adiroiban:
@mthuurne thanks for the Azure re-run link :) I was searching for it on the Azure Devops page:)

But I hope that this IOCP test is now fixed for good :)

I will wait for all the tests to be green and will merge ...then we can see if iocp test continues to fail

graingert:
@adiroiban it passed!

adiroiban:
@mthuurne please see latest commit and let me know if that drive-by comment is ok.

For the non-related flaky test discovered during this PR, I have created a separate ticket but add the link to the ticket as a drive by in this PR.

Do you still approve this PR?

Thanks!

mthuurne:
Yes, I approve of its current state.

wsanchez:
Coolness.  Thanks for working on this, Adi.

adiroiban:
After the merge... many tests failed. @mthuurne  I see the "re-run" link/button, but it looks like it is not doing anything.

mthuurne:
> After the merge... many tests failed. @mthuurne I see the "re-run" link/button, but it looks like it is not doing anything.

When I click the "re-run" link, I see "You have successfully requested twisted.twisted be rerun." appear at the top of the screen, but nothing starts re-running or even leaves the error state. I don't have any ideas why or how to debug this, unfortunately.

adiroiban:
@wsanchez this is where you can see all the checks for trunk 

image

PR #1428, opened 2020-10-08
upgrade to declarative setuptools

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9999


src/twisted/python/_setup.py
```diff
@@ -223,71 +173,25 @@ def getSetupArgs(
...
+        # actually used by our custom builder.  distutils deep-down checks
+        # to see if there are any ext_modules defined before invoking
+        # the build_ext command.  We need to trigger build_ext regardless
+        # because it is the thing that does the conditional checks to see
+        # if it should build any extensions.  The reason we have to delay
+        # the conditional checks until then is that the compiler objects
+        # are not yet set up when this code is executed.
+        "ext_modules": extensions,
+        "cmdclass": {"build_ext": my_build_ext},
+        "extras_require": _EXTRAS_REQUIRE,
```
graingert, line 182: currently _EXTRAS_REQUIRE is in python to save on duplication eg with `twisted[all_non_platform]`

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    Source=https://github.com/twisted/twisted
+    Issues=https://twistedmatrix.com/trac/report
+license = MIT
+classifiers =
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
mthuurne, line 21: 3.9 was released this week and CI passes on 3.9-rc2, so I think it's safe to add 3.9 to the classifiers as well.

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+long_description_content_type = text/x-rst
+
+[options]
+python_requires = >=3.5
```
mthuurne: Since `typing.TYPE_CHECKING` was added in 3.5.2, I think the minimum version needs a tiny bump.

src/twisted/python/_setup.py
```diff
@@ -182,32 +140,24 @@ def __init__(self, *args: object, **kwargs: object) -> None:
 ]
 
 
-def _longDescriptionArgsFromReadme(readme: str) -> Dict[str, str]:
+def _longDescriptionFromReadme(readme: str) -> str:
     """
     Generate a PyPI long description from the readme.
 
     @param readme: Path to the readme reStructuredText file.
```
mthuurne: Since `readme` contains a path, why not make `Path` its type?

src/twisted/python/test/test_setup.py
```diff
@@ -366,34 +305,33 @@ class LongDescriptionTests(SynchronousTestCase):
...
+            textwrap.dedent(
+                """\
+                Twisted
+                =======
+
+                Changes: `NEWS <https://github.com/twisted/twisted/blob/trunk/NEWS.rst>`_.
+                Read `the docs <https://twistedmatrix.com/documents/>`_.
+                """
+            ),
+            getSetupArgs(extensions=[], readme=str(path))["long_description"],
```
mthuurne: Since `getSetupArgs()` just passes on the result of `_longDescriptionFromReadme()` unchanged, you could instead call the latter directly. That avoids this test case creating an inflated amount of coverage.

src/twisted/python/test/test_setup.py
```diff
@@ -40,7 +38,9 @@ def test_conditionalExtensions(self):
             "whatever", ["whatever.c"], condition=lambda b: False
         )
 
-        args = getSetupArgs(extensions=[good_ext, bad_ext], readme="")
+        path = pathlib.Path(self.mktemp())
+        path.write_text("", encoding="utf8")
```
mthuurne:
```suggestion
        path.touch(exist_ok=False)
```

mthuurne (commented):
I added some comments suggesting potential cleanups. Overall it looks good.

I only reviewed the code by reading; I didn't test whether this produces a working wheel with all the correct metadata.

One concern I have though is that `test_mail.py` and `test_options.py` are removed, but the code that they're testing is not. This is probably related to the removal of the `notPortedModules` list, but it seems like something that should be resolved in a more thorough matter, perhaps in a separate PR.

src/twisted/python/test/test_setup.py
graingert: this is intentional to avoid coupling getSetupArgs to its specific implementation

src/twisted/python/test/test_setup.py
```diff
@@ -60,7 +60,9 @@ def test_win32Definition(self):
             "whatever", ["whatever.c"], define_macros=[("whatever", 2)]
         )
 
-        args = getSetupArgs(extensions=[ext], readme="")
+        path = pathlib.Path(self.mktemp())
+        path.write_text("", encoding="utf8")
```
graingert:
```suggestion
        path.touch(exist_ok=False)
```

setup.cfg
graingert: doesn't this stuff get backported without changing the version in some platforms?

I think it will be changed to 3.6 soon though

src/twisted/python/test/test_setup.py
graingert: now that _longDescriptionFromReadme is one expression - I refactored it into getSetupArgs

setup.cfg
graingert:
```suggestion
    Programming Language :: Python :: 3.8
    Programming Language :: Python :: 3.9
```

graingert:
> One concern I have though is that test_mail.py and test_options.py are removed, but the code that they're testing is not. This is probably related to the removal of the notPortedModules list, but it seems like something that should be resolved in a more thorough matter, perhaps in a separate PR.

I think reverting the tests and fixing them is on the table. Or deleting these `notPortedModules`. I don't think they do any harm in the dist though

setup.cfg
mthuurne: - With `>=3.5` on a system that runs actual Python 3.5.0, `pip` will install the new Twisted release and then Twisted will break when importing `typing.TYPE_CHECKING`.
- With `>=3.5.2` on a system that runs Python that identifies as 3.5.0 but is actually patched with bits from later releases, `pip` won't install the new Twisted release.

Since the latter is the least disruptive for the user, I think it's preferable.

There are plans to drop 3.5, see #1386, but that isn't scheduled to be merged until after the next release.

graingert: I've pushed that change now

wsanchez:
> > One concern I have though is that test_mail.py and test_options.py are removed, but the code that they're testing is not. This is probably related to the removal of the notPortedModules list, but it seems like something that should be resolved in a more thorough matter, perhaps in a separate PR.
> 
> I think reverting the tests and fixing them is on the table. Or deleting these `notPortedModules`. I don't think they do any harm in the dist though

I agree with @mthuurne that removing them is problematic…

wsanchez:
WHOAH oops didn't mean to close

wsanchez:
I think this is a lot more readable and I like moving the static config back up to the top of the project.  I've never really used `setup.cfg` so I'm not aware of what pitfalls may exist, if any, or what the best way to verify the built result is OK, though I guess that's what pre-releases are for…?

graingert:
> or what the best way to verify the built result is OK, though I guess that's what pre-releases are for…?

a pypi pre-release or a test.pypi.org upload would work. Tox does run this so it mostly works and the wheels seem fine.

you can run `python setup.py sdist bdist_wheel` and inspect the results

wsanchez:
> a pypi pre-release or a test.pypi.org upload would work. Tox does run this so it mostly works and the wheels seem fine.

I like that.

@rodrigc @glyph How do you feel about giving one of us upload permissions to Twisted on `test.pypi.org`?

graingert:
@wsanchez you could rename the `setup.cfg[metadata].name` key

rodrigc:
> @rodrigc @glyph How do you feel about giving one of us upload permissions to Twisted on `test.pypi.org`?

I can't help you with that.  I also am blocked on that: https://github.com/twisted/twisted/pull/1423#issuecomment-703744080

graingert:
@rodrigc you appear to be soul maintainer of https://test.pypi.org/project/Twisted/ you could add @wsanchez as a maintainer

rodrigc:
Interesting.  I don't know when I was granted that access to test.pypi.org.  For now I am going to focus on moving forward with the release, before giving additional people access to PyPI resources.  If Glyph or Hawkowl want to grant access, I will defer to their wishes.

glyph:
@rodrigc You weren't "granted access" - you own the project in its entirety on test.pypi.org; the data there does not derive from PyPI at all.

https://test.pypi.org/project/Twisted/

I didn't even have an account on test.pypi.org — I just made one, so at least I got "glyph" :).  Please go ahead and give everybody who needs it access to test PyPI (including me :)).

rodrigc:
@glyph OK, I granted you and @wsanchez access to https://test.pypi.org/ .  I couldn't find an account for hawkowl on test.pypi.org to grant access.

graingert:
> I agree with @mthuurne that removing them is problematic…

I've restored the tests that were "commented out" in src/twisted/python/_setup.py, ported just enough code to get them to run the test setUps and skipped every test that fails

wsanchez (approved):
cool, thanks

PR #1430, opened 2020-10-08
Add type hints to twisted.internet.interfaces and twisted.internet.base

wsanchez:
* Complete partial type hints for `twisted.internet.*`
 * Add type hints to `twisted.internet.interfaces` and `twisted.internet.base`

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10002


mypy.ini
```diff
@@ -54,45 +54,435 @@ check_untyped_defs = False
 allow_untyped_defs = True
 check_untyped_defs = False
 
-[mypy-twisted.internet.*]
```
wsanchez, line 70: This replaces config for `twisted.internet.*` and more specific config for a few files therein with individual entries for each file, so that we can pick them off one at a time.

src/twisted/internet/abstract.py
```diff
@@ -361,7 +361,7 @@ def write(self, data: bytes):
             self._maybePauseProducer()
             self.startWriting()
 
-    def writeSequence(self, iovec: Sequence[bytes]):
```
wsanchez, line 364: Any `Iterable`, not just `Sequence`s.

src/twisted/internet/abstract.py
```diff
@@ -361,7 +361,7 @@ def write(self, data: bytes):
...
+    def writeSequence(self, iovec: Iterable[bytes]) -> None:
```
wsanchez, line 364: Any `Iterable`, not just `Sequence`s.

src/twisted/internet/abstract.py
```diff
@@ -483,26 +483,22 @@ def fileno(self):
...
         addresses.)
-    @type family: C{int}
 
-    @rtype: C{bool}
     @return: C{True} if C{addr} represents an IPv4 address, C{False} otherwise.
     """
-    if isinstance(addr, bytes):
-        try:
+    if isinstance(addr, bytes):  # type: ignore[unreachable]
+        try:  # type: ignore[unreachable]
```
wsanchez, line 501: `addr` is typed as `str`, but code if left to continue to handle `bytes`.

src/twisted/internet/base.py
```diff
@@ -316,6 +360,16 @@ def getHostByName(self, name, timeout=(1, 3, 11, 45)):
...
+_ThreePhaseEventTriggerCallable = Callable[..., Any]
+_ThreePhaseEventTrigger = Tuple[
+    _ThreePhaseEventTriggerCallable, Tuple[object, ...], Dict[str, object]
+]
+_ThreePhaseEventTriggerHandle = NewType(
+    "_ThreePhaseEventTriggerHandle",
+    Tuple[str, _ThreePhaseEventTriggerCallable, Tuple[object, ...], Dict[str, object]],
+)
+
+
```
wsanchez, line 375: Internal types.
I used `NewType` for the handle that is returned to clients to make it more strictly opaque.

src/twisted/internet/base.py
```diff
@@ -494,43 +554,39 @@ def installNameResolver(self, resolver):
...
         Implementation of read-only
         L{IReactorPluggableNameResolver.nameResolver}.
         """
         return self._nameResolver
 
 
+_SystemEventID = NewType("_SystemEventID", Tuple[str, _ThreePhaseEventTriggerHandle])
+_ThreadCall = Tuple[Callable[..., Any], Tuple[object, ...], Dict[str, object]]
+
+
```
wsanchez, line 571: Again using `NewType` for handle returned to clients

src/twisted/internet/base.py
```diff
@@ -662,80 +720,104 @@ def crash(self):
         self.running = False
         self.addSystemEventTrigger("during", "startup", self._reallyStartRunning)
 
-    def sigInt(self, *args):
+    def sigInt(self, number: int, frame: Optional[FrameType] = None) -> None:
```
wsanchez, line 726: This matches the signature described by `signal.signal`.

src/twisted/internet/base.py
```diff
@@ -757,26 +839,30 @@ def startRunning(self):
...
+    def callLater(
+        self, delay: float, callable: Callable[..., Any], *args: object, **kw: object
+    ) -> DelayedCall:
+        """
+        See twisted.internet.interfaces.IReactorTime.callLater.
+        """
         assert builtins.callable(callable), "{} is not callable".format(callable)
         assert delay >= 0, "{} is not greater than or equal to 0 seconds".format(delay)
-        tple = DelayedCall(
+        delayedCall = DelayedCall(
```
wsanchez, line 868: What TF was `tple` supposed to mean? I assume that once upon a time, `DelayedCall` was a tuple.

src/twisted/internet/base.py
```diff
@@ -1074,65 +1164,79 @@ def _stopThreadPool(self):
...
 
 
 if platform.supportsThreads():
     classImplements(ReactorBase, IReactorThreads)
 
 
 @implementer(IConnector)
-class BaseConnector:
-    """Basic implementation of connector.
+class BaseConnector(ABC):
```
wsanchez, line 1212: Making this an ABC so that I can define the type signature of `_makeTransport`.

There are other classes around here that could probably benefit from becoming ABCs.

src/twisted/internet/base.py
```diff
@@ -1253,34 +1357,35 @@ def _handleSignals(self):
             )
             return
 
+        reactorBaseSelf = cast(ReactorBase, self)
```
wsanchez, line 1363: This because of mix-in supertype

src/twisted/internet/interfaces.py
```diff
@@ -890,21 +814,17 @@ class IReactorWin32Events(Interface):
     @since: 10.2
     """
 
-    def addEvent(event, fd, action):
+    def addEvent(event: object, fd: FileDescriptor, action: str) -> None:
```
wsanchez: Using `object` to type `event` for lack of anything else…

src/twisted/internet/interfaces.py
```diff
@@ -1689,7 +1591,7 @@ class IFileDescriptor(ILoggingContext):
     An interface representing a UNIX-style numeric file descriptor.
     """
 
-    def fileno():
+    def fileno() -> object:
```
wsanchez, line 1589: Using `object` since the docstring says the type is platform-specific.

src/twisted/mail/relaymanager.py
```diff
@@ -71,7 +71,7 @@ def connectionLost(self, reason: Failure = connectionDone):
         self.manager.notifyDone(self.factory)
 
 
-class SMTPManagedRelayer(ManagedRelayerMixin, relay.SMTPRelayer):
+class SMTPManagedRelayer(ManagedRelayerMixin, relay.SMTPRelayer):  # type: ignore[misc]
```
wsanchez, line 74: Mix-in defines `factory` as a property. Will deal with later.

src/twisted/test/test_tcp.py
```diff
@@ -120,9 +120,7 @@ class MyProtocolFactoryMixin:
 
     protocolConnectionMade = None
     protocolConnectionLost = None
-    protocol = (
-        None
-    )  # type: Optional[Union[Type[protocol.Protocol],Type[protocol.AbstractDatagramProtocol]]]  # noqa
```
wsanchez, line 125: `protocol.AbstractDatagramProtocol` should be a `Protocol`. Or something.
Removing it from here and working around it with `type: ignore` (eg. see `test_tuntap.py`) to make it more obvious that it needs fixing.

src/twisted/internet/base.py
```diff
@@ -75,24 +111,25 @@ def __init__(self, time, func, args, kw, cancel, reset, seconds=runtimeSeconds):
         self.canceller = cancel
         self.seconds = seconds
         self.cancelled = self.called = 0
-        self.delayed_time = 0
+        self.delayed_time = 0.0
```
wsanchez, line 117: Note that there are many places in `twisted.internet` where time is represented by "a number" and others where it it more clearly a `float` and others where it's documented as an `int`.

I'm trying to consistently use `float`.

wsanchez:
@glyph I suspect you'll want to at least scan this PR

src/twisted/internet/base.py
```diff
@@ -1368,8 +1368,9 @@ def _handleSignals(self) -> None:
         signal.signal(signal.SIGTERM, reactorBaseSelf.sigTerm)
 
         # Catch Ctrl-Break in windows
-        if hasattr(signal, "SIGBREAK"):
-            signal.signal(signal.SIGBREAK, reactorBaseSelf.sigBreak)
+        SIGBREAK = getattr(signal, "SIGBREAK", None)
```
rodrigc, line 1371: Yeah, I noticed that this workaround was necessary to appease mypy, since
`signal.SIGBREAK` is only on Windows.

src/twisted/internet/defer.py
```diff
@@ -91,7 +91,7 @@ def logError(err):
     return err
 
 
-def succeed(result):
+def succeed(result: object) -> "Deferred":
```
rodrigc, line 94: For these particular functions where you added type annotations, do you want to keep  the  @rtype's in the docstrings, or do you want to get rid of them?

src/twisted/internet/defer.py
```diff
@@ -113,7 +113,7 @@ def succeed(result):
     return d
 
 
-def fail(result=None):
+def fail(result: object = None) -> "Deferred":
     """
     Return a L{Deferred} that has already had C{.errback(result)} called.
```
rodrigc, line 117: There are a few other functions in this file, such as `execute()` and `maybeDeferred()`.  Do you want to take a whack at annotating those functions in this PR, or do you want to wait for a subsequent PR?

src/twisted/internet/iocpreactor/tcp.py
```diff
@@ -141,17 +140,13 @@ def logPrefix(self):
...
+        return bool(self.socket.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY))
 
     def setTcpNoDelay(self, enabled):
         self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, enabled)
 
     def getTcpKeepAlive(self):
-        return operator.truth(
-            self.socket.getsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE)
-        )
+        return bool(self.socket.getsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE))
```
rodrigc, line 149: Replacing `operator.truth()` with `bool()` is fine.

src/twisted/internet/protocol.py
```diff
@@ -30,9 +30,7 @@ class Factory:
     """
 
     # Put a subclass of Protocol here:
-    protocol = (
-        None
-    )  # type: Optional[Union[Type['Protocol'], Type['AbstractDatagramProtocol']]]  # noqa
+    protocol = None  # type: Optional[Type["Protocol"]]
```
rodrigc: I previously added `AbstractDatagramProtocol` to this annotation, because I ran into mypy problems elsewhere without it.  But if you can get rid of that here, that would be definitely simplify things.

src/twisted/test/test_tcp.py
rodrigc: Yeah I had problems with AbstractDatagramProtocol when initially adding mypy support to the tree.

wsanchez: Right.  When we add hints to `protocol.py` I think we need to figure out what to do with `AbstractDatagramProtocol`.  The current state of things seems incorrect to me, but I didn't want to spend too much time on that one just yet.

src/twisted/internet/base.py
wsanchez: Right, I'm trying not to add platform checks when checking for the attribute seems more robust.

Same sort of deal as we ran into with `select` module attributes.

src/twisted/internet/defer.py
wsanchez: I meant to get rid of them. Perhaps I forgot to do a final scan… yeah I totally skipped that for `interfaces.py`… and in doing so found a better type hint for one thing….

But this is `defer.py`… a bit harder to catch the ancillary type additions.  Fixing.

src/twisted/test/test_tcp.py
rodrigc: `AbstractDatagramProtocol` has been around since about 2003: https://github.com/twisted/twisted/commit/869198ef461818414b794ae86cfc1e065ccea6fe

I'm not sure how easy it will be to "fix" a base class in a library that has been around so long, yet preserve backwards compatibility.  It may not be possible.

glyph: Type signatures are not beholden to the compatibility policy, only runtime behavior.

wsanchez: @glyph I think the issue is that `AbstractDatagramProtocol` is weird in that it does not subclass `Protocol`, and it seems to be the only example of that. But again, I didn't look into it.

src/twisted/internet/defer.py
wsanchez: Different PR

src/twisted/internet/iocpreactor/tcp.py
wsanchez: Right. `operator.truth` is documented as being the same as `bool`.

src/twisted/internet/protocol.py
wsanchez: Yeah, [this ignore comment](https://github.com/twisted/twisted/pull/1430/files#diff-ad9e865d47a56bc625d757669be5663cL1331) is an example of fallout from undoing this.

I prefer that, though as ignore comments are a thing to look back at and fix, whereas I wouldn't find this annotation in a search for ways to improve the type hints; I'd have to run across it and wonder why it's so complex and then look into it and then decide I care…

glyph: This is why it's always a good idea to write up a bug and include a link to it next to any type:ignore's it affects :)

rodrigc: @glyph can you clarify what you would prefer the type of the protocol variable to be here?

1.   Leave it as `Optional[Union[Type['Protocol'], Type['AbstractDatagramProtocol']]] `
2.  change it to `type: Optional[Type["Protocol"]]`, and add `type:ignore[assignment]` comments in the tree as per: https://github.com/twisted/twisted/pull/1430/files#diff-ad9e865d47a56bc625d757669be5663cL1331

glyph: All of these types seem to be wrong, based on the intent of the attribute.  Hopefully `Optional[Callable[[], IProtocol]]` won't provoke any errors?  (It is depressing that `Factory` itself doesn't document this attribute when it's clearly so widely used…)

rodrigc: In my tree, I tried changing to:

```
diff --git a/src/twisted/internet/protocol.py b/src/twisted/internet/protocol.py
index 058644eef..1433105d6 100644
--- a/src/twisted/internet/protocol.py
+++ b/src/twisted/internet/protocol.py
@@ -11,12 +11,12 @@ Twisted.  The Protocol class contains some introductory material.
 
 
 import random
-from typing import Optional, Type, Union
+from typing import Optional, Type, Union, Callable
 
 from twisted.python import log, failure, components
 from twisted.internet import interfaces, error, defer
-from twisted.internet.interfaces import ITransport
+from twisted.internet.interfaces import ITransport, IProtocol
 from twisted.logger import _loggerFor
 
 
@@ -32,7 +32,7 @@ class Factory:
     # Put a subclass of Protocol here:
     protocol = (
         None
-    )  # type: Optional[Union[Type['Protocol'], Type['AbstractDatagramProtocol']]]  # noqa
+    )  # type: Optional[Callable[[], IProtocol]]
 
     numPorts = 0
     noisy = True
diff --git a/src/twisted/test/test_tcp.py b/src/twisted/test/test_tcp.py
index 4ae90494b..04bf70ffb 100644
--- a/src/twisted/test/test_tcp.py
+++ b/src/twisted/test/test_tcp.py
@@ -11,7 +11,7 @@ import random
 import errno
 import hamcrest
 from functools import wraps
-from typing import Optional, Type, Union
+from typing import Optional, Type, Union, Callable
 from unittest import skipIf
 
@@ -22,7 +22,7 @@ from twisted.python.log import msg, err
 from twisted.internet import protocol, reactor, defer, interfaces
 from twisted.internet.address import IPv4Address
-from twisted.internet.interfaces import IHalfCloseableProtocol, IPullProducer
+from twisted.internet.interfaces import IHalfCloseableProtocol, IPullProducer, IProtocol
 from twisted.protocols import policies
 from twisted.test.proto_helpers import AccumulatingProtocol
@@ -122,7 +122,7 @@ class MyProtocolFactoryMixin:
     protocolConnectionLost = None
     protocol = (
         None
-    )  # type: Optional[Union[Type[protocol.Protocol],Type[protocol.AbstractDatagramProtocol]]]  # noqa
+    )  # type: Optional[Callable[[], IProtocol]]
     called = 0
 
     def __init__(self):
```


and got these mypy errors:

```
src/twisted/words/protocols/irc.py:2826:16: error: Incompatible types in assignment (expression has type "Type[DccSendProtocol]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = DccSendProtocol
                   ^
src/twisted/words/protocols/irc.py:2938:16: error: Incompatible types in assignment (expression has type "Type[DccChat]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = DccChat
                   ^
src/twisted/internet/endpoints.py:199:16: error: Incompatible types in assignment (expression has type "Type[_WrappingProtocol]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = _WrappingProtocol
                   ^
src/twisted/words/protocols/jabber/xmlstream.py:688:16: error: Incompatible types in assignment (expression has type "Type[XmlStream]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = XmlStream
                   ^
src/twisted/names/server.py:64:16: error: Incompatible types in assignment (expression has type "Type[DNSProtocol]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = dns.DNSProtocol
                   ^
src/twisted/web/proxy.py:91:16: error: Incompatible types in assignment (expression has type "Type[ProxyClient]", base class "Factory" defined the type as "Optional[Callable[[], IProtocol]]")  [assignment]
        protocol = ProxyClient
                   ^
src/twisted/pair/test/test_tuntap.py:1331:24: error: Incompatible types in assignment (expression has type "Type[IPRecordingProtocol]", variable has type "Optional[Callable[[], IProtocol]]")  [assignment]
        factory.protocol = IPRecordingProtocol
                           ^
src/twisted/pair/test/test_tuntap.py:1353:24: error: Incompatible types in assignment (expression has type "Type[EthernetRecordingProtocol]", variable has type "Optional[Callable[[], IProtocol]]")  [assignment]
        factory.protocol = EthernetRecordingProtocol
                           ^
Found 8 errors in 6 files (checked 840 source files)
```

rodrigc: Also `AbstractDatagramProtocol` does not implement `connectionLost()` which is required by the `IProtocol` interface.  That's why I used the type `Optional[Union[Type['Protocol'], Type['AbstractDatagramProtocol']]]`, because there are a bunch of places that assume you can set `protocol` to an `AbstractDatagramProtocol`

glyph: OK let's go through these one at a time.

glyph: - `DccSendFactory` appears to be … totally unused. Undocumented everywhere. `DccSendProtocol`, similarly so. Not clear what is trying to be expressed here.  The type error is real; this only works because `DccSendProtocol` redefines the expected type of `protocol` and thereby breaks LSP. Also it looks like it might be a security nightmare if anybody actually tried to use this, so maybe just… deprecate & delete?

glyph: - `DccChatFactory` is similarly defining a completely new interface. It shouldn't be subclassing `ClientFactory`, it's not compatible.  It should be *implementing* an `IClientFactory` maybe.  Although it's similarly unused, at least it's tested.  Again fixing the types on the IRC side would make sense here.

glyph: - `_WrappingFactory` once again demonstrates the need for an `IClientFactory` which doesn't have this protocol attribute at all.  It's also changing the contract and should change the type of the declared attribute.

glyph: - `XmlStream` just has a bogus type declaration; delete it and it should work

glyph: - `DNSServerFactory`'s `buildProtocol` once again changes the contract; this is a real type error

glyph: - `ProxyClientFactory` once again changes the contract in its `buildProtocol`. Same as other instances.

glyph: - `test_tuntap.py`'s usages are goofy nonsense.  In real life these datagram protocols are constructed and passed directly to `TuntapPort`.  (See for example `docs/pair/examples/pairudp.py`).  The real implementation does not use `Factory` and here `Factory` with `.protocol` set to `lambda: y` having `buildProtocol` called is merely being used as a shortcut for `(lambda: y)()`.  Eliminate the `factory` from the tests entirely, just instantiate the `Protocol` or make a method that returns it for each test.

glyph: So yeah, `AbstractDatagramProtocol` is a complete red herring here, the type should remain `Optional[Callable[[], IProtocol]]`.

wsanchez: @glyph: I'm assuming that we're going to need to go through all of the `type: ignore` items after we've typed most of Twisted, so I feel like it's a bit premature to file tickets for all of them, because some of those get fixed by better typing elsewhere in the code, and others as actual bugs, etc.

My feeling on this one is to leave it as it is now, and fix the weirdness in `AbstractDatagramProtocol` in a different PR.

wsanchez: Sorry, let me back-pedal a bit.  I see why `Optional[Callable[[], IProtocol]]` seems more correct, and we should start with more correct.  Let me start there.

wsanchez: One problem with `IProtocol`:

```console
src/twisted/internet/protocol.py:132:9: error: "IProtocol" has no attribute "factory"  [attr-defined]
            p.factory = self
```

It seems like `Factory`  and `Protocol` are rather coupled, so `IProtocol` seems inappropriate here.

rodrigc: If you can, I would suggest reverting all protocol-related changes from this PR.
I am OK with all other changes in this PR, so if we could get those in that would be good.

Then in a separate PR, tackle all of @glyph's feedback related to protocol.

wsanchez: OK. Check out 0a10890 for fixes to the above, which includes comments referring to new tickets about wrongness noted by @glyph.

9b6fc8a deprecates the unused stuff in IRC.  I included `DccChatFactory` because it's not used in Twisted and I also couldn't find users on GitHub.

rodrigc: My suggestion is that https://github.com/twisted/twisted/commit/0a10890695d0236f9171883f8dfaa934e97f7221 and https://github.com/twisted/twisted/commit/9b6fc8a5a53a92d4e29335b1898ebe48d2e98546 be pulled out of this PR into separate PR's, since this is beginning to have too many changes beyond just adding type hints.

Deprecated requires a `.removal` newsfragment and tests as per the [Twisted compatibilty policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html)

wsanchez: 0a10890 is required it's just ignores plus comments referring to the tickets, as @glyph requested.

I can pull 9b6fc8a if you like, but it's a pretty small change. Lemme know and I'll either do that or add the `.removal` thing.

src/twisted/internet/defer.py
glyph: Worth noting here that this type annotation will need to be updated when we make Deferred a generic.  I'd almost rather we do that *before* adding types all over the place in this module, since this gives the illusion of type safety without actually ensuring anything about the result type in the Deferred.

src/twisted/test/test_tcp.py
glyph: After having looked at it, it seems like AbstractDatagramProtocol is only expected here by the wacky twisted.pair tests, right?

src/twisted/internet/protocol.py
rodrigc: https://github.com/twisted/twisted/commit/9b6fc8a5a53a92d4e29335b1898ebe48d2e98546 needs to be a separate PR with a proper newsfragment and tests as per the [Twisted compatibilty policy](https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html)

src/twisted/test/test_tcp.py
wsanchez: …correct

wsanchez: Well, no… it's used in `twisted.pair` non-test code. But it's only in `twisted.pair`.

src/twisted/internet/protocol.py
wsanchez: OK I'll pull it

wsanchez: Done. I'll file a PR for it after this is merged.

src/twisted/test/test_tcp.py
glyph: I guess I'll need to run this for myself locally since I didn't see it in the mypy errors craig posted. My non-tool-assisted reading of the app code was that things take datagram protocols directly; do you have a pointer to where this is happening?

wsanchez: I just did a search on the.sources:

```console
Searching 842 files for "AbstractDatagramProtocol" (case sensitive)

/Users/wsanchez/Dropbox/Developer/Twisted/twisted/src/twisted/internet/protocol.py:
  678  
  679  
  680: class AbstractDatagramProtocol:
  681      """
  682      Abstract protocol for datagram-oriented transports, e.g. IP, ICMP, ARP,
  ...
  754  
  755  @implementer(interfaces.ILoggingContext)
  756: class DatagramProtocol(AbstractDatagramProtocol):
  757      """
  758      Protocol for datagram-oriented transport, e.g. UDP.
  ...
  891      "FileWrapper",
  892      "ServerFactory",
  893:     "AbstractDatagramProtocol",
  894      "DatagramProtocol",
  895      "ConnectedDatagramProtocol",

/Users/wsanchez/Dropbox/Developer/Twisted/twisted/src/twisted/pair/ethernet.py:
   35  
   36  @implementer(IEthernetProtocol)
   37: class EthernetProtocol(protocol.AbstractDatagramProtocol):
   38      def __init__(self):
   39          self.etherProtos = {}

/Users/wsanchez/Dropbox/Developer/Twisted/twisted/src/twisted/pair/ip.py:
   44  
   45  @implementer(raw.IRawPacketProtocol)
   46: class IPProtocol(protocol.AbstractDatagramProtocol):
   47      def __init__(self):
   48          self.ipProtos = {}

/Users/wsanchez/Dropbox/Developer/Twisted/twisted/src/twisted/pair/rawudp.py:
   23  
   24  @implementer(raw.IRawDatagramProtocol)
   25: class RawUDPProtocol(protocol.AbstractDatagramProtocol):
   26      def __init__(self):
   27          self.udpProtos = {}

/Users/wsanchez/Dropbox/Developer/Twisted/twisted/src/twisted/pair/test/test_tuntap.py:
   38  from twisted.python.log import addObserver, removeObserver, textFromEventDict
   39  from twisted.internet.interfaces import IAddress, IReactorFDSet
   40: from twisted.internet.protocol import AbstractDatagramProtocol, Factory
   41  from twisted.internet.task import Clock
   42  from twisted.trial.unittest import SkipTest, SynchronousTestCase
   ..
 1305  
 1306  @implementer(IRawPacketProtocol)
 1307: class IPRecordingProtocol(AbstractDatagramProtocol):
 1308      """
 1309      A protocol which merely records the datagrams delivered to it.

8 matches across 5 files
```

Note also the subclass `DatagramProtocol`, which is used in even more places.

glyph: Those seem to be usages of the class itself; what I meant by "expected here" (rather than "used") is that `Factory.protocol` (or <any-subclass-of-`Factory`>`.protocol`) is only expected to return `AbstractDatagramProtocol`s in that one case.  I do recognize that we have datagram protocols all over the place :)

glyph: Going back up to the top of the thread here:

> `protocol.AbstractDatagramProtocol` should be a `Protocol`

Read from the perspective of "in the signature of `protocol.Factory`", then sure.  But from a higher-level perspective of "should the class exist": no, definitely not.  Its methods and lifecycle are totally incompatible.

The problem here is that `protocol.Factory` should never, ever return a `DatagramProtocol` (nor its unfortunately-named ancestor) because `protocol.Factory` is implicitly a factory for *stream-transport-oriented* Protocols, and `DatagramProtocol` is for processing unsequenced datagrams from a variety of sources; hence the fact that `datagramReceived` takes `addr` as well as `datagram`.

The reason it's `protocol.Factory` rather than `protocol.StreamFactory` on one side and `protocol.DatagramFactory` on the other is that datagram sockets do not get accepted to create connections; so there's no need for factory-level indirection or connection-level instances.  The socket just receives a datagram and the application gets told about it directly; that's it.  Hence, the various objects in `twisted.pair` and the reactor's own `listenUDP` just take an instance of a `DatagramProtocol` directly.

rodrigc (approved):
I think this PR should be merged as-is,
and the comments that @glyph made should be addressed in follow-up PR's.

src/twisted/test/test_tcp.py
wsanchez: OK, for that: https://twistedmatrix.com/trac/ticket/10013#ticket

Out of scope here.

src/twisted/internet/base.py
glyph: 👍 , `float` is the correct type.

wsanchez: Cool thanks for confirming. I was pretty sure but not 100%

PR #1432, opened 2020-10-10
configure pep517 isolated builds

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10010


adiroiban:
Hi @graingert . Can you please update to description of this PR to provide some info on why do we need this change?
What is wrong with the current build system ? :)

Thanks

graingert:
@adiroiban in the future all sdists will be built using pep517 isolated builds

tox.ini
```diff
@@ -28,9 +28,11 @@
 ; on travis in --diff mode
 ;
 [tox]
-minversion=2.4
+minversion=3.20.1
+requires=
+    virtualenv>=20.0.33
+    tox-wheel>=0.5.0
 skip_missing_interpreters=True
-toxworkdir=build/
```
adiroiban, line 33: is this change required for PEP517 ?

tox.ini
```diff
@@ -139,7 +143,7 @@ commands =
     coveralls-push: coveralls
 
     ; Build in a root folder so that buildbot can pick the result
-    wheel: python setup.py bdist_wheel --dist-dir {toxinidir}/dist
+    wheel: python -m pip wheel {toxinidir} --no-deps --use-pep517 --wheel-dir {toxinidir}/dist
```
adiroiban, line 147: my understanding is that pep517 is related to sdist and not to wheels.

Why do we have this change here?

There might be a good reason for that, but please update the description of this PR with a short summary of the changes.

For example explain why we need o update the wheel generation, why we need to change the tox build dir.

Maybe also add a link to https://www.python.org/dev/peps/pep-0517/ to make it easier for people to find the info required to review this PR.

adiroiban (changes requested):
Thanks for working on this.

I think that is good to get Twisted build up to latest specs.

I don't understand all the changes from this PR.

I left inline comments for that.

How can I manually test this? I think that the PR should description should have some instructions and not assume that anyone is familiar with PEP517.

I have run `python setup.py` on this branch, but I saw some warnings and I saw that it was copying the built docs inside the sdist

```
copying docs/_build/api/twisted.python.filepath.LinkError.html -> Twisted-20.3.0.dev0/docs/_build/api
```

This branch still has 

```
    apidocs: {toxinidir}/bin/admin/build-apidocs {toxinidir}/src/ apidocs
    narrativedocs: sphinx-build -aW -b html -d {toxinidir}/docs/_build {toxinidir}/docs {toxinidir}/docs/_build/
```

So maybe the sdist manifest should exclude the docs/_build ...and maybe docs at all.
I don't know why the docs are needed for the source dist.

Maybe a better title for this PR would be: 

"Build source distribution using PEP517 tools" ... or something like that.

Is this PR only about configuring the isolated builds, or is also about creating an sdist based on PEP517 tools?

So as part of this PR, maybe we should also add an sdist tox environment that will trigger 

Thanks!

tox.ini
graingert: seems to be a bug in tox-wheel

tox.ini
graingert: Isolated builds are an "optional" feature of pep517. pep517 relates to wheels and sdists

> A build frontend SHOULD, by default, create an isolated environment for each build, containing only the standard library and any explicitly requested build-dependencies.

tox.ini
adiroiban: Is this bug reported upstream? Is ok to have the build in .tox but I think that we should have a comment with the link to the bug... in case we want to come back the a custom build dir.

graingert: setuptools' bdist_wheel (ab)uses the `./build` dir as a build cache, which can be flakey (cache invalidation problems) so tox-wheel `rm -rf`s it. However when using `pip wheel --use-pep517` pip makes a fresh tmpdir and setuptools uses the `${tmpdir}/build` dir instead

https://github.com/ionelmc/tox-wheel/pull/9/files#diff-975220820dfc0998d4b34b2c84b5b94b55c8cb2b8f4172735905fd3ab5a194aaL148

tox.ini
```diff
@@ -28,9 +28,11 @@
 ; on travis in --diff mode
 ;
 [tox]
-minversion=2.4
+minversion=3.20.1
+requires=
+    virtualenv>=20.0.33
```
graingert:
```suggestion
    virtualenv>=20.0.35
```

tox.ini
adiroiban: Are we expecting to see developers creating their own wheels for Twisted?

And if a developer wants to create a wheel,  maybe they will use their own tools to create the wheel.

I am not familiar with the differences between `python setup.py bdist_wheel` and `pip wheel`

Maybe remove the `tox -e wheel` and just update .github/workflows/wheels.yaml

We don't have a `tox -e sdist` env.

graingert: This can be removed and replaced with a `tox -e release` job like `ldaptor`

graingert:
> How can I manually test this? I think that the PR should description should have some instructions and not assume that anyone is familiar with PEP517.

Any call to tox will use the new `pep517` build process (via tox-wheel)

> I have run `python setup.py` on this branch, but I saw some warnings and I saw that it was copying the built docs inside the sdist

You should avoid calling `setup.py` directly. `pip` will call `setuptools.build_meta` which will call `./setup.py` for you.


> So maybe the sdist manifest should exclude the docs/_build ...and maybe docs at all.
> I don't know why the docs are needed for the source dist.
> 
> Maybe a better title for this PR would be:
> 
> "Build source distribution using PEP517 tools" ... or something like that.

This is about both the sdist and the wheel

> Is this PR only about configuring the isolated builds, or is also about creating an sdist based on PEP517 tools?

It's about both

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = Twisted
-version = attr: src.twisted.__version__
+version = attr: twisted.__version__
```
graingert, line 3: `version = attr: twisted.__version__` works with or without pep517, `version = attr: src.twisted.__version__` only works without pep517

adiroiban (approved):
All good. Thanks!

PR #1433, opened 2020-10-11
move twisted.test.raiser to cython-test-exception-raiser dev extra

graingert:
another crack at this allows a single source .whl for mac/linux

https://github.com/twisted/twisted/pull/812/files

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/7945


graingert:
auditwheel should detect that it can build a plain source wheel...

adiroiban:
I am happy to see this move.

so... the code was moved here https://github.com/graingert/twisted-raiser .... 

The original code was introduced here in April 2008 as a workaround for pyrex and cpython - https://github.com/twisted/twisted/commit/7e748513728b24801ec91cfe98e0613f8807eb19

Does pyrex and cpython still have those issues?

The code from Twisted-raiser doesn't look like Twisted  specific at all... maybe rename it as test-cython-exception-raiser or something like that :)

Do we really need to move it in a separate repo?

maybe have all the code from the external twisted-raise it in src/test_cython_exception

src/twisted/python/_setup.py
```diff
@@ -121,23 +122,41 @@ def __init__(self, *args: object, **kwargs: object) -> None:
...
+_EXTENSIONS = (
+    [
+        ConditionalExtension(
+            "twisted.internet.iocpreactor.iocpsupport",
+            sources=[
+                "src/twisted/internet/iocpreactor/iocpsupport/iocpsupport.c",
+                "src/twisted/internet/iocpreactor/iocpsupport/winsock_pointers.c",
+            ],
+            libraries=["ws2_32"],
+            condition=lambda _: True,
```
adiroiban, line 154:
```suggestion
            condition=lambda _: _isCPython and sys.platform == "win32",
```

src/twisted/python/_setup.py
```diff
@@ -121,23 +122,41 @@ def __init__(self, *args: object, **kwargs: object) -> None:
...
+        ),
+    ]
+    if _isCPython and sys.platform == "win32"
```
adiroiban, line 157: I think that this condition can be removed. The ConditionalExtension was introduced to handle this type of conditions :) and with that, I think the original _checkCPython can be kept at the end

src/twisted/test/test_failure.py
```diff
@@ -22,10 +22,7 @@
 from twisted.trial.unittest import SynchronousTestCase
 
 
-try:
-    from twisted.test import raiser  # type: ignore[attr-defined]
-except ImportError:
-    raiser = None
+from cython_test_exception_raiser import raiser
```
adiroiban, line 25: Maybe it is worth adding a short comment here that this is a test helper which is hosted outside of the main twisted repo...or just a link to the repo.

adiroiban (changes requested):
Looks good. Happy to see this move.

Only minor comments.

Please update the PR title to the new name and description to include the link to the moved repo :)

I had an IRC chat with @graingert and we agreed that is best to keep the helper outside of the twisted/twisted repo.

I expect that the raiser code will never change :) ... so there is not much use to have all the PYPI publishing code in twisted/twisted

src/twisted/python/_setup.py
graingert: If this isn't present you get a tagged wheel

src/twisted/test/test_failure.py
graingert: My text editor already does this (go to definition)

src/twisted/python/_setup.py
adiroiban: Thanks. Then maybe it is worth adding a comment about why this extra condition is required :)

src/twisted/python/_setup.py
```diff
@@ -121,23 +122,41 @@ def __init__(self, *args: object, **kwargs: object) -> None:
...
+    else []
```
graingert:
```suggestion
    else []  # avoid adding ext_modules kwarg, to prevent platform tags
```

src/twisted/python/_setup.py
adiroiban: I see the code from  _extension_kwargs ... well... _extension_kwargs might need a docstring

src/twisted/python/_setup.py
adiroiban: If condition is alwasy True, this can be replaced with a normal extension 
We can remove the usage of ConditionalExtension

graingert: there's a load of stuff in there I don't understand about solaris, so I hesitated to remove it
https://github.com/twisted/twisted/blob/4a74c436e4d359f88ca5ff602c22ed740fe5012b/src/twisted/python/_setup.py#L235-L244

I think it can all go?  what's the WIN32 macro for?

adiroiban (approved):
This looks good.

It can be merged.

src/twisted/test/test_failure.py
adiroiban: there is no such feature in mousepad :)... it ok. I don't expect this code will be read soon. so don't worry

src/twisted/python/_setup.py
adiroiban: solaris can be removed as Twisted no longer supports Solaris

src/twisted/test/test_failure.py
adiroiban: one questions...  is cython also supported on PYPY? I see that the ImportError is no longer handled and that the test on pypy is green.

how come?

PR #1434, opened 2020-10-11
move extras to setup.cfg

graingert:
Currently the twisted.python._setup is very complicated and has a lot of somewhat unusual tests.

This moves the extras config from that `twisted.python._setup` into the more modern setuptools declarative format enabling us to remove a fair amount of python code and tests.

This is part of a series of work to move to a more "hypermodern" pep517/pep518 pure python Twisted wheel

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10011

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,6 +38,60 @@ zip_safe = False
 package_dir = =src
 packages = find:
 
+[options.extras_require]
+dev =
+    pyflakes >= 1.0.0
+    twisted-dev-tools >= 0.0.2
+    python-subunit
+    towncrier >= 17.4.0
+    twistedchecker >= 0.7.2
+    # force upgrades for rtd default packages: https://git.io/JU73V
+    alabaster~=0.7.12
+    commonmark~=0.9.1
+    docutils~=0.16.0
+    mock~=4.0
+    pillow~=7.2
+    readthedocs-sphinx-ext~=2.1
+    recommonmark~=0.6.0
+    sphinx~=3.2
+    sphinx-rtd-theme~=0.5.0
+tls =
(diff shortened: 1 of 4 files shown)


adiroiban:
I think that the description  of a PR should contain more details about why we have this PR.

My understanding is that we have this PR to reduce the number of packages we need for Twisted develop mode.

At the same time, with this change we will no longer need pillow / PIL for simple Twisted dev mode.

I am happy to see the packages moved to setup.cfg and outside of the python/_release.py

graingert:
@adiroiban I could split the setup.cfg move and the docs/rtd devolution into two differnt prs?

setup.cfg
```diff
@@ -38,6 +38,62 @@ zip_safe = False
 package_dir = =src
 packages = find:
 
+[options.extras_require]
+docs =
+    alabaster~=0.7.12
```
adiroiban: do we need to pin this?

I think that it can be removed as twisted is not using `alabaster` directly.

adiroiban (approved):
Looks good.

Please update the PR title and description to better describe why this change is needed.

I think that before merging this in trunk, we first need to fix trunk.

adiroiban:
> I could split the setup.cfg move and the docs/rtd devolution into two differnt prs?

no need to split, but I don't see any real description of this PR. This PR has a link to the trac ticket and the trac ticket has a link ot this PR :) ... looks like just working around the red tape  :)

The `move extras to setup.cfg` is like a commit message and not a PR / task description.

adiroiban (approved):
All good. Many thanks for the updated description!

PR #1446, opened 2020-10-14
Move iocpsupport to external dist

graingert:
another crack at this allows a single source .whl for mac/linux *and windows* :tada: https://github.com/twisted/twisted/pull/812/files

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/7945


setup.cfg
```diff
@@ -88,6 +88,7 @@ macos_platform =
     %(all_non_platform)s
 windows_platform =
     pywin32 != 226
+    twisted-iocpsupport ~= 0.0.1
```
graingert: Once there's support this distribution approach I'll do a v1 release

```suggestion
    twisted-iocpsupport ~= 1.0
```

adiroiban (changes requested):
Thanks for working on this.

This should make Twisted deployment simpler and faster.

I am +1 for moving iocpsupport as a separate package.
I am -1 for moving the source code outside of twisted/twisted

The tests for iocpsupport are part of the main twisted tests...so is best to have the code in the same repo.

Maybe have the twisted-iocpsupport source code moved to src/twisted_iocpsupprt package.
In root we can have a setup_twisted_iocpsupport.py which takes care of the wheel build.

We can have a github action to build the iocpsupport wheel when any file is modified inside the src/twisted_iocpsupport dir.
We can have a github action to publish the iocpsupport wheel when a tag starting with 'iocpsupport-VNNN' is created.

In tox, when tests are executed, iocpsupport is installed from the binary wheel.
In tox, we can have an option (commented / uncommented line) to build and use the code from src/twisted_iocpsupport when we want to develop iocpsupport

adiroiban:
For reference... and I think this should be in the PR description.

The code was moved here https://github.com/graingert/twisted-iocpsupport

src/twisted/internet/iocpreactor/iocpsupport.py
```diff
@@ -0,0 +1,27 @@
...
+    "get_accept_addrs",
+    "have_connectex",
+    "makesockaddr",
+    "maxAddrLen",
+    "recv",
+    "recvfrom",
+    "send",
+]
+
+from twisted_iocpsupport.iocpsupport import (
```
adiroiban, line 15: can we have all this as `from twisted_iocpsupport import recv` ?

And maybe add a deprecation so that in the future the code will import directly from twisted_iocpsupport and not from this redirection.

adiroiban:
For the newsfragment. My suggestion.

```
twisted.internet.iocpreactor.iocpsupport was moved to a separate package.
With this change the main Twisted package is a OS independent wheel.
```

src/twisted/internet/iocpreactor/iocpsupport.py
graingert: probably worth doing https://github.com/asottile/reorder_python_imports everywhere in a followup

setup.cfg
adiroiban: yes. I think that 1.0 wil do. I guess this will allow for 1.0.1 or a future 1.0.2 bugfix release.

src/twisted/internet/iocpreactor/iocpsupport.py
adiroiban: Well.. in Twisted we have backward compatibility policy, so you can't remove something from the public namespace without a prior deprecation warning.

But I can create the deprecation part after this is merged.

setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     Automat >= 0.8.0
     hyperlink >= 17.1.1
     attrs >= 19.2.0
+    twisted-iocpsupport ~= 0.0.1; platform_system == "Windows"
```
graingert:
```suggestion
    twisted-iocpsupport ~= 1.0; platform_system == "Windows"
```

adiroiban (approved):
Thanks. Let's merge this.

Have you executed blank on setup.py ?

graingert:
> Have you executed blank on setup.py ?

nope

PR #1448, opened 2020-10-14
Add type hints to twisted.internet.defer

wsanchez:
## Contributor Checklist:

* [x] The associated tickets in Trac are here: https://twistedmatrix.com/trac/ticket/10017 and here: https://twistedmatrix.com/trac/ticket/9816


src/twisted/python/compat.py
```diff
@@ -37,7 +37,7 @@
 from io import StringIO as NativeStringIO
 from io import TextIOBase
 from sys import intern
-from types import MethodType as _MethodType
+from types import FrameType, MethodType as _MethodType
```
wsanchez, line 40: Should all imports be prefixed with '_' or is '_MethodType' special?
It would be safer to move to `_foo.py` files with `foo.py` importing the public symbols if we are trying to keep the namespace cleaner than "look at `__all__`", no?

src/twisted/internet/defer.py
```diff
@@ -198,16 +223,61 @@ def setDebugging(on):
...
 
-def getDebugging():
+def getDebugging() -> bool:
     """
     Determine whether L{Deferred} debugging is enabled.
     """
     return Deferred.debug
 
 
+def _cancelledToTimedOutError(value: _T, timeout: float) -> _T:
```
wsanchez, line 233: Moved up for easier to manage definition order

src/twisted/internet/defer.py
```diff
@@ -198,16 +223,61 @@ def setDebugging(on):
...
+        raise TimeoutError(timeout, "Deferred")
+    return value
+
+
+# type note: this should be Callable[[object, ...], object] but mypy doesn't allow
+#   Callable[[object], object] is next best, but disallows valid callback signatures
+DeferredCallback = Callable[..., object]
+# type note: this should be Callable[[Failure, ...], object] but mypy doesn't allow
+#   Callable[[Failure], object] is next best, but disallows valid callback signatures
+DeferredErrback = Callable[..., object]
```
wsanchez: These two are used in public type signatures and I thought they would be useful to clients, so I didn't prefix them with `_`… but I'm not super happy with these signatures, since they aren't providing a good descriptions due to the limitations of `Callable`.

I think they remain useful in that when we eventually fix these, the user of these variables will then benefit from the checking.  I used the "next best" examples above when doing the initial typing work do distinguish callbacks from errbacks, for example, but we can't ship with those definitions because they don't allows for callbacks with extra arguments.

src/twisted/internet/defer.py
```diff
@@ -198,16 +223,61 @@ def setDebugging(on):
...
+
+_CallbackOrderedArguments = Tuple[object, ...]
+_CallbackKeywordArguments = Mapping[str, object]
```
wsanchez, line 282: These are standard for `*args` and `**kwargs`… so less compelling for clients, and maybe overkill here, but I think they help with reading the intended uses.

src/twisted/internet/defer.py
```diff
@@ -198,16 +223,61 @@ def setDebugging(on):
...
+_Callbacks = Tuple[
```
wsanchez: This one is only used internally.

src/twisted/internet/defer.py
```diff
@@ -232,32 +302,24 @@ class Deferred:
...
-
     @ivar _chainedTo: If this L{Deferred} is waiting for the result of another
         L{Deferred}, this is a reference to the other Deferred.  Otherwise,
         L{None}.
     """
 
     called = False
-    paused = False
-    _debugInfo = None
+    paused = 0
```
wsanchez, line 341: old code sure likes to pretend that `False is 0` and `True is 1`

src/twisted/internet/defer.py
```diff
@@ -766,103 +858,102 @@ def send(self, value=None):
...
+        class Adaptor:
+            def __init__(self, actual: Deferred) -> None:
+                self.actual = actual
+
+            def __call__(self, result: Future) -> None:
+                try:
+                    extracted = result.result()
+                except:
+                    extracted = Failure()
+                self.actual.callback(extracted)
```
wsanchez: This to avoid a function that sets an attribute on itself

src/twisted/internet/defer.py
```diff
@@ -1456,44 +1540,64 @@ def _inlineCallbacks(result, g, status):
...
+            assert appCodeTrace is not None
 
             # The contextvars backport and our no-op shim add an extra frame.
             if version_info < (3, 7):
                 appCodeTrace = appCodeTrace.tb_next
 
+            # type note: a number of ignore[union-attr] cases follow…
+            # It seems mypy isn't processing "assert is not None" as expected
+            # in some cases here.
+
```
wsanchez: I feel like I must be doing something wrong here but I don't see it

src/twisted/internet/defer.py
```diff
@@ -1636,36 +1747,41 @@ def loadData(url):
...
         except _DefGen_Return:
             raise TypeError(
                 "inlineCallbacks requires %r to produce a generator; instead"
                 "caught returnValue being used in a non-generator" % (f,)
             )
-        if not isinstance(gen, types.GeneratorType):
+        # type note: FIXME: error suggests type of gen is wrong:
+        #     Subclass of "Generator[Deferred, object, None]" and "GeneratorType"
+        #     cannot exist: would have incompatible method signatures
+        if not isinstance(gen, GeneratorType):  # type: ignore[unreachable]
```
wsanchez, line 1830: This suggests that `Generator[Deferred, object, None]]` is the wrong return type for `f`.
I could use other eyes on this.

src/twisted/python/compat.py
graingert: you can use https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

src/twisted/internet/defer.py
graingert: does this work better?

```python
class DeferredErrback(Protocol):
    def __call__(v: Failure, *args: object) -> object: ...
```

src/twisted/internet/defer.py
```diff
@@ -198,16 +223,61 @@ def setDebugging(on):
...
+        _CallbackKeywordArguments,
+    ],
+]
+
+_NONE_KWARGS = MappingProxyType({})  # type: _CallbackKeywordArguments
+
 # See module docstring.
 _NO_RESULT = object()
-_CONTINUE = object()
+_CONTINUE = cast(DeferredCallback, object())
```
graingert: an enum might be better here

```
class _Sentinel(enum.Enum):
    NO_RESULT = auto()
    CONTINUE = auto()
```

then you can use `Union[DeferredCallback, Literal[_Sentinel.CONTINUE]]`

src/twisted/internet/defer.py
```diff
@@ -741,21 +832,22 @@ def __str__(self) -> str:
 
     __repr__ = __str__
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator["Deferred"]:
```
```suggestion
    def __iter__(self) -> _DeferredT:
```

`Deferred` already has a `__next__`

src/twisted/internet/defer.py
graingert: what does `reveal_type(appCodeTrace)` say here?

graingert: might also be nicer with https://docs.python.org/3/library/traceback.html#traceback.walk_tb

src/twisted/internet/defer.py
wsanchez: Interesting… thanks, I'll try that

src/twisted/internet/defer.py
```diff
@@ -1636,40 +1747,53 @@ def loadData(url):
...
+    # It is:
+    #     "".join([
+    #         string.ascii_lowercase[int(c)] if c in string.digits else c
+    #         for c in "319AA2A8B18F4B8EA296D75F279EB07F"
+    #     ])
+    # …which is to say that it is a UUID encoded as letters.
+    # …which is to say: "it's not a string that will ever be used as a name in kwargs"
+    def run(
+        dbjAAcAiBbiFeBiEAcjgDhfFchjEBahF, f: Callable[..., object], *args: object, **kwargs: object
+    ) -> Deferred:
```
wsanchez, line 1792: OK perhaps this looks like a weird change.

The tests [`OtherPrimitivesTests.testLock`](https://github.com/twisted/twisted/blob/10017-mypy-defer/src/twisted/test/test_defer.py#L2131) and [`OtherPrimitivesTests.testSemaphore`](https://github.com/twisted/twisted/blob/10017-mypy-defer/src/twisted/test/test_defer.py#L2191) pass in `self=` to this method, which presumably means that `kwargs` here needs to accept any argument, hence the original method signature, which lacked `self` entirely, and the funky argument handling removed below to get it back.

That, of course, makes it hard for a tool like `mypy` to know what's going on with the method signature.  So here we put both `self` and the required first argument (which we can now type as a `Callable`) back in, but then we have to deal with the need to be able to pass `self` in as a keyword to the callable…

So we change the name of `self`, and we want that name not to be some other thing you might want to pass into the callable.  So I took a UUID and encoded it to letters.  If someone ends up trying to pass `dbjAAcAiBbiFeBiEAcjgDhfFchjEBahF` as a keyword here then I HATE THAT JERK SHUT UP GO AWAY.  So there.

src/twisted/internet/defer.py
```diff
@@ -1636,40 +1747,56 @@ def loadData(url):
...
+    #         for c in "319AA2A8B18F4B8EA296D75F279EB07F"
+    #     ])
+    # …which is to say that it is a UUID encoded as letters.
+    # …which is to say: "it's not a string that will ever be used as a name in kwargs"
+    def run(
+        dbjAAcAiBbiFeBiEAcjgDhfFchjEBahF,
+        f: Callable[..., object],
+        *args: object,
+        **kwargs: object
+    ) -> Deferred:
```
wsanchez: OK perhaps this looks like a weird change.

The tests [`OtherPrimitivesTests.testLock`](https://github.com/twisted/twisted/blob/10017-mypy-defer/src/twisted/test/test_defer.py#L2131) and [`OtherPrimitivesTests.testSemaphore`](https://github.com/twisted/twisted/blob/10017-mypy-defer/src/twisted/test/test_defer.py#L2191) pass in `self=` to this method, which presumably means that `kwargs` here needs to accept any argument, hence the original method signature, which lacked `self` entirely, and the funky argument handling removed below to get it back.

That, of course, makes it hard for a tool like `mypy` to know what's going on with the method signature.  So here we put both `self` and the required first argument (which we can now type as a `Callable`) back in, but then we have to deal with the need to be able to pass `self` in as a keyword to the callable…

So we change the name of `self`, and we want that name not to be some other thing you might want to pass into the callable.  So I took a UUID and encoded it to letters.  If someone ends up trying to pass `dbjAAcAiBbiFeBiEAcjgDhfFchjEBahF` as a keyword here then I HATE THAT JERK SHUT UP GO AWAY.  So there.

arnimarj:
It's great to see this annotation effort progressing. When my company started its typing efforts, we stubbed out the parts of Twisted we use regularly in our codebase. One thing we did in particular was to make `Deferred` generic:

```python
class Deferred(Awaitable[_T]):
	...

	def __await__(self) -> Generator[None, None, _T]:
		...

	def callback(self, value: _T) -> None:
		...
```

There is a bit more work figuring out for every twisted api we used which type a given deferred gets fire'd with, but it does thyen give accurate type information when awaiting a given deferred. Constructs like `DeferredQueue` can also be generic as a result:

```python
class DeferredQueue(Generic[_T]):
	def __init__(self, size: Optional[int] = ..., backlog: Optional[int] = ...):
		...

	def get(self) -> Deferred[_T]:
		...

	def put(self, obj: _T) -> None:
		...
```
Just throwing this in as an idea. Cheers.

src/twisted/python/compat.py
wsanchez: Neat. That does seem useful when running `mypy`. The concern here (and I assume the reason for the `_`-prefixed  `_MethodType`) is that symbols in the module's namespace may be accessed by client applications, and removing them arguably requires a deprecation.

Some time ago on a project at work, @glyph started creating separate implementation files and public files that imported only the public symbols, which makes it necessary to be explicit about what is exported, as opposed to `__all__`, which is sort of advisory, and I always do that now.

This is a bit of a rathole, but I'm really just wondering if `_MethodType` suggests something I'm not aware of.

src/twisted/internet/defer.py
wsanchez: OK here is what I tried for `DeferredErrback`:

```python
class DeferredErrback(Protocol):
    def __call__(self, failure: Failure, *args: object) ->object:
        ...
```

That yields some errors.  One class of those is from cases where a single function is being used as both a callback and errback (for example: `passthru`), and I think those just need to be cast.

Then we have these, and I'm wondering if this is due to the protocol being defined incorrectly, since I *think* these should match:

```console
src/twisted/internet/defer.py:545:46: error: Argument 2 to "addCallbacks" of "Deferred" has incompatible type "Callable[[Union[Failure, BaseException, None]], None]"; expected "DeferredErrback"  [arg-type]
            return self.addCallbacks(d.callback, d.errback)
                                                 ^
src/twisted/internet/defer.py:903:41: error: Argument 2 to "addCallbacks" of "Deferred" has incompatible type "Callable[[Failure], None]"; expected "DeferredErrback"  [arg-type]
            self.addCallbacks(maybeSucceed, maybeFail)
                                            ^
src/twisted/internet/defer.py:1652:23: error: Argument 1 to "addErrback" of "Deferred" has incompatible type "Callable[[Failure], Deferred]"; expected "DeferredErrback"  [arg-type]
            it.addErrback(handleCancel)
                          ^
```

src/twisted/internet/defer.py
wsanchez: Aha! `appCodeTrace` gets redefined and I wasn't re-checking it.

src/twisted/internet/defer.py
wsanchez: Nice, yes, that's better.

src/twisted/internet/defer.py
wsanchez: OK, yes.

src/twisted/internet/defer.py
wsanchez: @graingert `Literal` is new in Python 3.8… so importing it blows up in earlier versions.

I can't think of a correct alternative pre-3.8… I could do a hack like:

```python
try:
    from typing import Literal as _Literal
except ImportError:
    Literal = Optional
else:
    Literal = _Literal
```

Which I think will make `mypy` work fine, but lie at runtime about the type.  In this case (especially since this is an internal thing), I think that in practice that turns out to be fine…?  But it's a hack so I'm wondering if you are aware of a better option.

wsanchez: Maybe `if TYPE_CHECKING` instead of `try`/`except`.

mthuurne: > `Literal` is new in Python 3.8… so importing it blows up in earlier versions.

`Literal` exists in [`typing-extensions`](https://pypi.org/project/typing-extensions/) and mypy supports that. It's likely we'll want to use `typing-extensions` at some point, so this might be a good moment to add it as a dependency.

src/twisted/internet/defer.py
graingert: You could use a positional only argument, or `_v_`

wsanchez: Posargs are Py3.8+. I'll add a comment that it would be a better alternative so that when we eventually deprecate Py37-, we can use that instead.

By `_v_`, I assume you mean that as a literal name.  I also thought about `__self` or similar, but went with the above to make it basically impossible for there to be a conflict.  Perhaps overkill, but it's done and doesn't cost anything other than some cognitive load fo someone reading the code, which the comment hopefully abates.

wsanchez:
@arnimarj: Thanks for the input.

> It's great to see this annotation effort progressing. When my company started its typing efforts, we stubbed out the parts of Twisted we use regularly in our codebase. One thing we did in particular was to make `Deferred` generic:
> 
> ```python
> class Deferred(Awaitable[_T]):
> 	...
> 
> 	def __await__(self) -> Generator[None, None, _T]:
> 		...
> 
> 	def callback(self, value: _T) -> None:
> 		...
> ```

Does the use of `Awaitable` above add anything?  I ask because:

```python
>>> from twisted.internet.defer import Deferred
>>> d = Deferred()
>>> from typing import Awaitable
>>> isinstance(d, Awaitable)
True
```

I'm in favor of being explicit about the intended compliance with `Awaitable`, so I think it's good; just wondering if there's more to it, since I didn't think of doing that perviously.

src/twisted/internet/defer.py
glyph: I haven't had time to try this out yet, but I believe what you want here is something like

```
DeferredCallback = TypeVar("DeferredCallback", bound=DeferredCallbackCallableProtocol)
DeferredErrback = TypeVar("DeferredErrback", bound=DeferredErrbackCallableProtocol)
```

This doesn't quite address the binding of the result-type typevar into the callback chain.

I checked Pilot's temporary Deferred stubs, and apparently we just `Any` away this problem, so I think we can start off pretty lax here and tighten up in subsequent PRs if it proves too much of a puzzle.

src/twisted/python/compat.py
glyph: My current attempt at a solution to this problem is https://pypi.org/project/publication/ but it doesn't really interact sensibly with mypy yet.

wsanchez:
> I'm in favor of being explicit about the intended compliance with `Awaitable`, so I think it's good; just wondering if there's more to it, since I didn't think of doing that perviously.

OK, having tried it, I now see this error from Mypy:

```console
src/twisted/internet/defer.py:866:17: error: Incompatible types in assignment (expression has type "Callable[[], Deferred[_DeferredResultT]]", base class "Awaitable" defined the type as "Callable[[], Generator[Any, None, _DeferredResultT]]")  [assignment]
        __await__ = __iter__
```

So that Mypy is performing more checks based on this explicit declaration of intent, we see a compelling reason to do so.

Now to figure out what the correct fix to that is… :-)

src/twisted/python/compat.py
wsanchez: OK let's put a pin in that for now.

wsanchez:
I took at stab at making `Deferred` generic, and I think that it will be easier for review purposes if we do that separately from the work already present here, so I'm going to put this up for review now, and take on the generic work in a followup PR.

src/twisted/internet/defer.py
wsanchez: I gave that a shot and it's… better but also having issues.  Let's do that in a separate PR.

Ideally, we'd have this in the toolbox soon: https://github.com/python/mypy/issues/3157

glyph:
> I took at stab at making `Deferred` generic, and I think that it will be easier for review purposes if we do that separately from the work already present here, so I'm going to put this up for review now, and take on the generic work in a followup PR.

I'm definitely in favor of incremental progress, but there are a bunch of changes here (treating `self` as a type var) that will need to be reversed when we make `Deferred` generic, so it seems like there might be a bunch of extra lateral movement doing it that way.

src/twisted/internet/defer.py
```diff
@@ -295,60 +362,80 @@ def __init__(self, canceller=None):
...
         Add a pair of callbacks (success and error) to this L{Deferred}.
 
         These will be executed when the 'master' callback is run.
 
         @return: C{self}.
-        @rtype: a L{Deferred}
         """
+        # Default value used to be None and callers may be using None
+        if errback is None:
+            errback = passthru  # type: ignore[unreachable]
```
glyph: Why make this (incorrectly) unreachable rather than making the arguments properly `Optional`?

glyph: To be clear I'm probably in favor of making this signature change later, `None`s are a bummer, but it seems like the first iteration of the type signature should be more descriptive of the existing behavior.

src/twisted/internet/defer.py
```diff
@@ -295,60 +362,80 @@ def __init__(self, canceller=None):
...
-        errbackKeywords=None,
-    ):
+        self: _DeferredT,
+        callback: DeferredCallback,
+        errback: DeferredErrback = passthru,
+        callbackArgs: _CallbackOrderedArguments = (),
+        callbackKeywords: _CallbackKeywordArguments = _NONE_KWARGS,
+        errbackArgs: _CallbackOrderedArguments = (),
+        errbackKeywords: _CallbackKeywordArguments = _NONE_KWARGS,
+    ) -> _DeferredT:
```
glyph: In addition to self-as-typevar having some issues with mypy's internals, there's also the issue here that this signature is conceptually wrong too; `_DeferredT.addCallback(x->y)`  does *not* return `_DeferredT` in the future generic version, it returns `Deferred[Y]`.  I think if we want to go with simpler stubs for the first iteration, it would be better to simply type this as `Deferred`.  (I think the idea here is to capture `DeferredList`'s subclass behavior, but `DeferredList` being a subclass ought to be irrelevant; the public attributes `fireOnOneCallback` etc. are exposed by accident from the constructor, ought to be private, and it has no differing public behaviors beyond those.

glyph:
> It's great to see this annotation effort progressing. When my company started its typing efforts, we stubbed out the parts of Twisted we use regularly in our codebase. One thing we did in particular was to make `Deferred` generic:

@arnimarj this has always been part of the plan :). https://twistedmatrix.com/trac/ticket/9816

src/twisted/internet/defer.py
glyph: (One reason to make this change later is that if there's going to be this kind of runtime/type-check-time distinction, we should add some deprecation warnings to the runtime cases as well so that we can get the ball rolling on the removal, and that's a bunch of additional code.)

arnimarj:
> @arnimarj: Thanks for the input.
> 
> > It's great to see this annotation effort progressing. When my company started its typing efforts, we stubbed out the parts of Twisted we use regularly in our codebase. One thing we did in particular was to make `Deferred` generic:
> > ```python
> > class Deferred(Awaitable[_T]):
> > 	...
> > 
> > 	def __await__(self) -> Generator[None, None, _T]:
> > 		...
> > 
> > 	def callback(self, value: _T) -> None:
> > 		...
> > ```
> 
> Does the use of `Awaitable` above add anything? I ask because:
> 
> ```python
> >>> from twisted.internet.defer import Deferred
> >>> d = Deferred()
> >>> from typing import Awaitable
> >>> isinstance(d, Awaitable)
> True
> ```
> 
> I'm in favor of being explicit about the intended compliance with `Awaitable`, so I think it's good; just wondering if there's more to it, since I didn't think of doing that perviously.

For our purposes subclclassing `Awaitable` was enough to make this a useful annotation without having to stub out `__await__`. Since our strategy involved having separate stubs, we opted for brevity. Given the strategy in this PR, subclassing `Generic` would be sufficient to achieve the same results, I believe.

arnimarj:
> > It's great to see this annotation effort progressing. When my company started its typing efforts, we stubbed out the parts of Twisted we use regularly in our codebase. One thing we did in particular was to make `Deferred` generic:
> 
> @arnimarj this has always been part of the plan :). https://twistedmatrix.com/trac/ticket/9816

Glad to hear it. We have a sizable twisted codebase, and this trick has proved to be immensely useful. Again, its a pleasure to watch this effort progress.

src/twisted/internet/defer.py
wsanchez: I'm not sure why you are objecting to `self`-as-typevar being bad for mypy, since the mypy docs suggest it: https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self

But yes, ok, the signature is wrong.  I'll fix that.

glyph: Mostly this annotation:

> This feature is experimental. Checking code with type annotations for self arguments is still not fully implemented. Mypy may disallow valid code or allow unsafe code.

Plus some personal experiences with bugs when attempting to do this.  For all I know all the bugs *I* hit are fixed by now but the presence of that warning suggests to me that it's not fully cooked yet.

(I *want* this feature to be done; our work codebase has a bajillion classmethod constructors or generative modifiers which have slightly-inaccurate types because we've been burned by this and we don't want to start depending on it until they at least *claim* it works right)

wsanchez: Somehow that box was rendered invisibly to me until now.  :-)

OK… well, that's a real bummer.  OK.  Fine.  Yeah.  OK.  I'll stop doing that for now.

src/twisted/internet/defer.py
wsanchez: I guess I was assuming that there weren't many callers passing in `None` versus not passing in the argument at all, and if they did so… a type warning wouldn't be too onerous.

I agree that a runtime issue would require a deprecation and that should be another PR.

But I see a type error as a "very soft" deprecation that you can ignore your way past until such time as you get a formal deprecation.

If you feel that's a bridge too far, I'll make these optional, but as a client, I wouldn't mind cleaning this up as part of better type checking.

glyph: I think I am committed to having no further opinion on this.  Mainly I think it's unnecessary churn which would also cause valid clients adopting type-checking for the first time to have to rewrite stuff in the process (particularly it's annoying to have to replace a `None` with a `lambda x:x` or an `import passthru`) but also I think this is an edge case, and the *main* problem is that it provokes too much discussion about the pros & cons, so I'll cut that short and say "do what you like with it, either option is OK" :).

src/twisted/internet/defer.py
glyph: Might I suggest the slightly-less-elaborate name `self_319AA2A8B18F4B8EA296D75F279EB07F` ?

glyph: Also, this sounds like it ought to be a bug on mypy?

wsanchez: > Also, this sounds like it ought to be a bug on mypy?

Why? The error from mypy was in the test that passed in `self=`, and the warning was that `self` is being passed in twice, which is correct, I think.

wsanchez: > Might I suggest the slightly-less-elaborate name `self_319AA2A8B18F4B8EA296D75F279EB07F` ?

Sure, I don't care what the color of the bike shed ends up being.

src/twisted/internet/defer.py
wsanchez: Unless there's a consensus otherwise, I think it's better to clean up the type here as long as it stays compatible. `lambda x:x` is a minor drag for `errback`, sure, but the other values are `()` and `{}`… it just seems easy enough to deal with.

wsanchez: We can also have a change of heart later if it's causing headaches.

wsanchez:
#1454 adds to this PR by making Deferred generic

src/twisted/internet/defer.py
glyph: > Why? The error from mypy was in the test that passed in `self=`, and the warning was that `self` is being passed in twice, which is correct, I think.

Ah, I misunderstood, I thought that that actually worked at runtime given that the test was doing it. I see from a quick experiment that it doesn't, so, nevermind!

README.rst
```diff
@@ -1,5 +1,5 @@
 Twisted
-=======
+========
```
wsanchez: Old thing has false positive for "merge failure" check

src/twisted/internet/defer.py
```diff
@@ -726,140 +838,145 @@ def _runCallbacks(self):
...
+        class Adaptor:
+            def __init__(self, actual: Deferred[Any]) -> None:
+                self.actual = actual
+
+            def __call__(self, result: Future) -> None:
+                try:
+                    extracted = result.result()
+                except BaseException:
+                    extracted = Failure()
+                self.actual.callback(extracted)
```
wsanchez: This to avoid a function that sets an attribute on itself

glyph: It might be less disruptive to use `nonlocal`; the only reason this used a function attribute was to avoid py3-only features.

src/twisted/internet/base.py
```diff
@@ -354,13 +356,16 @@ def getHostByName(
...
+    ) -> Deferred[str]:
         try:
             address = socket.gethostbyname(name)
         except OSError:
             msg = "address {!r} not found".format(name)
             err = error.DNSLookupError(msg)
-            return defer.fail(err)
+            # type note: returning a failing Deferred is akin to raising...
+            #   it's awkward to express that with typing a return value, where
+            #   exceptions are not factored in
```
glyph: I think that the right answer here is to have `defer.fail` explicitly become a `Deferred[Any]`, so that it will match shapes like this one.

src/twisted/internet/base.py
```diff
@@ -354,13 +356,16 @@ def getHostByName(
...
+            return defer.fail(err)  # type: ignore[return-value]
```
glyph: If you can get rid of this modification you can avoid worrying about the codecov error here ;-)

src/twisted/internet/defer.py
```diff
@@ -674,14 +784,16 @@ def _runCallbacks(self):
...
-                    current.result = failure.Failure(captureVars=self.debug)
+                    current.result = Failure(captureVars=self.debug)
                 else:
                     if isinstance(current.result, Deferred):
                         # The result is another Deferred.  If it has a result,
                         # we can take it and keep going.
-                        resultResult = getattr(current.result, "result", _NO_RESULT)
+                        resultResult = getattr(
+                            current.result, "result", _Sentinel._NO_RESULT
+                        )
```
glyph: This code path is extremely hot, so I'd be a little wary of adding the two extra `getattr` calls here. (I wish speed center could quickly just, you know, tell us the performance impact instead of guessing like this, but c'est la vie)

glyph: (In other words, let's still cache `_NO_RESULT` in a global even if we want its type to be an enum for signature friendliness)

glyph (commented):
My main wonder is "why are there so many `Deferred[Any]`s in here, when they should clearly be `Deferred[_DeferredResultT]`s" ?  I have a vague memory of having talked about this before, so maybe I'm just failing to remember.

src/twisted/internet/defer.py
wsanchez: OK [replaced](https://github.com/twisted/twisted/pull/1448/commits/1c415f9d2977fd9e2a4acb51f46f872254408c8a) with use of `nonlocal`

src/twisted/internet/defer.py
```diff
@@ -892,57 +1004,27 @@ def main(reactor):
...
-    @raise TimeoutError: If C{value} is a L{Failure} that is a L{CancelledError}.
-    @raise Exception: If C{value} is a L{Failure} that is not a L{CancelledError},
-        it is re-raised.
-
-    @since: 16.5
-    """
-    if isinstance(value, failure.Failure):
-        value.trap(CancelledError)
-        raise TimeoutError(timeout, "Deferred")
-    return value
```
wsanchez, line 929: See above; this was moved up in the file

src/twisted/internet/defer.py
wsanchez: New diff only adds type hints, no code changes other than to make `actual` its own variable (marked `nonlocal` in `uncancel`, where it is set), instead of an attribute on the `adapt` function.

src/twisted/internet/base.py
wsanchez: Him, OK.  It's unfortunate, since this does weaken the type checking significantly, but it's probably best not to over-do it at this point.

src/twisted/internet/base.py
wsanchez: Done: https://github.com/twisted/twisted/pull/1448/commits/9f6d028459aa14c425bffdab1da4e84f3b3429f7

src/twisted/internet/base.py
wsanchez: Done: https://github.com/twisted/twisted/pull/1448/commits/9f6d028459aa14c425bffdab1da4e84f3b3429f7

src/twisted/internet/defer.py
wsanchez: Done: https://github.com/twisted/twisted/pull/1448/commits/f35df2083e6a78c13230ca689832654d0d636d9f

src/twisted/internet/defer.py
```diff
@@ -1393,24 +1535,30 @@ class _CancellationStatus:
...
 
-@failure._extraneous
-def _inlineCallbacks(result, g, status):
+@_extraneous
+def _inlineCallbacks(
+    result: object,
+    gen: Union[
+        Generator[Deferred[_T], object, None],
+        Coroutine[Deferred[_T], object, None],
+    ],
```
wsanchez, line 1545: Repeating a comment from @mstojcevich in @glyph's suggestion PR here so we don't lose it:

> Is None correct for the generator/coroutine return type for these? I think it should be Optional[object] since this handles pulling the value off of StopIteration(x) when the generator returns x.
> 
> And I don't think YieldType is necessarily a Deferred when gen is a Generator.
> 
> I think the same concerns apply to the original PR fwiw.

wsanchez:
Regarding code coverage: I've added tests for some new code here in https://github.com/twisted/twisted/pull/1448/commits/aa46f26d8170425c4e336f4ebe36b4461fed10ae.  The rest is pre-existing issues that are kind of out-of-scope for this PR.

I did try to address one of those here: https://github.com/twisted/twisted/pull/1517

wsanchez:
Now that Twisted exports type hints, I'm seeing errors in my projects like:

```console
error: Returning Any from function declared to return "Deferred"  [no-any-return]
```

Because, for example, `Deferred.fromFuture()` isn't typed, but other parts of Twisted are.  It would really be great to move thing PR along…

glyph:
> Now that Twisted exports type hints

To be honest, I'm wondering if we should roll this back.  Our type hints are in pretty rough shape, and are definitely going to be evolving over the next couple of releases.  Is there a way to make this easily opt-in for mypy consumers?

src/twisted/internet/defer.py
```diff
@@ -379,44 +520,44 @@ def addTimeout(self, timeout, clock, onTimeoutCancel=None):
...
 
-        def convertCancelled(value):
+        def convertCancelled(value: object) -> object:
             # if C{deferred} was timed out, call the translation function,
             # if provdied, otherwise just use L{cancelledToTimedOutError}
             if timedOut[0]:
-                toCall = onTimeoutCancel or _cancelledToTimedOutError
+                # Default value used to be None and callers may be using None
+                if onTimeoutCancel is None:
+                    toCall = _cancelledToTimedOutError  # type: ignore[unreachable]
```
glyph: If we want to support this behavior, we should be testing it.  Rather than a `ignore[unreachable]`, specify something like `onTimeoutCancelLegacy: Optional[Callable[[object, float], object] = onTimeoutCancel` up at the top of the function, then explain that we don't want it publicly typed to allow None but we may want to allow non-mypy users to keep passing it

src/twisted/internet/defer.py
```diff
@@ -659,7 +804,15 @@ def _runCallbacks(self):
                 try:
                     current._runningCallbacks = True
                     try:
-                        current.result = callback(current.result, *args, **kw)
+                        # type note: mypy sees `callback is _CONTINUE` above and
+                        #    then decides that `callback` is not callable.
+                        #    This goes away when we use `_Sentinel._CONTINUE`
+                        #    instead, but we don't want to do that attribute
```
glyph, line 807: What?  Why does mypy think these types are distinct at all?  It should be able to see them as a pure type alias?

src/twisted/internet/defer.py
```diff
@@ -741,125 +894,129 @@ def __str__(self) -> str:
 
     __repr__ = __str__
 
-    def __iter__(self):
+    def __iter__(self) -> "Deferred[_DeferredResultT]":
```
glyph, line 894: Ugh I can't wait to drop 3.6 so we don't have all this gratuitous quoting

src/twisted/internet/defer.py
```diff
@@ -741,125 +894,129 @@ def __str__(self) -> str:
...
             result.value.__failure__ = result
             raise result.value
         else:
             raise StopIteration(result)
 
     # For PEP-492 support (async/await)
-    __await__ = __iter__
+    # type note: base class "Awaitable" defined the type as:
+    #     Callable[[], Generator[Any, None, _DeferredResultT]]
+    __await__ = __iter__  # type: ignore[assignment]
```
glyph: I don't understand the issue being ignored here.  Isn't this compatible with `Awaitable.__await__` ?  That would seem to be the whole point?

src/twisted/internet/defer.py
```diff
@@ -741,125 +894,129 @@ def __str__(self) -> str:
...
-        @rtype: L{asyncio.Future}
+        @return: A L{Future} which will fire when the L{Deferred} fires.
         """
         try:
             createFuture = loop.create_future
         except AttributeError:
-            from asyncio import Future
 
-            def createFuture():
+            def createFuture() -> Future:
```
glyph: Do we support a version of Python where this `AttributeError` can still happen?  Can we just nuke the whole `try`?

src/twisted/internet/defer.py
```diff
@@ -1577,7 +1760,11 @@ class _InternalInlineCallbacksCancelledError(Exception):
     """
 
 
-def inlineCallbacks(f):
+# type note: "..." is used here because we don't have a better way to express
+#     that the same arguments are accepted by the returned callable.
```
glyph, line 1755: Is there a mypy bug we could link to here?

src/twisted/internet/defer.py
```diff
@@ -1638,40 +1825,58 @@ def loadData(url):
...
         except _DefGen_Return:
             raise TypeError(
                 "inlineCallbacks requires %r to produce a generator; instead"
                 "caught returnValue being used in a non-generator" % (f,)
             )
-        if not isinstance(gen, types.GeneratorType):
+        # type note: FIXME: error suggests type of gen is wrong:
+        #     Subclass of "Generator[Deferred, object, None]" and "GeneratorType"
+        #     cannot exist: would have incompatible method signatures
+        if not isinstance(gen, GeneratorType):  # type: ignore[unreachable]
```
glyph, line 1830: Can you file / link a ticket for this?

src/twisted/internet/defer.py
```diff
@@ -1681,37 +1886,37 @@ def run(*args, **kwargs):
...
-    def __aexit__(self, exc_type, exc_val, exc_tb):
+    def __aexit__(self, exc_type: bool, exc_val: bool, exc_tb: bool) -> Deferred[bool]:
         self.release()
         # We return False to indicate that we have not consumed the
         # exception, if any.
         return succeed(False)
 
+    @abstractmethod
+    def acquire(self: _ConcurrencyPrimitiveT) -> Deferred[_ConcurrencyPrimitiveT]:
+        pass
```
glyph, line 1902: This needs a docstring, which will hopefully also make it stop complaining about coverage.

src/twisted/internet/defer.py
```diff
@@ -1681,37 +1886,37 @@ def run(*args, **kwargs):
...
+
+    @abstractmethod
+    def release(self) -> None:
+        pass
```
glyph, line 1906: Also docstring here.

src/twisted/internet/defer.py
```diff
@@ -1907,7 +2117,7 @@ def put(self, obj):
         else:
             raise QueueOverflow()
 
-    def get(self):
+    def get(self) -> Deferred[Any]:
```
glyph: Seems like this should also be a generic, and this should be `Deferred[T]`?  If users are using it for something heterogenous, `T` can always be `Any`.

src/twisted/internet/defer.py
```diff
@@ -1966,16 +2173,15 @@ def __init__(self, name, scheduler=None):
         if scheduler is None:
             from twisted.internet import reactor
 
-            scheduler = reactor
+            scheduler = cast(IReactorTime, reactor)
```
glyph, line 2167: Hmm. This `cast` seems like it's a problem, but probably not *this* PR's problem.

src/twisted/internet/interfaces.py
```diff
@@ -48,7 +48,7 @@
         ServerFactory,
     )
     from twisted.internet.ssl import ClientContextFactory
-    from twisted.names.dns import Query
+    from twisted.names.dns import Query, RRHeader
```
glyph, line 51: We can add this to `.coveragerc` to avoid this type of bogus coverage gap:
```
[report]
exclude_lines =
    pragma: no cover
    if TYPE_CHECKING
    \s*\.\.\.$
```

src/twisted/test/test_defer.py
```diff
@@ -176,6 +176,44 @@ def testTwoCallbacks(self):
         self.assertEqual(self.callbackResults, (("hello",), {}))
         self.assertEqual(self.callback2Results, (("hello",), {}))
 
+    def testAddCallbacksNoneErrback(self):
```
glyph: Nitpick: new test names should be `test_wordsWordsWords`. Not gonna block on that though ;-)

mypy.ini
```diff
@@ -20,6 +20,18 @@ warn_return_any          = True
 warn_unreachable         = True
 warn_unused_ignores      = True
 
+# These are too strict for us at the moment
```
glyph, line 23: "... but we should gradually enable them as time allows" :)

glyph (approved):
If you could close a couple more of the coverage gaps (mainly the uncovered code, and the `_ConcurrencyPrimitive` changes) I think this looks good to me!  I suspect we'll find some issues once it lives in the wild, but this looks like a solid first cut.

src/twisted/internet/defer.py
wsanchez: Yeah I hope `pyupgrade` deals with them.

src/twisted/internet/defer.py
graingert: https://github.com/python/mypy/issues/8645 ?

glyph: I'd rather link to a ticket than a PR, but presumably if the PR gets forked or reopened then someone will link from there, so that sounds fine.

graingert: That's the tracking issue not a PR?

glyph: Wow, cosmic rays flipping bits in my brain I guess. Great, yes, perfect thing to link to, thank you!

src/twisted/internet/defer.py
wsanchez: Part of why I changed this is that I mis-read the doctoring and thought that passing in `None` wasn't documented, but I see it there now.

I'm going to revert this and leave it as a future exercise to make it not an optional callable. I think that doing so properly should probably include making `_cancelledToTimedOutError` public so that one can pass it in instead of `None`.

src/twisted/internet/defer.py
wsanchez: Type aliases are improving in Python 3.10, I think.

src/twisted/internet/defer.py
wsanchez: Here is the full error message:

```console
src/twisted/internet/defer.py:919:17: error: Incompatible types in assignment (expression has type "Callable[[], Deferred[_DeferredResultT]]", base class "Awaitable" defined the type as
"Callable[[], Generator[Any, None, _DeferredResultT]]")  [assignment]
```

So the issue seems to be that `Awaitable` expects `__iter__` to return a `Generator`, and not any `Iterable`.  We could make `__await__` a generator function, but that kind of seems gratuitous?

src/twisted/internet/defer.py
wsanchez: Docs say `create_future` was new in 3.5.2, so… I'll nuke it.

src/twisted/internet/defer.py
wsanchez: The problem here is that mypy thinks the reactor is `ModuleType`.  It should be fixed in another PR, though I'm not sure what more module voodoo we can do about it.
My guess is that it would really be better if the normal way to get the installed reactor was via a callable and not a weird module import that's really a reactor.

src/twisted/test/test_defer.py
wsanchez: Fine by me. DeferredTests doesn't… ok it does sometimes.  I was doing "when in Rome". Fixing.

mypy.ini
wsanchez: "at the moment" :-)

src/twisted/internet/interfaces.py
wsanchez: OK adding.  Skipping `pragma: no cover`, since we don't appear to use it at all in Twisted, which I think is intentional, so if that changes, we can add it?

src/twisted/internet/defer.py
glyph: I agree that it's gratuitous, and if this were new code I'd probably say "let's just make it compatible" but given that I don't want any more runtime changes than necessary, can you just file a typeshed bug and link it here?  Given that this works fine at runtime right now this does seem wrong.

glyph: Sorry, by "this" I mean the stdlib `__await__` stub, not our code

src/twisted/test/test_defer.py
glyph: Thanks!  And yeah, `DeferredTests` is *super* old.

src/twisted/internet/defer.py
glyph: Yeah I'd be very happy if Mypy just encouraged us to get rid of the gross `from twisted.internet import reactor` idiom entirely.

src/twisted/internet/interfaces.py
glyph: SGTM

src/twisted/internet/defer.py
wsanchez: https://github.com/python/typeshed/issues/5125

PR #1453, opened 2020-10-16
Fix incorrect fields in docstrings

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10021

### DIFF
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -1582,7 +1582,7 @@ class Key:
         @type comment: L{bytes}
 
         @param passphrase: Passphrase for a private key.
-        @type comment: L{bytes}
+        @type passphrase: L{bytes}
 
         @rtype: L{bytes}
         """
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -1570,13 +1570,19 @@ class Key:
     def _toString_OPENSSH(self, subtype=None, comment=None, passphrase=None):
         """
         Return a public or private OpenSSH string.  See
-        _fromString_PUBLIC_OPENSSH and _fromPrivateOpenSSH_PEM for the
-        string formats.  If extra is present, it represents a comment for a
-        public key, or a passphrase for a private key.
+        L{_fromString_PUBLIC_OPENSSH} and L{_fromPrivateOpenSSH_PEM} for the
+        string formats.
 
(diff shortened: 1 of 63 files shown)


src/twisted/conch/ssh/keys.py
```diff
@@ -1570,13 +1570,19 @@ def _toPrivateOpenSSH_PEM(self, passphrase=None):
...
+        @param subtype: A subtype to emit.  Only supported for private keys,
+            for which the currently supported subtypes are C{'PEM'} and C{'v1'}.
+            If not given, an appropriate default is used.
+        @type subtype: L{str} or L{None}
+
+        @param comment: Comment for a public key.
+        @type comment: L{bytes}
+
+        @param passphrase: Passphrase for a private key.
+        @type comment: L{bytes}
```
twm:
```suggestion
        @type passphrase: L{bytes}
```

src/twisted/internet/interfaces.py
```diff
@@ -1588,9 +1588,6 @@ class IFileDescriptor(ILoggingContext):
 
     def fileno() -> object:
         """
-        @raise: If the descriptor no longer has a valid file descriptor
-            number associated with it.
```
twm, line 1592: It seems like a deficiency in the interface that there is no exception type defined for this. Oh well.

twm (approved):
This looks good. I didn't run pydoctor, but all these changes are clearly improvements. I did note one typo.

The only other thought I had is that there are several places documented as `Exception` that might technically produce `BaseException`. However that's always true so 🤷‍♂️ .

Thank you!

mthuurne:
> The only other thought I had is that there are several places documented as `Exception` that might technically produce `BaseException`. However that's always true so man_shrugging .

I think the purpose of `@raise` is to document exceptions the caller might want to catch and the exceptions that inherit directly from `BaseException` are not ones you should be catching.

src/twisted/conch/ssh/keys.py
mthuurne: Ah, that's indeed a copy-paste error on my part.

I guess that also answers the question of whether twisted/pydoctor#229 is worth implementing.

src/twisted/internet/interfaces.py
mthuurne: None of the implementations within Twisted raise an exception though. Some return -1 if the descriptor is no longer valid.

PR #1456, opened 2020-10-21
 Convert parameters popped from **kwargs to keyword-only

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10025

### DIFF
--- a/src/twisted/protocols/amp.py
+++ b/src/twisted/protocols/amp.py
@@ -2109,21 +2109,25 @@ class StartTLS(Command):
 
     responseType = _TLSBox
 
-    def __init__(self, **kw):
+    def __init__(self, *, tls_localCertificate=None, tls_verifyAuthorities=None, **kw):
         """
         Create a StartTLS command.  (This is private.  Use AMP.callRemote.)
 
-        @keyword tls_localCertificate: the PrivateCertificate object to use to
-        secure the connection.  If it's None, or unspecified, an ephemeral DH
+        @param tls_localCertificate: the PrivateCertificate object to use to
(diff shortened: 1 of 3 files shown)


src/twisted/protocols/amp.py
```diff
@@ -2109,21 +2109,25 @@ class StartTLS(Command):
...
+        @param tls_verifyAuthorities: a list of Certificate objects which
         represent root certificates to verify our peer with.
         """
         if ssl is None:
             raise RuntimeError("TLS not available.")
-        self.certificate = kw.pop("tls_localCertificate", _NoCertificate(True))
-        self.authorities = kw.pop("tls_verifyAuthorities", None)
+        self.certificate = (
+            _NoCertificate(True)
+            if tls_localCertificate is None
```
graingert, line 2127: @mthuurne this is a bit different to the previous behaviour:


```python
# this test passes on trunk and fails on this PR
from twisted.protocols.amp import StartTLS
import twisted

class TestStartTLS(unittest.TestCase):
    def test_starttls_tls_localCertificate(self):
        assert StartTLS(tls_localCertificate=None).certificate is None
```

probably desired as a None certificate will crash here: 

https://github.com/twisted/twisted/blob/720f59dc970d8a32fdf9f97d1dc69855a04646e4/src/twisted/protocols/amp.py#L2496

graingert: ah woops it was in the commit message:

> There is a subtle behavioral change: the docstring states that
passing None for tls_localCertificate is the same as not passing
anything, but the implementation didn't handle None. Now it does.

graingert (approved):

PR #1457, opened 2020-10-21
#9553 Fix IOCP FileHandler loseConnection to remove the writer.

### LINKED ISSUE #5300: IOCP.loseConnection has "stopWriting" instead of "stopWriting()"
|[itamarst's avatar](https://github.com/itamarst)| @itamarst reported|
|-|-|
|Trac ID|trac#5300|
|Type|defect|
|Created|2011-10-09 20:24:32Z|

`FileDescriptor.loseConnection` in IOCP does `self.stopWriting` instead of `self.stopWriting()` as it should. The fix is trivial, but this also needs a test.

Jerub claims to have seen errors caused by this, so assigning to him.

```
trac-id__5300 5300
type__defect defect
reporter__itamar itamar
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__ 
time__1318191872000000 1318191872000000
changetime__1318191872000000 1318191872000000
version__None None
owner__jerub jerub

```
</details>

### LINKED ISSUE #9553: iocpreactor.FileHandle.loseConnection doesn't call stopWriting, but intends to
|[wiml's avatar](https://github.com/wiml)| @wiml reported|
|-|-|
|Trac ID|trac#9553|
|Type|defect|
|Created|2018-10-16 06:38:52Z|

Line 362 is "`self.stopWriting`" (sic, without any parentheses). I assume it was intended to actually call `self.stopWriting()` here.

```
trac-id__9553 9553
type__defect defect
reporter__wiml wiml
priority__normal normal
milestone__None None
branch__None None
branch_author__adiroiban adiroiban
status__new new
resolution__None None
component__core core
keywords__None None
time__1539671932461358 1539671932461358
changetime__1615820059526502 1615820059526502
version__None None
owner__adiroiban adiroiban

```
</details>

adiroiban:
Fixes #5300 
Fixes #9553 

I start working on this by first trying to see a test failing with the current implementation.

For the base twisted.internet.abstrat.FileDescriptor calling stopWriting in loseConnection is redundand.
_writeDisconnected is set to True only inside doWrite and in doWrite before setting _writeDisconnected , stopWriting is called.

I have also ran all the tests and in loseConnection stopWriting is never called when the reactor has a writer.

I have applied the following diff and executed all the tests and the breakpoint was never hit.. this means that stopWriting was always called with an empty reactor.

```diff
diff --git a/src/twisted/internet/abstract.py b/src/twisted/internet/abstract.py
index 8bf6a10e1..1b47e3e0e 100644
--- a/src/twisted/internet/abstract.py
+++ b/src/twisted/internet/abstract.py
@@ -403,6 +403,8 @@ class FileDescriptor(_ConsumerMixin, _LogOwner):
 
         if self.connected and not self.disconnecting:
             if self._writeDisconnected:
+                if self.reactor.getWriters():
+                    import pdb; import sys; sys.stdout = sys.__stdout__; pdb.set_trace()
                 # doWrite won't trigger the connection close anymore
                 self.stopReading()
                 self.stopWriting()
```

I was not able to reproduce the error with a test... so it might be that the stopWriting is not needed ... and this is why nobody complained about this bug before.

I have executed the test on IOCP with a broken implementation and the test is ok 
https://dev.azure.com/twistedmatrix/twisted/_build/results?buildId=3055&view=logs&j=ad8802da-9216-58da-0ca0-f3ed2f645295&t=ee0e3de9-91cc-5adc-e679-e484e1ce2c5a&l=2266

-------

My suggestion is to merge this PR as it is, with some updated tests and a blind fix for IOCP.

Also for the other reactors the stopWriting from loseConnection  does nothing...at least inside the current test suite.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9553


src/twisted/internet/iocpreactor/abstract.py
```diff
@@ -329,6 +329,8 @@ def loseConnection(self, _connDone=failure.Failure(main.CONNECTION_DONE)):
         """
         Close the connection at the next available opportunity.
 
+        This is the same code from twisted.internet.abstract.FileDescriptor.loseConnection
```
glyph: What is the point of this element of the docstring? What do you expect a user to do with it?  Should it be a maintenance note (in a comment) mentioning that this should be deduplicated?

src/twisted/internet/iocpreactor/abstract.py
```diff
@@ -344,7 +346,7 @@ def loseConnection(self, _connDone=failure.Failure(main.CONNECTION_DONE)):
             if self._writeDisconnected:
                 # doWrite won't trigger the connection close anymore
                 self.stopReading()
-                self.stopWriting
+                self.stopWriting()
```
glyph, line 347: 🤦

src/twisted/internet/test/connectionmixins.py
```diff
@@ -127,29 +127,28 @@ def buildProtocol(self, addr):
...
-def runProtocolsWithReactor(
-    reactorBuilder, serverProtocol, clientProtocol, endpointCreator
+def runProtocolsWithReactorInstance(
+    reactor, reactorBuilder, serverProtocol, clientProtocol, endpointCreator
 ):
     """
-    Connect two protocols using endpoints and a new reactor instance.
+    Connect two protocols using endpoints and an existing reactor instance.
 
     A new reactor will be created and run, with the client and server protocol
```
glyph: This docstring seems to be wrong now; no new reactor is created in this function.

src/twisted/internet/test/test_iocp.py
```diff
@@ -210,3 +226,39 @@ def cb(self, rc, bytes, evt):
...
         ir.doIteration(0)
         self.assertEqual(fd.counter, EVENTS_PER_LOOP + 1)
+
+    def test_doubleHalfClose(self):
+        """
+        If one side half-closes its connection, and then the other side of the
+        connection calls C{loseWriteConnection}, and then C{loseConnection} in
+        {writeConnectionLost}, the connection is closed correctly.
+
+        This covers a code branch for ticket #9553.
```
glyph: Can you make this an actual hyperlink (epytext syntax `U{https://...}` ?  Eventually we're going to have to migrate bugtrackers, and it'll be nice to just have hyperlinks that the server can redirect rather than have to preserve the numbering scheme.

src/twisted/internet/test/test_iocp.py
```diff
@@ -210,3 +226,39 @@ def cb(self, rc, bytes, evt):
...
+
+        # If test fails, reactor won't stop and we'll hit timeout:
+        server = ListenerProtocol()
+        client = Client()
+        reactor = self.buildReactor()
+        reactor._handleSignals()
+        # We might have the waker readers.
+        initial_handles = reactor.handles.copy()
+
+        runProtocolsWithReactorInstance(reactor, self, server, client, TCPCreator())
```
glyph: Rather than writing a whole new duplicated function which requires a matching reactor/reactorBuilder, it looks like what you want here is the ability to customize the reactor after it's built and before it's run.  Could you modify `runProtocolsWithReactor` to optionally take a callable of a function to run with the built reactor?  I feel that in addition to avoiding a ton of docstring duplication, having a callback would also make it clear what the "customize the reactor" bits are in the given callback and make it easier to see if there are other ways to unify those too.

src/twisted/newsfragments/9553.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet.iocpreactor.FileHandle.loseConnection now calls stopWriting.
```
glyph, line 1: Can you phrase this in more behavior-like terms rather than just "it calls this method"?  How would it misbehave before, how does it behave now?

glyph (commented):
I'm just a hair shy of "approve" here, if you could address the duplicated signature on the test callable, this looks like a great fix. Thank you!

src/twisted/internet/iocpreactor/abstract.py
adiroiban: True. I will move it to a comment. Thanks!

src/twisted/internet/test/connectionmixins.py
adiroiban: Yep. I will revert this and add a setup callback.

src/twisted/internet/test/test_iocp.py
adiroiban: I have updated to a link.

src/twisted/newsfragments/9553.bugfix
adiroiban: I don't know what was now working before :( ... 

The modified code from this PR is not triggered by any test.

I was trying to find a bug and see a test fail... but I was not able to do that.

Rather than discard all the new tests ... I though that they might be useful as they extend a bit the coverage...so I have created this PR.

I can also turn it into a .misc if you think that would be less confusing.

adiroiban (commented):
Thanks for the review. I tried to address the comments raised during the review.

adiroiban:
This is ready for a new review.

src/twisted/internet/test/test_iocp.py
```diff
@@ -210,3 +226,44 @@ def cb(self, rc, bytes, evt):
...
+            reactor._handleSignals()
+            # We might have the waker readers.
+            initial_handles.extend(reactor.handles.copy())
+
+        reactor = runProtocolsWithReactor(
+            self, server, client, TCPCreator(), reactorSetUp
+        )
+
+        # Check that the reactor is clean.
+        self.assertCountEqual(initial_handles, reactor.handles)
```
graingert, line 269: image

still not covered?

src/twisted/internet/test/connectionmixins.py
```diff
@@ -147,9 +151,12 @@ def runProtocolsWithReactor(
 
     @param endpointCreator: An instance of L{EndpointCreator}.
 
+    @param reactorSetUp: Callable called with the reactor as argument, after the reactor is created.
```
graingert, line 154: I think it's cleaner to pass a subclass of reactorBuilder with a custom reactorFactory

src/twisted/internet/test/test_tcp.py
```diff
@@ -2054,6 +2054,101 @@ def connectionMade(self):
...
+            def writeConnectionLost(self):
+                self.transport.loseConnection()
+
+        class Client(ConnectableProtocol):
+            def connectionMade(self):
+                self.transport.loseConnection()
+
+        initial_readers = []
+
+        def reactorSetUp(reactor):
```
graingert, line 2088: runProtocolsWithReactor already provides a hook to customize the reactor:

```python
def reactorSetUp(self, reactor):
    pass

def buildReactor(self, *args, **kwargs):
    reactor = super().buildReactor()
    self.reactorSetUp(reactor)
    return reactor

def test_doubleHalfCloseReactorCleanup(self):
    ...
    def reactorSetUp(reactor):
        ...

    self.reactorSetUp = reactorSetUp
```

graingert: or a delegate proxy:

```python
@attr.s(frozen=True, slots=True, auto_attribs=True)
class SetupReactorBuilder:
    _reactorBuilder: ReactorBuilder
    _setupReactor: Callable[[IReactorBase, None]

    def buildReactor(self, *args, **kwargs):
        reactor = self._reactorBuilder.buildReactor(*args, **kwargs)
        self._setupReactor(reactor)
        return reactor

    def runReactor(self, *args, **kwargs):
        return self._reactorBuilder.runReactor(*args, **kwargs)

class TestFoo:
    def testFoo(self):
        def reactorSetUp(reactor: IReactorBase) -> None:
            ...

        reactor = runProtocolsWithReactor(
            SetupReactorBuilder(self, reactorSetUp), server, client, TCPCreator()
        )
```

adiroiban:
This was just an attempt to save some tests that I wrote while trying to find the test case that triggers the code with the bug.

I don't plan to spend more time on refactring this PR.

If merging the code from this PR will put the trunk branch in a worst state, I think it's better to just close this and don't merge it.

Cheers

src/twisted/internet/test/test_iocp.py
adiroiban: Looks like windows coverage is not reported.

I have to inspect Codecov, but on coveralls I only see the posix builds... 

image

iocp is zero

image

-------

I will have to check if coveralls has some merge feature... or we can normalize the paths before pushing the reports.

I know that codecov can do that for you

adiroiban: same on codecov .... but paths are normalized
image

azure-pipelines/run_test_steps.yml
```diff
@@ -56,5 +56,5 @@ steps:
     env:
       TWISTED_REACTOR: ${{ parameters.windowsReactor }}
 
-- script: 'python -m tox -e coverage-prepare,coveralls-push'
+- script: 'python -m tox -e coverage-prepare,codecov-push,coveralls-push'
```
glyph, line 59: In the future please put changes like this into a separate PR so we can merge them independently.

src/twisted/internet/iocpreactor/abstract.py
glyph: Hmm. Codecov is still complaining about this. Possibly the codecov-push thing *needs* to be landed on trunk first?

src/twisted/internet/test/connectionmixins.py
glyph: Hmm, good point, `reactorBuilder` effectively already has this functionality?

src/twisted/internet/test/test_iocp.py
```diff
@@ -33,6 +45,7 @@
     if sys.platform == "win32":
         raise
 
+    IOCPReactor = None  # type: ignore[assignment, misc]
```
glyph, line 48: should this be under a `if not TYPE_CHECKING` so mypy won't complain?

src/twisted/internet/test/test_iocp.py
```diff
@@ -163,7 +176,10 @@ def test_ipv6AcceptAddress(self):
         self._acceptAddressTest(AF_INET6, "::1")
 
 
-class IOCPReactorTests(TestCase):
+class IOCPReactorTests(ReactorBuilder, TestCase):
+
+    reactorFactory = IOCPReactor  # type: ignore[assignment]
```
glyph, line 181: Huh, just curious - what's the error here?

src/twisted/internet/test/test_tcp.py
glyph: Thanks for pointing this out, @graingert !  I should have looked it up before giving my feedback.

glyph (commented):
I think I'd be basically ready to stamp this but I'd really like to see the coverage issue addressed first.

glyph:
(I'd be happy to punt on the reactorBuilder fix, I think the state of things in the branch as it is are fine)

adiroiban:
> I think I'd be basically ready to stamp this but I'd really like to see the coverage issue addressed first.

@glyph coverage is not reported for Windows and MacOS 

As part of this PR https://github.com/twisted/twisted/pull/1545 I have assigned custom names to each CI build coverage and I can see that Windows and macOS are missing.

I am planing to fix Windows and MacOS coverage in a separate PR.

adiroiban:
This needs-changes :

* Add/fix type annotation
* Refactor setupReactor
* Fix tests
* Fix code coverage

PR #1458, opened 2020-10-21
create a github action to release the wheel and sdist to pypi 

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9531


.github/workflows/test.yaml
```diff
@@ -36,14 +38,17 @@ jobs:
...
+      uses: actions/cache@v2
       with:
         path: ${{ steps.pip-cache.outputs.dir }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        key:
+          ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml', 'setup.py',
+          'setup.cfg') }}
         restore-keys: |
-            ${{ runner.os }}-pip-
+          ${{ runner.os }}-pip-
```
```suggestion
            ${{ runner.os }}-pip-
```

.github/workflows/test.yaml
```diff
@@ -77,14 +82,17 @@ jobs:
...
```
```suggestion
            ${{ runner.os }}-pip-
```

.github/workflows/docs.yaml
```diff
@@ -21,14 +21,17 @@ jobs:
...
+      uses: actions/cache@v2
       with:
         path: ${{ steps.pip-cache.outputs.dir }}
-        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        key:
+          ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml', 'setup.py',
+          'setup.cfg') }}
         restore-keys: |
-            ${{ runner.os }}-pip-
+          ${{ runner.os }}-pip-
```
graingert:
```suggestion
            ${{ runner.os }}-pip-
```

.github/workflows/test.yaml
```diff
@@ -3,14 +3,16 @@ name: Run tests
...
+      - twisted-*
   pull_request:
     branches: [ trunk ]
 
 jobs:
   testing:
     runs-on: ubuntu-20.04
     env:
-      TOXENV: "alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push"
+      TOXENV: "alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push,release"
```
adiroiban, line 15: test.yaml is a very generic name...

and this is only testing on Ubuntu 20.04 .... we don't test on Windows and macOS

Do we need to run the tests here?

I was thinking that we can have a simple wheel + sdist publishing workflow which will run a single job which will call ciwheelbuilder and then publish

graingert:
We don't need cibuildwheel because it's a pure python wheel

pyproject.toml
```diff
@@ -74,3 +74,205 @@ multi_line_output = 3
...
+    ; Code quality checkers
+    manifest-checker: check-manifest>=0.42
+
+    twine: twine
+
+    lint: pyflakes
+
+    black: black==20.8b1
+
+    wheels: cibuildwheel==1.6.1
```
adiroiban: cibuildwheel is here :)

I think is best to keep the tox.ini in the old location and have a separate pr to clean it

adiroiban:
As part of this PR we should also check the current release documentation and update it to cover the new process 

The release documentation is at docs/core/development/policy/release-process.rst

I would prefer to have that documentation removed from the docs and moved into CONTRIBUTING or somewhere outside of the public documentation

adiroiban:
Or instead of having the wheel documentation in docs/core/development/policy/release-process.rst we can have it as a comment inside the GitHub Actions wheel publisher workflow.

I think that Twisted tests are too complicated to run all the tests before publishing the wheels

So we should just have all the tests executed (and manually re-run)...then when someone things that the branch is green
will push a tag right away.

I guess that we can also re-run the wheel publish action that failed due to a flaky tests...but I think that for a start is better to have a simple release process ... my 2 cents

graingert:
> Or instead of having the wheel documentation in docs/core/development/policy/release-process.rst we can have it as a comment inside the GitHub Actions wheel publisher workflow.

Done

> I think that Twisted tests are too complicated to run all the tests before publishing the wheels
> 
> So we should just have all the tests executed (and manually re-run)...then when someone things that the branch is green
> will push a tag right away.
> 
> I guess that we can also re-run the wheel publish action that failed due to a flaky tests...but I think that for a start is better to have a simple release process ... my 2 cents

I think it's important that we run the tests against *the actual* artefact that gets released

.github/workflows/test.yaml
```diff
@@ -105,3 +113,33 @@ jobs:
...
+
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-20.04
+    needs: [testing]
+    steps:
+      - name: Download dists for PyPI
+        uses: actions/download-artifact@v2
+        with:
+          name: py3.9_dist
```
adiroiban, line 132: is this ok?

.github/workflows/test.yaml
```diff
@@ -105,3 +113,33 @@ jobs:
...
         tox -q
+
+    - name: upload dist
+      uses: actions/upload-artifact@v2
+      with:
+        name: py${{ matrix.python-version }}_dist
+        path: dist
+
+
+  all-successful:
```
adiroiban: can we have a more descriptive name for this.
Maybe 'pypi_publish' ?

tox.ini
```diff
@@ -85,8 +85,6 @@ deps =
 
     black: black==20.8b1
 
-    wheels: cibuildwheel==1.6.1
```
adiroiban, line 88: maybe also remove `twine`

tox.ini
```diff
@@ -159,10 +152,6 @@ commands =
 
     twine: twine check {distdir}/*.*
```
adiroiban, line 145: maybe also remove twine ...as it is no longer needed

.github/workflows/test.yaml
graingert: >  Windows and macOS

I think the plan is that we're going to run this workflow on MacOS and Windows soon, and remove azure pipelines

.github/workflows/test.yaml
graingert:
```suggestion
  pypi-publish:
```

docs/core/development/policy/release-process.rst
```diff
@@ -184,31 +184,12 @@ Prepare the branch
 #. Commit the version and ``README.rst`` changes.
 #. Submit the ticket for review
 #. Pause until the ticket is reviewed and accepted.
-#. Tag the release.
+#. Tag the release. Github Actions will upload the dist to PyPI
```
adiroiban, line 187: Maybe instead of `Tag the release' we should suggest using the GitHub Release UI and add the changelog there
and create a tag via the GItHub UI at https://github.com/twisted/twisted/releases/new
This is to replace the removed  ' Cut the tarballs & installers '

And as a separate paragraph inform that Twisted package is distribute only via PyPI.... to replace the removed `Distribute` section.

The `How to do a release candidate` section should also be updated.


-------
```suggestion
#. Use the `GitHub Create Release UI <https://github.com/twisted/twisted/releases/new>`_ the make a new release and tag.
#. Create a tag using the format `twisted-VERSION` based on the latest commt on the release branch that was approved after the review.
#. Use `Twisted VERSION` as the name of the release.
#. Add the release notes to GitHub Release page.
#. Make sure 'This is a pre-release` is not checked.
#. Github Actions will upload the dist to PyPI when a new tag is pushed to the repo. PyPI is the only canonical source for Twisted packages.
```

docs/core/development/policy/release-process.rst
```diff
@@ -184,31 +184,12 @@ Prepare the branch
 #. Commit the version and ``README.rst`` changes.
```
adiroiban, line 184: I think the previous info about the Quotes should be removed.

Also, I think that instead of reverting the pre-release changes, we can just manually update the release date and version inside the NEWS file

adiroiban (changes requested):
Thanks for the update.

Ii think that as part of this PR, beside creating the GitHub action workflow we should also update the documentation and explain how we plan to use it.

There were some docs updated...but i think they are not complete.

We should also take into consideration the pre-release process.

The process so far, was that before the final release, a pre-release was created.
People could tests and send feedback.

But maybe we can redefine the release process and remove the need for a pre-release.

Since it is not easy to do a release...if a problem is found, we will just do a new bugfix release.

What do you say?

graingert:
> But maybe we can redefine the release process and remove the need for a pre-release.

I'd rather have more pre-releases rather than fewer

graingert:
> Ii think that as part of this PR, beside creating the GitHub action workflow we should also update the documentation and explain how we plan to use it.

I just want to get y'all to the extent of: push a tag get a pypi release. And in exchange I'd like an a0 of what's in trunk.

I think we can defer the decisions on changing the release process because, with an a0 there's no abdication of doing all the extra stuff like mailing list pushes, blog posts, signatures and squashing/production of changelogs from NEWS files.

I'd be happy to add a line saying: "for .aXXX releases we change the version and push a tag, for .rcXXX and final releases we follow the following steps:"

tox.ini
```diff
@@ -185,3 +174,16 @@ commands =
         --cache-dir="{toxworkdir}/mypy_cache"  \
         {tty:--pretty:}                        \
         {posargs:src}
+
```
adiroiban, line 164: Can you please add a comment explaining the purpose / goal of this tox env?

I am not sure why we need this.

It looks like it creates an sdist based on PEP517 and together with the wheel created tox it will validate the files.

Will a wheel be created if we just run release, or it depends on a previous build?

.github/workflows/test.yaml
adiroiban: OK... but do we need to create the sdist and test using twine for each job from the matrix?

Can we move `tox -e release` as a step in the `pypi-publish` job?

There are a lot of tox environments executed in the matrix...if possible, I would prefer to have all the wheel/sdist related task in the dedicated pypi publish job. Does it make sense?

.github/workflows/test.yaml
```diff
@@ -55,6 +60,12 @@ jobs:
         python --version
         tox -q
 
+    - name: upload dist
```
adiroiban, line 63: Can we skip this upload + download part and instead create the sdist + wheels in the pypi-publish job?

I don't see what is the difference between creating the wheels via tox or ciwheelbuilder.

My expectation is that we will get the exactly same wheel... and since we use tox-wheel to run the tests we will be testing the same wheel.

.github/workflows/test.yaml
adiroiban: Why was this marked as resolved? Just asking.

But I hope this can be implemented without upload + download of artifacts.

adiroiban (commented):
Thanks for the updates.

I left a set of questions... we can follow up over IRC.

But is ok to focus on just the release.

I think that we can do an alpha release directly from a PR.
Why do we need an alpha from trunk?

Also, there is a PYPI testing token, so we can release on the pypi staging site, just for test.

I don't think that we can release from this PR, as this PR is from a clone... but you should have permissions to create a branch inside Twisted.

.github/workflows/test.yaml
graingert: > ciwheelbuilder

you mean cibuildwheel? that's for setting up specific docker containers with specific libraries for manylinux1 for statically linking against. We're just using plain old `pip wheel`

.github/workflows/test.yaml
graingert: ah I thought you were pointing out a bug - I was uploading artefacts in the wrong build stage

docs/core/development/policy/release-process.rst
graingert: sure, I don't want to change the release process here around changelogs/RC/final in this PR

.github/workflows/test.yaml
adiroiban: ah.... true :)

tox.ini
graingert: > Can you please add a comment explaining the purpose / goal of this tox env?
> 
> I am not sure why we need this.

it copies the wheel from the tox dist directory to the "./dist" directory for the upload

> It looks like it creates an sdist based on PEP517 and together with the wheel created tox it will validate the files.
> 
> Will a wheel be created if we just run release, or it depends on a previous build?

yes a wheel will be created and checked if you run `tox -e release`

adiroiban:
The code was moved to a twisted repo own branch to access the secrets.
This can be closed.

PR #1461, opened 2020-10-22
lint (black/pyflakes) everything that looks like python code using pre-commit

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10032


graingert:
I was somewhat conflicted on fixing the historic/2003 python docs - but the changes made to make them syntactically valid under python3 are so minimal that I think it's worth doing.

I'd rather delete them than shade them from the light of black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+        args: ["--target-version", "py35"]
```
wsanchez: Target version should be in `pyproject.toml`, and ideally not also in other places.

graingert:
```suggestion
```

woops this line is redundant

.github/workflows/lint.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     env:
-      TOXENV: "lint,black,twine"
+      TOXENV: "lint,twine"
```
adiroiban, line 13: do we still need twine here?

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -7,7 +7,7 @@ Please note that the trac ticket, news fragment, and review submission portions
 ## Contributor Checklist:
 
 * [ ] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/<!-- Create a new one at https://twistedmatrix.com/trac/newticket and replace this comment with the ticket number. -->
-* [ ] I ran `tox -e black-reformat` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
+* [ ] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
```
adiroiban, line 10: an we ask people to run pre-commit  rather than lint ? or have a `tox -e precommit`

I think that we can add precomit to the dev dependency lists in setup.cfg

I think that all the black reformating should be done in a single commit and that commit added to .git-blame-ignore-revs

I would keep `tox -e lint` for CI usage ... but local devs should run pre-commit via pipx or tox or custom virtualenv

README.rst
```diff
@@ -74,17 +74,12 @@ Static Code Checkers
...
 
-To automatically format code according to the coding standards::
+Or, for speed, use pre-commit directly::
 
-  $ tox -e black-reformat
-
-Or, for speed, specify exactly what to format::
-
-  $ tox -e black-reformat -- path/to/file.py
+  $ pipx run pre-commit run
```
adiroiban, line 82: can we add a tox pre-commit env as the "standard" tool... pipx is great, but I think that is best to keep all in tox.

People who will want to run pipx or just custom virtualenv could then use the rules from tox to see how to run-precommit

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.3
+    hooks:
+      - id: flake8
```
adiroiban, line 10: flake8 runs only pyflakes by default ?

adiroiban (changes requested):
Wow! Great stuff :)

Happy to see this and I think that this PR is on the right path

A have a few major comments before approving this.

See inline comments.

Thanks!

adiroiban:
> I'd rather delete them than shade them from the light of black

I think we can keep them...since they are historic. ... but is ok to update then with black.
If we remove, them I think that should be done in a separate PR.

Thanks!

.pre-commit-config.yaml
graingert: yes, flake8 runs pycodestyle and pyflakes, Ive ignored all the pycodestyle violations that currently stand, and double checked the pyflakes violations are not ignored

README.rst
graingert: `tox -e lint -- run` does that already

.github/PULL_REQUEST_TEMPLATE
graingert: I'm not sure what you're suggesting here

.github/workflows/lint.yaml
graingert: Yep. twine runs other checks on the built wheel and doesn't really fit with the pre-commit model

.github/PULL_REQUEST_TEMPLATE
adiroiban: sorry. my bad. This is ok.

.github/workflows/lint.yaml
adiroiban: But I guess this will be removed in the followup PR with the new wheel+sdist build workflow... or am O lost ?  ... again :)

setup.cfg
```diff
@@ -116,3 +116,76 @@ console_scripts =
     *.pxi
     *.pyx
     build.bat
+
+
+[flake8]
+disable-noqa = True
+max-line-length = 182
```
adiroiban, line 123: do we need this? Should black take care of long lines?

My undertanding is that we have stopped talking about style as black is taking care of that.

.github/workflows/lint.yaml
graingert: I'm keeping twine still in the release PR too

README.rst
adiroiban: Them maybe suggest

```suggestion
  $ tox -e lint -- run
```

I would prefer `tox -e lint` to only run on changes since trunk.  and have `tox -e lint --all-files` to be used by CI.

admin/fix-for-src-mv.py
```diff
@@ -4,7 +4,7 @@
 Twisted moved the C{twisted} hierarchy to the C{src} hierarchy, but C{git}
 doesn't know how to track moves of directories, only files.  Therefore any
 files added in branches after this move will be added into ./twisted/ and need
-to be moved over into 
+to be moved over into
```
adiroiban:
```suggestion
to be moved over into.
```

adiroiban (approved):
Thanks. Only minor comments.

This is ready for merge.

README.rst
graingert: running via `tox` instead of `pipx` adds quite a bit of slowness - `pipx` here is more of a demo on how it could be installed.

setup.cfg
graingert: black doesn't make changes that can't be verified by checking that the AST didn't change. there's ways to manually split a 1024+ column line to split into 88 cols that black simply won't ever be able to do

wsanchez (approved):

rodrigc:
@graingert You removed the `tox -e black` target, but you didn't update the places on the wiki where this target is mentioned: https://twistedmatrix.com/trac/wiki/TwistedDevelopment

Do you plan to update the wiki?

@adiroiban / @wsanchez When reviewing this type of stuff, please double check that any documentation changes on the wiki are updated as well.

adiroiban:
Thanks for the followup. Well, we do our best with the review

I think that anyone can update the Wiki without the need for a review... so feel free to update the wiki page.

-------

I would like to stop using the wiki and instead use the repo's docs for documentation.

In this way, everything can be done in a single PR.

We now have up to date docs based on the trunk branch at 

https://docs.twistedmatrix.com/en/latest/core/development/index.html

graingert:
> I would like to stop using the wiki and instead use the repo's docs for documentation.

I tend to agree, in this case I think we should delete the outdated wiki version of these docs and link to the updated version on https://docs.twistedmatrix.com/

rodrigc:
@graingert for now, can you update the wiki as I have asked?    While not perfect, the wiki has been around for a long time,
and for better or worse, many people in the project, and outside the project still refer to it.

The Twisted Development process is a very important part of the Twisted project.  If this is deleted from the wiki
in favor of documentation inside the tree, then this needs to be done via a ticket and PR (just like anything else in Twisted),
so that there is an audit trail that this has been decided and agreed upon.

PR #1466, opened 2020-10-23
[10035] Build apidocs inside Read the docs environment.

adiroiban:
## Scope

The result is available here https://twisted--1466.org.readthedocs.build/en/1466/

To simplify the release process we want to host the API docs inside read the docs.

I find that the Read the Docs theme is easier to read than the Twisted Trac theme for Sphinx.

This provides support for creating API docs via both the old `bin/admin/build-apidocs` and the Read The Docs build.
In this way we can transition with the next release

## Followup work

* Update the CI workflow to fail on apidocs + narrative docs errors - https://twistedmatrix.com/trac/ticket/10066
* Remove usage of apilinks and replace with intersphinx. -  https://twistedmatrix.com/trac/ticket/10057
* Update Twisted own hosted docs to redirect to read the docs for latest version - In Braid

## How to test

Make sure the old script continues to work `tox -e apidocs`

Check that Sphinx docs now has API docs included in the index this is the before last.

Also check that source links for API work.

```
tox  -e apidocs
firefox docs/_build/index.html
```

Check read the docs

Main https://twisted--1466.org.readthedocs.build/en/1466/ 
API https://twisted--1466.org.readthedocs.build/en/1466/api/index.html

Apilinks https://twisted--1466.org.readthedocs.build/en/1466/core/howto/servers.html  link at "Your protocol handling class will usually subclass twisted.internet.protocol.Protocol. "

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10035


adiroiban:
This is ready for review... the deprecated release script is still failing, but that might be a bug in pydoctor.
Those tests were enabled for a very long time.

The result is available here https://twisted--1466.org.readthedocs.build/en/1466/

setup.cfg
```diff
@@ -38,12 +38,25 @@ package_dir = =src
...
+    sphinx~=3.2
+    sphinx-rtd-theme~=0.5.0
+    pillow~=7.2
+    readthedocs-sphinx-ext~=2.1
+    recommonmark~=0.6.0
+
 dev =
+    %(test)s
+    %(docs)s
+    %(tls)s
```
graingert: I think the `tls =` lines should be above the `%(tls)s` line

src/twisted/python/_release.py
```diff
@@ -502,12 +503,26 @@ def main(self, args):
...
         """
         if len(args) != 2:
             sys.exit(
                 "Must specify two arguments: " "Twisted checkout and destination path"
             )
-        self.buildAPIDocs(FilePath(args[0]), FilePath(args[1]))
+
+        original_stdout = sys.stdout
+        try:
+            capture = StringIO()
```
graingert: https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout

src/twisted/python/test/test_release.py
```diff
@@ -19,12 +19,14 @@
...
 
 from io import BytesIO, StringIO
+from unittest import skipIf
 
 from twisted.trial.unittest import TestCase, FailTest, SkipTest
 
 from twisted.python.procutils import which
 from twisted.python import release
 from twisted.python.filepath import FilePath
+from twisted.python.reflect import requireModule
```
graingert: I think this can be `importlib.import_module` now. But we can save that for another pr

src/twisted/python/test/test_release.py
```diff
@@ -453,7 +449,7 @@ def test_build(self):
         )
         self.assertIn(
             '<a href="{}">{}</a>'.format(projectURL, projectName),
-            indexPath.getContent(),
+            indexPath.getContent().decode("ascii"),
```
graingert: Do these need to be "ascii"?
```suggestion
            indexPath.getContent().decode(),
```

adiroiban: well.. I just want to convert from bytes to str... and I want to go with a fixed encoding which will never fail.

For example utf-16 decoding will fail for an odd number of bytes.

As a best practice, I never used the default encoding.

Let me know if there is a better way of converting from bytes to ascii str.

setup.cfg
adiroiban: I tried  all this and they work... I don't think that the order . I put them in dev as I think that developers wants them 99% of the time.

```
pip install -e .[tls]
pip install -e .[dev]
pip install -e .[docs]
```

src/twisted/python/_release.py
adiroiban: True.. I will use it. I am not familiar with latest stdlib :) Nice to have this helper in stdlib.

src/twisted/python/test/test_release.py
adiroiban: I don't see  import_module used anywhere in Twisted.... so yes. maybe update all  requireModule calls in separate PR

tox.ini
```diff
@@ -156,8 +150,40 @@ commands =
...
+#
+[testenv:narrativedocs]
+extras = docs
+setenv =
+    # Set this to `True` to run similar to Read the docs.
+    READTHEDOCS=False
+    READTHEDOCS_LANGUAGE=en
+    READTHEDOCS_VERSION=1.0.0.dev0
+commands = sphinx-build -aW -b html -d {toxinidir}/docs/_build {toxinidir}/docs {toxinidir}/docs/_build/
+#{toxinidir}/bin/admin/build-docs {toxinidir}
```
adiroiban: this is here as a reminder for reviewers to see if we will still need bin/admin/build-docs or it can be removed.... or it should be updated so that tox will call bin/admin/build-docs instead of the custom sphinx-build

setup.cfg
graingert: oh I just think it's less confusing to have the ConfigParser keys in topological order

src/twisted/python/test/test_release.py
graingert: the recommended way from pyupgrade is to just rely on the default - as it will always be `utf8`

https://github.com/asottile/pyupgrade#encodeutf-8

graingert:
related, https://github.com/twisted-infra/braid/pull/264

> There will be a second PR to redirect twistedmatrix.com/documents/current to the new domain, but first we should have this pushed to PROD and make sure it works and later we can push the redirection.

@adiroiban when we do the redirect there will be no apidocs for any of the older builds

adiroiban:
My plan is to only redirect "current" URL fragment. Older version docs will continue to be on the same server without redirecting.

Later we can see how we can push to RTD tags to rebuild the docs for very old versions.

We already have some older version on RTD and the have absolute URL to apidocs.

src/twisted/python/test/test_release.py
adiroiban: Thanks. True stdlib docs - https://docs.python.org/3/library/stdtypes.html#bytes.decode

I am still a python2 developer and I am not used to burn py2 bridges :) but I will update this. Happy to see a fixed default in py3.

]

src/twisted/python/_release.py
```diff
@@ -501,12 +503,22 @@ def main(self, args):
...
         """
         if len(args) != 2:
             sys.exit(
                 "Must specify two arguments: " "Twisted checkout and destination path"
             )
-        self.buildAPIDocs(FilePath(args[0]), FilePath(args[1]))
+
+        capture = StringIO()
+        with redirect_stdout(capture):
+            self.buildAPIDocs(FilePath(args[0]), FilePath(args[1]))
```
mthuurne: I think the best long-term solution is to use a logging API in pydoctor, instead of writing directly to stdout. However, that is not something that needs to happen right now and the stdout capture will still work when pydoctor does make the switch.

adiroiban:
This is ready for review. See PR description.

.readthedocs.yml
```diff
@@ -1,7 +1,11 @@
 version: 2
 
 sphinx:
-  fail_on_warning: true
+  # This should not fail on errors so that we can view the rended Read The Docs.
```
mthuurne:
```suggestion
  # This should not fail on errors so that we can view the rendered Read The Docs
```

tox.ini
```diff
@@ -125,18 +127,36 @@ commands =
...
 
-
-[testenv:apidocs]
+#
+# `apidocs` environment is designed to build only the API doc HTML files.
+#
+# API docs build errors are visible to stdout
+# The resulting files can be viewed with web browsers from the local `apidocs`
+# path.
+[testenv:apidocs    ]
```
mthuurne: Those spaces before `]` are probably harmless, but it looks a bit odd.

mthuurne (approved):
Looks good, only some minor comments (typo + formatting).

I don't know much about Sphinx, so I skimmed over most of the `conf.py` file.

.readthedocs.yml
```diff
@@ -1,7 +1,11 @@
...
+  # pages even on errors.
```
graingert:
```suggestion
  # This should not fail on errors so that we can view the rendered Read The Docs.
  # pages even on errors.
```

.readthedocs.yml
```diff
@@ -1,7 +1,11 @@
...
+  # We have a separate build process which fails on errors and shows all
+  # the errors in one go.
```
graingert, line 7:
```suggestion
  # the errors in one go.
  # we can re-enable this when https://github.com/readthedocs/readthedocs.org/issues/7116
  # is fixed
```

graingert (approved):

docs/conf.py
```diff
@@ -166,23 +167,29 @@
 
 
 # -- Extension configuration ----------------------------------------------
+_git_reference = subprocess.getoutput("git rev-parse --abbrev-ref HEAD")
```
graingert: this is a legacy subprocess call: https://docs.python.org/3/library/subprocess.html#legacy-shell-invocation-functions

```suggestion
_git_reference = subprocess.run(
    ["git", "rev-parse", "--abbrev-ref", "HEAD"],
    text=True,
    encoding="utf8",
    capture_output=True,
    check=True,
).stdout
```

graingert (changes requested):

graingert (approved):

PR #1473, opened 2020-10-28
Fix TwistedSystem (plugin for pydoctor)

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10044


src/twisted/python/test/test_release.py
```diff
@@ -912,15 +877,26 @@ def test_export(self):
...
+try:
+    gitVersion = runCommand(["git", "--version"]).split(b" ")[2].split(b".")
+except FileNotFoundError:
+    gitSkip, gitSkipText = True, "git is not present."
+else:
+    # We want git 2.0 or above.
+    if int(gitVersion[0]) >= 2:
+        gitSkip, gitSkipText = False, ""
+    else:
+        gitSkip, gitSkipText = True, "old git is present"
```
adiroiban: I think that in these days all the supported environment if they have git, will have a git older than v2


And maybe we can also assume that all testing system have git... so we can always run it.

I think that this might be a reason why the tests were skipped on Windows, as getting the git path on Windows might be a little bit more complicated.

mthuurne: One reason it failed on Windows was that `/tmp` was hardcoded as the temp dir. That's easy to fix though.

Windows using "git-bash" (msys) is actually close enough to a POSIX environment for a lot of tasks. But I don't know if our CI runs under bash or `cmd.exe`.

mthuurne: After fixing the temp dir, more issues are popping up. So I disabled the git tests on Windows.

Once the pydoctor and sphinx tests are running in CI, we can see whether those do run correctly on Windows or should be disabled as well.

setup.cfg
```diff
@@ -41,6 +41,8 @@ packages = find:
 test =
     cython-test-exception-raiser ~= 1.0
     PyHamcrest >= 1.9.0
+    pydoctor >= 20.7.0
```
adiroiban, line 52: Thanks for the great start.


The updates to `test` are not quite ok.

In Twisted we are running tests without dependencies to make sure that twisted can handle soft dependencies.

We should not add more deps to this... and the long term plan is to also remove pyhamcrest or skip hamcrest test when pyhamcrest is not installed.

Yes... this did the trick, but I think that we should add a dev_docs extra requirements.

I will try to see if I can make it right  :)

adiroiban:
Test are failing on windows due to the source path URL.
On windows we end up with: https://github.com/twisted/twisted/tree/twisted-1.0.0/src\\twisted

I think that the issue is here:

```
            mod.sourceHref = self.sourcebase + source_path[len(projBaseDir):]
```

https://github.com/twisted/pydoctor/blob/678441a505b4ee8700b29dfd6c264e1d21e294a8/pydoctor/model.py#L572

we should do a string replace on source_path and replace \ with / to normalize the URL

adiroiban (approved):
This looks good.

I can see that test_buildWithDeprecated is now executed and is green

https://github.com/twisted/twisted/pull/1473/checks?check_run_id=1323782792#step:8:7444

Some tests on windows are failing ... and I think that due to pydoctor.

Can we fix pydoctor and do a new release of pydoctor ?

adiroiban:
So... my plan for this PR is the following:

* Merge with skiping all release tests on windows.
* Fix pydoctor source URL on windows and release a new version of pydoctor
* Update trial to skip module on empty string so that we don't need the typing comment
* Have a separate PR in which we use the updated pydoctor and enable the release tests on Windows.

 I see that release tests are executed on Linux https://github.com/twisted/twisted/pull/1473/checks?check_run_id=1323917764#step:8:7473

@mthuurne If you are happy with the plan feel free to merge this.

Thanks for your help! :)

setup.cfg
```diff
@@ -38,25 +38,37 @@ package_dir = =src
...
     sphinx-rtd-theme~=0.5.0
+
+; All the extra tools used to help with the development process.
+dev =
+    %(dev_release)s
+    pyflakes >= 1.0.0
+    twisted-dev-tools >= 0.0.2
+    python-subunit
+    twistedchecker >= 0.7.2
+    mock~=4.0
```
graingert: mock is for readthedocs

adiroiban: Thanks.  I will move it...

But do we need to pin the mock value?

I guess that we only need to pin the high level read the docs and that package will pull its dependencies. Isn't that the case?

Thanks!

adiroiban: Thanks. I tried to simplify the dependencies. Since these are dev and release only tools I think that is ok to be more flexible with the dependencies and don't pin everything.


I see that Read The docs build is successful with relaxed dependencies :)

setup.cfg
```diff
@@ -38,25 +38,30 @@ package_dir = =src
...
-    # force upgrades for rtd default packages: https://git.io/JU73V
-    alabaster~=0.7.12
-    commonmark~=0.9.1
-    docutils~=0.16.0
-    mock~=4.0
-    pillow~=7.2
-    readthedocs-sphinx-ext~=2.1
-    recommonmark~=0.6.0
-    sphinx~=3.2
-    sphinx-rtd-theme~=0.5.0
```
graingert, line 59: so we're going to be using the old deps now? https://github.com/readthedocs/readthedocs.org/blob/046c4bd20fd875807cb3e1840846249160c37079/readthedocs/doc_builder/python_environments.py#L354-L389

src/twisted/python/test/test_release.py
```diff
@@ -47,42 +50,9 @@
 )
 
 if os.name != "posix":
```
graingert: you should be able to get mypy to infer the type with:
```python
skip = (
    None
    if os.name == "posix"
    else "Release toolchain only supported on POSIX."
)
```

also I think `sys.platform != "win32"` is preferred?

https://gitter.im/python-trio/general?at=5da80653c87a1d28acc46108

adiroiban: Thanks. I will apply the changes

src/twisted/python/test/test_release.py
```diff
@@ -912,15 +883,29 @@ def test_export(self):
...
+    try:
+        gitVersion = runCommand(["git", "--version"]).split(b" ")[2].split(b".")
+    except FileNotFoundError:
+        gitSkip, gitSkipText = True, "git is not present."
+    else:
+        # We want git 2.0 or above.
+        if int(gitVersion[0]) >= 2:
+            gitSkip, gitSkipText = False, ""
+        else:
+            gitSkip, gitSkipText = True, "old git is present"
```
graingert: I think this could be more clear

```python
def _gitSkip() -> str:
    if os.name != "posix":
        return "Release toolchain only supported on POSIX."
    try:
        gitVersion = runCommand(["git", "--version"]).split(b" ")[2].split(b".")
    except FileNotFoundError:
        return "git is not present"
    if int(gitVersion[0]) < 2:
        return "git 2.0 or above is required"
    return ""

gitSkipText = _gitSkip()
gitSkip = bool(gitSkipText)
```

adiroiban: I  went with removing the whole git existence check.

I think that any env in which the tests are executed will have git...and will have a newer version of git.

setup.cfg
graingert:
```
readthedocs=
    %(dev_release)s
    # force upgrades for rtd default packages: https://git.io/JU73V
    alabaster~=0.7.12
    commonmark~=0.9.1
    docutils~=0.16.0
    mock~=4.0
    pillow~=7.2
    recommonmark~=0.6.0
```

and then update the `extra` in `./readthedocs.yml`

graingert (approved):

PR #1474, opened 2020-10-29
[10045] Don't skip test on empty string skip attribute.

adiroiban:
## Scope

This adds support in trial to not skip a test when the skip attribute is an empty string.

In previous version it was required to be None to be skipped.

I have created this since mypy is complaining about this code and wants Optional extra type hinting.

```
if non contextvars:
    contextvarsSkip = "contextvars is not available"
else:
    contextvarsSkip = None
```

Later I found out that mypy is happy with this syntax

```
if contextvars:
    contextvarsSkip = None
else:
    contextvarsSkip = "contextvars is not available"
```

When possible we should use the `@skipIf` decorator ...but the skip attribute is still supported and convenient for top level module skip.

I am pushing this since I have already created the implementation and the test.

Feel free to reject it.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10045

### DIFF
--- a/src/twisted/newsfragments/10045.feature
+++ b/src/twisted/newsfragments/10045.feature
@@ -1,2 +1,2 @@
 Trial will now skip a test when the `skip` attribute is an empty string.
-In previous version the test was skipped only when the attribute was None.
+In previous versions the test was not skipped in this case.
--- a/src/twisted/newsfragments/10045.feature
+++ b/src/twisted/newsfragments/10045.feature
@@ -1,0 +1,2 @@
+Trial will now skip a test when the `skip` attribute is an empty string.
+In previous version the test was skipped only when the attribute was None.
--- a/src/twisted/trial/_synctest.py
+++ b/src/twisted/trial/_synctest.py
@@ -980,7 +980,7 @@ class SynchronousTestCase(_Assertions):
         See L{TestCase} docstring for more details.
         """
         skipReason = util.acquireAttribute(self._parents, "skip", None)
-        doSkip = skipReason is not None
+        doSkip = skipReason not in [None, ""]
         if skipReason is None:
             doSkip = getattr(self, "__unittest_skip__", False)
             if doSkip:
--- a/src/twisted/trial/test/skipping.py
+++ b/src/twisted/trial/test/skipping.py
@@ -16,18 +16,51 @@ from twisted.trial.unittest import SynchronousTestCase, TestCase, SkipTest, Fail
 
 
 class SkippingMixin:
+    """
+    Tests that are skipped at the method level.
+
+    Make sure the docstring of each tests will start with the skip reason,
+    as this is use in the test assertion.
+    """
+
+    def test_no_skip(self):
+        """
+        This is a test that is not skipped.
+        here to make sure that the whole class is not skipped.
+        """
+
     def test_skip1(self):
-        raise SkipTest("skip1")
+        """
+        The skip1 reason.
+
+        The test is skipped when SkipTest is raised inside.
+        """
+        raise SkipTest("The skip1 reason.")
 
     def test_skip2(self):
+        """
+        skip2-reason
+
+        If executed, this test will fail, but is skipped via the skip
+        attribute.
+        """
         raise RuntimeError("I should not get raised")
 
-    test_skip2.skip = "skip2"  # type: ignore[attr-defined]
+    test_skip2.skip = "skip2-reason"  # type: ignore[attr-defined]
 
-    def test_skip3(self):
-        self.fail("I should not fail")
+    def test_no_skip_empty_attribute(self):
+        """
+        This test is not skipepd as the skip attribute is the empty string.
+        """
 
-    test_skip3.skip = "skip3"  # type: ignore[attr-defined]
+    test_no_skip_empty_attribute.skip = ""
+
+    def test_no_skip_none_attribute(self):
+        """
+        This test is not skipepd as the skip attribute is None.
+        """
+
+    test_no_skip_empty_attribute.skip = None
 
 
 class SynchronousSkipping(SkippingMixin, SynchronousTestCase):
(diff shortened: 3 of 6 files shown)


src/twisted/newsfragments/10045.feature
```diff
@@ -0,0 +1,2 @@
+Trial will now skip a test when the `skip` attribute is an empty string.
+In previous version the test was skipped only when the attribute was None.
```
exarkun: In previous versions the test was skipped when the attribute was a non-empty string or when it was None, I guess.  So maybe a good second sentence would be:

> In previous versions the test was not skipped in this case.

src/twisted/trial/test/test_skip.py
```diff
@@ -90,3 +90,11 @@ def test_one(self):
...
     def test_shouldNotSkip(self):
         self.assertTrue(True, "Test should run and not be skipped")
+
+    def test_emptyStringNotSkip(self):
+        """
+        A skip attribute with empty string should not skip the tests.
+        """
+        self.assertTrue(True, "Test should run and not be skipped")
+
+    test_emptyStringNotSkip.skip = ""  # type: ignore[attr-defined]
```
exarkun: All of the tests in this module seem to be implemented incorrectly.  What if trial doesn't run any of them?  It won't report a test suite failure - because none of them ran.  But that would clearly be broken skip handling behavior.

The good trial tests elsewhere in the suite define a victim `TestCase`  and then *another* `TestCase` that manually runs the test discovery/execution process on the victim and asserts the correct results.

This is sort of a pre-existing defect in the `test_skip.py` suite but this new test could at least be written correctly.

adiroiban:
Thanks @exarkun for the review.

I have updated the release note and updated the tests to count for skipped methods.

I have renamed the test_skip.py to test_skipif.py and updated the docstring to inform that the tests are only for `@skipIf` decorator and that we have skipping.py for other skip tests.

Please give it another try and let me know if you are happy to see this merged.

Thanks!

src/twisted/trial/_synctest.py
```diff
@@ -980,7 +980,7 @@ def getSkip(self) -> Tuple[bool, Optional[str]]:
         See L{TestCase} docstring for more details.
         """
         skipReason = util.acquireAttribute(self._parents, "skip", None)
-        doSkip = skipReason is not None
+        doSkip = skipReason not in [None, ""]
```
graingert, line 983: `__contains__` runs `__eq__` and is quite different from `is not None`

Might be worth doing `doSkip = not skipReason` though anyway

src/twisted/trial/test/skipping.py
```diff
@@ -16,18 +16,51 @@
...
+    def test_no_skip_empty_attribute(self):
+        """
+        This test is not skipepd as the skip attribute is the empty string.
+        """
+
+    test_no_skip_empty_attribute.skip = ""  # type: ignore[attr-defined]
+
+    def test_no_skip_none_attribute(self):
+        """
+        This test is not skipepd as the skip attribute is None.
```
exarkun, line 60: typo in skipped here and above

src/twisted/trial/test/test_tests.py
```diff
@@ -161,7 +161,7 @@ def test_counting(self):
         """
         Assert that there are three tests.
         """
-        self.assertCount(3)
+        self.assertCount(5)
```
exarkun, line 164: https://www.youtube.com/watch?v=NxZ7XAYkVo4&ab_channel=GuruMediator

src/twisted/trial/test/test_tests.py
```diff
@@ -174,8 +174,8 @@ def test_results(self):
         self.assertTrue(self.reporter.wasSuccessful())
         self.assertEqual(self.reporter.errors, [])
         self.assertEqual(self.reporter.failures, [])
-        self.assertEqual(len(self.reporter.skips), 3)
-        self.assertEqual(self.reporter.successes, 0)
+        self.assertEqual(len(self.reporter.skips), 2)
+        self.assertEqual(self.reporter.successes, 3)
```
exarkun, line 178: This was really confusing until I looked at what the implementation actually does and what it does is *not* skip a test that has a skip attribute set to `""`.  Isn't this the opposite of the goal of the PR?

src/twisted/trial/test/test_skipif.py
```diff
@@ -61,32 +63,3 @@ def test_shouldShouldRunWithSkipIfFalse(self):
...
         self.assertTrue(True, "Test should run and not be skipped")
-
-
-class SkipAttributeOnClass(TestCase):
-    """
-    All tests should be skipped because skip attribute is set on
-    this class.
-    """
-
-    skip = "'skip' attribute set on this class, so skip all tests"
```
exarkun, line 72: Where did test coverage of handling skip set on a class object go?

adiroiban: Good question. In skipping.py there is SkippedClassMixin  which is used in SynchronousSkippedClass and  AsynchronousSkippedClass... which are used in test_tests.py in SynchronousSkipClassTests and AsynchronousSkipClassTests with assertions in SkipClassesMixin  ...

I tried to keep  this PR as small as possible but I can also add docstrings to SkippedClassMixin in skipping.py

src/twisted/trial/test/test_tests.py
adiroiban: Thanks for the comment. The title of the PR was wrong.
The idea is to handle the empty string as don't skip...
so that you can have

```
if SOMEMTHING:
  skip = 'Shere reason'
else:
  skip = ""
```

Instead of the following code that will trigger a mypy warning.

```
if SOMEMTHING:
  skip = 'Shere reason'
else:
  skip = None
```

PR #1477, opened 2020-11-12
fix typo in LoopingCall.withCount.__doc__

graingert:


adiroiban (approved):
Thanks.

You will need to add the .misc news fragment and wait for all tests to be green and then you can merge this.

PR #1478, opened 2020-11-16
9933 - fix twisted.internet.process.pauseProducing/resumeProducing fo…

ndg63276:
…r python 3

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9933


auvipy (approved):

src/twisted/internet/test/test_process.py
```diff
@@ -843,6 +843,26 @@ def start():
...
+        """
+
+        def pauseAndResume(reactor):
+            try:
+                protocol = ProcessProtocol()
+                transport = reactor.spawnProcess(
+                    protocol, pyExe, [pyExe, b"-c", b""], usePTY=self.usePTY
+                )
+                transport.pauseProducing()
+                transport.resumeProducing()
```
graingert, line 858: we should test this streaming to an object with actual flow control

graingert: here's a test using a MemoryProcessReactor that could be updated to use a real reactor

https://github.com/twisted/twisted/blob/c6c863bf8430f5f1e017af874ffdc89b2cc1b710/src/twisted/internet/test/test_endpoints.py#L822-L1140

exarkun: I'm not totally clear about this review. Is this a request to change the test in the PR?

graingert: To add an additional test that uses a real reactor and process

exarkun: I think the test in the PR does use a real reactor and a real process already.  It uses the "reactor builder" testing framework which runs the test against *all* of the real reactors, even.

graingert: ah ok, that makes sense

graingert (approved):
This looks like a good patch and the test does cover the error here.

I'd like to see a test that checked the side effect of pause/resumeProducing as currently it's just a smoke test

wiml (approved):
This looks good to me.

I largely agree with @graingert  that it would be nice for the test to verify that `pauseProducing`/`resumeProducing` in fact stop-and-start producing data, but presumably the underlying `stopReading`/`startReading` calls are already tested in the reactor tests (and it doesn't seem trivia to add that here), so I don't think that should block merging this PR.

PR #1480, opened 2020-11-19
 Have twisted.mail.pop3 import exceptions from actual source module

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10054

---

I also renamed `twisted.mail.pop3client` to `twisted.mail._pop3client`, to reflect the fact that it's a private implementation module. Its docstring is explicit about that:

```
Don't use this module directly.  Use twisted.mail.pop3 instead.
```

I don't know why it wasn't using the naming convention for private modules, but maybe that wasn't adopted yet in 2004, which is when this module was added (commit 095b750003a3fce6419d7397c90e03a8dd015b33).

Since this was effectively a private module from the start, I don't think we need a deprecation period.

### DIFF
--- a/src/twisted/mail/pop3.py
+++ b/src/twisted/mail/pop3.py
@@ -1677,11 +1677,13 @@ class POP3Client(basic.LineOnlyReceiver):
 
 
 from twisted.mail.pop3client import POP3Client as AdvancedPOP3Client
-from twisted.mail.pop3client import InsecureAuthenticationDisallowed
-from twisted.mail.pop3client import ServerErrorResponse
-from twisted.mail.pop3client import LineTooLong
-from twisted.mail.pop3client import TLSError
-from twisted.mail.pop3client import TLSNotSupportedError
+from twisted.mail._except import (
+    InsecureAuthenticationDisallowed,
+    ServerErrorResponse,
+    LineTooLong,
+    TLSError,
+    TLSNotSupportedError,
+)
 
 __all__ = [
     # Interfaces


graingert:
~@mthuurne you might be able to just delete it as it wasn't ported to py3~

nope this was actually ported

mthuurne:
> ~@mthuurne you might be able to just delete it as it wasn't ported to py3~
> 
> nope this was actually ported

I don't have any production code that uses it, but the unit tests pass under Python 3 at least.

adiroiban (changes requested):
Looks good. Thanks.

The problem is that this break backward compatibility.
pop3client.py is documented as private, but it has a public name.

In theory we should keep a stub pop3client.py which reexport `POP3Client` `OK`, and `ERR`  and the stub module raise a deprecation warning when imported.

twisted.mail.pop3lient.POP3Client and twisted.mail.pop3client.OK were public API.

I don't know how many people are using this code, so since this is documented as private, I think that we can ask for an exception and it shouldn't be hard to get the exception approved.

I can ask for the exception.
The procedure is documented here

https://twistedmatrix.com/documents/current/core/development/policy/compatibility-policy.html#procedure-for-exceptions-to-this-policy

I am placing this as needs changes...and we can merge it after we get the exception approved.

Thanks!

mthuurne:
I split off the renaming of `pop3client` into [Twisted ticket 10055](https://twistedmatrix.com/trac/ticket/10055). I'll create a new PR once this one has been merged, since there is overlap in the touched code.

adiroiban (approved):
Thanks. Much appreciated!

Ready to merge

PR #1481, opened 2020-11-19
Make twisted.mail.pop3client a private module

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10055

---

This was split off from #1480 so that could get merged quicker.

I thought deprecation was unnecessary since the intent of this module was always that it was private, but there is no consensus around that. Therefore we must either go through a deprecation phase or get an official agreement that deprecation is unnecessary in this case.


adiroiban:
I have added the tests for the deprecation.

I see that twisted.mail.pop3.POP3Client is also deprecated.

Maybe we can remove that code as port of this PR as the code is deprecated for a long time.

src/twisted/mail/pop3client.py
```diff
@@ -1,1238 +1,24 @@
...
-            L{bytes} or fails with L{ServerErrorResponse}
-        @return: A deferred which fires when the server response is received.
-            On an OK response, the deferred succeeds with the server
-            response minus the status indicator.  On an ERR response, the
-            deferred fails with a server error response failure.
-        """
-        return self.sendShort(b"QUIT", None)
-
+warnings.warn(
+    "twisted.mail.pop3client was deprecated in NEXT. " "Use twisted.mail.pop3 instead.",
```
graingert:
```suggestion
    "twisted.mail.pop3client was deprecated in NEXT. Use twisted.mail.pop3 instead.",
```

mthuurne: In other deprecations, `Twisted NEXT` is used instead of just `NEXT`. I don't know if that's a requirement for the version substitution to work, but it's probably a good idea to be consistent in any case.

adiroiban: Yes. My bad. It should be "Twisted NEXT" as NEXT is relaced with just the version not "product_name + version "

adiroiban: The `NEXT. " "Use` part was done by black .

adiroiban:
Please check the updated version.

I have searched the code for docstrings with "private" in their content and I was not able to find any other occurrence of a private API which has a public name.

There are some private constructors `__init__`  twisted.python.modules.PythonAttribute and PathEntry

src/twisted/mail/pop3client.py
```diff
@@ -1,1238 +1,24 @@
...
-        """
-        return self.sendShort(b"QUIT", None)
-
+warnings.warn(
+    "twisted.mail.pop3client was deprecated in Twisted NEXT. Use twisted.mail.pop3 instead.",
+    DeprecationWarning,
+    stacklevel=2,
+)
 
-__all__ = []  # type: List[str]
```
mthuurne, line 1238: It seems very unlikely that someone would have `from twisted.mail.pop3client import *` in their code, but if that line does somehow exist somewhere, keeping `__all__ = []` would be better, since it preserves the old situation.

adiroiban: I have added an empty all.

src/twisted/mail/pop3client.py
graingert: @adiroiban @mthuurne you can catch this with https://github.com/keisheiled/flake8-implicit-str-concat

but `black --experimental-string-processing` should fix them soon

graingert (approved):

PR #1482, opened 2020-11-23
Correctly handle an non-zero non-pending error code from recvfrom in …

xdesai:
…IOCP reactor

Summary:
The normal processing loop for the UDP listener with an IOCP reactor is as follows.
1. Client code calls listenUdp
2. After some stack layers we eventually call doRead from _resumeReading(). This will happen exactly once. We do not come back to resumeReading again.
3. doRead creates an IOCP event object and registers a callback called cbRead. It then gives this event to the windows IOCP and says "Call me when something comes into our port"
4. doRead rolls off the end and we return control back to the reactor
5. After some time when data has come into our port our cbRead function is called by the windows OS
6. cbRead alerts the upper layer twisted protocols about our new data.
7. cbRead then calls doRead again. This creates a new IOCP event and sets us up to listen for new data again.
8. We roll off the end of doRead and return control back to the reactor.
9. Repeat forever :)

How the bug was exposed:
There is one edge case that was not handled correctly when this code was initially written.
When we create our IOCP event and ask windows to alert us by calling our callback there are situations when windows will
actually respond immediately instead of calling our registered callback at a later time. In our situation where we discovered the bug the
Windows OS returned a WSAECONNREFUSED or WSAECONNRESET immediately. The code as it was written before this fix would alert
the upper level twisted protocols of the result, but it would NOT create a new IOCP event and register it with the windows. This means
that after we processed the reject message we would never attempt to `recvfrom()` any more data from our port.
In the wild this led to our application essentially ignoring all traffic that came in our UDP port until we restarted the application.

Our proposed solution:
After reading this code to understand it's intended flow and exploring possible solutions we believe that answer to this bug is to simply
call the callback (cbRead) manually. Normally we can expect windows to call our callback for us. But in the case where we get a response immediately
we can just pretend to be the windows OS and call our own callback.

Test Plan:
Wrote a new unit test. When it is run with the IOCP reactor it fails. After our fix it succeeds.

It exposes the problem because it attempts to send a UDP packet to a port that is closed. Windows alerts us by sending an
ICMP packet back to the UDP listener. The recvfrom call returns a WSAECONNRESET which is one of the immediate response types.
So windows never calls the callback we registered.

## Contributor Checklist:

* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10052 
* [X] I ran additional checks listed at [Getting Your Patch Accepted](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted)


src/twisted/test/test_udp.py
```diff
@@ -307,6 +307,66 @@ def cbFinished(ignored):
...
         return d
 
+    def test_server_read_failure(self):
+        """
+        When a server fails to successfully read a packet the server should
+        still be able to process future packets.
+        The IOCP reactor had a historical problem where a failure to read caused
+        the reactor to ignore any future reads. This test should prevent a regression.
+
+        Note: This test assumes no one is listening on port 80 UDP.
```
glyph, line 316: This should allocate a port in the safe way for tests (i.e.: listen on port zero and get the bound port).

src/twisted/test/test_udp.py
```diff
@@ -307,6 +307,66 @@ def cbFinished(ignored):
...
+        """
+        client = GoodClient()
+        clientStarted = client.startedDeferred = defer.Deferred()
+        client_port = reactor.listenUDP(0, client, interface="127.0.0.1")
```
glyph: Twisted naming convention would have this as `clientPort` - this should be locally consistent.

src/twisted/test/test_udp.py
```diff
@@ -307,6 +307,66 @@ def cbFinished(ignored):
         d.addCallback(cbFinished)
         return d
 
+    def test_server_read_failure(self):
```
glyph: Twisted coding standard for method names has this as `test_serverReadFailure`; I'm sorry the docs don't make this clearer and more discoverable.

src/twisted/internet/iocpreactor/udp.py
```diff
@@ -167,7 +167,12 @@ def doRead(self):
         )
 
         if rc and rc != ERROR_IO_PENDING:
-            self.handleRead(rc, data, evt)
+            # If the error was not 0 or IO_PENDING then that means recvfrom() hit a
+            # failure condition. In this situation recvfrom() gives us our response
+            # right away and we don't need to wait for Windows to call the callback
+            # on our event. In fact, windows will not call it for us so we must call it
+            # ourselves manually
+            self.reactor.callLater(0, self.cbRead, rc, data, evt)
```
glyph, line 176: Hmm. I love the big comment here but are there any re-ordering concerns with this `callLater`? It feels a bit odd to defer this in a way which might get called out of order; why not call it synchronously?

glyph (commented):
This mostly looks great. I'm sorry that the review process takes so long; but, all-volunteer project, global sanity-shattering pandemic, you know the drill 😉 .

If you (or anyone else) can address the minor test and naming issues (and answer the ordering question - pretty sure that's just a failure of my own understanding) then it should be good to merge.

src/twisted/test/test_udp.py
xdesai: Oh, okay! Will fix!

src/twisted/test/test_udp.py
xdesai: So, I believe in this test I actually am correctly using the "listen on zero" when trying to write to the actual port.
However, I see how my comment about port 80 is confusing. Port 80 in this test actually represents a port that intentionally has no listener. So I don't actually want the port to be allocated and the test never does try to allocate on it.

The reason I need a guarantee that some port is not listening is so that we can trigger windows to respond immediately to the recvfrom and trigger the buggy situation.

I agree that just assuming nobody is listening on port 80 UDP is probably not ideal though. Do you have any ideas on how we might be able to figure out a port that we can guarantee has no listener?

src/twisted/test/test_udp.py
xdesai: on it!

src/twisted/internet/iocpreactor/udp.py
xdesai: Great question!
I had originally created this as a synchronous call, but some of my teammates pointed out that there might a potential for a tight infinite loop here so we intentionally added in the callLater to give the reactor a chance to butt in.

The infinite loop we were worried about:
- Synchronously call `cbRead()` on this line
- `cbRead()` calls the `doRead()` function (basically recursion with one function inbetween)
- `recvfrom()` has another early response
- Synchronously call `cbRead()` 
- Repeat forever

I'm very curious to hear your thoughts!

xdesai (commented):
Thanks for getting back to us on this :)
I totally understand that things are a bit slower these days and I know IOCP is a more niche use case so I'm very grateful for your thoughts.

src/twisted/internet/iocpreactor/udp.py
wiml: Looking at the surrounding code, I think the asynchronous version is probably actually better, not just because it avoids the recursion possibility that @xdesai mentions, but also because it means that the immediate-error case is presented to the caller in the same way that a delayed-error (or success) case would be. In my opinion this reduces one opportunity for the caller to have a bug. The caller would have had to handle delayed errors as asynchronous calls even without this change, so this doesn't seem likely to expose any bad behavior in the caller.

+1 as far as I'm concerned, although I should point out I'm not familiar with IOCP, I'm just reading the code.

src/twisted/test/test_udp.py
wiml: I can't think of an official way to reserve a port without listening on it, and your approach seems reasonably reliable to me (using a port in the reserved range, but which has no defined use).

I had the same immediate misunderstanding as @glyph, so it might be worth amending the `Note:` comment to make it clearer. I think it's fine as it is, though.


```suggestion
        Note: This test assumes no one is listening on port 80 UDP, so that we
        can be sure to elicit a failure when we write.
```

wiml (approved):

mbish:
Would it be possible to merge this into master soon? We're using this patch internally already but it'd be great to have it in the mainline repo. Just wanted to check back in before we forgot about it. Thanks again for looking into this!

glyph:
Hi @mbish,

It looks like @wiml approved this but didn't merge it.  @wiml — given that this is an external contributor, if it's got your stamp of approval, can you merge it in?

@mbish, if Wim is not responsive (we are all volunteers here after all, sometimes availability is spotty) then you can get it back into the queue at https://twisted.reviews/ for someone else with spare capacity to do Twisted reviews to notice by re-adding the "review" keyword over on the Trac ticket.

twm:
Merging this as it appears to have been cleared to land. Thanks everyone!

PR #1483, opened 2020-11-24
[#7338] Add support for HTTP to HTTPS redirection on a TLS port.

adiroiban:
This is still a proof of concept stage.

The ticket is now https://github.com/twisted/twisted/issues/7338

## Scope

This adds an HTTPS Site that handles TLS and HTTP to HTTPS redirection.

## Use case

You have a Twisted based application which runs the web admin dashboard on a custom port, ex 1234.

443 port is used by the user facing app.

Admin user remember that the admin dashboard is on port 1234.
The open a browser and type :  NAME.OF.THE.HOST:1234 
The browser (IE, Firefox, and Chrome as of today) will default to HTTP.
The connection fails as port 1234 is waiting for TLS handshake.
The admin user is panicked and calls  support :)

## To do

* Make mypy happy

## How to test

```
$ trial twisted.web.test.test_web
$ trial --without-module OpenSSL twisted.web.test.test_web
```

For manual test on Linux. Generate the server key+cert and then run the example in a terminal

``` 
$ openssl req -new -newkey rsa:2048 -days 3 -nodes -x509 \
        -keyout server-key.pem -out server-cert.pem
$ python docs/web/examples/https_http_redirect_server.py
```
In another terminal or a web browser try to access localhost over port 8443 with HTTP and HTTPS

```
    # See HTTP to HTTPS redirection.
    curl -v http://localhost:8443
    # See that HTTPS works
    curl -kv https://localhost:8443
```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/490


src/twisted/web/server.py
```diff
@@ -878,3 +884,85 @@ def acceptableProtocols(self):
...
+        self.transport.loseConnection()
+        return b""
+
+
+class HTTPSSiteWrapper(TLSMemoryBIOFactory):
+    """
+    An HTTPS L{Site} wrapper that handles TLS setup and HTTP to HTTPS redirection.
+    """
+
+    def __init__(self, contextFactory, site):
```
adiroiban, line 961: the actual site is initiated outside of the `HTTPSSiteWrapper` to reduce any confusion
The HTTP requestion will have the `request.site` attribute set to the outside site and not to this HTTPSSiteWrapper.

src/twisted/web/server.py
```diff
@@ -30,17 +30,23 @@
...
 from twisted.web.error import UnsupportedMethod
-
 from incremental import Version
 from twisted.python.deprecate import deprecatedModuleAttribute
 from twisted.logger import Logger
 
+try:
+    from twisted.protocols.tls import TLSMemoryBIOFactory
+except ImportError:
+    TLSMemoryBIOFactory = object
```
adiroiban, line 47: mypy complains about this . but I have no idea how to make mypy happy while still having a code that is easy to read and use.

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+
+                segments = part.split(b":", 1)
+                if len(segments) != 2:
+                    return b""
+
+                return segments[1].strip()
+
+            return b""
+
+        # We only want to handle the first package.
```
glyph, line 935: The word is "packet", not "package"

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+        After that, the dataReceived method of the wrapped protocol is called.
+
+        It checks if an HTTP request is made over HTTPS port, instead of an
+        SSL handshake.
+
+        @param data: The data received from the remote HTTP client peer.
+        @type data: L{bytes}.
+        """
+
+        def get_http_host(data):
```
glyph, line 908: `getHTTPHost` would be the Twisted-naming-standard version of this

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+
+        self.transport.write(
+            b"HTTP/1.1 301 Moved Permanently\r\n"
+            b"Location: https://%s\r\n"
+            b"Cache-Control: no-store, no-cache\r\n"
+            b"Connection: close\r\n"
+            b"Content-Length: 8\r\n"
+            b"\r\n"
+            b"redirect" % (host,)
+        )
```
glyph, line 952: Don't we have like, a lot of code for synthesizing HTTP/1.1 responses in this codebase somewhere?  Could we do this without just having a big static string?  (`Content-Length: 8` seems *particularly* brittle, here)

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+        self.transport.loseConnection()
+
+
+class HTTPSSiteWrapper(TLSMemoryBIOFactory):
```
glyph, line 956: 1. Could you do this with composition rather than inheritance?  The whole point of `TLSMemoryBIOFactory` is that it's a regular Factory can you can wrap it with `WrappingFactory` et. al.
2. Calling this `HTTPSSiteWrapper` seems like a *very* misleading name; you absolutely don't want this and shouldn't use it most of the time.  In fact in some load-balancer deployments it's going to cause a lot of trouble.  Perhaps something more like `PortConfusionFixerForHTTPS`?

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+        @return: A protocol to handle the new connection.
+        @rtype: L{twisted.internet.interfaces.IProtocol}
+        """
+        protocol = self.protocol(self, self.wrappedFactory.buildProtocol(addr))
+        return _HTTPSRedirectProtocolWrapper(
+            factory=WrappingFactory(self._site), wrappedProtocol=protocol
+        )
+
+    def __getattr__(self, name):
+        return getattr(self._site, name)
```
glyph, line 993: Is this level of delegation really necessary?

src/twisted/web/server.py
glyph, line 47: I think this should work?

```suggestion
from typing import TYPE_CHECKING
try:
    from twisted.protocols.tls import TLSMemoryBIOFactory
except ImportError:
    if not TYPE_CHECKING:
        TLSMemoryBIOFactory = object
```

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+        # We only want to handle the first package.
+        self.dataReceived = self.wrappedProtocol.dataReceived
+
+        host = get_http_host(data)
+
+        if not host:
+            # Just forward the data to the wrapped protocol.
+            return self.dataReceived(data)
+
+        self.transport.write(
```
glyph, line 944: We should *really* emit a noisy log message here, since this is a bad and very odd misconfiguration that you should probably try to fix.  Maybe log something about the client's configuration?

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+            """
+            Return the value of the Host header found inside `data` HTTP
+            raw request.
+
+            @param data: The first data chunk received from the remote HTTP client peer.
+            @type data: L{bytes}.
+
+            @return: The value of the Host header. It returns the empty string when the value could not be extracted.
+            @rtype: L{bytes}
+            """
```
glyph, line 918: I feel like this level of docstring detail is a bit overkill in a callback (nested callables like this don't need docstrings per the coding standard)

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
...
+            parts = data.split(b"\r\n")
+            if len(parts) == 1:
+                return b""
+
+            for part in parts:
+                if not part.lower().startswith(b"host"):
+                    continue
```
glyph, line 925: The `Host` header isn't necessary *guaranteed* to show up in the first packet, so this is a bit of a gross heuristic.  I wonder if this might not be better served by a precompiled binary regex, rather than splitting everything up into lines, since we only want to process *one* header here?

src/twisted/web/server.py
```diff
@@ -878,3 +884,110 @@ def acceptableProtocols(self):
             baseProtocols.insert(0, b"h2")
 
         return baseProtocols
+
+
+class _HTTPSRedirectProtocolWrapper(ProtocolWrapper):
+    """
+    Wraps the TLS protocol and redirect to HTTPS if an HTTP request is
+    detected after a connection is made.
```
glyph, line 892: I feel like this doesn't really cover the inherently potentially-broken nature of the one-reassembled-TCP-segment heuristic being used here.  (Or maybe that's more of an issue for the public docstring, but it's not covered there, either...)

glyph (changes requested):
Is this really a common type of misconfiguration?  I feel like the Internet was a different kind of place when this ticket was filed, where http/https confusion might have been possible and custom port numbers were really common.

But if we were going to solve this problem on today's internet, I almost think that the right way to do it is to attempt parsing HTTP on *any* kind of connection, not just HTTPS.  I can't remember the last time that I wanted this, but I *do* frequently want to run SSH and HTTPS on the same port.

If you want to merge this as-is, we need a better name for it, better documentation of the heuristic nature of the protocol detection, and ideally a better way to construct the response that doesn't hard-code "redirect" as the body of the response: if this specialized use-case is something you care about, presumably you might want to customize the response as well. (And for that matter you might want to stick a whole plain-text HTTP site in there, not just a redirect; if you're having this type of problem your clients must be a special kind of broken).

But I wonder if it might not be better to go back to the drawing board, add a custom content-sniffing protocol detector, and have this work across imap/imap-over-tls/http/https/ssh etc; sniff content XYZ via a regex, tell me about the groups you found in the regex, and I'll make a new protocol factory, with a configurable timeout / buffer maximum.

src/twisted/web/server.py
adiroiban: True.

The use case here is an HTTP request made by a web browser.
AFAIK all web browser will send the host header.

The use case for this patch is confused users that forget to use `HTTPS://` when accessing a web server.

If you don't have a host, it probably means  that you have a custom HTTP client and in that case you don't want a redirect.

src/twisted/web/server.py
adiroiban: I will remove the docstring.

Regarding the name... I thought that getHTTPHost is  only for methods ...and functions names are with underscore.

Will review the coding style and update future patches.

Thanks

src/twisted/web/server.py
adiroiban: Well, the client configuration in this use case is the "human brain" 

The user types 'service.example.com:1234` in the browser  and the browser makes an HTTP request.
The fix for  the user to remember to type `https://service.example.com:1234`

I don't know how to fix the human brain so this is why I have not emitted a log.
But I think that a log should be emitted , to detect misconfigurations made by automated requests.

src/twisted/web/server.py
adiroiban: True. I was lazy :)

I will see what code can be reused... but I guess that this will need instantiating a Request to send the response.

src/twisted/web/server.py
adiroiban: Yep. PortConfusionFixerForHTTPS and composition is the  way to go. Thanks

src/twisted/web/server.py
adiroiban: I remember that  it was necessary .  But I can remove it, run the tests and see what breaks :)

adiroiban (commented):
Thanks for the review. 

The patch was created as quick fix after a confused users called support in the middle of the night to complain that the service is down :)

I don't know if a general protocol detection framework will work.
Some protocol expect the server to send the first message before the client sending its request.

For example, an SSH client will connect to the server, will wait to receive the initial data from the server and only then sent its own initial data. AFAIK the SSH spec doesn't require the client to make the first step.

But it guess that it can work with HTTP and other protocol in which  and have a fallback protocol.

------------

This was a quick hack and I wanted to share the patch to see if others have the same issue.

If this is not a common issue with Twisted users, I can keep the patch in my fork and there is no point in adding it to Twisted core functionality.

adiroiban:
Thanks for your review. It is very useful.

> But if we were going to solve this problem on today's internet, I almost think that the right way to do it is to attempt parsing HTTP on any kind of connection, not just HTTPS. I can't remember the last time that I wanted this, but I do frequently want to run SSH and HTTPS on the same port.

I am going to close this PR and will rethink it and I have no idea when I will have time to work on this.... since it's low impact.

I will see if there is a way to implement it as general functionality that does generic factory dispatching.

I think that a generic dispatcher for the factory to be used for a connection is the best way to implement it.

For HTTP to HTTPS redirection it can dispatch the connection it to a factory that will respond with a site and implement the redirection as a resource on that site.

PR #1484, opened 2020-11-25
[9944] Enable and fix the skipped TLS tests.

adiroiban:
## Trac ticket copy/paste

According to a recent coveralls.io report:

​https://coveralls.io/builds/32817266/source?filename=src/twisted/protocols/test/test_tls.py

twisted.protocols.test.test_tls is essentially completely unexecuted across coverage-gathering CI jobs.

This appears to be due to a change in pyOpenSSL 19.1.0 which removed ConnectionType. The whole test_tls module is skipped if this can't be imported.

This suggests the module may have been skipped on CI since sometime in 2019.

## Changes

Update to latest Python OpenSSL package.

One test was failing as renegotiation is only available in TLSv1.2 or less.

------

The ALPN negotiation tests were failing as I have no idea why ClientNegotiationFactory is not working.
I tried to debug it and make sure we got the same context .. and that ALPN protocols are set...but I don't understand why it is not working.

I have removed the code and created a separate ticket for that https://twistedmatrix.com/trac/ticket/10067

## Testing

```
$ trial twisted.protocols.test.test_tls
$ trial --without-module OpenSSL twisted.protocols.test.test_tls
```

I have also executed manual ALPN tests by starting

```
$ python docs/core/examples/tls_alpn_npn_server.py
```

and the in another terminal run

```
$ python docs/core/examples/tls_alpn_npn_client.py
```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9944


adiroiban:
There is a test with a side effect

```
$ trial twisted.protocols.test.test_tls.IProtocolNegotiationFactoryTests.test_negotiationWithProtocolOverlap
twisted.protocols.test.test_tls
  IProtocolNegotiationFactoryTests
    test_negotiationWithProtocolOverlap ...                             [ERROR]

[ERROR]
  File "/home/adi/chevah/twisted/src/twisted/protocols/test/test_tls.py", line 1958, in checkNegotiatedProtocol
    self.assertEqual(client.negotiatedProtocol, b"h2")
  File "/home/adi/chevah/twisted/src/twisted/trial/_synctest.py", line 424, in assertEqual
  File "/usr/lib/python3.8/unittest/case.py", line 912, in assertEqual
  File "/usr/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
twisted.trial.unittest.FailTest: None != b'h2'

twisted.protocols.test.test_tls.IProtocolNegotiationFactoryTests.test_negotiationWithProtocolOverlap
Ran 1 tests in 0.008s

```

The test that currently fails is ok

```
$ trial twisted.protocols.test.test_tls.NonStreamingProducerTests
twisted.protocols.test.test_tls
  NonStreamingProducerTests
    test_interface ...                                                     [OK]
    test_pause ...                                                         [OK]
    test_resume ...                                                        [OK]
    test_resumeProducingRaiseAndUnregisterProducerRaises ...               [OK]
    test_resumeProducingRaises ...                                         [OK]
    test_stopProducing ...                                                 [OK]
    test_stopStreamingTwice ...                                            [OK]
    test_writeUntilDone ...                                                [OK]

Ran 8 tests in 0.011s

PASSED (successes=8)

```

adiroiban:
Another test failing

```
$ trial twisted.protocols.test.test_tls.TLSProducerTests.test_producerDuringRenegotiation
twisted.protocols.test.test_tls
  TLSProducerTests
    test_producerDuringRenegotiation ...                                [ERROR]

[ERROR]
  File "/home/adi/chevah/twisted/src/twisted/protocols/test/test_tls.py", line 1182, in test_producerDuringRenegotiation
    tlsc.renegotiate()
  File "/home/adi/chevah/twisted/.tox/build3/lib/python3.8/site-packages/OpenSSL/SSL.py", line 1921, in renegotiate
    _openssl_assert(_lib.SSL_renegotiate(self._ssl) == 1)
  File "/home/adi/chevah/twisted/.tox/build3/lib/python3.8/site-packages/OpenSSL/_util.py", line 67, in openssl_assert
    exception_from_error_queue(error)
  File "/home/adi/chevah/twisted/.tox/build3/lib/python3.8/site-packages/OpenSSL/_util.py", line 54, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.SSL.Error: [('SSL routines', 'SSL_renegotiate', 'wrong ssl version')]

twisted.protocols.test.test_tls.TLSProducerTests.test_producerDuringRenegotiation
```

adiroiban:
I have no idea how to fix the ALPN negotiation tests so that it will not require the call to _setAcceptableProtocols for the client.

I am asking for a review hoping that someone might know what is wrong there.

`_setAcceptableProtocols` is called for ClientNegotiationFactory but I don't know why it is not active.

I run a manual tests from the example for ALPN and it is working.. but that test is using SSL4ServerEndpoint and listenSSL

src/twisted/protocols/test/test_tls.py
```diff
@@ -1899,108 +1922,111 @@ def dataReceived(self, data):
...
         clientFactory = ClientNegotiationFactory(clientProtocols)
-        clientFactory.protocol = lambda: NotifyingSender(clientDataReceived)
+        clientFactory.protocol = lambda: NotifyingSender(
+            clientDataReceived, b"data from client"
+        )
 
-        clientContextFactory, _ = HandshakeCallbackContextFactory.factoryAndDeferred()
+        clientContextFactory = ClientTLSContext()
+        # See why ClientNegotiationFactory is not working.
+        _setAcceptableProtocols(clientContextFactory.getContext(), clientProtocols)
```
adiroiban: this call should not be made here... as ClientTLSContext trigger the automatic call.

src/twisted/protocols/test/test_tls.py
```diff
@@ -1163,38 +1171,56 @@ def test_producerDuringRenegotiation(self):
...
         producer exactly once.
+
+        Renegotiation only works on TLSv1.2 or less.
         """
-        c, ct, cp = self.setupStreamingProducer()
-        s, st, sp = self.setupStreamingProducer(server=True)
+        clientProtocol, clientTransport, _ = self.setupStreamingProducer()
+        serverProtocol, serverTransport, serverProducer = self.setupStreamingProducer(
+            server=True, serverMethod=TLSv1_2_METHOD
+        )
```
exarkun, line 1189: The new locals names are better but reviewing this method was a bit trickier with both renaming and other changes mixed together, for what it's worth.

src/twisted/protocols/test/test_tls.py
```diff
@@ -1899,108 +1922,111 @@ def dataReceived(self, data):
...
-        clientDataReceived.addCallback(lambda ignored: serverDataReceived)
-        serverDataReceived.addCallback(checkNegotiatedProtocol)
+        return deferred.addCallback(checkNegotiatedProtocol)
+
+    def test_negotiationNoOverlap(self):
+        """
+        When there is no overlap between client and server the connetion fails.
+
+        This test might fail on older OpenSSL version.
+        Prior to version 1.1.0 of OpenSSL, failing the ALPN negotiation does not
```
exarkun: It seems like this could be expressed as a skip?  The OpenSSL version in use should be available to us via either PyOpenSSL or cryptography.  I don't know if this is really a good idea or not because I don't know if there's a clear divide between OpenSSL where this test works and where it doesn't.  Is handshake failure because of ALPN failure a proper feature of OpenSSL now or just a random accident?

But a version-based skip seems a little bit better than just some words in the docstring.

src/twisted/test/ssl_helpers.py
```diff
@@ -18,21 +18,38 @@
...
-        ctx = SSL.Context(self._method)
-        ctx.use_certificate_file(self.filename)
-        ctx.use_privatekey_file(self.filename)
-        return ctx
+        if self._context is None:
+            self._context = SSL.Context(self._method)
+            self._context.use_certificate_file(self.filename)
+            self._context.use_privatekey_file(self.filename)
+
+        return self._context
```
exarkun: Are any of these caching changes actually necessary?  It doesn't seem like they should be.

Ahh.  I guess this was an attempt to fix the ALPN issue.  It doesn't seem like it should be necessary and I don't think we can *force* applications to cache their context anyway.  So we should probably back this out and find another solution to the failing ALPN test.

src/twisted/protocols/test/test_tls.py
```diff
@@ -1899,108 +1922,111 @@ def dataReceived(self, data):
...
+        deferred = self.handshakeProtocols(
+            clientProtocols=[b"http/1.1", b"h2", b"client"],
+            serverProtocols=[b"h2", b"http/1.1", b"server"],
+        )
 
-        def checkNegotiatedProtocol(ignored):
+        def checkNegotiatedProtocol(result):
+            client, server = result
             self.assertEqual(client.negotiatedProtocol, b"h2")
             self.assertEqual(server.negotiatedProtocol, b"h2")
```
exarkun: So this seems to be the test where the explicit `_setAcceptableProtocols` call matters.  It seems like our ALPN implementation is doing what we think it should be doing - ie, even without the explicit `_setAcceptableProtocols` in the tests, the implementation makes sure to call the right OpenSSL ALPN APIs.  But then they don't seem to have any effect.

I'll try to dig into this more, I guess.  It seems like we've misunderstood the OpenSSL ALPN APIs or they have a bug or they've subtly changed their behavior somehow.

exarkun (commented):
Thanks.  I've left some initial comments.  I think I see the ALPN problem and I'll try to look at it some later today.

src/twisted/protocols/test/test_tls.py
adiroiban: This tests is always green an all "supported" OpenSSL version.
Where supported mean OpenSSL version for which we run automated tests.

I left if without a skip, so that first time when we see it fails, we can review which version of OpenSSL is supported.
If that version of OpenSSL is supported and this test still fails, only then add the skip.

If the handshake doesn't fail, you end up with a "stalled"  / hangled / idle connection.

And I think that is important to know if the connection hangs and you can end up with a server with many such connections.

I can add a skip...but since so far this tests never fails... I decided that is best to left the comment for future debugging.

I guess that NPN is dead and people will use ALPN with 1.1.1 OpenSSL.

Also OpenSSL 1.1.0 is EOL  August 31, 2018 ... so maybe  just delete this comment without skipping this test as we will not support OpenSSL 1.1.0..

RHEL 7 has 1.0.2k-fips which was the LTS with EOL  December 31, 2019

RHEL 8 has 1.1.1c

The bottom line,  I don't know on which version I should  skip :) ... and I think that with current and future OpenSSL supported version this should never be skipped.

exarkun: > This tests is always green an all "supported" OpenSSL version.

I was only thinking of a skip for the known unsupported case.  But spelling it out explicitly like this, it doesn't seem worth it.  There are tons of unsupported versions of libraries that will make Twisted do wrong things and we  don't skip other tests for those cases.  So nevermind.

> Also OpenSSL 1.1.0 is EOL August 31, 2018 ... so maybe just delete this comment without skipping this test as we will not support OpenSSL 1.1.0..

This seems reasonable.  We don't go out of our way to make notes about other obsolete library versions.  OTOH, OpenSSL is complex and hard so the extra info isn't a bad thing.  Maybe make the comment more explicit about the fact that the information is left for future maintainers (actually, maybe a comment would be more appropriate for this information, compared to a docstring) and that the mentioned version of OpenSSL is explicitly not supported.

src/twisted/test/ssl_helpers.py
adiroiban: No.
They should be reverted once the fix this... or even sooner.

Thanks for the headup...as I would have most probably forgot about this.

I left them there to help with debugging. As far as I remember, in some places I only has a reference to the factory and not to the context...so I cached them to make it easy to retrieve them, without adding other helpers in the implementation code.

I think that is important not to cache them during tests as it helps detect implementation issues.
So maybe add a comment to this implementation that the context are not cached on purpose.

One of my guess was that the connection is not set up with the same context...so I have updated this code here.

But yes. This should be reverted.

I am curious if there is any production server which uses a different context for each connection.

I guess that you can have an ACME (Let's Encrypt) context factory that will automatically update the cert in the background. The new connections can get a refreshed certificate without the need to restart the port. 


I can imagine that you can have a server that will server one certificate if the source is an internal IP and another certificate if is an external IP ... but the current API does not support this use case, as getContext() is called without knowing the peer that will handle TLS

src/twisted/protocols/test/test_tls.py
```diff
@@ -16,19 +16,20 @@
...
         Context,
-        ConnectionType,
+        Connection,
         WantReadError,
     )
+    from twisted.internet._sslverify import _setAcceptableProtocols
 except ImportError:
     # Skip the whole test module if it can't be imported.
-    skip = "pyOpenSSL 0.10 or newer required for twisted.protocol.tls"
+    skip = "pyOpenSSL 16.0.0 or newer required for twisted.protocol.tls"
```
exarkun, line 31: I suppose this is the anti-pattern that led to the ticket being resolved in the first place.

This shows up in Twisted over and over so it's hardly a problem specific to either this module or this PR (and this PR is only changing some of the details of the pre-existing instance of this anti-pattern).

So.  This comment is not a blocker.  That said, I think a good replacement would be a pattern where the skip only happens if the dependency isn't met, not if *any* `ImportError` is raised.  It would be nice if this case fit into the model that the various skip decorators provider... but since there are possibly failing imports it's not clear that it does.  Or, I dunno, what about:

```
from OpenSSL import __version__ as pyOpenSSLVersion
try:
    from OpenSSL.SSL import (
        ...
    )
except ImportError:
    pass

@skipIf(pyOpenSSLVersion < "16.0.0", "pyOpenSSL 16.0.0 or newer required for twisted.protocol.tls")
class SomeSSLTests(...):
    ...
```

If the version check is the correct one to make then the tests won't run and it doesn't matter if the imports succeeded or not.

If the version check is out of step with reality then the tests will fail and someone can look at the problem.  That's exactly what we would have wanted to happen when the first `ConnectionType`-lacking pyOpenSSL release came out, I think.

I guess in this case we want to skip the whole module worth of tests so maybe this actually ends up looking like:

```
from OpenSSL import __version__ as pyOpenSSLVersion
try:
    from OpenSSL.SSL import (
        ...
    )
except ImportError:
    pass

if pyOpenSSLVersion < "16.0.0":
    skip ="pyOpenSSL 16.0.0 or newer required for twisted.protocol.tls"
```

src/twisted/protocols/test/test_tls.py
```diff
@@ -140,7 +141,9 @@ def connectionLost(self, reason):
             log.err(reason)
 
 
-def buildTLSProtocol(server=False, transport=None, fakeConnection=None):
+def buildTLSProtocol(
+    server=False, transport=None, fakeConnection=None, serverMethod=None
```
exarkun, line 144: At least the new parameter should be covered in the docstring.

src/twisted/protocols/test/test_tls.py
```diff
@@ -1114,7 +1117,9 @@ def drain(self, transport, allowEmpty=False):
         self.assertEqual(bool(allowEmpty or value), True)
         return value
 
-    def setupStreamingProducer(self, transport=None, fakeConnection=None, server=False):
+    def setupStreamingProducer(
+        self, transport=None, fakeConnection=None, server=False, serverMethod=None
```
exarkun, line 1122: docstring

src/twisted/protocols/test/test_tls.py
```diff
@@ -1851,17 +1877,6 @@ def acceptableProtocols(self):
 class IProtocolNegotiationFactoryTests(TestCase):
```
exarkun: Isn't this the TestCase (Or at least one of its test methods) that was being deleted and then fixed later in https://twistedmatrix.com/trac/ticket/10067 ?

exarkun (commented):
Thanks.  Sorry I let this drop for a while.  I like the plan to split this into the easy parts and the harder ALPN parts.  I'm not sure the plan has been executed here though?

Anyway, some comments inline.

src/twisted/protocols/test/test_tls.py
twm: This looks like a great idea to me. I filed a ticket to make it policy: https://twistedmatrix.com/trac/ticket/10082

adiroiban:
Thanks for the review. I have it on my to do list. Will address the comments and ask a new review.

src/twisted/protocols/test/test_tls.py
adiroiban: I left my comment at https://twistedmatrix.com/trac/ticket/10082

src/twisted/protocols/test/test_tls.py
adiroiban: True.

src/twisted/protocols/test/test_tls.py
adiroiban: True :)

src/twisted/protocols/test/test_tls.py
adiroiban: Understood. 

I don't know what to say... there is a lot of overhear to create a ticket + pr + wait for tests.

Yes. Reviewing just the renaming would be easier for the reviewer ... but I would say that is a lot of wasted time for the PR author.

At the same time, I understand that with Twisted the review time is scarce.

I will try  to make as little change as possible and move typos/renaming in separate PRs.

src/twisted/protocols/test/test_tls.py
```diff
@@ -1846,161 +1871,3 @@ def acceptableProtocols(self):
         @rtype: L{list} of L{bytes}
         """
         return self._acceptableProtocols
-
-
-class IProtocolNegotiationFactoryTests(TestCase):
```
adiroiban, line 1851: @exarkun  this was moved for a separate ticket https://twistedmatrix.com/trac/ticket/10067

adiroiban (commented):
This should be ready for a new view.

Local tests look ok

```
$ trial -u -j 14 src/twisted/protocols/test/test_tls.py
```

Maybe trial should have something like `-u 50` to automatically stop after 50 re-runs.

Looks like there is a bug in distrial.

If I try to stop these tests with Ctrl+C I get

```
  File "/home/adi/chevah/twisted/src/twisted/trial/_dist/disttrial.py", line 258, in runUntilFailure
    return self.run(suite, untilFailure=True)
  File "/home/adi/chevah/twisted/src/twisted/trial/_dist/disttrial.py", line 248, in run
    reactor.run()
  File "/home/adi/chevah/twisted/src/twisted/internet/base.py", line 1423, in run
    self.mainLoop()
  File "/home/adi/chevah/twisted/src/twisted/internet/base.py", line 1433, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/home/adi/chevah/twisted/src/twisted/internet/base.py", line 974, in runUntilCurrent
    f(*a, **kw)
  File "/home/adi/chevah/twisted/src/twisted/internet/base.py", line 713, in stop
    raise error.ReactorNotRunning("Can't stop reactor that isn't running.")
twisted.internet.error.ReactorNotRunning: Can't stop reactor that isn't running.
```

graingert (approved):

graingert:
@adiroiban can you review the changes I've made before merging

adiroiban:
Thanks @graingert  for the updates. Since I am the original author of this PR, I can't review + approve the changes.

I did a quick review and the changes look ok.

Why is types-pyOpenSSL not part of the latest pyOpenSSL package?

I hope that @exarkun or another @twisted/twisted-contributors will have time to review and approve it.

adiroiban:
@graingert if you are happy with the PR I think that we can merge it.

There is no change in production code. Only typing annotations.

The changes are in a testing code that  in trunk is never actually executed.

If we found there are issue with the changes, we can revert or fix them.

I think that this PR is an improvement over trunk and it can be merged.

is better to have at least half of the test executed.

cheers

graingert:
> Why is types-pyOpenSSL not part of the latest pyOpenSSL package?


@adiroiban

https://pypi.org/project/types-pyOpenSSL/ was extracted from typeshed automatically, and https://pypi.org/project/pyOpenSSL/ would need to merge them in

PR #1485, opened 2020-11-27
 Fix documentation of non-existing parameters

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10059

---

Something I noticed while doing this is that `ITransport.writeSequence()` uses `data` as its argument name, but a lot of implementations use different names such as `seq` and `iovec`. I think these should be considered bugs, since these arguments can currently only be passed safely positionally, not via keywords. It will also cause problems once type annotations are added and mypy no longer ignores these methods.

### DIFF
--- a/src/twisted/application/internet.py
+++ b/src/twisted/application/internet.py
@@ -571,6 +571,10 @@ class _ClientMachine:
     """
     State machine for maintaining a single outgoing connection to an endpoint.
 
+    @ivar _awaitingConnected: notifications to make when connection
+        succeeds, fails, or is cancelled
+    @type _awaitingConnected: list of (Deferred, count) tuples
+
     @see: L{ClientService}
     """
 
@@ -583,10 +587,6 @@ class _ClientMachine:
         @param log: The logger for the L{ClientService} instance this state
             machine is associated to.
         @type log: L{Logger}
-
-        @ivar _awaitingConnected: notifications to make when connection
-            succeeds, fails, or is cancelled
-        @type _awaitingConnected: list of (Deferred, count) tuples
         """
         self._endpoint = endpoint
         self._failedAttempts = 0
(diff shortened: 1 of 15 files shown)


adiroiban:
>  I think these should be considered bugs, since these arguments can currently only be passed safely positionally, not via keywords.

+1 I think that zope mypy  extension can detect this type of errors.

adiroiban (approved):
Wow. Great work. Thanks!

This can be merged.

Looking forward for updating pydoctor and have it fail on workings

RIght now, there are some warinings in pydoctor but the Documentation run is still green.

https://github.com/twisted/twisted/runs/1464285550#step:8:29

mthuurne:
> Looking forward for updating pydoctor and have it fail on workings
> 
> RIght now, there are some warinings in pydoctor but the Documentation run is still green.
> 
> https://github.com/twisted/twisted/runs/1464285550#step:8:29

We still have the issue that the processing order of circular dependencies can determine whether a reference is found or not. I'm not sure if that is the cause for this particular failure, but it could cause failures like this, in a flaky way.

So while I agree with the goal, I don't think activating "strict mode" is feasible yet with the upcoming pydoctor release. It will require bigger changes, such as separating the code model (implementation side) from the doc model (public API side).

mthuurne:
Looking at the error message, it does seem like this is an outdated version of pydoctor being used (we improved the phrasing of that particular message), so perhaps this error will be gone after updating.

adiroiban:
> Looking at the error message, it does seem like this is an outdated version of pydoctor being used (we improved the phrasing of that particular message), so perhaps this error will be gone after updating.

Yep.
Right now twisted uses  :

```
[testenv:apidocs]
deps=https://github.com/twisted/pydoctor/archive/3f9c64829dfa040b334c9ae27c332c7078356e79.zip
```
see https://github.com/twisted/twisted/blob/trunk/tox.ini#L135

PR #1488, opened 2020-12-04
Fix domain names resolution

taroved:
I'm going to implement dns names resolution. The current fix replace dns ANY request with A and AAAA requests.

Current ANY request doesn't provide IP addresses for most domain names

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10062


src/twisted/names/newsfragments/10062.bugfix
```diff
@@ -0,0 +1 @@
+Replace dns client ANY request with A and AAAA requests
```
exarkun, line 1: This doesn't follow the style-guide for newsfragments: https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles

Perhaps:

> twisted.names.common.ResolverBase.getHostByName now uses A and AAAA queries instead of ANY queries.

src/twisted/names/common.py
```diff
@@ -153,13 +153,21 @@ def lookupAllRecords(self, name, timeout=None):
...
-        # XXX - this should do A and AAAA lookups, not ANY (see RFC 8482).
-        # https://twistedmatrix.com/trac/ticket/9691
-        d = self.lookupAllRecords(name, timeout)
-        d.addCallback(self._cbRecords, name, effort)
+        d = self.lookupAddress(name, timeout)
+        d.addCallback(self._cbRecordsA, name, effort, timeout)
         return d
 
-    def _cbRecords(self, records, name, effort):
+    def _cbRecordsA(self, records, name, effort, timeout):
```
exarkun, line 163: A docstring for this method would be nice.

src/twisted/names/common.py
```diff
@@ -153,13 +153,21 @@ def lookupAllRecords(self, name, timeout=None):
...
+        (ans, auth, add) = records
+        result = extractRecord(self, dns.Name(name), ans + auth + add, effort)
+        if not result:
+            # if A lookup doesn't provide IP then lookup AAAA
+            d = self.lookupIPV6Address(name, timeout)
+            d.addCallback(self._cbRecordsAAAA, name, effort)
+            return d
+        return result
+
+    def _cbRecordsAAAA(self, records, name, effort):
```
exarkun, line 173: A docstring for this method would be nice.

exarkun (changes requested):
Thanks.  This looks like a good start.  I've left some comments inline.  Additionally, it would be good to have a test for the A and AAAA codepaths here.

I see that `getHostByName` does have test coverage now but whatever existing tests there are were passing before so they're not proving the use of A/AAAA.

I wonder if `twisted.names.test.test_client.FakeResolver` could just go away, removing the now-uncommon behavior of guaranteeing an A response to an ANY query.

Thanks again for your efforts so far.

src/twisted/names/common.py
```diff
@@ -153,13 +153,21 @@ def lookupAllRecords(self, name, timeout=None):
...
+        d = self.lookupAddress(name, timeout)
+        d.addCallback(self._cbRecordsA, name, effort, timeout)
         return d
 
-    def _cbRecords(self, records, name, effort):
+    def _cbRecordsA(self, records, name, effort, timeout):
+        (ans, auth, add) = records
+        result = extractRecord(self, dns.Name(name), ans + auth + add, effort)
+        if not result:
+            # if A lookup doesn't provide IP then lookup AAAA
```
ShadowJonathan, line 167: I think it should be wise to lookup AAAA records regardless of A results, it makes IPv6 stand on even field as IPv4, as perceived DNS requests (A vs AAAA) would even out, and it could help further adoption of IPv6

ShadowJonathan (changes requested):
Small objection, but otherwise, looks fine

src/twisted/names/common.py
exarkun: This seems like it might make a good follow-up.

ShadowJonathan: What do you mean by that? Make another PR which applies this suggestion?

exarkun: File a ticket that describes the problem or the improved desired behavior and then make a PR that resolves it, yes.

wiml: I agree with @ShadowJonathan that it should make queries for both A and AAAA by default, probably in parallel. It might also be worth making this behavior configurable for hosts that are ipv4-only, That seems like a good thing to put into a separate ticket+PR unless @taroved feels the need to do it in this one.

src/twisted/names/newsfragments/10062.bugfix
wiml:
```suggestion
twisted.names.common.ResolverBase.getHostByName now makes specific queries for A and AAAA records instead of an ANY query.
```

src/twisted/names/common.py
taroved: @wiml I don't know how to do it in parallel. I think that it good thing to put into a separate ticket+PR

PR #1490, opened 2020-12-07
[10027] Add Python 3.9 for Windows and macOS on Azure pipelines.

adiroiban:
This add py 3.9 test runs to make sure we support this version of Python on Windows and macOS.

So far, only Linux 3.9 is executed in GitHub Actions.

This branch continues to uses Azure for Windows and MacOS.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10027


adiroiban:
I think that this is ready for review... py3.9 should be ok... but I need to wait a lot for all tests to be green.

altendky (approved):
Checking the build I see a macOS 3.9 job and a Windows 3.9 job for each of select and iocp reactors.  Each job looks to be reporting 3.9 as desired in the tests.  I confirmed that `twisted-iocpsupport` does have a released wheel for 3.9 on Windows for both 32 and 64 bit.  I am not weighing in on the skipping of the tests but that is pre-existing and this just extends to 3.9 which seems however reasonable the original skip is.

It'll be nice to get caught up with Python!  Thanks.

PR #1491, opened 2020-12-07
[10063] Update versions of mypy and co to work with pip.

adiroiban:
This is an updated of mypy so that there are no conflicts between pinned version.

It updates the mypy version as I guess we will want to upgrade at some point.

It updates the twisted own code to not trigger any mypy errors.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10063

### DIFF
--- a/src/twisted/application/twist/_twist.py
+++ b/src/twisted/application/twist/_twist.py
@@ -102,8 +102,12 @@ class Twist:
         options = cls.options(argv)
 
         reactor = options["reactor"]
+        # If subCommand is None, TwistOptions.parseOptions() raises UsageError
+        # and Twist.options() will exit the runner, so we'll never get here.
+        subCommand = options.subCommand
+        assert subCommand is not None
         service = cls.service(
-            plugin=options.plugins[options.subCommand],
+            plugin=options.plugins[subCommand],
             options=options.subOptions,
         )
 
--- a/src/twisted/application/twist/test/test_options.py
+++ b/src/twisted/application/twist/test/test_options.py
@@ -47,7 +47,7 @@ class OptionsTests(twisted.trial.unittest.TestCase):
         """
         self.opened = []  # type: List[Tuple[str, Optional[str]]]
 
-        def fakeOpen(name: str, mode: Optional[str] = None) -> NotImplemented:
+        def fakeOpen(name: str, mode: Optional[str] = None) -> Exception:
             if name == "nocanopen":
                 raise OSError(None, None, name)
 
--- a/src/twisted/application/twist/test/test_options.py
+++ b/src/twisted/application/twist/test/test_options.py
@@ -47,7 +47,7 @@ class OptionsTests(twisted.trial.unittest.TestCase):
         """
         self.opened = []  # type: List[Tuple[str, Optional[str]]]
 
-        def fakeOpen(name: str, mode: Optional[str] = None) -> Exception:
+        def fakeOpen(name: str, mode: Optional[str] = None) -> object:
             if name == "nocanopen":
                 raise OSError(None, None, name)
 
--- a/src/twisted/application/twist/test/test_twist.py
+++ b/src/twisted/application/twist/test/test_twist.py
@@ -108,8 +108,10 @@ class TwistTests(twisted.trial.unittest.TestCase):
         options = Twist.options(["twist", "web"])
 
         reactor = options["reactor"]
+        subCommand = options.subCommand
+        assert subCommand is not None
         service = Twist.service(
-            plugin=options.plugins[options.subCommand],
+            plugin=options.plugins[subCommand],
             options=options.subOptions,
         )
 
--- a/src/twisted/python/runtime.py
+++ b/src/twisted/python/runtime.py
@@ -108,7 +108,7 @@ class Platform:
         @return: C{True} if the current platform has been detected as Vista
         """
         if getattr(sys, "getwindowsversion", None) is not None:
-            return sys.getwindowsversion()[0] == 6
+            return sys.getwindowsversion()[0] == 6  # type: ignore[no-any-return, attr-defined]
         else:
             return False
 
--- a/src/twisted/python/runtime.py
+++ b/src/twisted/python/runtime.py
@@ -107,10 +107,7 @@ class Platform:
 
         @return: C{True} if the current platform has been detected as Vista
         """
-        if getattr(sys, "getwindowsversion", None) is not None:
-            return sys.getwindowsversion()[0] == 6  # type: ignore[no-any-return, attr-defined]
-        else:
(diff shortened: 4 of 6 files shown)


src/twisted/application/twist/test/test_options.py
```diff
@@ -47,7 +47,7 @@ def patchOpen(self) -> None:
         """
         self.opened = []  # type: List[Tuple[str, Optional[str]]]
 
-        def fakeOpen(name: str, mode: Optional[str] = None) -> NotImplemented:
```
graingert, line 50: this doesn't return Exception

src/twisted/python/runtime.py
```diff
@@ -108,7 +108,7 @@ def isVista(self) -> bool:
         @return: C{True} if the current platform has been detected as Vista
         """
         if getattr(sys, "getwindowsversion", None) is not None:
```
graingert:
```
windowsversion = getattr(sys, "getwindowsversion", None)
return isinstance(windowsversion, tuple) and windowsversion[0] == 6
```

src/twisted/application/twist/test/test_options.py
adiroiban: What is your suggestion for fixing this? :) thanks

src/twisted/python/runtime.py
adiroiban: Is it worth doing this `isinstance` check instead of just ignoring it ?

mthuurne: That suggestion would work around the `attr-defined` warning but not the `no-any-return` warning, right?

tox.ini
```diff
@@ -139,9 +139,9 @@ deps=https://github.com/twisted/pydoctor/archive/3f9c64829dfa040b334c9ae27c332c7
 description = run Mypy (static type checker)
 
 deps =
-    mypy==0.782
+    mypy==0.790
     mypy_extensions==0.4.3
```
mthuurne:
```suggestion
```

src/twisted/application/twist/test/test_options.py
mthuurne: I wanted to suggest `Literal[NotImplemented]`, but mypy doesn't like that either:
```
error: Parameter 1 of Literal[...] is invalid  [misc]
```

mthuurne: You could use `object` as the return type, that is both correct and will make mypy complain if the returned value is used as another type.

src/twisted/python/runtime.py
graingert: It's certainly odd to lookup the attribute twice

mthuurne:
```suggestion
        if hasattr(sys, "getwindowsversion"):
```

src/twisted/application/twist/test/test_options.py
```diff
@@ -47,7 +47,7 @@ def patchOpen(self) -> None:
...
+        def fakeOpen(name: str, mode: Optional[str] = None) -> Exception:
```
graingert:
```suggestion
        def fakeOpen(name: str, mode: Optional[str] = None) -> object:
```

https://github.com/python/mypy/issues/363

src/twisted/python/runtime.py
graingert:
```
windowsversion = getattr(sys, "getwindowsversion", None)
return bool(isinstance(windowsversion, tuple) and windowsversion[0] == 6)
```

mthuurne: I'd say inefficient, but not odd, since it happens a lot in code.

Actually, that suggestion is incorrect, since `getwindowsversion` is a function, not a tuple.

graingert: @mthuurne ah of course, I think this should typecheck:

`return sys.platform == "win32" and sys.getwindowsversion().major == 6`

adiroiban:
Thanks for the feedback.

I don't plan to continue working on this.
Feel free to commit your suggestions. I will be happy to review them.

Thanks!

src/twisted/python/runtime.py
graingert: yep: https://mypy-play.net/?mypy=latest&python=3.6&flags=strict&gist=bca7bed5e8445b6e8e9e10a1f2aec229

graingert:
```suggestion
        return sys.platform == "win32" and sys.getwindowsversion().major == 6
```

src/twisted/python/runtime.py
```diff
@@ -108,7 +108,7 @@ def isVista(self) -> bool:
...
-            return sys.getwindowsversion()[0] == 6
+            return sys.getwindowsversion()[0] == 6  # type: ignore[no-any-return, attr-defined]
         else:
             return False
```
graingert:
```suggestion
```

graingert (approved):

mthuurne (approved):
Looks good, but now we have to figure out what makes `pip` crash so we can get CI to pass.

src/twisted/application/twist/_twist.py
```diff
@@ -102,8 +102,12 @@ def main(cls, argv: Sequence[str] = sys.argv) -> None:
         options = cls.options(argv)
 
         reactor = options["reactor"]
+        # If subCommand is None, TwistOptions.parseOptions() raises UsageError
+        # and Twist.options() will exit the runner, so we'll never get here.
+        subCommand = options.subCommand
+        assert subCommand is not None
```
adiroiban, line 108: I think that this needs an inline comment to note that this line was added to make mypy happy... as if you remove it, mypy wil fail.

```suggestion
        assert subCommand is not None  # Here to help mypy.
```

src/twisted/application/twist/test/test_twist.py
```diff
@@ -108,8 +108,10 @@ def test_startService(self) -> None:
         options = Twist.options(["twist", "web"])
 
         reactor = options["reactor"]
+        subCommand = options.subCommand
+        assert subCommand is not None
```
adiroiban, line 112: Maybe inline commnent about mypy ?

adiroiban (commented):
All good. Approved. Fenomenal work :)

I think that is best to merge this as it is and fix the docs in https://github.com/twisted/twisted/pull/1492

src/twisted/application/twist/_twist.py
graingert: I think the comment is covered in 105-106, also it's very idiomatic to use an `assert x is not None` to unpack an Optional

adiroiban:
Thanks. I will merge it as it is.

PR #1494, opened 2020-12-08
 Sync with pydoctor 20.12.1 release

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10065


mthuurne:
The failed "Documentation" and "readthedocs" builds are due to a problem with `pip`'s new resolver, so unrelated to this PR.

The problem on macOS might be an actual problem in pydoctor. I'll investigate it tomorrow.

mthuurne:
pydoctor 20.12.1 should solve the problem where `Path.relative_to()` raises `ValueError` on macOS.

src/twisted/mail/imap4.py
```diff
@@ -223,8 +223,7 @@ def last(self):
         largest number in use.  Must be set before attempting to
         use the MessageSet as a container.
 
-        @raises: L{ValueError} if a largest value has already
-        been set.
+        @raises ValueError: if a largest value has already been set.
```
adiroiban, line 234: well... maybe move the docs to setter?

But I think that is best to keep all the property docs includding setter docstring inside the main getter method.

I would say that this is a defect in pydoctor, as pydoctor should observer that we have a property and allow such a docstring..

if we don't have a setter pydoctror can highlight the property as read-only  ... otherwise it will be default read/write.

src/twisted/python/test/test_release.py
```diff
@@ -377,7 +377,14 @@ def inner():
...
 
 
-@skipIf(not requireModule("pydoctor"), "Pydoctor is not present.")
+pydoctor = requireModule("pydoctor")
+
+
+@skipIf(pydoctor is None, "Pydoctor is not present.")
+@skipIf(
+    pydoctor is not None and pydoctor.__version__ < Version("pydoctor", 20, 12, 1),
+    "Pydoctor 20.12.1 or later is required.",
```
adiroiban, line 386: do we need  this extra skip?

the setup.cfg asks for a newer pydoctor, so I think that we should be safe.

setup.cfg
```diff
@@ -49,7 +49,7 @@ test =
 ; release scripts and process.
 dev_release =
     towncrier >= 17.4.0
-    pydoctor == 20.7.0
+    pydoctor >= 20.12.1; python_version >= "3.6"
```
adiroiban, line 52: maybe install for 3.7 and up.
The reasoning is that pydoctor is only used by the release manager ... or the automated release process.

We don't really need to support any python version.... and can save some seconds during the automated tests.

adiroiban (approved):
Nicely done. Thanks. Only minor comments. merge a will :)

src/twisted/mail/imap4.py
mthuurne: I think it makes sense to have separate docstrings in the source for getters and setters, but to merge them in the output. I'll move it to the setter.

I agree that the presentation of properties in pydoctor is not ideal, but it is an improvement over ignoring them completely and I didn't want to delay the release even more.

src/twisted/python/test/test_release.py
mthuurne: We need it when running on Python 3.5, where pydoctor is excluded from the `dev_release` extra, but might still be installed anyway.

setup.cfg
mthuurne: The extra isn't just for CI, so I think it should reflect the actual dependencies, not our testing policy. I'm in favor of streamlining CI, since having to wait half an hour for test results is not good. But that should be done by not asking certain tests to be run, rather than pretending they cannot be run.

Also, I don't think we'd get significant speedups from just cutting off a few cases here and there. As is true for optimization in general, we need good profiling tools first, so we can see where the bottlenecks are. Are we not running more in parallel because we don't have the capacity or because the definitions don't allow it? How much time do we spend on actual tests and how much on setting up environments?

One fundamental change we could consider is the matrix approach: we currently test with 4 Python versions (3.5 - 3.8) on 4 operating systems (Windows twice, with different reactor) on every run. It might be sufficient to pick a subset of combinations that does include every Python version and every OS, but not every possible combination. That could halve the number of runs for PRs. There could be a full matrix test run on `trunk` on a daily basis to catch the problems that only occur in specific combinations, but I expect those to be rare.

Anyway, that is a much bigger discussion that should be had on the mailinglist.

PR #1496, opened 2020-12-11
Remove use of API function pages

mthuurne:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10057


src/twisted/newsfragments/10057.doc
```diff
@@ -0,0 +1 @@
+Fix a few dozen broken links to API documentation pages.
```
adiroiban, line 1: Maybe. "The documentation was updated to use the Sphinx intersphinx extension for linking between our narrative documentation and the Twisted own API documentation."

docs/conf.py
```diff
@@ -240,7 +239,7 @@
 intersphinx_mapping = {
     "py3": ("https://docs.python.org/3", None),
     "zopeinterface": ("https://zopeinterface.readthedocs.io/en/latest", None),
-    "api": (apilinks_base_url, "../apidocs/objects.inv"),
+    "api": (api_base_url, "../apidocs/objects.inv"),
```
adiroiban, line 241: Maybe rename "api" with "twisted" ... to know that this is the twisted on API...
I am not sure... "api" looks very generic... but at the same time, this is just a random name.

adiroiban:
Todo for myself - see how link to trial's private API work.

adiroiban:
fabulous job... I will do another review later...but this looks solid work :)

adiroiban (approved):
Excellent work! This can be merged :) Thanks a lot!

src/twisted/newsfragments/10057.doc
mthuurne: Your suggestion is a much better description of the PR, but the news fragments are for the release notes and I figured that the only thing the user is going to notice is that there are fewer broken links.

docs/conf.py
mthuurne: The names where can be used in references for disambiguation, but since the API objects have unique names anyway, this isn't used.

Since this inventory only contains API objects and not narrative objects, "api" is specific about which part of Twisted it covers, although the fact that it covers Twisted in the first place is implicit.

PR #1499, opened 2020-12-28
Use FS encoding and errors for unicode args

twm:
This partially reverts 85069b5ccdc92745809a8287ab24d6127c1a75da but should have the same effect.

## Scope and purpose

The error reported [Twisted #10070](https://twistedmatrix.com/trac/ticket/10070):

```pytb
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/defer.py", line 1445, in _inlineCallbacks
    result = current_context.run(g.send, result)
  File "/buildbot/buildbot-job/build/master/buildbot/test/unit/scripts/test_start.py", line 117, in test_start_timeout_number_string
    res = yield self.runStart(start_timeout='10')
  File "/buildbot/buildbot-job/build/master/buildbot/test/unit/scripts/test_start.py", line 87, in runStart
    return getProcessOutputAndValue(sys.executable, args=args, env=env)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/utils.py", line 174, in getProcessOutputAndValue
    return _callProtocolWithDeferred(
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/utils.py", line 28, in _callProtocolWithDeferred
    reactor.spawnProcess(p, executable, (executable,) + tuple(args), env, path)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/posixbase.py", line 386, in spawnProcess
    args, env = self._checkProcessArgs(args, env)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/base.py", line 1088, in _checkProcessArgs
    _arg = argChecker(arg)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.8/site-packages/twisted/internet/base.py", line 1074, in argChecker
    arg = arg.encode(defaultEncoding)
builtins.TypeError: encode() argument 'encoding' must be str, not None
```

This points at this code: ​https://github.com/twisted/twisted/blob/c8064075a207af1fc9ce19d8f885a986bc5ab00c/src/twisted/internet/base.py#L1055-L1074

It looks like `defaultEncoding` used to default to `sys.getfilesystemencoding()`: ​https://github.com/twisted/twisted/commit/85069b5ccdc92745809a8287ab24d6127c1a75da It's also [documented](https://docs.python.org/3/library/sys.html#sys.__stdin__) to be possible for `std.stdout` to be `None` on Windows so using `sys.stdout.encoding` seems broken in the general case.

It seems like just the environment variables added for #9858 should be sufficient to solve the problem there, as they activate UTF-8 mode on all supported Python versions. However it also looks off to me that we aren't using the system-configured encoding error handler (via the sublimely named `os.getfilesystemencodeerrors()`). That's what should allow non-text stuff to round-trip

I have no idea how to add tests for this, but I think that the additional coverage added in PR #1302 may be sufficient.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10070

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -7,14 +7,16 @@ Very basic functionality for a Reactor implementation.
 """
 
 
-from abc import ABC, abstractmethod
 import builtins
-from heapq import heappush, heappop, heapify
 import socket  # needed only for sync-dns
 import sys
+import warnings
+from abc import ABC, abstractmethod
+from heapq import heapify, heappop, heappush
 from traceback import format_stack
 from types import FrameType
 from typing import (
+    TYPE_CHECKING,
     Any,
     AnyStr,
     Callable,
@@ -26,15 +28,13 @@ from typing import (
     Sequence,
     Set,
     Tuple,
-    TYPE_CHECKING,
     Union,
     cast,
 )
-import warnings
 
-from zope.interface import implementer, classImplements
+from zope.interface import classImplements, implementer
 
-from twisted.internet import fdesc, main, error, abstract, defer, threads
+from twisted.internet import abstract, defer, error, fdesc, main, threads
 from twisted.internet._resolver import (
     ComplexResolverSimplifier as _ComplexResolverSimplifier,
     GAIResolver as _GAIResolver,
@@ -42,7 +42,6 @@ from twisted.internet._resolver import (
 )
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.internet.interfaces import (
-    _ISupportsExitSignalCapturing,
     IAddress,
     IConnector,
     IDelayedCall,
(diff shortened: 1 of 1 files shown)


adiroiban:
Here is my take on this https://github.com/twisted/twisted/pull/1501  ... don't revert and try to fix and update the tests

twm:
Ooops, I forgot 3.5 is still supported.

As I look at this post-sleep I'm beginning to wonder why this `str` to `byte` coercion stuff is running on Windows at all. Isn't the environment natively UTF-16 on Windows?

rodrigc:
This patch is not correct.  This will break certain versions of Python

adiroiban:
Well, we were trying to find the right formula for Win, Linux and macOS. I am not sure if closing the PR without any comment on what would be the right fix is the best way to move this forward.

Note that there is also #1501 for which the progress was stalled as I didn't had time to setup a macOS and WIndows dev VM with all the supported python version.
My local Win and macOS VM are only at 2.7 ...but plan to upgrade them to 3.9.

PR #1500, opened 2020-12-28
Typo in mail/pb.py

NKuzmic97:
mail/pb.py
Line 64: 
if self.virtualDirectoy == '.': 

should be 

if self.virtualDirectory == '.': 

https://twistedmatrix.com/trac/ticket/10074


wsanchez:
Needs a newsfragment.
Code path remains untested.

wiml:
In fact, it doesn't look like *any* of the `t.mail.pb` code is exercised by existing tests. We should probably fix that.

PR #1505, opened 2021-01-22
Fix or work around stricter pydoctor warnings about __all__ and __doc__

mthuurne:
## Scope and purpose

Fix warnings that will be reported when upgrading to the next pydoctor release. See ticket for details.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10080
* [x] ~I have updated the automated tests.~

### DIFF
--- a/src/twisted/test/proto_helpers.py
+++ b/src/twisted/test/proto_helpers.py
@@ -10,7 +10,22 @@ instead.
 from twisted.internet import testing
 
 
-__all__ = testing.__all__
+__all__ = [
+    "AccumulatingProtocol",
+    "LineSendingProtocol",
+    "FakeDatagramTransport",
+    "StringTransport",
+    "StringTransportWithDisconnection",
+    "StringIOWithoutClosing",
+    "_FakeConnector",
+    "_FakePort",
+    "MemoryReactor",
+    "MemoryReactorClock",
+    "RaisingMemoryReactor",
+    "NonStreamingProducer",
+    "waitUntilAllDisconnected",
+    "EventLoggingObserver",
+]
 
 
 AccumulatingProtocol = testing.AccumulatingProtocol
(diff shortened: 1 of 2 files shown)


src/twisted/test/proto_helpers.py
```diff
@@ -10,7 +10,22 @@
 from twisted.internet import testing
 
 
-__all__ = testing.__all__
+__all__ = [
```
adiroiban, line 13: This looks good and I think it's the right thing to do.

For this PR, I think that it would help to add a comment that this is a snapshot of public API and we plan to deprecate this module... maybe add the link to the ticket asking for the deprecation.

Assuming that we agree do deprecate this module. 

The comment should also mention that no other member should be added to `__all__`

adiroiban (approved):
Thanks for the update. This looks good for the merge.
Only minor comments. No need for another round of reviews.

I think that `twisted.internet.testing` is the canonical place from where to import the testing helpers.

twisted.test.proto_helpers is there for legacy usage.

So I think that we should have a deprecation warning for the code imported via twisted.test.proto_helpers.

I can take care of writing the deprecation code... if we agreed to  deprecate t.test.proto_helpers.

mthuurne:
> I think that `twisted.internet.testing` is the canonical place from where to import the testing helpers.
> 
> twisted.test.proto_helpers is there for legacy usage.
> 
> So I think that we should have a deprecation warning for the code imported via twisted.test.proto_helpers.
> 
> I can take care of writing the deprecation code... if we agreed to deprecate t.test.proto_helpers.

The `twisted.test.proto_helpers` module is already deprecated, see 4cade8bb. The deprecation code is located in [`src/twisted/test/__init__.py`](https://github.com/twisted/twisted/blob/trunk/src/twisted/test/__init__.py) rather than in the deprecated module itself.

adiroiban:
Thanks for the followup. Then this can be merged :)

PR #1506, opened 2021-01-29
make the types on defer.py more precise

glyph:
In lieu of trying to describe this blind in PR feedback, I wrote some types. You can add this to your branch if you like it.

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -32,7 +32,7 @@ from typing import (
     Union,
     cast,
 )
-from typing_extensions import Literal, Protocol as _Protocol
+from typing_extensions import Literal
 import warnings
 
 import attr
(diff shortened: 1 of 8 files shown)


glyph:
this is an add-on to #1448

src/twisted/internet/defer.py
```diff
@@ -1505,14 +1550,13 @@ class _CancellationStatus:
...
-    CoroutineType, Generator[Deferred[object], object, None]
-]
-
-
 @_extraneous
 def _inlineCallbacks(
-    result: object, gen: _InlineCallbacksGenerator, status: _CancellationStatus
+    result: object,
+    gen: Union[
+        Generator[Deferred[_T], object, None], Coroutine[Deferred[_T], object, None]
```
mstojcevich: Is None correct for the generator/coroutine return type for these? I think it should be Optional[object] since this handles pulling the value off of StopIteration(x) when the generator returns x.

And I don't think YieldType is necessarily a Deferred when gen is a Generator.

I think the same concerns apply to the original PR fwiw.

wsanchez:
OK all of this is now applied to https://github.com/twisted/twisted/tree/10017-mypy-defer

PR #1511, opened 2021-02-09
[10086] Constant compare time for check_password

giadram:
## Scope and purpose
To use constant compare time for hashes/passwords

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10086


markrwilliams:
Hi! Thanks for contributing a security fix to Twisted!

It looks like there are two test failures because those tests are passing `str` credentials instead of `bytes` to `checkPassword` as documented by the relevant interfaces ([1](https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/cred/credentials.py#L74), [2](https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/cred/credentials.py#L95-L98))

Unfortunately it looks like we never documented [`IUsernameDigestHash`](https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/cred/credentials.py#L48) as accepting `bytes`, even though it practically _should_ ( [`calcResponse` returns `bytes` so all comparisons to `str` will be False!](https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/cred/_digest.py#L108-L133) )

Given that the tests passed with `str` objects and we didn't fully document these interfaces, I'm pretty sure there are users out there who are using `str` credentials and who will be mad about this change.

I think there are a couple of steps to move forward:
1) You (or me, or someone else with commit access to this PR) can fix those two tests - practically that means replacing `"fake"` with `b"fake"` in these two places:
https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/conch/test/test_tap.py#L119
https://github.com/twisted/twisted/blob/ba00e96f3ba99c011fbec9882b34e7790e888bc9/src/twisted/test/test_ftp_options.py#L60
2) You (or me, or someone else with commit access to this PR) also adds a deprecation warning to the modified code telling users they can no longer pass `str` credentials to `checkPassword` or `checkHash`, and that calls `theStr.encode('charmap')` on the offending string(s) prior to passing them to `hmac.compare_digest`.
3) You (or me, or someone else with commit access to this PR) also updates the `IUsernameDigestHash.checkHash` documentation to point out that it requires `byte` objects.

I know this is a lot to ask! If you'd prefer not to do that work, I totally understand and can take over this PR to make those changes over the next few days.  I'll be sure to include your original commit and credit you in the news entry.

If you're up for it, the Twisted contributors and users would _deeply_ appreciate it and thank you even more profusely :)

exarkun:
See also https://twistedmatrix.com/trac/ticket/4536

giadram:
@markrwilliams ouch... "strings and bytes are not the same anymore" again :(

hmac.compare_digest doesn't care if its arguments are strings or bytes, only for both being of the same type.

Not sure what would be the best way to keep it as much backward compatible as possible...
If args are different types, then decode bytes' one as 'ascii' (or other way around -- encode string to bytes assuming ascii)?

markrwilliams:
> @markrwilliams ouch... "strings and bytes are not the same anymore" again :(

@giadram Yeah :( If they're both `str`s, we can `charmap` them both, and if one is and the other isn't, we can return `False`, since that's what's actually happening today anyway. I _think_ that's fully compatible, at least.

giadram:
Oh.. cool... taking into account 4536, maybe it would make more sense to just close this one as a duplicate.

My rationale to move away from '==' is more social than technical.  
When/If the code which uses twisted is audited I don't want to argue with an auditor about it and I will not have a benefit of saying "prove it!".

alex:
FWIW, my opinion is that this patch opts for the straight forward and correct approach to solving this, I don't think closing this as a dupe of something last touched 4 years ago is the best route.

exarkun:
> FWIW, my opinion is that this patch opts for the straight forward and correct approach to solving this, I don't think closing this as a dupe of something last touched 4 years ago is the best route.

Is there a difference in the issue being reported?  "This ticket is newer than that ticket" doesn't seem like a good reason to ignore the old ticket and press on.

Note this has nothing to do with whether this PR should be merged or not, although I think the branch associated with the older ticket addresses more than this PR does - dealing with bytes vs unicode and mismatched lengths and such.  It's really just about tracker cleanliness.

Of course, if the underlying issues are different, then certainly it would be a mistake to mark the newer ticket as a duplicate of the older.

graingert:
> calls `theStr.encode('charmap')` on the offending string(s) prior to passing them to `hmac.compare_digest`.

also worth bearing in mind that `hmac.compare_digest` only supports ascii str
```
>>> 'caf\N{LATIN SMALL LETTER E WITH ACUTE}' == 'caf\N{LATIN SMALL LETTER E WITH ACUTE}'
True
>>> hmac.compare_digest('caf\N{LATIN SMALL LETTER E WITH ACUTE}', 'caf\N{LATIN SMALL LETTER E WITH ACUTE}')
TypeError: comparing strings with non-ASCII characters is not supported
```

~so I think you need `theStr.encode('ascii', errors='backslashreplace')` but I don't know the security implications of that~

`theStr.encode("utf8", "surrogateescape")` is probable most compatible

giadram:
@graingert  I guess it could be better to forcefully encode strings to bytes with 'utf-8' then

exarkun:
https://github.com/twisted/twisted/compare/password-comparison-4536-2#diff-d7d7a89998bf4a38ce60f8b2a3ee76a699351662a0cc902f09b1bc96265fea4bR968 does some things to handle unicode.

adiroiban:
see also https://github.com/twisted/twisted/issues/4536  JP already commented,  but I am re-commenting to create GitHub  UI links

glyph (commented):
Sadly this now requires conflicts resolved, so I am just kicking it out of the review queue until that can happen. @giadram do you still have interest in pursuing this?

src/twisted/cred/credentials.py
```diff
@@ -466,6 +466,9 @@ def __init__(self, username: bytes, password: bytes) -> None:
         self.username = username
         self.password = password
 
+    def checkPassword(self, password: bytes) -> bool:
+        return hmac.compare_digest(self.password, password)
+
     def checkPassword(self, password: bytes) -> bool:
         return self.password == password
```
glyph:
```suggestion
    def checkPassword(self, password: bytes) -> bool:
        return hmac.compare_digest(self.password, password)
```

glyph: Logical conflict didn't result in physical conflict.

glyph (dismissed):

glyph:
please review

glyph (changes requested):
Someone needs to add a news fragment to turn the tests green.

PR #1512, opened 2021-02-11
Fix NoneType encode error when multipart body part does not include h…

vtapia:
…eaders

## Scope and purpose

Before commit ​https://github.com/python/cpython/commit/cc3fa204d357be5fafc10eb8c2a80fe0bca998f1 in cpython, cgi.parse_multipart ignored parts without a name defined in "content-disposition" (or parts without headers for that matter) but after 3.7+, to make it consistent with FieldStorage, the return of the function can now contain NoneType keys, which fail to encode.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10084
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard]
* [x] I have created a newsfragment in src/twisted/newsfragments/10084.bugfix

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -980,6 +980,7 @@ class Request:
                                     for z in y
                                 ]
                                 for x, y in cgiArgs.items()
+                                if isinstance(x, str)
                             }
                         )
                     else:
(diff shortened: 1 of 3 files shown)


twm (approved):
I see the `cgi.parse_multipart` debacle continues... thank you for the patch, it looks great!

PR #1516, opened 2021-02-15
Support and test PyPy 3.7

altendky:
## Scope and purpose

This adds PyPy 3.7 to the test matrix and addresses a false positive `returnValue()` `DeprecationWarning` when running in PyPy 3.7.  This showed up originally in pytest-twisted tests.

https://twistedmatrix.com/pipermail/twisted-python/2021-February/065456.html

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10093

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,7 +15,7 @@ jobs:
       TOXENV: "alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push"
     strategy:
       matrix:
-          python-version: [3.5, 3.6, 3.7, 3.8, 3.9, pypy3]
+          python-version: [3.5, 3.6, 3.7, 3.8, 3.9, pypy-3.6, pypy-3.7]
 
     steps:
     - uses: actions/checkout@v2
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1462,7 +1462,7 @@ def _inlineCallbacks(result, g, status):
             if version_info < (3, 7):
                 # The contextvars backport and our no-op shim add an extra frame.
                 appCodeTrace = appCodeTrace.tb_next
-            elif implementation.name == 'pypy':
+            elif implementation.name == "pypy":
                 # PyPy as of 3.7 adds an extra frame.
                 appCodeTrace = appCodeTrace.tb_next
 
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -22,7 +22,7 @@ import types
 import warnings
 from asyncio import iscoroutine
 from functools import wraps
-from sys import exc_info, version_info
+from sys import exc_info, implementation, version_info
 from typing import Optional
 
 import attr
@@ -1459,8 +1459,11 @@ def _inlineCallbacks(result, g, status):
             # code.
             appCodeTrace = exc_info()[2].tb_next
 
-            # The contextvars backport and our no-op shim add an extra frame.
-            if version_info < (3, 7):
+            if implementation.name == 'pypy' and version_info >= (3, 7):
+                # PyPy as of 3.7 adds an extra frame.
+                appCodeTrace = appCodeTrace.tb_next
+            elif version_info < (3, 7):
+                # The contextvars backport and our no-op shim add an extra frame.
                 appCodeTrace = appCodeTrace.tb_next
 
             if isFailure:
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1459,11 +1459,11 @@ def _inlineCallbacks(result, g, status):
             # code.
             appCodeTrace = exc_info()[2].tb_next
 
-            if implementation.name == 'pypy' and version_info >= (3, 7):
-                # PyPy as of 3.7 adds an extra frame.
-                appCodeTrace = appCodeTrace.tb_next
-            elif version_info < (3, 7):
+            if version_info < (3, 7):
                 # The contextvars backport and our no-op shim add an extra frame.
+                appCodeTrace = appCodeTrace.tb_next
+            elif implementation.name == 'pypy':
+                # PyPy as of 3.7 adds an extra frame.
                 appCodeTrace = appCodeTrace.tb_next
 
             if isFailure:
--- a/azure-pipelines/run_test_steps.yml
+++ b/azure-pipelines/run_test_steps.yml
@@ -14,7 +14,6 @@ parameters:
   - '3.7'
   - '3.8'
   - '3.9'
-  - 'pypy3'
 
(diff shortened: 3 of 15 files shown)


altendky:
Reverted out testing with PyPy (3.6) on macOS and Windows due to https://github.com/pyca/cryptography/issues/5696 and https://github.com/tox-dev/tox/issues/1850.

.github/workflows/test.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       TOXENV: "alldeps-withcov-posix,coverage-prepare,codecov-push,coveralls-push"
     strategy:
       matrix:
-          python-version: [3.5, 3.6, 3.7, 3.8, 3.9, pypy3]
+          python-version: [3.5, 3.6, 3.7, 3.8, 3.9, pypy-3.6, pypy-3.7]
```
rodrigc, line 18: You should keep pypy3 here, in addition to adding pypy-3.6 and pypy-3.7.

Then after this release, we can remove pypy3 and keep pypy-3.6+

src/twisted/internet/defer.py
```diff
@@ -1459,8 +1459,11 @@ def _inlineCallbacks(result, g, status):
             # code.
             appCodeTrace = exc_info()[2].tb_next
 
-            # The contextvars backport and our no-op shim add an extra frame.
             if version_info < (3, 7):
+                # The contextvars backport and our no-op shim add an extra frame.
+                appCodeTrace = appCodeTrace.tb_next
+            elif implementation.name == "pypy":
```
rodrigc, line 1465: Looks like pypy 3.5, pypy 3.6, and pypy 3.7 behave differently.
Let's have this logic work on all three.  Then after this release we can drop pypy 3.5

rodrigc (changes requested):

src/twisted/internet/defer.py
altendky: Current PyPy doesn't support 3.5.  Do we need to add it into the mix?  I'm not sure offhand if the action provides it but we can try.

.github/workflows/test.yaml
altendky: It seems that `pypy3` refers to 3.6.  What do we need to keep it around for?

https://github.com/twisted/twisted/runs/1905976630?check_suite_focus=true
image

rodrigc: Yeah looks like specifying `pypy3` in GitHub actions installs pypy3 compatible with Python 3.6.12.
So I think your change is fine, not worth fighting with GitHub about this.

src/twisted/internet/defer.py
rodrigc: Looks like https://github.com/actions/setup-python#available-versions-of-pypy
specifies the versions of Pypy.  I would say going through hoops to get Pypy 3.5  in Twisted CI is not worth the effort at this point.

Since we we specified `pypy3` before, and that was installing pypy3 compatible with Python 3.6.12, then I see no point in introducing Pypy 3.5 at this point in the release cycle, when we are moving to drop support for Python 3.5/Ppy 3.5 after the release.

rodrigc: Can you manually install the last Pypy 3.5 release from:
https://downloads.python.org/pypy/pypy3.5-v7.0.0-linux64.tar.bz2
https://downloads.python.org/pypy/pypy3.5-v7.0.0-osx64.tar.bz2

And manually run `test_inlinecb.py` in your environment with your patch with this version of pypy?

I just want to double check, just in case there is a remote possibility that someone out there is running Pypy 3.5

altendky:
```console
$ git rev-parse HEAD
a225560007f04235596992da561adcf686c118ab
```

```console
$ git diff
diff --git a/src/twisted/logger/_buffer.py b/src/twisted/logger/_buffer.py
index f90f8f982..3334f97e9 100644
--- a/src/twisted/logger/_buffer.py
+++ b/src/twisted/logger/_buffer.py
@@ -7,7 +7,7 @@ Log observer that maintains a buffer.
 """
 
 from collections import deque
-from typing import Deque, Optional
+from typing import Optional
 
```

```console
$ venvpp35/bin/python --version --version
Python 3.5.3 (928a4f70d3de7d17449456946154c5da6e600162, Feb 09 2019, 11:50:43)
[PyPy 7.0.0 with GCC 8.2.0]
```

```console
$ venvpp35/bin/trial twisted.internet.test.test_inlinecb
twisted.internet.test.test_inlinecb
  CancellationTests
    test_asynchronousCancellation ...                                      [OK]
    test_cascadeCancellingOnCancel ...                                     [OK]
    test_errbackCancelledErrorOnCancel ...                                 [OK]
    test_errorToErrorTranslation ...                                       [OK]
    test_errorToSuccessTranslation ...                                     [OK]
  ForwardTraceBackTests
    test_forwardLotsOfTracebacks ...                                       [OK]
    test_forwardTracebacks ...                                             [OK]
  NonLocalExitTests
    test_returnValueNonLocalDeferred ...                                   [OK]
    test_returnValueNonLocalWarning ...                                    [OK]
  StopIterationReturnTests
    test_returnWithValue ...                                               [OK]

Ran 10 tests in 0.135s

PASSED (successes=10)
```

altendky: Also full tox passed.

https://gist.github.com/altendky/8fab0cac9a0be55f2d836bcc018c6637

rodrigc: That's good enough for me.
No need change the CI to add Pypy 3.5.  If someone encounters a problem, we can point them to this
log, and ask them to upgrade to Pypy 3.6 which is better tested/supported.

.github/workflows/test.yaml
rodrigc: This change is fine.

rodrigc (approved):

src/twisted/internet/defer.py
```diff
@@ -1459,8 +1459,11 @@ def _inlineCallbacks(result, g, status):
             # code.
             appCodeTrace = exc_info()[2].tb_next
 
-            # The contextvars backport and our no-op shim add an extra frame.
             if version_info < (3, 7):
+                # The contextvars backport and our no-op shim add an extra frame.
+                appCodeTrace = appCodeTrace.tb_next
```
rodrigc, line 1464: so is the logic basically like:

```
 if (implementation.name == "cpython" and version_info < (3, 7)) or
     (implementation.name == "pypy" and version_info > (3, 6):
                # The contextvars backport and our no-op shim add an extra frame.
                # PyPy as of 3.7 adds an extra frame.
                appCodeTrace = appCodeTrace.tb_next
 ```

PR #1522, opened 2021-02-23
Stop using the deprecated int_from_bytes from cryptography.

clokep:
## Scope and purpose

cryptography 3.4.0 deprecated using `cryptography.utils.int_from_bytes` and suggests replacing it with `int.from_bytes` instead (which has the exact same signature). According to [the Python docs](https://docs.python.org/3/library/stdtypes.html#int.from_bytes) `int.from_bytes` has existed since Python 3.2, so should be safe to use in Twisted.

Looking at the deprecation PR (pyca/cryptography#5609) it was just passing through anyway, so this should just save an import and such and have no noticeable change.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10097
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles)) --> I put it in `src/twisted/conch/newsfragments/`, maybe that's not correct though.
* [x] I have updated the automated tests. --> Should be covered by the current conch tests.

### DIFF
--- a/src/twisted/conch/ssh/common.py
+++ b/src/twisted/conch/ssh/common.py
@@ -11,7 +11,7 @@ Maintainer: Paul Swartz
 
 import struct
 
-from cryptography.utils import int_from_bytes, int_to_bytes
+from cryptography.utils import int_to_bytes
 
 from twisted.python.deprecate import deprecated
 from twisted.python.versions import Version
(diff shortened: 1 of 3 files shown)


clokep:
I believe I've updated the trac ticket correctly. Let me know if any changes are necessary!

graingert (approved):

twm (approved):
It looks like you got them all, thanks!

clokep:
Thanks for double checking (and merging)!

PR #1407, opened 2020-09-27
Convert some Travis jobs to GitHub Actions

rodrigc:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9981
* [ ] I ran `tox -e black-reformat` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [ ] I ran additional checks listed at [Getting Your Patch Accepted](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [ ] I have updated the automated tests.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.

PR #1533, opened 2021-03-01
[10109] Add nodeps and  fix no-ipv6 job.

adiroiban:
## Scope and purpose

This adds the nodeps to the CI matrix.

I tried to refactor the GHA job to reuse as much as possible from the code...

The result is a bit harder to read...so I am ok to go back to code duplication in test.yaml file.

The `nodeps` run ironical, still has `coverage` as a dependency so that we can report the coverage and make sure we cover the extra code branches.

------

I am trying to reduce the size of the matrix so I went with this:

* latest 3.7 is used for non-ipv6 tests
* latest 3.6 used for nondeps tests
* minumim 3.6 and the rest are used for with ipv6 and with all deps tests

------

As a drive by, it fixes the no IPV6 tests as there were not working...
They were not working at lease since they were migrated to GHA - https://github.com/twisted/twisted/pull/1407/files
I have reviewed that PR but I have no idea why I have approved it :( ... I am trying to fix that damage here.

See coverage report which shows that ipv6 skip was never reached

https://codecov.io/gh/twisted/twisted/src/trunk/src/twisted/internet/test/test_tcp.py#L111

I have added a better description to ipv6 skip reason, as it was not that easy to understand what was going on there.

-------

As a drive by, I have used shorter names for GHA job names as with  older names it was harder to see what is failing

image

## How to test

For no-IPV6 job.
Check that loopback ipv6 is down 
https://github.com/twisted/twisted/pull/1533/checks?check_run_id=2008044713#step:6:14

Then check the logs and see that tests are skipped due to

```
Does not work on systems without IPv6 support.
and
IPv6 not available. [Errno 99] Cannot assign requested address
```

At least the diff from this PR has coverage  https://app.codecov.io/gh/twisted/twisted/compare/1533/diff
so it looks like noipv6 code is triggerred.

------

For nodeps check... check the logs and you should see a lot of skipped test.
In theory, the coverage report should see that non SSL code branches are also covered now

For trunk https://codecov.io/gh/twisted/twisted/src/trunk/src/twisted/internet/tcp.py#L42

And check the similar link for this branch... 

https://app.codecov.io/gh/twisted/twisted/compare/1533/tree/src/twisted/internet/tcp.py#D1L43

--------

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10109


adiroiban:
I am happy with the current version... so this is ready for review. Thanks!

.github/workflows/test.yaml
```diff
@@ -54,67 +60,23 @@ jobs:
...
-        python -c "print('\nENVIRONMENT VARIABLES\n=====================\n\n')"
-        python -c "import os; [ print(e,v) for (e,v) in os.environ.items() ]"
+        sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1
+        sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1
+        sudo sysctl -w net.ipv6.conf.lo.disable_ipv6=1
+        ip a
 
+    - uses: twisted/python-info-action@v1
     - name: Install dependencies
+      shell: bash
```
altendky: Perhaps we'd like this as a global default?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun
```yaml
defaults:
  run:
    shell: bash
```

.github/workflows/test.yaml
```diff
@@ -12,14 +14,28 @@ jobs:
...
-          # Run on the minimum micro Python version that we can get on CI.
-          # When updating the minimum Python version here, also update the
-          # `python_requires` from `setup.cfg`.
-          # Run on latest minor release of each major python version.
-          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, pypy-3.6, pypy-3.7]
+        # Run on the minimum micro Python version that we can get on CI.
+        # When updating the minimum Python version here, also update the
+        # `python_requires` from `setup.cfg`.
+        # Run on latest minor release of each major python version.
+        python-version: ["3.6.7", 3.8, 3.9, pypy-3.6, pypy-3.7]
```
altendky: I prefer 'complete' matrices...  but I haven't been dealing with Twisted lately so I don't know how painful the builds are.  For what is being done, it seems to be done well.

altendky (approved):
I'll admit I'm not sure why this passed while https://github.com/twisted/twisted/commit/85ba7bd1cb7bfb6b439f032ad4d981621f9ae6d5 failed.  So, either my review ain't great because this is meant to fix that failure in PyPy 3.6...  or that failure ought to be fixed before merging this.

Also, can't say I'm a fan that we seem to now be allowing out-of-date PRs to be merged?  As in, I don't see the expected up-to-date confirmation nor complaint on this PR.

Thanks for getting some coverage back.  `:]`

.github/workflows/test.yaml
```diff
@@ -12,14 +14,28 @@ jobs:
...
+        tox-env: [alldeps]
+        noipv6: ['']
+
+        include:
+          # Run nodeps tests and add some extra coverage for latest 3.6 micro
+          # version.
+          - python-version: 3.6
+            tox-env: nodeps
+          # Run nodeps tests and add some extra coverage for latest 3.6 micro
```
```suggestion
          # Run nodeps tests and add some extra coverage for latest 3.7 micro
```

.github/workflows/test.yaml
```diff
@@ -22,18 +28,21 @@ jobs:
...
         tox-env: [alldeps]
         noipv6: ['']
 
         include:
-          # Run nodeps tests and add some extra coverage for latest 3.6 micro
-          # version.
+
+          # `nodeps` is created to make sure we don't have import errors in the
+          # runtime and production code.
+          # Any supported Python version is OK for this job.
```
altendky:
```suggestion
          # The minimum supported Python version should be used to maximize
          # coverage of code that otherwise depends on backports.
```

.github/workflows/test.yaml
```diff
@@ -22,18 +28,21 @@ jobs:
...
           - python-version: 3.6
             tox-env: nodeps
-          # Run nodeps tests and add some extra coverage for latest 3.6 micro
-          # version.
-          - python-version: 3.7
+          # `noipv6` is created to make sure all is OK on an OS which don't
```
altendky:
```suggestion
          # `noipv6` is created to make sure all is OK on an OS which doesn't
```

admin/dump_all_version_info.py
```diff
@@ -1,7 +1,19 @@
...
 print(sys.executable)
 print(sys.version)
 print(sys.platform)
+print("::endgroup::")
+
+if os.environ.get("CI", "").lower() == "true":
+    # On CI show the exact deps found at runtime.
+    # Skipped on local dev to speed up the test execution.
+    print("::group::Deps info")
+    main(["freeze", "--all"])
```
altendky: Just subprocess instead of using the private Python API?  Easy enough with `sys.executable`.

.github/workflows/test.yaml
```diff
@@ -22,18 +28,21 @@ jobs:
         # When updating the minimum Python version here, also update the
         # `python_requires` from `setup.cfg`.
         # Run on latest minor release of each major python version.
-        python-version: ["3.6.7", 3.8, 3.9, pypy-3.6, pypy-3.7]
+        python-version: [3.6.7, 3.7, 3.8, 3.9, 3.10.0-alpha.5, pypy-3.6, pypy-3.7]
```
altendky: This probably ought to be a separate PR, but since I already mentioned it in IRC I ought to go ahead and put it here too...  Anyways, with the incantation below you can avoid having to update Python versions from alpha through final release.

https://github.com/pytest-dev/pytest-twisted/blob/5b1fa21b5d92ce1b32310dd0ad2010e0232be017/.github/workflows/ci.yml#L149
```yaml
        uses: actions/setup-python@v2
        with:
          python-version: '${{ matrix.python.action }}.0-alpha - ${{ matrix.python.action }}.X'
```

altendky (changes requested):

.github/workflows/test.yaml
adiroiban: @altendky I am happy with your suggest and I think that we should only pin the minimum python micro version that we support. For all other version we should get automatic updates.

Please check with @graingert to see if we can get consensus ...but I guess that a new PR would provide the right place to discuss this and take a decision.

--------

I guess that is a question between CI stability and real life usage.

I expect that most people will use in real life the latest python micro release. 
We want to run our tests in a CI environment as close as possible to that one.

Now, automatic updates might break the CI ... but then we will have a CI that is solid green, yet Twisted fails in real life :)

admin/dump_all_version_info.py
adiroiban: I don't know why I don't like forking another process... I went for the entry_points API ... I hope this is considered public API :)

altendky: Welp, I double checked that the entry point doesn't `sys.exit()` itself (at least not on success) so we do get the close tag for the group...  Hopefully they don't add another `console_script`.

.github/workflows/test.yaml
altendky: For completeness, here's an updated incantation that accounts for that incantation not working with PyPy.  Yeah, since we're calling out a minimum patch version, that'd be an extra special bit to deal with.  Aside from that, all this does is save you from having to bump from alpha to beta to rc to release.

https://github.com/pytest-dev/pytest-twisted/pull/135/files
```yaml
          # This allows the matrix to specify just the major.minor version while still
          # expanding it to get the latest patch version including alpha releases.
          # This avoids the need to update for each new alpha, beta, release candidate,
          # and then finally an actual release version.  actions/setup-python doesn't
          # support this for PyPy presently so we get no help there.
          # 
          # CPython -> 3.9.0-alpha - 3.9.X
          # PyPy    -> pypy-3.7
          python-version: ${{ fromJSON(format('["{0}", "{1}"]', format('{0}.0-alpha - {0}.X', matrix.python.action), matrix.python.action))[startsWith(matrix.python.action, 'pypy')] }}
```

altendky:
I don't understand the build failure...  otherwise I'll sign off.

adiroiban:
The pypy runs are failing even if there are no test failures.

```
PASSED (skips=1595, expectedFailures=3, successes=10354)
ERROR: InvocationError for command /home/runner/work/twisted/twisted/.tox/alldeps-withcov-posix/bin/coverage run -p --rcfile=/home/runner/work/twisted/twisted/.coveragerc -m twisted.trial --temp-directory=/home/runner/work/twisted/twisted/.tox/alldeps-withcov-posix/tmp/_trial_temp --reactor=default --reporter=verbose twisted (exited with code 1)
```

altendky (approved):

adiroiban:
Thanks for the review.

I am going to merge this. It will be revisited soon while I check the 

I see that the nodeps coverage is reported on codecov - https://codecov.io/gh/twisted/twisted/commit/fcc29fdfcf1c0a12dbbcd88010cbca8ee004b549/build

I don't know why coveralls is reporting a decrease in coverage as the nodeps is also published to coveralls

image

PR #1535, opened 2021-03-02
support py3.10a5 (by fixing twisted.trial.unittest.TestCase.flushWarnings) 

graingert:
It took a little while from 3.9's release to Twisted supporting it, I think it's worth trying to get support available before the release.

This is *just* the GHA ubuntu-latest machines

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10111

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,7 +19,7 @@ jobs:
           # When updating the minimum Python version here, also update the
           # `python_requires` from `setup.cfg`.
           # Run on latest minor release of each major python version.
-          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, pypy-3.6, pypy-3.7]
+          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, 3.10.0-alpha.5, pypy-3.6, pypy-3.7]
 
     steps:
     - uses: actions/checkout@v2
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,6 +63,7 @@ dev_release =
 dev =
     %(dev_release)s
     pyflakes ~= 2.2
+    # TODO: support python-subunit in py3.10 https://twistedmatrix.com/trac/ticket/10115
     python-subunit ~= 1.4; python_version < "3.10"
     twistedchecker ~= 0.7
     coverage ~= 5.5
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,7 @@ dev_release =
 dev =
     %(dev_release)s
     pyflakes ~= 2.2
-    python-subunit ~= 1.4
+    python-subunit ~= 1.4; python_version < "3.10"
     twistedchecker ~= 0.7
     coverage ~= 5.5
 
--- a/src/twisted/newsfragments/10111.feature
+++ b/src/twisted/newsfragments/10111.feature
@@ -1,1 +1,1 @@
-support py3.10a5
+support py3.10a5 (by fixing twisted.trial.unittest.TestCase.flushWarnings)
--- a/src/twisted/newsfragments/10111.feature
+++ b/src/twisted/newsfragments/10111.feature
@@ -1,0 +1,1 @@
+support py3.10a5
--- a/src/twisted/newsfragments/10111.misc
+++ b/src/twisted/newsfragments/10111.misc
@@ -1,1 +1,0 @@
-test on py3.10a5
--- a/src/twisted/newsfragments/10111.misc
+++ b/src/twisted/newsfragments/10111.misc
@@ -1,0 +1,1 @@
+test on py3.10a5
--- a/src/twisted/python/deprecate.py
+++ b/src/twisted/python/deprecate.py
@@ -604,21 +604,15 @@ def warnAboutFunction(offender, warningString):
     # inspect.getmodule() is attractive, but somewhat
     # broken in Python < 2.6.  See Python bug 4845.
     offenderModule = sys.modules[offender.__module__]
-    filename = inspect.getabsfile(offenderModule)
-    lineStarts = list(findlinestarts(offender.__code__))
-    lastLineNo = lineStarts[-1][1]
-    globals = offender.__globals__
-
-    kwargs = dict(
+    warn_explicit(
+        warningString,
         category=DeprecationWarning,
-        filename=filename,
-        lineno=lastLineNo,
+        filename=inspect.getabsfile(offenderModule),
+        lineno=max(lineNumber for _, lineNumber in findlinestarts(offender.__code__)),
         module=offenderModule.__name__,
-        registry=globals.setdefault("__warningregistry__", {}),
+        registry=offender.__globals__.setdefault("__warningregistry__", {}),
         module_globals=None,
     )
-
-    warn_explicit(warningString, **kwargs)
 
 
 def _passedArgSpec(argspec, positional, keyword):
--- a/src/twisted/python/test/test_deprecate.py
+++ b/src/twisted/python/test/test_deprecate.py
@@ -377,6 +377,29 @@ def callTestFunction():
         deprecate.warnAboutFunction(testFunction, "A Warning String")
 """
         )
+        self.package.child("pep626.py").setContent(
+            b"""
+"A module string"
+
+from twisted.python import deprecate
+
+def noop():
+    pass
+
+def testFunction(a=1, b=1):
+    "A doc string"
+    if a:
+        if b:
+            noop()
+        else:
+            pass
+
+def callTestFunction():
+    b = testFunction()
+    if b is None:
+        deprecate.warnAboutFunction(testFunction, "A Warning String")
+"""
+        )
         # Python 3 doesn't accept bytes in sys.path:
         packagePath = self.package.parent().path
         sys.path.insert(0, packagePath)
@@ -425,6 +448,26 @@ def callTestFunction():
         # Line number 9 is the last line in the testFunction in the helper
         # module.
         self.assertEqual(warningsShown[0]["lineno"], 9)
+        self.assertEqual(warningsShown[0]["message"], "A Warning String")
+        self.assertEqual(len(warningsShown), 1)
+
+    def test_warningLineNumberPEP626(self):
+        """
+        L{deprecate.warnAboutFunction} emits a C{DeprecationWarning} with the
+        number of a line within the implementation of the function passed to it
+        on py3.10 with PEP 626.
+        """
+        from twisted_private_helper import pep626
+
+        pep626.callTestFunction()
+        warningsShown = self.flushWarnings()
+        self.assertSamePath(
+            FilePath(warningsShown[0]["filename"].encode("utf-8")),
+            self.package.sibling(b"twisted_private_helper").child(b"pep626.py"),
+        )
+        # Line number 15 is the last line in the testFunction in the helper
+        # module.
+        self.assertEqual(warningsShown[0]["lineno"], 15)
         self.assertEqual(warningsShown[0]["message"], "A Warning String")
         self.assertEqual(len(warningsShown), 1)
 
--- a/src/twisted/python/test/test_deprecate.py
+++ b/src/twisted/python/test/test_deprecate.py
@@ -451,11 +451,11 @@ def callTestFunction():
         self.assertEqual(warningsShown[0]["message"], "A Warning String")
         self.assertEqual(len(warningsShown), 1)
 
-    def test_warningLineNumberPEP626(self):
+    def test_warningLineNumberDisFindlinestarts(self):
         """
         L{deprecate.warnAboutFunction} emits a C{DeprecationWarning} with the
-        number of a line within the implementation of the function passed to it
-        on py3.10 with PEP 626.
+        number of a line within the implementation handling the case in which
+        dis.findlinestarts returns the lines in random order.
         """
         from twisted_private_helper import pep626
 
--- a/src/twisted/trial/test/test_warning.py
+++ b/src/twisted/trial/test/test_warning.py
@@ -365,9 +365,14 @@ def foo():
         # Flush it
         self.assertEqual(len(self.flushWarnings([module.foo])), 1)
 
-    def test_pep626(self):
+    def test_offendingFunctions_deep_branch(self):
         """
-        dis.findlinestarts is not guaranteed to be sorted ascending.
+        In Python 3.6 the dis.findlinestarts documented behaviour
+        was changed such that the reported lines might not be sorted ascending.
+        In Python 3.10 PEP 626 introduced byte-code change such that the last
+        line of a function wasn't always associated with the last byte-code.
+        In the past flushWarning was not detecting that such a function was
+        associated with any warnings.
         """
 
         def foo(a=1, b=1):
--- a/src/twisted/trial/test/test_warning.py
+++ b/src/twisted/trial/test/test_warning.py
@@ -365,6 +365,24 @@ def foo():
         # Flush it
         self.assertEqual(len(self.flushWarnings([module.foo])), 1)
 
+    def test_pep626(self):
+        """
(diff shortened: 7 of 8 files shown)


src/twisted/python/deprecate.py
```diff
@@ -604,22 +604,16 @@ def warnAboutFunction(offender, warningString):
     # inspect.getmodule() is attractive, but somewhat
```
graingert, line 604: @exarkun is `inspect.getmodule()` safe to use now?

src/twisted/python/test/test_deprecate.py
```diff
@@ -428,6 +451,26 @@ def test_warningLineNumber(self):
...
+
+        pep626.callTestFunction()
+        warningsShown = self.flushWarnings()
+        self.assertSamePath(
+            FilePath(warningsShown[0]["filename"].encode("utf-8")),
+            self.package.sibling(b"twisted_private_helper").child(b"pep626.py"),
+        )
+        # Line number 15 is the last line in the testFunction in the helper
+        # module.
+        self.assertEqual(warningsShown[0]["lineno"], 15)
```
graingert, line 470: Proof of TDD:

```python
[FAIL]
  File "/home/graingert/projects/twisted/.tox/py310-alldeps-nocov/lib/python3.10/site-packages/twisted/python/test/test_deprecate.py", line 470, in test_warningLineNumberPEP626
    self.assertEqual(warningsShown[0]["lineno"], 15)
  File "/home/graingert/projects/twisted/.tox/py310-alldeps-nocov/lib/python3.10/site-packages/twisted/trial/_synctest.py", line 424, in assertEqual
  File "/usr/lib/python3.10/unittest/case.py", line 839, in assertEqual
  File "/usr/lib/python3.10/unittest/case.py", line 832, in _baseAssertEqual
twisted.trial.unittest.FailTest: 11 != 15

twisted.python.test.test_deprecate.WarnAboutFunctionTests.test_warningLineNumberPEP626
```

src/twisted/trial/test/test_warning.py
```diff
@@ -365,6 +365,24 @@ def foo():
...
+                if b:
+                    warnings.warn("oh no")
+                else:
+                    pass
+
+        # Generate the warning
+        foo()
+
+        # Flush it
+        self.assertEqual(len(self.flushWarnings([foo])), 1)
```
graingert, line 389: Proof of TDD:
```python
[FAIL]
  File "/home/graingert/projects/twisted/.tox/py310-alldeps-nocov/lib/python3.10/site-packages/twisted/trial/test/test_warning.py", line 384, in test_pep626
    self.assertEqual(len(self.flushWarnings([foo])), 1)
  File "/home/graingert/projects/twisted/.tox/py310-alldeps-nocov/lib/python3.10/site-packages/twisted/trial/_synctest.py", line 424, in assertEqual
  File "/usr/lib/python3.10/unittest/case.py", line 839, in assertEqual
  File "/usr/lib/python3.10/unittest/case.py", line 832, in _baseAssertEqual
twisted.trial.unittest.FailTest: 0 != 1

twisted.trial.test.test_warning.FlushWarningsTests.test_pep626
```

setup.cfg
```diff
@@ -63,7 +63,7 @@ dev_release =
 dev =
     %(dev_release)s
     pyflakes ~= 2.2
-    python-subunit ~= 1.4
+    python-subunit ~= 1.4; python_version < "3.10"
```
adiroiban, line 67: I guess that python-subunit is not yet ported to 3.10 ... is there an upstream ticket  that we can follow and enable subunit on 3.10 later?

src/twisted/python/test/test_deprecate.py
```diff
@@ -375,6 +375,29 @@ def callTestFunction():
     b = testFunction()
     if b == 3:
         deprecate.warnAboutFunction(testFunction, "A Warning String")
+"""
+        )
+        self.package.child("pep626.py").setContent(
```
adiroiban, line 380: I think that is best to define it in test_warningLineNumberPEP626 ...

 or move it to a helper method and cal iit from insid the test so that it can be moved outside of setup :)

It should make the test easier to read.

We don't need pytest to stop abusing setup/teardown :p

src/twisted/python/test/test_deprecate.py
```diff
@@ -375,6 +375,29 @@ def callTestFunction():
...
+            b"""
+"A module string"
```
adiroiban, line 382: I think that this module string  can be used to document this text fixture.

Why we need pep626.py and why the previous module.py is not enough.

src/twisted/python/test/test_deprecate.py
```diff
@@ -375,6 +375,29 @@ def callTestFunction():
...
+
+from twisted.python import deprecate
+
+def noop():
```
adiroiban, line 386: is this really needed for this example?

src/twisted/python/test/test_deprecate.py
```diff
@@ -375,6 +375,29 @@ def callTestFunction():
...
+    pass
+
+def testFunction(a=1, b=1):
+    "A doc string"
+    if a:
+        if b:
+            noop()
+        else:
+            pass
```
adiroiban, line 395: Do we need a real implementation for this test?
 
```suggestion
def testFunction():
    "The target for which the deprection is raised."

```

src/twisted/python/test/test_deprecate.py
```diff
@@ -375,6 +375,29 @@ def callTestFunction():
...
+
+def callTestFunction():
+    b = testFunction()
+    if b is None:
+        deprecate.warnAboutFunction(testFunction, "A Warning String")
```
adiroiban, line 400: Would this minimal example work?

```suggestion
def callTestFunction():
    deprecate.warnAboutFunction(testFunction, "A Warning String")
    
def otherFunctionNotCalledInTest():
   """
   Here to make sure we don't accidentally just report last line of the file.
   """
   deprecate.warnAboutFunction(testFunction, "Shoulnd't be called")
   more_calls_here()
```

src/twisted/trial/test/test_warning.py
```diff
@@ -365,6 +365,24 @@ def foo():
...
+            if a:
+                if b:
+                    warnings.warn("oh no")
+                else:
+                    pass
+
+        # Generate the warning
+        foo()
+
+        # Flush it
```
adiroiban, line 388: does this comment has any value :) ?

src/twisted/python/test/test_deprecate.py
```diff
@@ -428,6 +451,26 @@ def test_warningLineNumber(self):
         self.assertEqual(warningsShown[0]["message"], "A Warning String")
         self.assertEqual(len(warningsShown), 1)
 
+    def test_warningLineNumberPEP626(self):
```
adiroiban: is this pep636 or if just about a change in API on python 3.10?

Rather than mentioning PEP636 I think that it's more useful to explain that for some code internal Python API will report line number in non ascending orders

src/twisted/trial/test/test_warning.py
```diff
@@ -365,6 +365,24 @@ def foo():
         # Flush it
         self.assertEqual(len(self.flushWarnings([module.foo])), 1)
 
+    def test_pep626(self):
+        """
+        dis.findlinestarts is not guaranteed to be sorted ascending.
```
adiroiban: Is this test for Twisted code or stdlib code ? :)

src/twisted/python/test/test_deprecate.py
```diff
@@ -428,6 +451,26 @@ def test_warningLineNumber(self):
         self.assertEqual(warningsShown[0]["message"], "A Warning String")
```
adiroiban, line 451: can we have the module.py code defined in this test... it should make the test easier to read.

But.... do we need 2 tests here?

src/twisted/python/deprecate.py
```diff
@@ -604,22 +604,16 @@ def warnAboutFunction(offender, warningString):
...
-    globals = offender.__globals__
-
-    kwargs = dict(
+    warn_explicit(
+        warningString,
         category=DeprecationWarning,
-        filename=filename,
-        lineno=lastLineNo,
+        filename=inspect.getabsfile(offenderModule),
+        lineno=max(lineNumber for _, lineNumber in findlinestarts(offender.__code__)),
```
adiroiban, line 611: I don't understand dis.findlinestarts :)
https://docs.python.org/3/library/dis.html#dis.findlinestarts

I have to get this branch and play with unittest to understand what is going on here... it all looks like magic for me.

adiroiban (commented):
Thanks for working on this.

warnAboutFunction is a nice function... but I think the docstring or test's docstring should say that it only provides an approximation for the line number.

This is screenshot from trunk with 3.9 .... function called at 140 ... report for 150

image

--------

I don't understand why we need another test for this and why the current tests are not enough.

Also, I think that the test example should contain warnAboutFunction not as the line call in the file and not as the last line in the method/function from where it called.

It might give the impression that warnAboutFunction can pinpoint the line offended line of code .

.github/workflows/test.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
           # When updating the minimum Python version here, also update the
           # `python_requires` from `setup.cfg`.
           # Run on latest minor release of each major python version.
-          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, pypy-3.6, pypy-3.7]
+          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, 3.10.0-alpha.5, pypy-3.6, pypy-3.7]
```
adiroiban, line 22: can we use just 3.10 and it will automatically pic last 3.10 version?

```suggestion
          python-version: ["3.6.7", 3.6, 3.7, 3.8, 3.9, 3.10, pypy-3.6, pypy-3.7]
```

src/twisted/newsfragments/10111.misc
```diff
@@ -0,0 +1 @@
+test on py3.10a5
```
adiroiban: Can this be a feature news fragment like announcing that 3.10 is supported on Linux?

.github/workflows/test.yaml
graingert: I'd rather bump the pre-release version manually, so we don't subject Twisted CI to the whims of every single alpha breaking change

src/twisted/python/test/test_deprecate.py
graingert: https://www.python.org/dev/peps/pep-0626/

It's 626 not 636

src/twisted/python/test/test_deprecate.py
graingert: Yes, otherwise it won't test the change

src/twisted/python/test/test_deprecate.py
graingert: I'm just copying the existing pattern

graingert:
> should say that it only provides an approximation for the line number

It's designed to find the last line number

> I don't understand why we need another test for this and why the current tests are not enough.

Because the test failure I found triggered this pep626 behaviour by accident, I then fixed that specific case, then I found all uses of dis.findlinestarts and fixed them too, so each needed a new test

src/twisted/python/test/test_deprecate.py
graingert: I'm copying the original style

src/twisted/python/test/test_deprecate.py
graingert: No, we need a specific branching structure to trigger the change in dis.findlinestarts output

src/twisted/python/test/test_deprecate.py
adiroiban: Can we have a docstring that explain why this function needs to be structured this way and why a simpler one will not help with the test?

Just by looking at the 2 modules I can't spot the difference and I would say that they are duplicate tests :)

src/twisted/python/test/test_deprecate.py
adiroiban: Understood... but in this case I think that a docstring or some comment would help. Thanks!

adiroiban (approved):
Looks good.

I would say that the testing need better documentation

Maybe the docstring of `warnAboutFunction` should inform that last line is reported.


```diff
- Issue a warning string, identifying C{offender} as the responsible code. 
+ Issue a warning string, identifying last line of C{offender} definition as the responsible code.
```

Assuming that `test_pep626` will be removed as it's not a test for Twisted, I think that you can address the existing comments and merge without requiring another review.

Thanks!

setup.cfg
graingert: @adiroiban https://github.com/testing-cabal/subunit/pull/32

graingert:
> Assuming that test_pep626 will be removed as it's not a test for Twisted, 

this is a test for twisted! twisted was previously misusing `dis.findlinestarts`

setup.cfg
adiroiban: Can we also have a comment to the link to issue here.
In this way, if someone wants to look again at enabling subunit in 3.10 they don't have to search git blame and then look at the comments from the PR. Thanks!

src/twisted/trial/test/test_warning.py
```diff
@@ -365,6 +365,24 @@ def foo():
...
+        """
+
+        def foo(a=1, b=1):
+            if a:
+                if b:
+                    warnings.warn("oh no")
```
adiroiban, line 381: Maybe add a comment here that this needs to be constructed in this way to trigger findlinestarts behaviour

src/twisted/trial/test/test_warning.py
graingert:
```suggestion
    def test_offendingFunctions_deep_branch(self):
        """
        In Python 3.6 the dis.findlinestarts documented behaviour
        was changed such that the reported lines might not be sorted ascending.
        In Python 3.10 PEP 626 introduced byte-code change such that the last
        line of a function wasn't always associated with the last byte-code.
        In the past flushWarning was not detecting that such a function was
        associated with any warnings.
```

src/twisted/python/test/test_deprecate.py
```diff
@@ -428,6 +451,26 @@ def test_warningLineNumber(self):
...
         self.assertEqual(len(warningsShown), 1)
 
+    def test_warningLineNumberPEP626(self):
+        """
+        L{deprecate.warnAboutFunction} emits a C{DeprecationWarning} with the
+        number of a line within the implementation of the function passed to it
+        on py3.10 with PEP 626.
+        """
```
adiroiban:
```suggestion
    def test_warningLineNumberDisFindlinestarts(self):
        """
        L{deprecate.warnAboutFunction} emits a C{DeprecationWarning} with the
        number of a line within the implementation handling the case in which
        dis.findlinestarts returns the lines in random order.
        """
```

adiroiban (approved):
I think that this is an improvement over what we have in trunk and it can be merged.

Feel free to address the docstring changes and the title of the PR

Thanks!

src/twisted/newsfragments/10111.feature
```diff
@@ -0,0 +1 @@
+support py3.10a5
```
graingert:
```suggestion
support py3.10a5 (by fixing twisted.trial.unittest.TestCase.flushWarnings)
```

src/twisted/newsfragments/10111.feature
```diff
@@ -0,0 +1 @@
+support py3.10a5 (by fixing twisted.trial.unittest.TestCase.flushWarnings)
```
adiroiban, line 1: I think that this is not hurting to much... the convention is to write the release note starting with an object path.

twisted.trial.unittest.TestCase.flushWarnings is part of the release message , but warnAboutFunction is not included in the release notes... and I think that this PR also fixed warnAboutFunction ... and this is public API

PR #1537, opened 2021-03-03
Make cython-test-exception-raiser extension optional

kulikjak:
## Scope and purpose

The function `test_failureConstructionWithMungedStackSucceeds` has a `@skipIf` decorator [[here](https://github.com/kulikjak/twisted/blob/625be70a220eb528f75e9f4d3c9aec15aa782fc9/src/twisted/test/test_failure.py#L671)] if `cython-test-exception-raiser` is not available, but the import is not as permissive. Since the rest of the test suite is not affected by the missing extension in any way, and the test itself is ready for that, it should be made optional.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10113

### DIFF
--- a/src/twisted/test/test_failure.py
+++ b/src/twisted/test/test_failure.py
@@ -21,8 +21,10 @@ from twisted.python import failure
 
 from twisted.trial.unittest import SynchronousTestCase
 
-
-from cython_test_exception_raiser import raiser
+try:
+    from cython_test_exception_raiser import raiser
+except ImportError:
+    raiser = None
 
 
 def getDivisionFailure(*args, **kwargs):


graingert:
I intended to make cython_test_exception_raiser mandatory in tests, and I left that `@skipIf` present by accident.

kulikjak:
> I intended to make cython_test_exception_raiser mandatory in tests, and I left that `@skipIf` present by accident.

Well, that's another option.

Speaking for the Oracle Solaris FOSS team, I would prefer it to be optional because all python components/libraries used to test twisted internally must go through an approval process and then require some maintenance, and the addition of a simple extension just to go through one additional test doesn't seem to be worth it.

But I don't know what are your longer-term plans with it, and very well understand that our internal processes should not affect your decisions about twisted testing ;).

graingert:
> and the addition of a simple extension just to go through one additional test doesn't seem to be worth it.

I moved the extension module from twisted/twisted into twisted/cython-test-exception-raiser to make twisted installable from pypi.org as a pure python wheel. How did you manage this module when it was part of twisted/twisted?

graingert:
> the Oracle Solaris FOSS team

for reference, and my interest, do you have a link to the associated downstream ticket?

kulikjak:
> I moved the extension module from twisted/twisted into twisted/cython-test-exception-raiser to make twisted installable from pypi.org as a pure python wheel. 

I see, that makes sense.

> How did you manage this module when it was part of twisted/twisted?

Not sure if I understand this question correctly. When it was part of twisted, it was all just one component - one approval, single component to update when necessary, single package to deliver.

> for reference, and my interest, do you have a link to the associated downstream ticket?

We don't have public issue tracker for Oracle Solaris.

graingert:
> When it was part of twisted, it was all just one component - one approval, single component to update when necessary, single package to deliver.

what I mean is that the source code itself was already approved - can you not fast track it on those grounds?

You could also re-vendor the package as part of your build system

kulikjak:
> what I mean is that the source code itself was already approved - can you not fast track it on those grounds?

A little bit maybe.

> You could also re-vendor the package as part of your build system

Yes, that is also an option - we can always patch it back in. I am already patching in this PR, so we have several solutions. I just wanted to offer it to upstream since it looked like that is the expected behavior (and every patch that gets accepted upstream and we don't have to add ourselves is a good one).

But it looks like it is the other way around; I should've checked previous versions...

adiroiban:
for reference, in setup.cfg  are the test extra dependencies 

```
[options.extras_require]
; This is list of minimum dependencies required to run the test.
; The plan is to remove the `test` extra requirement and skip tests that
; depend on then
; To prevent trunk failures we should pin all dev dependencies to minor
; versions.
; We trust semantic versioning and auto-upgrading to a bugfix release
; should be OK.
test =
    cython-test-exception-raiser ~= 1.0
    PyHamcrest >= 1.9.0
```

with the change from this PR, I think the cython-test-exception-raiser can be removed from test extras.

I think that as long as we have a code coverage that fails if the extension is not successfully imported in a test, we are fine.

And the long term goal is to have all test extensions options... including pyhamcrest :)

kulikjak:
> And the long term goal is to have all test extensions options... including pyhamcrest :)

Ah, well in that case this PR is useful no matter our process!

> with the change from this PR, I think the cython-test-exception-raiser can be removed from test extras.

Do I remove it completely? I mean, it is no longer required, but it is still an optional import in the test suite, so maybe it should be moved to another section? I am not familiar with this stuff much...

adiroiban:
> Do I remove it completely? I mean, it is no longer required, but it is still an optional import in the test suite, so maybe it should be moved to another section? I am not familiar with this stuff much...

@kulikjak  my understanding is that Twisted runtime is fine and works without the c extension raiser.

AFAIK, cython-test-exception-raiser provides already compiled wheels for all the supported platforms and as @graingert mentioned it's the same code .

We moved it to an external package to make it easier to install Twisted at runtime.

If you want to run the Twisted tests from the source code, you need to install Twisted in `test mode` first... `pip install .[test]` 

Now Solaris 10 is not supported by Twisted , so you will have to compile the extension first....but if your Solaris build system has python-dev packages, that should not be a problem.
 
We could skip more tests, and to an extreme have a test suite in which all tests are skipped and the test run is green... but when what is the point of the tests.

The extension is used to test that Failure traceback reporting works and Failure is a core functionality of Twisted.

So... can you see if you can build the extension on Solaris during normal Twisted build time?

If we skip this test Solaris tests might pass but there is a risk that Failure traceback reporting will be broken in Solaris for CPython extensions code.

--------

So, I think the long term plan is to remove as many deps from the standard test run... but that is only done to check that core code is sane.

cython-test-exception-raiser  will remain as an exception.

-----

If you think that cython-test-exception-raiser  should still be removed, please reach to Twisted mailing list so that other devs can decide.  Not all Twisted developers are checking the PR comments.

graingert:
> The extension is used to test that Failure traceback reporting works and Failure is a core functionality of Twisted.

Yes to be clear this isn't testing an integration with an optional cython-test-exception-raiser, Twisted is using cython-test-exception-raiser in a self test

kulikjak:
I see; I misunderstood the first @adiroiban comment a little. It makes sense that since this is testing the core functionality of twisted, it should not be optional.

Thanks for the help and your comments. I am closing this, and we will handle it internally.

adiroiban:
Sorry @kulikjak  for the confusion . It was my fault. I wasn't sure at first how to proceed. 

If you have any issues with building the extension let me know. I had a chat with my colleague from $work and we are building Python and Twisted ourselves on Sparc Solaris 11 ..and used to do it also for Solaris 10... so I guess that all should work fine.

Note that at this point, Solaris (Sparc or X64) is not a supported OS for Twisted.

The Twisted only supports the OS/environments on which we run CI tests.

If the Solaris project can donate a GitHub Action runner for Solaris Sparc we can try to set up the CI to also run the tests on Solaris.

----------------

Thanks @graingert for the help to find the reason why things are the same... I will try to update the internal docs/comments as a drive by in a future PR.

kulikjak:
@adiroiban No problem ;).

I tried myself, and I can confirm that both **Twisted** and the **cython-test-exception-raiser** extension build without issues on Solaris; the test suite also passes without problems.

As for the Action runner, we are currently not able to provide Solaris machines for FOSS communities, but we are working on a solution.

PR #1539, opened 2021-03-05
[10116] Allow ByteLikes to be written for FileDescriptor instead of specifically bytes

wongpat:
## Scope and purpose

Somewhere between twisted 20.3.0 and 21.2.0, I noticed errors popping up complaining that the data sent must be bytes when using `twisted.internet.abstract.FileDescriptor`. It turns out in 20.3, the condition was

```py
        if isinstance(data, unicode): # no, really, I mean it
            raise TypeError("Data must not be unicode")
```

but was changed to use `_dataMustBeBytes` which only checks for `bytes` instances. However, it also meant it would fail when trying to write a `memoryview` (which can only be backed by `bytes` or a `bytearray`). This PR addresses it by adding support for `memoryview`s.

TRAC ticket https://twistedmatrix.com/trac/ticket/10116#ticket

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/


graingert:
Currently the type annotation for this is bytes so I suspect this needs updating too

Also I think this should be any bytes like object: https://docs.python.org/dev/glossary.html#term-bytes-like-object

src/twisted/internet/abstract.py
```diff
@@ -19,7 +19,7 @@
 
 
 def _dataMustBeBytes(obj):
```
graingert: so FileDescriptor.write calls `self._tempDataBuffer.append(data)` which is used in `doWrite` which calls:

```python
self.dataBuffer = _concatenate(
    self.dataBuffer, self.offset, self._tempDataBuffer
)
```

Which calls `bytes.join` which is defined as taking a `typing.ByteString | memoryview`:
https://github.com/python/typeshed/blob/16ae4c61201cd8b96b8b22cdfb2ab9e89ba5bcf2/stdlib/builtins.pyi#L440

So maybe we should be type-checking (at runtime and with mypy) for that instead of bytes

graingert: looks like this is still wrong: https://github.com/python/typing/issues/593#issuecomment-605516078

> Also - typing.Bytestring (mentioned in the original post) doesn't seem appropriate in all cases since it looks like it is Sequence[int] in there. A sequence of ints is not accepted by zlib.decompress, for example, although my dusty memory uncertainly thinks that a sequence of ints was supposed to be legit for a true buffer protocol (I'm really not sure, though).

graingert: I think we want: https://github.com/python/typeshed/blob/c92e3cb228d478a4cbd8f53072cae608dcf2f6c4/stdlib/_typeshed/__init__.pyi#L171

```python
ReadableBuffer = Union[bytes, bytearray, memoryview, array.array, mmap.mmap]
READABLE_BUFFER = (bytes, bytearray, memoryview, array.array, mmap.mmap)

def _dataMustBeBytes(obj: ReadableBuffer):
    if not isinstance(obj, READABLE_BUFFER):
        raise TypeError("Data must be bytes")
```

and then we need to go through and replace `data: bytes` with `data: ReadableBuffer` everywhere

wongpat: I agree that the protocol should accept any byte-like object (or something that fulfills the buffer protocol); my commits were limited to just supporting `memoryview` but this makes more sense. Just have to expand on the unit tests as well, but I am not sure if twisted's integration with unit tests support sub tests or parameterization.

wsanchez: I think that changing the type hint to `ReadableBuffer` as above is good and that removing the `_dataMustBeBytes` entirely would be the best path here…

The type hint will steer people the right way and the type check was there to catch string issues common in Py2, which is no longer in play.

graingert: Ok so a new public: `twisted.internet.abstract.ReadableBuffer`

```python
ReadableBuffer = Union[bytes, bytearray, memoryview, array.array, mmap.mmap]
```

and _dataMustBeBytes removed with all consumers updated to use the new type annotation.

@wongpat would you like to make these changes?

wongpat: > _dataMustBeBytes removed with all consumers updated to use the new type annotation.

Do we not want to throw an error if someone passes in a unicode value (without calling `encode`)? I'm fine doing as suggested, but wanted to confirm this behavior change. I guess those using mypy will be able to cache the errors too.

src/twisted/internet/abstract.py
```diff
@@ -186,7 +186,7 @@ def __init__(self, reactor: Optional[interfaces.IReactorFDSet] = None):
             reactor = _reactor  # type: ignore[assignment]
         self.reactor = reactor
         # will be added to dataBuffer in doWrite
-        self._tempDataBuffer: List[bytes] = []
+        self._tempDataBuffer: List[ReadableBuffer] = []
```
wongpat, line 189: This already existed, but wouldn't this fail with python 3.5 since variable annotations started with 3.6?

graingert: twisted no longer supports Python3.5 because it is EOL

src/twisted/_version.py
```diff
@@ -7,5 +7,5 @@
 
 from incremental import Version
 
-__version__ = Version("Twisted", 20, 3, 0, dev=0)
+__version__ = Version("Twisted", 21, 2, 1)
```
graingert:
```suggestion
__version__ = Version("Twisted", 20, 3, 0, dev=0)
```

graingert: we bump the version as part of the release process - if every PR changed the version then every PR would have merge conflicts

src/twisted/internet/abstract.py
wongpat: @wsanchez @graingert I implemented the suggestions. The automatic mypy tests are failing unfortunately, but drilling down to the errors they weren't lines I modified.

graingert:
@wongpat the fix is here https://github.com/twisted/twisted/compare/trunk...graingert:allow-all-readable-buffer-in-write

but I can't push it to your PR because [you created this PR from the default branch of your repository](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/)

wongpat:
> @wongpat the fix is here [trunk...graingert:allow-all-readable-buffer-in-write](https://github.com/twisted/twisted/compare/trunk...graingert:allow-all-readable-buffer-in-write)
> 
> but I can't push it to your PR because [you created this PR from the default branch of your repository](https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/)

Should I close and open a new PR?

graingert:
You can pull my changes in yourself, or recreate the PR and I'll push them

wongpat:
> You can pull my changes in yourself, or recreate the PR and I'll push them

Just did that. Let me know if there's anything else you want me to look into. Thanks

exarkun:
The interface that this code implements is pretty clear: the argument is `bytes`: https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.ITransport.html#write

wsanchez:
> The interface that this code implements is pretty clear: the argument is `bytes`: https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.ITransport.html#write

True, and we've been discussion recently how it's not really feasible (unfortunately out not) to modify an interface like this.  In this case, for example, we're talking about one implementation of `ITransport` that is provided by Twisted, but other clients may have built their own implementations and aren't require to handle types other than `bytes`.

That means we shouldn't change the type hint; the documentation has always said the argument should be `bytes`.

I can see an argument for changing the implementation to allow other types, but that would be a "legacy problem", would have to be noted as such by a comment, and we can't guarantee that this will remain the case indefinitely, but we could undergo deprecation process for that before changing it again.

src/twisted/internet/test/test_filedescriptor.py
```diff
@@ -55,22 +55,6 @@ class FileDescriptorTests(SynchronousTestCase):
     Tests for L{FileDescriptor}.
     """
 
-    def test_writeWithUnicodeRaisesException(self):
```
twm, line 58: Please restore these tests; the implementation here definitely shouldn't expect `str`.

twm (commented):
Note that a type annotation of `bytes` [also allows `bytearray` and `memoryview`](https://mypy.readthedocs.io/en/stable/duck_type_compatibility.html?highlight=memoryview#duck-type-compatibility).

glyph:
Taking this out of review for now since it needs conflicts resolved.

Do we want to punt on this entirely? It seems like we may want to have some loosening of the interface or something here, but perhaps we need some nuanced upgrade to the type hint in order to make this do what we want.

glyph:
@wongpat I'm sorry that this seems to have gotten lost in the shuffle and buried in conflicts, but at this point (with no response in 2 years) I think I'm going to have to call it. I would be happy to accept another fix for #10116 — and feel free to reopen this one — but having someone on the core team resurrect this one seems like more trouble than it's worth at this point.

PR #1540, opened 2021-03-07
Sync templates with pydoctor 21.2.2 release

mthuurne:
## Scope and purpose

Fixes several issues in the generated API docs. See ticket for details.

I think it would be worthwhile to also merge this to either the release-21.2.0-10091 branch or a branch spun off from that and then re-generate the docs for Twisted 21.2.0.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10105


mthuurne:
I don't know why the pypy-3.6 run fails: `trial` doesn't report any failed test cases, yet the `coverage` run returns a non-zero exit code.

adiroiban:
> I don't know why the pypy-3.6 run fails: `trial` doesn't report any failed test cases, yet the `coverage` run returns a non-zero exit code.

Sorry..,. That is a know issue and I have no idea how to fix it.

setup.cfg
```diff
@@ -54,7 +54,7 @@ test =
 ; release scripts and process.
 dev_release =
     towncrier ~= 19.2
-    pydoctor ~= 21.2; python_version >= "3.6"
+    pydoctor ~= 21.2.2; python_version >= "3.6"
```
adiroiban, line 57: shoulnd't `pydoctor ~= 21.2` automatically update to 21.2.2 once released?

src/twisted/newsfragments/10105.doc
```diff
@@ -0,0 +1,2 @@
+ Sync API docs templates with pydoctor 21.2.2 release.
+
```
adiroiban, line 2:
```suggestion
Update Twisted API docs templates to work with changes from the pydoctor 21.2.2 release.

```

adiroiban (approved):
Thanks. Looks good.

I was using this page for testing 

https://twisted--1540.org.readthedocs.build/en/1540/api/twisted.enterprise.adbapi.ConnectionPool.html?private=1#_runWithConnection

i can see that anchor goes to a visible docs...the private state is preserved

Also, the header looks the same like on https://twisted--1540.org.readthedocs.build/en/1540/api/index.html

Thanks! This can be merged.

I have only left minor comments 

For reference - pydoctor ticket for having project version on each page - https://github.com/twisted/pydoctor/issues/333
I was not able to find the Twisted ticket that ask for including Twisted version on each API docs page.
I found this ticket https://twistedmatrix.com/trac/ticket/3945 which is for narrative Docs... so I guess that we should either create a new Twisted ticket to update the template to include the version.

setup.cfg
mthuurne: It will install 21.2.2 if nothing was installed yet, but I don't know if it forces an update if an older version was already installed.

In any case, we need 21.2.2 or higher for the docs to be rendered correctly, so I think `~= 21.2.2` is a more accurate description of the requirements even if it doesn't make much of a difference in practice.

The `python_version >= "3.6"` part is only needed for the release branch, since `trunk` is now 3.6+. But it's harmless to have it on `trunk`.

src/twisted/newsfragments/10105.doc
mthuurne: I agree that your wording suggestion is slightly better, but I really don't want to go through another CI cycle of over half an hour to change one line that may or may not end up in the release notes one day...

adiroiban: 100% understand the decision and the pain :(
 I can update the release notes as a drive by on one of my PRs.

-------

Many thanks for merging this PR.

I have removed this branch... it should keep the repo a bit cleanup and auto-completion less noisy :)

-------

Regarding re-releasing the documentation. Please check with Craig.

If this is cherry-picked in the release branch I can update the documentation at https://twistedmatrix.com/documents/current/api/

For Read The Docs... there is not much we can do other than creating a  new 21.2.0.post1 tag to trigger a new build on the updated code... but since RTD is still WIP, I think that we can let it be :)

I have created this ticket to track the RTD migration progress https://twistedmatrix.com/trac/ticket/10117

PR #1543, opened 2021-03-09
[10121] Fix random PyPy CI failures. Improve GHA CI speed. Improve coverage reporting.

adiroiban:
## Scope and purpose

We got random failures on pypy (both 3.6 and 3.7 versions)

Also pypy jobs with coverage are very slow...30minutes.

See what can be done to speed up the CI.

### PYPY coverage

Coverage was disabled for most of the PyPy..
The tests are much faster now without coverage 4m vs 30m
We still run coverage on PYPY, but only for compat and a few other test that are exercising the PYPY specific code.

In this way we can monitor if we continue to see random CI failures at the end of the test run on PY.

These are the tests required to get the extra coverage on pypy
* twisted.test.test_compat
* twisted.test.test_defer
* twisted.internet.test.test_socket
* twisted.trial.test.test_tests

### Linux / MacOS general CI speedup

All tests are now executed in parallel, including pypy.

To gain more speed, I have removed `tox --notests` as it was just wasting time.
We can later show the dependencies using the python-info action.

PR are waiting to be reviewed to also get parallel jobs on Windows.

### Coverage reporting

I have updated coverage XML generation and codecov.io reporting to make the check red on failures.
Only coveralls failures are ignored for now.
In this way it should be easier to observer coverage issues. 

I tried to fix the coveralls report for trunk merges.
I did some work in #1574  but there is no way to test this unless merged in trunk.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10121
* [NA] I have updated the automated tests.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -104,12 +104,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip tox coverage coveralls
-        tox --notest
 
     - name: Test
       run: |
-        python --version
-        tox -q ${{ matrix.trial-target }}
+        tox ${{ matrix.trial-target }}
 
     - name: Prepare coverage
       if: always() && contains(matrix['tox-env'], 'withcov')
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -36,6 +36,8 @@ jobs:
         # As of 9 March 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
         # Trial distributed jobs enabled to speed up the CI jobs.
         trial-args: ['-j 4']
+        # Tests are executed with the default target which is the full test suite.
+        trial-target: ['']
 
         include:
 
@@ -61,6 +63,10 @@ jobs:
             tox-env: alldeps-nocov-posix
           - python-version: pypy-3.7
             tox-env: alldeps-nocov-posix
+          # We still run some tests with coverage
+          - python-version: pypy-3.7
+            tox-env: alldeps-withcov-posix
+            trial-target: twisted.test.test_compat twisted.test.test_defer twisted.trial.test.test_tests
 
 
     steps:
@@ -103,11 +109,10 @@ jobs:
     - name: Test
       run: |
         python --version
-        tox -q
+        tox -q ${{ matrix.trial-target }}
 
     - name: Prepare coverage
       if: always() && contains(matrix['tox-env'], 'withcov')
-      continue-on-error: true
       run: |
         # sub-process coverage are generated in separate files so we combine them
         # to get an unified coverage for the local run.
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,6 +21,8 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       TOXENV: "${{ matrix.tox-env }}"
+      # As of April 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
+      # Trial distributed jobs enabled to speed up the CI jobs.
       TRIAL_ARGS: "-j 4"
     name: ${{ matrix.python-version }}${{ matrix.noipv6 }}-${{ matrix.tox-env }}
     strategy:
@@ -32,10 +34,8 @@ jobs:
         # Run on latest minor release of each major python version.
         python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-alpha.5]
         tox-env: ['alldeps-withcov-posix']
+        # By default, tests are executed without disabling IPv6.
         noipv6: ['']
-        # As of 9 March 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
-        # Trial distributed jobs enabled to speed up the CI jobs.
-        trial-args: ['-j 4']
         # Tests are executed with the default target which is the full test suite.
         trial-target: ['']
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,7 +66,7 @@ jobs:
           # We still run some tests with coverage
           - python-version: pypy-3.7
             tox-env: alldeps-withcov-posix
-            trial-target: twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket
+            trial-target: twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests
 
 
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,7 +66,7 @@ jobs:
           # We still run some tests with coverage
           - python-version: pypy-3.7
             tox-env: alldeps-withcov-posix
-            trial-target: twisted.test.test_compat twisted.test.test_defer twisted.trial.test.test_tests
+            trial-target: twisted.test.test_compat twisted.test.test_defer
 
 
     steps:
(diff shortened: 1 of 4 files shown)


wsanchez:
`codecov/project` shows a slight drop.  I'm getting a 403 error when I try to view it on Codecov, but it would be good to know where the dropped lines are.

codecov.yml
```diff
@@ -12,4 +19,12 @@ coverage:
       default:
         target: 100%
 
-comment: off
+# We don't want to receive general PR comments about coverage.
```
adiroiban, line 35: These are already the defaults, but I have added them here to make it explicit and help undestand how the whole codecov.io  process works

codecov.yml
```diff
@@ -1,9 +1,16 @@
...
+# Twisted settings: https://codecov.io/gh/twisted/twisted/settings/yaml
 #
 # We want 100% coverage for new patches to make sure we are always increasing
 # the coverage.
 #
+codecov:
+  require_ci_to_pass: yes
+  notify:
+    after_n_builds: 7
+    wait_for_ci: yes
```
adiroiban, line 14: Here I am trying to see if I can configure codecov.io to wait a bit more before sending the coverage reports so that we don't have transient red checks

adiroiban:
I have no idea why `codecov/project` is red ... maybe it will need time to be updated -
i have checked the page https://app.codecov.io/gh/twisted/twisted/compare/1543/changes and the PR has no missing coverage (when compared to trunk)

This is ready for review. See PR description

.github/workflows/test.yaml
```diff
@@ -99,36 +103,32 @@ jobs:
...
-      continue-on-error: true
+      if: always() && contains(matrix['tox-env'], 'withcov')
       run: |
         # sub-process coverage are generated in separate files so we combine them
         # to get an unified coverage for the local run.
         # The XML is generate to be used with 3rd party tools like diff-cover.
         python -m coverage combine
         python -m coverage xml -o coverage.xml -i
-        python -m coverage report
+        python -m coverage report --skip-covered
```
twm, line 120: 👍

.github/workflows/test.yaml
```diff
@@ -36,6 +36,8 @@ jobs:
         # As of 9 March 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
         # Trial distributed jobs enabled to speed up the CI jobs.
         trial-args: ['-j 4']
```
twm: Should this be removed as you removed the reference to it above?

adiroiban:
@twm thanks for the review. I have updated the GHA config. Let me know if you see any issues with it.

I see the codecov.io check is still sent before all checks are done

The codecov.io docs say that cryptography project is a good example 
https://docs.codecov.io/docs/merging-reports 
https://github.com/pyca/cryptography/blob/main/codecov.yml


-------------

The codecov/project is now red ...but I guess that after some time codecov.io is preocessing the reports and will update it.

The codecov page looks green https://app.codecov.io/gh/twisted/twisted/compare/1543/changes

I happened  with my previous comment. By the time I left the comment the check was red for a5326cc and now it's green

My plan is to rewrite the coverage reporting in a separate PR, so that we collect all GHA coverage in a separate job and combine ourselves and then send a single coverage.xml for the whole GHA builds... I will test in in a separate branch ... but might need moving one Windows and MacOS job to GitHub

wiml:
> The codecov/project is now red

It looks like the change is due to a now-untested branch in `TestDecoratorMixin.test_decorateTestSuiteReferences()`, near lines 1167–1169, which skips a test if `sys.getrefcount` is not available … I'd guess that PyPy is the only tested target that doesn't have getrefcount. Maybe add `twisted.trial.test.test_tests` (or all of `twisted.trial`) to the mypy suite?

adiroiban:
> It looks like the change is due to a now-untested branch in `TestDecoratorMixin.test_decorateTestSuiteReferences()`,

Thanks. You are my hero :)

Updated pypy coverage only take 3 minutes..

7 minutes is the total run time for GitHub Actions tests.

We still have Azure tests which take  14m.

So with the changes from this PR GHA should be ok and we should look at Azure speedup next

codecov.yml
wiml: I don't think the existing behavior is bad, necessarily, it's just confusing for people who (like me) assume that a check is complete once it shows either a check or X.

wiml (approved):
LGTM, although I'm not terribly familiar with codecov.io+GHA interactions.

codecov.yml
exarkun: In my experience these two options don't actually work.  I never got to the bottom of why and codecov.io declined to assist.

codecov.yml
```diff
@@ -1,15 +1,49 @@
...
+        # Allow for a bit of slack in overall project report.
+        # As long as the patch is 100% the PR should be ok.
+        # We report coverage even on errors.
+        # So a run on trunk might fail and during the failure process will
+        # exercise a different code branch (ex timeout errors).
+        # This code branch is reported at the end of the run.
+        # Future test runs will not trigger the error and the error handling
+        # code branch is not called, and if no other tests is touching that
+        # branch we are left with overall reduced code coverage.
+        threshold: 0.02%
```
exarkun, line 32: This setting may be required if coverage is a required check but the reasoning here is invalid.  The threshold value allows an ongoing slip in coverage.  Build N of trunk can have (0.02 × M) greater coverage than build (N+M) of trunk and this configuration tells codecov.io that this is fine.

So this setting should be considered a temporary fix while the real problem is being addressed - non-deterministic coverage in the test suite.

altendky: There are various cases where coverage can be reduced.  Even just deleting lines that have been covered causes this.  It's just a result of not having '100%' coverage (after including consideration of coverage ignores).  I don't know if there's a way to deal with that separately from the commented situation.

altendky (approved):
> To gain more speed, I have removed tox --notests as it was just wasting time.

I guess that should be brought back in https://github.com/twisted/twisted/pull/1577 after this PR is merged?

codecov.yml
altendky: What are the odds, that is the case in my next review.  https://github.com/twisted/towncrier/pull/334  They deleted a line that is useless and the coverage dropped by 0.01%.

adiroiban: Thanks wiml for the review. Just monitoring future PR and checking for overall project coverage regression can help with future fixed for this issue.

Thanks exarkun for the headup. I forgot to file the followup ticket.

Thanks altendky for the code delete info. I was hoping that codeco.io is smart and will not trigger such an error. 
I have reported this issue 
https://community.codecov.io/t/project-coverage-decrease-on-deleted-line/2726

-----------------

I agree that this should be a temporary change. 
I have added to link to the ticket in the comment.

If coverage/project will continue to be red, people will most probably ignore it by default.
With that, I think that is a higher probability of seeing an overall decrease in coverage.

We had the project red for some time, and I am not aware of any secondary tickets or actions to fix the overall project coverage.

-----------

Regarding 0.02 x M.
We have 100% patch coverage , so is not that easy to find a series of PR  with 100% patch coverage but witch will always decreased the overall project by 0.02 due to different tests.

I am monitoring overall coverage changes  as there is a gang of usual suspect tests.

For example

image

With this swinging line

image

There is also this swinging test helpers to prevent a flaky test failure

image

codecov.yml
adiroiban: I will try to find some time and report this issue via the internal / paid support codecov channel.

It is true that codecov are not helping too much with open source projects... so we are on a free ride here.

But I see if I can reproduce it for my private project and try to get help via that project.

adiroiban:
> I guess that should be brought back in #1577 after this PR is merged?

@altendky yep... looking forward for that PR. It would be nice to see a separate timing for deps install vs actual test run. Thanks

PR #1545, opened 2021-03-09
[10123] Run parallel test jobs on Linux and macOS.

adiroiban:
## Scope and purpose

Since the GHA and Azure VMs have 2 CPU I think that it makes sense to run the jobs on both CPUs.

Also, Twisted own tests should run in parallel to make it easier (aka faster) to run on local computers.
I guess most Twisted dev have 4CPUs and up.

## Changes

I added `TRIAL_ARGS` and not just a more specific `TRIAL_JOBS` as even when using 1 job trial still spawn a separate process and debugging is not enabled.

I added names for coverage codecov builds to make it easier to debug coverage issues.

I have fixed the coverage reporting on Azure pipelines for Codecov. So we now get macOS and Windows coverage.

I have fixed the tests that were  failing with running in distributed or debug mode.

For now I have enabled CPython with 4 jobs as it gets an 10% (from my very quick calculations) over j2 and no random failures.

I have also enabled parallel jobs on macOS.

Parallel jobs are not supported on Windows... yet.

On PYPY there are random failures with -j2 and -j4 ... on my pypy 3.6 I have executed the tests 50 times with 16 concurrent jobs and there were no failures ...

##  Coverage change

You can see the coverage report here https://app.codecov.io/gh/twisted/twisted/compare/1545/changes

There are changes since we now test or skip the code for the `Long Usage Description` that is automatically pulled from `__main__` 

Also, there are changes due to the way the tests are started from the distributed jobs.

## Random failures

No failures for a single parallel run with 8 workers.

```
for i in `seq 50`; do trial -j16 twisted; done
```

I have executed a few times `trial -u -j8 twisted`  and on the second run there are some random tests that are failing...
maybe they are doing some caching

```
twisted.test.test_rebuild.RebuildTests.test_FileRebuild
twisted.test.test_twistd.AppLoggerTests.test_legacyObservers
twisted.test.test_main.MainTests.test_twisted

twisted.python.test.test_release.APIBuilderTests.test_build
twisted.test.test_sslverify.ExpandCipherStringTests.test_doesNotSwallowOtherSSLErrors

twisted.python.test.test_shellcomp.ZshTests.test_zshCode

twisted.mail.test.test_pop3.POP3ClientDeprecationTests.test_deprecation
```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10123


adiroiban:
On my laptop this is pretty fast

```
$ pip install -e .[all_non_platform,dev]
$ trial -j twitsed
Ran 11952 tests in 22.008s

PASSED (skips=1568, successes=10384)
```

---------

The coverage part was already handled by tox.ini .

This is ready for review

adiroiban:
codecov build names:

image

adiroiban:
Below is the summary of the diff changes caused by running jobs with dist trial 

workertrial code is no longer executed... ... so I guess that we would still need a test run that will  be triggered in non-distributed way to have a better end to end coverage

image

.github/workflows/test.yaml
```diff
@@ -35,15 +34,21 @@ jobs:
         python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-alpha.5, pypy-3.6, pypy-3.7]
         tox-env: ['alldeps-withcov-posix']
         noipv6: ['']
+        # As of 9 March 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
+        # Enable trial distributed jobs to speed up the jobs.
+        trial-args: ['-j 2']
```
altendky: Apparently trial needs an 'auto-parallelism' feature.  `:|`

altendky (dismissed):
8 -> 5 minutes for the 3.9 run I checked.  That's nice.  I don't use trial or know how robust its parallel runs are but I trust you do and have already considered that.  I guess this will help expose what problems there are.  `:]`

src/twisted/test/test_main.py
```diff
@@ -47,7 +53,10 @@ def test_trial(self):
         cmd = sys.executable
         p = Accumulator()
         d = p.endedDeferred = defer.Deferred()
-        reactor.spawnProcess(p, cmd, [cmd, "-m", "twisted.trial", "--help"], env=None)
+        # Enforce 80 columns as otherwise the test are failing when executed in debug mode.
+        reactor.spawnProcess(
+            p, cmd, [cmd, "-m", "twisted.trial", "--help"], env={"COLUMNS": "80"}
```
glyph: We should never unilaterally clear `env` like this; this value should be something more like `dict(os.environ, COLUMNS='80')`

.github/workflows/test.yaml
```diff
@@ -20,31 +20,40 @@ jobs:
...
       matrix:
         # Run on the minimum micro Python version that we can get on CI.
         # When updating the minimum Python version here, also update the
         # `python_requires` from `setup.cfg`.
         # Run on latest minor release of each major python version.
         python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-alpha.5, pypy-3.6, pypy-3.7]
-        tox-env: [alldeps]
+        tox-env: ['alldeps-withcov-posix']
         noipv6: ['']
+        # As of 9 March 2021 GHA VM have 2 CPUs - Azure Standard_DS2_v2
```
glyph, line 37: Typically you want to go a little bit above the raw CPU count, since processes can and do also block on I/O, so you want to give the scheduler a bit more ready work to play with.  It really depends on the workload though, so my suggestion would be to bump it to 3 and look at the timings.

src/twisted/test/test_main.py
adiroiban: May I ask why? :)

Since this is a test I was thinking you might want full control over the environment... to reduce any external dependency.

I have no idea why spawnProcess has different behaviour for empty env :)

```
                - C{env} is not set:
                  - On POSIX: pass an empty environment.
                  - On Windows: pass L{os.environ}.
                - C{env} is L{None}:
                  - On POSIX: pass L{os.environ}.
                  - On Windows: pass L{os.environ}.
                - C{env} is a L{dict}:
                  - On POSIX: pass the key/value pairs in C{env} as the
                    complete environment.
                  - On Windows: update L{os.environ} with the key/value
                    pairs in the L{dict} before passing it. As a
                    consequence of U{bug #1640
                    <http://twistedmatrix.com/trac/ticket/1640>}, passing
                    keys with empty values in an effort to unset
                    environment variables I{won't} unset them.
```

----------

I have no idea why we get a difference behavior for COLUMNS  (and environment variables in general) as this env var is not set in the GitHub Runner environment:

https://github.com/twisted/twisted/pull/1545/checks?check_run_id=2075976401#step:7:21

The tests are also failing when calling `trial --debug` ... and I can see COLUMNS and LINES env var being defined in the spawned process... but they are not defined in my shell process or in the parent trial process.

.github/workflows/test.yaml
adiroiban: I did some test on my local system before going with this value :) ... but increasing above CPU count yields no improvements... But I see that on GitHub VM there is a small improvement.

The GitHub VM are slower... so 3minutes for a run vs 20seconds for local run

Below are my local results on a 8core 16 threads cpu... i did multiple runs for each job count... here are just some examples:

* -j8 Ran 11952 tests in 27.908s
* -j16 Ran 11952 tests in 17.072s
* -j17 Ran 11952 tests in 17.965s
* -j24 Ran 11952 tests in 17.671s

---------

I went with -j2 as this is the CPU count on Azure, but at the same time, to reduce the probability of running at the same time tests that are not isolated.

Here are the logs for -j2 https://github.com/twisted/twisted/runs/2075976363
Here are the logs for -j4 https://github.com/twisted/twisted/pull/1545/checks?check_run_id=2085348540

There is an increase in performance for -j4

On -j4 pypy tests failed 

For the initial change I would say that is better to do baby steps and just go with 2 concurrent jobs to not introduce (just yet) too many random failures on trunk.

altendky:
FWIW, I'm confused about my review.  I remember seeing GHA and tox changes, not code changes.  But, my review seems to show up with those other changes present.  I'm going to go ahead and rescind my approval since I haven't looked over the code changes and @glyph has already raised issues with them.

.github/workflows/test.yaml
glyph: > Below are my local results on a 8core 16 threads cpu... i did multiple runs for each job count... here are just some examples:

FWIW I would expect the top speed to be somewhere around `-j11` or `-j10` for an 8-core machine, not all the way up at `-j16`.  (In my experience hyperthreading doesn't help Python very much, since there's too much cache contention; it's real cores that give you real concurrency.)

glyph: > For the initial change I would say that is better to do baby steps and just go with 2 concurrent jobs to not introduce (just yet) too many random failures on trunk.

Yep, that seems fine.

.github/workflows/test.yaml
```diff
@@ -20,32 +20,50 @@ jobs:
   testing:
     runs-on: ubuntu-20.04
     env:
-      TOXENV: "${{ matrix.tox-env }}-withcov-posix,coverage-prepare,codecov-push,coveralls-push"
-    name: ${{ matrix.python-version }}${{ matrix.noipv6 }}-${{ matrix.tox-env }}-withcov-posix
+      TOXENV: "${{ matrix.tox-env }}"
+      CODECOV_OPTIONS: "-n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}'"
+      TRIAL_ARGS: "${{ matrix.trial-args }}"
```
wsanchez, line 25: I would call this TRIAL_JOBS and in the Tox config use `--jobs="{env:TRIAL_JOBS}"` when invoking Trial.  But I'd also set TRIAL_JOBS in a `setenv` clause in Tox, and have it default to `2`, so that when one runs locally, they are also testing in parallel (which is faster but also matches CI).

src/twisted/test/test_main.py
wsanchez: Well, if the goal of this change is to set `COLUMNS`, then stomping the environment possibly does extra unintended/unrelated things.

As to why not do that; we should not make it impossible to pass environment variables through here; we should (and do) control the environment via Tox (or otherwise in the test environment).

I'll not that these fixes for `--debug` are not related to the number of jobs, so I'd rather see them in a separate PR instead of rolling them in here.  That said, I wouldn't block a review for this, but I agree with Glyph about the environment.

wsanchez (changes requested):
Please don't step on the environment.

TRIAL_JOBS thing is a suggestion; feel free to ignore.

.github/workflows/test.yaml
adiroiban: From PR description :)

> I added TRIAL_ARGS and not just a more specific TRIAL_JOBS as even when using 1 job trial still spawn a separate process and debugging is not enabled.

My experience using tox for local TDD is horrible... so I stopped using tox at all on my local system.

Do you use tox when developing twisted on your local system? Isn't it super slow?

.github/workflows/test.yaml
adiroiban: With -j10 I only got 19 seconds... and j11 got 18 seconds... but yes j14 got the same results as j16 --- 17 seconds.

On my local system I have executed the full Twisted tests with python 3.9 and pypy3.6 hundreds of times without any random failure.

adiroiban:
Thanks for your reviews.
I have updated the tests to keep the existing environment and only update COLUMNS

src/twisted/application/twist/_options.py
```diff
@@ -59,6 +59,10 @@ def __init__(self) -> None:
         self["reactorName"] = self.defaultReactorName
         self["logLevel"] = self.defaultLogLevel
         self["logFile"] = stdout
+        # An empty long description is explicitly set here as otherwise
+        # when executing from distributed trial twisted.python.usage will
+        # pull the description from `__main__` which is another entry point.
```
exarkun, line 64: It seems like the real condition that applies here is something other than "executing from distributed trial"?  Is it really *specifically* an interaction between disttrial and Options?  Or is it about `__main__` being set unexpected in that case?

Put another way, is `Options` broken when used with `python -m` in general or is it broken against disttrial?

src/twisted/test/test_main.py
```diff
@@ -19,12 +19,21 @@
...
         p = Accumulator()
         d = p.endedDeferred = defer.Deferred()
-        reactor.spawnProcess(p, cmd, [cmd, "-m", "twisted", "--help"], env=None)
+        # Enforce 80 columns as otherwise the test are failing when executed in debug mode
+        # or other places in which COLUMNS env var is defined.
+        # The environment is set for both parent and child processes.
+        self.patch(os, "environ", dict(os.environ, COLUMNS="80"))
+        reactor.spawnProcess(
+            p, cmd, [cmd, "-m", "twisted", "--help"], env=dict(os.environ, COLUMNS="80")
+        )
```
exarkun: It would probably be nice to factor out the duplicate `dict(...)` since the point is to make the environment identical in the parent and child (using the same object makes this clearer and more likely).

src/twisted/application/twist/_options.py
adiroiban: Is it about __main__ being a an unexpected in that case.

Then running with distributed worker, `__main__` is the distributed worker main and not the command main,

> Options broken when used with python -m in general or is it broken against disttrial?

Not sure if broken in general or just with distrial.

It works with `python -m` ... but not when called via disttrial

I would say that the automatic extraction of the long description from  `__main__` is broken in general... is just that it works with a direct invocation ffrom `bin/python -m `

-------------

This is hitting this branch
https://github.com/twisted/twisted/blob/7252b5e6d46ec5a834f8d79f6faba0c8252a36e3/src/twisted/python/usage.py#L538

Since the long description was not used anyway, I think it's best to explicitly set it here and don't rely on the automatic extraction of the long description.

Now... in the future, if a long description is added to the docstring of `twist` then this should be updated.

src/twisted/test/test_main.py
adiroiban: Based on the conversation over IRC I went and simplified the assertion to not check for the full help message.

I hope this is better :)

src/twisted/application/twist/_options.py
exarkun: I looked a little more closely and I think I understand this better now.

The test that hits this problem asks the `twist` options to describe themselves in two execution contexts.  (1) in the in-process context of the running test suite and (2) in a newly launched `python -m twisted` child process.

For (a), there are now two sub-cases we've identified.  (a) the test suite is being run by `trial ...` and (b) the test suite is being run by `trial -jN ...`.

For (1a), `__main__` is `twisted.scripts.trial` which has no docstring.  For (1b), `__main__` is `twisted.trial._disttrial.workertrial` which has a docstring.  For (2) `__main__` is `twisted.__main__` which has no docstring.

Thus, the test was (accidentally, I assume) correct because cases (1a) and (2) are in the same state with respect to their docstrings whereas case (1b) differs.

For now, it seems like the work-around is probably alright.  It only fixes one example of the general problem here - which is that `__main__.__doc__` doesn't seem to be particularly indicative of what an `Options` class is for.  Also, it's a bad API because it's action-at-a-distance, it's global state, and it's not even documented anywhere.

I think this behavior should be removed from `Options` and if folks want to set a longdesc, they should set a longdesc.

Do you agree?  Will you file a ticket for deprecation and removal?

adiroiban: Yes. I agreed to deprecate the automatic long description feature in t.python.usage . I asked about this over IRC a few days ago.

https://twistedmatrix.com/trac/ticket/10133

---------------

Yes. Those are the 2 subclasses.

For (1a trial script) the test was not failing as twisted.scripts.trial._BasicOptions defined its own `longdesc` so the `__main__` docstring is not used at all.

https://github.com/twisted/twisted/blob/9a59de7abfe27f6c2ae1545374493f430938fb5b/src/twisted/scripts/trial.py#L165

This change should not be considered a workaround. It's an explicit definition for `longdesc` for `twist` CLI options.
For `twist` script I defined an empty `longdesc` as this was the automatically pulled value anyway.

adiroiban:
This is ready for a new review. We got coverage codecov.io reporting working on Linux, Win and macOS

https://codecov.io/gh/twisted/twisted/commit/52472097f4aaac29e6c7d42567d5534de09f8640/build

image

adiroiban:
I have observed `twisted.test.test_twistd.AppLoggerTests.test_legacyObservers` test failing with parallel jobs,

I was able to replicate them on my local system with `$ trial -u -j 8 twisted.test.test_twistd`

I have fixed the tests to do cleanup and I hope that now all is ok.

src/twisted/test/test_twistd.py
```diff
@@ -1542,9 +1542,10 @@ def test_getLogObserverFile(self):
         """
         logFiles = _patchTextFileLogObserver(self.patch)
         filename = self.mktemp()
-        logger = app.AppLogger({"logfile": filename})
+        sut = app.AppLogger({"logfile": filename})
```
adiroiban, line 1545: this was shadowing the global `logger` module.
I named it `system under test` :) open to suggestions for better names

src/twisted/test/test_twistd.py
```diff
@@ -1542,9 +1542,10 @@ def test_getLogObserverFile(self):
...
 
-        logger._getLogObserver()
+        observer = sut._getLogObserver()
+        self.addCleanup(logger.globalLogPublisher.removeObserver, observer)
```
adiroiban, line 1548: this is the global logger and this test was adding an observer but not removing it at the end

the created observer was opening a temp file. which was not closed at the end

this resulted in random failures in other tests as the garbage collector was complaining about the file that was left open.

src/twisted/test/test_twistd.py
```diff
@@ -1595,7 +1596,7 @@ def __call__(self, x):
 
         self.assertIn("starting up", textFromEventDict(logs[0]))
         warnings = self.flushWarnings([self.test_legacyObservers])
-        self.assertEqual(len(warnings), 0)
+        self.assertEqual(len(warnings), 0, warnings)
```
adiroiban, line 1599: this was added here to help debugger. intead of seeing `1 != 0` you see the actual warning and get a clue of what went wrong.

.github/workflows/test.yaml
wsanchez: OK, sorry I missed that text.

I use tox locally. And yes, it is way more painful with Twisted than with other projects, and in my experience actually wrong if you don't always use `-r`, which is insane.

All that said, making tests when run locally match what's happening in CI is probably a good idea, and "Tox sucks for me" seems like a non-argument, and "it's super slow" suggests that defaulting to `-j2` would be an improvement.

I don't know what "debugging is not enabled" means; I probably just never use the enabled debugging things.

All *that* said, I'm fine with leaving `TRIAL_ARGS ` alone here if "debugging is not enabled" causes people grief.

wsanchez (approved):

PR #1546, opened 2021-03-10
asynchronous filesystem access

ihaywood3:
see https://twistedmatrix.com/trac/ticket/10079

##


src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+from typing import Tuple, Iterable, Optional, Callable, cast, Any
+import os
+import os.path
+import threading
+
+log = Logger()
+
+
+DEFAULT_BUFFER_MAX = 2 ** 12  # four megabytes
+DEFAULT_CHUNK_SIZE = 4096
```
exarkun: Why make these globals instead of part of the interface?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+
+class IAsyncFilesystem(Interface):
+
+    """
+    an asynchronous filesystem
+    based closely on L{twisted.conch.interfaces.ISFTPServer}
```
exarkun: How many people are going to benefit from hearing about SFTP here?  This might make an interesting implementation note in a comment rather than a docstring.  How about some more high-level information about what folks can expect from the interface here?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def openFile(
+        filename: str, flags: int = 0, attrs: Optional[dict] = None
+    ) -> "Deferred":
+        """
+        Called when the clients asks to open a file.
+
+        @param filename: a string representing the file to open.
+
+        @param flags: an integer of the flags to open the file with, ORed
```
exarkun: Bitfields in integers is pretty old-school.  How about something using an enum, something using constantly, etc?  Whatever it is, the actual valid values and their meaning should be documented somewhere.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+            size: the size of the file in bytes
+            uid: the user ID of the file as an integer
+            gid: the group ID of the file as an integer
+            permissions: the permissions of the file with as an integer.
+            the bit representation of this field is defined by POSIX.
+            atime: the access time of the file as seconds since the epoch.
+            mtime: the modification time of the file as seconds since the epoch.
+            ext_*: extended attributes.  The server is not required to
+            understand this, but it may.
```
exarkun: Instead of a dictionary, this should be some kind of more structured object.  Arbitrary keys are not allowed (according to the docs) so the data type should not allow arbitrary keys.

Also, the docs here aren't complete enough for someone to know what to do here.  Why is `size` supplied when opening a file?  Do I have to make it agree with the actual size on disk of an existing file?  Is it for extending/truncating a file?  I can't tell from these docs.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+    based closely on L{twisted.conch.interfaces.ISFTPServer}
+    """
+
+    def openFile(
+        filename: str, flags: int = 0, attrs: Optional[dict] = None
+    ) -> "Deferred":
+        """
+        Called when the clients asks to open a file.
+
+        @param filename: a string representing the file to open.
```
exarkun: A for representing a path is an anti-pattern.  `FilePath` (Twisted's preferred path type) should be used here.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+class IAsyncFilesystem(Interface):
+
+    """
+    an asynchronous filesystem
+    based closely on L{twisted.conch.interfaces.ISFTPServer}
+    """
+
+    def openFile(
+        filename: str, flags: int = 0, attrs: Optional[dict] = None
+    ) -> "Deferred":
```
exarkun: I'm not sure why the return type is indicated as `"Deferred'` instead of `Deferred`.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        or a Deferred that is called back with same.
+
+        The close() method is called when the client is finished reading
+        from the directory.  At this point, the iterable will no longer
+        be used.
+
+        The iterable should return tuples of the form (filename,
+        attrs) or Deferreds that return the same.  The
+        sequence must support __getitem__, but otherwise may be any
+        'sequence-like' object.
```
exarkun: This was a bad interface when Conch defined it. It's still bad. :)  "iterable of tuples or Deferreds, also maybe indexable" is complicated and confusing and awkward to use.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+        attrs is a dictionary in the format of the attrs argument to openFile.
+
+        @param path: the directory to open.
+        """
+
+    def getAttrs(path: str, followLinks: bool) -> "Deferred":
+        """
+        Return the attributes for the given path.
```
exarkun: What are path "attributes"?  I can't tell from reading this docstring.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+        This method returns a dictionary in the same format as the attrs
+        argument to openFile or a Deferred that is called back with same.
```
exarkun: Just return a Deferred, always.  That's what the type signature says it's going to do.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        """
+
+    def removeDirectory(path: str) -> "Deferred":
+        """
+        Remove a directory (non-recursively)
+
+        It is an error to remove a directory that has files or directories in
+        it.
+
+        This method returns when the directory is removed, or a Deferred that
```
exarkun: There's `@return:`

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        called back when they are.
+
+        @param path: the path to set attributes for as a string.
+        @param attrs: a dictionary in the same format as the attrs argument to
+        L{openFile}.
+        """
+
+    def readLink(path: str) -> "Deferred":
+        """
+        Find the root of a set of symbolic links.
```
exarkun: The root?  What's that?  A set?  Where's a set?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        This method returns when the link is made, or a Deferred that
+        returns the same.
+
+        @param linkPath: the pathname of the symlink as a string.
+        @param targetPath: the path of the target of the link as a string.
+        """
+
+    def realPath(path: str) -> "Deferred":
+        """
+        Convert any path to an absolute path.
```
exarkun: What makes a path "real"?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+        This method returns the absolute path as a string, or a Deferred
+        that returns the same.
+
+        @param path: the path to convert as a string.
+        """
+
+    def unregister() -> None:
+        """
+        release system resources
```
exarkun: What resources?  Why do I want to release them?  What can I do afterwards?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def close() -> "Deferred":
+        """
+        Close the file.
+
+        This method returns nothing if the close succeeds immediately, or a
+        Deferred that is called back when the close succeeds.
+        """
+
+    def readChunk(offset: int, length: int) -> "Deferred":
```
exarkun: How about just `read`?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        """
+        Read from the file.
+
+        If EOF is reached before any data is read, raise EOFError.
```
exarkun: How about just returning an empty string, like all the other reads?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        @type consumer: C{IConsumer}
+        @param start: offset to begin reading from
+        @param chunkSize: size, in bytes, of each read
+        @return: A Deferred which fires when the file has been
+        consumed completely.
+
+        adapted from L{twisted.protocols.ftp.IReadFile}
+        """
+
+    producer = Attribute("a read-only IPushProducer object available during send()")
```
exarkun: Attributes are often not that great an idea on an interface.  I'd try to get rid of this.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+log = Logger()
+
+
+DEFAULT_BUFFER_MAX = 2 ** 12  # four megabytes
+DEFAULT_CHUNK_SIZE = 4096
+
+
+class IAsyncFilesystem(Interface):
+
```
exarkun, line 37: The interfaces should probably go somewhere else.  Maybe twisted.internet.interfaces.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        @param attrs: a dictionary in the same format as the attrs argument to
+        L{openFile}.
+        @type attrs: L{dict}
+        """
+        self.fso = fso
+        self._write_consumer: Optional["_ThreadFileConsumer"] = None
+        self._producer = _ThreadFileProducer()
+        if attrs is None:
+            attrs = {}
+        if "permissions" in attrs and not self.fso.read_only:
```
exarkun: Having conflicting ways to specify state/parameters isn't quite optimal.  I'd look for a way to change the interface to get rid of this condition.

One possibility could be to split the read and write functionality into two interfaces.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        fd = os.open(filename, flags, mode)
+        if attrs and not self.fso.read_only:
+            self.fso._setAttrs(filename, attrs)
+        self.fd = fd
+
+    def fileno(self) -> int:
+        return self.fd
+
+    def close(self) -> "Deferred":
+        if self._write_consumer:
```
exarkun, line 174: What happens if `close` is called twice?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+            d = self._write_consumer.write_deferred
+            d.addCallback(lambda _: self.fso._deferToThread(os.close, self.fd))
+            return d
+        else:
+            return self.fso._deferToThread(os.close, self.fd)
+
+    def readChunk(self, offset: int, length: int) -> "Deferred":
+        def int_read() -> bytes:
+            os.lseek(self.fd, offset, 0)
+            return os.read(self.fd, length)
```
exarkun: Looks not-thread-safe.  What if `readChunk` is called twice concurrently or called concurrently with `writeChunk`?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        s = self.fso._deferToThread(os.stat, self.fd)
+        s.addCallback(self.fso._getAttrs)
+        return s
+
+    def setAttrs(self, attrs: dict) -> "Deferred":
+        if self.fso.read_only:
+            raise AccessViolation()
+        return self.fso._deferToThread(self.fso._setAttrs, self.fd, attrs)
+
+    def flush(self) -> "Deferred":
```
exarkun: This isn't defined on the interface.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+        def int_read_loop() -> None:
+            os.lseek(self.fd, start, 0)
+            while True:
+                buf = os.read(self.fd, chunkSize)
+                if buf:
+                    blockingCallFromThread(reactor, consumer.write, buf)
+                if len(buf) < chunkSize or self._producer.stop_flag:
+                    break
+                self._producer.event.wait()
```
exarkun: This looks like it could be factored as a separate object wrapping an object with `readChunk`.  Then it could be composed with such an object to provide a basic implementation of `send` (a) without any more threading and (b) to provide a basic implementation for free for many possible implementations of `readChunk`.  It should also be more easily testable factored in such a way.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        assert self.producer
+        if self.streaming and not self._paused:
+            cast("IPushProducer", self.producer).pauseProducing()
+            self._paused = True
+        self.producer = None
+
+    def write(self, data: bytes) -> None:
+        with self.lock:
+            self._buffer += data
+            self.event.set()
```
exarkun, line 253: It would be pretty nice if there were a way to do a lock-free write so the reactor thread doesn't have to block on the non-reactor file I/O thread.  Also having to think about locking logic always makes things harder...

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+            and len(self._buffer) > self.buffer_max
+        ):
+            cast("IPushProducer", self.producer).pauseProducing()
+            self._paused = True
+
+    def close(self) -> None:
+        with self.lock:
+            self._closed = True
+            self.event.set()
+        if self.producer and self.streaming and (not self._paused):
```
exarkun, line 268: What if `close` is called twice?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        if threadpool:
+            self.threadpool = threadpool
+        else:
+            self.threadpool = reactor.getThreadPool()  # type: ignore
+
+    def unregister(self) -> None:
+        pass
+
+    def _setAttrs(self, path: str, attrs: dict) -> bool:
+        if "uid" in attrs and "gid" in attrs:
```
exarkun: What if only one is there?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+        @param read_only: if C{True}, changes to filesystem forbidden
+        @type read_only: C{bool}
+        """
+        self.root = root
+        self.read_only = read_only
+        if threadpool:
+            self.threadpool = threadpool
+        else:
+            self.threadpool = reactor.getThreadPool()  # type: ignore
```
exarkun: I suspect sharing the reactor threadpool here is going to be problematic.  We should carefully consider this default (and explain it with a comment if we decide to keep it).  The reactor threadpool is generally for light-duty usage by short-duration.  Things like `_ThreadFileConsumer` are going to consume a thread from the pool for a longer period of time.  The reactor thread pool is small.  It's kind of easy to imagine an application where this becomes a problem.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+from twisted.internet.threads import (
+    deferToThreadPool,
+    blockingCallFromThread,
+)
+from twisted.internet.defer import Deferred
+from twisted.logger import Logger
+from typing import Tuple, Iterable, Optional, Callable, cast, Any
+import os
+import os.path
+import threading
```
exarkun, line 10: imports in Twisted code are arranged, top to bottom:

 * stdlib
 * third-party
 * twisted imports

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+            permissions: the permissions of the file with as an integer.
+            the bit representation of this field is defined by POSIX.
+            atime: the access time of the file as seconds since the epoch.
+            mtime: the modification time of the file as seconds since the epoch.
+            ext_*: extended attributes.  The server is not required to
+            understand this, but it may.
+
+        NOTE: there is no way to indicate text or binary files.
+
+        This method returns an object that meets the IFile interface.
```
exarkun: We say that an object that has methods and attributes as described by an interface **provides** that interface.  A class which has instances that provide an interface **implements** the interface.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+            uid: the user ID of the file as an integer
+            gid: the group ID of the file as an integer
+            permissions: the permissions of the file with as an integer.
+            the bit representation of this field is defined by POSIX.
+            atime: the access time of the file as seconds since the epoch.
+            mtime: the modification time of the file as seconds since the epoch.
+            ext_*: extended attributes.  The server is not required to
+            understand this, but it may.
+
+        NOTE: there is no way to indicate text or binary files.
```
exarkun: So all files are text?  All files are binary?  Files type is automatically detected?  File type is platform dependent?  Filesystem dependent?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        Produce the contents of the file to the given consumer.
+
+        @type consumer: C{IConsumer}
+        @param start: offset to begin reading from
+        @param chunkSize: size, in bytes, of each read
+        @return: A Deferred which fires when the file has been
+        consumed completely.
+
+        adapted from L{twisted.protocols.ftp.IReadFile}
+        """
```
exarkun: Depending on various implementation details, I think it will probably make the most sense to implement `readChunk` in terms of send.  If `send` also accepts a length then all `readChunk` does is adapt the output/return type to a single string (from a bunch of writes on a consumer).  The underlying implementation should otherwise be identical (and therefore should just be the same).

This also means `readChunk` can come for free for all implementations that provide `send`.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+    producer = Attribute("a read-only IPushProducer object available during send()")
+
+    def receive(append: bool = False) -> "IConsumer":
```
exarkun: Analogous comment here about `receive` / `writeChunk` as I left above about `send` / `readChunk`.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        if "permissions" in attrs and not self.fso.read_only:
+            mode = attrs["permissions"]
+            del attrs["permissions"]
+        else:
+            mode = 0o644
+        if self.fso.read_only:
+            flags = os.O_RDONLY
+        elif flags == 0:
+            flags = os.O_RDWR
+        fd = os.open(filename, flags, mode)
```
exarkun: At first I thought this I/O was happening in the wrong thread.  Eventually I noticed that `_ThreadFile` is called in a non-reactor thread.  So that's fine, I guess.  It would be good to document that `_ThreadFile.__init__` runs in a non-reactor thread.

Also, I/O inside `__init__` is kind of an anti-pattern.  I suggest a classmethod constructor that does the I/O and leaves `__init__` to just setting attributes.

Speaking of just setting attributes, how about some `@attr.s` usage here?  It would probably shave down the linecount.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+        read_only: bool = False,
+    ) -> None:
+        """
+        @param root: the base directory
+        @type root: L{str}
+
+        @param threadpool: the threadpool to use, default the reactor's threadpool
+        @type threadpool: L{twisted.python.threadpool.ThreadPool}
+
+        @param read_only: if C{True}, changes to filesystem forbidden
```
exarkun: This docstring looks incomplete, maybe?  It reads as incomplete, anyway.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def unregister(self) -> None:
+        pass
+
+    def _setAttrs(self, path: str, attrs: dict) -> bool:
+        if "uid" in attrs and "gid" in attrs:
+            os.chown(path, attrs["uid"], attrs["gid"])
+        if "permissions" in attrs:
+            os.chmod(path, attrs["permissions"])
+        if "atime" in attrs and "mtime" in attrs:
```
exarkun: What if only one of them is present?

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,709 @@
...
+            if s.st_birthtime:
+                d["ext_birthtime"] = s.st_birthtime
+        except AttributeError:
+            # not available on all platforms
+            pass
+        return d
+
+    def _absPath(self, path: str) -> str:
+        p = os.path.normpath(os.path.join(self.root, path))
+        if not p.startswith(self.root):
```
exarkun: What if `self.root == "/foo"` and `p == "/foobar"`?  It would be good to use FilePath everywhere to avoid this kind of problem.

src/twisted/internet/test/test_asyncfs.py
```diff
@@ -0,0 +1,341 @@
...
+            d2.addCallback(cb_pro2)
+            return d2
+
+        d1 = self.fso.openFile("bigfile.txt", os.O_CREAT | os.O_WRONLY)
+        d1.addCallback(cb_pro1)
+        return d1
+
+
+class TestThreadFs(BaseTestFs, unittest.TestCase):
+    fso_class = asyncfs.ThreadFs
```
exarkun: Mixins are an anti-pattern.  For this kind of test sharing (speculative at this point, I guess), composition is preferable.  I know unittest doesn't make it easy but it's still better than dealing with the complexity of mixins.

Also, I'd suggest that stateful Hypothesis tests might just make this test suite shorter, easier to read, and more comprehensive.

exarkun (commented):
Thanks for your work on this.  I've left some comments inline.  I ran out of time before I got to have a good look at the test suite.  However, I notice codecov is reporting a lot of uncovered lines so maybe there's more work to do on the test suite anyway.

adiroiban (commented):
src/twisted/newsfragments/9947.misc  -  I think that this should be removed.

Maybe it's better to start working on this API  in a repo like twisted/txasyncfs

we can then play with the API without having to worry about backward compatibility.

I am not sure what is the scope / use cases for this API.

Do we need readLink support for every filesystem type ?

I would go with the most simple interface that can read or write data and then extend to more complex interfaces

Is this for local OS filesystem access or also remote filesystem,  server-side usage only, client-side usage only, mixed ?

Should it handle Unix, Windows, ZOS (`//'PDEB.OU11.DRSAFG(+1)'`) paths?

-------

I have implementation for server-side and client-side async file access API for FTP, SFTP, HTTP(Rest,WebDAV,AS2), Azure Files REST API ... but code is a mess and not proud of it.

Below is a crude export 

https://gist.github.com/adiroiban/f8c0324662aa4bbff12fe959e40ed43f

I am happy to see a better interface... I don't know how the perfect interface should look like, but I can help with feedback of what went wrong with each implementation

maybe you have some time to drop by IRC and we can have a chat

Thanks!

ihaywood3:
> src/twisted/newsfragments/9947.misc - I think that this should be removed.

whoops this should have been removed. 

> Maybe it's better to start working on this API in a repo like twisted/txasyncfs
> we can then play with the API without having to worry about backward compatibility.

I would really like this to be in twisted proper as I see it as a base for other work such as a io_uring reactor. 

> I am not sure what is the scope / use cases for this API.

its for protocols that make heavy use of the local filesystem (FTP, t.web.static.File) to have a suitable abstraction to do async file I/O

> Do we need readLink support for every filesystem type ?

It's basically an async wrapper around the Python `os`  module so the underlying filesystem type should be opaque. If `os` doesn't offer it a Failure wrapping AttributeError would be returned.  

> Is this for local OS filesystem access or also remote filesystem, server-side usage only, client-side usage only, mixed ?

local OS only.  

> Should it handle Unix, Windows, ZOS (`//'PDEB.OU11.DRSAFG(+1)'`) paths?

paths get passed straight through to `os.open`, etc so it will handle whatever those functions handle.

src/twisted/internet/asyncfs.py
ihaywood3: I would prefer the other way round as the reading primitive on `io_uring` (and `os` for threads) is like `readChunk`. For the `io_uring` implementation (and probably others) `send` I would implement in `readChunk`.  This could be done for threads also but I assumed (possibly wrongly) that a single long-running task was important for efficiency.

ihaywood3:
Thank you both for your reviews. The general answer to "why is this way?" is that I had assumed it was important to stick closely to the SFTP interface. To my pleasant surprise you are happy to deviate from this. I will get on with implementing exarkun's recommendations.

src/twisted/internet/asyncfs.py
adiroiban: I guess that at the low level this will call scandir  - https://docs.python.org/3.10/library/os.html#os.scandir  so maybe defined it based on that API https://docs.python.org/3.10/library/os.html#os.DirEntry

adiroiban (commented):
Thanks for the update.

> I would really like this to be in twisted proper as I see it as a base for other work such as a io_uring reactor.
n
It can be added later to twisted/twisted, once the API is demonstrated to be stable and we have various implementation of the interface and code using those implementations.

Even if this is added directly to twisted/twisted I don't think that such a big interface is a good idea.

I prefer to split this into multiple smaller interface .... IAsyncRead, IAsyncWrite, IAsyncList, IAsyncGetAttributes,  IAsyncSetAttributes  .. golang style :)

> its for protocols that make heavy use of the local filesystem (FTP, t.web.static.File) to have a suitable abstraction to do async file I/O

But it's for the server-side or client-side of the protocols?

Why limit the interface to network protocols?

Why the interface shouldn't be used for other background jobs/task that are not necessarily tied to a remote peer?

> It's basically an async wrapper around the Python os module so the underlying filesystem type should be opaque. If os doesn't offer it a Failure wrapping AttributeError would be returned.

If it's just a wrapper for a specific os API then why create the interface in the first place?

My understanding is that you create an interface when you want to use Python OS stdlib as in implementation , but also allow other non-stdlib implementations ... maybe something based on win32file 

-----------

We already have FilePath which to me looks like an object oriented wrapper around the same os module code.

The IFilePath is generic but its not async.

I have never used IFilePath in my code as I needed async methods, but to mee the IFilePath API doesn't look that bad.

If a new local os file access API interface is required I think that it would be important to at a "current state of the art" analysis and see what is bad and what is good with that API.

Having a similar API might make the existing code depending on IFIlePath easier to port to the new Async API.

This is just my external feedback as I have not used IFilePath.

> local OS only.

but local OS only like  just mirror the low level Python OS an implementation or a portable interface and implementation that uses the lower denominator between posix and nt features.

I would lilke to see an interface in which I can do asysnc / consume-producer file operations and don't care whether the "file" is local or remote... also I don't want to care (add extra ifs) for linux vs unix vs mac access.

To give an example in both Windows and Linux a file has an inode ID or symbolic/hard link... but for windows you need to use win32file to handle junction points or read the  hard links count or node id

----------

Then there is also the aspect of impersonated file access for server-side code.
For example on Windows, I am running the main code as `srv_twsited` local account, but the files access is made  based on each authenticated user.

Does the new interface / impementation care for that use case?
As long as you don't use thread it shouldn't matter... but once you start using thread with impersonation things get ugly.

----------

As a conclusion I think that it's important to take into consideration the current IFilePath interface and AbstractFilePath implementation and see what they did good and did bad.

For example I see that walk() is not part of any IFilePath related interface

-----------

I am not interested in creating a local OS filesystem only async interface as I don't plan to use such a code.
So I can't provide much feedback or constructive comments.

This doesn't mean that I am against such an implementation or interface. 
It's just that I can't be a good reviewer for this PR.

My first suggestion based on past experience with Twisted development is to try to push work in small changes.
Break the interface and push small PR.
Otherwise it will be hard to find another Twisted dev to review and approve a huge PR.

There is also a technical reason for smaller interfaces.
With Twisted backward compatibility policy it's impossible to make any changes to an existing interface... the only option is to deprecated it and add a new one.

ihaywood3:
> Even if this is added directly to twisted/twisted I don't think that such a big interface is a good idea.
> 
> I prefer to split this into multiple smaller interface .... IAsyncRead, IAsyncWrite, IAsyncList, IAsyncGetAttributes, IAsyncSetAttributes .. golang style :)

Now it's clear there's no tie to an existing interface this sounds good. I will start with just `IAsyncRead` / `IAsyncWrite`. I assume the constructors would take an `IFilePath`? 

> > its for protocols that make heavy use of the local filesystem (FTP, t.web.static.File) to have a suitable abstraction to do async file I/O
> 
> But it's for the server-side or client-side of the protocols?
> 
> Why limit the interface to network protocols?
> 
> Why the interface shouldn't be used for other background jobs/task that are not necessarily tied to a remote peer?

That's simply my main anticipated use-case because my interest comes from writing a SMB server, but it's not restricted to network servers at all. 
 

> We already have FilePath which to me looks like an object oriented wrapper around the same os module code.
> 
> The IFilePath is generic but its not async.

It would be possible to make an "async version" of `IFilePath` but it would run into the same problem as basing it on `ISFTPServer`: it would be too much functionality in one class, I (now) prefer your earlier approach. 

> I would lilke to see an interface in which I can do asysnc / consume-producer file operations and don't care whether the "file" is local or remote... also I don't want to care (add extra ifs) for linux vs unix vs mac access.

If you want /remote/ access to Windows servers, you're reviewing the wrong PR of mine ;-) 

https://github.com/twisted/twisted/pull/1274

> Then there is also the aspect of impersonated file access for server-side code.
> For example on Windows, I am running the main code as `srv_twsited` local account, but the files access is made based on each authenticated user.
> 
> Does the new interface / impementation care for that use case?
> As long as you don't use thread it shouldn't matter... but once you start using thread with impersonation things get ugly.

I had relied on the `os` module to abstract away Windows / posix issues. Other implementations wouldn't use threads and would be OS-specific.  Thanks for the heads up about threads and impersonation. 
 
> I am not interested in creating a local OS filesystem only async interface as I don't plan to use such a code.

later on it would be possible to extend the interface to accept URLs instead of local paths, and so the appropriate network client, veiled behind `IAsyncRead` / `IAsyncWrite`. 

> My first suggestion based on past experience with Twisted development is to try to push work in small changes.

reading / writing is the minimum sensible functionality so I will pare this PR back to just that.

adiroiban:
@ihaywood3 
> If you want /remote/ access to Windows servers, you're reviewing the wrong PR of mine ;-)
> 
> #1274

Hi Ian. I am not very interested in a Twisted only SMB protocol implementation. Sorry about that.

I am following the development of https://github.com/jborean93/smbprotocol and I hope that I could find some time to contribute some patched to that project to make it easier to use it with twisted or asyncio or trio

See https://github.com/jborean93/smbprotocol/issues/38#issuecomment-643663275

src/twisted/internet/asyncfs.py
ihaywood3: It has quite different semantics to a traditional `read()` there's no file pointer the offset is required for every call.

exarkun: That doesn't seem like a big deal to me.  The reverse question would be "What's a chunk?"  The function name will never convey all important information about the function - that's why there's other documentation.

src/twisted/internet/asyncfs.py
exarkun: Twisted's threading primitives are pool-based so there's little or no difference between a long-lived task and N short-lived tasks from an efficiency perspective.  Either way you have to pass messages between the reactor thread and a worker thread.  There _is_ a difference in contention for threads in the pool though.  If you have long-lived tasks hogging all of the threads in the pool then no short-lived tasks can run until some long-lived tasks finish.

So it's preferable to have short-lived tasks in default/global thread pool.

If there is an efficiency argument to be made (and maybe there *shouldn't* be - at least not without benchmarks), it is probably that the interface should not involved the creation of an additional Deferred for every piece of data that moves through the system because creating, using, and destroying a Deferred is significantly more expensive than shoveling a `bytes` object from one place to another.  With `send` I can make anywhere from 1 to N Deferreds by choosing how much data to work on but *without* committing myself to holding all of that data in memory at once.  With `writeChunk` I have to balance number of Deferreds created against amount of data held in memory at a time.

If there is some argument based on the underlying system primitives (ie `io_uring` or Windows IOCP) then perhaps `send` and `writeChunk` should be on separate interfaces.  An implementer can choose whichever best fits the underlying abstraction and we can provide adapters from each to the other for use when one interface or the other is better suited to the application code (although speaking of application code, I wonder if txtubes is the *actually* correct application-facing interface).

exarkun:
> > I am not sure what is the scope / use cases for this API.
> its for protocols that make heavy use of the local filesystem (FTP, t.web.static.File) to have a suitable abstraction to do async file I/O

This information would make a great addition to the ticket and the module / interface class docstrings. :)

> Now it's clear there's no tie to an existing interface this sounds good.
> I will start with just IAsyncRead / IAsyncWrite. >
 I assume the constructors would take an IFilePath?

Keep in mind that Zope Interfaces typically don't make any statement about constructors.
There is (and should be) no ``IAsyncRead.__init__``.
If there is a particular implementation of ``IAsyncRead`` for local filesystem operations
(like the thread-based one here)
then accepting an ``IFilePath`` as an argument may indeed make sense.

> It would be possible to make an "async version" of ``IFilePath`` but ...

I agree that the smaller interfaces are better than trying to make a monolithic async variation of ``IFilePath``.
Also, it was probably a mistake for ``IFilePath`` to mix path manipulation operations with I/O operations.

Since this new async functionality implements all of the I/O operations itself,
it seems likely that it would *only* want to consume the path manipulation parts of ``IFilePath``.
I think it should do that I just wanted to note that not all of ``IFilePath`` is going to be useful.

> > Then there is also the aspect of impersonated file access for server-side code.
> > For example on Windows, I am running the main code as srv_twsited local account,
> > but the files access is made based on each authenticated user.
> > Does the new interface / impementation care for that use case?
> > As long as you don't use thread it shouldn't matter...
> > but once you start using thread with impersonation things get ugly.

If the platform doesn't offer any actual support for this kind of thing then there's not much Twisted can do about it,
is there?

Conch tried to deal with this by playing with set(e)uid but it's basically broken because:

(1) Conch, a library, cannot guarantee that there will be no threaded use anywhere in the process using it; and                                                                                                                                                                                                                                                                             
(2) Conch itself uses threads!

If Windows IOCP or io_uring or whatever the thing on macOS is/will be supports this kind of impersonation then we can expose that...
If they don't, we're probably not going to invent a system that adds it.
If some do and some don't we can offer an API with graceful degradation but I think that's about it?
If you have specific ideas in this area, please feel free to share. :)

> I had relied on the os module to abstract away Windows / posix issues.
> Other implementations wouldn't use threads and would be OS-specific.
> Thanks for the heads up about threads and impersonation.

The particular interface we settle on should account for the need for graceful degradation -
if there are actually any platforms that do anything useful here.

The only thing you can do on POSIX is set(e)uid to change the permissions of your whole process.
On Windows, maybe ``RunImpersonated`` is relevant?
https://docs.microsoft.com/en-us/dotnet/api/system.security.principal.windowsidentity.runimpersonated?view=netframework-4.8

That interface works by having you submit work for execution and the executor makes sure identity/permissions are set correctly for that work (and not other work).

Note the interface is not specific to filesystem access so in some ways this goes beyond the scope of this async filesystem I/O effort.
Instead, we could probably imagine some cross-platform::

```
runImpersonated(identity, functionReturningDeferred)
```

where ``functionReturningDeferred`` and all its callbacks added before `runImpersonated` receives the `Deferred` run with some altered credentials.

But you still have problems with threads so you still may not be able to do this anywhere except on Windows.

But I suggest further efforts on the impersonation feature get split off to a separate ticket/PR.

> > My first suggestion based on past experience with Twisted development is to try to push work in small changes.
> reading / writing is the minimum sensible functionality so I will pare this PR back to just that.

Maybe you can get even smaller than that - "reading" is sometimes useful all by itself, for example.
Perhaps this is taking the idea of minimization further than necessary but considering the large *conceptual* surface area of this change,
it might be worth driving the size of the *implementation* down as far as possible,
at least for the first PR or two.

exarkun:
Sorry, I know I wrote a lot already, but this one comment got lost in my attempts to edit out other redundancy:

The eventual plan for an io_uring-based reactor be fleshed out just a little bit so it is clear how these pieces will eventually fit with that.  I'm not suggesting there's going to be any kind of problem with that eventual integration but as far as I know right now the plan exists only in your (ihaywood3) head - so it's difficult or impossible for anyone else to review PRs like this one with respect to that goal (so you're going to miss out on at least part of the value of the code review process).

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,287 @@
...
+    L{twisted.internet.interfaces.IReactorFileAsync} this is used,
+    otherwise fallback to the threaded implementation.
+
+    @param pool: the threadpool to use if threads are used. Default is the reactor's
+    own threadpool. B{WARNING:} Reactor threadpools are generally for light work, you will
+    usually need to provide a pool.
+
+    @return: a Deferred returning L{IAsyncWriter} or L{IAsyncReader}
+    """
+    if not user_reactor:
```
exarkun: `if user_reactor is None:`

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,287 @@
...
+        user_reactor = reactor
+    if IReactorFileAsync in providedBy(user_reactor):  # pragma: no cover
```
exarkun: Why not IReactorFileAsync.providedBy(user_reactor)?  Or, preferably:

```
IReactorFileAsync(user_reactor, threadedFallbackReactor).openAsync(path, flags)
```

Also, this doesn't seem like a reasonable place for a `no cover` pragma.

src/twisted/internet/asyncfs.py
```diff
@@ -0,0 +1,287 @@
...
+from twisted.internet.defer import Deferred
+from twisted.logger import Logger
+
+log = Logger()
+
+
+def openAsync(
+    path: FilePath,
+    flags: FileAsyncFlags,
+    user_reactor: Any = None,
```
exarkun: `user_reactor` should be `userReactor`.  Or probably just `reactor`?  That is the conventional name.

ihaywood3: surely just `reactor`  would clash with `from twisted.internet import reactor`?
`userReactor` is fine

exarkun: Yea.  Maybe don't have a default at all and remove the global reactor import from the module entirely.

ihaywood3: `openAsync` is intended as a conveience function, it seems odd to require callers to provide the reactor every time.

ihaywood3:
Is there anything else to do to progress this PR?

src/twisted/internet/interfaces.py
```diff
@@ -2707,3 +2710,130 @@ class _ISupportsExitSignalCapturing(Interface):
...
+DEFAULT_CHUNK_SIZE = 4096
+DEFAULT_BUFFER_MAX = 2 ** 22  # 4 megabytes
+
+
+class IAsyncReader(Interface):
+    """
+    a file or file-like object opened for reading asynchronously
+    """
+
+    def close() -> "Deferred":
```
graingert, line 2814: I think this should return an Awaitable or a Coroutine

graingert:
```suggestion
    def aclose() -> Awaitable[object]:
```

then it can be used in ```async with contextlib.aclosing(reader):```

graingert: IAsyncReader should also be a subtype of contextlib.AsyncContextManager[T]

src/twisted/internet/test/test_asyncfs.py
```diff
@@ -0,0 +1,230 @@
...
+        self.oldpath = os.getcwd()
+        self.tpath = tempfile.mkdtemp()
+        os.chdir(self.tpath)
+        with open("one.txt", "w") as fd:
+            fd.write("blah" * 3)
+        with open("two.txt", "w") as fd:
+            fd.write("blah")
+
+    def tearDown(self):
+        for i in os.listdir():
```
graingert, line 77: use tempfile.TemporaryDirectory

src/twisted/internet/interfaces.py
```diff
@@ -2707,3 +2710,130 @@ class _ISupportsExitSignalCapturing(Interface):
...
+    CREATE = 8
+    EXCLUSIVE = 16
+
+
+class IReactorFileAsync(Interface):
+    """
+    interface for reactors where the OS offers an async file I/O API
+    """
+
+    def openAsync(path: "FilePath", flags: FileAsyncFlags) -> "Deferred":
```
graingert, line 2917:
```python

if TYPE_CHECKING:
    from _typeshed import OpenBinaryModeWriting, OpenBinaryModeReading, OpenBinaryModeUpdating,  OpenBinaryMode
else:
    OpenBinaryModeWriting = OpenBinaryModeReading = OpenBinaryModeUpdating = OpenBinaryMode = object

class IReactorFileAsync(Interface):
    @overload
    def openAsync(path: os.PathLike, flags: OpenBinaryModeReading) -> Awaitable[IReadAble]: ...

    @overload
    def openAsync(path: os.PathLike, flags: OpenBinaryModeWriting) -> Awaitable[IWriteAble]: ...

    @overload
    def openAsync(path: os.PathLike, flags: OpenBinaryModeUpdating) -> Awaitable[IReadWriteAble]: ...

    def openAsync(path: os.PathLike, flags: OpenBinaryMode) -> Awaitable[Union[IReadAble, IReadWriteAble, IReadWriteAble]]):
        """..."""
    
```

src/twisted/internet/test/test_asyncfs.py
```diff
@@ -0,0 +1,230 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details..
+# type: ignore
```
graingert, line 3:
```suggestion
```

src/twisted/internet/asyncfs.py
graingert: could you use a `queue.Queue(1).put_nowait` here ?

glyph (changes requested):
This PR still has a bunch of existing un-responded-to feedback and failing tests, so it doesn't really need another review until all that is dealt with.

PR #1548, opened 2021-03-14
Report both runner and tox python info on GHA

altendky:
Draft for:
- [ ] ~~Remove existing in-tox reporting~~
  - Discussed in https://github.com/twisted/twisted/pull/1548#issuecomment-806029186
- [x] Add an empty newsfragment
- [x] Relevant changes from #1545
- [x] Test fix in https://github.com/twisted/twisted/pull/1557

## Scope and purpose

When debugging it is useful to have piles of info you don't care about...  because that's about the only way to also have the one tidbit you do need.  This PR adds reporting of information about the Python and environment used in the tox run as well as still reporting the same for the outer GitHub Actions runner.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10129


altendky:
Kick CI.

altendky:
@adiroiban, so perhaps the real value would be not in the `python-env` parameter but in `tox-env`?  Or maybe this is the point I missed in your request.  Having `python-info-action` take `tox-env` as it can be passed on the cli to tox would allow the action to comma-split and report for each specified env...

adiroiban:
I think that reporting tox via python info action only once is enough :)

Note that here https://github.com/twisted/twisted/pull/1545/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R23  my suggestion is to have a simple tox-env variables and not a list

Also in the same PR admin/dump_all_version_info.py was updated to show deps.

altendky:
I checked and the runtime was 1s in the action's own tests so I figured we may as well have both outside and inside info.  Especially what with tox stripping environment variables.  Anyways, I can wait until #1545 gets shaken out and then see what it looks like here.

altendky:
Looks like reasonable overhead here as well.

image

My comment about tox stripping env vars is irrelevant for a couple reasons.

https://github.com/twisted/twisted/blob/1cf3b6ca4901efa16e0292fc3462ed3fb2e33911/tox.ini#L79-L80

Also, the action is running the Python from the env, but not 'in' the tox env where variables would be stripped.  None the less, it does get you the Python info for each case including prefixes and installed packages.  If the action grew the ability to run 'in the tox env' then that'd fall through with this present setup (and then it would show the extra env vars that get added).

In the draft list in the OP I added removal of the existing reporting of info in the tox env.  But, then you don't get anything locally.  At some point when the action grows a locally runnable copy of the code (theoretically, maybe, someday, 'ish) then perhaps this duplication would be revisited.  For now, leave it I guess.

altendky:
I can't say I understand why coverage is dropping...

PR #1551, opened 2021-03-16
[3795] Fix quadradics in _ChunkedTransferDecoder

twm:
## Scope and purpose

There were two issues here:

- Buffering was done by appending to a `bytes`, which as an immutable object may trigger reallocation and quadratic amounts copying. This is resolved by using `bytearray`, which is mutable with amortized O(n) append complexity. It's also [amortized O(n) to truncate it at the front ][1], which we make use of.
- Processing of the length line searched for `\r\n` on the entire input each time data arrived, resulting in quadratic work if input trickled in. This is solved by tracking the extent of the prior search and resuming it at that point.

The fix can be verified by running the proof-of-concept script attached to the ticket.

[1]: https://github.com/python-hyper/h11/blob/82d5e461bd8a3eca6f1aaa51a1c785a1779721a5/h11/_receivebuffer.py#L22-L26

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/3795

```
Merge pull request #1551 from twisted/3795-quadratic-http-chunking

Author: twm
Reviewer: adiroiban
Fixes: ticket:3795

Fix quadratic time complexity in twisted.web.http._ChunkedTransferDecoder
```

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1870,12 +1870,13 @@ class _ChunkedTransferDecoder:
 
     def _dataReceived_CRLF(self) -> bool:
         """
-        Await the carriage return and line feed characters that follow the
+        Await the carriage return and line feed characters that are the end of chunk marker that follow the
         chunk data.
 
         @returns: C{True} when the CRLF have been read, otherwise C{False}.
         """
         # TODO: https://twistedmatrix.com/trac/ticket/10137
+        # It should raise an error when receiving malformed end of chunk markers.
         if self._buffer.startswith(b"\r\n"):
             self.state = "CHUNK_LENGTH"
             del self._buffer[0:2]
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1869,14 +1869,29 @@ class _ChunkedTransferDecoder:
         return True
 
     def _dataReceived_CRLF(self) -> bool:
+        """
+        Await the carriage return and line feed characters that follow the
+        chunk data.
+
+        @returns: C{True} when the CRLF have been read, otherwise C{False}.
+        """
+        # TODO: https://twistedmatrix.com/trac/ticket/10137
         if self._buffer.startswith(b"\r\n"):
             self.state = "CHUNK_LENGTH"
             del self._buffer[0:2]
             return True
-        else:
-            return False
+        return False
 
     def _dataReceived_TRAILER(self) -> bool:
+        """
+        Await the carriage return and line feed characters that follow the
+        terminal zero-length chunk. Then invoke C{finishCallback} and switch to
+        state C{'FINISHED'}.
+
+        @returns: C{False}, as there is either insufficient data to continue,
+            or no data remains.
+        """
+        # TODO: https://twistedmatrix.com/trac/ticket/10137
         if self._buffer.startswith(b"\r\n"):
             data = memoryview(self._buffer)[2:].tobytes()
             del self._buffer[:]
@@ -1885,6 +1900,12 @@ class _ChunkedTransferDecoder:
         return False
 
     def _dataReceived_BODY(self) -> bool:
+        """
+        Deliver any available chunk data to the C{dataCallback}. When all the
+        remaining data for the chunk arrives, switch to state C{'CRLF'}.
+
+        @returns: C{True} to continue processing of any buffered data.
+        """
         if len(self._buffer) >= self.length:
             chunk = memoryview(self._buffer)[: self.length].tobytes()
             del self._buffer[: self.length]
@@ -1893,11 +1914,16 @@ class _ChunkedTransferDecoder:
         else:
             chunk = bytes(self._buffer)
             self.length -= len(chunk)
-            self._buffer = bytearray()
+            del self._buffer[:]
             self.dataCallback(chunk)
         return True
 
     def _dataReceived_FINISHED(self) -> bool:
+        """
+        Once C{finishCallback} has been invoked receipt of additional data
+        raises L{RuntimeError} because it represents a programming error in
+        the caller.
+        """
         raise RuntimeError(
             "_ChunkedTransferDecoder.dataReceived called after last "
             "chunk was processed"
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1888,12 +1888,13 @@ class _ChunkedTransferDecoder:
         if len(self._buffer) >= self.length:
             chunk = memoryview(self._buffer)[: self.length].tobytes()
             del self._buffer[: self.length]
-            self.dataCallback(chunk)
             self.state = "CRLF"
+            self.dataCallback(chunk)
         elif self._buffer:
-            self.dataCallback(bytes(self._buffer))
-            self.length -= len(self._buffer)
+            chunk = bytes(self._buffer)
+            self.length -= len(chunk)
             self._buffer = bytearray()
+            self.dataCallback(chunk)
         return True
 
     def _dataReceived_FINISHED(self) -> bool:
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1890,7 +1890,7 @@ class _ChunkedTransferDecoder:
             del self._buffer[: self.length]
             self.state = "CRLF"
             self.dataCallback(chunk)
-        elif self._buffer:
+        else:
             chunk = bytes(self._buffer)
             self.length -= len(chunk)
             self._buffer = bytearray()
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -100,8 +100,6 @@ import cgi
 import math
 import os
 import re
-
-# system imports
 import tempfile
 import time
 import warnings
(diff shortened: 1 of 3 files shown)


twm:
Whoa, CodeCov is good now.

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,97 @@ class _ChunkedTransferDecoder:
...
-        return b""
+        return False
 
-    def _dataReceived_BODY(self, data):
-        if len(data) >= self.length:
-            chunk, data = data[: self.length], data[self.length :]
-            self.dataCallback(chunk)
+    def _dataReceived_BODY(self) -> bool:
+        if len(self._buffer) >= self.length:
+            chunk = memoryview(self._buffer)[: self.length].tobytes()
```
adiroiban, line 1911: Maybe a comment would help to explain what each branch is handling.

```suggestion
        if len(self._buffer) >= self.length:
            # All data received for the current chunk.
            chunk = memoryview(self._buffer)[: self.length].tobytes()
```

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,97 @@ class _ChunkedTransferDecoder:
...
-            self.length -= len(data)
-            self.dataCallback(data)
-            return b""
+            self.dataCallback(chunk)
+        elif self._buffer:
+            chunk = bytes(self._buffer)
+            self.length -= len(chunk)
+            self._buffer = bytearray()
+            self.dataCallback(chunk)
+        return True
```
adiroiban: Just a minor style suggestion.

I think that this refactoring will also solve the missing branch coverage that is preventing this diff for being 100%

In Codecov, yellow is missing branch coverage.

https://app.codecov.io/gh/twisted/twisted/compare/1551/diff#D1L1893

```suggestion
            self.dataCallback(chunk)
            return True

        if not self._buffer:
             # No data received yet for the current chunk.
            return True

        # We have data and we are not at the end of the chunk.
        chunk = bytes(self._buffer)
        self.length -= len(chunk)
        self._buffer = bytearray()
        self.dataCallback(chunk)
        return True
```

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,97 @@ class _ChunkedTransferDecoder:
...
-            return b""
+            return False
 
-    def _dataReceived_TRAILER(self, data):
-        if data.startswith(b"\r\n"):
-            data = data[2:]
+    def _dataReceived_TRAILER(self) -> bool:
+        if self._buffer.startswith(b"\r\n"):
+            data = memoryview(self._buffer)[2:].tobytes()
+            del self._buffer[:]
```
adiroiban, line 1898: I am not  sure why use `del` vs `self._buffer = bytearray()`

I understand the different between them, but is there a difference for the logic of this function?

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,97 @@ class _ChunkedTransferDecoder:
...
             self.state = "FINISHED"
             self.finishCallback(data)
-        else:
-            self._buffer = data
-        return b""
+        return False
```
adiroiban, line 1901: Maybe raise an exception if we are in the trailer, and the buffer and 2 or more bytes...or buffer has one byte and that byte is not `\r`

I guess that in this case the handling will raise a timeout, as the END marker is never receive.

But maybe raising an exception will prevent receiving more invalid data from the connection.

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,97 @@ class _ChunkedTransferDecoder:
...
+    def _dataReceived_CRLF(self) -> bool:
+        if self._buffer.startswith(b"\r\n"):
             self.state = "CHUNK_LENGTH"
-            return data[2:]
+            del self._buffer[0:2]
+            return True
         else:
-            self._buffer = data
-            return b""
+            return False
```
adiroiban: Maybe remove else and reduce indentation.

A similar comment to `TRAILER` if self._buffer has size 1 or 2 and don't start with `\r` or \r\n` maybe fail fast and receiving more data make no sense.

I see that this is the same behaviour as trunk... so it's ok to have it fixed in a separate ticket.

src/twisted/web/test/test_http.py
```diff
@@ -1240,6 +1240,34 @@ def test_short(self):
         self.assertEqual(L, [b"a", b"b", b"c", b"1", b"2", b"3", b"4", b"5"])
         self.assertEqual(finished, [b""])
 
+    def test_long(self):
```
adiroiban, line 1243: Very minor suggestion
```suggestion
    def test_fragmented_chunk(self):
```

adiroiban (approved):
This looks good.
I only have minor comments.

I am now running some manual tests and will report later.

Thanks!

adiroiban:
Manual tests looks good.

In one terminal I have `$ twistd -n web --port tcp:8080 --path .`
In another one `$ python poc-3795-py3.py`

In trunk I can see the CPU usage increasing for  the `twistd` process.
In this branch the CPU is minimal

src/twisted/web/http.py
twm: I don't have a coherent rationale; I was going to check if `del self._buffer[:]` retains the allocated buffer at all before adding the review keyword.

src/twisted/web/http.py
twm: I'll add a docstring, since this function lacks one.

src/twisted/web/http.py
twm: I changed this to consistently use `del self._buffer[:]` instead of allocating a fresh buffer.

src/twisted/web/test/test_http.py
twm: `test_long` is the opposite of `test_short` though. :)

src/twisted/web/http.py
twm: Thanks for pointing this out. I filed https://twistedmatrix.com/trac/ticket/10137 and added some comments referencing it.

There is a potential similar issue in the chunk length line. As I was tinkering with it I wondered if there should be a length limit.

twm:
Hi @adiroiban, thank you for the review. Unfortunately I wasn't quite ready for it &mdash; I was working through the checklist when I noticed a legitimate coverage miss, and forgot to remove the review request. I have addressed your comments and submitted for review properly, so I hope this one is easy.

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,123 @@ class _ChunkedTransferDecoder:
...
+        return True
+
+    def _dataReceived_CRLF(self) -> bool:
+        """
+        Await the carriage return and line feed characters that follow the
+        chunk data.
+
+        @returns: C{True} when the CRLF have been read, otherwise C{False}.
+        """
+        # TODO: https://twistedmatrix.com/trac/ticket/10137
```
adiroiban, line 1878: thanks for the todo.
When adding a TODO/FIXME i prefer to also leave a short comment.
In this way, I don't have to check Trac each time I see a todo.
Just a short to help you decide if you should check the ticket or ignore it

```suggestion
        # TODO: https://twistedmatrix.com/trac/ticket/10137
        # It should raise an error when receiving malformed end of chunk markers.
```

src/twisted/web/http.py
```diff
@@ -1825,75 +1823,123 @@ class _ChunkedTransferDecoder:
...
-        if data.startswith(b"\r\n"):
+        self.length = length
+        del self._buffer[0 : eolIndex + 2]
+        self._start = 0
+        return True
+
+    def _dataReceived_CRLF(self) -> bool:
+        """
+        Await the carriage return and line feed characters that follow the
+        chunk data.
```
adiroiban: Maybe 

```suggestion
        Await the carriage return and line feed characters that are the end of chunk marker that follow the
        chunk data.
```

src/twisted/web/http.py
adiroiban: Thanks for the update. Just a very minor comment in terms of style.

I find it more helpful if the docstring of a function describes how,when,why to use it it and then inside comments explains what it does :)

For example thinks like `Then invoke C{finishCallback} and switch to state C{'FINISHED'}.` don't seem to useful for me as I can already read that from the code.

My issue was with the condition that was not obvious for a new developers just starting to read the code.

So maybe something like this:

```suggestion
        full_chunk_received = len(self._buffer) >= self.length
        if full_chunk_received:
            chunk = memoryview(self._buffer)[: self.length].tobytes()
```

src/twisted/web/test/test_http.py
adiroiban: Maybe also rename `test_short` to `test_on_shot_chunk`  or something like that.
But really, feel free to ignore.

adiroiban (commented):
Thanks.
This code looks so much better.

I have already approved it... so I am leaving it on the review queue in case some other developer wants to take a look.

src/twisted/web/http.py
twm: Agreed, but I intend to fix this in the next few days so I didn't bother.

src/twisted/web/test/test_http.py
twm: That name doesn't comply with the coding standard. ;)

twm:
Thank you again, @adiroiban!

PR #1553, opened 2021-03-18
[10138] Fix 'dict_keys' object is not subscriptable

nielsavonds:
## Scope and purpose

Fix a crash that occurs when using Dropbear to connect to a Twisted SSH server.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10138

```
Merge pull request #10138 from nielsavonds/10138-nielsavonds-dropbearfix

Author: nielsavonds
Reviewer: 
Fixes: ticket:10138

Assign list instead of dict_keys to transport supportedPublicKeys in t.c.ssh.Factory
```


adiroiban:
Thanks Niels for the PR.
I am trying to run some manual and automated tests for this PR.

Is this an issue only with dropbear client? Does it required a specific dropbear version?
Do you know why this is not an issue with OpenSSH client?

I am doing a root cause analysis trying to find answer myself.. but maybe you already have the answers of some questions  and could speed up the review :)

Thanks!

nielsavonds:
Hi and thanks for your response,

This is an issue that only occurs when connecting with the Dropbear client. We're setting up port forwards, so I'm not sure if it would work for any other connection. The crash occurs in ssh_KEXINIT, here:

```
        if ord(rest[0:1]):  # Flag first_kex_packet_follows?
            if (
                kexAlgs[0] != self.supportedKeyExchanges[0]
                or keyAlgs[0] != self.gi[0]
            ):
                self.ignoreNextPacket = True  # Guess was wrong
```

I traced this a bit and with OpenSSH the first if-test is false, so we never get to the part where the supportedKeyExchanges is indexed. However, when using OpenSSH, the supportedKeyExchanges also contains a dict_keys object, it's just never indexed.

The dropbear client command I used to test was:
```
dbclient -p 2225 -N -R 8000:localhost:8000  user@localhost
```

For the moment, I managed to work around this in our code, by doing this:
```
class NobiFactory(factory.SSHFactory):

    def buildProtocol(self, addr):
        result = super().buildProtocol(addr)
        result.supportedPublicKeys = list(result.supportedPublicKeys)
        return result
```

Let me know if you have any additional questions.

wiml:
I think it's worth adding a clarifying comment by the change to `ssh/factory.py`, because that ivar is explicitly documented as being an *ordered* list, and here we are not imposing any order. I think it's okay, but only because a bunch of things are true:

* This is only used by the server side, and my reading of [RFC4253 §7.1](https://tools.ietf.org/html/rfc4253#section-7.1) is that only the client's preference order matters for the host-key selection algorithm;
* Conch never sends a speculative (guessed) KEXINIT, so the fact that we're not being consistent about which key alg is our preferred one never matters.

An even better approach might be something along the lines of (untested):

```python
         t.supportedPublicKeys = [ 
             keyAlg 
             for keyAlg in t.supportedPublicKeys
             if keyAlg in self.privateKeys
         ]
```

This would preserve the existing order in `SSHTransportBase.supportedPublicKeys` as well as allowing someone to remove key algorithms from that list without having them re-added by `buildProtocol()`.

adiroiban:
This is still on my todo list, but for now I am busing with fixing some generic Twisted CI issues

src/twisted/conch/ssh/factory.py
```diff
@@ -56,7 +56,7 @@ def buildProtocol(self, addr):
         @return: The built transport.
         """
         t = protocol.Factory.buildProtocol(self, addr)
-        t.supportedPublicKeys = self.privateKeys.keys()
+        t.supportedPublicKeys = list(self.privateKeys.keys())
```
wiml: (See my [discussion comment](https://github.com/twisted/twisted/pull/1553#issuecomment-805348538) and [proposed change](https://github.com/nielsavonds/twisted/pull/1) )

wiml (changes requested):

src/twisted/conch/ssh/factory.py
wiml: No response from the requestor in two months, so I went ahead and added my proposed fix to this PR.

src/twisted/conch/ssh/factory.py
```diff
@@ -56,7 +56,13 @@ def buildProtocol(self, addr):
         @return: The built transport.
         """
         t = protocol.Factory.buildProtocol(self, addr)
-        t.supportedPublicKeys = self.privateKeys.keys()
+        # Advertise only the host-key algorithms that correspond to
+        # keys we have
```
adiroiban:
```suggestion
        # Advertise only the host-key algorithms that correspond to
        # the intersection of what we have with what we support.
```

I don't know what to say about this change.... but I am not against it :)

This will silently ignore a "configured" private key when python cryptography/openssl does not support it or for example if you accidentally add  a private key with a typo in the name (ssh-dsa vs ssh-dss :) 

I guess that generating a log when a non-supported key is configured can help detect some configuration / deployment errors.

I think that converting this to a list here is OK and adding the log should not be a blocker.

-------------

src/twisted/conch/newsfragments/10138.bugfix
```diff
@@ -0,0 +1 @@
+t.c.ssh.Factory assign list instead of dict_keys to transport supportedPublicKeys
```
adiroiban:
```suggestion
twisted.conch.ssh.Factory.buildProtocol now sets the  protocol.supportedPublicKeys as a list.  EXPLANATION of why the previous dict_key was bad
```

src/twisted/conch/ssh/factory.py
adiroiban: Hi @wiml thanks for the update. Let me know if you need a review on this.

adiroiban:
I have executed a small manual test:

```
$ mkdir ssh-keys
$ ckeygen -t rsa -f ssh-keys/ssh_host_rsa_key
$ ckeygen -t rsa -f ssh-keys/client_rsa
$ python docs/conch/examples/sshsimpleserver.py
```

From another terminal then do  `$ telnet localhost 5022`

I can see that `ssh-rsa` is advertised ... so all fine here.

Then I do `$ dbclient -p 5022 user@localhost` and here fails.

On Ubuntu you can install dropbear  client via  on `sudo apt-get install dropbear-bin`

-----------

The next step is to write an automated unit test and integration test for this.

----------

The SSHTransportBase.supportedPublicKeys is documented as a list, but maybe type hinting can be added as part of this PR.

src/twisted/conch/ssh/factory.py
adiroiban: Also, let me know if you need help with writing an automated tests for this.

adiroiban:
Documentation for  `first_kex_packet_follows` boolean from SSH_MSG_KEXINIT  https://datatracker.ietf.org/doc/html/rfc4253#section-7.1

When a new connection is made both client and server will send its SSH_MSG_KEXINIT without waiting for any other message from the other peer.

Is not like other protocols in which the server need to send the first message... in with HTTP in which the client needs to  send the first message.

My understanding is that dropbear (and any other ssh-client) can send its list of supported host keys and kex algorithm with SSH_MSG_KEXINIT and and informs the server that the client has already decided what algorithms to use... so the client will send a KEX message like MSG_KEXDH_INIT righ away, without waiting for the server's SSH_MSG_KEXINIT

correct me if I am wrong :)

-----------

Twisted SSH client will send its client side SSH_MSG_KEXINIT but then will wait to receive the SSH_MSG_KEXINIT from the server before proceeding with SSH_MSG_KEXINIT or other KEX method.

------------

FWIW  
I see that paramiko is also affected by this dropbear client behaviour  https://github.com/paramiko/paramiko/issues/418

And also golang https://github.com/golang/go/issues/16962

src/twisted/conch/ssh/factory.py
wiml: I think this does not significantly change the behavior if a private key is misconfigured — we would never have been able to use that key, since our connection peer wouldn't have negotiated for `ssh-dsa`.

(Converting to a list is the main motivation for this PR as I understand it. It was a list on Python2, but on Python3 it turned into a dict keys view, which doesn't support indexing. The list-ification just restores the status-quo-ante.)

wiml:
My understanding is that the client doesn't force a choice, but that it's a sort of speculative key exchange. If the client correctly guesses the negotiated cipher, then it can reduce the number of roundtrips needed to set up the connection. TLS1.3 does something similar.

We don't really do kex negotiation correctly anyway (I'm sure it's already in the trac backlog) but implementing that is for another PR, I think.

adiroiban:
> but that it's a sort of speculative key exchange.

Yes. speculative on the client side :)

> We don't really do kex negotiation correctly anyway (I'm sure it's already in the trac backlog) but implementing that is for another PR, I think.

Why?  Do you have a link or a ticket number?

--------

For this PR to land I think that we need at least one unit test to get code coverage for the affected code.
@wiml  do you plan to create the test or do you want me to create the test?

Thanks!

wiml:
> > We don't really do kex negotiation correctly anyway (I'm sure it's already in the trac backlog) but implementing that is for another PR, I think.
> 
> Why? Do you have a link or a ticket number?

Just comparing RFC4253 §7.1 against `SSHTransportBase.ssh_KEXINIT()` — I think conch's algorithm would have been correct in an RSA-only world, but when there are multiple host key algorithms the negotiation algorithm has to take that into account.

A quick look through trac doesn't turn up an entry for this particular bug; I could be wrong about it being an error.

> For this PR to land I think that we need at least one unit test to get code coverage for the affected code.
> @wiml do you plan to create the test or do you want me to create the test?

I've gotten really busy lately; I could take a stab at it in a couple of weeks, but probably not before then.

src/twisted/conch/newsfragments/10138.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh.Factory.buildProtocol now sets transport's supportedPublicKeys to a list, like it was in Python 2.x. This prevents an exception when we try to subscript that list in SSHServerTransport.ssh_KEXINIT().
```
twm, line 1: This message should mention Dropbear compatibility as that's the user-visible symptom of the issue.

twm (changes requested):
I'm not familiar with SSH key exchange, but from a process POV I think that there should be some sort of test change in this PR, so I'm marking it as change requested and removing it from the review queue. Thanks!

adiroiban:
Thanks for the cleanup Tom.

This was on my todo for a long time.

I have assigned it to myself as a reminder...but if anyone wants to continue working on this, feel free to  do so and I can help with a review.

PR #1557, opened 2021-03-18
[10143] Close file at end of test for twisted.test.test_twistd.

adiroiban:
## Scope and purpose

Some tests in twisted.test.test_twistd are leaving the local log file open af the end of the test.

This will generate a warning in the garbage collection.

There are other tests that are checking for warning and the garbage collection warnings interfere with the expected warnings and are causing random test failures.

I am not reverting the fix for ticket 10123 as the changes from there are not introducing any new error. is just that they are not doing anything.

I was fooled by the fact that with `trial -j 18` the probability of seeing this error is much lower.

I ran the test with `$ trial -u -j 2 twisted.test.test_twistd ` and have observed the failured.

Then after the fix I tried `$ trial -u -j 16 twisted.test.test_twistd` and I was no longer observing the failure...and I thought that it was fixed.

I have tested this test with  `$ trial -u -j 2 twisted.test.test_twistd ` running for 5 minutes ... in trunk this fails in 10 seconds.

```
Ran 91168 tests in 312.525s

PASSED (skips=46620, successes=44548)

```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10143

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10143

Update twisted.test.test_twistd to close the log file at the end of the test.
```

### DIFF
--- a/src/twisted/test/test_twistd.py
+++ b/src/twisted/test/test_twistd.py
@@ -1545,10 +1545,12 @@ class AppLoggerTests(TestCase):
         sut = app.AppLogger({"logfile": filename})
 
         observer = sut._getLogObserver()
-        self.addCleanup(logger.globalLogPublisher.removeObserver, observer)
+        self.addCleanup(observer._outFile.close)
 
         self.assertEqual(len(logFiles), 1)
         self.assertEqual(logFiles[0].path, os.path.abspath(filename))
+
+        logger.globalLogPublisher.removeObserver(observer)
 
     def test_stop(self):
         """
@@ -1687,7 +1689,7 @@ class UnixAppLoggerTests(TestCase):
         sut = UnixAppLogger({"logfile": filename})
 
         observer = sut._getLogObserver()
-        self.addCleanup(logger.globalLogPublisher.removeObserver, observer)
+        self.addCleanup(observer._outFile.close)
 
         self.assertEqual(len(logFiles), 1)
         self.assertEqual(logFiles[0].path, os.path.abspath(filename))
@@ -1721,7 +1723,7 @@ class UnixAppLoggerTests(TestCase):
         sut = UnixAppLogger({"logfile": filename})
 
         observer = sut._getLogObserver()
-        self.addCleanup(logger.globalLogPublisher.removeObserver, observer)
+        self.addCleanup(observer._outFile.close)
 
         self.assertEqual(self.signals, [])
 
--- a/src/twisted/test/test_twistd.py
+++ b/src/twisted/test/test_twistd.py
@@ -1550,8 +1550,6 @@ class AppLoggerTests(TestCase):
         self.assertEqual(len(logFiles), 1)
         self.assertEqual(logFiles[0].path, os.path.abspath(filename))
 
-        logger.globalLogPublisher.removeObserver(observer)
-
     def test_stop(self):
         """
         L{app.AppLogger.stop} removes the observer created in C{start}, and


src/twisted/test/test_twistd.py
```diff
@@ -1545,11 +1545,13 @@ def test_getLogObserverFile(self):
         sut = app.AppLogger({"logfile": filename})
 
         observer = sut._getLogObserver()
-        self.addCleanup(logger.globalLogPublisher.removeObserver, observer)
+        self.addCleanup(observer._outFile.close)
 
         self.assertEqual(len(logFiles), 1)
         self.assertEqual(logFiles[0].path, os.path.abspath(filename))
 
+        logger.globalLogPublisher.removeObserver(observer)
```
altendky: Is there a reason to not register both as cleanup activities?  As is, won't a failed assertion result in leaving the observer in place?  In the changes below, the observer doesn't get removed in any case anymore.  The inconsistency isn't obviously good.

(not that I know anything about twisted logging...)

exarkun: Seems correct to me.

exarkun: (I mean @altendky 's comment seems correct to me.)

altendky: Since it isn't just my naiveté, I'll formally elevate this to 'request changes'.

altendky (changes requested):

src/twisted/test/test_twistd.py
adiroiban: Good question.
Sorry for my bad code over here.

`logger.globalLogPublisher.removeObserver(observer)` does nothing in this case and should be removed.

I was reading the `AppLogger._getLogObserver()` implementation
 https://github.com/twisted/twisted/blob/trunk/src/twisted/application/app.py#L228

I saw the file is opened in `logfile.LogFile.fromFullPath(self._logfilename)` 
and then I saw that that remove observer is called in `AppLogger.stop()` and I think that this is how you can clean/close what was open by `_getLogObserver`

But I was wrong.
The stop doesn't closes the file.

I tried to read the code for the logging framework, but I am always reading it after 10PM and I can't make sense of it.

I have no idea how this logging API should be used.

This test was leaving open files for a long time, is just that it was not noticed since we haven't executed parallel tests or garbage collection checks.

You can run this on trunk
```
$ python3 -Wall -m twisted.trial --force-gc twisted.test.test_twistd.AppLoggerTests

    test_getLogObserverFile ...                                            [OK]
/home/adi/chevah/twisted/src/twisted/trial/_asyncrunner.py:141: ResourceWarning: unclosed file <_io.FileIO name='/home/adi/chevah/twisted/_trial_temp/twisted.test.test_twistd/AppLoggerTests/test_getLogObserverFile/q_avmsk1/temp' mode='rb+' closefd=True>
    test_getLogObserverStdout ...                                          [OK]

```

and then run the same test on this branch and you will not see the ResourceWarning.


I was fooled by calling the test direclty and not seeing the warning and thinking that it was fixed...

```
$ python3 -Wall -m twisted.trial --force-gc twisted.test.test_twistd.AppLoggerTests.test_getLogObserverFile
```

this is what you happend when you try to skip some TDD steps

adiroiban:
Sorry for my initial bad code. This is ready for a new review.

You can check that the resources are cleaned at the end of the test by forcing a gc and enabling warnings

```
$ python3 -Wall -m twisted.trial --force-gc twisted.test.test_twistd.AppLoggerTests
```

src/twisted/test/test_twistd.py
altendky: Ok, so `logger.globalLogPublisher.removeObserver()` is irrelevant because the observer is never getting added to the global publisher to begin with...  right?  So then yeah, I finally get that it doesn't matter.

https://github.com/twisted/twisted/blob/6fc21a15d345d32328e2df622870ee1e98868f97/src/twisted/application/app.py#L228-L237

It crossed my mind that there's some 'hazard' of the file being `stdout`, but we are explicitly setting the file in each of these cases so it would probably be overly cautious to worry about a change that would result in ...  meh...  yeah.  Looks good.  Thanks.

altendky (approved):

src/twisted/test/test_twistd.py
adiroiban: even if the observer would have been added to the global log publisher, I think that the API call for remove will just "stop forwarding logs to this observer". It will not take care of observer close and cleanup.

PR #1558, opened 2021-03-18
10141: Avoid trying to call cancel() on the timeout if it is None.

wiml:
Also, try to clear out callback variables a little earlier to avoid the possibility of them being callback()'d twice.

This fixes 10141, and might also fix [#9494](https://twistedmatrix.com/trac/ticket/9494); I can't be sure without being able to replicate it, but the backtrace in the trac ticket is suggestive.

## Contributor Checklist:

* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10141
* [X] I ran `tox -e lint`

```
Merge pull request #1558 from twisted/wiml-namesTimeout-10141

Author: wiml
Reviewer: <github_username>, <github_usernames_if_more_reviewers>
Fixes: ticket:10141
Refs: ticket:9494 ticket:3428

Avoid trying to call cancel() on the timeout if it is None, and clean up
better during abnormal situations.
```


wiml:
I think this is ready. The check failures are:

* Codecov.io says that the PR decreases overall test coverage, but I don't see how it could. If I look at the coverage report, about 2/3 of the time I get an authorization error from codecov and can't see anything; the other 1/3 of the time it reports coverage decreases in completely unrelated code, like the TLS tests. I think this is spurious.
* Codecov.io complains the diff test coverage isn't 100%. As far as I can tell, 100% of the diff *lines* are covered by tests; however, one defensive `if` statement is not fully exercised. I can't make both codecov and mypy happy here. I think the current code is more correct than if I didn't have the `if`.
* I frequently get one or two spurious failures in other parts of Twisted, *e.g.* in disttrial. They move around if I re-run the tests, and don't seem to be related to the changes in this PR. Sometimes I don't get any of these spurious failures. I assume they are an intermittent bug elsewhere in Twisted or an instability in the CI infra but I can't be sure. In any event I can't think of a way they could be connected to these AXFR fixes.


You'll also notice one TODO-skipped test; this is because these new unit tests also elicit the failure described in [trac #3428](https://twistedmatrix.com/trac/ticket/3428). I have another patch queued up to fix that, after this one is merged, but I want to get this fix merged so I can get on with my actual work.

wsanchez:
You do need to test the code past where `connector` is `None`, or remove the defensive check if it's not necessary.

What's `mypy` complaining about there?

wiml:
Ah, you're right. I was thinking of a similar class where `connector` is an ivar and therefore has to be typed `Optional`. But in this class it's threaded through the deferreds so it clearly won't be `None`.

adiroiban:
@wiml thanks for the pr.

you  can ping me if you found random/spurious/flaky tests. My goal in life is to get a solid green for all tests from trunk :)

I see some tests failing , but not all failures are due to parallel jobs.

For example this fails even without parallel
```
$ trial -u twisted.trial._dist.test.test_disttrial
```

but pass with

```
$ trial twisted.trial._dist.test.test_disttrial
```

-----------

coverage reporting should be broken, unless this is solved/merged https://github.com/twisted/twisted/pull/1574

I have no idea how you got a green build for https://github.com/twisted/twisted/pull/1558/commits/79181664e0eccaab91ec5f31898ea3e0eaa9fc6b

wiml:
@adiroiban In my most recent push c68b1db there's a spurious-looking failure in one of the pypy runs, which looks like it might simply have timed out after half an hour.

Regarding that one green commit I'm not sure either, since codecov only occasionally lets me see the results of runs. :man_shrugging: One thing I have noticed, though, is that `codecov/project` usually succeeds/fails long before the various test runs do, so perhaps its pass/fail depends on which test runs have completed when it gathers coverage data?

adiroiban:
I am working on PYPY failures at https://github.com/twisted/twisted/pull/1543

For codecov/project yes, you get a report as soon as the first report is sent  by a job... and then it is updated.
I can look at configuring that to delay the report.

wiml:
In that case I think I'll wait for the CI issues to get sorted out on trunk and then come back to this fix.

src/twisted/names/test/test_names.py
```diff
@@ -472,6 +506,65 @@ def test_zoneTransferConnectionFails(self):
...
+            EasilyStartledProtocol,
+            DomainError,
+        )
+
+    test_zoneTransferConnectionDrops.todo = "Bug #3428"  # type: ignore[attr-defined]
+
+    def test_zoneTransferTimesOut(self) -> defer.Deferred:
+        """
+        An AXFR TCP transfer which never receives a response from the
+        server should cleanly time out.
```
adiroiban: What does  "cleanly time out" mean? 

What is the target / subject / system under test here?

Would this be a better description of this test.

```
L{Resolver,lookupZone} returns a deferred which fails with L{DNSQueryTimeoutError}
when the remote DNS server is not responding before the timeout interval.
```

adiroiban:
I don't think that you should block on the random CI failures. Just make sure the code from this PR is rock solid and I am sure that it will be approved.

There will be always CI failures, as keeping a CI happy is a full time job description :)

You can try to review #1543 :)

I have triggered an update on this branch to get latest CI fixes.

I am dedicating all my current energy to improving the overall Twisted CI dev experience and I don't have the energy required to do a review :(

src/twisted/names/test/test_names.py
wiml: That sounds good, I've clarified the docstring.

wiml:
Finally a green CI check (and faster, too) — thanks for all your work, Adi!

wiml:
Ping?

src/twisted/names/test/test_names.py
```diff
@@ -469,6 +502,66 @@ def test_zoneTransferConnectionFails(self):
...
         )
 
+    def _testHelper_zoneTransferFailure(
+        self,
+        protocolClass: Type[AccumulatingProtocol],
+        expectedFailure: Type[Exception],
+        **kwargs: Any,
+    ) -> defer.Deferred:
+        factory = RecordingServerFactory.forProtocol(protocolClass)
+        nonresponsiveTCP = reactor.listenTCP(0, factory, interface="127.0.0.1")  # type: ignore[attr-defined]
```
twm, line 512: Could this be done with [`MemoryReactor`](https://twistedmatrix.com/documents/current/api/twisted.internet.testing.MemoryReactor.html) instead of the real reactor? There is a version that provides `IReactorUDP` in `twisted.names.test.test_util`.

src/twisted/names/test/test_names.py
```diff
@@ -469,6 +502,66 @@ def test_zoneTransferConnectionFails(self):
...
+            resolver.lookupZone("test-domain.com", **kwargs), expectedFailure
+        )
+
+        def _waitAnd(r, fn, *args):
+            # We need to wait a moment before calling _furtherChecks
+            # for the TCP closure to work its way through the stack
+            # and show up as accumulator.closed==True.
+            d = defer.Deferred()
+            d.addCallback(fn, *args)
+            reactor.callLater(0.01, d.callback, r)
```
twm, line 526: Blind sleeps lead to flaky tests. :grimacing: This wouldn't be necessary with a memory reactor.

src/twisted/names/test/test_names.py
```diff
@@ -469,6 +502,66 @@ def test_zoneTransferConnectionFails(self):
...
+        """
+        If the TCP connection initiated by L{Resolver.lookupZone} is lost before any
+        response is received, the deferred will errback.
+        """
+        return self._testHelper_zoneTransferFailure(
+            EasilyStartledProtocol,
+            DomainError,
+        )
+
+    test_zoneTransferConnectionDrops.todo = "Bug #3428"  # type: ignore[attr-defined]
```
twm, line 553: _Technically_ we're not supposed to have TODOs in trunk, but I'd allow it.

twm (changes requested):
The fix here looks good to me! However I'm concerned about the use of the real reactor in the test, particularly the sleep. Listening on ports like this tends to cause flakiness on the scale of Twisted's suite, so we should avoid it whenever possible, which looks quite possible here.

src/twisted/names/newsfragments/10141.bugfix
```diff
@@ -0,0 +1 @@
+t.n.client.Resolver will no longer throw an AttributeError if a zone transfer times out.
```
exarkun, line 1: Please expand the abbreviations and use "raise" instead of "throw".

src/twisted/names/test/test_names.py
```diff
@@ -469,6 +502,66 @@ def test_zoneTransferConnectionFails(self):
             resolver.lookupZone("impossible.invalid"), error.DNSLookupError
         )
 
+    def _testHelper_zoneTransferFailure(
```
exarkun, line 505: This should be `_testHelperZoneTransferFailure` because it is not dynamically dispatched.

src/twisted/names/test/test_names.py
```diff
@@ -469,6 +502,66 @@ def test_zoneTransferConnectionFails(self):
...
+        self,
+        protocolClass: Type[AccumulatingProtocol],
+        expectedFailure: Type[Exception],
+        **kwargs: Any,
```
exarkun, line 509: I think that `object` should be preferred over `Any`?

src/twisted/names/test/test_names.py
exarkun: `listenTCP(0, ...)` will fail sometimes with the real reactor which is further reason to use a test double instead.

src/twisted/names/test/test_names.py
exarkun: Why? :slightly_smiling_face: It could go on the ticket or in a new branch associated with the ticket...

PR #1562, opened 2021-03-21
10137: Reject malformed HTTP chunk boundaries

twm:
## Scope and purpose

Following up on https://github.com/twisted/twisted/pull/1551/files#r596460182, reject invalid chunk boundaries.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10137

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1870,18 +1870,23 @@ class _ChunkedTransferDecoder:
 
     def _dataReceived_CRLF(self) -> bool:
         """
-        Await the carriage return and line feed characters that are the end of chunk marker that follow the
-        chunk data.
+        Await the carriage return and line feed characters that are the end of
+        chunk marker that follow the chunk data.
 
         @returns: C{True} when the CRLF have been read, otherwise C{False}.
+
+        @raises _MalformedChunkedDataError: when anything other than CRLF are
+            received.
         """
-        # TODO: https://twistedmatrix.com/trac/ticket/10137
-        # It should raise an error when receiving malformed end of chunk markers.
-        if self._buffer.startswith(b"\r\n"):
-            self.state = "CHUNK_LENGTH"
-            del self._buffer[0:2]
-            return True
-        return False
+        if len(self._buffer) < 2:
+            return False
+
+        if not self._buffer.startswith(b"\r\n"):
+            raise _MalformedChunkedDataError("Chunk did not end with CRLF")
+
+        self.state = "CHUNK_LENGTH"
+        del self._buffer[0:2]
+        return True
 
     def _dataReceived_TRAILER(self) -> bool:
         """
@@ -1891,13 +1896,20 @@ class _ChunkedTransferDecoder:
 
         @returns: C{False}, as there is either insufficient data to continue,
             or no data remains.
+
+        @raises _MalformedChunkedDataError: when anything other than CRLF is
+            received.
         """
-        # TODO: https://twistedmatrix.com/trac/ticket/10137
-        if self._buffer.startswith(b"\r\n"):
-            data = memoryview(self._buffer)[2:].tobytes()
-            del self._buffer[:]
-            self.state = "FINISHED"
-            self.finishCallback(data)
+        if len(self._buffer) < 2:
+            return False
+
+        if not self._buffer.startswith(b"\r\n"):
+            raise _MalformedChunkedDataError("Chunk did not end with CRLF")
+
+        data = memoryview(self._buffer)[2:].tobytes()
+        del self._buffer[:]
+        self.state = "FINISHED"
+        self.finishCallback(data)
         return False
 
     def _dataReceived_BODY(self) -> bool:
--- a/src/twisted/web/newsfragments/10137.bugfix
+++ b/src/twisted/web/newsfragments/10137.bugfix
@@ -1,0 +1,2 @@
+The server-side HTTP/1.1 chunking implementation now rejects invalid chunk boundaries, preventing unbounded buffering.
+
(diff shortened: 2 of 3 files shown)


adiroiban (approved):
Looks good. Thanks!

I have tested this server on trunk and this branch
```
$ twistd -l - -n web --port tcp:8080 --path .
```

And then used this code

```python
from twisted.internet.task import react, deferLater
from twisted.internet.protocol import Factory, Protocol
from twisted.internet.endpoints import TCP4ClientEndpoint


class Greeter(Protocol):
    def dataReceived(self, data):
        print("Got:", repr(data))


async def txmain(reactor):
    point = TCP4ClientEndpoint(reactor, "localhost", 8080)
    protocol = await point.connect(Factory.forProtocol(Greeter))
    protocol.transport.setTcpNoDelay(True)
    protocol.transport.write(
        b"POST / HTTP/1.1\r\n"
        b"Transfer-Encoding: chunked\r\n"
        b"\r\n"
        )

    protocol.transport.write(
        b"3\r\n"
        b"abcEXTRA\r\n"
        b"0\r\n\r\n"
        )

    # Wait for response.
    await deferLater(reactor, 1)

react(txmain, ())
```

adiroiban:
It looks like there is no limit for the chunk length .
The only limit is HEX value.

Not sure if Twisted should  reject a non-hex value right away by default.

Right now it reads the whole length field and then does the validation... but if you receive a size like `123g`  you don't need to wait for the reset of the data to know that this is an invalid request.

For example if in the client I do

```python
    protocol.transport.write(
        b"POST / HTTP/1.1\r\n"
        b"Transfer-Encoding: chunked\r\n"
        b"\r\n"
        )

    # Write a huge chunk size... that is also invalid.
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
```

I see that the connection is not rejected right away... and there is a huge memory usage on the server.

For my production code I use a custom (forked) Request handler that by default will reject any request bigger than 100kb ... and for special URL and conditions I "unblock" the size.


Below is my interface.
It also has a fixed sized for headers...so you can't bomb the server memory just by sending a huge header.

headersReceived allows you to implement streamed consumers for the body.
maximumBodyLength allows you to prevent this type of memory issue.


```
class IChevahRequest(IRequest):
    """
    Extended request used in Chevah project.
    """
    # FIXME:1370:
    # Check if upstream interface was updated.

    maximumBodyLength = Attribute(
        """
        Set the maximum accepted body length.

        Clients sending more data are disconnected.

        Setting value `0` or None will allow the request to receive an
        unlimited body.
        """)

    def headersReceived(command, path, version):
        """
        Called by channel when all headers has been received.

        When a non-continue (non 100) response is returned, this method should
        set all the required headers for the response.
        """

```

twm:
> It looks like there is no limit for the chunk length .

There's no need to limit the chunk size, since we don't buffer the whole chunk before delivering it to the application.

> Right now it reads the whole length field and then does the validation...

Yup, that's https://twistedmatrix.com/trac/ticket/10144 which I have changes for that I will submit shortly. :)

> It also has a fixed sized for headers...so you can't bomb the server memory just by sending a huge header.

I thought we'd added a limit for that? If not, yikes.

On the general HTTP request size limit thing: I do agree that Twisted should have some mechanism here. Since we end up buffering the request in any case there should really be a policy here. I can imagine that this limit could propagate down to the chunked decoder so it could reject an oversize request as soon as the chunk length line arrives, though I'm not sure if that's worth the complexity. It seems like there might be some overlap with the 100-continue issue here? The desired request size limit will vary by path, but we lack a generalized way to communicate that back from e.g. a `t.w.r.Resource`.

Thank you for the review!

adiroiban:
yes. there is an overlap with 100-continue.
I have created the `headersReceived` for 100-continue.

The important part here is `maximumBodyLength`

I have code that starts with a default 100k ... then it reads the first data in the payload in memory and decides what to do...
this is a per resource decision. For example with a multi-part form. Some fields are kept in memory (10kb) while a file upload is redirected to S3 and can have any size.


------------

For the buffer part, since chunked decode is low-level code I think that we need to place limits on any buffer.
And as a general code security rule, maybe any code using a buffer should have a memory limited buffer.

------------


For the chunk size I have used this test... it's brutal :)
And I have observed that Twisted HTTP server uses GBs of memory  and is not disconnecting the client.
The memory is freed only after closing the web servee.

server
```
$ twistd -l - -n web --port tcp:8080 --path .
```

client

```
from twisted.internet.task import react, deferLater
from twisted.internet.protocol import Factory, Protocol
from twisted.internet.endpoints import TCP4ClientEndpoint


class Greeter(Protocol):
    def dataReceived(self, data):
        print("Got:", repr(data))


async def txmain(reactor):
    point = TCP4ClientEndpoint(reactor, "localhost", 8080)
    protocol = await point.connect(Factory.forProtocol(Greeter))
    protocol.transport.setTcpNoDelay(True)
    protocol.transport.write(
        b"POST / HTTP/1.1\r\n"
        b"Transfer-Encoding: chunked\r\n"
        b"\r\n"
        )

    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    protocol.transport.write(
        b"3g13344" * 100000000
        )
    # Wait for response.
    await deferLater(reactor, 10)

react(txmain, ())

```

PR #1550, opened 2021-03-15
[10130] Disallow negative HTTP chunk size

twm:
## Scope and purpose

`_ChunkedTransferDecoder` uses `int()` to parse the chunk size, which permits a leading `-`, resulting in a nonsensical negative chunk size. A chunk size is supposed to [consist only of HEXDIGits](https://tools.ietf.org/html/rfc7230#section-4.1), so treat a negative chunk size like any other malformed chunk size.

In addition to this fix, this PR includes a few docstring updates and typo fixes.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10130
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] I have updated the automated tests.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.

PR #1563, opened 2021-03-22
10144 Limit ChunkedTransferDecoder buffering

twm:
## Scope and purpose

Following up on #1551, #1550, and #1562, this PR imposes a length limit on the chunk size line. That is, the first line in the `chunk` production:

     chunk          = chunk-size [ chunk-ext ] CRLF
                      chunk-data CRLF

Previously, `_ChunkedTransferDecoder` would buffer an unbounded amount while searching for the terminal CRLF. The new limit, 4 KiB, doesn't meaningfully limit the size of the chunk that can be encoded, but restricts the amount of extension data that can be associated with the chunk.

The 4 KiB limit is an arbitrary one intended to maximize compatibility should extensions be present. The decoder discards extensions, as allowed by [the RFC](https://tools.ietf.org/html/rfc7230#section-4.1), so size is the only concern here.. As far as I know extensions are useless and unused in the wild, so I'm open to setting the limit lower. I put the limit in a global constant (rather than a class variable) so that it could be overridden without touching private symbols, if necessary.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10144

### DIFF
--- a/src/twisted/web/test/test_http.py
+++ b/src/twisted/web/test/test_http.py
@@ -1289,6 +1289,38 @@ class ChunkedTransferEncodingTests(unittest.TestCase):
         p.dataReceived(b"3; x-foo=bar\r\nabc\r\n")
         self.assertEqual(L, [b"abc"])
 
+    def test_oversizedChunkSizeLine(self):
+        """
+        L{_ChunkedTransferDecoder.dataReceived} raises
+        L{_MalformedChunkedDataError} when the chunk size line exceeds 4 KiB.
+        """
+        p = http._ChunkedTransferDecoder(
+            lambda b: None,  # pragma: nocov
+            lambda b: None,  # pragma: nocov
+        )
+        self.assertRaises(
+            http._MalformedChunkedDataError,
+            p.dataReceived,
+            b"3;" + b"." * 4096 + b"\r\nabc\r\n",
+        )
+
+    def test_oversizedChunkSizeLinePartial(self):
+        """
+        L{_ChunkedTransferDecoder.dataReceived} raises
+        L{_MalformedChunkedDataError} when the amount of data buffered while
+        looking for the end of the chunk size line exceeds 4 KiB so
+        that buffering does not continue without bound.
+        """
+        p = http._ChunkedTransferDecoder(
+            lambda b: None,  # pragma: nocov
+            lambda b: None,  # pragma: nocov
+        )
+        self.assertRaises(
+            http._MalformedChunkedDataError,
+            p.dataReceived,
+            b"." * 4097,
+        )
+
     def test_malformedChunkSize(self):
         """
         L{_ChunkedTransferDecoder.dataReceived} raises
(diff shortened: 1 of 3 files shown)


src/twisted/web/test/test_http.py
```diff
@@ -1289,6 +1289,34 @@ def test_extensions(self):
...
+        L{_ChunkedTransferDecoder.dataReceived} raises
+        L{_MalformedChunkedDataError} when the amount of data buffered while
+        looking for the end of the chunk size line exceeds 4 KiB so
+        that buffering does not continue without bound.
+        """
+        p = http._ChunkedTransferDecoder(None, None)
+        self.assertRaises(
+            http._MalformedChunkedDataError,
+            p.dataReceived,
+            b"." * 4097,
```
adiroiban: Maybe write this test using  `http.maxChunkSizeLineLength + 1`

adiroiban (approved):
It looks good. Thanks for the fix. It can be merge.


Does this qualify for a security bugfix.

It's not data loss, but it's an easy DOS attack.

I think that any unbound buffer should be considered a security issue.

I think that  we should also have a `.security` news fragment to highlight the very sensitive changes.

I would go with a default like 32+2 characters  ... but maybe is  safe to have 1024.

If my math is right, with 32 hex digits you should be able to represent a number larger than yottabytes

From RFC.
```
The chunked encoding is specific to each connection and is likely to
   be removed or recoded by each recipient (including intermediaries)
   before any higher-level application would have a chance to inspect
   the extensions.  Hence, use of chunk extensions is generally limited
   to specialized HTTP services such as "long polling" (where client and
   server can have shared expectations regarding the use of chunk
   extensions) or for padding within an end-to-end secured connection.
```

And if someone wants to use extension, they can explicitly increase the limit.

I guess that in these days many are using Twisted web beside a HTTP App load balancer and these security issue is handled by the balancer.

I don't have time know to set up a test, but it would be interesting to see what size is accepted by AWS / Azure / Google cloud load balancer before dropping the connection.

Thanks again for the PR .

twm:
> I would go with a default like 32+2 characters ... but maybe is safe to have 1024.

I reduced it to 1024. I'm not inclined to seek the absolute minimal value as that would be unnecessarily incompatible if anyone actually _is_ using chunk extensions.

> From RFC.
> 
>> The chunked encoding is specific to each connection and is likely to
   be removed or recoded by each recipient (including intermediaries)
   before any higher-level application would have a chance to inspect
   the extensions.  Hence, use of chunk extensions is generally limited
   to specialized HTTP services such as "long polling" (where client and
   server can have shared expectations regarding the use of chunk
   extensions) or for padding within an end-to-end secured connection.
>
> And if someone wants to use extension, they can explicitly increase the limit.

The use case described in the RFC is presently impossible with t.w.http because we don't communicate chunk extensions to the application in any way. So I think compatibility with some client that is sending useless extensions for reasons unknown is the only concern here.

Thank you for the review! I'll merge this as the build greens up.

PR #1570, opened 2021-03-27
10155 Add type hints to twisted.internet.task.

wsanchez:
Add type hints for `twisted.internet.task`

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10155

```

(as long as you wish)
```


src/twisted/internet/base.py
```diff
@@ -237,13 +238,14 @@ def __repr__(self) -> str:
...
-                    func = self.func.im_class.__name__ + "." + func  # type: ignore[attr-defined]
-            else:
+            func = getattr(self.func, "__qualname__", None)
+            if func is None:
+                func = getattr(self.func, "__name__", None)
+                if func is not None:
+                    imClass = getattr(self.func, "im_class", None)
+                    if imClass is not None:
+                        func = f"{imClass}.{func}"
+            if func is None:
```
wsanchez: Is this code obsolete?  `__qualname__` was added in Python 3.3.

src/twisted/internet/defer.py
```diff
@@ -631,7 +631,7 @@ def chainDeferred(self, d: "Deferred[_DeferredResultT]") -> "Deferred[None]":
         d._chainedTo = self
         return self.addCallbacks(d.callback, d.errback)
 
-    def callback(self, result: _DeferredResultT) -> None:
+    def callback(self, result: Union[_DeferredResultT, Failure]) -> None:
```
adiroiban, line 638: I don't understand why we have _DeferredResultT in the first place.

Shouldn't this be `object` as you can pass anything to a callback?

```suggestion
    def callback(self, result: object) -> None:
```

and we have  `def _startRunCallbacks(self, result: object) -> None:`

----------------

Maybe also add a docstring update that `result` can also be a Failure, and in this case this has the same effect as calling `errback`.

src/twisted/internet/task.py
```diff
@@ -40,39 +55,37 @@ class LoopingCall:
...
 
-    def __init__(self, f, *a, **kw):
+    def __init__(self, f: Callable[..., object], *a: object, **kw: object) -> None:
         self.f = f
         self.a = a
         self.kw = kw
         from twisted.internet import reactor
 
-        self.clock = reactor
+        self.clock: IReactorTime = cast(IReactorTime, reactor)
```
adiroiban: Just asking. Wouldn't this do the job... as the explicit casting will tell mypy that the result is IReactorType ?

```suggestion
        self.clock = cast(IReactorTime, reactor)
```

src/twisted/internet/task.py
```diff
@@ -20,11 +31,15 @@
...
 
-from twisted.internet import base, defer
-from twisted.internet.interfaces import IReactorTime
+from twisted.internet.base import DelayedCall
+from twisted.internet.defer import Deferred, ensureDeferred, maybeDeferred
+from twisted.internet.interfaces import IDelayedCall, IReactorCore, IReactorTime
 from twisted.internet.error import ReactorNotRunning
 
 
+_T = TypeVar("_T")
```
adiroiban, line 41: I have no idea why you need this. 
Is there a tutorial or a guideline for when to use this?

Maybe it's commons sense .

I read the docs and `A = TypeVar('A', str, bytes)` make sense... but this usage, don't :)

https://docs.python.org/3/library/typing.html#typing.TypeVar

src/twisted/internet/task.py
```diff
@@ -130,7 +139,9 @@ def counter():
             lastTime = self._realLastTime
             if lastTime is None:
                 lastTime = self.starttime
+                assert lastTime is not None
```
adiroiban: why these asserts ? I guess they are there to instruct mypy that lastTime can't be None.. so that you don't need Optional.

But is this a workaround/hack or the final code pattern used by mypy? 

I search https://docs.python.org/3/library/typing.html for info about this pattern... nothing found.

src/twisted/internet/task.py
```diff
@@ -369,61 +381,56 @@ class CooperativeTask:
...
 
-        @return: a L{defer.Deferred} that fires with the C{iterator} that this
+        @return: a L{Deferred} that fires with the C{iterator} that this
             L{CooperativeTask} was created with when the iterator has been
             exhausted (i.e. its C{next} method has raised C{StopIteration}), or
             fails with the exception raised by C{next} if it raises some other
             exception.
 
-        @rtype: L{defer.Deferred}
+        @rtype: L{Deferred}
```
adiroiban, line 428: do we need this?

adiroiban (commented):
Looks good. Thanks!


I don't understand why TypeVar is used , so I only left a comment.

But if this is not reviewed by someone else in 2 weeks, I think that this can be merged.



------------------

In the docstring for `errback` I see

```
Passing a string as `fail' is deprecated, and will be punished with
a warning message.
```

I think that t his was already deprecated and maybe can be removed in this PR as a drive by.

src/twisted/internet/base.py
```diff
@@ -907,7 +895,7 @@ def _moveCallLaterSooner(self, delayedCall: DelayedCall) -> None:
     def _cancelCallLater(self, delayedCall: DelayedCall) -> None:
         self._cancellations += 1
 
-    def getDelayedCalls(self) -> List[IDelayedCall]:
+    def getDelayedCalls(self) -> Iterable[IDelayedCall]:
```
wiml: If this is returning an iterable, I'd want the docstring to tell me if I need to be concerned about the effects of creating/removing deferreds before I'm done iterating. Is this a snapshot generated at call-time or is it lazily generated on iteration?

Perhaps this should be typed as `Sequence` ? That would address the concern in PR #995 and would effectively indicate that it's a static snapshot (which it is) but not constrain us to returning a `list` if we don't want to. (And it allows the caller to easily make assertions about length, which I suspect is a useful thing for callers of this function to be able to do in unit tests.)

src/twisted/internet/defer.py
wiml: Type variables are bound to a specific type at some point during type-checking. The scope of that binding is poorly documented (i.e. it's obvious to people who write typecheckers). But in this case it's a type parameter of the Deferred object's type. In the same way that calling `pop()` on a `List[sometype]` returns specifically `sometype` and not `object`, this annotation says you're supposed to call `callback()` with the specific type that it's expecting.

The type annotation on Deferred isn't quite correct anyway (see my comment in [trac 9816](https://twistedmatrix.com/trac/ticket/9816#comment:1) but it may be the best we can do within mypy's limitations.

I do think that if we want to explicitly support the behavior that calling `callback` with a `Failure` is equivalent to calling `errback` then it should be documented.

src/twisted/internet/interfaces.py
```diff
@@ -1194,11 +1194,11 @@ def callLater(
                  C{reset()} methods.
         """
 
-    def getDelayedCalls() -> List["IDelayedCall"]:
+    def getDelayedCalls() -> Iterable["IDelayedCall"]:
         """
         Retrieve all currently scheduled delayed calls.
 
-        @return: A list of L{IDelayedCall} providers representing all
+        @return: An iterable of L{IDelayedCall} providers representing all
```
wiml: See PR #995 and my comment on t.i.base.py; this is a potentially breaking semantic change, probably should have its own ticket

src/twisted/internet/task.py
wiml: Yes, I think this would be better written as an assert about the state of the `LoopingCall` object, before `starttime` is assigned to `lastTime`:

```
                assert self.starttime is not None, "LoopingCall called before it was started?"
```

Mypy should be able to propagate its understanding of the non-None-ness into `lastTime`, and this would tell the reader (or eventual debugger) what actual invariant has been violated if this assert were to fail.

src/twisted/internet/task.py
```diff
@@ -40,39 +55,37 @@ class LoopingCall:
...
-    interval = None
+    _deferred: Optional[Deferred["LoopingCall"]] = None
+    interval: Optional[float] = None
     _runAtStart = False
-    starttime = None
+    starttime: Optional[float] = None
+    _realLastTime: Optional[float] = None
 
-    def __init__(self, f, *a, **kw):
+    def __init__(self, f: Callable[..., object], *a: object, **kw: object) -> None:
```
wiml, line 80: I wonder if it would be better to type `f` as returning `Union[object, Deferred[object]]` ? It wouldn't restrict the set of types at all, but it would clarify the expected return values.

I'd argue that it would make more sense to type it as `Union[None, Deferred[None]]` except that would probably end up requiring nuisance casts in calling code.

src/twisted/internet/task.py
```diff
@@ -110,17 +123,13 @@ def withCount(cls, countCallable):
...
         @return: An instance of L{LoopingCall} with call counting enabled,
             which provides the count as the first positional argument.
 
-        @rtype: L{LoopingCall}
-
         @since: 9.0
         """
 
-        def counter():
+        def counter() -> object:
```
wiml, line 133: A docstring here would be nice.

src/twisted/internet/task.py
```diff
@@ -130,7 +139,9 @@ def counter():
             lastTime = self._realLastTime
             if lastTime is None:
                 lastTime = self.starttime
+                assert lastTime is not None
                 if self._runAtStart:
+                    assert self.interval is not None
```
wiml: Same consideration as the assertion about `self.starttime`

src/twisted/internet/task.py
```diff
@@ -568,20 +579,24 @@ def __init__(
...
-        self._delayedCall = None
+        self._delayedCall: Optional[IDelayedCall] = None
         self._stopped = False
         self._started = started
 
-    def coiterate(self, iterator, doneDeferred=None):
+    def coiterate(
+        self,
+        iterator: Iterator[_TaskResultT],
+        doneDeferred: Optional[Deferred[Iterator[_TaskResultT]]] = None,
```
wiml, line 598: This (and the return value) should be typed to indicate that the deferred fires with the same-type object as was passed to `iterator`, rather than just an arbitrary iterator

src/twisted/internet/task.py
```diff
@@ -628,7 +643,7 @@ def _removeTask(self, task):
             self._delayedCall.cancel()
             self._delayedCall = None
 
-    def _tasksWhileNotStopped(self):
+    def _tasksWhileNotStopped(self) -> Iterable[CooperativeTask]:
```
wiml, line 651:
```suggestion
    def _tasksWhileNotStopped(self) -> Iterator[CooperativeTask]:
```

src/twisted/internet/task.py
```diff
@@ -696,7 +711,7 @@ def running(self):
 _theCooperator = Cooperator()
 
 
-def coiterate(iterator):
+def coiterate(iterator: Iterator[_T]) -> Deferred[Iterator[_T]]:
```
wiml, line 719: Same comment as for `Cooperator.coiterate()`

src/twisted/internet/task.py
```diff
@@ -797,25 +812,27 @@ def advance(self, amount):
...
         """
         for amount in timings:
             self.advance(amount)
 
 
-def deferLater(clock, delay, callable=None, *args, **kw):
+def deferLater(
+    clock: IReactorTime,
+    delay: float,
+    callable: Optional[Callable[..., _T]] = None,
```
wiml, line 831: If this can take `None`, it should probably be documented as such

src/twisted/internet/task.py
wiml: A good example of the difference, which bit me a few times before I understood it, is the [AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr) type var, which is defined like your type `A`. Consider these two functions:

```python

def foo(arg: AnyStr) -> AnyStr:
  ...

def bar(arg: Union[str, bytes]) -> Union[str, bytes]:
  ...
```

Both functions can take either str or bytes and can return either str or bytes. However, if you call `foo` with a str, it is constrained to return a str, not bytes — for the scope of that call, `AnyStr` is bound specifically to str. On the other hand, it's perfectly valid for `bar` to return bytes if you pass it a str, or even to return a random choice between strs and byteses at runtime. `bar`'s arg type and return type aren't connected to each other (at least, not connected by the type annotations).

Another, more subtle, difference, is that `AnyStr` can't be bound to `Union[str, bytes]`! So if a *caller* has some value that could be str or bytes but it doesn't know which (i.e., the caller's variable has type `Union[str, bytes]`) it can't call `foo()`. I ran into that situation [here](https://github.com/twisted/twisted/pull/1576/commits/8f483dc7e6ef0340ec86a9ee950185a537c54671).

wiml (dismissed):

src/twisted/internet/task.py
adiroiban: `AnyStr` usage is clear ... but what is the purpose of `_T = TypeVar("_T")` ?

Does it mean that with this `foo` accepts any type, but arg1 and arg2 and return must be of same type   ?

```python
def foo(arg1: _T, arg2: _T) -> _T:
    ...
```

Maybe rename it to `_SameType = TypeVar('_SameType')

src/twisted/internet/base.py
twm: Agreed, this should be `Sequence` &mdash; I have asserted on the length of this value.

src/twisted/internet/task.py
wiml: > Does it mean that with this `foo` accepts any type, but arg1 and arg2 and return must be of same type ?

Yes exactly.

I think it's reasonable keeping the name since `T` and `U` are pretty much universally used for generic type variables like these. A name like `SameType` would be confusing because it's redundant; *all* type variables are the same type as themselves, so I'd go looking for what *else* `SameType` was the same type as.

If the variable is used over many lines or is constrained in some interesting way, etc., then I think it's worth giving it a descriptive name. I've been using `_TSomething` but there might be a better convention out there.

src/twisted/internet/base.py
wsanchez: OK changing to `Sequence`

src/twisted/internet/defer.py
wsanchez: I did think it was odd to pass a `Failure` to `callback`, but that exists in Twisted.  Let me look again as to what code does that, but I don't think we can change the behavior of `callback` without breaking compatibility.

src/twisted/internet/task.py
wsanchez: yup

src/twisted/internet/task.py
wsanchez: `_T` is used all other the place in type hints, so I'd rather not depart from that convention, which is also in use Twisted elsewhere.

If Twisted want to adopt a different convention, that's fine, but this isn't the PR for that.  Perhaps a discussion, then edit to contributor docs, then fix the existing use cases.

src/twisted/internet/task.py
wsanchez: Yes, since we aren't importing `defer`. Also, I think it's more readable that way.

src/twisted/internet/task.py
wsanchez: I think the client has to rely on the doctoring for anything useful here.

Since it doesn't cause the type checker to do anything new, I don't think changing the return type to include `Deferred[object]` is useful.

src/twisted/internet/task.py
wsanchez: It's a nested callable… perhaps in a different PR.

src/twisted/internet/task.py
wsanchez: @adiroiban The issue here is that `self.starttime` is optional, and while the code is written such that it shouldn't be `None` at this point, Mypy isn't in the general bug detection business, so it wants you to make that an explicit assertion, which I think is a good thing.

@wiml I'll make that change for clarity.

src/twisted/internet/task.py
wsanchez: Can you suggest something, cuz I see an iterator of `_TaskResultT`… that's not specific enough?

src/twisted/internet/task.py
wsanchez: Iterator is broader than the single use case for this private callable…

src/twisted/internet/task.py
wsanchez: Same question…

src/twisted/internet/task.py
wsanchez: done

src/twisted/internet/defer.py
wsanchez: Docstring updated.

We don't really have a choice about supporting that behavior; changing it would be a compatibility issue, and probably not a small one.

wsanchez:
I think I've addressed all of the comment except for @wiml's suggestion on `coiterate`, since I'm not sure what to change there.

src/twisted/internet/task.py
wsanchez: @wiml poke poke

src/twisted/internet/task.py
wsanchez: Perhaps the question here is whether `Iterator[_TaskResultT]` used throughout `CooperativeTask` should instead be a `_TaskIteratorT` which is defined as `TypeVar("_TaskIterableT", bound=Iterator[_TaskResultT])`.

I'm not sure that the type of the `Iterator` (of `_TaskResultT`) being preserved here is important (that it: part of the API).  And I'm not sure if `_TaskIteratorT` as I suggested above will accomplish that…  @graingert: thoughts?

The documentation for `coiterate` just says that you get back a `Deferred` that fires on completion, not what that it's return is… so perhaps we should be actually typing return value as `Deferred[None]`?

adiroiban (approved):
Thanks for the update.

I think that this can be merged and there is not much use keep it pending without a review.

if someone is not happy with it, we can always have a followup PR.

Thanks again!

wsanchez:
Thanks

PR #1574, opened 2021-03-28
[10159] Fix codecov.io reporting. Move codecov.io / coveralls reporting outside of tox environments.

adiroiban:
## Scope and purpose

Reporting coverage to codecov.io and coveralls should only be needed for CI.

For CI we already have python virtual env created.

That venv can be reused for codecov/coveralls operations so that we don't need to add tox extra complexity for those calls.

It should also make the tox.ini cleaner and move the coverage publish code closer to where it's used.

Coverage reporting also broken with codecov reports for windows linux and mac not merged.
Coveralls reports are not normalized.

## Changes

There is a problem with codecov.io reporting as the coverage is reported for the PR auto-merge commit.
I saw this error for GitHub Coverage and Azure Coverage.

The error is
```
 Issue detecting commit SHA. Please run actions/checkout with fetch-depth > 1 or set to 0
```

`fetch-depth: 2` was enabled for codecov bash uploaded.

---------

Generating XML+ codecov.io publish was taking 40 seconds with the most time taken by XML generation.
Only coveralls publish was also taking 40 seconds.

If coveralls is not used by Twisted devs we can stop publishing it and save 40 seconds with each build.

-------

With coverage publishing handling moved outside of tox.ini there is no speed improvement.
I still prefer the coverage publishing outside of tox as I find it easier to understand how things works.

----------

The coverage config was updated to normalize the path and ignore the trial temp file.

The CI steps were updated to show the status of the local coverage... to help debugging.

----------

Any errors generated during coverage reporting are ignored.
The idea is to reduce random failures.
Once we have the other CI jobs with a solid green we can look at requiring strict rules.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10159
* [NO] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10159

Fix codecov reporting for a PR. Move coverage CI reporting outside of tox.
```

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,1 +1,0 @@
-repo_token: "JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -119,7 +119,7 @@ jobs:
       run: |
         python -m coveralls
       env:
-        GITHUB_TOKEN: ${{ github.token }}
+        COVERALLS_REPO_TOKEN: 'JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'
 
 
   pypi-publish:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -109,6 +109,7 @@ jobs:
 
     - uses: codecov/codecov-action@v1
       if: always()
+      continue-on-error: true
       with:
         files: coverage.xml
         name: lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}
@@ -118,6 +119,7 @@ jobs:
     - name: Publish to Coveralls
       # We want to publish coverage even on failure.
       if: always()
+      continue-on-error: true
       run: |
         python -m coveralls
       env:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -112,15 +112,14 @@ jobs:
       run: |
         python -m coverage xml -o coverage.xml -i
         python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
-      env:
-        COVERALLS_REPO_TOKEN: 'JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'
 
     - name: Publish coveralls
       # We want to publish coverage even on failure.
       if: contains(matrix['tox-env'], 'withcov') || failure()
       run: |
         python -m coveralls
-
+      env:
+        COVERALLS_REPO_TOKEN: 'JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'
 
 
   pypi-publish:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -106,16 +106,16 @@ jobs:
         python --version
         tox -q
 
-    - name: Generate xMl and publish codecov.io
+    - name: Generate XML and publish to Codecov
       # We want to publish coverage even on failure.
-      if: contains(matrix['tox-env'], 'withcov') || failure()
+      if: always()
       run: |
         python -m coverage xml -o coverage.xml -i
         python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
 
-    - name: Publish coveralls
+    - name: Publish to Coveralls
       # We want to publish coverage even on failure.
-      if: contains(matrix['tox-env'], 'withcov') || failure()
+      if: always()
       run: |
         python -m coveralls
       env:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -111,7 +111,9 @@ jobs:
       if: always()
       run: |
         python -m coverage xml -o coverage.xml -i
-        python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
+        # Bash is the recommended uploader.
+        # See: https://github.com/codecov/example-python#faq
+        bash <(curl -s https://codecov.io/bash) -X search -X gcov -f coverage.xml -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
 
     - name: Publish to Coveralls
       # We want to publish coverage even on failure.
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -98,7 +98,7 @@ jobs:
     - uses: twisted/python-info-action@v1
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip tox codecov coveralls
+        python -m pip install --upgrade pip tox coveralls
         tox --notest
 
     - name: Test
@@ -110,7 +110,6 @@ jobs:
       # We want to publish coverage even on failure.
       if: always()
       run: |
-        python -m coverage xml -o coverage.xml -i
         # Bash is the recommended uploader.
         # See: https://github.com/codecov/example-python#faq
         bash <(curl -s https://codecov.io/bash) -X search -X gcov -f coverage.xml -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,13 +47,14 @@ jobs:
           # end to end functional test usage for non-distributed trial runs.
           - python-version: 3.6.7
             tox-env: nodeps-withcov-posix
-            trial-args: ''
+            trial-args: '-j 4'
           # `noipv6` is created to make sure all is OK on an OS which doesn't
           # have IPv6 available.
           # Any supported Python version is OK for this job.
           - python-version: 3.6
             tox-env: alldeps-withcov-posix
             noipv6: -noipv6
+            trial-args: '-j 4'
           # On PYPY concurrent test jobs result in random failures so for now
           # run non-distributed tests.
           - python-version: pypy-3.6
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -107,8 +107,7 @@ jobs:
         python --version
         tox -q
 
-    - name: Prepare coverate
-      # We want to publish coverage even on failure.
+    - name: Prepare coverage
       if: always()
       continue-on-error: true
       run: |
@@ -129,11 +128,10 @@ jobs:
         functionalities: gcov,search
 
     - name: Publish to Coveralls
-      # We want to publish coverage even on failure.
       if: always()
       continue-on-error: true
       run: |
-        python -m coveralls
+        python -m coveralls -v
       env:
         COVERALLS_REPO_TOKEN: 'JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -106,19 +106,21 @@ jobs:
         python --version
         tox -q
 
-    - name: Publish coverage
+    - name: Generate xMl and publish codecov.io
       # We want to publish coverage even on failure.
       if: contains(matrix['tox-env'], 'withcov') || failure()
       run: |
         python -m coverage xml -o coverage.xml -i
-        python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml &&
-        PID1=$!
-        python -m coveralls &&
-        PID2=$!
-        wait $PID1
-        wait $PID2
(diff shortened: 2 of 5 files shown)


adiroiban:
The builds are now reported for both GHA and Azure https://codecov.io/gh/twisted/twisted/commit/4794ef2b15244904655b4c9dd20eae352fe88ae4/build

image

.coveralls.yml
```diff
@@ -1 +0,0 @@
-repo_token: "JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX"
```
adiroiban, line 1: This should be private.
For now, I have moved it inside each CI config. In this way we don't need to install pyyaml just to read this line.

.github/workflows/test.yaml
```diff
@@ -99,18 +98,28 @@ jobs:
...
+        python -m pip install --upgrade pip tox codecov coveralls
         tox --notest
 
     - name: Test
       run: |
         python --version
         tox -q
 
-    - name: Publish coverage
+    - name: Generate xMl and publish codecov.io
```
altendky:
```suggestion
    - name: Generate XML and publish to Codecov
```

.github/workflows/test.yaml
```diff
@@ -99,18 +98,28 @@ jobs:
...
       # We want to publish coverage even on failure.
       if: contains(matrix['tox-env'], 'withcov') || failure()
-      run: tox -e coverage-prepare,codecov-push,coveralls-push
+      run: |
+        python -m coverage xml -o coverage.xml -i
+        python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
+
+    - name: Publish coveralls
+      # We want to publish coverage even on failure.
+      if: contains(matrix['tox-env'], 'withcov') || failure()
```
altendky: Won't this publish on failure even if `withcov` isn't in the tox env?  I do see this already existed a few lines up.

.github/workflows/test.yaml
```diff
@@ -99,18 +98,28 @@ jobs:
...
         tox -q
 
-    - name: Publish coverage
+    - name: Generate xMl and publish codecov.io
       # We want to publish coverage even on failure.
       if: contains(matrix['tox-env'], 'withcov') || failure()
-      run: tox -e coverage-prepare,codecov-push,coveralls-push
+      run: |
+        python -m coverage xml -o coverage.xml -i
+        python -m codecov -n 'lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}' -X search -X gcov -f coverage.xml
```
altendky: My understanding is that Codecov prefers use of the bash uploader.

azure-pipelines/run_test_steps.yml
```diff
@@ -28,7 +28,9 @@ parameters:
 steps:
 - checkout: self
   clean: true
-  fetchDepth: 1
+  # A minimum dept of 2 is required, as for PR head is the auto-merge commit
+  # and we need the last commit from PR to report the coverage.
```
altendky: This is very possibly a topic for a separate PR, but what about just skipping the automatic merge commit and [checking out the actual branch](https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit) instead?  It's not like the automatic merge commit actually keeps the CI report up to date with the target branch.  Relatedly, having dropped the requirement that branches are up to date seems a bit hazardous.

```suggestion
  # A minimum depth of 2 is required, as the default is the auto-merge commit
  # and we need the last commit from the PR to report the coverage.
```

What code actually looks back up to the parent to use this?  Maybe this relates to the long-standing complaints about sometimes getting wrong delta coverage reports?

azure-pipelines/run_test_steps.yml
```diff
@@ -45,7 +47,7 @@ steps:
    python -c "import os; [ print(e,v) for (e,v) in os.environ.items() ]"
   displayName: 'Get Python Information'
 
-- script: 'python -m pip install -U pip setuptools tox'
+- script: 'python -m pip install -U pip setuptools tox coveralls pyyaml'
```
altendky: Wasn't the dropping of `.coveralls.yml` so that we didn't need `pyyaml`?  https://github.com/twisted/twisted/pull/1574/files#r602889624  In case that is so...

```suggestion
- script: 'python -m pip install -U pip setuptools tox coveralls'
```

azure-pipelines/run_test_steps.yml
```diff
@@ -62,7 +64,18 @@ steps:
...
       TWISTED_REACTOR: ${{ parameters.windowsReactor }}
 
-- bash:  |
-    python -m tox -e coverage-prepare,coveralls-push
-    bash <(curl -s https://codecov.io/bash) -n "${{ parameters.platform }}-${{ parameters.pythonVersion }}-alldeps-withcov"
-  displayName: 'Report coverage'
+- bash: |
+    python -m coverage xml -o coverage.xml -i
+    bash <(curl -s https://codecov.io/bash) -X search -X gcov -f coverage.xml -v -n "${{ parameters.platform }}-${{ parameters.pythonVersion }}-alldeps-withcov"
+  displayName: 'Generate XMl and report codecov.io'
```
altendky: Capitalization and may as well match above I guess...

```suggestion
  displayName: 'Generate XML and publish to Codecov'
```

altendky (commented):
If it doesn't speed anything up then I'm not clear how it is better to repeat the (little bit of) code in two CI configurations than to just have it in `tox.ini`.  Though sure, it doesn't presently seem to be actually the same between the two since GHA uses the Python uploader and AP uses the bash uploader.  `:[`

There are some minor changes requested, but at this point mostly I'm interested in some discussion.

azure-pipelines/run_test_steps.yml
adiroiban: Thanks for the typo fixes.

Note that this is an issue with Azure Pipelines and not GitHub actions ... so we need different docs  and I don't think there is this option

https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#checkout


The codecov bash script in Azure needs this.

Running the test on branch HEAD vs  merge-commit should be debated outside of this PR.

adiroiban:
Thanks @altendky for the review

> If it doesn't speed anything up then I'm not clear how it is better to repeat the (little bit of) code in two CI configurations than to just have it in `tox.ini`. Though sure, it doesn't presently seem to be actually the same between the two since GHA uses the Python uploader and AP uses the bash uploader. ``

On Azure we need special handling for coveralls... and for azure we also need 2 separate tox env -- to fake CircleCI only for Covealls and not for Codecov.

I tried to move the code to tox.ini but I found the result e bigger mess.
Even if the commands are the same the comand line arguments and environment variables are defined in different ways


The CI configuration already complicated enough with job matrix on GHA and templates in Azure, and the env VARS that need to be created inside the CI YAML config and then passed to TOX.
For me, keeping everything inside the CI YAML config is better.

You can give it a try if you want.

```
[testenv:ci-coverage-gha]
description = Publish the coverage from an GitHub Actions CI job.
skip_install = True
deps =
    codecov
    coveralls
setenv =
    COVERALLS_REPO_TOKEN='JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'

commands =
    # Convert from internal coverage format to universal Cobertura XML.
    python -m coverage xml -o coverage.xml -i
    python -m codecov -X search -X gcov -f coverage.xml $CODECOV_ARGS
    python -m coveralls


[testenv:ci-coverage-azure-covdecov]
description = Publish to codecov from an Azure CI job.
skip_install = True
deps =
    codecov

commands =
    # Convert from internal coverage format to universal Cobertura XML.
    python -m coverage xml -o coverage.xml -i
    python -m codecov -X search -X gcov -f coverage.xml $CODECOV_ARGS


[testenv:ci-coverage-azure-coveralls]
description = Publish to Coveralls from an Azure CI job.
skip_install = True
deps =
    coveralls

setenv =
    COVERALLS_REPO_TOKEN='JFDTIRUVOQ8jCM3zcajrZALlpKXyiXGAX'

commands =
    # Coveralls has no support for Azure so we pretend to be CircleCI.
    # See https://github.com/lemurheavy/coveralls-public/issues/1272
    export CIRCLECI=1
    CIRCLE_BUILD_NUM=$BUILD_BUILDNUMER
    export CIRCLE_BRANCH=$BUILD_SOURCEBRANCH
    python -m coveralls

```

and bash version from tox

```
[testenv:ci-coverage-azure-codecov]
description = Publish to Codecov via bash from an Azure CI job.
skip_install = True

allowlist_externals =
    bash
    curl

commands =
    # Convert from internal coverage format to universal Cobertura XML.
    python -m coverage xml -o coverage.xml -i

    curl -s https://codecov.io/bash -o codecov.sh
    bash codecov.sh  -X search -X gcov -f coverage.xml $CODECOV_ARGS
```

adiroiban:
Codecov project fails due to

* https://app.codecov.io/gh/twisted/twisted/compare/1574/changes#D1L6188  - looks like we don't have tests for `imap4_utf_7` codec and a previous report didn't report the branch coverage for it
* https://app.codecov.io/gh/twisted/twisted/compare/1574/changes#D2L405 - looks like a timeout was not longer triggered. i suspect the timeout was reported from a previous test failure

.github/workflows/test.yaml
adiroiban: true. Thanks!

altendky: My point was that this should only publish if `withcov` was specified.  I don't know how to do this in a simple way, and I haven't thought through the not-simple way.  So, just always doing it and ignoring any potential errors when there's no coverage data (which apparently there either aren't or they have already been dealt with, what with the passing CI (or we don't exercise this case)) may be the best that's worth doing anyways.  And that's what we have now.

I guess the tl;dr is that I think `if: always()` is 'wrong' but maybe GHA isn't friendly enough for it to be worth making it right.

azure-pipelines/run_test_steps.yml
altendky: Sorry, too much GHA on my mind.  Thanks for the correction.  I would expect that AP has some variable that could be used to specify the branch instead of the merge commit.  But agreed, no need for a debate here.

.coveralls.yml
altendky: If it were 'private' it wouldn't be available to PRs from forks so we wouldn't get coverage on them.  (for some definition of private)

altendky (approved):
Thanks for the detailed response.  Everything looks to be addressed except maybe one bit.  The use of the the Python vs. bash codecov uploader on GHA.  If you did consider that and decline to change, alrighty.  `:]`  No need to hold up on that.

I'll trust your call on the coverage issues.

.github/workflows/test.yaml
adiroiban: Thanks for the headup and the review.

I though that I have moved to bash, and this is why I haven't left another comment. My bad.

I will update this to use the bash uploader

from https://github.com/codecov/example-python

```
Q: What's the difference between the codecov-bash and codecov-python uploader?
A: As far as python is concerned, nothing. You may choose to use either uploader. Codecov recommends using the bash uploader when possible as it supports more unique repository setups. 
```

adiroiban:
Thanks @altendky for the review.


It looks like the merge commit is still pushed... but after some time, codecov.io removes it and merges it with the main PR commit.

image


So I am not sure if depth=2 is required.

But it looks like it takes a lot of time for codecov.io to send to final report... and I think that this is due to pypy reports that are sent that late.

It's funny, as now the GitHub Actions reports are not visible.


so... still WIP. I will ping you over IRC when things are ready for another check

Thanks!

.github/workflows/test.yaml
```diff
@@ -48,13 +47,14 @@ jobs:
           # end to end functional test usage for non-distributed trial runs.
           - python-version: 3.6.7
             tox-env: nodeps-withcov-posix
-            trial-args: ''
+            trial-args: '-j 4'
```
adiroiban, line 50: this is a drive-by fix, as parallel coverage should be enabled here.

It was accidentally disabled thinking that if left empty it will use the default.

.github/workflows/test.yaml
```diff
@@ -48,13 +47,14 @@ jobs:
...
           # `noipv6` is created to make sure all is OK on an OS which doesn't
           # have IPv6 available.
           # Any supported Python version is OK for this job.
           - python-version: 3.6
             tox-env: alldeps-withcov-posix
             noipv6: -noipv6
+            trial-args: '-j 4'
```
adiroiban, line 57: this is a drive-by fix, as parallel coverage should be enabled here.

It was accidentally disabled thinking that if missing it will use the default.

adiroiban:
I think that this should be merged as it is as it already has enough improvements.

## Coveralls fixes
* Publish from Azure as a fake Circle-CI https://coveralls.io/builds/38390418
* Publish from Github actions https://coveralls.io/builds/38390167
* Paths for both Azure and GitHub are now normalized 
* NOT-DONE - PR coverage not yet merged

## Codecov
* Coverage merged for both Azure and GitHub to have unified view.

## General changes

* Coverage publishing from CI moved outside of tox

altendky (approved):
I'll say that I'm feeling pretty itchy about all this.  But, given the discussion, I feel like that's just the way it is going to be regardless because of the multiple CI systems * the multiple coverage presenters * the myriad oddities.  Oh well, it's mutable, let's go for it.

adiroiban:
Thanks for the review... This is not over yet :) Part 2 at https://github.com/twisted/twisted/pull/1543

PR #1576, opened 2021-04-01
10166: Improve type annotations of some Twisted utility functions.

wiml:
## Scope and purpose

I've been working on type-annotating some other twisted code and it's nice to have correct annotations on these small utility functions.

Notes:

* I didn't bother annotating a lot of the deprecated functions, since I figure they'll be gone before too long anyway.
* I had to rewrite `_matchingString()` so that mypy could understand it, but that's a small change. It has the tiny advantage that the new implementation won't encode/decode `constantString` if it doesn't have to.
* I also converted a `map(lambda...)` in `stringyString()` into a generator expression, again because mypy didn't understand the previous code, but also it reads clearer to me.

## Contributor Checklist:

* [X] The associated [ticket in Trac](https://twistedmatrix.com/trac/ticket/10166) 
* [X] I ran `tox -e lint` 
* [X] I have created a newsfragment in src/twisted/newsfragments/10166.misc 
* [X] I have submitted the associated Trac ticket for review by adding the word `review`

```

(as long as you wish)
```


src/twisted/python/compat.py
```diff
@@ -205,7 +205,11 @@ def currentframe(n: int = 0) -> FrameType:
     return f
 
 
-def execfile(filename, globals, locals=None):
+def execfile(
+    filename: Union[str, bytes, os.PathLike],
+    globals: Dict[str, Any],
```
graingert, line 214:
```suggestion
    globals: Optional[Dict[str, Any]],
```
exec seems to take an optional globals?

src/twisted/python/compat.py
```diff
@@ -429,7 +435,9 @@ def intToBytes(i: int) -> bytes:
     return b"%d" % (i,)
 
 
-def lazyByteSlice(object, offset=0, size=None):
+def lazyByteSlice(
+    object: Any, offset: int = 0, size: Optional[int] = None
```
graingert:
```python
    object: ReadableBuffer, offset: int = 0, size: Optional[int] = None
```

memoryview takes a ReadableBuffer https://github.com/python/typeshed/blob/16ae4c61201cd8b96b8b22cdfb2ab9e89ba5bcf2/stdlib/builtins.pyi#L608

defined, https://github.com/python/typeshed/blob/16ae4c61201cd8b96b8b22cdfb2ab9e89ba5bcf2/stdlib/_typeshed/__init__.pyi#L171

You'll need to put this type alias somewhere:

`ReadableBuffer = Union[bytes, bytearray, memoryview, array.array, mmap.mmap]`

src/twisted/internet/_idna.py
```diff
@@ -28,10 +28,10 @@ def _idnaBytes(text):
     except ImportError:
         return text.encode("idna")
     else:
-        return idna.encode(text)
+        return idna.encode(text)  # type: ignore[no-any-return]
```
graingert, line 31: this has a type annotation, so I think the twisted mypy setup might be wrong

https://github.com/kjd/idna/blob/master/idna/core.py#L354

I think the all_deps_nonplatform extra will need to be installed in the mypy toxenv

graingert: @wsanchez ^

mthuurne: The last release of `idna` was on Jan 4 this year, while `py.typed` was added on Jan 19. So unless you'd pin a pre-release snapshot, mypy won't see any type annotations.

src/twisted/python/compat.py
wiml: That's true. Expanding a bit on `execfile`'s doc comment, though, it doesn't seem like useful behavior for `t.p.compat.execfile` to support: it would execute the file in the context of the `twisted.python.compat` module, which is almost certainly not something that people want to be doing. (If they *do* want that behavior, it's probably good practice for it to be done explicitly at the call site.)

I'd even consider adding a runtime check in `execfile` to prevent people from passing None despite the typechecker.

src/twisted/python/compat.py
wiml: Ah, nice find. The typeshed file says it's ok to use its definition from outside of typeshed but it doesn't guarantee stability — do you think it's better to copy the definition out into Twisted (meaning, Twisted's definition will become silently out of date if Python changes), or reference typeshed (meaning, Twisted may fail to typecheck if `_typeshed.py` changes)? I'm inclined to the latter, because it means that breakage would be immediately visible instead of hidden. (And typeshed doesn't have so much code churn that I would expect the definition to change soon.)

src/twisted/python/compat.py
```diff
@@ -355,7 +359,7 @@ def nativeString(s: AnyStr) -> str:
     return s
 
 
-def _matchingString(constantString, inputString):
+def _matchingString(constantString: Union[bytes, str], inputString: AnyStr) -> AnyStr:
```
wiml: On further consideration, this isn't the correct type for `_matchingString`. The `typing.AnyString` typevar can be bound to str or bytes, but not to `Union[str, bytes]`, which prevents `_matchingString` from typechecking if it's used in a function which must support both possibilities. I'll be pushing another commit which fixes this.

wiml:
Currently the only failing check is `codecov/patch` and its only complaint is that the import guarded by `if typing.TYPE_CHECKING:` is never executed. I think that's a nonissue...

mthuurne:
> Currently the only failing check is `codecov/patch` and its only complaint is that the import guarded by `if typing.TYPE_CHECKING:` is never executed. I think that's a nonissue...

In `.coveragerc` there is an exclude for `if TYPE_CHECKING`; perhaps a line should be added there to exclude `if typing.TYPE_CHECKING` as well.

Or maybe `if (typing\.)?TYPE_CHECKING:$`, since the exclude list is a regex.

src/twisted/python/compat.py
```diff
@@ -27,6 +27,7 @@
 import platform
 import socket
 import sys
+import typing
```
wsanchez, line 30: Import `TYPE_CHECKING` with the `typing` imports below instead.

src/twisted/python/text.py
```diff
@@ -7,8 +7,11 @@
 Miscellany of text-munging functions.
 """
 
+import typing
```
wsanchez, line 10: import `IO` in the list below instead.

wsanchez: (Or `TextIO`)

wiml: I'd rather have the `IO` type from `typing` be explicitly qualified, because it's easy to confuse it with the more concrete imports from the `io` module. It should be clear at the point of use that it refers to an arbitrary "file-like object" rather than something that descends from a particular class.

src/twisted/python/compat.py
wiml: Addressed in 8f483dc I think.

src/twisted/internet/_idna.py
wsanchez: I remain confused about Twisted's build setup, but if `all_deps_nonplatform` installs dependencies, we do want it in the `mypy` Tox environment, so that it can use whatever annotations are (or become) available there.

src/twisted/python/compat.py
wsanchez: I'm supportive of a narrower type bus we think that's correct, but I'd avoid the runtime check, as that will technically trigger a compatibility concern that's unrelated to the type hint goal of this PR.

That also gives a user with a use case we hadn't considered a change to complain about the type hints without also getting runtime errors.

src/twisted/python/compat.py
```diff
@@ -377,13 +399,17 @@ def _matchingString(constantString, inputString):
     @rtype: the type of C{inputString}
     """
     if isinstance(constantString, bytes):
-        otherType = constantString.decode("ascii")
-    else:
-        otherType = constantString.encode("ascii")
-    if type(constantString) == type(inputString):
-        return constantString
+        uniString = constantString.decode("ascii")
```
wsanchez, line 402: Why decode here instead of in the `else` clause? Seems like it might get computed here and then thrown away.

src/twisted/python/compat.py
```diff
@@ -377,13 +399,17 @@ def _matchingString(constantString, inputString):
...
+        if isinstance(inputString, bytes):
+            return constantString
+        else:
+            return uniString
     else:
-        return otherType
+        byteString = constantString.encode("ascii")
```
wsanchez, line 408: Same here. Might get computed and then tossed.

src/twisted/python/util.py
```diff
@@ -6,6 +6,7 @@
 import errno
 import os
 import sys
+import typing
```
wsanchez, line 9: Just import `TypeVar` with the other typing imports…?

.coveragerc
```diff
@@ -15,6 +15,6 @@ source=
 precision = 2
 ignore_errors = True
 exclude_lines =
-    if TYPE_CHECKING
+    if (typing\.)?TYPE_CHECKING
```
wsanchez, line 18: Seems unnecessary if we consistently import from `typing` instead of typing itself, which so far seems the preferred practice (to me, anyway).

src/twisted/python/compat.py
```diff
@@ -355,7 +363,21 @@ def nativeString(s: AnyStr) -> str:
...
+# bytes, it can't be a Union[] at the call site.)
+#
+# We can't use a typevar with bound=Union[str, bytes] here, either,
+# since even if the caller passes in some subclass of str or bytes,
+# our return value will be a plain old str or bytes.
+_WhateverString = TypeVar("_WhateverString", str, bytes, Union[str, bytes])
+
+
+def _matchingString(
+    constantString: Union[bytes, str], inputString: _WhateverString
```
graingert, line 379: I think this is cleaner with `@overload`

src/twisted/python/compat.py
```diff
@@ -487,7 +515,7 @@ def bytesEnviron():
     return {encodekey(x): encodevalue(y) for x, y in os.environ.items()}  # type: ignore[call-arg]
 
 
-def _constructMethod(cls, name, self):
+def _constructMethod(cls: type, name: str, self: object) -> _MethodType:
```
graingert, line 518: How about?

```python
def _constructMethod(cls: Type[T], name: str, self: T) -> _MethodType:
```

src/twisted/python/util.py
```diff
@@ -546,18 +567,28 @@ def __init__(self, intervals, default=60):
...
     def __init__(self, i, d):
 
+        # Note that we're using list as a "mutable tuple" here; the
+        # first and third element of each entry change as we
+        # execute. The actual type of each entry is [float|int, Real, int]
+
+        # However, mypy doesn't understand non-uniform lists,
+        # so it can't properly typecheck this variable.
+        # Also note that mypy doesn't correctly handle the types from
+        # numbers.* yet, at least in version 0.812.
```
graingert, line 584: mypy explicitly won't support numbers module

src/twisted/python/compat.py
wiml: I mentioned this in the commit message, but: the function is explicitly documented as encoding or decoding its argument even i it doesn't have to. I'd rather not break that stated contract, even if it is a silly one.

src/twisted/python/compat.py
wiml: I don't think `@overload` can provide the stated semantics? I could be wrong.

src/twisted/python/compat.py
wiml: (See above)

src/twisted/python/compat.py
graingert:
```
@overload
def matchingString(constantString: str | bytes, inputString: str) -> str: ...

@overload
def matchingString(constantString: str | bytes, inputString: bytes) -> bytes: ...
 
def matchingString(constantString: str | bytes, inputString: str | bytes) - > str | bytes:
    ...
```

PR #1577, opened 2021-04-01
Use tox --skip-pkg-install

altendky:
## Scope and purpose

## Contributor Checklist:

* [ ] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/
* [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [ ] The title of the PR starts with the associated Trac ticket number (without the `#` character).

```

(as long as you wish)
```

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -47,4 +47,4 @@ jobs:
 
     - name: Check the API documentation
       run: |
-        tox -q -e apidocs
+        tox --quiet --skip-pkg-install -e apidocs
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -43,8 +43,8 @@ jobs:
 
     - name: Check the narrative documentation
       run: |
-        tox -q -e narrativedocs
+        tox --skip-pkg-install -e narrativedocs
 
     - name: Check the API documentation
       run: |
-        tox --quiet --skip-pkg-install -e apidocs
+        tox --skip-pkg-install -e apidocs
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -31,4 +31,4 @@ jobs:
     - name: Lint
       run: |
         python --version
-        tox -q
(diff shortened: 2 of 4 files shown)


altendky:
Kick CI

wsanchez:
Why put this in the CI invocation instead of the `tox.ini` file?  Local Tox runs should be similar to CI, ideally.

I'm not sure this is appropriate for the `mypy` environment; you need to install dependencies in order to get type stubs… looks like that's not apparent because `ignore_missing_imports` is always True… https://github.com/twisted/twisted/pull/1578 should fix that.

.github/workflows/docs.yaml
```diff
@@ -47,4 +47,4 @@ jobs:
 
     - name: Check the API documentation
       run: |
-        tox -q -e apidocs
+        tox --quiet --skip-pkg-install -e apidocs
```
adiroiban: I think that with --skip-pkg-install we can enable back non-quiet mode.

The -q was there to suppress the deps installation info

altendky:
@wsanchez first we run with `--notest` to build the environment, then we report information about that environment, then we (would now) `--skip-pkg-install` to use the already-built env.  It's just separating the env build from the env run.  So, the deps should have been installed.  Just in the first invocation and not (reinstalled) in the second.

altendky:
Err...  we don't yet report the info from the env.  But maybe will. https://github.com/twisted/twisted/pull/1548

altendky:
Looks like it might be time for a ticket...

adiroiban:
Thanks. 

I like that "install deps" is a separate step and we can have quick view on how long it takes  ... and if there is no performance penalty I am happy to see it in trunk.

Is funny there is not much difference between nodeps (https://github.com/twisted/twisted/runs/2335304305?check_suite_focus=true#step:8:1) and alldeps  

Thanks again for this work

PR #1578, opened 2021-04-02
10168: Enable ignore_missing_imports only where it is needed.

wsanchez:
The current `mypy.ini` file sets `ignore_missing_imports=True` globally.

We should instead disable it specifically for packages that are known to lack type stubs. That ensures that for those where we expect type stubs, the Mypy error is not suppressed if they are missing.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10168

```

(as long as you wish)
```

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -821,7 +821,15 @@ warn_return_any = False
 
 # Don't complain about imports of fake modules used in tests
 
+[mypy-__main__]
+ignore_missing_imports = True
+
 [mypy-automat]
+ignore_missing_imports = True
+
+[mypy-dummy]
+ignore_missing_imports = True
+[mypy-dummy.*]
 ignore_missing_imports = True
 
 [mypy-goodpackage]
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,8 +3,6 @@
 namespace_packages = True
 plugins=mypy_zope:plugin
 
-ignore_missing_imports = True
-
 # Increase our expectations
 
 check_untyped_defs       = True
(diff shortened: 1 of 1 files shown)


mypy.ini
```diff
@@ -821,3 +819,213 @@ allow_incomplete_defs = True
 [mypy-twisted.words.protocols.jabber.jid]
 warn_return_any = False
 
+# Don't complain about imports of fake modules used in tests
+
+[mypy-__main__]
+ignore_missing_imports = True
+
+[mypy-automat]
```
adiroiban, line 827: Is this only testing?

I see this in src/twisted/application/internet.py

```
from automat import MethodicalMachine
```

mypy.ini
```diff
@@ -821,3 +819,213 @@ allow_incomplete_defs = True
...
+[mypy-twisted_renamed_helper]
+ignore_missing_imports = True
+
+[mypy-twisted_toptobottom_temp]
+ignore_missing_imports = True
+
+[mypy-uberpackage]
+ignore_missing_imports = True
+
+[mypy-utmp]
```
adiroiban, line 856: Looks like this is used in `src/twisted/conch/unix.py` , outside of testing

adiroiban (approved):
I guess that this is ok.

I don't think that it make sense to manually double check if any of the dependencies do have mypy hints.

I guess that once one Twisted developer cares about typing and knows that for example h2
has typing, the list can be later updated.

Maybe @graingert  might want to double check... if he has time.

Thanks!

graingert:
I've created https://github.com/twisted/twisted/pull/1584 which uses `# type: ignore[import]` which will  automatically remind us to remove the `ignore` annotation when stubs are made available

graingert (changes requested):
I'm going to request changes because I think my solution in #1584 will be more convenient, because mypy will warn us when modules gain type hints/stubs, and "weird"/technical imports like `__main__`, `package.module`, `twisted_rebuild_fakelib` make more sense to be ignored locally than globally as they will never be stubbed and just clutter the mypy.init

https://github.com/twisted/twisted/pull/1584/files#diff-8f4abfb76eeb666cb2782ad79aee6d2bfaa9eaa0911b28edcd57f6e601f9c8d4R538

if you disagree feel free to dismiss my review

graingert:
I'm closing this in favour of #1584 for now

PR #1580, opened 2021-04-05
8199 Use constant-time compare_digest() for MACs and similar secret data.

wiml:
## Scope and purpose

Using a non–constant-time comparison for digest verification can leak some secret information to an attacker. This patch makes sure to use the `compare_digest()` function for such comparisons.

The motivating change is the one in `transport.py`. Probably not all of these comparisons are sensitive, but since the cost is minimal I erred on the side of caution.

## Contributor Checklist:

* [X] The [associated ticket in Trac](https://twistedmatrix.com/trac/ticket/8199)
* [X] I ran `tox -e lint`
* [X] The title of the PR starts with the associated Trac ticket number

```
Merge pull request #123 from wiml/8199-conch-constant-time-compare
Author: wiml
Fixes: ticket:8199
Refs: ticket:4536

Use constant-time compare_digest() for MACs and similar secret data.
```


graingert:
is there a reason to use hmac.compare_digest vs secrets.compare_digest?

wiml:
Not really — they're actually the same function. I'm using `hmac` when comparing a digest that was just generated by a HMAC because the module is already right there in scope and familiar to the reader.

Anyway it looks like one change is in code that isn't covered by unit tests so I'm going to withdraw this PR until I can write up a test that exercises it.

adiroiban:
Many thanks for this PR

Just for reference, there is a very old discussion on this topic https://twistedmatrix.com/trac/ticket/4536

But I think that this PR should proceed without waiting for that ticket.

As long as we have tests, this should be good to merge :)

As the same time, this is already an improvement and is updating code that don't previously has unit tests.

so @wiml if you find it hard to write a test let me know... I will see if I can help with the test or maybe we can get it merged as it is by asking for an exception.

wiml:
On further inspection, it's not that there is no test for that code, it's that (a) the test is skipped on all of our CI runners because (b) the code has been deprecated in favor of a different class that's easier to test properly.

Since it's in deprecated code and is a change that is "probably not" necessary for security anyway, I'm going to re-do this PR without the changes to `conch/checkers.py`.

(The corresponding comparison in the non-deprecated code is delegated to the `Key` object, which maybe deserves its own investigation for timing leaks but I think is out of scope for this particular PR.)

wiml:
> Just for reference, there is a very old discussion on this topic https://twistedmatrix.com/trac/ticket/4536

In fact, two of the changes I made here are discussed in that ticket! But were backed out in 30d2169 for compatibility reasons. Since we're on py3+ now (and can rely on the existence of {`hmac`,`secrets`}`.compare_digest()`) and these two comparisons are fairly guaranteed to be `bytes` I think this can go forward.

I'll look at that trac bug some more, though — it may have become easier to fix in the few years since it was last looked at.

glyph (approved):
Looks simple, straightforward, and good!

PR #1581, opened 2021-04-06
[10171] Fix coverage run random failures on PYPY due to SQLite error.

adiroiban:
## Scope and purpose

Coverage runs on PYPY fails

https://github.com/twisted/twisted/pull/1577/checks?check_run_id=2274177993#step:9:751

Experimental fix discussed here https://github.com/nedbat/coveragepy/issues/1010

This is the Twisted integration branch.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10171

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10171

Fix coverage run on PYPY due to SQLite error.
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,7 @@ dev =
     # TODO: support python-subunit in py3.10 https://twistedmatrix.com/trac/ticket/10115
     python-subunit ~= 1.4; python_version < "3.10"
     twistedchecker ~= 0.7
-    coverage ~= 5.5
+    #WIP: coverage ~= 5.5
 
 tls =
     pyopenssl >= 16.0.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,9 @@ dev =
     # TODO: support python-subunit in py3.10 https://twistedmatrix.com/trac/ticket/10115
     python-subunit ~= 1.4; python_version < "3.10"
     twistedchecker ~= 0.7
-    #WIP: coverage ~= 5.5
+    # To be pinned to a minor version once PYPY `coverage run` error is fixed.
+    # See: https://twistedmatrix.com/trac/ticket/10171
+    coverage
 
 tls =
     pyopenssl >= 16.0.0
--- a/tox.ini
+++ b/tox.ini
@@ -56,6 +56,10 @@ extras =
 ;; dependencies that are not specified as extras
 deps =
     lint: pre-commit
+    # For now we use a dev version of coverage with a fix for `coverage run`
+    # on PYPY.
+    # This should be moved
+    # See: https://twistedmatrix.com/trac/ticket/10171
     withcov: https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
 
 ; All environment variables are passed.
--- a/tox.ini
+++ b/tox.ini
@@ -56,6 +56,7 @@ extras =
 ;; dependencies that are not specified as extras
 deps =
     lint: pre-commit
+    withcov: https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
 
 ; All environment variables are passed.
 passenv = *


adiroiban:
My plan is to have this is trunk for a while... 1 month. If we no longer observer pypy coverage errors. we can consider it fixed. report upstream and use the released version with the fix, when ready,

Thanks!

setup.cfg
```diff
@@ -67,7 +67,9 @@ dev =
     # TODO: support python-subunit in py3.10 https://twistedmatrix.com/trac/ticket/10115
     python-subunit ~= 1.4; python_version < "3.10"
     twistedchecker ~= 0.7
-    coverage ~= 5.5
+    # To be pinned to a minor version once PYPY `coverage run` error is fixed.
+    # See: https://twistedmatrix.com/trac/ticket/10171
+    coverage
```
altendky, line 72: Can we drop the `tox.ini` change if we do this here?

```suggestion
    coverage @ https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
```

altendky: Hum, actually I suppose...

```suggestion
    coverage ~= 5.5 @ https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
```

Or describe the version it presently lists, though this is going a bit far probably.  The first suggestion in the previous comment is likely sufficient and useful since it isolates the change to only needing to be undone in one place.

https://github.com/nedbat/coveragepy/blob/nedbat/another-1010/coverage/version.py#L8
```python
version_info = (5, 5, 1, "alpha", 0)
```

```suggestion
    coverage >= 5.5.1, < 6 @ https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
```

adiroiban: I don't think that we can drop it.

I tried your last suggestion and I got

```
  pkg_resources.extern.packaging.requirements.InvalidRequirement: Parse error at "'@ https:'": Expected stringEnd
```

--------------

I think that a bit of duplication is ok and there is not much gain by "doing it right(tm)" 

I think that this is no longer supported...and it required an extra `dependency_links`

https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html?highlight=dependency_links#dependencies-that-aren-t-in-pypi

altendky: `@` is a new syntax but apparently it doesn't work there...  I'll have to go learn.  We can move on.

https://www.python.org/dev/peps/pep-0508/#examples

altendky: (this is not in `setup.cfg`, just an example in a place without `dependency_links`.)

```console
$ venv/bin/pip install 'coverage @ https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip'
Collecting coverage@ https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
  Downloading https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
     / 1.0 MB 1.5 MB/s
Building wheels for collected packages: coverage
  Building wheel for coverage (setup.py) ... done
  Created wheel for coverage: filename=coverage-5.5.1a0-cp39-cp39-linux_x86_64.whl size=243403 sha256=de379aeb239ebb6321429001d7085a07748ffb5277bd9692e50c72bf692b885a
  Stored in directory: /home/altendky/.cache/pip/wheels/3e/b1/2f/368d0a88f96d28bf2f5f1e74203b32ac575b293fe0d40f4bcd
Successfully built coverage
Installing collected packages: coverage
Successfully installed coverage-5.5.1a0
```

altendky (approved):

setup.cfg
altendky: Mm, looks like maybe just the version spec is the issue.  Either version or @ but not both.  Which seems reasonable.

adiroiban:
Thanks for the review. I hope a new coverage release will be made soon and we can then remove this.

PR #1582, opened 2021-04-10
10173  Remove py2.4 sets module support from jelly.py

wiml:
Pretty straightforward: this code can never be executed any more, nor can the tests for it ever not be skipped.

* [X] The [associated ticket in Trac](https://twistedmatrix.com/trac/ticket/10173#ticket)
* [X] I ran `tox -e lint`

```

Author: wiml
Fixes: ticket:10173
Refs: ticket:8616

Remove py2.4 sets module support from jelly.py
```

### DIFF
--- a/src/twisted/spread/jelly.py
+++ b/src/twisted/spread/jelly.py
@@ -86,19 +86,6 @@ from incremental import Version
 _SetTypes = [set]
 _ImmutableSetTypes = [frozenset]
 
-with warnings.catch_warnings():
-    warnings.simplefilter("ignore", category=DeprecationWarning)
-    try:
-        import sets as _sets
-    except ImportError:
-        # sets module is deprecated in Python 2.6, and gone in
-        # Python 3
-        _sets = None
-    else:
-        _SetTypes.append(_sets.Set)
-        _ImmutableSetTypes.append(_sets.ImmutableSet)
-
-
 DictTypes = (dict,)
 
 None_atom = b"None"  # N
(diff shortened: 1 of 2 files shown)


src/twisted/spread/jelly.py
```diff
@@ -86,19 +86,6 @@
 _SetTypes = [set]
 _ImmutableSetTypes = [frozenset]
```
graingert: I'd be tempted to remove these too

graingert (approved):

src/twisted/spread/jelly.py
wiml: I considered that, or perhaps changing them to tuples at least. I'm on the fence about whether it's likely that we'll want some other classes in there in the future, but perhaps simplicity now is a better option

src/twisted/spread/jelly.py
```diff
@@ -538,9 +533,9 @@ def jelly(self, obj):
                     sxp.append(dictionary_atom)
                     for key, val in obj.items():
                         sxp.append([self.jelly(key), self.jelly(val)])
-                elif objType in _SetTypes:
+                elif objType is set:
                     sxp.extend(self._jellyIterable(set_atom, obj))
-                elif objType in _ImmutableSetTypes:
+                elif objType is frozenset:
```
graingert, line 538: @wiml this is a slightly breaking change:

```python
class TypeEqAllMeta(type):
    def __eq__(self, other):
        return True

class TypeEqAll(metaclass=TypeEqAllMeta):
    pass


v = TypeEqAll()

print(f"{type(v) in (set,)=} {type(v) in (frozenset,)=}") # type(v) in (set,)=True type(v) in (frozenset,)=True
print(f"{type(v) is set=} {type(v) is frozenset=}")  # type(v) is set=False type(v) is frozenset=False
```

which is probably fine - but something to be aware of

wiml: That's a good point I hadn't considered (classes which may be `==` without being `is`). As you say this is unlikely to break anything but I appreciate the feedback.

PR #1584, opened 2021-04-12
10168: do not ignore missing imports (and use incremental from trunk for mypy toxenv)

graingert:
## Scope and purpose

The current mypy.ini file sets ignore_missing_imports=True globally.

We should instead disable it specifically for packages that are known to lack type stubs. That ensures that for those where we expect type stubs, the Mypy error is not suppressed if they are missing.

It's extremely inconvenient adding type annotations when you do something like:

```
from typeing_extensions import Protocol

class FooPortocol(Protocol[_T_IFileHandle_contra):
    ... 

... # 500 lines of Generic type soup
```

only to realise that Protocol was considered an `Any` and so all the type-soup simply appears to work but actually short circuits to Any.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10168 

```

(as long as you wish)
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -116,6 +116,12 @@ windows_platform =
 osx_platform =
     %(macos_platform)s
 
+mypy =
+    mypy==0.812
+    mypy-zope==0.2.13
+    %(dev)s
+    %(all_non_platform)s
+
 [options.packages.find]
 where = src
 
--- a/src/twisted/application/runner/_runner.py
+++ b/src/twisted/application/runner/_runner.py
@@ -13,7 +13,7 @@ from typing import Any, TextIO, Callable, Mapping
 
 from attr import attrib, attrs, Factory
 
-from constantly import NamedConstant  # type: ignore[import] # type: ignore[import]
+from constantly import NamedConstant  # type: ignore[import]
 
 from twisted.internet.interfaces import IReactorCore
 from twisted.logger import (
--- a/src/twisted/application/runner/_runner.py
+++ b/src/twisted/application/runner/_runner.py
@@ -13,7 +13,7 @@ from typing import Any, TextIO, Callable, Mapping
 
 from attr import attrib, attrs, Factory
 
-from constantly import NamedConstant
+from constantly import NamedConstant  # type: ignore[import] # type: ignore[import]
 
 from twisted.internet.interfaces import IReactorCore
 from twisted.logger import (
--- a/src/twisted/copyright.py
+++ b/src/twisted/copyright.py
@@ -8,9 +8,9 @@ Copyright information for Twisted.
 
 __all__ = ["copyright", "disclaimer", "longversion", "version"]
 
-from twisted import __version__ as version, version as longversion
+from twisted import __version__ as version, version as _longversion
 
-longversion = str(longversion)
+longversion = str(_longversion)
 
 copyright = """\
 Copyright (c) 2001-2021 Twisted Matrix Laboratories.
--- a/src/twisted/internet/gireactor.py
+++ b/src/twisted/internet/gireactor.py
@@ -24,7 +24,7 @@ On Python 3, pygobject v3.4 or later is required.
 from twisted.internet.error import ReactorAlreadyRunning
 from twisted.internet import _glibbase
 from twisted.python import runtime
-import gi.pygtkcompat  # type: ignore[import] # type: ignore[import]
+import gi.pygtkcompat  # type: ignore[import]
 from gi.repository import GLib  # type: ignore[import]
 
 # We require a sufficiently new version of pygobject, so always exists:
--- a/src/twisted/internet/gireactor.py
+++ b/src/twisted/internet/gireactor.py
@@ -24,8 +24,8 @@ On Python 3, pygobject v3.4 or later is required.
 from twisted.internet.error import ReactorAlreadyRunning
 from twisted.internet import _glibbase
 from twisted.python import runtime
-import gi.pygtkcompat
-from gi.repository import GLib
+import gi.pygtkcompat  # type: ignore[import] # type: ignore[import]
+from gi.repository import GLib  # type: ignore[import]
 
 # We require a sufficiently new version of pygobject, so always exists:
 _pygtkcompatPresent = True
--- a/src/twisted/internet/test/process_helper.py
+++ b/src/twisted/internet/test/process_helper.py
@@ -13,7 +13,7 @@ def grandchild():
 def main():
     if sys.argv[1] == "child":
         if sys.argv[2] == "windows":
-            import win32api as api, win32process as proc
+            import win32api as api, win32process as proc  # type: ignore[import] # type: ignore[import]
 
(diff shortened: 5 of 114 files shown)


tox.ini
```diff
@@ -60,7 +60,7 @@ deps =
     # on PYPY.
     # This should be moved
     # See: https://twistedmatrix.com/trac/ticket/10171
-    withcov: https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.zip
+    withcov: https://github.com/nedbat/coveragepy/archive/refs/heads/nedbat/another-1010.tar.gz
```
graingert, line 63: the .tar.gz is 168KB smaller

tox.ini
```diff
@@ -156,13 +156,10 @@ commands = {toxinidir}/bin/admin/build-apidocs {toxinidir}/src/ apidocs
...
 [testenv:mypy]
-
 description = run Mypy (static type checker)
 
-deps =
-    mypy==0.812
-    mypy-zope==0.2.13
-
+extras = mypy
+deps = incremental @ https://github.com/twisted/incremental/archive/refs/heads/trunk.tar.gz
```
adiroiban, line 163: Is there a ticket on incremental for the issue why we need to use trunk ? Can you add it as comment?
Maybe instead of trunk we should pin on a certain commit?
Thanks!

tox.ini
adiroiban: thanks :)

setup.cfg
```diff
@@ -116,6 +116,12 @@ windows_platform =
 osx_platform =
     %(macos_platform)s
 
+mypy =
```
adiroiban, line 119: Minor comment.
Maybe add all mypy and mypy-zope directly to dev.
When I do local twisted development, I find it easier to just create a virtual env with [.dev] and get everything.

But if installing mypy in the other dev env takes a  long time, I am ok with using this separate extra.

adiroiban (commented):
Thanks the the PR.

The comment from here https://github.com/twisted/twisted/pull/1578#pullrequestreview-633653477 make sense to me.

I don't understand the description of this PR.

If by using the `# type: ignore[import]` we will get an automatic warning/error once the ignore package has type definition, I think that  this solution is better than #1578.

But I think that is best if @wsanchez weight his opinion. 

------

I would say that this needs minor changes the explain the drive-by incremental fix.

tox.ini
graingert:
```suggestion
# https://github.com/twisted/incremental/pull/69
deps = incremental @ https://github.com/twisted/incremental/archive/refs/heads/trunk.tar.gz
```

setup.cfg
graingert: Yeah mypy and mypy-zope are very tiny wheels I'd be happy to do that

graingert: @adiroiban ah that doesn't work because mypy needs `all_non_platform` and `dev`

adiroiban: Yes... so maybe add `all_non_platform` as part of `dev` :) as when you are doing development you most probably want it all

graingert: Probably a separate ticket

adiroiban:
@wiml do you have time to take a look as this PR  and weight your opinion ? Thanks!

setup.cfg
```diff
@@ -116,6 +116,12 @@ windows_platform =
...
+    mypy==0.812
+    mypy-zope==0.2.13
+    %(dev)s
+    %(all_non_platform)s
```
wsanchez, line 123: Ew `%`-formatting. `setup.cfg` doesn't allow f-strings?

src/twisted/application/runner/_runner.py
```diff
@@ -13,7 +13,7 @@
 
 from attr import attrib, attrs, Factory
 
-from constantly import NamedConstant
+from constantly import NamedConstant  # type: ignore[import] # type: ignore[import]
```
wsanchez: Is this doubled up intentionally?

src/twisted/internet/gireactor.py
```diff
@@ -24,8 +24,8 @@
 from twisted.internet.error import ReactorAlreadyRunning
 from twisted.internet import _glibbase
 from twisted.python import runtime
-import gi.pygtkcompat
-from gi.repository import GLib
+import gi.pygtkcompat  # type: ignore[import] # type: ignore[import]
```
wsanchez: Doubled up again here

src/twisted/internet/test/process_helper.py
```diff
@@ -13,7 +13,7 @@ def grandchild():
 def main():
     if sys.argv[1] == "child":
         if sys.argv[2] == "windows":
-            import win32api as api, win32process as proc
+            import win32api as api, win32process as proc  # type: ignore[import] # type: ignore[import]
```
wsanchez: doubled up

src/twisted/internet/tksupport.py
```diff
@@ -31,7 +31,7 @@
 
 from twisted.internet import task
 
-import tkinter.simpledialog as tkSimpleDialog
+import tkinter.simpledialog as tkSimpleDialog  # type: ignore[import] # type: ignore[import]
```
wsanchez: doubled up

src/twisted/web/_http2.py
```diff
@@ -22,12 +22,12 @@
...
 from zope.interface import implementer
 
-import priority
-import h2.config
-import h2.connection
-import h2.errors
-import h2.events
-import h2.exceptions
+import priority  # type: ignore[import]
+import h2.config  # type: ignore[import] # type: ignore[import]
```
wsanchez: doubled up

wsanchez (approved):
Looks good.
I'd give a pass to the coverage complaints for the `#type: ignore` annotations.

setup.cfg
wiml: This is an f-string–like interpolation, really, it's not `%`-formatting: https://docs.python.org/3/library/configparser.html#interpolation-of-values

src/twisted/copyright.py
```diff
@@ -8,9 +8,9 @@
 
 __all__ = ["copyright", "disclaimer", "longversion", "version"]
 
-from twisted import __version__ as version, version as longversion
+from twisted import __version__ as version, version as _longversion
 
-longversion = str(longversion)
+longversion = str(_longversion)
```
wiml, line 13: Duplicates part of pr #1576 ;)

src/twisted/test/test_failure.py
```diff
@@ -22,7 +22,7 @@
 from twisted.trial.unittest import SynchronousTestCase
 
 
-from cython_test_exception_raiser import raiser
+from cython_test_exception_raiser import raiser  # type: ignore[import]
```
wiml, line 25: This is a Twisted-organization module, perhaps we should open a ticket to annotate it so we can remove this `ignore` in the future

PR #1585, opened 2021-04-16
10172 Bump required version of mypy-zope to 0.3.0

graingert:
## Scope and purpose

mypy-zope 0.3.0 adds support for using the @overload decorator in interfaces ( see ​https://github.com/Shoobx/mypy-zope/issues/46 ), which is needed in at least one place in Twisted.

mypy-zope is also stricter on interface definitions, and so I've had to type: ignore those places

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10172

```

(as long as you wish)
```

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -1144,7 +1144,7 @@ class ReactorBase(PluggableResolverMixin):
             self.installNameResolver(_GAIResolver(self, self.getThreadPool))
             self.usingThreads = True
 
-        def callFromThread(
+        def callFromThread(  # type: ignore[override]
             self, f: Callable[..., Any], *args: object, **kwargs: object
         ) -> None:
             """
@@ -1199,7 +1199,7 @@ class ReactorBase(PluggableResolverMixin):
                 assert self.threadpool is not None
             return self.threadpool
 
-        def callInThread(
+        def callInThread(  # type: ignore[override]
             self, _callable: Callable[..., Any], *args: object, **kwargs: object
         ) -> None:
             """
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -1144,6 +1144,9 @@ class ReactorBase(PluggableResolverMixin):
             self.installNameResolver(_GAIResolver(self, self.getThreadPool))
             self.usingThreads = True
 
+        # `IReactorFromThreads` defines the first named argument as
+        # `callable: Callable[..., Any]` but this defines it as `f`
+        # really both should be defined using py3.8 positional only
         def callFromThread(  # type: ignore[override]
             self, f: Callable[..., Any], *args: object, **kwargs: object
         ) -> None:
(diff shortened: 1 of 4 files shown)


src/twisted/internet/base.py
```diff
@@ -1144,7 +1144,7 @@ def _initThreads(self) -> None:
             self.installNameResolver(_GAIResolver(self, self.getThreadPool))
             self.usingThreads = True
 
-        def callFromThread(
+        def callFromThread(  # type: ignore[override]
             self, f: Callable[..., Any], *args: object, **kwargs: object
         ) -> None:
```
graingert, line 1152: the trouble is `twisted.internet.interfaces.IReactorFromThreads` defines the first named argument as `callable: Callable[..., Any]` but this defines it as `f`

really both should be defined using py3.8 positional only
```
        def callFromThread(
            self, f: Callable[..., Any], \, *args: object, **kwargs: object
        ) -> None:
```

wsanchez: Can you copy this explanation to a comment?
We should be documenting ignores that aren't obvious.

src/twisted/python/components.py
```diff
@@ -337,7 +337,7 @@ def __init__(self, original):
     for name in iface:
         contents[name] = _ProxyDescriptor(name, originalAttribute)
     proxy = type("(Proxy for {})".format(reflect.qual(iface)), (object,), contents)
-    declarations.classImplements(proxy, iface)
```
wsanchez, line 340: This would probably benefit from a comment; not clear to me why it's needed.

wsanchez (approved):
OK, with a couple of comments added

src/twisted/internet/base.py
```diff
@@ -1144,7 +1144,7 @@ def _initThreads(self) -> None:
             self.installNameResolver(_GAIResolver(self, self.getThreadPool))
             self.usingThreads = True
 
-        def callFromThread(
+        def callFromThread(  # type: ignore[override]
```
graingert, line 1150:
```suggestion
        # `IReactorFromThreads` defines the first named argument as
        # `callable: Callable[..., Any]` but this defines it as `f`
        # really both should be defined using py3.8 positional only
        def callFromThread(  # type: ignore[override]
```

src/twisted/internet/base.py
```diff
@@ -1199,7 +1199,7 @@ def getThreadPool(self) -> ThreadPool:
                 assert self.threadpool is not None
             return self.threadpool
 
-        def callInThread(
+        def callInThread(  # type: ignore[override]
```
graingert, line 1208:
```suggestion
        # `IReactorInThreads` defines the first named argument as
        # `callable: Callable[..., Any]` but this defines it as `_callable`
        # really both should be defined using py3.8 positional only
        def callInThread(  # type: ignore[override]
```

src/twisted/python/components.py
```diff
@@ -337,7 +337,7 @@ def __init__(self, original):
...
+    declarations.classImplements(proxy, iface)  # type: ignore[misc]
```
graingert, line 342:
```suggestion
    # mypy-zope declarations.classImplements only works when passing
    # a concrete class type
    declarations.classImplements(proxy, iface)  # type: ignore[misc]
```

wsanchez (approved):
great thanks

src/twisted/python/components.py
wiml: I ran into this recently and filed [this PR against mypy-zope](https://github.com/Shoobx/mypy-zope/issues/50); not enturely sure if it's a mypy problem or a mypy-zope problem.

PR #1590, opened 2021-04-27
10186: Fix cancellation of deferred flattening

mthuurne:
## Scope and purpose

Fixes a regression introduced in #1549.

I ran into this problem when mypy flagged it while I was adding type annotations. Those type annotations will come in a future PR, but @glyph preferred this to get fixed quickly.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10186
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] The title of the PR starts with the associated Trac ticket number (without the `#` character).
* [x] I have updated the automated tests and checked that all checks for the PR are green. -> The red test is a known issue unrelated to this PR.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.
* [x] The merge commit will use the below format
  The first line is automatically generated by GitHub based on PR ID and branch name.
  The other lines generated by GitHub should be replaced.

```
Merge pull request #123 from twisted/4356-branch-name-with-trac-id

Author: mthuurne
Reviewer: glyph
Fixes: ticket:10186

Fix cancellation of deferred flattening.
Add a test case to demonstrate the bug.
```

PR #1591, opened 2021-04-27
Revert "10186: Fix cancellation of deferred flattening"

mthuurne:
Reverts twisted/twisted#1590

The fix should have read `lambda _: d.cancel()` instead of `lambda _: d.cancel`. Yet the new test passed, so that wasn't specific enough. And if I fix the code, the test itself starts causing another test case to fail, so it is probably leaving junk behind in the reactor.


mthuurne:
Actually, it is probably easier to merge the fix to the fix directly, so I'll close this PR and open a new one for that.

PR #1592, opened 2021-04-28
[10187] Fix IRC ctcpExtract for pypy 3.7.4.

adiroiban:
## Scope and purpose

Our CI is failing in trunk after the auto-upgrade to pypy 3.7.4 which uses py3.10.

The IRC tests are failing.

I have found that the source is twisted.words.protocols.irc.ctcpExtract

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/

```

Author: adiroiban
Reviewer: wiml
Fixes: ticket:10187

Update  twisted.words.protocols.irc.ctcpExtract for PYPY 3.7.4 new list and iterator behaviour.
```

### DIFF
--- a/src/twisted/words/protocols/irc.py
+++ b/src/twisted/words/protocols/irc.py
@@ -3678,10 +3678,10 @@ def ctcpExtract(message):
             normal_messages.append(messages.pop(0))
         odd = not odd
 
-    extended_messages[:] = filter(None, extended_messages)
-    normal_messages[:] = filter(None, normal_messages)
+    extended_messages[:] = list(filter(None, extended_messages))
+    normal_messages[:] = list(filter(None, normal_messages))
 
-    extended_messages[:] = map(ctcpDequote, extended_messages)
+    extended_messages[:] = list(map(ctcpDequote, extended_messages))
     for i in range(len(extended_messages)):
         m = extended_messages[i].split(SPC, 1)
         tag = m[0]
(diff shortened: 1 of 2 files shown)


src/twisted/words/protocols/irc.py
```diff
@@ -3678,10 +3678,10 @@ def ctcpExtract(message):
             normal_messages.append(messages.pop(0))
         odd = not odd
 
-    extended_messages[:] = filter(None, extended_messages)
-    normal_messages[:] = filter(None, normal_messages)
+    extended_messages[:] = list(filter(None, extended_messages))
```
adiroiban, line 3681: My suggestion is to have something like this

```suggestion
    extended_messages = list(filter(None, extended_messages))
```

I don't know why we need to update the list "in-place"

wiml: @adiroiban because the two lists are inserted into the `retval` dictionary earlier. (I'm not sure if it's written that way for a reason... The whole function could probably be tidied up a lot using some list comprehensions, if someone wanted to do that.)

wiml (approved):
LGTM. The style of `ctcpExtract` is a bit odd I agree but I don't think it's important to clean it up beyond what's here.

glyph:
I will land this since it's blocking builds.

src/twisted/words/protocols/irc.py
adiroiban: Ah... I see.

so the issue is here
```
extended_messages = []
normal_messages = []
retval = {"extended": extended_messages, "normal": normal_messages}
``` 

and the return dict could have been created at the end with

```
return {"extended": extended_messages, "normal": normal_messages}
```

I think that we can live with that.

adiroiban:
For reference upstream bug report https://foss.heptapod.net/pypy/pypy/-/issues/3451

PR #1593, opened 2021-04-28
10186: Re-fix cancellation of deferred flattening

mthuurne:
## Scope and purpose

The previous fix (#1590) was incorrect and even though it included a test case, the test wasn't sensitive enough to detect the problem.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10186
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] The title of the PR starts with the associated Trac ticket number (without the `#` character).
* [x] I have updated the automated tests and checked that all checks for the PR are green. -> failing test is a known issue unrelated to this PR
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.
* [x] The merge commit will use the below format
  The first line is automatically generated by GitHub based on PR ID and branch name.
  The other lines generated by GitHub should be replaced.

```
Merge pull request #123 from twisted/4356-branch-name-with-trac-id

Author: mthuurne
Reviewer: glyph
Fixes: ticket:10186

Properly fix the cancellation of deferred flattening and improve the test case.
```

PR #1594, opened 2021-04-29
10184: Add type annotations to twisted.web.template

mthuurne:
## Scope and purpose

The main goal is to provide type hints to projects using Twisted's template system or that are using just Stan.

Adding the type hints increased the amount of checking mypy can do on Twisted itself and this uncovered two bugs. One was fixed in #1590 + #1593, the other was pretty minor and is fixed in a dedicated commit in this PR.

Things to consider during the review:
(mostly copy-pasted from the commit comments)

Since a lot of functions consume or produce "Stan objects" or "objects can be flattened" and a lot of types can be flattened, I added a type alias for that: `Flattenable`. This may be useful for projects using Twisted as well, so I exported it as a public name from `twisted.web.template`.

The documentation for the `TagLoader` class used to state that it could load `IRenderable` providers, but the implementation actually accepts any flattenable object. The instance variable in which the to-be-loaded object is stored is even named `tag` while the `Tag` class does not implement `IRenderable`. Therefore I assume the docstring was incorrect and changed it to accept `Flattenable` instead. Note that `IRenderable` is a member of `Flattenable`, so the previously documented interface is supported by the annotation.

The `render` attribute of `Tag` will now raise `TypeError` when anything other than a string is assigned to it via `__call__`. Previously, the assignment was accepted but it would break during rendering.

The `request` argument to `twisted.web.template.renderElement()` can now be any `IRequest` implementer instead of only `twisted.web.server.Request`. The only reason it was limited to `server.Request` before was that the `site` attribute was inspected to decide whether to print a minimal or detailed error message. We now pick the minimal message for other request types, to be on the safe side (the detailed message might leak sensitive information).

For the `XMLFile` path, the annotation does not include types for which support is deprecated. I assume projects type checking their code do not want to use a feature that was deprecated 9 years ago.

I removed `@type` and `@rtype` docstring fields that provided no additional value over the type annotations. In quite a few cases, the documented types were incorrect, so this will increase the accuracy of the documentation.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10184

```

Author: mthuurne
Reviewer: wsanchez
Fixes: ticket:10184

Add type annotations to `twisted.web.template`.
```

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1262,9 +1262,17 @@ class DeferredList(Deferred[List[Tuple[bool, _DeferredResultT]]]):  # type: igno
             logged.  This does not prevent C{fireOnOneErrback} from working.
         """
         self._deferredList = list(deferredList)
+
         self.resultList: List[Optional[Tuple[bool, _DeferredResultT]]] = [None] * len(
             self._deferredList
         )
+        """
+        The final result, in progress.
+        Each item in the list corresponds to the L{Deferred} at the same
+        position in L{_deferredList}. It will be L{None} if the L{Deferred}
+        did not complete yet, or a C{(success, result)} pair if it did.
+        """
+
         Deferred.__init__(self)
         if len(self._deferredList) == 0 and not fireOnOneCallback:
             self.callback([])
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -959,7 +959,11 @@ class Deferred(Awaitable[_DeferredResultT]):
     # type note: base class "Awaitable" defined the type as:
     #     Callable[[], Generator[Any, None, _DeferredResultT]]
     #     See: https://github.com/python/typeshed/issues/5125
-    __await__ = __iter__  # type: ignore[assignment]
+    #     When the typeshed patch is included in a mypy release,
+    #     this method can be replaced by `__await__ = __iter__`.
+    def __await__(self) -> Generator[Any, None, _DeferredResultT]:
+        return self.__iter__()  # type: ignore[return-value]
+
     __next__ = send
 
     def asFuture(self, loop: AbstractEventLoop) -> "Future[_DeferredResultT]":
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -16,12 +16,14 @@ from sys import exc_info, version_info
 import traceback
 from types import GeneratorType, MappingProxyType
 from typing import (
+    TYPE_CHECKING,
     Any,
     Awaitable,
     Callable,
     Coroutine,
     Generator,
     Generic,
+    Iterable,
     List,
     Mapping,
     NoReturn,
@@ -31,6 +33,7 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    overload,
 )
 from typing_extensions import Literal
 import warnings
@@ -1163,10 +1166,43 @@ class FirstError(Exception):
         return -1
 
 
-_DeferredListResult = Tuple[bool, object]
+if TYPE_CHECKING:
+
+    # The result type is different depending on whether fireOnOneCallback
+    # is True or False.  The type system is not flexible enough to handle
+    # that in a class definition, so instead we pretend that DeferredList
+    # is a function that returns a Deferred.
+
+    @overload
+    def _DeferredList(
+        deferredList: Iterable[Deferred[_T]],
+        fireOnOneCallback: Literal[True],
+        fireOnOneErrback: bool = False,
+        consumeErrors: bool = False,
+    ) -> Deferred[Tuple[_T, int]]:
+        ...
+
+    @overload
+    def _DeferredList(
+        deferredList: Iterable[Deferred[_T]],
+        fireOnOneCallback: Literal[False] = False,
+        fireOnOneErrback: bool = False,
+        consumeErrors: bool = False,
+    ) -> Deferred[List[Tuple[bool, _T]]]:
+        ...
+
+    def _DeferredList(
+        deferredList: Iterable[Deferred[_T]],
+        fireOnOneCallback: bool = False,
+        fireOnOneErrback: bool = False,
+        consumeErrors: bool = False,
+    ) -> Deferred[Any]:
+        ...
+
+    DeferredList = _DeferredList
 
 
-class DeferredList(Deferred[object]):
+class DeferredList(Deferred[List[Tuple[bool, _DeferredResultT]]]):  # type: ignore[no-redef]
     """
     L{DeferredList} is a tool for collecting the results of several Deferreds.
 
@@ -1194,7 +1230,7 @@ class DeferredList(Deferred[object]):
 
     def __init__(
         self,
-        deferredList: List[Deferred[object]],
+        deferredList: Iterable[Deferred[_DeferredResultT]],
         fireOnOneCallback: bool = False,
         fireOnOneErrback: bool = False,
         consumeErrors: bool = False,
@@ -1202,7 +1238,7 @@ class DeferredList(Deferred[object]):
         """
         Initialize a DeferredList.
 
-        @param deferredList: The list of deferreds to track.
+        @param deferredList: The deferreds to track.
         @param fireOnOneCallback: (keyword param) a flag indicating that this
             L{DeferredList} will fire when the first L{Deferred} in
             C{deferredList} fires with a non-failure result without waiting for
@@ -1226,12 +1262,12 @@ class DeferredList(Deferred[object]):
             logged.  This does not prevent C{fireOnOneErrback} from working.
         """
         self._deferredList = list(deferredList)
-        self.resultList: List[Optional[_DeferredListResult]] = [None] * len(
+        self.resultList: List[Optional[Tuple[bool, _DeferredResultT]]] = [None] * len(
             self._deferredList
         )
         Deferred.__init__(self)
         if len(self._deferredList) == 0 and not fireOnOneCallback:
-            self.callback(self.resultList)
+            self.callback([])
 
         # These flags need to be set *before* attaching callbacks to the
         # deferreds, because the callbacks use these flags, and will run
@@ -1251,7 +1287,9 @@ class DeferredList(Deferred[object]):
             )
             index = index + 1
 
-    def _cbDeferred(self, result: _T, index: int, succeeded: bool) -> Optional[_T]:
+    def _cbDeferred(
+        self, result: _DeferredResultT, index: int, succeeded: bool
+    ) -> Optional[_DeferredResultT]:
         """
         (internal) Callback for when one of my deferreds fires.
         """
@@ -1260,12 +1298,16 @@ class DeferredList(Deferred[object]):
         self.finishedCount += 1
         if not self.called:
             if succeeded == SUCCESS and self.fireOnOneCallback:
-                self.callback((result, index))
+                self.callback((result, index))  # type: ignore[arg-type]
             elif succeeded == FAILURE and self.fireOnOneErrback:
                 assert isinstance(result, Failure)
                 self.errback(Failure(FirstError(result, index)))
             elif self.finishedCount == len(self.resultList):
-                self.callback(self.resultList)
+                # All the None values have been replaced by results, so we can
+                # safely cast away the Optional[].
+                self.callback(
+                    cast(List[Tuple[bool, _DeferredResultT]], self.resultList)
+                )
 
         if succeeded == FAILURE and self.consumeErrors:
             return None
@@ -1293,8 +1335,8 @@ class DeferredList(Deferred[object]):
 
 
 def _parseDeferredListResult(
-    resultList: List[_DeferredListResult], fireOnOneErrback: bool = False
-) -> List[object]:
+    resultList: List[Tuple[bool, _T]], fireOnOneErrback: bool = False
+) -> List[_T]:
     if __debug__:
         for success, value in resultList:
             assert success
@@ -1302,8 +1344,8 @@ def _parseDeferredListResult(
 
 
 def gatherResults(
-    deferredList: List[Deferred[object]], consumeErrors: bool = False
-) -> Deferred[Any]:
+    deferredList: Iterable[Deferred[_T]], consumeErrors: bool = False
+) -> Deferred[List[_T]]:
     """
     Returns, via a L{Deferred}, a list with the results of the given
     L{Deferred}s - in effect, a "join" of multiple deferred operations.
@@ -1317,8 +1359,6 @@ def gatherResults(
     This differs from L{DeferredList} in that you don't need to parse
     the result for success/failure.
 
-    @type deferredList:  L{list} of L{Deferred}s
-
     @param consumeErrors: (keyword param) a flag, defaulting to False,
         indicating that failures in any of the given L{Deferred}s should not be
         propagated to errbacks added to the individual L{Deferred}s after this
@@ -1329,7 +1369,7 @@ def gatherResults(
     """
     d = DeferredList(deferredList, fireOnOneErrback=True, consumeErrors=consumeErrors)
     d.addCallback(_parseDeferredListResult)
-    return d
+    return cast(Deferred[List[_T]], d)
 
 
 # Constants for use with DeferredList
--- a/src/twisted/web/_element.py
+++ b/src/twisted/web/_element.py
@@ -3,7 +3,7 @@
 # See LICENSE for details.
 
 
-from typing import TYPE_CHECKING, Callable, Optional, TypeVar, Union, overload
+from typing import TYPE_CHECKING, Callable, List, Optional, TypeVar, Union, overload
 from zope.interface import implementer
 
 from twisted.web.iweb import IRenderable, IRequest, ITemplateLoader
@@ -55,7 +55,7 @@ class Expose:
         if not funcObjs:
             raise TypeError("expose() takes at least 1 argument (0 given)")
         for fObj in funcObjs:
-            exposedThrough = getattr(fObj, "exposedThrough", [])
+            exposedThrough: List[Expose] = getattr(fObj, "exposedThrough", [])
             exposedThrough.append(self)
             setattr(fObj, "exposedThrough", exposedThrough)
         return funcObjs[0]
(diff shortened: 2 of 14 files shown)


src/twisted/internet/defer.py
```diff
@@ -959,7 +962,11 @@ def send(self, value: object = None) -> "Deferred[_DeferredResultT]":
     # type note: base class "Awaitable" defined the type as:
     #     Callable[[], Generator[Any, None, _DeferredResultT]]
     #     See: https://github.com/python/typeshed/issues/5125
```
wsanchez, line 964: Calling out https://github.com/python/typeshed/issues/5125 in this PR

src/twisted/internet/defer.py
```diff
@@ -1159,10 +1166,43 @@ def __cmp__(self, other: object) -> int:
         return -1
 
 
-_DeferredListResult = Tuple[bool, object]
+if TYPE_CHECKING:
 
+    # The result type is different depending on whether fireOnOneCallback
+    # is True or False.  The type system is not flexible enough to handle
+    # that in a class definition, so instead we pretend that DeferredList
+    # is a function that returns a Deferred.
```
wsanchez, line 1174: OK.  Shouldn't there be an `else` before the class def? I guess the `type: ignore` deal with that?

src/twisted/internet/defer.py
```diff
@@ -1222,12 +1262,12 @@ def __init__(
             logged.  This does not prevent C{fireOnOneErrback} from working.
         """
         self._deferredList = list(deferredList)
-        self.resultList: List[Optional[_DeferredListResult]] = [None] * len(
+        self.resultList: List[Optional[Tuple[bool, _DeferredResultT]]] = [None] * len(
```
wsanchez, line 1266: A comment here explaining why `bool` is in here would be helpful, I think.

src/twisted/web/_element.py
```diff
@@ -19,15 +22,9 @@ class Expose:
     Instances of this class can be called with one or more functions as
     positional arguments.  The names of these functions will be added to a list
     on the class object of which they are methods.
-
-    @ivar attributeName: The attribute with which exposed methods will be
-    tracked.
     """
 
-    def __init__(self, doc=None):
-        self.doc = doc
```
wsanchez, line 28: Is this a public class? That is, is this perhaps a compat issue?

src/twisted/web/_element.py
```diff
@@ -58,13 +55,24 @@ def frob(self, a, b):
         if not funcObjs:
             raise TypeError("expose() takes at least 1 argument (0 given)")
         for fObj in funcObjs:
-            fObj.exposedThrough = getattr(fObj, "exposedThrough", [])
-            fObj.exposedThrough.append(self)
+            exposedThrough = getattr(fObj, "exposedThrough", [])
+            exposedThrough.append(self)
+            setattr(fObj, "exposedThrough", exposedThrough)
```
wsanchez, line 63: If this is to avoid a "no such attribute error" from Mypy, I wonder whether we prefer a `type: ignore` (which notes we are intentionally overriding the type system) is better. @glyph ^^

My thinking is that `type: ignore` invites occasional scrutiny in the future, and maybe that's worthwhile?

src/twisted/web/_element.py
```diff
@@ -181,3 +181,7 @@ def render(self, request):
         if loader is None:
             raise MissingTemplateLoader(self)
         return loader.load()
+
+
+if TYPE_CHECKING:
+    from twisted.web.template import Flattenable, Tag
```
wsanchez: Curious why not move this up so you don't need to put quotes around them in the file.

src/twisted/web/_flatten.py
```diff
@@ -12,29 +12,72 @@
...
+
+
+T = TypeVar("T")
 
+FlattenableRec = Any
+"""
+For documentation purposes, read C{FlattenableRec} as L{Flattenable}.
+However, since mypy doesn't support recursive type definitions (yet?),
+we'll put Any in the actual definition.
+"""
```
wsanchez: Yeah I think JSON has this problem also.

wsanchez: What's `Rec` stand for?

src/twisted/web/_stan.py
```diff
@@ -210,13 +214,13 @@ def _clone(self, obj, deep):
         @return: a clone of C{obj}.
         """
         if hasattr(obj, "clone"):
-            return obj.clone(deep)
+            return obj.clone(deep)  # type: ignore[attr-defined, no-any-return]
```
wsanchez: We probably should `cast` instead of `no-any-return`, I think.

src/twisted/web/_stan.py
```diff
@@ -210,13 +214,13 @@ def _clone(self, obj, deep):
...
         elif isinstance(obj, (list, tuple)):
-            return [self._clone(x, deep) for x in obj]
+            return [self._clone(x, deep) for x in obj]  # type: ignore[return-value]
```
wsanchez: `cast` here also, I think

src/twisted/web/_stan.py
```diff
@@ -288,54 +292,51 @@ def __repr__(self) -> str:
...
-        self.ordinal = ordinal
+    ordinal: int
+    """The ordinal value of the unicode character to which this object refers."""
 
     def __repr__(self) -> str:
         return "CharRef(%d)" % (self.ordinal,)
+
+
+if TYPE_CHECKING:
+    from twisted.web.template import Flattenable
```
wsanchez: Why not import at the top?

src/twisted/web/iweb.py
```diff
@@ -825,3 +825,7 @@ def endpointForURI(uri):
     "IClientRequest",
     "UNKNOWN_LENGTH",
 ]
+
+
+if TYPE_CHECKING:
+    from twisted.web.template import Flattenable, Tag
```
wsanchez: Top?

src/twisted/web/test/_util.py
```diff
@@ -55,9 +62,9 @@ def assertFlattensImmediately(self, root, target):
             L{target}.
         @rtype: L{bytes}
         """
-        return self.successResultOf(self.assertFlattensTo(root, target))
+        return self.successResultOf(self.assertFlattensTo(root, target))  # type: ignore[no-any-return]
```
wsanchez, line 65: `cast` instead

src/twisted/web/template.py
```diff
@@ -109,40 +126,43 @@ class _ToStan(handler.ContentHandler, handler.EntityResolver):
...
-    def startDocument(self):
+    def startDocument(self) -> None:
         """
         Initialise the document.
         """
-        self.document = []
+        # Depending on our active context, the element type can be Tag, slot
+        # or str. Since mypy doesn't understand that context, it would be
+        # a pain to not use Any here.
+        self.document: List[Any] = []
```
wsanchez, line 150: "Leave as future exercise" seems totally reasonable in this PR

src/twisted/web/test/test_flatten.py
```diff
@@ -237,13 +258,13 @@ def test_serializedAttributeWithTagWithAttribute(self):
...
 
-    def test_serializeComment(self):
+    def test_serializeComment(self) -> None:
         """
         Test that comments are correctly flattened and escaped.
         """
         self.assertFlattensImmediately(Comment("foo bar"), b"<!--foo bar-->")
 
-    def test_commentEscaping(self):
+    def test_commentEscaping(self) -> Deferred[Any]:
```
wsanchez: `Deferred[object]` perhaps.

src/twisted/web/test/test_flatten.py
```diff
@@ -442,15 +473,15 @@ class FlattenerErrorTests(SynchronousTestCase):
...
+    def test_renderable(self) -> None:
         """
         If a L{FlattenerError} is created with an L{IRenderable} provider root,
         the repr of that object is included in the string representation of the
         exception.
         """
 
         @implementer(IRenderable)
-        class Renderable:
+        class Renderable:  # type: ignore[misc]
```
wsanchez, line 484: What's the `misc` error here?

wsanchez (approved):
Looks great, thanks for the work.
A few questions and nits, but nothing I'd block on.

src/twisted/web/_element.py
glyph: Just my 2¢ here, I'd rather not see type annotation PRs ever introduce additional runtime coupling if they can possibly avoid it. When we adopt 3.7 we can drop all the quotes with `from __future__ import annotations`

glyph: (In the meanwhile they seem like a small price to pay to avoid having to think hard about circular dependencies or import times)

src/twisted/web/_flatten.py
glyph: Presumably `Recursive` ?

src/twisted/web/template.py
glyph: I am not strongly recommending this, just mentioning it, but something that can sometimes be useful in this type of situation is adding a top-level type alias and commenting there, rather than at the call site; so like at module scope doing `TagSoupTerminal = Any` (or a better / more specific name if you can think of one) and then eventually replace it with `TagSoupTerminal = Union[Tag, slot, str]` (or similar) without changing the implementation logic

src/twisted/internet/defer.py
mthuurne: I picked `type: ignore` over `else` because this makes mypy check the implementation of `DeferredList`.

src/twisted/internet/defer.py
mthuurne: I'll add a docstring for `resultList`.

src/twisted/web/_element.py
mthuurne: `Expose` is in a private module and not re-exported in any public module, so the class itself is private API.

src/twisted/web/_element.py
mthuurne: I was going to argue that `getattr()` allows type checking of the attribute, but I see that I didn't actually annotate `exposedThrough`. I think that's the cleaner solution though: `type: ignore` will make the warning go away but will also prevent mypy from checking the code.

I'll add an annotation of `exposedThrough`. If it is later decided that `type: ignore` is preferred, that's fine too.

src/twisted/web/_element.py
mthuurne: Circular dependencies are indeed a potential issue here: `template` imports `_element`, but it does so at the end of the module, so it would probably be fine but it would be brittle.

Therefore I'd rather leave the `TYPE_CHECKING` in place. ~I could move the statement to the top of the file, if you think that makes it easier to read.~ I have moved the import to the top of the module.

src/twisted/web/_flatten.py
mthuurne: > Yeah I think JSON has this problem also.

Every tree structure has this problem.

I once tried to do some tricky workaround and that crashed mypy. The crash was fixed quickly, but there is no solution for recursive types yet.

> What's `Rec` stand for?

> Presumably `Recursive` ?

Yes, recursive. I'll write it out in full to make it clearer.

src/twisted/web/_stan.py
mthuurne: Agreed.

src/twisted/web/_stan.py
mthuurne: Hmm, strictly speaking we don't preserve the original type here: if a tuple is passed, we return a list instead. Is it acceptable to paper over this detail? Or would it be better to tone down the claim that `_clone()` can clone an arbitrary object and instead annotate it as taking and returning `Flattenable`?

I have a preference for the latter, since `_clone()` is made for tag contents and cannot actually clone arbitrary Python objects. For example if you'd tell it to make a deep clone of a JSON object, it would make a shallow copy of the dictionaries. I'll implement this option so it can be reviewed.

mthuurne: Iterators are considered flattenable. Currently they are just copied, but they're not immutable. Should they be cloned into lists instead? We'd have to modify the original as well though, since we'd empty the iterator while cloning it, which might not be what the user expects.

Coroutines are also flattenable types, but I don't think we could clone those in any meaningful way. Should we just copy them as-is or raise an exception? Maybe raising an exception is the right solution for iterators too.

src/twisted/web/_stan.py
mthuurne: I think originally I didn't have the `TYPE_CHECKING` there and then importing at the bottom made sense to reduce the chance of cyclic import problems. I'll move it.

src/twisted/web/test/_util.py
mthuurne: I don't think I agree in this case. What we're working around here is not a limitation of mypy or the type system, but the fact that `successResultOf()` isn't annotated yet. I assume that will change at some point and having a `type: ignore` comment means that mypy will flag the suppression as a warning as soon as it can be removed. If we'd use a `cast`, we wouldn't be notified when the workaround becomes obsolete.

src/twisted/web/template.py
mthuurne: > "Leave as future exercise" seems totally reasonable in this PR

I think this is a case where annotations alone aren't going to be sufficient; it will probably require some restructuring of the parser to make it more mypy-friendly.

> I am not strongly recommending this, just mentioning it, but something that can sometimes be useful in this type of situation is adding a top-level type alias and commenting there, rather than at the call site; so like at module scope doing `TagSoupTerminal = Any` (or a better / more specific name if you can think of one) and then eventually replace it with `TagSoupTerminal = Union[Tag, slot, str]` (or similar) without changing the implementation logic

That sounds like a good strategy in other places, but here it wouldn't save much, since there are only two annotations with `Any` and they're both in the block directly below this comment. If this were a public class, there might be some value in assigning a name for documentation purposes (*), but it's a private class instead.

(*) I think support for documented aliases should be added to pydoctor. That's useful for type aliases in particular; for other objects aliases are typically used to export an object under a different name and in that case the documentation of the original can be moved.

src/twisted/web/test/test_flatten.py
mthuurne: `Deferred[object]` is actually not accepted by mypy. But in the error message, it tells me the actual type is `Deferred[List[bytes]]`, so I'll use that.

src/twisted/web/test/test_flatten.py
mthuurne:
```
src/twisted/web/test/test_flatten.py:484:9: error: 'Renderable' is missing following 'IRenderable' interface members: lookupRenderMethod, render.  [misc]
```

mthuurne:
I think the only issue left is how to deal with `Tag._clone()`:
- do you agree with the change from arbitrary type to flattenable?
- should we raise an exception for flattenable types that are neither cloneable nor immutable, such as iterators and coroutines?

Regarding the latter, for new code I'd be in favor of rejecting input we cannot handle properly. But I'm worried that there is code out there that clones a tag and then only uses the clone, in that case we'd break a working program if we'd reject the cloning of iterators and coroutines.

src/twisted/internet/defer.py
wsanchez: Aha, cool. I didn't know that was even an option.

src/twisted/web/_flatten.py
wsanchez: Cool.  I'm glad that you get some non-nested type checking here as a baseline.

src/twisted/web/_stan.py
wsanchez: Oof. Both sound problematic.  Since you can only drain an iterator once, if you copy it, one of the copies is going to get a different (empty) result when flattened, right?

wsanchez: I vote for exceptions.

src/twisted/web/test/_util.py
wsanchez: Ah, ok

wsanchez: An unnecessary `cast` will also cause complaint later, which is what I was angling for.  But I think you are right and both will.

src/twisted/web/test/test_flatten.py
wsanchez: I'm a little surprised about that, but ok

wsanchez:
> I think the only issue left is how to deal with `Tag._clone()`:
> 
> * do you agree with the change from arbitrary type to flattenable?
> * should we raise an exception for flattenable types that are neither cloneable nor immutable, such as iterators and coroutines?

I think so.  And I think we can take some latitude here since it's private; we can fix it if it upsets something else in Twisted, but that seems a non-issue.

> Regarding the latter, for new code I'd be in favor of rejecting input we cannot handle properly. But I'm worried that there is code out there that clones a tag and then only uses the clone, in that case we'd break a working program if we'd reject the cloning of iterators and coroutines.

Hrm.  But that's *so* dangerous…  If not an exception, then maybe emitting a warning, so be replaced with an exception later, like a deprecation.

src/twisted/web/template.py
wsanchez: Agreed with both ;-)

mthuurne:
> > I think the only issue left is how to deal with `Tag._clone()`:
> > 
> > * do you agree with the change from arbitrary type to flattenable?
> > * should we raise an exception for flattenable types that are neither cloneable nor immutable, such as iterators and coroutines?
> 
> I think so. And I think we can take some latitude here since it's private; we can fix it if it upsets something else in Twisted, but that seems a non-issue.

It's not as private as its name suggests, since it's called by the public `clone()` method to do a lot of the actual cloning.

> > Regarding the latter, for new code I'd be in favor of rejecting input we cannot handle properly. But I'm worried that there is code out there that clones a tag and then only uses the clone, in that case we'd break a working program if we'd reject the cloning of iterators and coroutines.
> 
> Hrm. But that's _so_ dangerous… If not an exception, then maybe emitting a warning, so be replaced with an exception later, like a deprecation.

I like the idea of a deprecation: it won't break applications but it will alert developers that they're doing something dangerous.

mthuurne:
The coverage regression is in `DNSDatagramProtocol.query()`, which should not be impacted at all by this PR. So I'm assuming that method is covered by a flaky tests.

wsanchez:
@mthuurne looks good, thanks!

PR #1597, opened 2021-05-11
10194 Add type hints for twisted.test.test_defer

wsanchez:
Add type hints for `twisted.test.test_defer`.

The main purpose of this is the better test the type hints in `twisted.internet.defer`, and a few small changes resulted there from this PR.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10194

```

(as long as you wish)
```

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1327,8 +1327,6 @@ class DeferredList(Deferred[_DeferredListResultListT]):  # type: ignore[no-redef
                 assert isinstance(result, Failure)
                 self.errback(Failure(FirstError(result, index)))
             elif self.finishedCount == len(self.resultList):
-                # All the None values have been replaced by results, so we can
-                # safely cast away the Optional[].
                 self.callback(self.resultList)
 
         if succeeded == FAILURE and self.consumeErrors:
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1159,16 +1159,18 @@ class FirstError(Exception):
         return -1
 
 
-# This wants to be TypeVar, but is set to Any for now, since we can't type
-# DeferredList's superclass the way we'd like to (see below), which means a
-# TypeVar would be unbound (which is an error).
+# _DeferredListResultT wants to be TypeVar, but is set to Any for now, since we
+# can't type DeferredList's superclass the way we'd like to (see below), which
+# means a TypeVar would be unbound (which is an error).
 _DeferredListResultT = Any
+_DeferredListResultItemT = Tuple[bool, _DeferredListResultT]
+_DeferredListResultListT = List[_DeferredListResultItemT]
 
 # type note for DeferredList:
 #
 # If fireOnOneCallback is False, this wants to be:
 #
-#   class DeferredList(Deferred[List[Tuple[bool, _DeferredListResultT]]]):
+#   class DeferredList(Deferred[_DeferredListResultListT]):
 #
 # ...but if fireOnOneCallback is True, it wants to be:
 #
@@ -1238,8 +1240,8 @@ class DeferredList(Deferred[Any]):
             logged.  This does not prevent C{fireOnOneErrback} from working.
         """
         self._deferredList = list(deferredList)
-        self.resultList: List[Tuple[bool, _DeferredListResultT]] = [
-            cast(Tuple[bool, _DeferredListResultT], None)
+        self.resultList: _DeferredListResultListT = [
+            cast(_DeferredListResultItemT, None)
         ] * len(self._deferredList)
         Deferred.__init__(self)
         if len(self._deferredList) == 0 and not fireOnOneCallback:
@@ -1307,7 +1309,7 @@ class DeferredList(Deferred[Any]):
 
 
 def _parseDeferredListResult(
-    resultList: List[Tuple[bool, _DeferredListResultT]], fireOnOneErrback: bool = False
+    resultList: _DeferredListResultListT, fireOnOneErrback: bool = False
 ) -> List[_DeferredListResultT]:
     if __debug__:
         for success, value in resultList:
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1180,7 +1180,7 @@ class FirstError(Exception):
 
 _DeferredListSingleResultT = Tuple[_DeferredResultT, int]
 _DeferredListResultItemT = Tuple[bool, _DeferredResultT]
-_DeferredListResultListT = List[Optional[_DeferredListResultItemT]]
+_DeferredListResultListT = List[_DeferredListResultItemT]
 
 if TYPE_CHECKING:
 
@@ -1281,7 +1281,12 @@ class DeferredList(Deferred[_DeferredListResultListT]):  # type: ignore[no-redef
         """
         self._deferredList = list(deferredList)
 
-        self.resultList: _DeferredListResultListT = [None] * len(self._deferredList)
+        # Note this contains optional result values as the DeferredList is
+        # processing its results, even though the callback result will not,
+        # which is why we aren't using _DeferredListResultListT here.
+        self.resultList: List[Optional[_DeferredListResultItemT]] = [None] * len(
+            self._deferredList
+        )
         """
         The final result, in progress.
         Each item in the list corresponds to the L{Deferred} at the same
@@ -1327,7 +1332,10 @@ class DeferredList(Deferred[_DeferredListResultListT]):  # type: ignore[no-redef
                 assert isinstance(result, Failure)
                 self.errback(Failure(FirstError(result, index)))
             elif self.finishedCount == len(self.resultList):
-                self.callback(self.resultList)
+                # At this point, None values in self.resultList have been
+                # replaced by result values, so we cast it to
+                # _DeferredListResultListT to match the callback result type.
+                self.callback(cast(_DeferredListResultListT, self.resultList))
 
         if succeeded == FAILURE and self.consumeErrors:
             return None
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1820,8 +1820,8 @@ class _InternalInlineCallbacksCancelledError(Exception):
 # type note: "..." is used here because we don't have a better way to express
 #     that the same arguments are accepted by the returned callable.
 def inlineCallbacks(
-    f: Callable[..., Generator[Any, Any, None]]
-) -> Callable[..., Deferred[Any]]:
+    f: Callable[..., Generator[Deferred[object], object, None]]
+) -> Callable[..., Deferred[object]]:
     """
     L{inlineCallbacks} helps you write L{Deferred}-using code that looks like a
     regular sequential function. For example::
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1058,7 +1058,11 @@ class Deferred(Awaitable[_DeferredResultT]):
 
     @classmethod
     def fromCoroutine(
-        cls, coro: Coroutine["Deferred[_T]", object, _T]
+        cls,
+        coro: Union[
+            Coroutine["Deferred[_T]", Any, _T],
+            Generator["Deferred[_T]", Any, _T],
+        ],
     ) -> "Deferred[_T]":
         """
         Schedule the execution of a coroutine that awaits on L{Deferred}s,
@@ -1095,14 +1099,19 @@ class Deferred(Awaitable[_DeferredResultT]):
 
         @raise ValueError: If C{coro} is not a coroutine or generator.
         """
-        if not iscoroutine(coro) and not isinstance(coro, GeneratorType):
+        # type note: Subclass of "Generator[Deferred[_T], object, _T]" and "GeneratorType" cannot exist
+        if not iscoroutine(coro) and not isinstance(coro, GeneratorType):  # type: ignore[unreachable]
             raise NotACoroutineError(f"{coro!r} is not a coroutine")
 
         return _cancellableInlineCallbacks(coro)
 
 
 def ensureDeferred(
-    coro: Union[Coroutine["Deferred[_T]", object, "_T"], Deferred[_T]]
+    coro: Union[
+        Coroutine[Deferred[_T], Any, _T],
+        Generator[Deferred[_T], Any, _T],
+        Deferred[_T],
+    ]
 ) -> Deferred[_T]:
     """
     Schedule the execution of a coroutine that awaits/yields from L{Deferred}s,
@@ -1489,7 +1498,7 @@ def _deferGenerator(
             result = None
 
 
-@deprecated(Version("Twisted", 15, 0, 0), "twisted.internet.defer.inlineCallbacks")
+@deprecated(Version("Twisted", 15, 0, 0), "twisted.internet.defer.x")
 def deferredGenerator(
     f: Callable[..., _DeferableGenerator]
 ) -> Callable[..., Deferred[object]]:
@@ -1813,8 +1822,8 @@ class _InternalInlineCallbacksCancelledError(Exception):
 # type note: "..." is used here because we don't have a better way to express
 #     that the same arguments are accepted by the returned callable.
 def inlineCallbacks(
-    f: Callable[..., Generator[Deferred[object], object, None]]
-) -> Callable[..., Deferred[object]]:
+    f: Callable[..., Generator[Any, Any, None]]
+) -> Callable[..., Deferred[Any]]:
     """
     L{inlineCallbacks} helps you write L{Deferred}-using code that looks like a
     regular sequential function. For example::
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1180,7 +1180,7 @@ class FirstError(Exception):
 
 _DeferredListSingleResultT = Tuple[_DeferredResultT, int]
 _DeferredListResultItemT = Tuple[bool, _DeferredResultT]
-_DeferredListResultListT = List[_DeferredListResultItemT]
+_DeferredListResultListT = List[Optional[_DeferredListResultItemT]]
 
 if TYPE_CHECKING:
 
@@ -1281,9 +1281,7 @@ class DeferredList(Deferred[_DeferredListResultListT]):  # type: ignore[no-redef
         """
         self._deferredList = list(deferredList)
 
-        self.resultList: _DeferredListResultListT = [
-            cast(_DeferredListResultItemT, None)
-        ] * len(self._deferredList)
+        self.resultList: _DeferredListResultListT = [None] * len(self._deferredList)
         """
         The final result, in progress.
         Each item in the list corresponds to the L{Deferred} at the same
@@ -1359,10 +1357,12 @@ class DeferredList(Deferred[_DeferredListResultListT]):  # type: ignore[no-redef
 
 
 def _parseDeferredListResult(
-    resultList: _DeferredListResultListT, fireOnOneErrback: bool = False
+    resultList: List[_DeferredListResultItemT], fireOnOneErrback: bool = False
 ) -> List[_T]:
     if __debug__:
-        for success, value in resultList:
+        for result in resultList:
+            assert result is not None
+            success, value = result
             assert success
     return [x[1] for x in resultList]
 
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1190,7 +1190,7 @@ class DeferredList(Deferred[object]):
 
     def __init__(
         self,
-        deferredList: List[Deferred[object]],
+        deferredList: List[Deferred[Any]],
         fireOnOneCallback: bool = False,
         fireOnOneErrback: bool = False,
         consumeErrors: bool = False,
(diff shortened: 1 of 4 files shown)


src/twisted/internet/defer.py
```diff
@@ -1058,7 +1058,11 @@ def uncancel(
 
     @classmethod
     def fromCoroutine(
-        cls, coro: Coroutine["Deferred[_T]", object, _T]
+        cls,
+        coro: Union[
+            Coroutine["Deferred[_T]", Any, _T],
+            Generator["Deferred[_T]", Any, _T],
+        ],
```
wsanchez, line 1065: Generators are also accepted.

src/twisted/internet/defer.py
```diff
@@ -1095,14 +1099,19 @@ def main(reactor):
 
         @raise ValueError: If C{coro} is not a coroutine or generator.
         """
-        if not iscoroutine(coro) and not isinstance(coro, GeneratorType):
+        # type note: Subclass of "Generator[Deferred[_T], object, _T]" and "GeneratorType" cannot exist
+        if not iscoroutine(coro) and not isinstance(coro, GeneratorType):  # type: ignore[unreachable]
```
wsanchez, line 1103: I think this is an oddity of `GeneratorType` and `Generator` not being the same.

src/twisted/internet/defer.py
```diff
@@ -1095,14 +1099,19 @@ def main(reactor):
...
         return _cancellableInlineCallbacks(coro)
 
 
 def ensureDeferred(
-    coro: Union[Coroutine["Deferred[_T]", object, "_T"], Deferred[_T]]
+    coro: Union[
+        Coroutine[Deferred[_T], Any, _T],
+        Generator[Deferred[_T], Any, _T],
+        Deferred[_T],
+    ]
```
wsanchez, line 1114: Generators are also accepted.

src/twisted/internet/defer.py
```diff
@@ -1266,9 +1281,9 @@ def __init__(
         """
         self._deferredList = list(deferredList)
 
-        self.resultList: List[Optional[Tuple[bool, _DeferredResultT]]] = [None] * len(
-            self._deferredList
-        )
+        self.resultList: _DeferredListResultListT = [
+            cast(_DeferredListResultItemT, None)
```
wsanchez: Casting to `None` here instead of typing `self.resultList` as optional because it's only allowed to be set to `None` here and all of the `None` values should be replaced by `_DeferredListResultItemT` values by the time the `DeferredList` fires, so we don't want the client to have to deal with the optional that they never see.

src/twisted/test/test_defer.py
```diff
@@ -676,119 +729,134 @@ def test_cancelGatherResults(self):
...
         """
-        S, E = [], []
-        d = defer.maybeDeferred((lambda x: x + 5), 10)
-        d.addCallbacks(S.append, E.append)
-        self.assertEqual(E, [])
-        self.assertEqual(S, [15])
 
-    def test_maybeDeferredSyncException(self):
+        def plusFive(x: int) -> int:
+            return x + 5
```
wsanchez, line 779: Was a lambda, change to function so it can be typed. More like this below.

src/twisted/test/test_defer.py
```diff
@@ -676,119 +729,134 @@ def test_cancelGatherResults(self):
...
             expected = str(e)
-        d = defer.maybeDeferred((lambda x: x + 5), "10")
-        d.addCallbacks(S.append, E.append)
-        self.assertEqual(S, [])
-        self.assertEqual(len(E), 1)
-        self.assertEqual(str(E[0].value), expected)
 
-    def test_maybeDeferredSyncFailure(self):
+        def plusFive(x: int) -> int:
+            return x + 5
```
wsanchez, line 787: Was a lambda

src/twisted/test/test_defer.py
```diff
@@ -1005,23 +1073,27 @@ def test_asynchronousImplicitChain(self):
...
         callback on a second L{Deferred}, the first L{Deferred}'s result is
         converted to L{None} and no unhandled error is logged when it is
         garbage collected.
         """
         first = defer.fail(RuntimeError("First Deferred's Failure"))
-        second = defer.Deferred()
-        second.addCallback(lambda ign, first=first: first)
+
+        def cb(_: None, first: Deferred[None] = first) -> Deferred[None]:
+            return first
```
wsanchez, line 1086: Was a lambda

src/twisted/test/test_defer.py
```diff
@@ -1142,34 +1214,38 @@ def secondCallback(result):
...
+                results.append(("firstCallback", result))
                 # Return a Deferred that definitely has not fired yet with a
                 # result-transforming callback so we can fire the Deferreds
                 # out of order and see how the callback affects the ultimate
                 # results.
-                return inner.addCallback(lambda x: [x])
 
-            def secondCallback(result):
+                def transform(result: str) -> List[str]:
+                    return [result]
```
wsanchez, line 1244: Was a lambda

src/twisted/test/test_defer.py
```diff
@@ -1268,83 +1344,103 @@ def circularCallback(result):
...
-            last = defer.Deferred()
+            last: Deferred[None] = Deferred()
 
             inner = innerDeferreds.pop()
-            top.addCallback(lambda ign, inner=inner: inner)
+
+            def cbInner(
+                ignored: object, inner: Deferred[None] = inner
+            ) -> Deferred[None]:
+                return inner
```
wsanchez, line 1428: Was a lambda

src/twisted/test/test_defer.py
```diff
@@ -1268,83 +1344,103 @@ def circularCallback(result):
...
             top.addCallback(recordStackDepth)
 
             while innerDeferreds:
                 newInner = innerDeferreds.pop()
-                inner.addCallback(lambda ign, inner=newInner: inner)
+
+                def cbNewInner(
+                    ignored: object, inner: Deferred[None] = newInner
+                ) -> Deferred[None]:
+                    return inner
```
wsanchez, line 1439: Was a lambda

src/twisted/test/test_defer.py
```diff
@@ -2191,77 +2297,88 @@ def testLock(self):
         firstUnique = object()
         secondUnique = object()
 
-        controlDeferred = defer.Deferred()
+        controlDeferred: Deferred[object] = Deferred()
 
-        def helper(self, b):
-            self.b = b
+        result: Optional[object] = None
```
wsanchez, line 2302: Use a variable (`nonlocal` in `helper()`) instead of setting an attribute on `self`.

src/twisted/test/test_defer.py
```diff
@@ -2191,77 +2297,86 @@ def testLock(self):
...
 
-        resultDeferred = lock.run(helper, self=self, b=firstUnique)
+        def helper(resultValue: object, returnValue: object = None) -> object:
+            nonlocal result
+            result = resultValue
+            return returnValue
+
+        resultDeferred = lock.run(
+            helper, resultValue=firstUnique, returnValue=controlDeferred
+        )
```
wsanchez: Use a variable (`nonlocal` in `helper()`) instead of setting an attribute on `self`.  More use of `nonlocal` below, to either avoid setting attributes on objects or instead of using one-element lists.

src/twisted/test/test_defer.py
```diff
@@ -2541,202 +2656,210 @@ class DeferredAddTimeoutTests(unittest.SynchronousTestCase):
...
-        innerDeferred = defer.Deferred()
-        dCallbacked = [None]
+        innerDeferred: Deferred[None] = Deferred()
+        dCallbacked: Optional[str] = None
 
-        def onCallback(results):
-            dCallbacked[0] = results
+        def onCallback(results: str) -> Deferred[None]:
+            nonlocal dCallbacked
+            dCallbacked = results
```
wsanchez, line 2721: Use `nonlocal`

src/twisted/test/test_defer.py
```diff
@@ -2541,202 +2656,210 @@ class DeferredAddTimeoutTests(unittest.SynchronousTestCase):
...
-        dErrbacked = [None]
+        innerDeferred: Deferred[None] = Deferred()
+        dErrbacked: Optional[Failure] = None
         error = ValueError("fail")
 
-        def onErrback(f):
-            dErrbacked[0] = f
+        def onErrback(f: Failure) -> Deferred[None]:
+            nonlocal dErrbacked
+            dErrbacked = f
```
wsanchez, line 2755: Use `nonlocal`

src/twisted/test/test_defer.py
```diff
@@ -2541,202 +2656,210 @@ class DeferredAddTimeoutTests(unittest.SynchronousTestCase):
...
```
wsanchez, line 2791: Use `nonlocal`

src/twisted/test/test_defer.py
```diff
@@ -2745,190 +2868,202 @@ def test_timedOutProvidedCancelFailure(self):
...
-        innerDeferred = defer.Deferred()
-        dCanceled = [None]
+        innerDeferred: Deferred[None] = Deferred()
+        dCanceled = None
 
-        def onErrback(f):
-            dCanceled[0] = f
+        def onErrback(f: Failure) -> Deferred[None]:
+            nonlocal dCanceled
+            dCanceled = f
```
wsanchez, line 2889: Use `nonlocal`

src/twisted/test/test_defer.py
```diff
@@ -2745,190 +2868,202 @@ def test_timedOutProvidedCancelFailure(self):
...
```
wsanchez, line 2924: Use `nonlocal`

src/twisted/internet/defer.py
```diff
@@ -1809,8 +1822,8 @@ class _InternalInlineCallbacksCancelledError(Exception):
 # type note: "..." is used here because we don't have a better way to express
 #     that the same arguments are accepted by the returned callable.
 def inlineCallbacks(
-    f: Callable[..., Generator[Deferred[object], object, None]]
-) -> Callable[..., Deferred[object]]:
+    f: Callable[..., Generator[Any, Any, None]]
+) -> Callable[..., Deferred[Any]]:
```
wsanchez: `object` ➜ `Any` because otherwise Mypy will complain about decorated functions take return a generator with a `None` send-type, or return a `Deferred[None]`, for example.

src/twisted/internet/defer.py
```diff
@@ -1169,6 +1178,10 @@ def __cmp__(self, other: object) -> int:
         return -1
 
 
+_DeferredListSingleResultT = Tuple[_DeferredResultT, int]
+_DeferredListResultItemT = Tuple[bool, _DeferredResultT]
+_DeferredListResultListT = List[_DeferredListResultItemT]
+
```
wsanchez, line 1184: These are used below, depending on `fireOnOnceCallback`, which make this API challenging to type well.

@mthuurne you might wanna double check me here, since you took the first pass.

src/twisted/internet/defer.py
mthuurne: There is a partial overlap: at least `Generator[T, None, None]` is accepted as an instance of `GeneratorType`. I found that out while working on Stan's annotations.

src/twisted/internet/defer.py
mthuurne: I'll dive into it when I'm a bit better rested.

src/twisted/internet/defer.py
mthuurne: I hadn't realized that the client might be looking at the `resultList` property rather than using the callback argument. I think the latter would be more common and more in line with Twisted's design, but both are valid.

src/twisted/internet/defer.py
```diff
@@ -1316,9 +1331,7 @@ def _cbDeferred(
             elif self.finishedCount == len(self.resultList):
                 # All the None values have been replaced by results, so we can
                 # safely cast away the Optional[].
```
mthuurne: That comment is obsolete now that you're typecasting on the creation of `resultList` instead of casting it on completion.

src/twisted/internet/defer.py
mthuurne: Maybe add a comment explaining this decision in the code.

src/twisted/internet/defer.py
mthuurne: One thing to consider is whether these type aliases should be private or public. If they're being returned by public methods, they should probably be public.

While pydoctor currently doesn't document type aliases as separate entities, I think it is inevitable that we'll implement that at some point. After all, we do show type annotations in the docs now and the user will want to know what concrete types an alias represents. Also I think it should become possible in the future to have docstrings on aliases. To prepare for that future, it would make sense to make all type aliases that the user might want to learn about public.

Another reason to make type aliases public when returned by public methods is that the user might want to use those same aliases to annotate their own code.

src/twisted/internet/defer.py
mthuurne: I was mistaken: in `twisted.web._flatten._flattenElement()` there is no warning issued because `GeneratorType` is tested in a single check together with other types: `isinstance(root, (tuple, list, GeneratorType))`; if I give it a dedicated `isinstance()` check, it fails with the same error as the one you suppressed here.

There is a similar suppression in `inlineCallbacks()`; it's marked as FIXME there, but I'm not sure it can be fixed. Or maybe it should be fixed in mypy instead of Twisted. Intuitively, I'd expect at least `Generator[Any, Any, Any]` to be accepted as a `GeneratorType`, but it is not. However, intuition doesn't always mean a lot when dealing with complex type systems. There are zero issues reported for mypy that contain the name `GeneratorType`, so it doesn't seem to be a known issue.

src/twisted/internet/defer.py
mthuurne: Looking at the code again, I do think these types should be made public, since they describe the deferred return value of `DeferredList`, which is a public interface.

Since `_DeferredListResultItemT` is used only twice, it could be inlined instead of having a name introduced for it. I think that would make the code easier to read, but it's subjective, so it's also fine with me if you decide against it.

src/twisted/internet/defer.py
```diff
@@ -1178,34 +1191,36 @@ def __cmp__(self, other: object) -> int:
 
     @overload
     def _DeferredList(
-        deferredList: Iterable[Deferred[_T]],
+        deferredList: Iterable[Deferred[_DeferredResultT]],
         fireOnOneCallback: Literal[True],
         fireOnOneErrback: bool = False,
         consumeErrors: bool = False,
-    ) -> Deferred[Tuple[_T, int]]:
+    ) -> Deferred[_DeferredListSingleResultT]:
```
mthuurne, line 1198: The difference between `_T` and `_DeferredResultT` is that the latter is marked as contravariant. Since the default is invariant, this is a more liberal annotation, so I guess that if mypy accepts it for the implementation, it is an improvement for the user.

I must admit that this is at the edge of my understanding, while being complicated even more by the fact that the result type can change across a `Deferred` chain and we're not able to model that aspect in the type system. I wanted to point out that this is more than a readability change and while I think it's an improvement, I don't fully understand the implications.

src/twisted/internet/defer.py
mthuurne:
```suggestion
```

src/twisted/internet/defer.py
mthuurne: Do you have an example of valid code that is rejected by the old signature? Because if I revert this change, mypy doesn't find any errors in Twisted.

src/twisted/test/test_defer.py
```diff
@@ -12,28 +12,55 @@
...
-def ensuringDeferred(f):
+def ensuringDeferred(f: Callable[..., Any]) -> Callable[..., Any]:
     @functools.wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: object, **kwargs: object) -> Deferred[Any]:
         result = f(*args, **kwargs)
-        return defer.ensureDeferred(result)
+        return ensureDeferred(result)
 
     return wrapper
```
mthuurne: This decorator is only applied to coroutines, so it could use `Deferred.fromCoroutine()` instead of `ensureDeferred()`. That simplifies its signature compared to `ensureDeferred()` which can take 3 different types:

```py
def ensuringDeferred(
    f: Callable[..., Coroutine[Deferred[_T], Any, _T]]
) -> Callable[..., Deferred[_T]]:
    @functools.wraps(f)
    def wrapper(*args: object, **kwargs: object) -> Deferred[_T]:
        coro = f(*args, **kwargs)
        return Deferred.fromCoroutine(coro)

    return wrapper
```

It's only used 3 times though and all those coroutines have `None` as the result type, so it's probably not all that important to get precise typing here, unless it could be shared with other test modules later.

src/twisted/test/test_defer.py
```diff
@@ -71,160 +98,190 @@ class ImmediateFailureMixin:
     Add additional assertion methods.
     """
 
-    def assertImmediateFailure(self, deferred, exception):
+    def assertImmediateFailure(
+        self, deferred: Deferred[Any], exception: Type[Exception]
+    ) -> Exception:
```
mthuurne: This could be made more precise using `ExceptionT = TypeVar("ExceptionT", bound=Exception)`. It doesn't really seem to make a difference in practice though, since there is only one case where the returned value is used and in that case `Exception` is passed as the exception type.

mthuurne (approved):
I think this is already an improvement in its current form, but I added some remarks about things that could be refined.

Note that for the test module, I only reviewed the test helpers at the top. I assume that since mypy accepts the rest of the annotations and the tests still pass there should be no problems there.

src/twisted/internet/defer.py
wsanchez: Yeah, I agree it's more common.  Because `resultList` is a public attribute, I think it's more important to type it in a way that makes sense to the client over the implementation.

(I have a feeling it might be better if it weren't public, but I'm not a big `DeferredList` user so I don't think I know enough about that.)

mthuurne: The `DeferredList` class should probably be replaced by two functions, one for each value of `fireOnOneCallback`. But that would have to go through a lengthy deprecation cycle, so even if someone would write those replacement functions today, having `DeferredList` annotated is going to be useful for the coming years.

src/twisted/internet/defer.py
mthuurne: I made a minimal test case from this and reported it to mypy: python/mypy#10501

mthuurne: It turns out to be an issue in typeshed and a fix was already being worked on: python/typeshed#5496

wsanchez: Cool; we'll remove the type hints when we bump to a Mypy with that fix.

src/twisted/internet/defer.py
wsanchez: Agreed.

wsanchez: OK I'm looking at this again, and the `None` values to show to the caller when the child `Deferred` objects have yet to fire, so I'm switching this back to optional, which requires a bit more checking elsewhere.  See commit 02110624550737ef58c34f623b5dd36d02f04479.

wsanchez: @mthuurne Can you sanity check that commit?

src/twisted/internet/defer.py
wsanchez: OK I was on the fence here, but will make them public.

wsanchez: Actually, one reason I hesitated is that these are `TypeVar`s (via _DeferredResultT), and I'm not sure what the convention we want to use for those symbols would be.  Is it `DeferredListResultListT` or `DeferredListResultList`?  I think the former, because `TypeVar`s are different and I think name should reflect it.

@graingert You might have an opinion on this? I think I've seen you use `T_Foo` in examples, but I don't know if that's in use in Twisted. (I don't see it in Twisted proper.)

(We could also defer this to a separate PR.)

src/twisted/internet/defer.py
wsanchez: poof

src/twisted/internet/defer.py
wsanchez: Hrm. I must have undone whatever I did that was triggering this… I'll revert it.

src/twisted/test/test_defer.py
wsanchez: Done, thanks!

src/twisted/test/test_defer.py
wsanchez: Done, thanks!

src/twisted/internet/defer.py
mthuurne: It looks mostly fine, except that the type of the value returned by `DeferredList` should not contain `Optional`, since when the callback fires, all the `None` elements have been replaced. In other words, `resultList` contains a partial result and can therefore contain `None`, but the list returned via the callback is a final result and cannot contain `None`.

wsanchez: OK done in d0d586ae287c6feaacb0c094a3bc9b1673dc890c

src/twisted/internet/defer.py
wsanchez: Since `mypy` will tell you the expended type, and that's largely usable, I'm going to defer this for another PR, as I'm hesitant to add public symbols without more confidence that they are what we want.

PR #1604, opened 2021-05-24
[10204] Include the requested min and max when callling getDHPrime.

adiroiban:
## Scope and purpose
This updates the Conch SSH server factory to also receive the r min and max values requested by the client.

The default implementation stills returns the value only based  on `ideal` but it allows implementing custom factories that might deviate from the "ideal" value requested by the client.

For example a factory can go with max when the ideal request bu the client is too small and max would be good enough. 

It is implemented in a backward compatible way, so that old SSH factory implementation will continue to work. 

## Mypy errors

I am not sure what to do with these mypy errors ... should I ignore them?

I guess that the issue is that `random.choice` is a generic type.

```
src/twisted/conch/ssh/factory.py:116:63: error: Returning Any from function declared to return "SupportsLessThan"  [no-any-return]
            primesKeys = sorted(self.primes.keys(), key=lambda i: abs(i - bits))
                                                                  ^
src/twisted/conch/ssh/factory.py:118:9: error: Returning Any from function declared to return "Tuple[int, int]"  [no-any-return]
            return random.choice(self.primes[realBits])
```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10204

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10204

When calling t.conch.ssh.factory.SSHFactory.getDHPrimes also Include the  min and max values requested by the client.
```


src/twisted/conch/test/test_transport.py
```diff
@@ -1966,6 +1948,31 @@ def test_KEX_DH_GEX_REQUEST(self, keyAlgorithm=b"ssh-rsa"):
         self.assertEqual(self.proto.g, 2)
         self.assertEqual(self.proto.p, dhPrime)
 
+    def test_KEX_DH_GEX_REQUEST_legacyFacotry(self):
```
glyph, line 1951: We shouldn't be adding new tests without type hints.

```suggestion
    def test_KEX_DH_GEX_REQUEST_legacyFacotry(self) -> None:
```

src/twisted/conch/test/test_transport.py
```diff
@@ -310,36 +310,18 @@ def getPrimes(self):
...
+class LegacyGetDHPrimesFactory(MockFactory):
     """
-    The old SSHFactory returned mappings from key names to strings from
-    getPublicKeys().  We return those here for testing.
+    A factory that implements L{twisted.conch.ssh.factory.SSHFactory.getDHPrime}
+    with a single argument.
     """
 
-    def getPublicKeys(self):
+    def getDHPrime(self, ideal):
```
glyph, line 319: It would still be nice to have the type-hint here for completeness:

```suggestion
    def getDHPrime(self, ideal: int) -> Tuple[int, int]:
```

src/twisted/conch/ssh/factory.py
```diff
@@ -102,13 +102,20 @@ def getPrimes(self):
         @rtype: L{dict}
         """
 
-    def getDHPrime(self, bits):
+    def getDHPrime(
+        self, bits: int, minimum: Optional[int] = None, maximum: Optional[int] = None
+    ) -> Tuple[int, int]:
```
glyph, line 108: Love to see all this detail getting fleshed out.

src/twisted/conch/ssh/factory.py
```diff
@@ -102,13 +102,20 @@ def getPrimes(self):
...
+
+        The C{minimum} and C{maximum} parameters are not used in the default implementation.
+
+        @param bits: The ideal size for a group key exchange algorithm or the actual size requested for a non-group key exchange algorithm.
+        @param minimum: The minimum accepted size for group key exchange algorithm. This is C{None} for non-group algorithms.
+        @param maximum: The maximum accepted size for group key exchange algorithm. This is C{None} for non-group algorithms.
 
-        @type bits: L{int}
-        @rtype:     L{tuple}
+        @return: A random tuple of prime and associated generator.
```
glyph: 👍

src/twisted/conch/test/test_transport.py
```diff
@@ -310,36 +310,18 @@ def getPrimes(self):
         return {2048: (group14,), 4096: ((5, 7),)}
 
 
-class MockOldFactoryPublicKeys(MockFactory):
+class LegacyGetDHPrimesFactory(MockFactory):
     """
-    The old SSHFactory returned mappings from key names to strings from
-    getPublicKeys().  We return those here for testing.
+    A factory that implements L{twisted.conch.ssh.factory.SSHFactory.getDHPrime}
+    with a single argument.
```
glyph, line 316: Can you include a Twisted version here, and why we might have 3rd-party implementations of this, so that it's more immediately clear when we can drop tests that use it?

src/twisted/conch/ssh/transport.py
```diff
@@ -1645,7 +1645,11 @@ def ssh_KEX_DH_GEX_REQUEST(self, packet):
             return
         self.dhGexRequest = packet
         min, ideal, max = struct.unpack(">3L", packet)
-        self.g, self.p = self.factory.getDHPrime(ideal)
+        try:
+            self.g, self.p = self.factory.getDHPrime(ideal, min, max)
+        except TypeError:
+            # Fallback to backward compatibly factory.
```
glyph, line 1651: This should emit a deprecation warning making it clear what has happened. Possibly even `log.failure` the traceback, so we don't lose information?

glyph (changes requested):
Requested changes are hopefully pretty minor; I addressed the failing `mypy` issue.

PR #1605, opened 2021-05-26
[10206] Use ed25519_supported to detect support for curve25519-sha256.

adiroiban:
## Scope and purpose

This is a fix to explicitly check for ED25119 support as part of  curve25519-sha256 key exchange algorithm support detection.

I think that `backend.x25519_supported()` was used due to an error.
In theory both ed25519 and x25519 are supported since 1.1.1 
See the python cryptography issue for more details https://github.com/pyca/cryptography/issues/5164 and links to OpenSSL bugs.

This was not detected and i think that all the tests were always executed with latest OpenSSL.

This bug is observed on Ubuntu 18.04 

No test was added as we don't have any test suite for Ubuntu 18.04 that is executed with a cryptography compiled to use the OS openssl library. 

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10206

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10206

 Use backend.ed25519_supported() to detect support for curve25519-sha256.
```

### DIFF
--- a/src/twisted/conch/ssh/_kex.py
+++ b/src/twisted/conch/ssh/_kex.py
@@ -276,7 +276,7 @@ def getSupportedKeyExchanges():
                 ec.ECDH(), _curveTable[keyAlgorithmDsa]
             )
         elif keyAlgorithm.startswith(b"curve25519-sha256"):
-            supported = backend.x25519_supported()
+            supported = backend.ed25519_supported()
         else:
             supported = True
         if not supported:
--- a/src/twisted/newsfragments/10206.bugfix
+++ b/src/twisted/newsfragments/10206.bugfix
@@ -1,0 +1,2 @@
+twisted.conch.ssh._kex.getSupportedKeyExchanges() will now return than
+curve25519-sha256 is supported only for OpenSSL 1.1.1b or greater.


cjwatson:
I don't understand your comment in the ticket that "we don't need X25519 , but Ed25519".  That's incorrect.  The `curve25519-sha256` key exchange algorithm is separate from support for the Ed25519 public key type, and if you look in `SSHTransportBase._generateECPrivateKey` and friends you'll see that the key exchange code does in fact use `x25519`.  Checking `x25519_supported` was intentional.

Do you have practical evidence that the Curve25519 exchange fails on Ubuntu 18.04?  If so, could you post a reproducer?

cjwatson:
Also, I've just done an interop test between OpenSSH on Ubuntu 20.04 and Twisted trunk on Ubuntu 18.04 (using `docs/conch/examples/sshsimpleserver.py`), and confirmed that it successfully does a `curve25519-sha256` key exchange.  Unless I'm missing something, I think this PR should be closed.

adiroiban:
Thanks. Sorry for the noise. Yes. I think that it can closed. I will try to reproduce and re-open if required.

adiroiban:
Note that the issue is only on Ubuntu 18.04 which has a bit older OpenSSL version.
But I will come back with a manual tests and will reopen if I can reproduce it.

cjwatson:
Yes, I tested with the server end on Ubuntu 18.04 because of that.

PR #1606, opened 2021-05-28
9792: Run inlineCallbacks callbacks outside of the _DefGen_Return exception context

mstojcevich:
## Scope and purpose

Python 3's implicit exception chaining will chain `_DefGen_Return` or `StopIteration` onto exceptions raised during `inlineCallbacks`-decorated function's callbacks because those exceptions are in the current exception context while the callbacks are running.

This PR changes `inlineCallbacks` to run the callbacks outside of the `_DefGen_Return`/`StopIteration` exception contexts.

## Contributor Checklist:

* [X] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9792
* [X] I have updated the automated tests and checked that all checks for the PR are green.

```

Author: mstojcevich, twm
Reviewer: graingert
Fixes: ticket:9792, ticket:10262

Run inlineCallbacks callbacks outside of the _DefGen_Return exception context

Python 3's implicit exception chaining will chain _DefGen_Return or StopIteration onto exceptions raised during inlineCallbacks-decorated function's callbacks because those exceptions are in the current exception context while the callbacks are running. This changes inlineCallbacks to run callbacks outside of the context of _DefGen_Return and StopIteration exceptions.
```


src/twisted/internet/defer.py
```diff
@@ -1661,8 +1663,8 @@ def _inlineCallbacks(
                 result = current_context.run(gen.send, result)
         except StopIteration as e:
             # fell off the end, or "return" statement
-            status.deferred.callback(getattr(e, "value", None))
-            return
+            should_callback = True
+            callback_val = getattr(e, "value", None)
```
graingert: I think this is always present from 3.3+

```suggestion
            callback_val = e.value
```

graingert: Asyncio also sets the result in the except block https://github.com/python/cpython/blob/3.9/Lib/asyncio/tasks.py#L265

src/twisted/test/test_defer.py
```diff
@@ -3557,3 +3558,96 @@ async def testFunction() -> bool:
...
+        there should not be any exception context if there wasn't any to begin
+        with.
+        """
+        for do_nothing in (self.doNothingReturn, self.doNothingReturnValue):
+            yield do_nothing()
+            # There were never any exceptions that this function was aware of
+            # (_DefGen_Return and StopIteration are implementation details
+            #  of 'do_nothing' and shouldn't leak into the exc_info of the
+            #  caller)
+            self.assertEqual(sys.exc_info(), (None, None, None))
```
graingert, line 3687: Ah amazing I've always seen weird StopIteration related Tracebacks and had no idea why

graingert (changes requested):
Can you test this with async/await also please

src/twisted/internet/defer.py
```diff
@@ -1735,12 +1737,16 @@ def _inlineCallbacks(
                     lineno,
                 )
 
-            status.deferred.callback(e.value)
-            return
+            should_callback = True
```
graingert: I think this is cleaner with `break`


```python
while 1:
    ...
    callbackVal = e.value
    break
    ...
    callbackVal = e.value
    break

status.deferred.callback(callbackVal)
```

graingert: @mstojcevich here's the change https://github.com/twisted/twisted/commit/39ee52c3db975308a0d0ba336d4adb2673d48c5d

graingert:
@mstojcevich these tests pass already without the changes to `defer.py`: https://github.com/twisted/twisted/pull/1654

mstojcevich:
> @mstojcevich these tests pass already without the changes to `defer.py`: #1654

Oh, interesting. I originally wrote these having them sleep instead of yielding something that does nothing. I think having a real callback that came back from the reactor might have been important to have them fail without the changes. I'll see if I can get them working right.

exarkun:
Just noting there is a lot of overlap between this PR and https://github.com/twisted/twisted/pull/1662

glyph:
@mstojcevich I'm sorry we don't have a ton of review bandwidth these days, but would you mind resolving conflicts and addressing the feedback here?

twm:
I have merged forward and added tests based on the reproducer attached to [Trac #9792](https://twistedmatrix.com/trac/ticket/9792).

adiroiban:
@twm there is also #1662 , maybe we can have that merged first as it looks to me like it overlaps a bit

graingert:
@twm can you demo that the tests fail without the fix by reverting the fix on a new branch?

adiroiban:
I think that the demo part should be done by the reviewer.

As part of the review for #1662 I did just that, kept the tests, reverted the implementation changes, the test failed

adiroiban (changes requested):
I am not sure if we still need this.

#1662 was merged and this PR now has conflicts.

Do we still need this PR?

thanks!

twm:
@adiroiban The fix in #1662 was equivalent, so we don't need the changes here. There are additional tests here, but as they didn't actually reproduce the bug I'm not sure of their value. At the least conflicts would need to be resolved, and the 0.1 second sleep in the tests reduced to 0.0. I don't plan to do any of that myself &mdash; my goal was to get whichever PR across the finish line.

twm:
Given the above I will close this for now, but if you are interested in landing these tests please feel free to reopen @mstojcevich!

PR #1607, opened 2021-06-16
10208: Add a PyNaCl-based fallback path for Ed25519

cjwatson:
## Scope and purpose

Some commonly-used platforms that Twisted still supports, such as Ubuntu 18.04 LTS, lack a version of OpenSSL new enough to support Ed25519 via cryptography.  Add a fallback path using PyNaCl to support these.

I went back and forward on the best way to handle installing PyNaCl.  I think it's probably reasonable to just add it to the `conch` extra as I've done here, because installing it doesn't too onerous (it uses `libsodium`, but has a built-in copy of it, and PyPI has plenty of wheels), and it results in the convenient property that Ed25519 is always supported rather than being conditional.  However, if this turns out to be a problem on some platforms, then we can drop it from the extra, import it conditionally, and just say that you need to install it manually if you don't have a new enough OpenSSL.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10208

```

Author: cjwatson
Fixes: ticket:10208

Add a PyNaCl-based fallback path for Ed25519

Some commonly-used platforms that Twisted still supports, such as Ubuntu
18.04 LTS, lack a version of OpenSSL new enough to support Ed25519 via
cryptography.  Add a fallback path using PyNaCl to support these.
```


setup.cfg
```diff
@@ -83,6 +83,7 @@ conch =
     cryptography >= 2.6
     appdirs >= 1.4.0
     bcrypt >= 3.0.0
+    PyNaCl
```
adiroiban, line 90: Maybe add a separate target.

Most users will have a modern cryptography and nacl is not needd

```suggestion
conch-nacl:
    %(conch)s
    PyNaCl
```

adiroiban:
Thanks for the PR and sorry for the late reply. You can ping me for any SSH related PR and I will try to make it a top priority.

I guess that this is related to https://github.com/twisted/twisted/pull/1605

--------

I think that using pynacl is the right solution.

I would prefer to have pynacl as a soft-dependency so that users with a modern openssl+crytography don't have to worry about pynacl on Alpine Linux.

setup.cfg
cjwatson: Seems reasonable.  Done (not quite via your suggestion, but similar).

cjwatson:
@adiroiban Thanks.  I agree that a soft dependency makes sense; for some reason I just hadn't really thought of the approach of adding a new extra.  I hope this is better now.

(I noticed #1605 while looking for possibly-related tickets before filing my own.  This bit has been in my sprawling plan since 2018 ...)

src/twisted/conch/ssh/keys.py
```diff
@@ -26,6 +27,8 @@
     load_pem_private_key,
     load_ssh_public_key,
 )
+from nacl.exceptions import BadSignatureError  # type: ignore[import]
```
adiroiban: I think that for a soft-dependency this needs to be wrapped in a `try/except ImportError` and define `BadSignatureError = object` or something like this

I am no what  is best way to implement it as a soft-dependency.
Because if you have it as a soft-dependency  and pynal is not installed and you have an older OpenSSL  you will not have support for ED25519

cjwatson: Oh, right, good point.  I'll sort this out as soon as I get a moment.

adiroiban: No hurry from my side. I did a quick review and the code looks very well written.
I am looking to do a 20.7.0 release ASAP...but we can also do a 20.8.0 release to include this PR.

cjwatson:
@adiroiban Sorry for my long delay on this!  I got badly bogged down in `mypy` issues, until I realized that if I used composition instead of inheritance to stitch in the `nacl.signing`-based implementation then everything got a lot simpler, because that made space for simply inheriting from `ed25519.Ed25519{Public,Private}Key` and then type inference has a much easier time of it.

This is ready for another review now.

adiroiban (approved):
Excellent work. Thanks!

I have updated the branch with master to trigger a new round of tests.

If all is green this can be merged. 

You left the commit message in the PR description, so anyone can see this PR green can merge it :)

adiroiban:
I have enabled the auto-merge. In this way we should not have to wait for the CI status in order to merge. 

Let me know if you prefer manual merge.
And I think that you can cancel the auto-merge for this PR at any time.

PR #1624, opened 2021-07-11
10222 avoid AssertionError in deferred when debugging is disabled

graingert:
## Scope and purpose

defect fix and reproduction in test

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10222

```

(as long as you wish)
```


src/twisted/test/test_defer.py
```diff
@@ -1516,6 +1516,23 @@ def test_errbackWithNoArgsNoDebug(self) -> None:
...
+            pass
+
+        async def c() -> None:
+            for _ in range(2):
+                with self.assertRaises(MyError):
+                    await d
+
+        d2 = Deferred.fromCoroutine(c())
+        d.errback(MyError())
+        self.assertIsNone(d._debugInfo)
```
graingert, line 1585: for some reason: `defer.fail(MyError())._debugInfo` is always truthy regardless of `defer.setDebugging(False)` setting see https://twistedmatrix.com/trac/ticket/10223#comment:1

graingert:
it looks like there's something more fundamentally wrong with reusing-deferreds in coros:

```python
from twisted.internet import defer, task


class MyError(Exception):
    pass


async def sleep(delay, result=None):

    await task.deferLater(clock=reactor, delay=delay)
    return result


async def foo(d):
    try:
        await d
    except MyError:
        pass

    # d.result should be None here!
    print(d.result)  # [Failure instance: Traceback (failure with no frames): <class '__main__.MyError'>: 
    await sleep(0)
    print(d.result)  # it's correctly None now


async def amain(_):
    d = defer.Deferred()
    d2 = defer.Deferred.fromCoroutine(foo(d))
    d.errback(MyError())
    result = await d2
    print(result)


task.react(amain)
```

it appears as though once a Deferred is awaited the result should always be replaced with `None`

https://github.com/twisted/twisted/blob/5c24e99e671c4082a1ddc8dbeb869402294bd0dc/src/twisted/internet/defer.py#L1672-L1677

but if you call `await d; await d;` the result isn't stripped between turns of the `_inlineCallbacks` trampoline

src/twisted/internet/defer.py
```diff
@@ -1743,14 +1751,14 @@ def _inlineCallbacks(
 
         if isinstance(result, Deferred):
             # a deferred was yielded, get the result.
-            def gotResult(r: object) -> None:
+            def gotResult(r: object, d: Deferred) -> None:
```
```suggestion
            def gotResult(r: object) -> None:
```

src/twisted/internet/defer.py
```diff
@@ -1743,14 +1751,14 @@ def _inlineCallbacks(
...
                 if waiting[0]:
                     waiting[0] = False
                     waiting[1] = r
                 else:
                     current_context.run(_inlineCallbacks, r, gen, status)
 
-            result.addBoth(gotResult)
+            result.addBoth(gotResult, result)
```
```suggestion
            result.addBoth(gotResult)
```

src/twisted/defer/newsfragments/10222.bugfix
```diff
@@ -0,0 +1 @@
+avoid AssertionError in deferred awaited twice when debugging is disabled
```
graingert:
```suggestion
avoid AssertionError in deferred awaited twice when debugging is disabled
always clear the result of awaited Deferreds
```

graingert:
so the problem is here:

https://github.com/twisted/twisted/blob/5c24e99e671c4082a1ddc8dbeb869402294bd0dc/src/twisted/internet/defer.py#L1670-L1679

when you await a Deferred you expect the result to be cleared as a result of `gotBoth` returning None. However there's two ways to get the result from a Deferred on `await d`:

```python
d = defer.succeed(object())
await d   # as an optimization, doesn't call d.addBoth because the deferred is already done
```

```python
d = defer.Deferred[object]()

async def c():
    v = await d   # calls d.addBoth which clears the result. v is mock.sentinel.done
    w = await d   # c().__await__.send() gets called before gotResult returns None. w is v is mock.sentinel.done
    await sleep(0)  # run around the Clock's trampoline so gotResult returns
    x = await d   # the d.result has been set to the return value of gotResult (None) x is None

d2 = defer.Deferred.fromCoroutine(c())
d.callback(mock.sentinel.done)
```

therefore `defer.Deferred.send` and `defer.Deferred.throw` should pre-empt `_inlineCallbacks`' `gotResult`s' return value of `None`

src/twisted/internet/defer.py
```diff
@@ -942,6 +942,19 @@ def __str__(self) -> str:
     def __iter__(self) -> "Deferred[_DeferredResultT]":
         return self
 
+    def close(self) -> None:
+        self.result = None
```
graingert: is there anything else we need to do when we close this IterableCoroutine?

src/twisted/test/test_defer.py
```diff
@@ -3183,7 +3226,7 @@ def test_ensureDeferredGenerator(self) -> None:
 
         def run() -> Generator[Deferred[str], None, str]:
             d = defer.succeed("foo")
-            res = cast(str, (yield from d))
+            res: str = (yield from d)
```
graingert, line 3274: `yield from d` is typed correctly now!

src/twisted/internet/defer.py
```diff
@@ -939,38 +939,23 @@ def __str__(self) -> str:
...
-            assert self._debugInfo is not None
-            self._debugInfo.failResult = None
-            result.value.__failure__ = result
-            raise result.value
-        else:
-            raise StopIteration(result)
+    def __iter__(
+        self,
+    ) -> "Generator[Deferred[_DeferredResultT], _DeferredResultT, _DeferredResultT]":
+        return (yield self)
```
graingert: from https://github.com/python-trio/trio/blob/f2e23123b196358b85556c592cfd26882261b738/trio/_core/_traps.py#L11-L21

src/twisted/internet/defer.py
```diff
@@ -939,38 +951,27 @@ def __str__(self) -> str:
...
+        # https://github.com/python-trio/trio/blob/ab75cd12a4db68e2fbb2e58f46900ebaa582aab2/trio/_core/_run.py#L2160-L2166
+        return (yield self.addBoth(_Outcome)).unwrap()  # type: ignore[no-any-return,attr-defined,arg-type]
+
+    if TYPE_CHECKING:
+        #     Callable[[], Generator[Any, None, _DeferredResultT]]
+        #     See: https://github.com/python/typeshed/issues/5125
+        #     When the typeshed patch is included in a mypy release,
+        #     this method can be replaced by `__await__ = __iter__`.
+        def __await__(self) -> Generator[Any, None, _DeferredResultT]:
+            return self.__iter__()  # type: ignore[return-value]
```
glyph, line 1004: A method body in an `if TYPE_CHECKING` block is redundant, and should just be `...`

src/twisted/internet/defer.py
```diff
@@ -939,38 +951,27 @@ def __str__(self) -> str:
...
-    def __await__(self) -> Generator[Any, None, _DeferredResultT]:
-        return self.__iter__()  # type: ignore[return-value]
+    def __iter__(
+        self,
+    ) -> "Generator[Deferred[_DeferredResultT], _DeferredResultT, _DeferredResultT]":
+        # Wrap failures in an _Outcome object here to avoid buggy .throw()
+        #   https://bugs.python.org/issue29587
+        #   https://bugs.python.org/issue29590
+        # https://github.com/python-trio/trio/blob/ab75cd12a4db68e2fbb2e58f46900ebaa582aab2/trio/_core/_run.py#L2160-L2166
+        return (yield self.addBoth(_Outcome)).unwrap()  # type: ignore[no-any-return,attr-defined,arg-type]
```
glyph, line 996: It seems like this should be a `cast` instead of all these `type:ignore`s? Any `type: ignore` should be scoped as tightly as possible to a single expression

src/twisted/internet/defer.py
graingert: This is actually a type assertion about `self.__iter__()` but I can remove it as there's no external effect

src/twisted/defer/newsfragments/10222.bugfix
```diff
@@ -0,0 +1,2 @@
...
+always clear the result of awaited Deferreds
```
glyph: If there are 2 bugs being fixed here, there should be 2 tickets and 2 news files, the format here is not multi-line.

src/twisted/internet/defer.py
```diff
@@ -345,6 +345,18 @@ def __del__(self) -> None:
...
+class _Outcome:
+    _result: object
+
+    @_extraneous
+    def unwrap(self) -> object:
+        result = self._result
+        if isinstance(result, Failure):
+            result.raiseException()
+        return result
+
```
glyph, line 392: Adding all these extra allocations and faffing around when it's only required for old versions of Python seems wasteful. All these bugs are fixed on pythons released in the last 5 years. I realize we may need to support these older versions but adding additional runtime overhead in a *very* hot path in every version of python seems like an odd choice. (Gosh I want speedcenter to put commit statuses on PRs so I don't have to hypothesize about performance like this, but allocating a new instance for every single `await` has gotta be bad.)

graingert: @glyph there's a lot of frame sniffing and tb_next munging going around here that I don't know how to fix otherwise.

src/twisted/defer/newsfragments/10222.bugfix
graingert:
```suggestion
always consume the result of awaited Deferreds
```

graingert: the assertion error in debug mode was actually just fallout from the result not being consumed correctly

src/twisted/internet/defer.py
graingert: it's this code here that confounds me the most https://github.com/twisted/twisted/blob/a4fe705c5b450a946966cd63d19b3ba0f8d6cba8/src/twisted/python/failure.py#L502-L574

glyph: > there's a lot of frame sniffing and tb_next munging going around here that I don't know how to fix otherwise

In the worst case scenario, just have a `_3_7_plus` version of the implementation and swap in the whole function at import time.

src/twisted/defer/newsfragments/10222.bugfix
glyph: In that case the feedback is "Make it a full, grammatically correct sentence" :)

src/twisted/defer/newsfragments/10222.bugfix
```diff
@@ -0,0 +1 @@
+always consume the result of awaited Deferreds
```
glyph: This reads more like a commit message than a news fragment. Who are you commanding to always consume the result of an awaited deferred? the user? what consumes it? how?

glyph: (i.e. when someone reads this in a changelog, how do they understand what behavior has actually changed?)

graingert:
```suggestion
``Deferred.fromCoroutine`` (and therefore also ``ensureDeferred``) now always consumes the result of awaited Deferreds. Previously awaiting the same Deferred synchronously multiple times in a row would resolve to the same result.
```

src/twisted/internet/defer.py
graingert: the test failure still happens on Python 3.10

glyph: Ah. Hm. In that case I guess this is unfortunate but inevitable. I thought the cpython bugs were fixed ages ago?

graingert: here's the error log: https://github.com/twisted/twisted/pull/1624/checks?check_run_id=3131779811#step:9:17950

src/twisted/internet/defer.py
wsanchez: I vote for keeping it so we notice if the ignore becomes unnecessary.

glyph:
Hopefully I didn't mess up that commit and pre-commit can deal with any linting…

src/twisted/internet/defer.py
glyph: Yeah, OK, suggestion withdrawn

src/twisted/python/failure.py
```diff
@@ -207,10 +209,19 @@ def co_positions(self):
         return ((None, None, None, None),)
 
 
-_inlineCallbacksExtraneous = []
+_inlineCallbacksExtraneous: List[types.CodeType] = []
 
 
-def _extraneous(f):
+class _HasCode(typing_extensions.Protocol):
```
glyph, line 215: This probably wants to have a docstring.

glyph (changes requested):
@graingert can you sort out conflicts and CI status?

PR #1616, opened 2021-07-08
[Fix #10216] Update the release documentation with feedback from the 21.7.0 release.

adiroiban:
## Scope and purpose

This is a minor release doc update  while doing an actual release.

Updated docs at https://twisted--1616.org.readthedocs.build/en/1616/core/development/policy/release-process.html

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10216
* [NA] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] The title of the PR starts with the associated Trac ticket number (without the `#` character).
* [NA] I have updated the automated tests and checked that all checks for the PR are green.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.
* [x] The merge commit will use the below format
  The first line is automatically generated by GitHub based on PR ID and branch name.
  The other lines generated by GitHub should be replaced.

```
Merge pull request #123 from twisted/4356-branch-name-with-trac-id

Author: adiroiban
Reviewer: 
Fixes: ticket:10216

Update the release documentation.
```

PR #1629, opened 2021-07-15
[#10228] On relase automate the default/stable version update on Read The Docs.

adiroiban:
## Scope and purpose

For now, after a release is done we need to manually log in to  RTD and update the default version.

This PR tries to automate that process, so that the release manager will not need to care about RTD settings.

So after a stable release we should have the stable branch updated https://github.com/twisted/twisted/tree/stable
And this will cause the RTD docs to also be updated https://docs.twistedmatrix.com/en/stable/

## Implementation

The release is only triggered by the creation o tag.
uses bash to differenciat between a release candidate tag and a final release tag.
Uses git CLI to force push and rebase the origin/stable branch.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10228

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10228

Automatically update the RTD default version on release.
```


adiroiban:
Please do an initial review and let me know if the changes make sense.

I don't know how to test them...as they are designed for the release automation.

But after 21.7.0 is release, I can delay the manual RTD update and try to run part of this code.

adiroiban:
This is ready for initial review to see if it's worth doing it.

But still in experimental state.

I am not sure how RTD is deciding for which tag to build ... the 21.7.0rc1 is a valid tag, but I don't see any version for it.

So the RTD code might need to be updated to create the version itself...

The the main issues might be due to the fact that the tag is named `twisted-1.2.3` and not just `1.2.3` ...

https://docs.readthedocs.io/en/stable/versions.html

And I might need support for RTD team as I think that the "stable" RTD feature is broken for the Twisted project.

adiroiban:
I think that this can be simplified by using a branch named "stable" and making sure that we push the latest release to that branch...and RTD will build the doc for it

I will see what we can do

adiroiban:
This is draft while I am waiting for feedback from RTD support team.

But if there is no workaround, the plan is to have a "stable" branch that is automatically pushed after a release  and use the RTD API to create the documentation for the new release tag.

adiroiban:
This is ready for review.

------------

Here is a test run, in which the "test-stable-branch" branch is forced updated

https://github.com/twisted/twisted/pull/1629/checks?check_run_id=3213885636

it uses a code in which the final-release precondition is ignored.

----------

I don't know how to test this in PROD.
I guess that we just have to review  the code and hope it will work with the next release.

The important part is to check that the regex is ok

You can manually use this script in bash

```bash
REGEX=refs/tags/twisted-1.2.3
STABLE_REF_RE='.*twisted-[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$'
if [[ $REGEX =~ $STABLE_REF_RE ]]; then echo "OK"; fi
```

--------

The documentation will be updated as part of https://github.com/twisted/twisted/pull/1616

which is a parallel PR

graingert (approved):

adiroiban:
Thanks for the review. Much appreciated.

I forgot to mention a minor issue with this change.

if you latest release is 21.7.0 and we will decide to do some sort of security release for 21.2.0 which will be released as 21.2.1 this will overwrite the "stable" branch.

I don't expect that we will have many situation like this... and in this case, you can always manually force push and reset the "stable" branch.

I  will document this as part of #10216

graingert:
We'd most likely have security issues with all releases, we'd just have to apply them in order

PR #1631, opened 2021-07-18
9555 sarhamam replace bytes with bytes

Sarhamam:
When replacing bytes, encode `os.linesep` which is str, and prefix `\r\n` with b.


twm:
Hi @Sarhamam, this looks like a reasonable fix but you'll have to [add some sort of test coverage of this function](https://twistedmatrix.com/trac/wiki/ReviewProcess#Authors:Thingsyourbranchorpatchmustcontain) before we can merge it.

src/twisted/protocols/ftp.py
```diff
@@ -667,7 +667,7 @@ def __init__(self, cons):
             self.write = cons.write
```
graingert, line 667: This conditional is wrong in the same way too

src/twisted/protocols/ftp.py
```diff
@@ -667,7 +667,7 @@ def __init__(self, cons):
...
 
     def write(self, bytes):
-        return self.cons.write(bytes.replace(os.linesep, "\r\n"))
+        return self.cons.write(bytes.replace(os.linesep.encode(), b"\r\n"))
```
graingert, line 670: Probably worth setting
```
_LINESEP_BYTES = os.linesep.encode()
```
somewhere

src/twisted/protocols/ftp.py
Sarhamam: I dont think this conditional is wrong, `os.linesep` return str type, and so is "\r\n", no need to convert to bytes here.

src/twisted/protocols/ftp.py
Sarhamam: Done at the top of the file, along with _CRLF_BYTES constant.

Sarhamam:
Having problems with fork - will create new PR once ready.

PR #1632, opened 2021-07-18
10231 Fix type annotation for `inlineCallbacks`

richvdh:
## Scope and purpose

Propagate the ReturnType of the generators wrapped by `inlineCallbacks`. 

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10231

```

Author: richvdh
Reviewer: 
Fixes: ticket:10231

Update the type annotation on `inlineCallbacks`.
```

### DIFF
--- a/src/twisted/newsfragments/10231.bugfix
+++ b/src/twisted/newsfragments/10231.bugfix
@@ -1,1 +1,1 @@
-twisted.internet.defer.inlineCallbacks has an improved type annotation, to avoidtyping errors when it is used on a function which returns a non-None result.
+twisted.internet.defer.inlineCallbacks has an improved type annotation, to avoid typing errors when it is used on a function which returns a non-None result.
--- a/src/twisted/newsfragments/10231.bugfix
+++ b/src/twisted/newsfragments/10231.bugfix
@@ -1,1 +1,1 @@
-Fix type annotations for inlineCallbacks.
+twisted.internet.defer.inlineCallbacks has an improved type annotation, to avoidtyping errors when it is used on a function which returns a non-None result.
--- a/src/twisted/newsfragments/10231.bugfix
+++ b/src/twisted/newsfragments/10231.bugfix
@@ -1,1 +1,1 @@
-Fix type annotations fro inlineCallbacks.
+Fix type annotations for inlineCallbacks.
--- a/src/twisted/newsfragments/10231.bugfix
+++ b/src/twisted/newsfragments/10231.bugfix
@@ -1,0 +1,1 @@
+Fix type annotations fro inlineCallbacks.
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1828,8 +1828,8 @@ class _InternalInlineCallbacksCancelledError(Exception):
 # type note: "..." is used here because we don't have a better way to express
 #     that the same arguments are accepted by the returned callable.
 def inlineCallbacks(
-    f: Callable[..., Generator[Deferred[object], object, None]]
-) -> Callable[..., Deferred[object]]:
+    f: Callable[..., Generator[Deferred[object], object, _T]]
+) -> Callable[..., Deferred[_T]]:
     """
     L{inlineCallbacks} helps you write L{Deferred}-using code that looks like a
     regular sequential function. For example::


adiroiban:
Thanks for the quick PR.
Can we have it targeting trunk?

I will the cherry pick it in the release branch.

Thanks

richvdh:
sure, done

graingert:
I'm not really sure what the point of setting an accurate return type annotation on inlineCallbacks is at all. Currently the lack of ParamSpec support means the arguments of the decorated function can't be preserved, and if you can use `return value` syntax in a generator in your source code then you can use the preferred `async/await` syntax that supports Deferred annotations fully

src/twisted/newsfragments/10231.bugfix
```diff
@@ -0,0 +1 @@
+Fix type annotations for inlineCallbacks.
```
graingert:
```suggestion
Improve type annotations for inlineCallbacks.
```

graingert: this is an improvement, but it's not fixed: It's not fixable without fixing `Deferred.__iter__` and using ParamSpec to propagate the decorated call signatute

adiroiban: If I understood this correctly.

```suggestion
twisted.internet.defer.inlineCallbacks has now improved type annotation that will not raise an error for non-object return values.
```


but basically this is like  no typing at all as it can return anything.

You can ignore my comment and I am still trying to figure it out why we need 100% typing in Python

graingert: @adiroiban this isn't correct either. There's no such thing as "non-object return values"

graingert: here's the issue and the fix more clearly: https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=64b950daeb501e6a6bd0592d9fe456d5

richvdh: I've updated the newsfile - hopefully it is better now?

richvdh:
> I'm not really sure what the point of setting an accurate return type annotation on inlineCallbacks is at all. Currently the lack of ParamSpec support means the arguments of the decorated function can't be preserved, and if you can use `return value` syntax in a generator in your source code then you can use the preferred `async/await` syntax that supports Deferred annotations fully

I don't agree with the last part of this. There are still situations where it is useful to use `inlineCallbacks` - and have it return a function which returns a `Deferred` of the right type. Two examples:
 * You need to maintain compatibility with older code which expects a `Deferred` rather than a coroutine - either because the caller hasn't yet been converted to `async/await` or because you're preserving a public API.
 * You're implementing an interface. For example, we have an implementation of `IAgent`, where `IAgent.request` is defined to return a `Deferred` - so an `async` function is not acceptable.

graingert:
You can still make a deferred returning function from an async function, and it preserves the call signature:

```python
async def _request(self, ham: int) -> None:
    await x

def request(self, ham: int) -> Deferred[None]:
    return Deferred.fromCoroutine(self._request(ham))
```

richvdh:
> You can still make a deferred returning function from an async function, and it preserves the call signature:

Well, I *can*, but the fact I can work around it seems a poor argument against improving the type annotation on `inlineCallbacks` anyway?

src/twisted/newsfragments/10231.bugfix
```diff
@@ -0,0 +1 @@
+twisted.internet.defer.inlineCallbacks has an improved type annotation, to avoidtyping errors when it is used on a function which returns a non-None result.
```
graingert:
```suggestion
twisted.internet.defer.inlineCallbacks has an improved type annotation, to avoid typing errors when it is used on a function which returns a non-None result.
```

graingert (changes requested):
Newsfragment typo

adiroiban (approved):
Thanks you all for the contribution.

The news fragment make sense for me now.

As long as this change is not braking anything else, I think that is OK to merge it.

@graingert any comment against merging it?

graingert (approved):

graingert:
closing and re-opening to kick off azure pipelines

graingert (approved):

PR #1641, opened 2021-07-26
10239: epytext again

twm:
## Scope and purpose

Fix syntax errors. Fix cross-references due to move of hyper-h2 docs. Addresses these messages:

```
apidocs run-test: commands[0] | .../twisted/bin/admin/build-apidocs .../twisted/src/ apidocs
Failed to get object inventory from https://hyper-h2.readthedocs.io/en/stable/objects.inv
.../twisted/src/twisted/web/server.py:705: bad docstring: Possible mal-formatted field item.
.../twisted/src/twisted/web/server.py:705: bad docstring: Fields must be the final elements in an epytext string.
.../twisted/src/twisted/protocols/haproxy/_parser.py:32: bad docstring: Bad link target.

L{Mapping[str, str]}
                   ^
.../twisted/src/twisted/web/_http2.py:68: Cannot find link target for "h2.connection.H2Connection" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:440: Cannot find link target for "h2.events.RequestReceived" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:472: Cannot find link target for "h2.events.DataReceived" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:484: Cannot find link target for "h2.events.StreamEnded" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:495: Cannot find link target for "h2.events.StreamReset" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:509: Cannot find link target for "h2.events.PriorityUpdated" (you can link to external docs with --intersphinx)
.../twisted/src/twisted/web/_http2.py:662: Cannot find link target for "h2.events.WindowUpdated" (you can link to external docs with --intersphinx)
these 2 objects' docstrings contain syntax errors:
    twisted.protocols.haproxy._parser.unparseEndpoint
    twisted.web.server.Session
```

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10239
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
[...]

PR #1643, opened 2021-07-27
10240 restore 'all-successful' CI job and simplify workflows files

graingert:
## Scope and purpose

we need a job to [select as a required build](https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert) so that github "automerge" doesn't just instantly merge the PR

This also adds  `--warnings-as-errors` to pydcotor to prevent https://github.com/twisted/twisted/pull/1641 happening again

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10240

```

(as long as you wish)
```


.github/workflows/lint.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Lint
       run: |
         python --version
-        tox -e lint
+        tox -e lint,apidocs
```
adiroiban: I think that instead of updating lint, we should fix the docs.yaml workflow to fail on pydoctor warnings

graingert: we already have way too many workflows, I'd like to push more complexity into tox and out of all the .github files

adiroiban: well...the docs.yaml just calls `tox -e apidocs` which calls `bin/admin/build-apidocs`  which calls `twisted.python._release.BuildAPIDocsScript` ... is just that for narrative docs and apidocs we have a separate workflow... so that for example if a flaky test fails we can re-run only the trial tests...without having to rerun lint and documentation generation

but yes.. I would say that mypy.yaml and docs.yaml and lint.yaml can be merged.

adiroiban: I see that lint now takes 7 minutes... I think that one of the reason why docs was moved to a separate workflow, was to have the docs executed in a separate VM... so that we can get a fast feedback.

But maybe we can run tox on 2 CPUs ... or build the docs as a separate job inside the lint.yaml

.github/workflows/test.yaml
```diff
@@ -168,3 +168,49 @@ jobs:
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip tox
+
+    - name: Lint
```
adiroiban: I think the name of the step (and also for the overall job) is no longer `lint` ... but I don't know what name to use :)
maybe "tox run" ?

graingert: It's stuff that's `not test`

graingert: What about calling it `nit`?

adiroiban: I think that any name, other than 'lint' would be ok :) ... but this is the tox environments run ... 
but nit is a bit gross :p

graingert: > this is the tox environments run 

the other stages run tox environments - specifically the `py` environments

.github/workflows/test.yaml
```diff
@@ -168,3 +168,49 @@ jobs:
       uses: pypa/gh-action-pypi-publish@v1.3.1
       with:
         password: ${{ secrets.PYPI_UPLOAD_TOKEN }}
+
+  lint:
```
adiroiban: mabye 
```suggestion
  StaticChecks:
```

.github/workflows/test.yaml
```diff
@@ -168,3 +168,49 @@ jobs:
...
+        git reset --hard $BRANCH_SHA
+
+    - name: Newsfragment
+      run: |
+        tox -q -e newsfragment
+
+  all-successful:
+    # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
+    runs-on: ubuntu-latest
+    needs: [testing, lint, pypi-publish]
```
```suggestion
    needs: [testing, StaticChecks, pypi-publish]
```

.github/workflows/test.yaml
adiroiban: I don't have a better name... but I think that this is a good description for the step

```suggestion
    - name: Run the checks
```

adiroiban (approved):
It looks good. Thanks.

Just a minor comment to update the name of the step, which is now called "lint" but it also runs many other checks.

PR #1645, opened 2021-07-28
#10210 Improve connection error handling in ESMTPSenderFactory

### LINKED ISSUE #10210: twisted.mail.smtp reports confusing exceptions when TLS fails
|[richvdh's avatar](https://github.com/richvdh)| @richvdh reported|
|-|-|
|Trac ID|trac#10210|
|Type|defect|
|Created|2021-06-22 21:32:55Z|

When there is a TLS error connecting to an SMTP server, `twisted.mail.smtp.sendmail` (and the lower-level classes in `twisted.mail.smtp`) raise a misleading exception:

Omitting application-specific frames:

```
Capture point (most recent call last):
...
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/base.py", line 1423, in run
    self.mainLoop()
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/base.py", line 1433, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/home/synapse/src/synapse/metrics/__init__.py", line 548, in f
    ret = func(*args, **kwargs)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/base.py", line 999, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/tcp.py", line 507, in connectionLost
    self.connector.connectionLost(reason)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/base.py", line 1312, in connectionLost
    self.factory.clientConnectionLost(self, reason)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/mail/smtp.py", line 1935, in clientConnectionLost
    self._processConnectionError(connector, err)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/mail/smtp.py", line 1952, in _processConnectionError
    self.result.errback(err.value)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/defer.py", line 517, in errback
    self._startRunCallbacks(fail)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/defer.py", line 580, in _startRunCallbacks
    self._runCallbacks()
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/defer.py", line 662, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/defer.py", line 1511, in gotResult
    current_context.run(_inlineCallbacks, r, g, status)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/internet/defer.py", line 1443, in _inlineCallbacks
    result = current_context.run(result.throwExceptionIntoGenerator, g)
  File "/home/synapse/env/lib/python3.7/site-packages/twisted/python/failure.py", line 500, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
...
twisted.internet.error.ConnectionAborted: Connection was aborted locally using ITCPTransport.abortConnection.
```

The problem seems to stem from `SMTPSenderFactory.clientConnectionFailed` being called with an unhelpful `ConnectionAborted` rather than anything more descriptive.

This seems to be thanks to [https://github.com/twisted/twisted/blob/twisted-21.2.0/src/twisted/protocols/tls.py#L427 `TLSMemoryBIOProtocol.failVerification`], which stashes the error and calls `Transport.abortConnection()`. Once the connection is closed, [https://github.com/twisted/twisted/blob/twisted-21.1.0/src/twisted/internet/tcp.py#L508 `tcp.Client.connectionLost`] is called with a `ConnectionAborted`; that propogates through `Connector.connectionLost` to the `Factory`'s `connectionLost` method.

```
trac-id__10210 10210
type__defect defect
reporter__richvdh richvdh
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__mail mail
keywords__review review
time__1624397575861341 1624397575861341
changetime__1643657348183696 1643657348183696
version__None None
owner__None None

```
</details>

richvdh:
## Scope and purpose

This fixes #10210 by passing the failure reason back to the SMTPSenderFactory out-of-band.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10210

```

Author: richvdh
Reviewer: 
Fixes: ticket:10210

Make sure we remember the reason that the connection failed, and raise that in
preference to "connection done" or whatever.
```


src/twisted/mail/test/test_smtp.py
```diff
@@ -674,6 +676,50 @@ def check(ignored):
...
+        def errback(f):
+            # check that we got an exception with a plausible error code.
+            f.trap(smtp.SMTPConnectError)
+            self.assertIn("certificate verify failed", str(f.value))
+
+        def callback(_res):
+            raise Exception("ESMTPSenderFactory unexpectedly completed")
+
+        sentDeferred.addCallbacks(callback, errback)
+        return sentDeferred
```
glyph: In general we now strongly prefer the style of test where you *don't* spin up a real reactor and do for-real network I/O; instead, consider using the various utilities in `twisted.internet.testing` to produce deterministic mocks for your network set-up attempts.  That said, this style of test is definitely better than nothing, so if you feel that the deterministic version would take you a bunch more time we can work with this for now (it's certainly not the first such test in this file).

src/twisted/mail/test/test_smtp.py
```diff
@@ -674,6 +676,50 @@ def check(ignored):
...
+        )
+        self.addCleanup(connector.disconnect)
+
+        def errback(f):
+            # check that we got an exception with a plausible error code.
+            f.trap(smtp.SMTPConnectError)
+            self.assertIn("certificate verify failed", str(f.value))
+
+        def callback(_res):
+            raise Exception("ESMTPSenderFactory unexpectedly completed")
```
glyph: Codecov is unhappy about this line never being exercised, which is giving you a ❌ .  To pacify it, you might consider using `self.assertFailure` (whose documentation sadly omits that it returns the `Exception` object), like this:

```python3
def checkVerifyFailedMessage(e: smtp.SMTPConnectError):
    self.assertIn("certificate verify failed", str(e))
return self.assertFailure(sentDeferred, smtp.SMTPConnectError).addCallback(checkVerifyFailedMessage)
```

This leaves the error reporting to the testing framework rather than inlining it within the test where it's going to not be exercised and leave us coverage gaps.

glyph (commented):
I don't have time for a full review now but hopefully these initial thoughts are useful!

src/twisted/mail/test/test_smtp.py
richvdh: done, thanks for the tip.

src/twisted/mail/test/test_smtp.py
richvdh: Thanks for the speedy initial review, @glyph - and sorry for the slow turnaround on my side.

Yeah, I'm aware of the preference. I initially just went with the flow in this file, but I've just taken a bit of a harder look at what it would take to use `twisted.test.iosim.FakeTransport` and friends to connect up the client and server.

It would be a bit fiddly, but the real problem is that what I really want to test here is how `_newtls.startTLS` interacts with the SMTP code, and `startTLS` relies on wrapping a real `FileDescriptor` object - or at least, something that behaves much the same as one. Possibly I could rejig `FakeTransport` enough to make `startTLS` work with it ... but at that point I start to feel like I'm just testing the tests.

So, as painful as it is, I think this might be a case where we're better off with a real reactor, at least for now.

glyph: No need to be sorry for a slow turnaround. We're all very time-constrained here in open source land, and we do things when we can.

src/twisted/mail/smtp.py
```diff
@@ -1923,21 +1928,34 @@ def clientConnectionLost(self, connector, err):
         self._processConnectionError(connector, err)
 
     def _processConnectionError(self, connector, err):
+        # If the initial connection succeeded, but there was a later error
+        # (such as TLS validation failure), we're more interested in that than
+        # "Connection was closed cleanly" or whatever.
+        if self.currentProtocol and self.currentProtocol.failureReason:
```
glyph: 1. We don't want to test `self.currentProtocol.__bool__` here, we just want to check if it `is not None`
2. I find that inline ternaries are more expressive than overloading boolean logic for tests like this, so, for clarity, I think I'd prefer this:

```suggestion
        if self.currentProtocol.failureReason is not None if self.currentProtocol is not None else False:
```

The first point is semantic, although the ternary thing is just stylistic, so feel free to skip it if you'd prefer `… is not None and … is not None`

src/twisted/mail/test/test_smtp.py
```diff
@@ -669,6 +671,47 @@ def check(ignored):
 
         return self.loopback(server, client).addCallback(check)
 
+    def test_ESMTPSenderFactory_TLSError(self):
+        """
+        Attempting to connect to an ESMTP server which presents an invalid certificate
+        should produce a meaningful error.
```
glyph: Love this docstring, very close to https://jml.io/pages/test-docstrings.html .

(It would be even nicer if it gave a crisper definition of "meaningful" but that is really nitpicking)

src/twisted/mail/test/test_smtp.py
```diff
@@ -669,6 +671,47 @@ def check(ignored):
...
+        )
+        self.addCleanup(connector.disconnect)
+
+        def checkVerifyFailedMessage(e: smtp.SMTPConnectError) -> None:
+            # check that the SMTPConnectError has a message relating to the TLS error
+            self.assertIn("certificate verify failed", str(e))
+
+        return self.assertFailure(sentDeferred, smtp.SMTPConnectError).addCallback(
+            checkVerifyFailedMessage
+        )
```
glyph, line 718: FYI, this sort of real-networking test that returns a Deferred is really not the right way to test protocol-parsing logic like this; it does unnecessarily non-deterministic I/O with the operating system.  If you can, it should be rewritten to be more robust using `twisted.test.iosim.IOPump` or similar.  At the very least however, please remove the direct call to `listenTCP` and instead use the higher-level `self.loopback(...)` to at least abstract out the connection-making logic.

glyph (commented):
Thanks so much for improving this error handling. I hope that this feedback is minor and you an address so we can land soon?

glyph:
Oops. I see i already mostly gave this review before. I wonder why the robot did not remove the `needs-review` label?

glyph (changes requested):
Hello robot. Please see comment review for requested changes.

adiroiban:
@glyph 
> Oops. I see i already mostly gave this review before. I wonder why the robot did not remove the needs-review label?

You only left a review with an "indecisive comment"
The robot only updates the label on "approved" or "request changes"

glyph (changes requested):
Next time escape the robot command in that comment @adiroiban  😉

src/twisted/mail/smtp.py
richvdh: Honestly I find the ternary expression utterly incomprehensible, but I've corrected this to check for Noneness.

src/twisted/mail/test/test_smtp.py
richvdh: Yeah. As I wrote [before](https://github.com/twisted/twisted/pull/1645#discussion_r682124969), I struggled to get this to work with `iosim` because `startTLS` wants to wrap a real file descriptor.

I had a look at making it work with `self.loopback`. It could be my failure to understand how to use `loopback`, but as far as I can tell, it wants to build its own `ClientFactory` (a `LoopbackClientFactory`), which seems like it makes it impossible to instantiate an `SMTPSenderFactory` (which is where the interesting changes are here).

glyph:
please review

src/twisted/mail/test/test_smtp.py
richvdh: @exarkun [says](https://matrix.to/#/!TlViTjmLDIWAUxzxsO:libera.chat/$TsVkMpUZ44hJl8S_qDYJoILEBQvlCeZXqtGbwLd6xow?via=libera.chat&via=matrix.org&via=gitter.im)

> You mentioned that there are problems using iosim because startTLS expects a "real life file descriptor"?  iosim has built-in support for startTLS though, so I wonder what happened there.
>
> src/twisted/test/test_amp.py tests AMP's startTLS usage and uses iosim
> any of the tests that use `connectedServerAndClient`

richvdh:
Sadly it's not looking like I'm going to have much time to work on this in the near future. My dayjob has taken me away from regular work on Twisted-based projects, and to be honest I've pretty much forgotten all the state I had on this when I worked on it two years ago.

I'd really love it if this could be merged as-is, despite the less-than-ideal test, since IMHO it fixes a real bug that bites users. That said, I do understand I'm not the one that will have to live with the headache of maintaining that test in future.

Anyway, in short, if anyone else has time to pick this up and see if the tests can be improved, it would be *amazing*.

glyph:
@richvdh Thanks for clarifying the status here. No worries that you don't have time to volunteer, it is, after all, volunteering :)

src/twisted/mail/smtp.py
```diff
@@ -950,6 +951,9 @@ def __init__(self, identity, logsize=10):
         self.code = -1
         self.log = util.LineLog(logsize)
 
+        # the reason we were unable to successfully send the message, if any.
+        self.failureReason: Optional[Failure] = None
```
glyph: This requires an `@ivar` if it's going to be made public, but I would suggest making it private, since it is exposed publicly at the relevant time in `self.result.errback(...)` below.

glyph:
```suggestion
        self._failureReason: Optional[Failure] = None
```

src/twisted/mail/smtp.py
```diff
@@ -968,6 +972,7 @@ def connectionLost(self, reason=protocol.connectionDone):
         """
         We are no longer connected
         """
+        self.failureReason = reason
```
glyph:
```suggestion
        self._failureReason = reason
```

src/twisted/mail/test/test_smtp.py
```diff
@@ -668,6 +670,47 @@ def check(ignored):
 
         return self.loopback(server, client).addCallback(check)
 
+    def test_ESMTPSenderFactory_TLSError(self):
```
glyph:
```suggestion
    def test_ESMTPSenderFactory_TLSError(self) -> None:
```

src/twisted/mail/smtp.py
```diff
@@ -1922,21 +1927,37 @@ def clientConnectionLost(self, connector, err):
...
 
     def _processConnectionError(self, connector, err):
+        # If the initial connection succeeded, but there was a later error
+        # (such as TLS validation failure), we're more interested in that than
+        # "Connection was closed cleanly" or whatever.
+        if (
+            self.currentProtocol is not None
+            and self.currentProtocol.failureReason is not None
+        ):
+            err = self.currentProtocol.failureReason
```
glyph:
```suggestion
        if (
            self.currentProtocol is not None
            and self.currentProtocol._failureReason is not None
        ):
            err = self.currentProtocol._failureReason
```

glyph: We are about to hand off the exception here, which means it would be a good time to *clear* `_failureReason` as well, otherwise it sticks around and potentially hangs on to a bunch of additional error-handling state we no longer need, including stack frames with copious circular references to the rest of the program, generating more work for the garbage collector.

src/twisted/mail/smtp.py
```diff
@@ -1922,21 +1927,37 @@ def clientConnectionLost(self, connector, err):
...
-            if err.check(error.ConnectionDone):
-                err.value = SMTPConnectError(-1, "Unable to connect to server.")
-            self.result.errback(err.value)
+            return
+
+        # do our best to wrap the error into something more meaningful.
+        exn = SMTPConnectError(
+            -1, "Unable to connect to SMTP server: " + str(err.value)
+        )
+        exn.__cause__ = err.value
```
glyph, line 1958: I appreciate the use of `__cause__` to chain the exception properly; we should probably make this a bit easier to do with `Failure`.

glyph (approved):
I think I agree with the reasoning of landing this even with a sub-optimal test. Whoever comes along to land it, please:

- [ ] fix up the public name to be private
- [ ] clear the exception when it's propagated to a `Deferred`
- [ ] file a follow-up to fix the test

src/twisted/mail/test/test_smtp.py
```diff
@@ -668,6 +670,47 @@ def check(ignored):
...
+        """
+        Attempting to connect to an ESMTP server which presents an invalid certificate
+        should produce a meaningful error.
```
adiroiban:
```suggestion
        Attempting to connect to an ESMTP server which presents an invalid certificate will trigger a failure that contains information that the failure was caused by the vertificate validation.
```

adiroiban:
Adding the todo pre-merge list here, to make it easier to see

- [ ] fix up the public name to be private
- [ ] clear the exception when it's propagated to a `Deferred`
- [ ] file a follow-up to fix the test

adiroiban:
Adding the todo pre-merge list here, to make it easier to see

- [ ] fix up the public name to be private
- [ ] clear the exception when it's propagated to a `Deferred`
- [ ] file a follow-up to fix the test

richvdh:
> Adding the todo pre-merge list here, to make it easier to see
> 
>     * [ ]  fix up the public name to be private
>     * [ ]  clear the exception when it's propagated to a `Deferred`

I believe both of these have happened, in https://github.com/twisted/twisted/pull/1645/changes/38660126fbc9ea125aff2df42730101bc456e9f2

>     * [ ]  file a follow-up to fix the test

Happy to do this myself, if it helps?

Any chance we can get this PR landed?

adiroiban:
Hi @richvdh . Thanks for the followup.

I see that this was approved by Glyph ... so I guess that we can merge this.

I will make sure all the automated tests pass and I will merge...

richvdh:
le sigh, looks like the test has bitrotted. I'll try and take a look.

glyph:
Thanks @richvdh , appreciate your persistence.

richvdh:
Looks like the test conflicted with #12506. The `__cause__` of the resultant `SMTPConnectError` was previously `Error([('SSL routines', '', 'certificate verify failed')])`; it is now `VerificationError(errors=[DNSMismatch(mismatched_id=DNS_ID(hostname=b'testdomain'))])`.

I've pushed an update which I hope will make the test pass.

glyph:
@richvdh looks like there are just a couple of very silly mypy errors left, one of which would be already addressed by https://github.com/twisted/twisted/pull/12577 if you could take a moment to review that one :)

Thanks!

glyph:
Let's see if the types actually check now…

glyph:
Okay, still a problem, per CI, but the errors are better now:

```
src/twisted/mail/smtp.py:955:9: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
src/twisted/mail/test/test_smtp.py:683:43: error: Argument 2 to "listenTCP" of "IReactorTCP" has incompatible type "Factory[DummyESMTP]"; expected "ServerFactory[_ProtoWithFactory]"  [arg-type]
src/twisted/mail/test/test_smtp.py:688:24: error: Need type annotation for "sentDeferred"  [var-annotated]
src/twisted/mail/test/test_smtp.py:703:13: error: "IAddress" has no attribute "host"  [attr-defined]
src/twisted/mail/test/test_smtp.py:703:40: error: "IAddress" has no attribute "port"  [attr-defined]
src/twisted/mail/test/test_smtp.py:711:9: error: No return value expected  [return-value]
```

richvdh:
Right, it's finally passing CI. @glyph PTAL?

adiroiban:
Hi Richard . Since Glyph has already approved this...and with them merge being blocked only by CI failures, I think that is safe to merge it, now that CI is ok.

Thanks and sorry for the delay.

I have enabled the automatic merge

glyph:
> Hi Richard . Since Glyph has already approved this...and with them merge being blocked only by CI failures, I think that is safe to merge it, now that CI is ok.

Yep, thanks for taking care of the merge.

PR #1648, opened 2021-08-01
[10246] Enable GitHub Actions concurency and cancel jobs for old PR code.

adiroiban:
## Scope and purpose

This tries to reduce the "wasted" time used on builds that executed for code that is no longer the head of a branch.

For example if you push some code and then you realized that you made a mistake , when you will push your fix all active test runs for the previous code are cancelled.

There is an exception for the "trunk" branch, for which we want to get a build for every push.

I have updated the coverage to no longer be always executed, but only executed when the job as not cancelled.
The coverage publish "cleanup" could have take more than 4 minutes and delay the whole cancellation process.

This feature is documented here 

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10246
* [NA] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10246

Enable GitHub Actions concurrency for the test workflow.
```


adiroiban:
This is a job that was first started.
Then while the job not done, I have pushed another change to this PR
The job as automatically cancelelled

https://github.com/twisted/twisted/pull/1648/checks?check_run_id=3214396992

The link to the whole run

https://github.com/twisted/twisted/actions/runs/1087631629

The run that has triggered the cancellation 

https://github.com/twisted/twisted/actions/runs/1087633024

.github/workflows/test.yaml
```diff
@@ -11,6 +11,15 @@ on:
     branches: [ trunk ]
 
 
+# Only have a run a single parallel for each branch.
+# Runs for trunk are queues.
+# Older runs for non-trunk branches are cancelled and the jobs are executed
+# only for the latest push to the branch.
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: ${{ github.head_ref != 'refs/heads/trunk' }}
```
altendky: It looks like `head_ref` gets set to `""` for...  manually...  re-run (?) trunk builds?  Just happened to be the first one I looked at.

https://github.com/twisted/twisted/runs/3274487611?check_suite_focus=true#step:7:152
```
      "head_ref": "",
```

Likewise though for what appears to be a merge-triggered build.

https://github.com/twisted/twisted/runs/3270940492?check_suite_focus=true#step:7:152
```
      "head_ref": "",
```

I think this check is wrong?

.github/workflows/test.yaml
```diff
@@ -110,7 +119,7 @@ jobs:
         tox ${{ matrix.trial-target }}
 
     - name: Prepare coverage
-      if: always() && contains(matrix['tox-env'], 'withcov')
+      if: ${{ !cancelled() && contains(matrix['tox-env'], 'withcov') }}
```
altendky, line 122: I'm always uncomfortable with the GitHub documentation's wording around `if:` and the conditions.  I thought `always()` was some weird thing and that without it you didn't get the run-even-if-a-previous-step-failed behavior.  But, I might be making this up so if you think it's OK, don't worry too much.  Just noting for completeness.

adiroiban: This is my understanding :)

`always` has some sort of cleanup semantics... the step is really always called.

`!cancelled` is similar to always... but the step is not executed on cancellation... 

I think that `!cancelled` is what we want.

We want to push coverage on test failures... but if the job was cancelled, we shouldn't bother pushing the coverage.

.github/workflows/test.yaml
adiroiban: Yes. The check is wrong. I think that `github.ref` is what we need :)
I have pushed an update.

altendky:
We do releases from commits on branches/in PRs, right?  Do we want any special consideration for that?  It would be a bummer to push a post-release-version-tweak commit or something while a release build is running and lose it.  Or, cancel it partway through and have some things published and others not.

adiroiban:
Good question :)

The releases are only done on tags push.

We don't release to PyPI on random commits.

The release builds should handled the same as PR builds.

If we someone end up with 2 builds for `twisted-21.7.0rc1` we want to cancel the old job.


here is a final release job https://github.com/twisted/twisted/runs/3179987482?check_suite_focus=true#step:6:153

here is a release candidate job https://github.com/twisted/twisted/runs/3142384008?check_suite_focus=true#step:6:153

They have different `github.ref` and will be part of 2 separate build groups

altendky (approved):
Thanks for the explanations.  And, I guess... maybe the work here too.  `:]`

adiroiban:
Thanks for the review... we will see how well it works.

PR #1653, opened 2021-08-03
10247: Make the request argument to twisted.web.iweb.IRenderable.render non-optional

wsanchez:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10247

```

(as long as you wish)
```

### DIFF
--- a/src/twisted/web/_stan.py
+++ b/src/twisted/web/_stan.py
@@ -23,13 +23,12 @@ cumbersome.
 
 
 from inspect import isgenerator, iscoroutine
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union
 from warnings import warn
 
 import attr
 
-if TYPE_CHECKING:
-    from twisted.web.template import Flattenable
+from ._flatten import Flattenable
 
 
 @attr.s(hash=False, eq=False, auto_attribs=True)
--- a/src/twisted/web/_stan.py
+++ b/src/twisted/web/_stan.py
@@ -23,12 +23,13 @@ cumbersome.
 
 
 from inspect import isgenerator, iscoroutine
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, TYPE_CHECKING, Union
 from warnings import warn
 
 import attr
 
-from ._flatten import Flattenable
+if TYPE_CHECKING:
+    from ._flatten import Flattenable
 
 
 @attr.s(hash=False, eq=False, auto_attribs=True)
--- a/src/twisted/web/resource.py
+++ b/src/twisted/web/resource.py
@@ -25,8 +25,10 @@ from twisted.python.compat import nativeString
 from twisted.python.reflect import prefixedMethodNames
 from twisted.python.components import proxyForInterface
 
+from twisted.web._flatten import Flattenable
 from twisted.web._responses import FORBIDDEN, NOT_FOUND
 from twisted.web.error import UnsupportedMethod
+from twisted.web.iweb import IRequest
 
 
 class IResource(Interface):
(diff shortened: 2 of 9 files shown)


src/twisted/web/resource.py
```diff
@@ -72,7 +74,7 @@ def putChild(path, child):
         @type path: C{bytes}
         """
 
-    def render(request):
+    def render(request: IRequest) -> Flattenable:
```
wsanchez, line 77: Added this type hint without doing the whole file just to ensure that the signatures and types match between `IRenderable` and `Resource`.

wsanchez:
@mthuurne Says it does make sense for `IRenderable` to allow a `None` request, and that it is used without requests being a thing by such software as `pydoctor`.

I thought I ran into Big Problems™ trying to have different views on this in the `render` on `IRenderable` and `Resource`, but I'll look again (in Klein, in particular) before we proceed here.

mthuurne:
> @mthuurne Says it does make sense for `IRenderable` to allow a `None` request, and that it is used without requests being a thing by such software as `pydoctor`.

Yes, there are scenarios where it is useful to render templates to HTML outside of a web server, for example when generating HTML files like `pydoctor` does.

To be precise, `pydoctor` doesn't call `render()` directly, but it does call `flattenString()` with `None` for the `request` argument, which eventually ends up being passed as the `request` argument to `render()`.

> I thought I ran into Big Problems™ trying to have different views on this in the `render` on `IRenderable` and `Resource`, but I'll look again (in Klein, in particular) before we proceed here.

I think it's fine if `IResource` does require `request` to not be `None`, since resources are specific to handling HTTP requests, unlike templates.

src/twisted/web/_stan.py
```diff
@@ -23,13 +23,13 @@
...
 from inspect import isgenerator, iscoroutine
-from typing import TYPE_CHECKING, Dict, List, Optional, Union
+from typing import Dict, List, Optional, TYPE_CHECKING, Union
 from warnings import warn
 
 import attr
 
 if TYPE_CHECKING:
-    from twisted.web.template import Flattenable
+    from ._flatten import Flattenable
```
mthuurne, line 32: Why this change? `Flattenable` is explicitly exported from `twisted.web.template` via `__all__`. Importing it from its public location should be more future proof than importing it from its implementation module.

src/twisted/web/resource.py
mthuurne: Must they match though? `twisted.web.resource.Resource` implements `IResource` but does not implement `IRenderable`. They are obviously designed to work together, but when using delegation rather than inheritance, it shouldn't be a problem if `IResource.render()` has a mandatory `request` parameter and `IRenderable.render()` has an optional `request` parameter.

mthuurne (changes requested):
The implementation of the patch looks OK, although I think the import of `Flattenable` would be better left as-is.

I don't agree with the goal of the patch though: it seems unnecessarily restrictive to require a request object for every use of templates.

Since the template implementation just passes through the request, an alternative to `Optional[IRequest]` might be to give `IRenderable` a generic type argument for the request type.

src/twisted/web/_stan.py
graingert: That usually makes cycles worse

graingert: I think using `from . import _flatten` and then `x: "_flatten.Flattenable"` is preferable to `if TYPE_CHECKING`

mthuurne: `mypy` doesn't care about import cycles, so that's only a factor if you want to remove the `if TYPE_CHECKING`. Even if you import from the implementation module, you'll still have to keep import cycles in mind, so I don't think removing `if TYPE_CHECKING` will simplify the code.

PR #1655, opened 2021-08-04
10248 configure isort pre-commit hook

graingert:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10248

```

(as long as you wish)
```


graingert:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -67,14 +67,12 @@ build-backend = "setuptools.build_meta"
...
 
 [tool.isort]
-line_length = 88
+profile = "black"
 sections = "FUTURE,STDLIB,ZOPE,OPENSSL,THIRDPARTY,FIRSTPARTY,LOCALFOLDER"
 default_section = "THIRDPARTY"
 no_lines_before = "LOCALFOLDER"
 known_first_party = "twisted"
 known_zope = "zope"
 known_openssl = "OpenSSL"
```
graingert, line 76:
```suggestion
sections = "FUTURE,STDLIB,ZOPE,OPENSSL,THIRDPARTY,FIRSTPARTY,LOCALFOLDER,REACTOR"
default_section = "THIRDPARTY"
no_lines_before = "LOCALFOLDER"
known_first_party = "twisted"
known_zope = "zope"
known_openssl = "OpenSSL"
known_reactor = "twisted.internet.reactor"
```

adiroiban:
Thanks for working on this.

I think that this is much needed change for the project.

------------

The changes looks good, with the exception of the newly created module.

It would have been nice to have a quick description of this error and the solution in the PR description.

I am not sure I fully understand to main cause that trigger the generation of a new module.

I am a bit worries that this import error was only detected by accident via pydoctor.

Can we also use the same import method as pydoctor in one of twisted own tests to make sure we cover this import path?

To keep the size of this PR low and avoid further changes, can this PR be merged with an `# isort: skip` so that we don't need to create a new module?

I am a bit worried that this change might trigger side-effects with project using Twisted.

Also, I think that if regressions are found due to import order, it's important to have available a quick fix option so that we don't have always do a code refactoring.

---------

Now, refactoring the code so that the import order is not that import is a good thing.

But, with this new refactoring I don't really see the purpose of twisted.web.template.
I now see it as a public gatekeeper to some private API from twisted_utils.py ... maybe `_template_util.py` should be just be named `_template.py`

-----------

After this lands, I think that it would help to send an email to the mailing list to let twisted users know that we made some changes to the import order and they should run a check of the new code now.

 Maybe even trigger a 21.8.0 release, with a pre-release on pypi. 

---------

I also tried to read the info from https://github.com/twisted/twisted/pull/1132 
I saw that in the end the PR was merged with isort configuration...but isort was not really used in the project.

-------------

I don't know if `narrativedocs` depends on `apidocs`.
`narrativedocs` will trigger an internal `apidocs` build, as the API documentation is now integrated into narrative documentation... to make it work with Read The Docs.

Maybe we can stop building `tox -e apidocs` in CI.
I left the `apidocs` tox env for the case in which you want a quicker debugging / testing of the API docs content. 

`$ tox -p2 -e narrativedocs,apidocs` works for me on trunk.

From the error of the `apidocs` build (https://github.com/twisted/twisted/runs/3260163989#step:5:80) I don't see any hint that the error is caused by narrativedocs 

I have done a few more (7 times while reviewing the module rename) runs `$ tox -p2 -e narrativedocs,apidocs` runs on my laptop and all looks ok.

graingert:
> I have done a few more (7 times while reviewing the module rename) runs $ tox -p2 -e narrativedocs,apidocs runs on my laptop and all looks ok.

the error occurs for me with `tox -p2 -re narrativedocs,apidocs`

graingert:
I've tried another re-run and without the tox depends change locally and it does indeed pass every time

graingert:
> Also, I think that if regressions are found due to import order, it's important to have available a quick fix option so that we don't have always do a code refactoring.

the pyproject.toml lists an allowlist of un-isorted modules. But really it's better to merge modules with a cyclic dependency

adiroiban (approved):
All good. Thanks.

I have only reviewed the manual changes... and not the automated changes made by isort.

Since this PR has no description, I assume that the purpose for using isort is to get consistent code, similar to `black`.

I don't have a better name for the new private module.
Since this is private module we can change it at any time.
Also, this PR is only moving code to a private module without changing the public API.

Even if things are not perfect now, since this is private we can update it later without generating any extra breakage.

The coverage is not 100% but that should be ok for this PR.

Thanks again @graingert for the hard work!

adiroiban:
One minor comment.
It would help to have a release note for this change.

Since the import order was changed, 3rd party users, like pydoctor, might end up with import errors.

This is why I plan to do a 21.8.0rc1 release soon to get some more feedback about this change.

Thanks!

adiroiban (approved):
All good. Thanks.

PR #1656, opened 2021-08-06
10249 Use enviroment markers for extra all

McSinyx:
## Scope and purpose

This will make it valid to install every extra on any platform.  This sounds odd but it is a [criteria we are experimentally mandate in IPWHL](https://man.sr.ht/~cnx/ipwhl/#dependencies), a WIP downstream repository for Python packages.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10249
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews now.

```

(as long as you wish)
```

### DIFF
--- a/docs/installation/howto/optional.rst
+++ b/docs/installation/howto/optional.rst
@@ -42,11 +42,7 @@ The following optional dependencies are supported:
 
 * **serial** - the `pyserial`_ package to work with serial data.
 
-* **all_non_platform** - installs **tls**, **conch**, **soap**, and **serial** options.
-
-* **macos_platform** - **all_non_platform** options and `pyobjc`_ to work with Objective-C apis.
-
-* **windows_platform** - **all_non_platform** options and `pywin32`_ to work with Windows's apis.
+* **all** - installs **tls**, **conch**, **soap**, and **serial** options.
 
 * **http2** - packages needed for http2 support.
 
--- a/docs/installation/howto/optional.rst
+++ b/docs/installation/howto/optional.rst
@@ -15,62 +15,109 @@ For a deeper explanation of what optional dependencies are and how they are decl
 
 The following optional dependencies are supported:
 
-* **dev** - packages that aid in the development of Twisted itself.
+tls
+   Packages that are needed to work with TLS:
 
-  * `pyflakes`_
-  * `twisted-dev-tools`_
-  * `python-subunit`_
-  * `Sphinx`_
-  * `TwistedChecker`_, only available on python2
-  * `pydoctor`_, only available on python2
+   * pyOpenSSL_
+   * service_identity_
+   * idna_
 
+conch
+   Packages for working with conch/SSH:
 
-* **tls** - packages that are needed to work with TLS.
+   * pyasn1_
+   * cryptography_
+   * bcrypt_
+   * appdirs_
 
-  * `pyOpenSSL`_
-  * `service_identity`_
-  * `idna`_
+conch_nacl
+   ``conch`` options and PyNaCl_ to support Ed25519 keys
+   on systems with OpenSSL < 1.1.1b.
 
-* **conch** - packages for working with conch/SSH.
+serial
+   The pyserial_ package to work with serial data.
+   On Windows, this also requires pywin32_.
 
-  * `pyasn1`_
-  * `cryptography`_
+http2
+   Packages needed for HTTP/2 support:
 
-* **conch_nacl** - **conch** options and `PyNaCl`_ to support Ed25519 keys on systems with OpenSSL < 1.1.1b.
+   * h2_
+   * priority_
 
-* **soap** - the `SOAPpy`_ package to work with SOAP.
+contextvars
+   The contextvars_ backport package to provide `context variables`_ support
+   for Python versions before 3.7.
 
-* **serial** - the `pyserial`_ package to work with serial data.
+all
+   Extras ``tls``, ``conch``, ``serial``, ``http2``, ``contextvars``,
+   and platform-specific interfacing such as pyobjc_ for Objective-C on macOS
+   and pywin32_ for Windows.
 
-* **all** - installs **tls**, **conch**, **soap**, and **serial** options.
+dev_release
+   Packages used in Twisted's release process:
 
-* **http2** - packages needed for http2 support.
+   * towncrier_
+   * pydoctor_
+   * Sphinx_, sphinx-rtd-theme_ and readthedocs-sphinx-ext_
 
-  * `h2`_
-  * `priority`_
+dev
+   Packages that aid in the development of Twisted itself,
+   including those in ``all`, ``dev_release`` and the following:
 
-* **contextvars** - the `contextvars`_ backport package to provide contextvars support for Python versions before 3.7.
+   * pyflakes_
+   * python-subunit_
+   * twistedchecker_
+   * coverage_
 
+mypy
+   Type checking facilities, including ``dev`` packages,
+   ``conch_nacl``, as well as the following:
 
-.. _pip: https://pip.pypa.io/en/latest/quickstart.html
-.. _TwistedChecker: https://pypi.python.org/pypi/TwistedChecker
-.. _pyflakes: https://pypi.python.org/pypi/pyflakes
-.. _twisted-dev-tools: https://pypi.python.org/pypi/twisted-dev-tools
-.. _python-subunit: https://pypi.python.org/pypi/python-subunit
-.. _Sphinx: https://pypi.python.org/pypi/Sphinx/1.3b1
-.. _pydoctor: https://pypi.python.org/pypi/pydoctor
-.. _pyOpenSSL: https://pypi.python.org/pypi/pyOpenSSL
-.. _service_identity: https://pypi.python.org/pypi/service_identity
-.. _pyasn1: https://pypi.python.org/pypi/pyasn1
-.. _cryptography: https://pypi.python.org/pypi/cryptography
+   * mypy_
+   * mypy-zope_
+   * types-setuptools_
+   * types-pyOpenSSL_
+
+.. _pip: https://pip.pypa.io
+.. _python packaging tutorial:
+   https://packaging.python.org/en/latest/installing.html#examples
+.. _setuptools documentation:
+   https://setuptools.readthedocs.io/en/stable/userguide/dependency_management.html#optional-dependencies
+
+.. _pyOpenSSL: https://pypi.org/project/pyOpenSSL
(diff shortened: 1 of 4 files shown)


docs/installation/howto/optional.rst
```diff
@@ -40,11 +40,7 @@ The following optional dependencies are supported:
 
 * **serial** - the `pyserial`_ package to work with serial data.
 
-* **all_non_platform** - installs **tls**, **conch**, **soap**, and **serial** options.
-
-* **macos_platform** - **all_non_platform** options and `pyobjc`_ to work with Objective-C apis.
-
-* **windows_platform** - **all_non_platform** options and `pywin32`_ to work with Windows's apis.
+* **all** - installs **tls**, **conch**, **soap**, and **serial** options.
```
graingert: there's a bunch of very outdated advice in this file eg "* `TwistedChecker`_, only available on python2"

that I'm going to gently suggest you also fix

src/twisted/newsfragments/10249.feature
```diff
@@ -0,0 +1,5 @@
+The ``all_non_platform`` extra has been renamed to ``all``,
```
graingert, line 1: all currently pulls in `    %(test)s` which is probably not what users of twisted want

graingert: eg if you do `pip install twisted[all]` it will also install `cython-test-exception-raiser ~= 1.0` which nobody needs and I'll get angry letters about it

graingert (changes requested):

src/twisted/newsfragments/10249.feature
McSinyx: Apparently this has not been documented, need we add a news fragment for it or we can pretend that the `test` extra never existed?

> The plan is to remove the `test` extra requirement and skip tests that depend on then [sic]

docs/installation/howto/optional.rst
McSinyx: I overhauled that page with updated links and package listing.  Should there be a news fragment for it?

src/twisted/newsfragments/10249.feature
```diff
@@ -0,0 +1,5 @@
...
+with the former name being a compatibility alias.  Platform-specific extras
+such as ``macos_platform`` and ``windows_platform`` have also been deprecated;
```
twm, line 3: Adding platform markers is good, but I don't that we should eliminate the platform integration extras as separate things. A user shouldn't have to opt in to *every* Twisted dependency just to get platform integration.

McSinyx: I don't disagree, just that currently that's how they're defined:

https://github.com/twisted/twisted/blob/5568fa24d60b470658b6cfc48be338f06cce57f0/setup.cfg#L96-L112

I don't have any problem defining e.g. `objc` and `win32` extras for just the `pyobjc-*` and `pywin32` packages respectively though, but someone will need to tell me to do it because I use neither Windows, macOS nor Twisted personally so I am unaware of how it is usually used.

twm: Oh, that's super weird. In that case your changes don't seem to represent a compatibility break so 👍

McSinyx:
Gentle ping!

src/twisted/newsfragments/10249.feature
adiroiban: That comment about skipping test is more about `pyHamcrest >= 1.9.0` 

The `cython-test-exception-raiser` is just an internal testing package that was created so that we can have non binary wheels for the main Twisted package.

setup.cfg
```diff
@@ -93,34 +94,36 @@ http2 =
...
+    pyobjc-framework-Cocoa; platform_system == "Darwin"
+    pywin32 != 226; platform_system == "Windows"
     %(tls)s
     %(conch)s
     %(serial)s
     %(http2)s
     %(contextvars)s
 
+all_non_platform =
+    %(all)s
```
adiroiban: Thanks for the update. I think that this is an important one.

I think that all_non_platform should be kept as it is (with test removed) and not install pywin32.

and maybe create something like `_platform_specific` and add it to the new `all` toge

```
all_non_platform =
  AS_BEFORE

_platform_specific =
    pyobjc-core; platform_system == "Darwin"
    pyobjc-framework-CFNetwork; platform_system == "Darwin"
    pyobjc-framework-Cocoa; platform_system == "Darwin"
    pywin32 != 226; platform_system == "Windows"

all =
    %(all_non_platform)s
    %(_platform_specifics)s
```

Does it make sense?

The concern raised by Thomas is that maybe you are already calling `pip twisted[all_non_platform]` on pypy on windows or macos and you might end up with a failure.

---------

We don't have pypy test runs on macOS or Windows (I think that we should have at least one) so I don't know if that is the case.

But I think that is best to play it safe and keep `all_non_platform` is it was before.

Then we can add `all` or `standard` or any new extra :)

McSinyx: Sorry for the late reply, do you think it would be preferable to use the `macos_platform` and `macos_platform` extra name or `_platform_specific` should be the way to go?

> [installing] on pypy on windows or macos and you might end up with a failure.

Yup pyobjc and pywin32 looks like CPython extension modules, I think I'll go head and add `platform_python_implementation == "CPython"` to the marker.

glyph:
Merging changes back to the branch appears to be disabled so I can't update it for current trunk.

McSinyx:
I've rebased it on top of trunk.

docs/installation/howto/optional.rst
```diff
@@ -15,63 +15,103 @@ For a deeper explanation of what optional dependencies are and how they are decl
...
-.. _pyserial: https://pypi.python.org/pypi/pyserial
-.. _pyobjc: https://pypi.python.org/pypi/pyobjc
-.. _pywin32: https://pypi.python.org/pypi/pywin32
-.. _`setuptools documentation`: https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies
-.. _`python packaging tutorial`: https://packaging.python.org/en/latest/installing.html#examples
-.. _idna: https://pypi.python.org/pypi/idna
-.. _h2: https://pypi.python.org/pypi/h2
-.. _priority: https://pypi.python.org/pypi/priority
-.. _contextvars: https://pypi.org/project/contextvars/
+dev_release
```
twm: `dev_release` and `dev` should probably be last in this list, as end users won't care about them, only Twisted developers.

twm (approved):
I had a minor note, but even so this looks like a real improvement and I'd like to merge it. Unfortunately it has drifted from trunk once more. Could you merge/rebase again?

docs/installation/howto/optional.rst
McSinyx: Both done!

adiroiban:
The changes looks good.

We have the deprecation info in the release notes. That is great.

We also raises deprecation warnings for things that are deprecated... so that people will that don't read the release notes will know what was changed.

Is there an easy way in which a deprecation error can be raised for using a specific extra?

I guess that we can try to keep them deprecated and without removal for as long as we can.

I guess that you can include a package that once installed it will trigger a deprecation warning...but looks complicated.

---------------

so @graingert any comments on the current PR?

Can it be merged?

Thanks!

twm:
@adiroiban I think the hold-up here is entirely that Twisted maintainers can't merge/rebase it easily, so it drifts out of sync with trunk. (Well, and now there are conflicts.) I'd have merged it a month ago otherwise.

setup.cfg
```diff
@@ -100,35 +101,37 @@ http2 =
...
```
adiroiban: To keep backward compatibility as suggested by Thomas

I know that I suggested to remove tests... but is best to keep it backward compatible, when possible.

We will leave "all_non_platform" as it was and let it rest in peace :)

And all the documentation is updated to use "all"


```suggestion
# Deprecated. Should not be used for new deployments.
all_non_platform =
    %(test)s
    THE-SAME-LIST-AS-BEFORE
```

setup.cfg
```diff
@@ -64,6 +64,7 @@ dev_release =
 
 ; All the extra tools used to help with the development process.
 dev =
+    %(test)s
```
adiroiban, line 68: Maybe also add all to dev. in this way, you can just install dev and you know that you are ready to rock :)

```suggestion
    %(all)s
    %(test)s
```

adiroiban (commented):
Many thanks for the update.

As commented by Thomas over gitter, all_non_platform should also install test as it was doing in the past to keep backward compatibility.

I left more comments inline.

setup.cfg
McSinyx: Thanks for the clarification, I've updated the patches.

setup.cfg
McSinyx: I've added this along with the change log.

adiroiban:
@graingert ... so. Thomas. Anything bad with this PR? Are you happy to see it merged?

Right now the PR is blocked as you have previously rejected  the changes.

Thanks!

docs/installation/howto/optional.rst
```diff
@@ -15,66 +15,109 @@ For a deeper explanation of what optional dependencies are and how they are decl
...
 
-* **dev** - packages that aid in the development of Twisted itself.
+tls
+   Packages that are needed to work with TLS:
 
-  * `pyflakes`_
-  * `twisted-dev-tools`_
-  * `python-subunit`_
-  * `Sphinx`_
-  * `TwistedChecker`_, only available on python2
```
glyph, line 24: Not the responsibility of this change, but, I think that we should *not* actually re-enumerate all dependencies in this document any more. Just list the extras, and let the package metadata speak for itself if people want to know specifics. We can't necessarily keep this up to the minute up-to-date if someone adds a transitive dependency, for example.

glyph (changes requested):
Sorry to see that this has drifted out of sync with trunk *yet again*, probably due to the big packaging overhaul. Hopefully that means that this will be easier to do, if somewhat annoying that it all needs to be done again.

glyph:
Note that you can ask the robot for a re-review at any time by saying "p-l-e-a-s-e review" without the dashes.

McSinyx:
Closing as bitrotten and no longer needed, cheers!

glyph:
@McSinyx Thank you for your contribution, and I'm sorry we couldn't fully address it while it was relevant. Glad that it's no longer needed though!

PR #1662, opened 2021-09-14
10260: Avoid confusing combined traceback

taliastocks:
In the current release, if an inlineCallback generator returns to a deferred callback that throws an exception, the traceback is confusing, e.g.:

```
  File "/mnt/mesos/sandbox/.pex/install/Twisted-20.3.0-cp37-cp37m-manylinux1_x86_64.whl.1128a7e2b98487eb88697f10720d146583a69992/Twisted-20.3.0-cp37-cp37m-manylinux1_x86_64.whl/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
StopIteration: Response(code=200, message=b'OK')

During handling of the above exception, another exception occurred:

  File "/mnt/mesos/sandbox/__main__-D229445.750890.pex/infra_compute/proxy/proxy.py", line 291, in _proxy_request
    simplified_response = yield self.proxy.process_request(simplified_request)
... # the real traceback we care about
```

This happens because the `status.deferred.callback()` is called from within the `StopIteration` exception handler. Python assumes that the `StopIteration` is the real exception we care about because that exception was being handled while the second exception was thrown. However, we don't care about the `StopIteration` exception, as it is only used to transmit a return value from the `inlineCallback` generator.

This can be avoided by first exiting the exception handler, then calling `status.deferred.callback()`. Python will no longer incorrectly tack on the `StopIteration` traceback on the top if there is an exception in the callback.

## Scope and purpose

This PR prevents Python from tacking on a confusing `StopIteration` traceback on top of the relevant traceback when a deferred callback throws an exception after being called with the result of an `inlineCallback` generator.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10260

```

Author: taliastocks,twm
Reviewer: adiroiban
Fixes: ticket:10260, ticket:9792

This PR prevents Python from tacking on a confusing `StopIteration` traceback on top of the relevant traceback when a deferred callback throws an exception after being called with the result of an `inlineCallbacks` generator.

Previously, if an inlineCallbacks generator returns to a deferred callback that throws an exception, the traceback is confusing, e.g.:

      File "/mnt/mesos/sandbox/.pex/install/Twisted-20.3.0-cp37-cp37m-manylinux1_x86_64.whl.1128a7e2b98487eb88697f10720d146583a69992/Twisted-20.3.0-cp37-cp37m-manylinux1_x86_64.whl/twisted/internet/defer.py", line 1418, in _inlineCallbacks
        result = g.send(result)
    StopIteration: Response(code=200, message=b'OK')
    
    During handling of the above exception, another exception occurred:
    
      File "/mnt/mesos/sandbox/__main__-D229445.750890.pex/infra_compute/proxy/proxy.py", line 291, in _proxy_request
        simplified_response = yield self.proxy.process_request(simplified_request)
    ... # the real traceback we care about

This happens because the `status.deferred.callback()` is called from within the `StopIteration` exception handler. Python assumes that the `StopIteration` is the real exception we care about because that exception was being handled while the second exception was thrown. However, we don't care about the `StopIteration` exception, as it is only used to transmit a return value from the `inlineCallback` generator.

This can be avoided by first exiting the exception handler, then calling `status.deferred.callback()`. Python will no longer incorrectly tack on the `StopIteration` traceback on the top if there is an exception in the callback.
```


taliastocks:
docs/readthedocs.org:twisted is throwing
```
AttributeError: 'ZopeInterfaceClass' object has no attribute 'append'
```
which doesn't seem relevant to my change.

tristanlatr:
The error `AttributeError: 'ZopeInterfaceClass' object has no attribute 'append'` has been fixed in the newer version of `pydoctor`, so that should not be an issue anymore.

exarkun:
Thanks.  This will be a great misbehavior to fix.  Can you also add a unit test that demonstrates the problem and that the changes here fix it?

glyph:
@taliastocks Sorry for the delays in review here, but would you mind updating the PR to resolve conflicts?

twm:
I have merged forward and added tests based on the reproducer attached to [Trac #9792](https://twistedmatrix.com/trac/ticket/9792).

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
 
         self.assertIn("inlineCallbacks", str(self.assertRaises(TypeError, _noYield)))
 
+    def test_internalDefGenReturnDoesntLeak(self):
+        """
+        When one inlineCallbacks calls another, the internal L{_DefGen_Return}
+        flow control exception doesn't leak into tracebacks captured in the
+        caller.
+        """
```
adiroiban:
```suggestion
    def test_internalDefGenReturnValueDoesntLeak(self):
        """
        When one inlineCallbacks calls another, the internal L{_DefGen_Return}
        flow control exception doesn't leak into tracebacks captured in the
        caller.
        
        This is the test in which the inner functions use L{defer.retrunValue},
        which is the original python 2.7 method.
        """
```

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+        clock = task.Clock()
+
+        @inlineCallbacks
+        def _returns():
+            yield task.deferLater(clock, 0)
+            returnValue(6)
```
adiroiban:
```suggestion
        def _returns():
            """
            This is the inner function using returnValue.
            """
            yield task.deferLater(clock, 0)
            returnValue('actual-value-not-used-for-the-test')
```

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+            try:
+                yield _returns()
+                raise TerminalException("boom")
+            except TerminalException:
+                return traceback.format_exc()
+
+        d = _raises()
+        clock.advance(0)
+        tb = self.successResultOf(d)
+        self.assertNotIn("_DefGen_Return", tb)
```
adiroiban, line 371: can we also add a "positive" assertion about how to traceback should look like ?

If the traceback is empty, the test will still pass...but maybe we have other tests that will fail...so no need for extra assertions here.

```suggestion
        # The internal function is in the traceback.
        self.assertNotIn("_DefGen_Return", tb)
        # No other extra exception is in the traceback.
        self.assertNotIn(
            "During handling of the above exception, "
            "another exception occurred",
            tb)
        # Our targeted exception is in the traceback
        self.assertIn('test_defgen.TerminalException: boom returnValue', tb)
```

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+        @inlineCallbacks
+        def _returns():
+            yield task.deferLater(clock, 0)
+            returnValue(6)
+
+        @inlineCallbacks
+        def _raises():
+            try:
+                yield _returns()
+                raise TerminalException("boom")
```
adiroiban: Very minor comment to make the exaction unique and make it easier to trace the origin of this exception, in the case in which things go very bad and we have random side effects

```suggestion
                raise TerminalException("boom returnValue")
```

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+        clock.advance(0)
+        tb = self.successResultOf(d)
+        self.assertNotIn("_DefGen_Return", tb)
+
+    def test_internalStopIterationDoesntLeak(self):
+        """
+        When one inlineCallbacks calls another, the internal L{StopIteration}
+        flow control exception generated when the inner generator returns
+        doesn't leak into tracebacks captured in the caller.
+        """
```
adiroiban:
```suggestion
        """
        When one inlineCallbacks calls another, the internal L{StopIteration}
        flow control exception generated when the inner generator returns
        doesn't leak into tracebacks captured in the caller.
        
        This is similar to C{test_internalDefGenReturnValueDoesntLeak}
        but the inner function uses the "normal" return statemement, and
        not the C{returnValue} helper.
        """
```

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+            try:
+                yield _returns()
+                raise TerminalException("boom")
+            except TerminalException:
+                return traceback.format_exc()
+
+        d = _raises()
+        clock.advance(0)
+        tb = self.successResultOf(d)
+        self.assertNotIn("StopIteration", tb)
```
adiroiban, line 409: Maybe also add the extra assertions and comments like in the previous test

src/twisted/test/test_defgen.py
```diff
@@ -338,6 +339,58 @@ def _noYield():
...
+        @inlineCallbacks
+        def _returns():
+            yield task.deferLater(clock, 0)
+            return 6
+
+        @inlineCallbacks
+        def _raises():
+            try:
+                yield _returns()
+                raise TerminalException("boom")
```
adiroiban:
```suggestion
                raise TerminalException("boom normal return")
```

adiroiban (approved):
It looks good. @twm see my comments and feel free to merge this in whatever form you prefer.

Thanks @taliastocks  and @twm for the update.

The tests were a bit hard to read at first.

I left some feeback on how maybe they can be easier to read.

Thanks!

src/twisted/test/test_defgen.py
twm: This is an excellent suggestion; I've applied your changes and added the additional assertions. I also changed the patch to use camelCase identifiers.

taliastocks:
Thanks everyone! Sorry I've been unreachable!

PR #1665, opened 2021-09-19
9555 sarhamam replace bytes with bytes

Sarhamam:
When replacing bytes, encode os.linesep which is str, and prefix \r\n with b


Sarhamam:
Check release/publish status is failing on `lists of files in version control and sdist do not match!` ,and read the docs build also is failing. no idea how to fix it,  can anyone help?

src/twisted/protocols/test/test_ftp
```diff
@@ -0,0 +1,33 @@
+import mock
+import twisted
+from twisted.trial import unittest
+from twisted.protocols.ftp import ASCIIConsumerWrapper
+
+class ASCIIConsumerWrapperTestCase(unittest.TestCase):
+    """ Test case for ASCIIConsumerWrapper, tests __init__ and write function ""
+    @mock.patch("twisted.protocols.ftp.os")
```
twm, line 8: Happily, there's no need to do this mocking &mdash; we run our tests on both Linux and Windows, so you'll get both possible `os.linesep` values.

In Twisted tests we avoid mocking as much as possible. If we were to patch something like this we'd use `SynchronousTestCase.patch()` rather than the `mock` package as the latter is a box full of footguns.

Please remove the mocking!

src/twisted/protocols/test/test_ftp
```diff
@@ -0,0 +1,33 @@
...
+    def test_init(self, os_mock):
+        """ Test non supported platform, and make sure that for '\r\n' we're not wrapping the write function.
+        cons_mock = mock.MagicMock()
```
twm, line 11: Don't use a mock for this &mdash; especially not a `MagicMock`, the footgunniest of mocks. I'd suggest defining a class right here in the test, but it looks like `twisted.protocols.ftp.FileConsumer` will do, like:

```py
buf = io.BytesIO()
wrapper = ASCIIConsumerWrapper(FileConsumer(buf))

wrapper.write(f"test{os.linesep}".encode())

self.assertEqual(b'\r\n', buf.getvalue())
```

twm (changes requested):
Hi @Sarhamam, there are a few changes are necessary before we can merge this:

1. Rename the test file with a `.py` extension so that it gets run.
2. Update the `assert` and `if` in `ASCIIConsumerWrapper.__init__` to compare with `b'\r\n'` rather than a `str`.
3. Address my other comments on the tests.

PR #1666, opened 2021-10-11
10266: Add support for SSH extension negotiation

cjwatson:
## Scope and purpose

This is defined in RFC 8308.  RFC 8332 ("Use of RSA Keys with SHA-256 and SHA-512") says that servers that accept rsa-sha2-* signatures for client authentication SHOULD implement this, so it makes sense to do this first.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10266
* [x] I have created a newsfragment in src/twisted/conch/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))

```

Author: cjwatson
Fixes: ticket:10266

This is defined in RFC 8308.  RFC 8332 ("Use of RSA Keys with SHA-256
and SHA-512") says that servers that accept rsa-sha2-* signatures for
client authentication SHOULD implement this, so it makes sense to do
this first.
```


cjwatson:
@adiroiban Would you mind having a look at this?  It's a prerequisite for `rsa-sha2-*` signature algorithm support.  Thanks!

src/twisted/conch/ssh/transport.py
```diff
@@ -525,11 +534,22 @@ def sendKexInit(self):
...
+        supportedKeyExchanges = list(self.supportedKeyExchanges)
+        # Advertise extension negotiation (RFC 8308, section 2.1).  At
+        # present, the Conch client processes the "server-sig-algs"
+        # extension (section 3.1), and the Conch server sends that but
+        # ignores any extensions sent by the client, so strictly speaking at
+        # the moment we only need to send this in the client case; however,
+        # there's nothing to forbid the server from sending it as well, and
+        # doing so makes things easier if it needs to process extensions
+        # sent by clients in future.
+        supportedKeyExchanges.append(b"ext-info-c" if self.isClient else b"ext-info-s")
```
adiroiban: Very minor comment.
Can we have these values as private constants ?
This should signal that they are not random string bytes but that there is a specific semantic to each value.
And using a constant should prevent someone using them as unicode/text.

Maybe something like
```
_EXT_INFO_C = b"ext-info-c"
```

src/twisted/conch/ssh/transport.py
```diff
@@ -873,11 +897,16 @@ def ssh_KEXINIT(self, packet):
...
+        if (
+            None
+            in (
+                self.kexAlg,
+                self.keyAlg,
+                self.outgoingCompressionType,
+                self.incomingCompressionType,
+            )
+            # RFC 8308, section 2.2
+            or self.kexAlg in (b"ext-info-c", b"ext-info-s")
```
adiroiban: Just more comments so that you don't have to open the RFC and see what 2.2 is :)
```suggestion
            # RFC 8308, section 2.2
            # Most probably a broken implementation on the remote peer,
            # as the extension markers should not be returned as the negotiated algorithm.
            or self.kexAlg in (b"ext-info-c", b"ext-info-s")
```

src/twisted/conch/ssh/transport.py
```diff
@@ -418,6 +419,11 @@ class SSHTransportBase(protocol.Protocol):
         passed to L{sendPacket} but could not be sent because it is not legal
         to send them while a key exchange is in progress.  When the key
         exchange completes, another attempt is made to send these messages.
+
+    @ivar _supportsExtensionNegotiation: a boolean indicating whether the
+        other side supports RFC 8308 extension negotiation.
+
+    @ivar otherExtensions: a dict of extensions supported by the other side.
```
adiroiban: Maybe 
```suggestion
    @ivar peerExtensions: a dict of extensions supported by the other side of the connection.
```

src/twisted/conch/ssh/transport.py
```diff
@@ -418,6 +419,11 @@ class SSHTransportBase(protocol.Protocol):
         passed to L{sendPacket} but could not be sent because it is not legal
         to send them while a key exchange is in progress.  When the key
         exchange completes, another attempt is made to send these messages.
+
+    @ivar _supportsExtensionNegotiation: a boolean indicating whether the
+        other side supports RFC 8308 extension negotiation.
```
adiroiban: I had to read the docs twice to see what "other" means.
 
I think that `peer` is the generic terminology in twisted to refer to the other side of the connection.

Maybe 

```suggestion
    @ivar _peerSupportsExtensions: a boolean indicating whether the
        other side supports RFC 8308 extension negotiation.
```

src/twisted/conch/ssh/transport.py
```diff
@@ -1036,6 +1087,20 @@ def sendDisconnect(self, reason, desc):
         )
         self.transport.loseConnection()
 
+    def sendExtInfo(self, extensions):
+        """
+        Send an RFC 8308 extension advertisement to the client, if permitted.
+
```
adiroiban: I think that this can also be used by the client to advertise its extensions.

Or if in twisted this should only be used by the server side, maybe move it to server code and make it private.

```suggestion
    def sendExtInfo(self, extensions):
        """
        Send an RFC 8308 extension advertisement to the remote peer.
        
        Nothing is sent if the peer doesn't support negotiations.
```

and if this should be used by the client side, I think that we should expose `_supportsExtensionNegotiation` as a private member so that the an external code using the client can query if the server supports any extensions.

src/twisted/conch/test/test_transport.py
```diff
@@ -797,6 +799,62 @@ def test_sendKexInitBlocksOthers(self):
...
+                    + common.NS(b"ssh-rsa,rsa-sha2-256")
+                    + common.NS(b"elevation")
+                    + common.NS(b"d"),
+                )
+            ],
+        )
+
+    def test_sendExtInfoUnsupported(self):
+        """
+        If the other end has not advertised support for extension
```
adiroiban: I am not sure where in section 2.2 is this specified.

Maybe is best to copy the relevant text from RFC in the test docstring

src/twisted/conch/test/test_transport.py
```diff
@@ -797,6 +799,62 @@ def test_sendKexInitBlocksOthers(self):
...
+        negotiation, no EXT_INFO message is sent.  See RFC 8308, section
+        2.2.
+        """
+        self.proto.sendExtInfo([(b"server-sig-algs", b"ssh-rsa,rsa-sha2-256")])
+        self.assertEqual(self.packets, [])
+
+    def test_EXT_INFO(self):
+        """
+        Test that EXT_INFO messages are received correctly.  See RFC 8308,
+        section 2.3.
```
adiroiban:
```suggestion
        Once MSG_EXT_INFO is received it will  store the extension information a mapping.
```

received correctly ... is a bit generic in the context of a test docstring :)

adiroiban (approved):
It looks good. Thanks!

Feel free to ping me for any SSH or FTP related PR.

I only have minor comments about doctrings and variable naming.

The logic is perfect :)

Let me know if you want to make any changes or want to merge it as it is.

Thanks again!

src/twisted/conch/ssh/transport.py
cjwatson: Your suggestion looks good; I've applied it, thanks.

I don't have a concrete use case for using this from the client side at the moment, so I'd prefer to keep the attribute private for now to avoid accidentally painting ourselves into a corner by making it public.  No problem with making it public later if that turns out to be useful.

cjwatson:
@adiroiban Thanks for the feedback!  I think I've resolved all of your comments in one form or another, and this should be ready for re-review now.

adiroiban (approved):
Many thanks for the excellent updates.

> I don't have a concrete use case for using this from the client side at the moment, so I'd prefer to keep the attribute private for now to avoid accidentally painting ourselves into a corner by making it public. No problem with making it public later if that turns out to be useful.

`sendExtInfo` is public but I think that is ok.
We should worry about future use case... but not too much :)
We can deprecate or make things public if needed.

---------

this can be merged.

I have added you to @twisted/twisted-contributors team so that you can merge this.

Let me know if you will receive too many notifications and I can remove you..

@twisted/twisted-contributors is also used to request reviews..
so if that too much noise, we can create a `@twisted/reviewers` with people that are willing to get notifications when things needs review in twisted

PR #1671, opened 2021-11-07
10273 use current_thread() instead of currentThread()

altendky:
## Scope and purpose

`current_thread()` was introduced in 2.6 as an alias for `currentThread()`.
The old `currentThread()` spelling raises a `DeprecationWarning` in 3.10.

https://docs.python.org/2.7/library/threading.html#threading.current_thread

```console
$ for v in 2.7 3.5 3.6 3.7 3.8 3.9 3.10; do echo -------------- Python $v; python$v -c 'import threading; print(threading.currentThread())'; done
-------------- Python 2.7
<_MainThread(MainThread, started 139976240273216)>
-------------- Python 3.5
<_MainThread(MainThread, started 140209924052800)>
-------------- Python 3.6
<_MainThread(MainThread, started 139896207542080)>
-------------- Python 3.7
<_MainThread(MainThread, started 139969139185472)>
-------------- Python 3.8
<_MainThread(MainThread, started 140299012052800)>
-------------- Python 3.9
<_MainThread(MainThread, started 140377659451200)>
-------------- Python 3.10
<string>:1: DeprecationWarning: currentThread() is deprecated, use current_thread() instead
<_MainThread(MainThread, started 139908902836032)>
```

Thanks to @whitslack for pointing this out over in https://github.com/pytest-dev/pytest-twisted/issues/146.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10273

(yep, this formatting is messed up due to including a code block in the message.  please edit this and then copy the source for use.)

```
Merge pull request #1670 from altendky/10273-altendky-current_thread

Author: altendky
Fixes: ticket:10273

`current_thread()` was introduced in 2.6 as an alias for `currentThread()`.
The old `currentThread()` spelling raises a `DeprecationWarning` in 3.10.

https://docs.python.org/2.7/library/threading.html#threading.current_thread

```console
$ for v in 2.7 3.5 3.6 3.7 3.8 3.9 3.10; do echo -------------- Python $v; python$v -c 'import threading; print(threading.currentThread())'; done
-------------- Python 2.7
<_MainThread(MainThread, started 139976240273216)>
-------------- Python 3.5
<_MainThread(MainThread, started 140209924052800)>
-------------- Python 3.6
<_MainThread(MainThread, started 139896207542080)>
-------------- Python 3.7
<_MainThread(MainThread, started 139969139185472)>
-------------- Python 3.8
<_MainThread(MainThread, started 140299012052800)>
-------------- Python 3.9
<_MainThread(MainThread, started 140377659451200)>
-------------- Python 3.10
<string>:1: DeprecationWarning: currentThread() is deprecated, use current_thread() instead
<_MainThread(MainThread, started 139908902836032)>
```

Thanks to @whitslack for pointing this out over in https://github.com/pytest-dev/pytest-twisted/issues/146.
```

### DIFF
--- a/src/twisted/internet/test/test_threads.py
+++ b/src/twisted/internet/test/test_threads.py
@@ -14,6 +14,7 @@ from twisted.internet.interfaces import IReactorThreads
 from twisted.internet.test.reactormixins import ReactorBuilder
 from twisted.python.threadable import isInIOThread
 from twisted.python.threadpool import ThreadPool
+from twisted.python.versions import Version
 
 
 class ThreadTestsBuilder(ReactorBuilder):
@@ -219,6 +220,12 @@ class ThreadTestsBuilder(ReactorBuilder):
         reactor.callInThread(check)
         self.runReactor(reactor)
         self.assertEqual([False], results)
+
+    def test_threadPoolCurrentThreadDeprecated(self):
+        self.callDeprecated(
+            version=(Version("Twisted", "NEXT", 0, 0), "ThreadPool.current_thread"),
+            f=ThreadPool.currentThread,
+        )
 
 
 globals().update(ThreadTestsBuilder.makeTestCaseClasses())
--- a/src/twisted/internet/test/test_threads.py
+++ b/src/twisted/internet/test/test_threads.py
@@ -7,7 +7,6 @@ Tests for implementations of L{IReactorThreads}.
 
 
 import gc
-import threading
 from weakref import ref
 
 from twisted.internet.interfaces import IReactorThreads
--- a/src/twisted/internet/test/test_threads.py
+++ b/src/twisted/internet/test/test_threads.py
@@ -109,13 +109,13 @@ class ThreadTestsBuilder(ReactorBuilder):
         result = []
 
         def threadCall():
-            result.append(threading.currentThread())
+            result.append(threading.current_thread())
             reactor.stop()
 
         reactor.callLater(0, reactor.callInThread, reactor.callFromThread, threadCall)
         self.runReactor(reactor, 5)
 
-        self.assertEqual(result, [threading.currentThread()])
+        self.assertEqual(result, [threading.current_thread()])
 
     def test_stopThreadPool(self):
         """
--- a/src/twisted/internet/test/test_threads.py
+++ b/src/twisted/internet/test/test_threads.py
@@ -12,7 +12,7 @@ from weakref import ref
 
 from twisted.internet.interfaces import IReactorThreads
 from twisted.internet.test.reactormixins import ReactorBuilder
-from twisted.python.threadable import isInIOThread
+from twisted.python.threadable import getThreadID, isInIOThread
 from twisted.python.threadpool import ThreadPool
 from twisted.python.versions import Version
 
@@ -110,13 +110,13 @@ class ThreadTestsBuilder(ReactorBuilder):
         result = []
 
         def threadCall():
-            result.append(threading.current_thread())
+            result.append(getThreadID())
             reactor.stop()
 
         reactor.callLater(0, reactor.callInThread, reactor.callFromThread, threadCall)
         self.runReactor(reactor, 5)
 
-        self.assertEqual(result, [threading.current_thread()])
+        self.assertEqual(result, [getThreadID()])
 
     def test_stopThreadPool(self):
         """
@@ -223,7 +223,7 @@ class ThreadTestsBuilder(ReactorBuilder):
 
     def test_threadPoolCurrentThreadDeprecated(self):
         self.callDeprecated(
-            version=(Version("Twisted", "NEXT", 0, 0), "ThreadPool.current_thread"),
+            version=(Version("Twisted", "NEXT", 0, 0), "twisted.python.threadable.getThreadID"),
             f=ThreadPool.currentThread,
         )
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/twisted/python/threadpool.py [hunk 1]
```diff
@@ -43,7 +43,7 @@ class ThreadPool:
     name = None
 
     threadFactory = Thread
-    currentThread = staticmethod(currentThread)
+    current_thread = staticmethod(current_thread)
```
cjwatson: This looks like externally-visible API.  Does this need to keep the old `currentThread` name around, perhaps with a Twisted deprecation notice?

cjwatson (dismissed):
I'm in favour of the general idea, but there's one bit where I need a little convincing.

src/twisted/python/threadpool.py
altendky: Of course this has to be even harder...  `:]`  Thanks for the review.  I agree.

twm: Why change the name in the Twisted API at all? An underscore name here is inconsistent with the rest of the class.

src/twisted/newsfragments/10273.bugfix
```diff
@@ -0,0 +1 @@
+Deprecated ``threading.currentThread()`` in favor of ``threading.current_thread()`` to avoid the deprecation warnings introduced for the former in Python 3.10.
```
twm: If you keep the deprecation please use its fully-qualified name (i.e. `twisted.python.threadpool.ThreadPool.currentThread()`) in this news fragment.

twm (changes requested):
The `ThreadPool.currentThread` alias is undocumented and appears completely unused, so I'd guess it's just vestigial API from before the `twisted._threads` rewrite. Deprecating it seems fine to me.

I suggest avoiding addition of the `ThreadPool.current_thread` API surface. If you do add this API surface you'll need to add a test for it, so I'm marking this PR as needing changes. If you remove the new API surface then it's good to merge.

Thanks for this!

adiroiban:
Many thanks for the cleanup.

I think that we have the old `twisted.python.threadable.getThreadID()` that can be used as a replacement.

```
def getThreadID():
    if threadingmodule is None:
        return _dummyID
    return threadingmodule.currentThread().ident
```
 
I see that ThreadPool.currentThread  it is only used inside the tests and only to get the id.

Would it make send to deprecated it and rename it as a private member? ?

I think that this was also requested by @twm

altendky:
Alrighty, thanks all.  I will switch to deprecating `ThreadPool.currentThread()` in favor of `getThreadID()` and _not_ add `ThreadPool.current_thread()`.

Sorry for trying to "just do this quickly".  In doing so I neglected various important aspects you have all caught, thankfully.

adiroiban:
Many thanks @altendky for the cleanup. 

As a developer I always try to fix the issue (in this case, the deprecation warning) as soon a possible . So most probably I would have come up with the same fix in the first place :)

Is then up to the reviewes to see if the fix has any drawbacks :)

I am not sure that `getThreadID` is a complete replacement for `ThreadPool.currentThread()`

it was my guess that it can be used as a replacement.

Thanks!

src/twisted/newsfragments/10273.bugfix
```diff
@@ -0,0 +1,2 @@
+Deprecated ``twisted.python.threading.ThreadPool.currentThread()`` in favor of ``threading.current_thread()``.
```
altendky, line 1: Does just recommending use of the underlying stdlib function seem better?

twm: Yes, that's strictly better.

src/twisted/python/threadpool.py
```diff
@@ -43,7 +45,12 @@ class ThreadPool:
     name = None
 
     threadFactory = Thread
-    currentThread = staticmethod(currentThread)
+    currentThread = staticmethod(
+        deprecated(
+            version=Version("Twisted", "NEXT", 0, 0),
+            replacement="twisted.python.threadable.getThreadID",
```
twm:
```suggestion
            replacement="threading.current_thread",
```

src/twisted/internet/test/test_threads.py
```diff
@@ -220,5 +221,14 @@ def check():
         self.runReactor(reactor)
         self.assertEqual([False], results)
 
+    def test_threadPoolCurrentThreadDeprecated(self):
+        self.callDeprecated(
+            version=(
+                Version("Twisted", "NEXT", 0, 0),
+                "twisted.python.threadable.getThreadID",
```
twm:
```suggestion
                "threading.current_thread",
```

twm (changes requested):
@adiroiban I think that you are overthinking this? `threading.getThreadID` returns something other than the current thread, the `ident` (an integer):

```py
def getThreadID():
    if threadingmodule is None:
        return _dummyID
    return threadingmodule.currentThread().ident
```
@altendky I'm sorry for my slow response here. I'm marking this as requiring changes, but I think that you are good to merge if you apply the suggested changes. Again, sorry I'm only circling back now. :disappointed:

altendky:
No worries twm, I've been all over the place and this isn't anything I'm waiting on.  I appreciate catching all my errors.

twm (approved):
Looks great! :shipit:

altendky:
@cjwatson, looks like you've still got requested changes.  Have they been satisfied?

PR #1675, opened 2021-12-16
5710 initial implementation of using posix_spawn

glyph:
## Scope and purpose

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/5710

```
Author: glyph
Reviewer: 
Fixes: ticket:5710

Twisted will now use the more efficient `posix_spawn` process-spawning mechanism when the parameters passed allow for it, replacing `fork`/`exec`.
```


glyph:
One follow-up I can think of here is making the usage of posix_spawn a bit more visible. It might be a *huge* performance boost for some applications, and those should be able to figure out what it was that made a difference.

src/twisted/internet/process.py
```diff
@@ -643,6 +677,7 @@ def __init__(
         nuances of setXXuid on UNIX: it will assume that either your effective
         or real UID is 0.)
         """
+        self._reactor = reactor
```
wsanchez, line 680: Rather than hanging a new attribute on the object, perhaps just pass this along to `_fork` and then `_trySpawnInsteadOfFork`?

wsanchez (approved):
Seems legit to me

src/twisted/internet/process.py
adiroiban: I think that having `_reactor` class/instance member is ok.
It's a common patter and makes testing much easier.

The `_trySpawnInsteadOfFork` is private API, so I think that doing a dependency injection there is more complicated.

adiroiban (approved):
All good.  Thanks. Ready to merge.

PR #1676, opened 2021-12-27
10287: Type annotate twisted.python.fakepwd

twm:
## Scope and purpose

We've had some bugs due to `bytes` vs. `str` issues in [#9130](https://twistedmatrix.com/trac/ticket/9130) and [#10286](https://twistedmatrix.com/trac/ticket/10286). This indicates a shortcoming of the fake implementation: it happily passes through `bytes` when it should require `str`. Therefore, add type annotations and a few runtime type checks to better resemble the real `pwd` and `spwd` modules ([#10287](https://twistedmatrix.com/trac/ticket/10287#ticket)).

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10287
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] The title of the PR starts with the associated Trac ticket number (without the `#` character).
* [x] I have updated the automated tests and checked that all checks for the PR are green.

    Failures appear unrelated to the content of the change.

* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.

PR #1678, opened 2021-12-28
10286: UNIXAuthorizedKeysFiles type issues

twm:
## Scope and purpose

It turns out all the tests for this code were skipped because the `skip` values were the empty string, rather than ``None``, when the conditional imports succeeded. I've rearranged the imports a bit so that the tests run, and made them typecheck with the changes from #1676.

The actual fix here is pretty simple: decode the username [as in the fix to #9130](https://twistedmatrix.com/trac/ticket/9130). As usual the stdlib's obsession with making everything `str` instead of `bytes` ~bites~ costs us in pointless copies and additional error cases. 🙄

Getting it all to type check... well, I'm hoping that #1677 will allow me to remove some of the casts here, as [typeshed has typings for `pwd`](https://github.com/python/typeshed/blob/master/stdlib/pwd.pyi) and [`spwd`](https://github.com/python/typeshed/blob/master/stdlib/spwd.pyi).

The layering here isn't great — the implementation mixes a pluggable `pwd` backend with the system-level concern of filesystem encoding — but I don't see a _simple_ fix that preserves backwards compatibility. Easier to replace ``UNIXAuthorizedKeysFiles`` entirely, which I've no ambition for.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10286


src/twisted/conch/newsfragments/10286.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conc.checkers.UNIXAuthorizedKeysFiles now uses the filesystem encoding to decode usernames before looking them up in the password database, so it works on CPython 3.
```
exarkun: type in the module name here.  also, pypy is still supported I think?  So maybe Python 3 instead of CPython 3.

twm: It actually already worked on PyPy 3 due to different (IMO better) behavior there, but this is perhaps more confusing than illuminating to mention here.

adiroiban:
```suggestion
twisted.conch.checkers.UNIXAuthorizedKeysFiles now uses the filesystem encoding to decode usernames before looking them up in the password database.
This is a fix for CPython 3. PyPy3 was not affected by this bug.
```

src/twisted/conch/checkers.py
```diff
@@ -47,34 +44,115 @@
...
+
+    @returns: A L{pwd.struct_passwd}, where field 1 may contain a crypted
+        password, or L{None} when the L{pwd} database is unavailable.
+
+    @raises KeyError: when no such user exists
     """
     if pwd is None:
         return None
-    return pwd.getpwnam(username)
+    return cast(CryptedPasswordRecord, pwd.getpwnam(username))
```
adiroiban, line 124: #1677 is merged. Can this cast be removed?

src/twisted/conch/checkers.py
```diff
@@ -47,34 +44,115 @@
...
+        password, or L{None} when the L{spwd} database is unavailable.
+
+    @raises KeyError: when no such user exists
     """
     if spwd is not None:
         f = spwd.getspnam
     else:
         return None
-    return runAsEffectiveUser(0, 0, f, username)
+    return cast(CryptedPasswordRecord, runAsEffectiveUser(0, 0, f, username))
```
adiroiban, line 145: is this cast still needed after #1677?

src/twisted/conch/checkers.py
```diff
@@ -47,34 +44,115 @@
...
+    Lookup a user by name in a L{pwd}-style database.
+
+    @param userdb: The user database.
+
+    @param username: Identifying name in bytes. This will be decoded according
+    to the filesystem encoding, as the L{pwd} module does internally.
+
+    @raises KeyError: when the user doesn't exist
+    """
+    return userdb.getpwnam(username.decode(sys.getfilesystemencoding()))
```
adiroiban, line 106: This looks OK as I see a similar code in `UNIXPasswordDatabase.requestAvatarId`

I was expecting the credential username and password to be `str`.
Who is using random bytes for this?

Or allow the whole Twisted cred framework to be setup with a custom encoding.

--------

I am not using the Twisted Cred framework myself.
I think that when I tried to use it 10 years ago, when I was just learning Twisted I ended up with lots of issued due to username encodings.
I had automated tests running for various encoding and at that time I didn't understood well how default encoding works and how to mock it.
...so I ended up with my custom credential code.

adiroiban (approved):
It looks good. Minor comments inline.

Please check the comments and can merge once all tests are green.

Many thanks!

I have updated the branch from UI to remove 3.6 and get the tests green.

src/twisted/conch/checkers.py
twm: Unfortunately it seems not, as MyPy still complains:

```
src/twisted/conch/checkers.py:155:5: error: Returning Any from function declared to return "Optional[CryptedPasswordRecord]" 
[no-any-return]
        return runAsEffectiveUser(0, 0, f, username)
```

src/twisted/conch/checkers.py
twm: Same story:

```
src/twisted/conch/checkers.py:134:12: error: Incompatible return value type (got "UserRecord", expected
"Optional[CryptedPasswordRecord]")  [return-value]
        return pwd.getpwnam(username)
```

src/twisted/conch/checkers.py
twm: The code it's actually inverting is this:

https://github.com/python/cpython/blob/a5451c96a14ac0c3ee7cef7b5c52ab1d783ec613/Modules/pwdmodule.c#L222-L226

As I grumbled on the PyPy bug I think it's weird to force us to deal in Unicode when the actual API is `bytes` with an undefined encoding, but :man_shrugging:.

twm:
I'll wait for https://twistedmatrix.com/trac/ticket/10304 for `Protocol`.

src/twisted/conch/checkers.py
adiroiban: Yes. I guess that in cpython3 due to the impulse to have everything as unicode, we ended up with this API.

This is life :)

adiroiban:
Thanks for the update. Yes. Once the CI is green please merge.
You can also enable auto-merge :)

PR #1680, opened 2022-01-11
7796: Set f_globals and f_locals on _Frame

pdunning-xilinx:
## Scope and purpose

Having f_globals and f_locals in traceback frames is useful for post
mortem debugging. Thus we should provide what we can.

If we're constructing a fake traceback then we are doing so for a
Failure that has been cleaned and therefore the variables in f_globals
and f_locals are the result of _safeReprVars(). The benefit of that is
not holding references to the original objects anymore but the downside
is that the post mortem debugger only has access to string
representations of the variables. That's still a lot better than showing
all the variables as not defined in the post mortem debugger.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/7796
One test fails due to my system not having IPv6 connectivity (twisted.python.test.test_sendmsg.GetSocketFamilyTests.test_inet6). The test is unrelated to the patch.

Do the `mypy` checks have to pass? There are some failures in that due to a missing dependency in the `tox` config and two failures that just seem to be there in `trunk`.

```

Author: pdunning-xilinx
Fixes: ticket:7796

Having f_globals and f_locals in traceback frames is useful for post
mortem debugging. Thus we should provide what we can.

If we're constructing a fake traceback then we are doing so for a
Failure that has been cleaned and therefore the variables in f_globals
and f_locals are the result of _safeReprVars(). The benefit of that is
not holding references to the original objects anymore but the downside
is that the post mortem debugger only has access to string
representations of the variables. That's still a lot better than showing
all the variables as not defined in the post mortem debugger.
```


pdunning-xilinx:
So the CI pipeline has pypy and MAC problems that I don't understand.

pdunning-xilinx:
I've added to the test for `_Frame` attributes to test this change.

pdunning-xilinx:
Looks like all the other recent PRs show the same CI failures for Pypy 3.6 and MacOS. I can't see a ticket for the problem.

glyph:
@pdunning-xilinx I filed https://twistedmatrix.com/trac/ticket/10291 and it's now fixed.

pdunning-xilinx:
Excellent, now all the checks are passing.

src/twisted/python/failure.py
```diff
@@ -163,8 +160,8 @@ def __init__(self, frameinfo, back):
         name, filename, lineno, localz, globalz = frameinfo
         self.f_code = _Code(name, filename)
         self.f_lineno = lineno
-        self.f_globals = {}
-        self.f_locals = {}
+        self.f_globals = dict(globalz or {})
+        self.f_locals = dict(localz or {})
```
exarkun, line 164: I think this `or` is untested.  Is it meant to handle `None`?  What if one of these values is `{}`?  Substituting a new empty dict in place of the given empty dict is probably surprise.  Also, why substitute an empty dict at all?  If the frameinfo had None, shouldn't the `_Frame` have None?

pdunning-xilinx: The old unit test used to pass None in. I suppose that means the or is not tested now.
That opens up the question of what should be passable as an argument.
I don't know whether the real frameinfo can have a None.

Certainly taking `None` or `{}` as an input and replacing them with a new empty dict maintains the existing behaviour
which always gave a new empty dict.
Also I think it makes sense to always produce a new dict, just to avoid mutability issues, hence the `dict()` call.

exarkun: Okay - "always a copy" is reasonable.

adiroiban:
I think that this can be closed as the changes were moved to https://github.com/twisted/twisted/pull/1684

Feel free to reopen is I am wrong.

Thanks!

PR #1682, opened 2022-01-17
10292: Fix bad test skips

twm:
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10292

### DIFF
--- a/src/twisted/conch/test/test_transport.py
+++ b/src/twisted/conch/test/test_transport.py
@@ -29,8 +29,9 @@ from twisted.trial.unittest import TestCase
 pyasn1 = requireModule("pyasn1")
 cryptography = requireModule("cryptography")
 
+dependencySkip: Optional[str]
 if pyasn1 and cryptography:
-    dependencySkip = ""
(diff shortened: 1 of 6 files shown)


glyph (approved):
This looks like good forward progress. Ship it.

That said: could we possibly get more principled about this process?  It's still easy to accidentally trip the skip logic in builds where we don't expect it by missing some transitive dependency or something.

My suggestion would be like a `TWISTED_TEST_SKIP_FEATURES` env var which must be set in order to skip certain sets of tests, so we set it explicitly in CI, or in tox.

Thoughts?

twm:
I definitely share your discomfort with the current approach. It's too easy to _think_ you've run the tests, especially locally. Particularly as there are _always_ skipped tests for other platforms in the run, cluttering the output.

An environment variable sounds reasonable to me, though the non-Tox devs among us may object.

adiroiban:
There is also this PR  related to the skip logic https://github.com/twisted/twisted/pull/1474

PR #1690, opened 2022-01-31
10301 Fix support for `contextvar.reset`

richvdh:
## Scope and purpose

We need to use the same copy of the context for each iteration of the generator function, to ensure that `.reset` works as expected.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10301

```

Author: richvdh
Fixes: ticket:10301

We need to use the same copy of the context for each iteration of the generator function,
to ensure that `.reset` works as expected.
```


itamarst:
I encountered the same problem (it impacts Eliot logging—https://eliot.readthedocs.io/en/stable/—which relies on contextvars). My reproducer:

```python
from twisted.internet.defer import inlineCallbacks
from twisted.internet.task import deferLater
from contextvars import ContextVar
from contextlib import contextmanager

MY_VAR = ContextVar("example")

@contextmanager
def set_var(name):
    print("setting!", name)
    token = MY_VAR.set(1)
    try:
        yield
    except Exception as e:
        print("resetting, exception!", name, e)
        MY_VAR.reset(token)
        raise
    else:
        print("resetting, success!", name)
        MY_VAR.reset(token)
        print("finished reseting", name)

@inlineCallbacks
def parent():
    with set_var("parent"):
        print("parent yielding")
        x = yield child()
        print("parent done yielding")
        return x

@inlineCallbacks
def child():
    with set_var("child"):
        print("child yielding")
        x = yield deferLater(reactor, 0.01, lambda: 123)
        print("child done yielding")
        return x

parent().addBoth(lambda r: print("Result", r) or reactor.stop())
reactor.run()
```

This PR fixes the problem, and conceptually makes sense to me as well.

src/twisted/test/test_defer.py
```diff
@@ -3364,10 +3364,26 @@ def test_withInlineCallbacks(self) -> None:
 
         @defer.inlineCallbacks
         def yieldingDeferred() -> Generator[Deferred[Any], Any, None]:
+            # first try setting the var
```
exarkun: This test is already excessively complex.  Can you please move testing for this fix into a new test method dedicated to this one specific behavior?

exarkun (changes requested):
Thanks!  The implementation change looks correct to me.  I left a comment about test factoring inline.  Once that is addressed I would be happy to see this merged.

src/twisted/test/test_defer.py
richvdh: @exarkun thanks for the review! I've split this test up as you suggest in 83c9a66.

Please could you take another look?

exarkun (approved):
Thanks!

PR #1692, opened 2022-02-06
9765: Add support for SSH RSA keys with SHA-2 signatures

cjwatson:
## Scope and purpose

This is defined in RFC 8332.  The keys themselves stay the same but now
support multiple signature algorithms.

OpenSSH 8.8 disabled RSA signatures using SHA-1 by default, so this is
needed to interoperate cleanly.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9765

```

Author: cjwatson
Fixes: ticket:9765

This is defined in RFC 8332.  The keys themselves stay the same but now
support multiple signature algorithms.
```


src/twisted/conch/ssh/factory.py
```diff
@@ -31,14 +33,35 @@ class SSHFactory(protocol.Factory):
...
+                    key = keys.Key.fromString(
+                        self.publicKeys[source].toString("OPENSSH")
+                    )
+                    key.hashAlgorithm = hashAlgorithm
+                    self.publicKeys[target] = key
         if not hasattr(self, "privateKeys"):
-            self.privateKeys = self.getPrivateKeys()
+            self.privateKeys = dict(self.getPrivateKeys())
+            for source, target, hashAlgorithm in self._extraHashAlgorithms:
+                if source in self.privateKeys and target not in self.privateKeys:
```
adiroiban:
```suggestion
            self.privateKeys = dict(self.getPrivateKeys())
            for source, target, hashAlgorithm in self._extraHashAlgorithms:
                 # Automatically generate SHA2 variants if they are not returned
                 # by the user.
                if source in self.privateKeys and target not in self.privateKeys:
```

src/twisted/conch/ssh/factory.py
```diff
@@ -31,14 +33,35 @@ class SSHFactory(protocol.Factory):
...
+                    key = keys.Key.fromString(
+                        self.privateKeys[source].toString("OPENSSH")
+                    )
+                    key.hashAlgorithm = hashAlgorithm
```
adiroiban: Since `fromString` is a constructor, I would go with something like this 

```suggestion
                    key = keys.Key.fromString(
                        self.privateKeys[source].toString("OPENSSH")
                        hashAlgorithm=hashAlgorithm,
                    )
```

but the current API is also OK and I not -1 on it.

The proposed API would meant that we don't need a setter for `hashAlgorithm` and all validation can be done at `init` or `fromString` level.

src/twisted/conch/ssh/keys.py
```diff
@@ -1114,6 +1144,59 @@ def size(self):
...
+                keySize = self.size()
+                if keySize <= 256:
+                    return hashes.SHA256()
+                elif keySize <= 384:
+                    return hashes.SHA384()
+                else:
+                    return hashes.SHA512()
+            elif keyType == "Ed25519":
+                return hashes.SHA512()
+        return self._hashAlgorithm
```
adiroiban: Very minor suggestion. I hope it helps with searching for if /elif pairs :)

```suggestion
        if self._hashAlgorithm:
            # We have an explicit algorithm.
            return self._hashAlgorithm

        # When no explicit algorithm is defined we use a default
        # one based on the key type.
        keyType = self.type()
        if keyType in ("RSA", "DSA"):
            return hashes.SHA1()

        if keyType == "EC":
            # Hash size depends on key size
            keySize = self.size()
            if keySize <= 256:
                return hashes.SHA256()
            if keySize <= 384:
                return hashes.SHA384()
            # For any other size go SHA512 as the default.
            return hashes.SHA512()

       if keyType == "Ed25519":
            return hashes.SHA512()
```

src/twisted/conch/ssh/keys.py
```diff
@@ -921,15 +921,21 @@ def _fromEd25519Components(cls, a, k=None):
...
         @param keyObject: Low level key.
         @type keyObject: C{cryptography.hazmat.primitives.asymmetric} key.
+
+        @param hashAlgorithm: The hash algorithm to use when signing and
+        verifying using this key, or L{None} to use a default appropriate
+        for the key type.
+        @type hashAlgorithm: L{hashes.HashAlgorithm} or L{None}
         """
         self._keyObject = keyObject
+        self.hashAlgorithm = hashAlgorithm
```
adiroiban: Do we need it as public writable member?

I would prefer to have it as a read-only property.

I think that hashAlgorithm is something that is defined once and then is only used internally.

I am hoping that by making it private and only setting it at construct time, we can prevent  some future strange bugs in which the hashAlgoritm is updated at runtime.

src/twisted/conch/newsfragments/9765.feature
```diff
@@ -0,0 +1 @@
+twisted.conch.ssh now supports using RSA keys with SHA-2 signatures (RFC 8332).
```
adiroiban: maybe also add a note that `rsa-sha2-512` and `rsa-sha2-256` are now automatically generated and set as the preferred host keys.

And in the release note, maybe add a note that the support is automatic and conch will use the existing RSA keys to generate SHA2 variants.

No API change is require for end users.

src/twisted/conch/test/test_ssh.py
```diff
@@ -864,12 +864,37 @@ class SSHFactoryTests(unittest.TestCase):
...
+            b"ssh-rsa": keys.Key.fromString(privateRSA_openssh)
+        }
         sshFactory.startFactory()
         return sshFactory
 
+    def test_keys(self):
+        """
+        startFactory() adds some key variants using stronger hash
+        algorithms.
+        """
```
adiroiban:
```suggestion
    def test_startFactoryAutomaticKeyVariants(self):
        """
        startFactory() automatically updates the internal keys to
        add variants using stronger hash algorithm.
        """
```

adiroiban (approved):
This looks very good. Thanks.

It would be nice for the release notes to make it clear that end users don't have to do anything to get SHA2 host keys.
They only need to update Twisted and it will auto-magically work :)

My only important comment is about the new public writable `Key.hashAlgorithm` attribute.

I am ok to have it merged as it is.
I would prefer to have it read-only and defined only at init/construct time.

----------------

I think that `self.publicKeys` and `self.privateKeys` are the old API that was supposed to be deprecated.

For the long term, I would like to see `self.publicKeys` and `self.getPublicKeys()` deprecated.

If you have the private key, you no longer need an separate source for the public keys.
You can just extract the public key from the private one.

... and it can lead to strange errors in which `getPublicKeys` returns one set of keys and `getPrivateKeys` returns another set of keys. 

Also, I am using a self.getHostKeys() API which only returns a list of `ssh.Key`.
the ssh type is added automatically by the transfer and end users don't have to worry about that.

src/twisted/conch/ssh/keys.py
cjwatson: This comment caused me to realize that this whole approach was fundamentally based on woolly thinking.  The public key signature algorithm (and hence the hash algorithm) is fundamentally not a property of the key; it's a property of how you use the key, by signing or verification.  When I went and looked at how the client equivalent of this work might work in future, I found that we'll really need to keep the concepts more clearly separated there, since a client picks a key and then picks an appropriate key signature algorithm for it; having to construct a copy of the key using a different hash algorithm would be very cumbersome.  So I'm glad you saved me from locking us into an inappropriate API!

Instead of all this, I've changed `Key.sign` to take an optional `signatureType` argument.  This turns out to be quite straightforward, and in particular it means there's no messing about in the factory to make variant copies of keys (although I did have to teach the factory to fill in `supportedPublicKeys` properly).  The patch is a bit longer, but I think it's now conceptually clearer.

src/twisted/conch/ssh/keys.py
cjwatson: Thanks - but I think I really prefer the `if`/`elif` chain in this case.  I find that clearer than repeatedly falling through like this.  Sometimes this sort of thing can be worth it to reduce indentation level, but I don't think it makes enough of a difference in this case to be worth (for me) the extra confusion in reading it.

src/twisted/conch/ssh/factory.py
cjwatson: This is now fixed by not having `hashAlgorithm` be a property of the key at all.

src/twisted/conch/newsfragments/9765.feature
cjwatson: I've extended this news fragment a bit to try to capture that.

cjwatson:
@adiroiban Thanks a lot for your very useful previous review.  I know you approved it, but since I've made quite extensive changes I'd appreciate it if you could take another look.

src/twisted/conch/ssh/keys.py
```diff
@@ -143,6 +150,22 @@ def _normalizePassphrase(passphrase):
         return passphrase
 
 
+def _getKeyFormat(keyAlg):
```
adiroiban: naming this helper is hard... I don't have a better suggest :)

One idea is to move this inside the transport and use it as a proxy for the factory key access 

```diff
- self.factory.publicKeys[self.keyAlg]
+ self._getPublicKey(self.keyAlg) 
```

And Transport._getPublicKey can return the public key from the factory, based on the name of the requested algorithm.

--------


Very minor comment

I think that this should be moved into the transport.py file as this is independent of the code from keys.py and is more related to the SSH protocol.

But it can be kept here.

---------

As a general comment, a lot of the code from keys.py is not very linked to Twisted or async networking

I would like to see a lot of the fromString and toString moved into `cryptography` serialization part where it can also be shared with paramiko or maybe in some soft of Django app in which you need to do various key conversions for end users.

The code from keys.py can only be a very thin wrapper around the `cryptography` library `self._keyObject` public or private key.

adiroiban (approved):
Thanks. Looks very good.
Happy to help by accident :p 

Only a minor comment, but if the tests are green this can be merge.

src/twisted/conch/ssh/keys.py
cjwatson: I went back and forth on having this in `keys` vs. `transport`, so I'm quite happy to move it.  Will do.

And yes, a lot of `keys` dates back to the old pycrypto days ...

adiroiban: This is private API and we can always update. So no need to worry to much about this.

Thanks.

I hope I will find some time to contribute a refactoring for the current keys.py file.

For my work, I am using a forked version which removes the Twisted deps and for convenience allows you to load Putty SSH keys or Tectia SSH keys or extract the public part from OpenSSL certs  - https://github.com/chevah/chevah-keycert/

The fork was made long time ago when cryptography didn't even existed ... but I think that now we have better serialization support and should reuse more code from

https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/

adiroiban:
Excellent work @cjwatson . Let me know if you need or want a new release of twisted that will include this change.

If OpenSSH is dropping (or has dropped) support for SHA1, I think that this might be important.

cjwatson:
@adroiban My main motivation is Launchpad, and since that's currently stuck on Python 3.5 for other reasons, I can't use current Twisted releases anyway and need to cherry-pick things to older releases - so I have no pressing need for a release and I expect that the next routine release will do just fine.  That's not to say nobody else might have a need for it, of course!

adiroiban:
ok. I am not in a hurry for a release... I am also doing cherry picking for python 2.7 :(

lilmattSB (approved):

adiroiban:
Only now I had the time to integrate these changes into my project.

All looks ok, with the exception of client-side SSH public key authentication.

conch.ssh will always request the SSH public key auth for RSA with the `ssh-rsa` algorithm

https://github.com/twisted/twisted/blob/2a33824557ed65d2241a51d6bba07ac76521b50f/src/twisted/conch/ssh/userauth.py#L612-L624

I am not sure how this can be solved in Twisted in a nice way.

I guess that we will need to have a `t.conch.ssh.userauth.SSHUserAuthClient.rsa_algorithm` or something similar.

Or create some clever detection based on the extension.

I will create [a new ticke](https://github.com/twisted/twisted/issues/11808)t and will add more info there.

this is just if someone is looking into this.

PR #1696, opened 2022-02-13
10308: Make SSHSession clean up both client and session

cjwatson:
## Scope and purpose

If an environment variable passing request and a subsystem request were
both sent on the same channel, then SSHSession only cleaned up the
ISession adapter when receiving EOF or closing the channel, and did not
call loseConnection on the client transport which is the only reasonable
way for a subsystem to be notified when a connection is closed.

SSHSession now cleans up both the client transport and the ISession
adapter if both are set.

I think this should be safe for non-subsystem sessions as well - I can't see a practical downside to doing more careful cleanup here.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10308
* [x] I have created a newsfragment in src/twisted/conch/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))

```

Author: cjwatson
Reviewer: adiroiban
Fixes: ticket:10308

If an environment variable passing request and a subsystem request were
both sent on the same channel, then SSHSession only cleaned up the
ISession adapter when receiving EOF or closing the channel, and did not
call loseConnection on the client transport which is the only reasonable
way for a subsystem to be notified when a connection is closed.

SSHSession now cleans up both the client transport and the ISession
adapter if both are set.
```

### DIFF
--- a/src/twisted/conch/ssh/session.py
+++ b/src/twisted/conch/ssh/session.py
@@ -180,16 +180,16 @@ class SSHSession(channel.SSHChannel):
             log.warn("Weird extended data: {dataType}", dataType=dataType)
 
     def eofReceived(self):
+        if self.client:
+            self.conn.sendClose(self)
         if self.session:
             self.session.eofReceived()
-        elif self.client:
-            self.conn.sendClose(self)
 
     def closed(self):
+        if self.client:
+            self.client.transport.loseConnection()
         if self.session:
             self.session.closed()
-        elif self.client:
-            self.client.transport.loseConnection()
 
     # def closeReceived(self):
     #    self.loseConnection() # don't know what to do with this
--- a/src/twisted/conch/ssh/session.py
+++ b/src/twisted/conch/ssh/session.py
@@ -186,7 +186,7 @@ class SSHSession(channel.SSHChannel):
             self.session.eofReceived()
 
     def closed(self):
-        if self.client:
+        if self.client and self.client.transport:
             self.client.transport.loseConnection()
         if self.session:
             self.session.closed()
--- a/src/twisted/conch/newsfragments/10308.bugfix
+++ b/src/twisted/conch/newsfragments/10308.bugfix
@@ -1,0 +1,1 @@
+twisted.conch.ssh.session.SSHSession now cleans up both the client transport and the ISession adapter if both are set.  Previously, a subsystem's connectionLost method was not called if a environment variable passing request was also sent on the same channel.
--- a/src/twisted/conch/test/test_session.py
+++ b/src/twisted/conch/test/test_session.py
@@ -542,6 +542,25 @@ class SessionInterfaceTests(RegistryUsingMixin, TestCase):
         self.assertTrue(self.session.client.transport.close)
         self.session.client.transport.close = False
 
+    def test_client_closed_with_env_subsystem(self):
+        """
+        If the peer requests an environment variable in its setup process
+        followed by requesting a subsystem, SSHSession.closed() should tell
+        the transport connected to the client that the connection was lost.
+        """
+        self.assertTrue(
+            self.session.requestReceived(b"env", common.NS(b"FOO") + common.NS(b"bar"))
+        )
+        self.assertTrue(
+            self.session.requestReceived(
+                b"subsystem", common.NS(b"TestSubsystem") + b"data"
+            )
+        )
+        self.session.client = StubClient()
+        self.session.closed()
+        self.assertTrue(self.session.client.transport.close)
+        self.session.client.transport.close = False
+
     def test_badSubsystemDoesNotCreateClient(self):
         """
         When a subsystem request fails, SSHSession.client should not be set.


src/twisted/conch/ssh/session.py
```diff
@@ -180,16 +180,16 @@ def extReceived(self, dataType, data):
             log.warn("Weird extended data: {dataType}", dataType=dataType)
 
     def eofReceived(self):
+        if self.client:
```
adiroiban, line 183: If it works for you. I think that is ok.

For me, the SSH session and subsystem experience is a bit of a mess ... both for server side and client side usage.


I have also hit this issue and did some fixed for my fork without knowing what is the right way to do it...so I have not sent a PR.

Happy to see this PR.

Basically, I found that having  `self.session` and `self.client` and `self.conn` was confusing.

And I think that `self.client` is kind of a `self.subsystem` thing.

So I did some refactoring and removed the usage of `self.session` so that I only have `self.client` and `self.conn` 

also... I have `self.client.eofReceived()` so that the client will know what happen and can take the appropriate actions... for example triggering a connection close. 

And `self.client` is in fact a session ... but I found that better to not name it session and have a session in a session.

and to not have to use both `self.client` and `self.session` we can have something like

```python
self.session.openSubsystem(client)
```

I am aware that this might not be backward compatible. 

.... so, in summary.

I am not happy with the current design of SSHSession ...but I don't know how we can untangle this in an easy way without breaking backward compatibility.

----------

In general I am not happy with the current logic.
`If we have a self.client, do something on the self.conn`

I would prefer something like

`If we have self.client, let the client know that eof was received.`
Then the client can do what it wants.

But, I think that if this works for we, we should be fine.

For the long term, I think the only way out of this is to have a `SSHSessionGen2` in which we don't have both session and client.

src/twisted/conch/ssh/session.py
```diff
@@ -180,16 +180,16 @@ def extReceived(self, dataType, data):
...
+            self.conn.sendClose(self)
         if self.session:
             self.session.eofReceived()
-        elif self.client:
-            self.conn.sendClose(self)
 
     def closed(self):
+        if self.client and self.client.transport:
+            self.client.transport.loseConnection()
```
adiroiban, line 190: If it works for you, this is fine.

For my fork, I think that I found out that by calling self.client.transport.loseConnection() will trigger any close notifications on the client itself.

So here I am triggering something like `self.client.closed()` to also let the client know that it's over.

But i don't remember all the details...I remember that it was a bit of a mess :(

-----------


Here, in my fork I only have self.session.closed() ... but the session is linked to the client and the session will close the transport and notify the client.

adiroiban (approved):
Looks good. Many thanks for the PR.

I can't think of any other backward compatible way to fix this.

For the long  term, I hope we can get rid of `self.client` and only use the `self.session` to handle all kind of operations exec/shell/subsystem/env .

------------

But maybe we don't need SSHSessionGen2 and we can deprecate `SSHSession.client` and add `SSHSessionProcessProtocol.setSubsystem` or something like that.

adiroiban:
@cjwatson if you have time, you can also check https://github.com/twisted/twisted/pull/1697 

If you don't have time for a full review, it would help just general feedback to make sure the PR is on the right track.

Thanks!

src/twisted/conch/ssh/session.py
cjwatson: Yeah, arranging for backward compatibility here is utterly miserable.  We ran into this pretty hard when adding environment variable passing support as well.

`self.client` isn't just a subsystem thing - it's also used for communicating with shells or commands.  But yes, I agree that this is all a mess.  A further complication that I'm not sure you've considered is that environment variable passing requires creating an `ISession` adapter (`self.session`) before we're in a position to create an `SSHSessionProcessProtocol` (`self.client`), and the resulting `ISession` adapter must be kept around since otherwise the environment variable request would have no effect.

One approach that might work in a backward-compatible way would be to leave `SSHSession.client` and `SSHSession.session` as-is (unfortunate though that layout is), but move the session handling in `SSHSession.eofReceived` and `SSHSession.closed` into `SSHSessionProcessProtocol`; that way at least the protocol controls all the cleanup.  However, for that to work properly we'd need to create an unconnected `SSHSessionProcessProtocol` a bit earlier in order to handle the case where an environment variable request has been sent but a shell/exec/subsystem request hasn't yet been sent, since currently `self.session` is set in that case but `self.client` isn't.

I think that could be done without a total backward-incompatible rewrite.  What do you think?  (But it feels like something that should be a separate PR.)

src/twisted/conch/ssh/session.py
cjwatson: Yeah, the problem is that the session adapter can't really gain a link to the client protocol without significant backward-compatibility difficulties, since the session adapter is entirely provided by the application; so I think we do need these as separate concepts, although as mentioned above maybe the client protocol could gain a link to the session adapter and then it would be able to deal with notifying the session adapter.

Notifying the actual SSH client of the closure is dealt with by the `self.conn.sendClose(self)` call in `SSHSession.eofReceived`, which sends `MSG_CHANNEL_CLOSE` on the wire.

src/twisted/conch/ssh/session.py
adiroiban: > A further complication that I'm not sure you've considered is that environment variable passing requires creating an ISession adapter (self.session) before we're in a position to create an SSHSessionProcessProtocol (self.client), and the resulting ISession adapter must be kept around since otherwise the environment variable request would have no effect.

Yes. I remember reviewing this. Thanks again for that fix :)


-----------

> I think that could be done without a total backward-incompatible rewrite. What do you think? (But it feels like something that should be a separate PR.)

yes. A separate review.
I am happy to review such a PR :)

I will not have time to work on that refactoring anytime soon :(

src/twisted/conch/ssh/session.py
adiroiban: Ah... sorry. I wanted to say. Notify the `self.client` subsystem... not the remote SSH peer.

Another reason to stop using `self.client` :)

----------

Thanks for the merge!

PR #1697, opened 2022-02-14
[Fix #10309] Refactor SSH version string parsing.

### LINKED ISSUE #10309: Refactoring of SSH protocol version exchange
|-|-|
|Trac ID|trac#10309|
|Type|enhancement|
|Created|2022-02-13 22:45:24Z|

Let's see if we can make the code better :)

```
trac-id__10309 10309
type__enhancement enhancement
reporter__adiroiban adiroiban
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__review review
time__1644792324634761 1644792324634761
changetime__1644799433811120 1644799433811120
version__None None
owner__None None

```
</details>

adiroiban:
## Scope and purpose

This is a refactoring of the SSH peer version string handling.

In 98387b39e9f0b21462f6abc7a1325dc370fcdeb1 we did a quick and dirty fix.

The goal of this PR is to look at providing a better fix.

The main issue with the previous fix, is that in theory a SSH peer can send with the first package the followings :

* SSH version string
* supported ciphers.

Now... the SSH version string is limited to 253 byte... but the supported ciphers can be provided in a packet that is more than 4K in length.
Especially if the peer support a lot of cipher and cipher aliases.

----------

I don't know why conch SSH was implemented to ignore the lines received before 'SSH-`.
I don't thinks that this is OK.
The test was implemented in https://github.com/twisted/twisted/commit/775318c4be8d4c9dce31abde0ae96b0e20f9afac

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10309

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10309

Refactor the SSH version string exchange to no longer ignore data sent before the version string.
The SSH version string no longer accepts null characters.
The SSH version string is now limited to 253 characters.
```

reviewers: @twisted/twisted-contributors

### DIFF
--- a/src/twisted/conch/ssh/transport.py
+++ b/src/twisted/conch/ssh/transport.py
@@ -730,34 +730,56 @@ class SSHTransportBase(protocol.Protocol):
         self.buf = self.buf + data
         if not self.gotVersion:
 
-            if len(self.buf) > 4096:
+            if len(self.buf) >= 255 and b"\n" not in self.buf:
+                # Looks like we have already received enough data,
+                # yet it doesn't look like the version string line
+                # is in the received data.
+                # RFC 4253 section 4.2
+                # The maximum length of the string is 255 characters,
+                # including the Carriage Return and Line Feed.
                 self.sendDisconnect(
                     DISCONNECT_CONNECTION_LOST,
-                    b"Peer version string longer than 4KB. "
-                    b"Preventing a denial of service attack.",
+                    b"Invalid peer version. Long value. Preventing a denial of service attack.",
                 )
                 return
 
-            if self.buf.find(b"\n", self.buf.find(b"SSH-")) == -1:
+            if b"\n" not in self.buf:
+                # We don't have yet the full version string.
+                return
+
+            if not self.buf.startswith(b"SSH-"):
+                # The identification string MUST be
+                # SSH-protoversion-softwareversion SP comments CR LF
+                self.sendDisconnect(
+                    DISCONNECT_CONNECTION_LOST, b"Invalid peer version format."
+                )
                 return
 
             # RFC 4253 section 4.2 ask for strict `\r\n` line ending.
             # Here we are a bit more relaxed and accept implementations ending
             # only in '\n'.
             # https://tools.ietf.org/html/rfc4253#section-4.2
-            lines = self.buf.split(b"\n")
-            for p in lines:
-                if p.startswith(b"SSH-"):
-                    self.gotVersion = True
-                    # Since the line was split on '\n' and most of the time
-                    # it uses '\r\n' we may get an extra '\r'.
-                    self.otherVersionString = p.rstrip(b"\r")
-                    remoteVersion = p.split(b"-")[1]
-                    if remoteVersion not in self.supportedVersions:
-                        self._unsupportedVersionReceived(remoteVersion)
-                        return
-                    i = lines.index(p)
-                    self.buf = b"\n".join(lines[i + 1 :])
+            version_line, self.buf = self.buf.split(b"\n", 1)
+            # Since the line was split on '\n' and most of the time
+            # it uses '\r\n' we may get an extra '\r'.
+            version_line = version_line.rstrip(b"\r")
+
+            if b"\x00" in version_line:
+                # RFC 4253 Section 4.2: The null character MUST NOT be sent.
+                self.sendDisconnect(
+                    DISCONNECT_CONNECTION_LOST,
+                    b"Invalid peer version. Null character found.",
+                )
+                return
+
+            self.otherVersionString = version_line
+            self.gotVersion = True
+
+            remoteVersion = self.otherVersionString.split(b"-")[1]
+            if remoteVersion not in self.supportedVersions:
+                self._unsupportedVersionReceived(remoteVersion)
+                return
+
         packet = self.getPacket()
         while packet:
             messageNum = ord(packet[0:1])
--- a/src/twisted/conch/test/test_transport.py
+++ b/src/twisted/conch/test/test_transport.py
@@ -515,10 +515,48 @@ class BaseSSHTransportTests(BaseSSHTransportBaseCase, TransportTestCase):
         )
         self.assertRegex(softwareVersion, softwareVersionRegex)
 
+    def test_dataReceiveVersionNullCharacter(self):
+        """
+        It will reject the connection when null characters are found in the
+        peer's SSH version string.
+        """
+        sut = MockTransportBase()
+        sut.makeConnection(self.transport)
+
+        # White the new line is not received, all looks fine so far.
+        sut.dataReceived(b"SSH-2-to-be-continued-")
+        self.assertFalse(self.transport.disconnecting)
+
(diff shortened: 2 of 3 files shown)


adiroiban:
This is part-2 of the SSH version string security issue.

In practice, most SSH implementations are sending just their SSH version string and then wait for more data.

But, an SSH implementation can also send its SSH version string and the list of supported algorithm all in the same package.

Most of the time, the list of algorithms host keys + kex + cipher + hmac + compressions is less than 2K .

But, if a peer supports many ciphers ... or aliases , it can send more than 4K in the first package without breaking the SSH protocol.

So... maybe we should not release 22.2.0 as it is now and have this merged and do a new release candidate.

src/twisted/conch/ssh/transport.py
```diff
@@ -730,34 +730,56 @@ def dataReceived(self, data):
...
+            if b"\n" not in self.buf:
+                # We don't have yet the full version string.
+                return
+
+            if not self.buf.startswith(b"SSH-"):
+                # The identification string MUST be
+                # SSH-protoversion-softwareversion SP comments CR LF
+                self.sendDisconnect(
+                    DISCONNECT_CONNECTION_LOST, b"Invalid peer version format."
+                )
```
cjwatson, line 755: This seems to violate a "MUST" in RFC 4253 section 4.2, in the case where the peer is a server:

```
   The server MAY send other lines of data before sending the version
   string.  Each line SHOULD be terminated by a Carriage Return and Line
   Feed.  Such lines MUST NOT begin with "SSH-", and SHOULD be encoded
   in ISO-10646 UTF-8 [RFC3629] (language is not specified).  Clients
   MUST be able to process such lines.  Such lines MAY be silently
   ignored, or MAY be displayed to the client user.  If they are
   displayed, control character filtering, as discussed in [SSH-ARCH],
   SHOULD be used.  The primary use of this feature is to allow TCP-
   wrappers to display an error message before disconnecting.
```

It would be OK to accept lines before the identification string only in the client case, though; and it may be permissible to limit the number of such lines even in the client case.  For example, OpenSSH does this, where `SSH_MAX_PRE_BANNER_LINES` is defined to 1024:

```c
                /* Accept lines before banner only on client */
                if (ssh->kex->server || ++n > SSH_MAX_PRE_BANNER_LINES) {
  bad:
                        if ((r = sshbuf_put(ssh_packet_get_output(ssh),
                            mismatch, strlen(mismatch))) != 0)
                                return r;
                        return SSH_ERR_NO_PROTOCOL_VERSION;
                }
```

Clients don't have the same DoS concerns that servers have.

src/twisted/conch/ssh/transport.py
```diff
@@ -730,34 +730,56 @@ def dataReceived(self, data):
...
+                # yet it doesn't look like the version string line
+                # is in the received data.
+                # RFC 4253 section 4.2
+                # The maximum length of the string is 255 characters,
+                # including the Carriage Return and Line Feed.
                 self.sendDisconnect(
                     DISCONNECT_CONNECTION_LOST,
-                    b"Peer version string longer than 4KB. "
-                    b"Preventing a denial of service attack.",
+                    b"Invalid peer version. Long value. Preventing a denial of service attack.",
```
cjwatson, line 742:
```suggestion
                    b"Peer identification string not found in first 255 characters.",
```

All other things being equal I prefer using the "identification string" language from the RFC.  I don't object to the DoS attack language, but it seems unnecessary - I'm sure there are various errors that serve the purpose of preventing DoS attacks, and we probably shouldn't imply in the logs that only this one does so.

src/twisted/conch/test/test_transport.py
```diff
@@ -515,22 +515,60 @@ def test_sendVersion(self):
...
+        self.assertTrue(self.transport.disconnecting)
+        self.assertIn(
+            b"Invalid peer version. Null character found.", self.transport.value()
+        )
+
+    def test_dataReceiveVersionInvalidVersionFormat(self):
+        """
+        It will reject the connection when remote peer SSH format doesn't start
+        with `SSH-`.
+        """
```
cjwatson, line 542: Can you make this test server-specific, and add a client-specific test that ensures that lines before the identification string are allowed?

src/twisted/conch/test/test_transport.py
```diff
@@ -515,22 +515,60 @@ def test_sendVersion(self):
...
+
+        self.assertTrue(self.transport.disconnecting)
+        self.assertIn(b"Invalid peer version format.", self.transport.value())
+
     def test_dataReceiveVersionNotSentMemoryDOS(self):
         """
         When the peer is not sending its SSH version but keeps sending data,
-        the connection is disconnected after 4KB to prevent buffering too
+        the connection is disconnected after 255B to prevent buffering too
         much and running our of memory.
```
cjwatson, line 560: Could you fix this typo for "running out of memory" while you're here?

cjwatson (changes requested):

src/twisted/conch/test/test_transport.py
```diff
@@ -1116,24 +1154,20 @@ def stubLoseConnection():
 
         testBad(b"SSH-1.5-OpenSSH")
         testBad(b"SSH-3.0-Twisted")
-        testBad(b"GET / HTTP/1.1")
 
     def test_dataBeforeVersion(self):
         """
-        Test that the transport ignores data sent before the version string.
+        Any data sent before the version string, is considered a protocol
```
cjwatson, line 1160:
```suggestion
        Any data sent before the version string is considered a protocol
```

src/twisted/conch/ssh/transport.py
adiroiban: Many thanks. I missed that part of the RFC.

I saw that there was an explicit test for this behaviour but I didn't knew why.

I will revert and document this. Thanks!

src/twisted/conch/ssh/transport.py
adiroiban: Agree. I will use your suggestion. Agree that DoS is not needed. Thanks!

src/twisted/conch/test/test_transport.py
adiroiban: Yes. Will do.

src/twisted/conch/ssh/transport.py
graingert: > It would be OK to accept lines before the identification string only in the client case, though; and it may be permissible to limit the number of such lines even in the client case. For example, OpenSSH does this, where SSH_MAX_PRE_BANNER_LINES is defined to 1024:

previously  SSHTransportBase [buffered the data then dropped the lines](https://github.com/twisted/twisted/pull/1697/files#diff-92b57e49a6a90fe8379778df32a08366febe251584794e1aa01ffd878a674690L748-L750), when restoring this behaviour it would be better if any data that doesn't start b"SSH-" is dropped immediately

adiroiban:
testing the automatic PR labeling 

needs-changes

PR #1699, opened 2022-02-18
#9443 IRCClient: Allow trying message decoding with multiple codecs

DefaultUser:
## Scope and purpose

The IRC protocol doesn't specify encoding, so occasionally codecs other
than "utf-8" are used. This patch allows IRCClients to specify multiple
possible codecs to use when decoding incoming messages.
Additionally, if a message can't be decoded with any of the specified
codecs, a final attempt with a non-strict Error Handler is attempted.
This defaults to "replace", but can be replaced by the user.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/9443

```

Author: DefaultUser
Reviewer:
Fixes: #9443
```


src/twisted/words/protocols/irc.py
```diff
@@ -2651,8 +2654,17 @@ def dataReceived(self, data):
 
     def lineReceived(self, line):
         if bytes != str and isinstance(line, bytes):
```
twm: Could you delete always-true condition while you're here? This is a Python 2 &rarr; 3 porting artifact.

src/twisted/words/protocols/irc.py
```diff
@@ -1238,6 +1238,9 @@ class IRCClient(basic.LineReceiver):
     _heartbeat = None
     heartbeatInterval = 120
 
+    decode_codecs = ["utf-8"]  # IRC protocol doesn't specify encoding
+    decode_fallback_errorhandling = "replace"  # see bytes.decode
```
twm: The names must match the Twisted convention: `decodeCodecs`, `decodeFallbackErrors`.

You also have to add `@ivar` annotations for them to the class docstring.

src/twisted/words/protocols/irc.py
```diff
@@ -2651,8 +2654,17 @@ def dataReceived(self, data):
...
         if bytes != str and isinstance(line, bytes):
-            # decode bytes from transport to unicode
-            line = line.decode("utf-8")
+            # decode bytes from transport to str
+            for codec in self.decode_codecs:
+                try:
+                    line = line.decode(codec)
+                    break
+                except ValueError:
+                    log.msg(f"Couldn't decode line '{line}' with codec {codec}")
```
twm: We definitely want to use `repr()` on the line here, as it could contain control characters etc.

```suggestion
                    log.msg(f"Couldn't decode line {line!r} with codec {codec}")
```

I'd also say this should be a debug message but the logging in this module hasn't been updated to use `twisted.logger` yet so no need to make that change right now.

src/twisted/words/test/test_irc.py
```diff
@@ -2572,6 +2572,71 @@ def test_dccChatIndecipherablePort(self):
...
+class ClientDecodingTests(IRCTestCase):
+    """
+    Tests for L{IRCClient} receiving data in different encodings.
+    """
+
+    def setUp(self):
+        methods = ["handleCommand", "badMessage"]
+        self.client = CollectorClient(methods)
+
+    def test_decoding_unspecified_codec(self):
```
twm: Thank you for writing such nice docstrings!

The test method names (and locals) should also be camelCase. Really minor nit: you can [remove "Test that" from the docstring](https://jml.io/pages/test-docstrings.html) as it's redundant.

src/twisted/words/test/test_irc.py
```diff
@@ -2572,6 +2572,71 @@ def test_dccChatIndecipherablePort(self):
...
+    def test_decoding_errorhandling(self):
+        """
+        Test that fallback errorhandling for message decoding can be overridden
+        by the user.
+        """
+        self.client.decode_codecs = ["utf-8"]
+        self.client.decode_fallback_errorhandling = "strict"
+        prefix = ":foo!bar@baz"
+        line_bytes = f"{prefix} PRIVMSG foo :ä".encode("latin-1")
+        self.assertRaises(UnicodeDecodeError, self.client.lineReceived, line_bytes)
```
twm: I'm a little torn here. It seems like this codepath should call `self.badMessage`. However, `badMessage`'s first argument is `str` (mis-documented as `bytes` :see_no_evil:) so doing so wouldn't be compatible. Given the default error handling strategy is `replace` I think this'll do okay.

twm (changes requested):
Hi @DefaultUser, sorry it took me so long to follow up here. The substance of these changes looks good to me! There are a few superficial things to fix, then we can get this merged.

Thanks again for giving IRCClient some love!

DefaultUser:
Hi @twm, I've incorporated the changes you requested. I'm happy to help if there is anything more that needs to be changed.

src/twisted/words/test/test_irc.py
```diff
@@ -2572,6 +2572,67 @@ def test_dccChatIndecipherablePort(self):
...
+    def testDecodingSpecifiedCodec(self):
+        """
+        Message decoding will try the specified codecs before using
+        the fallback errorhandling.
+        """
+        self.client.decodeCodecs = ["utf-8", "latin-1"]
+        self.client.decodeFallbackErrorhandling = IRCClient.decodeFallbackErrorhandling
+        prefix = ":foo!bar@baz"
+        line_bytes = f"{prefix} PRIVMSG foo :ä".encode("latin-1")
+        self.client.lineReceived(line_bytes)
```
twm: So right now our standard requires ASCII... I filed #11723 to change this. Also the variable should be named `lineBytes`.

The `\N{}` syntax can be handy here, though in this case it might be best to use a `b"\xe4"` literal as it's more obvious that this isn't valid UTF-8.

```suggestion
        lineBytes = f"{prefix} PRIVMSG foo :\N{LATIN SMALL LETTER A WITH DIAERESIS}".encode("latin-1")
        self.client.lineReceived(lineBytes)
```

The same goes for the other tests.

src/twisted/words/newsfragments/9443.bugfix
```diff
@@ -0,0 +1 @@
+twisted.words.protocols.irc.IRCClient now allows specifying multiple codecs that are tried when decoding messages and will use a non-strict Error Handler as last resort.
```
twm: Not sure why Error Handler is capitalized?

```suggestion
twisted.words.protocols.irc.IRCClient now allows specifying multiple codecs to try when decoding messages and will use a non-strict error handler as last resort.
```

src/twisted/words/protocols/irc.py
```diff
@@ -1238,6 +1246,9 @@ class IRCClient(basic.LineReceiver):
     _heartbeat = None
     heartbeatInterval = 120
 
+    decodeCodecs: list[str] = ["utf-8"]  # IRC protocol doesn't specify encoding
+    decodeFallbackErrorhandling: str = "replace"  # see bytes.decode
```
twm: This comment would be more useful in the docstring where it could help folks trying to use the class. Or maybe link directly to https://docs.python.org/3/library/codecs.html#error-handlers ?

src/twisted/words/protocols/irc.py
```diff
@@ -2650,9 +2661,18 @@ def dataReceived(self, data):
         basic.LineReceiver.dataReceived(self, data)
 
     def lineReceived(self, line):
-        if bytes != str and isinstance(line, bytes):
-            # decode bytes from transport to unicode
-            line = line.decode("utf-8")
+        if isinstance(line, bytes):
```
twm: It will _always_ be `bytes` &mdash; please remove this `isinstance` condition.

twm (changes requested):
Hey @DefaultUser, thanks for following up! This is very close, just a few minor changes left.

src/twisted/words/test/test_irc.py
DefaultUser: I didn't know about the `\N{}` syntax. You're right about using the bytes literal directly being more obvious, so I changed these test that way.

src/twisted/words/newsfragments/9443.bugfix
DefaultUser: No reason to have it capitalized. It's simply muscle memory since different languages have different rules on capitalization.

src/twisted/words/protocols/irc.py
DefaultUser: Moved to the docstring.

src/twisted/words/protocols/irc.py
DefaultUser: Removing this check broke some other tests that called `lineReceived` directly, but I fixed those.

src/twisted/words/protocols/irc.py
```diff
@@ -1197,6 +1197,16 @@ class IRCClient(basic.LineReceiver):
     @ivar heartbeatInterval: Interval, in seconds, to send I{PING} messages to
         the server as a form of keepalive, defaults to 120 seconds. Use L{None}
         to disable the heartbeat.
+
+    @type decodeCodecs: C{list} of C{str}
```
twm, line 1201: FYI for the future, you can omit the `@type` when there's a type annotation on the variable itself &mdash; Pydoctor will use the annotation now!

src/twisted/words/protocols/irc.py
```diff
@@ -2643,16 +2656,22 @@ def connectionMade(self):
         if self.performLogin:
             self.register(self.nickname)
 
-    def dataReceived(self, data):
+    def dataReceived(self, data: bytes | str):
```
twm: :grimacing: This is always `bytes`!

```suggestion
    def dataReceived(self, data: bytes) -> None:
```

src/twisted/words/protocols/irc.py
twm: Perfect, thank you!

twm (changes requested):
Thanks for your patience @DefaultUser! This looks good to go apart from the type annotation on `dataReceived`. Could you fix that and address any type issues that it shakes out? Then this should be good to land.

src/twisted/words/protocols/irc.py
```diff
@@ -1238,6 +1248,9 @@ class IRCClient(basic.LineReceiver):
     _heartbeat = None
     heartbeatInterval = 120
 
+    decodeCodecs: list[str] = ["utf-8"]
```
twm: Oh, and we still have to use `List[str]` because we support old Python yet.

src/twisted/words/protocols/irc.py
DefaultUser: Yes, you're totally right. Fixed.

src/twisted/words/protocols/irc.py
DefaultUser: I forgot that this is a quite new feature.

DefaultUser:
I just realized that I forgot to commit and push a change last year.

Edit: I should have run the tests locally. I'll fix those soon.

DefaultUser:
Hi @twm,
the remaining failing tests seem unrelated to my changes.
I hope that now all the issues are resolved.

twm:
@DefaultUser It looks like the MyPy check is failing.

DefaultUser:
@twm : I thought I fixed this previously. Anyways, now it's fixed.
I'm not sure why the pypy tests are failing, but this seems unrelated. I also got some test failures in unrelated tests (logging and conch) when running all tests locally, but they don't show up here, so I hope that everything is all right now.

glyph:
@DefaultUser In the intervening year we have made some test reliability improvements so I am updating open PRs with failing tests that might be flakes. Hopefully you will get luckier with this run at CI.

PR #1700, opened 2022-02-19
[#10285] IRCClient: Split NOTICEs the same way as PRIVMSGs before sending

DefaultUser:
## Scope and purpose

NOTICEs should follow the same length restrictions as PRIVMSGs.
Additionally they should also be automatically split on newline
characters to have a consistent interface with "msg".

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10285

```

Author: DefaultUser
Reviewer:
Fixes: ticket: 10285
```


src/twisted/words/protocols/irc.py
```diff
@@ -69,6 +70,9 @@
 CHANNEL_PREFIXES = "&#!+"
 
 
+MSG_OR_NOTICE = enum.Enum("MSG_OR_NOTICE", "PRIVMSG NOTICE")
```
twm: As this enum isn't used by any public API surface it should have a private name.

I think it'd be fine &mdash; even preferable &mdash; to pass the strings `'PRIVMSG'` and `'NOTICE'` directly without the enum at all, again because it's all private. Enums are slow after all.

src/twisted/words/protocols/irc.py
```diff
@@ -1713,15 +1717,18 @@ def _safeMaximumLineLength(self, command):
         fudge = 10
         return MAX_COMMAND_LENGTH - len(theoretical) - fudge
 
-    def msg(self, user, message, length=None):
+    def _send_msg_or_notice(self, msg_or_notice, user, message, length=None):
```
twm: By the Twisted naming convention this should be `_sendMsgOrNotice`. The same applies to the `msg_or_notice` argument &mdash; `msgOrNotice` &mdash; though the naming is awkward in both cases. I see the IRC3 spec refers to PRIVMSG and NOTICE as "messages". Perhaps call the function `_sendMessage` and the argument `command`?

src/twisted/words/test/test_irc.py
```diff
@@ -1846,55 +1846,82 @@ def test_invalidMaxLength(self):
 
     def test_multipleLine(self):
         """
-        Messages longer than the C{length} parameter to L{IRCClient.msg} will
-        be split and sent in multiple commands.
+        Messages and notices longer than the C{length} parameter to
```
twm: Please split these tests apart so that each one tests a single thing: `msg` or `notice`, or convert them to be table driven. The former is probably easier. As they are now constructed they're two tests glommed together, but we want to test the behavior of the API surface separately.

twm (changes requested):
Hi, thank you for giving IRCClient some love! The mechanics of this look good, but there are some minor changes to be made before merge. Thanks for your persistence with the process!

src/twisted/words/protocols/irc.py
DefaultUser: I wanted to avoid ambiguous names, but if the specs refer to both as "messages", it should be allright.

src/twisted/words/protocols/irc.py
DefaultUser: I prefer passing enums over string constants as it's easier to spot typos that way, but I guess that's not a prblem here as the usage is very limited.

twm (approved):
Looks good, and thanks again! I'll merge this once the builds finish.

PR #1701, opened 2022-03-04
10313: Send exit-status/exit-signal on EOF to SSHSession again

cjwatson:
## Scope and purpose

When I tried to upgrade https://git.launchpad.net/turnip (the hosting backend for git.launchpad.net) to a backport of https://github.com/twisted/twisted/pull/1696, I found that its SSH frontend tests failed with a series of extremely opaque errors which boiled down to git saying "fatal: the remote end hung up unexpectedly", and which on detailed comparison were because the "exit-status" channel request wasn't being sent after a "git upload-pack" subprocess exited.

This turns out to be because I drew an incorrect parallelism between SSHSession.eofReceived and SSHSession.closed when fixing https://twistedmatrix.com/trac/ticket/10308: while it is indeed necessary to lose the client transport's connection in SSHSession.closed even if there's a session adapter as well, it's *not* necessary to send a close message in SSHSession.eofReceived if there's a session adapter.  Firstly, the session adapter will typically already send a close message via SSHChannel.loseConnection once it's finished cleaning up (and would presumably have done so prior to the fix for https://twistedmatrix.com/trac/ticket/10308).  Secondly, and more critically, the SSH_MSG_CHANNEL_CLOSE message sent by SSHConnection.sendClose is a request to terminate the channel, and it's supposed to be the last thing that the party sending it sends on that channel; SSHConnection refuses to send any other messages after sending a close message.  As a result, any asynchronous cleanup work done by SSHSession.eofReceived, even a Deferred that fires on the next iteration of the reactor, will fail to send cleanup messages such as "exit-status" or "exit-signal", which the peer may interpret as abrupt termination; that is indeed what git does.

We failed to notice this because the session tests use a StubConnection which behaves differently: it's happy to continue to "send" messages (i.e. record them in its various dicts of messages that would have been sent to a real peer by the code under test) even if the channel has been closed, and thus fails to be an accurate stub.  Fixing the stub immediately exposes the failure in SessionInterfaceTests.test_requestExecWithData.

Reverting part of my previous change means that SSHSession correctly sends "exit-status"/"exit-signal" messages again when a program it started in response to a "shell" or "exec" request exits.  I've also checked that this doesn't cause https://twistedmatrix.com/trac/ticket/10308 to recur; only the change to SSHSession.closed was in fact needed there.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10313
* [x] I have created a newsfragment in src/twisted/conch/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles)) (empty because this regression never made it into a release)

```

Author: cjwatson
Fixes: ticket:10313

Send exit-status/exit-signal on EOF to SSHSession again
```

### DIFF
--- a/src/twisted/conch/ssh/session.py
+++ b/src/twisted/conch/ssh/session.py
@@ -180,10 +180,14 @@ class SSHSession(channel.SSHChannel):
             log.warn("Weird extended data: {dataType}", dataType=dataType)
 
     def eofReceived(self):
-        if self.client:
-            self.conn.sendClose(self)
+        # If we have a session, tell it that EOF has been received and
+        # expect it to send a close message (it may need to send other
+        # messages such as exit-status or exit-signal first).  If we don't
+        # have a session, then just send a close message directly.
         if self.session:
             self.session.eofReceived()
+        elif self.client:
+            self.conn.sendClose(self)
 
     def closed(self):
         if self.client and self.client.transport:
--- a/src/twisted/conch/test/test_session.py
+++ b/src/twisted/conch/test/test_session.py
@@ -342,18 +342,24 @@ class StubConnection:
         """
         Record the sent data.
         """
+        if self.closes.get(channel):
+            return
         self.data.setdefault(channel, []).append(data)
 
     def sendExtendedData(self, channel, type, data):
         """
         Record the sent extended data.
         """
+        if self.closes.get(channel):
+            return
         self.extData.setdefault(channel, []).append((type, data))
 
     def sendRequest(self, channel, request, data, wantReply=False):
         """
         Record the sent channel request.
         """
+        if self.closes.get(channel):
+            return
         self.requests.setdefault(channel, []).append((request, data, wantReply))
         if wantReply:
             return defer.succeed(None)
@@ -362,6 +368,8 @@ class StubConnection:
         """
         Record the sent EOF.
         """
+        if self.closes.get(channel):
+            return
         self.eofs[channel] = True
 
     def sendClose(self, channel):


src/twisted/conch/test/test_session.py
```diff
@@ -342,18 +342,24 @@ def sendData(self, channel, data):
         """
         Record the sent data.
         """
+        if self.closes.get(channel):
+            return
```
adiroiban, line 346: I guess that this is OK ... and if needed, we can later extend  `StubConnection` to raise an exception if something is sent after the connection is closed.

src/twisted/conch/ssh/session.py
```diff
@@ -180,10 +180,14 @@ def extReceived(self, dataType, data):
             log.warn("Weird extended data: {dataType}", dataType=dataType)
 
     def eofReceived(self):
-        if self.client:
-            self.conn.sendClose(self)
+        # If we have a session, tell it that EOF has been received and
```
adiroiban, line 183: Thanks.

I guess that this is OK. As mentioned in the other PR, I don't have a good idea for how to untangle this.

I remember that I was confused by using `self.conn.sendClose(self) on self.client` vs `self.session.eofReceived() on self.session` 

The API doc for `SSHChannle.eofRecieve` is `Called when the other side will send no more data.`

But I guess the main confusion is due to the fact that we have both `self.client` and `self.session`  ... and that Conch SSH documentation is not ideal.

--------

For my project, I have patched to code to only use `self.client` for everything.

The `self.client` can be a RFC session (exec, pty, subsystem) or port forwarding "session". 

And then each `self.client` depending on its type will call  `self.conn.sendClose(self)` if needed.

Also each `self.client` has the `eofReceived` and `closed` method bubbled/forwarded.

I can update my code to use `self.session` name instead of `self.client` and that should work with this update.

But maybe it would help to recommend using only `self.session` for everything  and define a common API.

My understanding is that a SSH channel can be closed in 2 ways

* Only SSH_MSG_CHANNEL_CLOSE
* SSH_MSG_CHANNEL_EOF followed by SSH_MSG_CHANNEL_CLOSE 

also, a SSH channel can only "host" a single session.
A SSH channel can't host/multiple multiple sessions or have both a session and a port forwarding connection - https://datatracker.ietf.org/doc/html/rfc4254#section-5

adiroiban (approved):
Thanks. All good.
Only side comments to make sure I understood the RFC.

src/twisted/conch/test/test_session.py
cjwatson: I agree that might be worth it at some point to catch mistakes, though we'd have to change `SSHConnection.sendData` (and maybe other methods) first, and I've no idea what the fallout of that would be.

src/twisted/conch/ssh/session.py
cjwatson: Yeah, your approach broadly seems better - it's just the difficult compatibility issues of getting there from here.  As I mentioned in my previous PR (#1696), the most practical approach may be to move the EOF/close handling into the process protocol (`self.client`) and just create an unconnected process protocol earlier for cases such as environment variable passing requests in advance of setting up the real session.  I don't think this PR would make that plan any harder to execute; it would just need the corresponding change to the client's `eofReceived` to only call `conn.sendClose` if it doesn't have a session.

But for now, this restores previous behaviour ...

PR #1702, opened 2022-03-11
10312 trial -jN -u iterations

ihaywood3:
https://twistedmatrix.com/trac/ticket/10312

this is a minimalist solution to the bug


adiroiban (approved):
Looks good.Thanks. I think that this can be landed once the tests are green.

PR #1708, opened 2022-03-27
10320 discard async cleanups when running them

exarkun:
## Scope and purpose

See https://twistedmatrix.com/trac/ticket/10320

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10320

```

Author: exarkun
Fixes: ticket:103200

Prevent cleanups from accumulating on TestCase.
```

### DIFF
--- a/src/twisted/trial/_asynctest.py
+++ b/src/twisted/trial/_asynctest.py
@@ -199,7 +199,8 @@ class TestCase(SynchronousTestCase):
         object.
         """
         failures = []
-        for func, args, kwargs in self._cleanups[::-1]:
+        while len(self._cleanups) > 0:
+            func, args, kwargs = self._cleanups.pop()
             try:
                 yield func(*args, **kwargs)
             except Exception:
--- a/src/twisted/trial/newsfragments/10320.bugfix
+++ b/src/twisted/trial/newsfragments/10320.bugfix
@@ -1,2 +1,1 @@
-
 twisted.trial.unittest.TestCase now discards cleanup functions after running them.  Notably, this prevents them from being run an ever growing number of times with `trial -u ...`.
--- a/src/twisted/trial/newsfragments/10320.bugfix
+++ b/src/twisted/trial/newsfragments/10320.bugfix
@@ -1,0 +1,2 @@
+
+twisted.trial.unittest.TestCase now discards cleanup functions after running them.  Notably, this prevents them from being run an ever growing number of times with `trial -u ...`.
--- a/src/twisted/trial/test/test_tests.py
+++ b/src/twisted/trial/test/test_tests.py
@@ -962,6 +962,21 @@ class AddCleanupMixin:
         self.assertEqual(error1.getErrorMessage(), "bar")
         self.assertEqual(error2.getErrorMessage(), "foo")
 
+    def test_cleanupRunsOnce(self):
+        """
+        A function registered as a cleanup is run once.
+        """
+        cleanups = []
+        self.test.addCleanup(lambda: cleanups.append(stage))
+        # It should get run this time.
+        stage = "first"
+        self.test.run(self.result)
+        # It should not get run the next time since it has not been
+        # re-registered.
+        stage = "second"
+        self.test.run(self.result)
+        self.assertEqual(cleanups, ["first"])
+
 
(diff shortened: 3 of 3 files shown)


meejah (approved):
LGTM :+1:

src/twisted/trial/newsfragments/10320.bugfix
```diff
@@ -0,0 +1,2 @@
+
+twisted.trial.unittest.TestCase now discards cleanup functions after running them.  Notably, this prevents them from being run an ever growing number of times with `trial -u ...`.
```
glyph:
```suggestion
twisted.trial.unittest.TestCase now discards cleanup functions after running them.  Notably, this prevents them from being run an ever growing number of times with `trial -u ...`.
```

glyph: Just removing a spurious blank line here.

src/twisted/trial/_asynctest.py
```diff
@@ -199,7 +199,8 @@ def deferRunCleanups(self, ignored, result):
         object.
         """
         failures = []
-        for func, args, kwargs in self._cleanups[::-1]:
+        while len(self._cleanups) > 0:
+            func, args, kwargs = self._cleanups.pop()
```
glyph, line 203: An idle thought for future reference; should we have a `destructivelyIterate` utility somewhere? This idiom seems common; would it be worthwhile to use it with `for` syntax?

src/twisted/trial/test/test_tests.py
```diff
@@ -962,6 +962,21 @@ def test_multipleErrorsReported(self):
...
+        cleanups = []
+        self.test.addCleanup(lambda: cleanups.append(stage))
+        # It should get run this time.
+        stage = "first"
+        self.test.run(self.result)
+        # It should not get run the next time since it has not been
+        # re-registered.
+        stage = "second"
+        self.test.run(self.result)
+        self.assertEqual(cleanups, ["first"])
```
glyph, line 978: This test is a just a little confusing because it took me a moment to realize that `addCleanup` is expected to be invoked *during* `run`, and not usually like this.  A more realistic test would be something that adds a cleanup each time *during* the synthetic test run, but asserts that the appropriate cleanup call is paired with each run and the streams don't get crossed.

This test would also pass if, for example, all cleanups just didn't run after the first test run.  But this is a pretty minor critique so I would not call it a blocker; if you want to clean up the test just a little before landing, go for it

glyph (approved):
LGTM too, just making it official

src/twisted/trial/_asynctest.py
meejah: maybe `consume(func, iter)`...? or `consumeIterator(func, iter)` .. or maybe the args go the other way, I didn't do a survey.

exarkun: I'd probably lean towards a refactoring that avoids the need to mutate state wherever possible.  Since mutation is used heavily throughout trial and unittest, this is probably at least a huge amount of work, though.

It's not obvious to me how `destructivelyIterate` would work with the order reversal here.  I don't think you could make `destructivelyIterate(self_cleanups[::-1])` work.  I suppose you could always add more parameters, `destructivelyIterate(self._cleanups, reverse=True)` or something.  Maybe if `_cleanups` were a stack instead of a list it would go a bit more smoothly.  I haven't looked over the rest of the implementation to see what other trouble such a change might make.

Likewise I'm not sure how `consume(func, iter(self._cleanups))` would work.  If a function is going to help, I think it needs the list object so it can pop items off of it.  (If passing in the list itself was the idea then it seems like it ends up equivalent to `destructivelyIterate`?  But maybe I'm overlooking something.)

src/twisted/trial/test/test_tests.py
exarkun: I agree that it's weird.  It seems to be the prevailing idiom for this test suite, though, so I'm inclined to go with consistency for now.  I would approve of a PR that rewrites all of `AddCleanupMixin` to use more realistic test cases (maybe I'll even get to it after some of the other functional fixes I'm working on).

exarkun:
Thanks for the reviews @meejah and @glyph

PR #1711, opened 2022-03-28
10322 deduplicate test.log opening code

exarkun:
## Scope and purpose

See https://twistedmatrix.com/trac/ticket/10322

I opted to make both codepaths open the file in append mode to avoid accidental data loss and to make them both always use UTF-8 instead of letting one of them rely on the platform to choose to make the behavior more consistent.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10322

```

Author: exarkun
Fixes: ticket:10322

Remove duplication and inconsistencies in trial's two `test.log` opening codepaths.
```

### DIFF
--- a/src/twisted/trial/util.py
+++ b/src/twisted/trial/util.py
@@ -18,8 +18,8 @@ Maintainer: Jonathan Lange
     asynchronous (ie, Deferred-returning) test methods, in seconds.
 """
 
-
 from random import randrange
+from typing import TextIO
 
 from twisted.internet import interfaces, utils
 from twisted.python.failure import Failure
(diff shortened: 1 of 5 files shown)


src/twisted/trial/util.py
```diff
@@ -387,3 +387,12 @@ def _listToPhrase(things, finalDelimiter, delimiter=", "):
             finalDelimiter,
             strThings[-1],
         )
+
+
+def openTestLog(path: FilePath) -> TextIO:
```
adiroiban, line 392: Should this be a public API? Just asking?

Very minor comment.

Maybe make it private and give it a generic name, so that it can also be used for non log files... if needed.

```suggestion
def _openUTF8Append(path: FilePath) -> TextIO:
```

adiroiban (approved):
Looks good. thanks for the cleanup. only minor comments.

src/twisted/trial/util.py
exarkun: Maybe `twisted.trial.util` should be deprecated and replaced by `twisted.trial._util`.  The current module docstring effectively says it is all meant to be private.

adiroiban: Ok. Just merge. don't worry. Thanks! We can deprecated the whole `util` later... and in general, I think that with `trial` code we can be more relaxed in terms of backward compatibility, as I expect that nobody uses trial code for production code.

PR #1718, opened 2022-04-11
10327 Support coroutine results in maybeDeferred

exarkun:
## Scope and purpose

https://twistedmatrix.com/trac/ticket/10327

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10327

```

Author: exarkun
Fixes: ticket:10327

Support a coroutine result in `maybeDeferred` by applying `Deferred.fromCoroutine`.
This makes it possible to convert functions which may or may not return a `Deferred`
(and which are called by `maybeDeferred`) into functions defined with `async def`
with minimal or no additional effort.

A partial list of APIs in Twisted which this causes to begin supporting coroutines is:

* twisted.internet.task.LoopingCall
* twisted.internet.threads.blockingCallFromThread 
* twisted.protocols.ident.IdentServer.lookup
* twisted.protocols.amp.IResponderLocator.locateResponder 
* twisted.protocols.ftp.FTP.processCommand (ftp_* methods)
* twisted.protocols.postfix.PostfixTCPMapServer's "factory.get"
* twisted.application.internet.ClientService
* twisted.application.service.IService.stopService
* twisted.conch.scripts.cftp.StdioClient (cmd_* methods)
* twisted.conch.ssh.userauth.SSHUserAuthClient.getPublicKey
* twisted.conch.ssh.connection.SSHChannel.requestReceived
* twisted.conch.interfaces.ISFTPServer.openFile
* all twisted.conch.interfaces.ISFTPFile methods
* twisted.cred.credentials.IUsernamePassword.checkPassword
* twisted.cred.credentials.IUsernameHashedPassword.checkPassword
* twisted.cred.credentials.IUsernameDigestHash.checkHash
* twisted.cred.checkers.ICredentialsChecker.requestAvatarId
* twisted.internet.interfaces.IListeningPort.stopListening

* Some APIs in twisted.mail, twisted.names, twisted.spread
* Some APIs in twisted.web.soap, twisted.web.xmlrpc
* Some APIs twisted.words.service
```

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -172,7 +172,7 @@ def _isAsyncDefValue(o: Any) -> bool:
     Determine whether to apply "coroutine" handling to a value in
     L{maybeDeferred}.
 
-    @return: C{True} if L{o} came from a function defined using C{async def},
+    @return: C{True} if C{o} came from a function defined using C{async def},
         C{False} otherwise.
     """
     # A note on how we identify such values...
@@ -202,17 +202,18 @@ def maybeDeferred(
 
     Call the given function with the given arguments.  Then:
 
-    - If the returned object is a L{Deferred}, return it.
+      - If the returned object is a L{Deferred}, return it.
 
-    - If the returned object is a L{Failure}, wrap it with L{fail} and return it.
+      - If the returned object is a L{Failure}, wrap it with L{fail} and
+        return it.
 
-    - If the returned object is a L{types.CoroutineType}, wrap it with
-      L{Deferred.fromCoroutine} and return it.
+      - If the returned object is a L{types.CoroutineType}, wrap it with
+        L{Deferred.fromCoroutine} and return it.
 
-    - Otherwise, wrap it in L{succeed} and return it.
+      - Otherwise, wrap it in L{succeed} and return it.
 
-    - If an exception is raised, convert it to a L{Failure}, wrap it in
-      L{fail}, and then return it.
+      - If an exception is raised, convert it to a L{Failure}, wrap it in
+        L{fail}, and then return it.
 
     @param f: The callable to invoke
     @param args: The arguments to pass to C{f}
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -167,33 +167,6 @@ def execute(
         return succeed(result)
 
 
-def _isAsyncDefValue(o: Any) -> bool:
-    """
-    Determine whether to apply "coroutine" handling to a value in
-    L{maybeDeferred}.
-
-    @return: C{True} if C{o} came from a function defined using C{async def},
-        C{False} otherwise.
-    """
-    # A note on how we identify such values...
-    #
-    # inspect.iscoroutinefunction(f) should be the simplest and easiest way to
-    # determine if we want to apply coroutine handling.  However, the value
-    # may be returned by a regular function that calls a coroutine function
-    # and returns its result.  It would be confusing if cases like this led to
-    # different handling of the coroutine (even though it is a mistake to have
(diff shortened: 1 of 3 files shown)


src/twisted/internet/defer.py
```diff
@@ -167,17 +167,53 @@ def execute(
         return succeed(result)
 
 
+def _isAsyncDefValue(o: Any) -> bool:
```
glyph: This is sort of moot in this specific case, but in general if you can get away with `o: object`  rather than `o: Any` you should generally go that way

src/twisted/internet/defer.py
```diff
@@ -195,6 +231,8 @@ def maybeDeferred(
         return result
     elif isinstance(result, Failure):
         return fail(result)
+    elif _isAsyncDefValue(result):
+        return Deferred.fromCoroutine(result)
```
glyph: Pushing this into a separate function seems oddly asymmetrical here, and given that we're already in for a penny here on `isinstance` calls, it seems like we might as well be in for a pound?  We already dropped the `type(result) is Deferred` optimization many years ago because it raised too many odd edge cases.

I'd rather leave this as a standard-looking `isinstance` ladder to make it more obvious that some optimization (maybe `singledispatch` will be super fast one day?) could be applied, and remove the indirection.

```suggestion
    elif isinstance(result, CoroutineType):
        return Deferred.fromCoroutine(result)
```

glyph: (Also this raises an interesting question: what about `Future`s 🤔)

glyph (approved):
I've got some stylistic comments about edge cases, but in general, LGTM

src/twisted/internet/defer.py
exarkun: I did have the check inline originally but that left me without a very satisfying place to put the large comment.  If I move the comment back I guess it breaks up the aesthetic of that nice wish-it-were-a-pattern-match structure. :)

Also I guess I originally imagined the function might have another caller in this module but that seems less likely now that I've looked at more of the pieces involved.  Happy to move it back, if you like.

I'm not sure about `isinstance(result, CoroutineType)` instead of `type(result) is CoroutineType` though.  If the idea is to support exactly whatever `async def ...` gives you then I think the latter is more correct.  The latter means that `__instancecheck__` and `__subclasscheck__` are potentially getting involved, right?

Speaking of that, that `Future` question is really a can of worms.  If `CoroutineType`, why not `Future`?  If `Future`, why not `Awaitable`?

I think the answer is that the goal of this change is to provide a transition path from "Deferred-or-not" to "async def" for certain interfaces.  The goal is _not_ "make maybeDeferred compatible with all of the random kinds of async operation APIs" - because maybeDeferred itself is a bit of a misfeature, we're not trying to make it great and powerful, just to provide a path to completely escape from it - and supporting "async def" is enough for that.

Something we should discuss making great and powerful, though, might be `Deferred._runCallbacks` ... should a Deferred callback be allowed to return any Awaitable?  But that is distinctly a separate question.

What do you think?

twm: I'd say stick with `type(result) is CoroutineType` as it's easy to relax that to an `isinstance()` check if necessary, but difficult to optimize in the other direction.

I'd also mildly prefer it were inline, but I think it's good either way.

twm (approved):
This looks great to me. Thank you for all your work on this @exarkun!

src/twisted/internet/defer.py
exarkun: now inline, w/ still using `type(result) ...`

adiroiban (approved):
Looks very good. Nicely documented and tests. Thanks!

PR #1721, opened 2022-04-20
Fix typo in twisted.internet.defer.Deferred.ensureDeferred

jugmac00:


jugmac00:
Test failure is due to a missing news fragment which was also skipped for the last couple of typo fix commits I had a look at.

adiroiban (approved):
Thanks.

I think that we should be "more agile" (tm), and allow typo fixes without an associated trac ticket or news fragment.

But,  I think that the PR title should have a suggestive name. something like.

```
Fix typo in twisted.internet.defer.Deferred.ensureDeferred.
```

It would be very nice if there is a tool that could detect if the path only touches the docstring ... or rst files.

We can then update towncier, not to cry about missing newsfragments.

jugmac00:
@adiroiban I rebased and updated the commit message/title.

PR #1724, opened 2022-04-26
[#10304] Remove py36 from CI and remove 3.6 specific code branches.

adiroiban:
## Scope and purpose

To simplify the CI and development process we will remove support for py3.6

As a drive-by, the CI now runs 3.10 instead of 3.10.0-rc.1

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10304

```

Author: adiroiban
Reviewer: 
Fixes: ticket:10304

Remove the support for Python 3.6.
The cryptography library update is blocked before 37.0, as with 37.0 the current test suite fails.
The cryptography version is pinned to make our existing tests happy.
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -60,7 +60,7 @@ jobs:
           # end to end functional test usage for non-distributed trial runs.
           # When updating the minimum Python version here, also update the
           # `python_requires` from `setup.cfg`.
-          - python-version: '3.7.0'
+          - python-version: '3.7.1'
             tox-env: nodeps-withcov-posix
           # On PYPY coverage is very slow (5min vs 30min) as there is no C
           # extension.
@@ -240,18 +240,5 @@ jobs:
     runs-on: ubuntu-latest
     needs: [testing, static-checks, release-publish]
     steps:
-
-    - name: Fail the job if testing is not success
-      if: ${{ needs.testing.result != 'success'}}
-      run: exit 1
-
-    - name: Fail the job if static-checks is not success
-      if: ${{ needs.static-checks.result != 'success'}}
-      run: exit 1
-
-    - name: Fail the job if release-publish is not success
-      if: ${{ needs.release-publish.result != 'success'}}
-      run: exit 1
-
     - name: note that all tests succeeded
       run: echo "🎉"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,11 +37,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Run on the minimum micro Python version that we can get on CI.
-        # When updating the minimum Python version here, also update the
-        # `python_requires` from `setup.cfg`.
         # Run on latest minor release of each major python version.
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-rc.1]
+        python-version: [3.8, 3.9, 3.10]
         tox-env: ['alldeps-withcov-posix']
         # By default, tests are executed without disabling IPv6.
         noipv6: ['']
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yaml
```diff
@@ -37,40 +37,39 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Run on the minimum micro Python version that we can get on CI.
-        # When updating the minimum Python version here, also update the
-        # `python_requires` from `setup.cfg`.
         # Run on latest minor release of each major python version.
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-rc.1]
+        python-version: [3.8, 3.9, 3.10]
```
adiroiban: As a drive-by, use 3.10 instead of RC.

Trying to reduce the size of the matrix, while still providing good coverage, 3.7 is used is now used for nodeps and noipv6 runs.

I hope that a smaller matrix will reduce the flaky tests.

src/twisted/web/test/test_http.py
```diff
@@ -2569,18 +2568,6 @@ def process(self):
 
 
 class QueryArgumentsTests(unittest.TestCase):
-    # FIXME: https://twistedmatrix.com/trac/ticket/10096
```
adiroiban, line 2572: this test was moved to the ticket https://twistedmatrix.com/trac/ticket/10096#comment:7

as without 3.6 we no longer run it.

adiroiban:
I think that codecov/project — 91.22% (-0.04%) compared to 09fa825  can be ignored .
As it just fails since here we only have code removals.,

setup.cfg
```diff
@@ -80,7 +79,7 @@ tls =
 
 conch =
     pyasn1
-    cryptography >= 2.6
+    cryptography >= 2.6, < 37.0
```
alex, line 82: Why?

adiroiban: Thanks @alex  for the review. Our tests are failing with the latest cryptography 

It might be because the tests are not well written for openssl 3 

But meanwhile we investigate why the tests are failing, I think that we should have a green trunk.

You can see the errors here  trunk was green at merge time, and then after a re-run tests are failing as the latest crytpography is used.

https://github.com/twisted/twisted/runs/6178994160?check_suite_focus=true

alex: Ok. FWIW I'd suggest this be in it's own PR, as it's not related to dropping 36 support.

I'd also encourage folks to get cryptography 37 working soon, as next week there'll be an openssl security release and cryptography 37.0.1 will be necessary for wheels to get a security update.

exarkun: I agree with Alex that, ideally, the cryptography pin would be a separate PR.

If it is not then the ticket and commit message for this PR should be explicit about why that change is being included since otherwise it doesn't make any sense.

exarkun (changes requested):
Thanks.  Apart from the issue about the cryptography pin changing this looks good.

If you don't want to do the extra ticket/PR for that then at least update the commit message in the ticket description.

exarkun (approved):

adiroiban:
Thanks for the review. I have created https://twistedmatrix.com/trac/ticket/10337 to keep a record on the test failures

.github/workflows/test.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
           # end to end functional test usage for non-distributed trial runs.
           # When updating the minimum Python version here, also update the
           # `python_requires` from `setup.cfg`.
-          - python-version: '3.7.0'
+          - python-version: '3.7.1'
```
DMRobertson, line 63: What was the rationale for this, out of interest?

adiroiban: AFAIK, there is no python 3.7.0 available on GitHub Actions.

See this error https://github.com/twisted/twisted/runs/6178400313?check_suite_focus=true#step:3:8

DMRobertson: Thanks, that makes sense. I'm not sure this means that Twisted itself needs to depend on Python >= 3.7.1 though? (See the change to setup.cfg in this commit)

DMRobertson: (For context, we found ourselves unable to create a Poetry lockfile when running against Twisted trunk: see https://github.com/matrix-org/synapse/issues/12575#issuecomment-1112020359)

adiroiban: ok. you can send a PR for 3.7.0 and I can approve it.

previously we had `python_requires = >=3.6.7` ... so it wasn't 3.6.0 ... so we went with the minium GHA could offer.

PR #1726, opened 2022-04-27
10311 trial j u count tests right

exarkun:
## Scope and purpose

See https://twistedmatrix.com/trac/ticket/10311

Also, because of the complex factoring of the disttrial implementation, this includes somewhat substantial refactoring of the way disttrial drives its worker processes and manages the resulting information.

The fix for the specific issue described in the ticket is really just to make a separate TestResult for each iteration instead of re-using the same one over and over.  The refactoring enables this change and makes it testable.
 
## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10311

```

Author: exarkun
Fixes: ticket:10311

Correctly count tests run for each iteration of `trial --jobs=N --until-failure ...`.
```


src/twisted/internet/protocol.py
```diff
@@ -623,6 +623,8 @@ class ProcessProtocol(BaseProtocol):
     stdin, stdout, and stderr file descriptors.
     """
 
+    transport: Optional[interfaces.IProcessTransport] = None
```
exarkun, line 626: `ProcessProtocol` code that uses `self.transport` doesn't type check without this.

src/twisted/scripts/trial.py
```diff
@@ -608,7 +608,7 @@ def _makeRunner(config):
         from twisted.trial._dist.disttrial import DistTrialRunner
 
         cls = DistTrialRunner
-        args["workerNumber"] = config["jobs"]
+        args["maxWorkers"] = config["jobs"]
```
exarkun, line 611: I kept finding myself confused by "workerNumber" which is not the number of a worker but an upper limit on the number of worker processes to create.

src/twisted/trial/_dist/disttrial.py
```diff
@@ -11,99 +11,169 @@
...
+from . import _WORKER_AMP_STDIN, _WORKER_AMP_STDOUT
+from .distreporter import DistReporter
+from .functional import countingCalls, iterateWhile, parallel, void
+from .worker import LocalWorker, LocalWorkerAMP, WorkerAction
+
+
+class IDistTrialReactor(IReactorCore, IReactorProcess):
+    """
+    The reactor interfaces required by disttrial.
+    """
```
exarkun, line 42: I don't know if a better way to represent intersection types for mypy.

src/twisted/trial/_dist/disttrial.py
```diff
@@ -141,113 +337,113 @@ def _driveWorker(self, worker, result, testCases, cooperate):
 
         @param testCases: The global list of tests to iterate.
 
-        @param cooperate: The cooperate function to use, to be customized in
-            tests.
-        @type cooperate: C{function}
```
exarkun, line 146: I could see no value brought by using `cooperator` in this code.  Sending jobs to the worker processes is asynchronous and already gives up execution control.

src/twisted/trial/_dist/functional.py
```diff
@@ -0,0 +1,120 @@
...
+        return fx(fy(a))
+
+    return g
+
+
+# Discard the result of an awaitable and substitute None in its place.
+#
+# Ignore the `Cannot infer type argument 1 of "compose"`
+# https://github.com/python/mypy/issues/6220
+void: Callable[[Awaitable[_A]], Deferred[None]] = compose(  # type: ignore[misc]
```
adiroiban: Minor comment.
This makes more sense to me.

```suggestion
discardResult: Callable[[Awaitable[_A]], Deferred[None]] = compose(  # type: ignore[misc]
```

src/twisted/trial/_dist/functional.py
```diff
@@ -0,0 +1,120 @@
...
+    Deferred.fromCoroutine,
+    partial(flip(sequence), succeed(None)),
+)
+
+
+async def parallel(actions: Iterable[Deferred[_A]]) -> Sequence[_A]:
```
adiroiban: We can keep this.
I don't know we gain by using this instead of `gatherResults`

But unfortunate I am not familiar with py3 async code :(
So maybe for people familiar with py3 async, this makes perfect sense :)

src/twisted/trial/_dist/disttrial.py
```diff
@@ -11,99 +11,169 @@
...
+class WorkerPool:
+    """
+    Manage a fixed-size collection of child processes which can run tests.
+
+    @ivar _config: Configuration for the precise way in which the pool is run.
+    """
+
+    _config: WorkerPoolConfig
+
+    def createLocalWorkers(
```
adiroiban: maybe also make it private... like `_launchWorkerProcesses`.
I don't see why this needed as a public/external method.

This is in a private package, so maybe is ok to have it public.
I am not sure what are the rules for classes from private packages.

src/twisted/trial/_dist/worker.py
```diff
@@ -283,32 +304,47 @@ class LocalWorker(ProcessProtocol):
...
         self._logFile = logFile
-        self.endDeferred = Deferred()
+        self.endDeferred: Deferred = Deferred()
+
+    async def exit(self) -> None:
+        """
+        Cause the worker process to exit.
+        """
+        if self.transport is None:
+            raise NotRunning()
```
adiroiban, line 326: I am not sure why there is no coverage for this.
Is it expected not to have coverage?

I guess that this is just some corner care for which writing a test is not that easy, and is not worth spending the extra effort to write a test for it.


But maybe it is worth to add a nocover pragma together with a comment.

src/twisted/trial/_dist/test/test_disttrial.py
```diff
@@ -369,146 +389,363 @@ def test_runWithoutTestButWithAnError(self):
...
+
+
+class WorkerBroken(Exception):
+    """
+    A worker tried to run a test case but the worker is broken.
+    """
+
+
+class _BrokenLocalWorker:
+    async def run(self, case: TestCase, result: TestResult) -> None:
```
adiroiban: Minor comment.
Maybe be explicit about what is broken with this worker
```suggestion
class _BrokenLocalWorker:
    """
    L{LocalWorker} that always fails to start. 
    """
    async def run(self, case: TestCase, result: TestResult) -> None:
```

src/twisted/trial/_dist/test/test_disttrial.py
```diff
@@ -369,146 +389,363 @@ def test_runWithoutTestButWithAnError(self):
...
+    _config: WorkerPoolConfig
+
+    async def start(
+        self, reactor: interfaces.IReactorProcess
+    ) -> StartedWorkerPoolBroken:
+        return StartedWorkerPoolBroken()
+
+
+class _LocalWorker:
+    async def run(self, case: TestCase, result: TestResult) -> None:
```
adiroiban: I think it would help to have a docstring here.

Why we need this, and why not user `t.trial._dist.worker.LocalWorker.`

adiroiban (approved):
Using `trial -jN -u` is so much better with this PR.
I can now press Ctrl+c and the process ends.
In trunk, the execution stops but the process is stalled.

I get the same error in trunk and on this branch when I press ctrl+c.
Just FYI, I don't think that it should be fixed in this PR.

```
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/adi/chevah/twisted/src/twisted/trial/__main__.py", line 9, in <module>
    sys.exit(load_entry_point("Twisted", "console_scripts", "trial")())
  File "/home/adi/chevah/twisted/src/twisted/scripts/trial.py", line 652, in run
    test_result = trialRunner.runUntilFailure(suite)
  File "/home/adi/chevah/twisted/src/twisted/trial/_dist/disttrial.py", line 454, in runUntilFailure
    return self.run(suite, untilFailure=True)
  File "/home/adi/chevah/twisted/src/twisted/trial/_dist/disttrial.py", line 436, in run
    if isinstance(result, Failure):
UnboundLocalError: local variable 'result' referenced before assignment
```

------------

It's a bit hard to review the mypy and async part of this PR as I don't have much experience with using them.

But I think that this looks good.

If there is no other twisted dev available to review this, I think that is best to merge this PR.

If later we found any issue, we can always fix them in a separate PR.

Also, I think that trial API should not follow the same backward compatibility policy as the other Twisted API...to reduce the development effort.
I assume that nobody is using trial code for production.
Trial code is only used as a development only tool.

Thanks again for the refactoring!

src/twisted/trial/_dist/functional.py
exarkun: Hm.  I don't think we gain anything, in fact.  A previous version of this function did more but then I simplified it for some reason (I forget why now).  I think I can delete it and just use `gatherResults` in its place.

src/twisted/trial/_dist/disttrial.py
exarkun: The symmetry with `_launchWorkerProcesses` makes sense.  I think that disttrial being private means we don't have to worry too much about compatibility guarantees for most of this code to end-users - but `start` is really the only public interface this object is meant to have, so having `_createLocalWorkers` does a better job of communicating internally (within Twisted) about the interface.

src/twisted/trial/_dist/worker.py
exarkun: I added a couple tests and fixed a bug they revealed. :)

src/twisted/trial/_dist/test/test_disttrial.py
exarkun: I added a docstring, maybe that helps a little bit.  I find that the naming of the pieces in disttrial makes all of the interactions a lot harder to understand.  Maybe I didn't improve things much with the new names I added. :/  The whole package might benefit from a change that is purely about giving things clearer names...

PR #1732, opened 2022-05-11
10319 loosen loader requirements

exarkun:
## Scope and purpose

Some trial tests fail in my development environment.  They appear to make assertions which are unreasonably narrow.  This makes all of trial's test suite pass which makes development easier.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10319

```

Author: exarkun
Fixes: ticket:10319

Loosen some assertions about internal trial implementation details.
```


src/twisted/trial/test/test_loader.py
```diff
@@ -66,6 +73,30 @@ def test_findFileWithImportError(self):
         )
 
 
+def loosely_resembles(module: ModuleType) -> Matcher[ModuleType]:
```
glyph: naming convention nit: should be `looselyResembles`. If you want to go for consistency with hamcrest though, feel free to leave it alone

src/twisted/trial/test/test_loader.py
```diff
@@ -66,6 +73,30 @@ def test_findFileWithImportError(self):
...
+    """
+    Match a module with a L{ModuleSpec} like that of the given module.
+
+    @return: A matcher for a module spec that has the same name and origin as
+        the given module spec, though the origin may be structurally inequal
+        as long as it is semantically equal.
+    """
+    expected = module.__spec__
+    if expected is None:
+        match_spec: Matcher[Optional[object]] = is_(None)
```
glyph: Would it be possible to hit this with a synthetic module object?  Or maybe just `assert` it out if we don't need to handle this case yet?

src/twisted/trial/test/test_loader.py
```diff
@@ -89,7 +120,7 @@ def test_moduleInPath(self):
 
         self.assertEqual(sample2, sample1)
 
-    def test_moduleNotInPath(self):
+    def test_moduleNotInPath(self) -> None:
```
glyph, line 122: always love to see some new type coverage in tests

src/twisted/trial/test/test_loader.py
glyph: Should this go into `twisted.trial.test.matchers` too?

glyph (approved):
Seems like this expresses the test's intent a bit better, ship it

src/twisted/trial/test/test_loader.py
exarkun: Oops, that was just force of habit.  I assume Twisted still requires camelCase even for something like "a function used with a library that uses underscores".

I put this one in the test module itself rather than the matchers module because it seems unlikely this will ever be useful for tests elsewhere - so I'd probably lean towards leaving it here.  Of course, someone writing tests elsewhere is not very likely to ever find this for re-use if it is in the test module so that might be an argument for having it in `matchers` instead...

PR #1734, opened 2022-05-25
fix typo in README

edison12a:


README.rst
```diff
@@ -74,9 +74,10 @@ Some of these tests may fail if you:
 Static Code Checkers
 --------------------
 
-You can ensure that code complies to Twisted `coding standards <https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html>`_::
+You can ensure that code complies to Twisted `coding 
+dards <https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html>`_::
```
twm, line 79: It looks like you've introduced a new typo here ("dards").

adiroiban:
I am testing the automatic PR  labeler here 

needs-changes

adiroiban (changes requested):
so it looks like this needs some changes :)

(just testing the review robot here)

README.rst
```diff
@@ -75,9 +75,10 @@ Some of these tests may fail if you:
...
```
adiroiban, line 79: I think that for coding standard we should use "latest" instead of "stable"

Also, we now have a new domain

```suggestion
You can ensure that code complies to Twisted `coding standards <https://docs.twisted.org/en/latest/development/coding-standard.html>`_::
```

PR #1736, opened 2022-06-05
10348 chunkier flattening

exarkun:
## Scope and purpose

See https://twistedmatrix.com/trac/ticket/10348

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10348

```

Author: exarkun
Fixes: 10348

In `twisted.web.template.renderElement`, buffer synchronous pieces
of the result to minimize the number of calls to `request.write`.  This
should be faster and make response sizes smaller.
```


src/twisted/web/test/test_flatten.py
```diff
@@ -453,6 +457,53 @@ def test_unknownTypeRaises(self) -> None:
...
+        When a L{Deferred} is encountered any buffered data is passed to the write
+        function.  After the result of the L{Deferred} is available it is
+        passed to another write along with following synchronous values.
+        """
+        c = succeed("c")
+        f = succeed("f")
+        value = ["a", "b", c, "d", "e", f, "g", "h"]
+        output: List[bytes] = []
+        self.successResultOf(flatten(None, value, output.append))
+        assert_that(output, equal_to([b"ab", b"cde", b"fgh"]))
```
adiroiban: very minor suggestion

I start the review by checking the tests without checking the implementation and I was expecting it to block/wait only for deferred that don't have a result yet.

So this is why I was a bit confused by this test.

the `c` variable name blends in :)


```suggestion
        first_wait = succeed("first-")
        second_wait = succeed("second-")
        value = [
            "already-available",
            "-chunks",
            first_wait,
            "chunks-already-",
            "computed",
            second_wait,
            "more-chunks-",
            "already-available"
            ]
        output: List[bytes] = []
        self.successResultOf(flatten(None, value, output.append))
        assert_that(output, equal_to([
            b"already-available-chunks",
            b"first-chunks-already-computed",
            b"second-more-chunks-already-available"
            ]))
```


Maybe have a new test in which there is a check that the buffering is paused

```python
        first_wait = succeed("first-")
        second_wait = Deferred()
        value = [
            "already-available",
            "-chunks",
            first_wait,
            "chunks-already-",
            "computed",
            second_wait,
            "more-chunks-",
            "already-available"
            ]
        output: List[bytes] = []

        d = flatten(None, value, output.append)

        # Check d is not yet called.
        # and that we have a partial write.
        assert_that(output, equal_to([
            b"already-available-chunks",
            b"first-chunks-already-computed",
            ]))

        # We got a result and the flatten opration is now done.
        second_wait.callback('second-')
        self.successResultOf(d)
        assert_that(output, equal_to([
            b"already-available-chunks",
            b"first-chunks-already-computed",
            b"second-more-chunks-already-available"
            ]))
```

src/twisted/web/_flatten.py
```diff
@@ -381,14 +385,40 @@ async def _flattenTree(
...
-        _flattenElement(request, root, write, [], None, escapeForContent)
+        _flattenElement(request, root, bufferedWrite, [], None, escapeForContent)
     ]
+
     while stack:
         try:
             frame = stack[-1].gi_frame
             element = next(stack[-1])
             if isinstance(element, Deferred):
+                flushBuffer()
```
adiroiban, line 423: I think that this is good enough.

```suggestion
                # For now, keep it simple and just flush on any deferred.
                # As a further optimization, left for future changes,
                # If deferred is already called and deferred.result
                # is not a defered, we can keep adding to the buffer,
                # without flushing it. 
                flushBuffer()
```
Maybe just add a comment for someone looking into further optimization of this part.

src/twisted/web/_flatten.py
```diff
@@ -381,14 +385,40 @@ async def _flattenTree(
 
     @return: A C{Deferred}-returning coroutine that resolves to C{None}.
     """
+    buf = []
+    bufSize = 0
+
+    # Accumulate some bytes up to the buffer size so that we don't annoy the
```
adiroiban, line 391: very minor comment regarding the style.
Why have a comment and not a docstring ?

src/twisted/web/test/test_flatten.py
```diff
@@ -453,6 +457,53 @@ def test_unknownTypeRaises(self) -> None:
...
+        assert_that(
+            output,
+            equal_to([b"x" * BUFFER_SIZE, b"y" * BUFFER_SIZE, b"z" * BUFFER_SIZE]),
+        )
+
+    def test_chunksSeparatedByAsync(self) -> None:
+        """
+        When a L{Deferred} is encountered any buffered data is passed to the write
+        function.  After the result of the L{Deferred} is available it is
+        passed to another write along with following synchronous values.
```
adiroiban:
```suggestion
        When a L{Deferred} is encountered any buffered data is passed to the write
        function.  After the result of the L{Deferred} is available it is
        passed to another write along with the following synchronous values.
        
        The write is triggered by any deferred,
        not only by deferreds that don't yet have a result.
```

adiroiban (approved):
Thanks for the performance improvement.
This looks very good and is already a big improvement.

The comments are very minor and should not block the merge of this code.

src/twisted/web/_flatten.py
exarkun: Usually I don't put docstrings on inner functions since it is extremely unlikely that anyone will ever see them as docstrings and if they are docstrings then they are held in memory.  I don't feel too strongly either way though.

src/twisted/web/test/test_flatten.py
exarkun: Hm, yes, but this test only tests the Deferred-that-already-has-a-result case.

src/twisted/web/_flatten.py
exarkun: Eh, I don't want to add anything about the `result` attribute.  The only people I want looking at that don't need any hints from me.

PR #11570, opened 2022-07-08
Don't skip all-success on deps failure.

### LINKED ISSUE #11569: Enforce protected branches
GitHub has an issue in which when a dependent job fails, the job depending on it will skip.

We want the downstream job to also fail.

In this way, in GitHub branch protection API we can define a single job for a workflow file.

In this way,  each time you modify a job in a worflow file, you don't have to also update the GitHub UI config.

adiroiban:
## Scope and purpose

Fixes #11569 

This applies the changes created by @altendky  from here https://github.com/Chia-Network/chia-blockchain/pull/11880/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R79 

The code was already tested in towncrier here https://github.com/twisted/towncrier/pull/392

Note that this only handles the jobs defined inside our `test.yaml` files

codecov.io, Azure Pipelines, Read The Docs, and pre-commit are external services and are not covered by this change.

In theory, we can enabled probot-settings app and have all the repo settings as YAML 
https://github.com/repository-settings/app

Anyone with push access to the repo will be admin and could change the branch protection via YAML config.

```
Author: adiroiban
Reviewer: 
Fixes #11569

Fail and don't skip CI branch protection when a job fails.
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -257,12 +257,29 @@ jobs:
           fi
 
 
+  # We have this job so that the PR can be blocked on a single job.
+  # In this way, each time a job is modified,
+  # we don't have to go to GitHub UI and reconfigure branch protection.
+  # See GitHub support answer for this hack.
+  # https://gist.github.com/altendky/2e3483a1f7e1ba21cc97de75db9b7d1c
   all-successful:
-    # We have this job so that the PR can be blocked on a single job.
-    # See GitHub support answer for this hack.
-    # https://gist.github.com/altendky/2e3483a1f7e1ba21cc97de75db9b7d1c
+    # Is very important to force running this always, as otherwise it will be
+    #  skipped by default.
+    if: always()
     runs-on: ubuntu-latest
-    needs: [testing, static-checks, release-publish]
+    # Here should be the list of all the other jobs defined in this file.
+    needs:
+      - testing
+      - static-checks
+      - release-publish
     steps:
-    - name: note that all tests succeeded
-      run: echo "🎉"
+      - name: Require all successes
+        shell: python
+        env:
+          RESULTS: ${{ toJSON(needs.*.result) }}
+        run: |
+          import json
+          import os
+          import sys
+          results = json.loads(os.environ["RESULTS"])
+          sys.exit(0 if all(result == "success" for result in results) else 1)


.github/workflows/test.yaml
```diff
@@ -257,12 +257,29 @@ jobs:
...
+      - name: Require all successes
+        shell: python
+        env:
+          RESULTS: ${{ toJSON(needs.*.result) }}
+        run: |
+          import json
+          import os
+          import sys
+          results = json.loads(os.environ["RESULTS"])
+          sys.exit(0 if all(result == "success" for result in results) else 1)
```
altendky, line 285: As I see this show up in more places, maybe it should be made into an action?  Or the whole job into a reusable workflow.  In this case the reusable workflows are used to get around the 256 job per matrix limit, but it is still an example with the `needs:` in use.

https://github.com/Chia-Network/chia-blockchain/blob/9a4316f96eeb92a06d5ceaf94387d349c5ea9537/.github/workflows/test.yml#L45-L55
```yaml
  macos:
    uses: ./.github/workflows/test-single.yml
    needs: configure
    with:
      emoji: 🍎
      matrix: macos
      name: macOS
      file_name: macos
      concurrency_name: macos
      configuration: ${{ needs.configure.outputs.configuration }}
      runs-on: macos-latest
```

altendky (approved):
It reads right at least.  The reusability comment can be taken on separately if we get to that.

.github/workflows/test.yaml
adiroiban: I have merged this as it should be better than what we have now.

I think that a shared action should be good enough.

--------

I am not sure how workflow interdependencies work.

adiroiban:
Many thanks for the review! :)

.github/workflows/test.yaml
altendky: I did make a first try at shared workflow.  You do have to pass the `needs:` and code the `if: always()` in the repeated location so...  It's something, but not what I had hoped.  https://github.com/Chia-Network/chia-blockchain/pull/11880

adiroiban: Thanks... I think that is nice, but for Twisted is a bit complicated.

I mean the current code, should be easier to read and understand by average Twisted dev that is not an expert into GiHub Actions.

webknjaz: > As I see this show up in more places, maybe it should be made into an action?

Yes: https://github.com/re-actors/alls-green

PR #11572, opened 2022-07-11
11571 enforce char encoding in testing

### LINKED ISSUE #11571: Enforce the character encoding in "test_newlineStyle"
The test `test_newlineStyle` of `twisted.trial` is failing on my Windows machine due to the following error.
> builtins.UnicodeEncodeError: 'charmap' codec can't encode character '\u2609' in position 15: character maps to <undefined>

It seems because the default `encoding` of the built-in function [`open()`](https://docs.python.org/3/library/functions.html#open) is platform dependent, and the default character encoding `cp1252` for `Windows` doesn't have a mapping to the Unicode character `sun` or `\u2609` which is used for testing.

The same error can be raised with `"Here comes the \N{sun}!".encode('cp1252')`.

jamwonro:
## Scope and purpose

Fixes #11571 

The test `trial._dist.test.test_worker.LocalWorkerTests.test_newlineStyle` fails on `Windows` machines, because the default character encoding `cp1252` doesn't have a mapping for the Unicode character `sun`. The failing test case can be resolved by enforcing a character encoding when it opens a file for writing.

## Contributor Checklist:

* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://docs.twistedmatrix.com/en/stable/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://github.com/twisted/trac-wiki-archive/blob/trunk//ReviewProcess.mediawiki#News_files))
* [x] The title of the PR starts with the associated issue number (without the `#` character).

```
Merge pull request #123 from twisted/4356-branch-name-with-issue-id

Author: jamwonro
Fixes #11571 

Fix the failure of "trial._dist.test.test_worker.LocalWorkerTests.test_newlineStyle" on Windows by forcing UTF-8 encoding.
```


adiroiban:
Thanks for the PR.

We already have the test green on Windows. See this build log for trunk. 

https://dev.azure.com/twistedmatrix/twisted/_build/results?buildId=4848&view=logs&j=a9f5be88-9f44-53f4-896b-687d3085a015&t=50e42789-57d9-5d74-e8a4-3a4e15f86668&l=10252

I see that we "force" UTF-8 encoding when running the tests

https://github.com/twisted/twisted/blob/trunk/azure-pipelines/windows_test_jobs.yml#L30

-----------

I am not sure what is the best way to run the tests... use default `cp1252` or force UTF-8

--------

Was this the only test failing with `cp1252` ?

Maybe as part of this PR we should run Windows 3.7 tests with UTF-8 and 3.10 with default encoding... just to improve the coverage and make sure all the other tests work with default Windows encoding.

--------

But I guess that as long as there is a comment for this change explaining why we don't use the default encoding, that should be fine.

-------

Maybe we should leave the test as it it, and instead update the documentation to inform Windows users that they should always run Twisted with UTF-8 default encoding...

I don't know what is best.

jamwonro:
@adiroiban It was the only test case that failed on my Windows 11 machine with the default character encoding `cp1252`. I think it is natural to specify the `encoding` simply because the test uses a `Unicode` character.

adiroiban:
Thanks for the feedback.
Is this ready for review?

What should be the merge message for this change?

The PR description only has  `Long description providing a summary of these changes.
(as long as you wish)`  :p

jamwonro:
@adiroiban Yes, this is ready for review. I updated the description. Many thanks for your comments.

adiroiban:
Thanks. I have enabled the auto-merge for this PR. I hope all will be ok with the CI and it will land soon.

Thanks again for helping with Twisted on Windows.
This is one important part that is a bit neglected, as the majority of Twisted devs are on Unix

adiroiban (approved):
All good. Thanks.

There is no need to advertise this change to the public release notes, as this is a private change.

End users of the library shouldn't care.

PR #11579, opened 2022-07-14
fix: adbapi.Transaction.reconnect

powo:
The `.reconnect()` method of `twisted.adbapi.Transaction` calls `.reconnect()` on the underlying (`self._connection`) DB-API Connection object, which results e.g. in:

```
exceptions.AttributeError: 'psycopg2.extensions.connection' object has no attribute 'reconnect'
```

This Bug seems to originate from a commit back in 2005: https://github.com/twisted/twisted/commit/77d242150ff3e2dffc1e2c3b9227f423cc311926#diff-35f94ca8beef39d853fda43e943779380c128ff0c29bb6fc939c163d527d0f9aR113

.. I assume this happened by accident, and the author wanted to call the Method of the `adbapi.Connection` Class instead.

The Fix exactly reverts this few lines of that prior commit and calls instead the same disconnect/connect like in the `adbapi.Connection` Class.


adiroiban:
Hi. Thanks for the change.

The tests needs to be green before a merge.

Also, having some tests to make sure we don't introduce regressions for this functionality would be great.

I am not using adbapi in Twisted and I am not sure what is the intended purpose of this reconnect method. Maybe this is a good time to add a docstring to this method.

PR #1234, opened 2020-03-25
#9782 Remove twisted.news.

hawkowl:
Trac link: https://twistedmatrix.com/trac/ticket/9782

PR #11578, opened 2022-07-12
#10066 Update pydoctor version and remove Twisted pydoctor custom code that is now available as standard features.

adiroiban:
## Scope and purpose

Fixes #10066

If the epydoc API markup is invalid, the CI is happy and green.

We need to update the CI to fail on pydoctor build errors.
This should help make sure the API doc is valid and sane :)

As part of this PR I have updated the pydoctor version.

Also, instead of using the custom python script to call pydoctor, this calls pydoctor directly from tox.
I hope that in this way it should be easier to understand what is going on there, with fewer redirection layers.

## Contributor Checklist:

* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://docs.twistedmatrix.com/en/stable/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://github.com/twisted/trac-wiki-archive/blob/trunk//ReviewProcess.mediawiki#News_files))
* [x] The title of the PR starts with the associated issue number (without the `#` character).
* [x] I have updated the automated tests and checked that all checks for the PR are green.
* [x] The merge commit will use the below format
  The first line is automatically generated by GitHub based on PR ID and branch name.
  The other lines generated by GitHub should be replaced.

```
Merge pull request #123 from twisted/4356-branch-name-with-issue-id

Author: adiroiban
Reviewer: 
Fixes #10066

Update pydoctor version.
Fail CI for a PR on pydoctor build errors.
```

PR #11577, opened 2022-07-11
#11575 Move development related docs into a single folder inside docs root.

adiroiban:
## Scope and purpose

Fixes #11575

This moves all the dev docs into a separate folder.
It keeps redirection pages.

Right now we have docs:

* Here - https://docs.twisted.org/en/stable/core/development/index.html
* and here - https://docs.twisted.org/en/stable/core/development/policy/index.html

I think that `/core/development/policy/` too deep, and has sub-documented, and we have sub-documetns also at `/core/development/`
This PR moves all the dev docs into a single place at `/development` 

And keeps redirection pages for the old URL.

## How to test

This PR is only about moving existing content.
No changes were done to the content itslef.
It's just an re-organization

* New pages at https://twisted--11577.org.readthedocs.build/en/11577/development/index.html
* Some redirection pages https://twisted--11577.org.readthedocs.build/en/11577/core/development/index.html
* And policy redirection https://twisted--11577.org.readthedocs.build/en/11577/core/development/policy/index.html
* A leaf redirection https://twisted--11577.org.readthedocs.build/en/11577/core/development/philosophy.html

## Contributor Checklist:

* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://docs.twistedmatrix.com/en/stable/core/development/policy/coding-standard.html)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://github.com/twisted/trac-wiki-archive/blob/trunk//ReviewProcess.mediawiki#News_files))
* [x] The title of the PR starts with the associated issue number (without the `#` character).
* [NA] I have updated the automated tests and checked that all checks for the PR are green.
* [X] The merge commit will use the below format
  The first line is automatically generated by GitHub based on PR ID and branch name.
  The other lines generated by GitHub should be replaced.

```
Merge pull request #123 from twisted/4356-branch-name-with-issue-id

Author: adiroiban
Reviewer: 
Fixes #11575

Move all dev docs into the root of the docs project.
```

PR #11586, opened 2022-07-19
#11585 Update the PR template and dev docs to not require a specific merge commit message.

### LINKED ISSUE #11585: Remove the requirement for a fixed merge message
> The commit format is no longer that important.
> 
> If the PR description contains `Fixes #1234` I think that the associated PR is automatically closed. This is why I am observing from here.
> 
> What we can do, is update the format of the PR title, to include the ticket id as `#1234` and not just `1234`. In this way, the commit message is auto-filled with the PR title and will have the reference to the ticket.
> 
> we can also configure GitHub to automatically delete the branch after a merge.

Yeah, I agree with all this. I think we should auto-delete branches and switch to the Github-friendly "fixes" line, and possibly dispense with the manual commit format entirely.

_Originally posted by @glyph in https://github.com/twisted/twisted/issues/11584#issuecomment-1188481007_

adiroiban:
## Scope and purpose

Fixes #11585

With the GitHub integration, a specific commit message is no longer required.

GitHub can use the PR title as the commit message.
And if the PR description contains a reference to a ticket, it is auto closed.

## Changes

This updates the PR template not to request the commit message.
The PR template is updated to talk about the PR title.
The PR template was updated with the latest links.

I have removed the contribution checklist as I think they are just noise.
We should use the check list for real tasks associated with a PR.
The CI checks should automatically force you to "check" those steps.

I have updated the CI to fail if the PR title has no ticket reference.
It is a very simple check.

As a drive-by I have move the docs into a separate job as docs + mypy was taking to long.
Also narrative docs now builds apidocs. So there is no need for both.
More work for docs CI is at https://github.com/twisted/twisted/pull/11578

## Trade-off

The PR title check runs on commit push. In this way we have it in a single YAML file.
The problem is that if the PR is wrong, you will need a new commit to have the PR title check pass.

The other option is to have the PR title check in a separate YAML file that runs on every PR title update.

## How to test

See a failure here https://github.com/twisted/twisted/runs/7401209770?check_suite_focus=true#step:8:24

----------

Testing the PR template is complicated...as it will be visible on the next run
Below is the same content to review the MD rendering

-----------

## Scope and purpose

Fixes #

You can just refer to the associated ticket(s).

## Contributor Checklist:

This process apply to *all* pull requests, no matter how small.

* The title of the PR should describe the changes and starts with the associated issue number (including the `#` character for GitHub auto-links).
* Once all checks are green, request a review by leaving a comment that contains the `please review` words.
  This will trigger the review process.


adiroiban:
I think that this is ready for an initial review

please review

thanks!

.github/workflows/test.yaml
```diff
@@ -191,9 +214,35 @@ jobs:
...
 
+    # This is called only on code changes.
+    # If you forgot to define the right PR title you should re-run the job.
+    - name: Check PR title format
+      if: github.event_name == 'pull_request'
+      shell: python
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      run: |
+        import os, re, sys
```
exarkun: IMO code in a yaml file is an anti-pattern.  At most this should reference an external tool that can be a regular Python file (and so gain editor support for proper 4 space indents :wink: ).

.github/workflows/test.yaml
```diff
@@ -191,9 +214,35 @@ jobs:
         git reset --hard $BRANCH_SHA
 
     - name: Newsfragment
+      if: ${{ !cancelled() }}
```
exarkun: I don't know why this is here.  Maybe a comment would help.

.github/workflows/test.yaml
```diff
@@ -179,7 +202,7 @@ jobs:
     - name: Run the checks
       run: |
         python --version
-        tox --parallel -e lint,apidocs,narrativedocs,mypy
```
exarkun, line 182: I see that there is a separate job for `narrativedocs` above now.  I like this.  CI jobs that generate 10,000 lines of output that are a mix of 5 different logical jobs are miserable to interact with.  I hope that `lint` and `mypy` will be similarly split apart someday.

I don't see a separate job for `apidocs` being created.  Maybe this is because the API docs are now generated exclusively by readthedocs?

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -20,22 +12,13 @@ Add mentions of things that are not covered here and are planed to be done in se
...
-
-Author: <github_username>, <github_usernames_if_more_authors>
-Reviewer: <github_username>, <github_username_if_more_reviewers>
-Fixes #<github_issue_number>
-Fixes #<another_if_more_in_one_PR>
-
-Long description providing a summary of these changes.
-(as long as you wish)
-```
+This process apply to *all* pull requests, no matter how small.
```
exarkun:
```suggestion
This process applies to *all* pull requests - no matter how small.
```

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,17 +1,9 @@
...
-Note that the GitHub issue, news fragment, and review submission portions of this process apply to *all* pull requests, no matter how small;
-if you don't do them, it's likely that nobody will even notice your PR needs a review.
--->
-
 ## Scope and purpose
 
 Fixes #<!-- ISSUE-ID -->
 
 Add a few words about why this PR is needed and what is its scope.
+You can just refer to the associated ticket(s).
```
exarkun:
```suggestion
If the associate ticket(s) fully explain the need you can just refer to it/them.
```

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -20,22 +12,13 @@ Add mentions of things that are not covered here and are planed to be done in se
 
 ## Contributor Checklist:
 
-* [ ] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://docs.twistedmatrix.com/en/stable/core/development/policy/coding-standard.html)
-* [ ] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://github.com/twisted/trac-wiki-archive/blob/trunk//ReviewProcess.mediawiki#News_files))
-* [ ] The title of the PR starts with the associated issue number (without the `#` character).
-* [ ] I have updated the automated tests and checked that all checks for the PR are green.
```
exarkun, line 26: This part of the template change doesn't seem to be in support of resolving the linked ticket but I don't particularly object to the changes (one version seems as good as the other to me).

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -20,22 +12,13 @@ Add mentions of things that are not covered here and are planed to be done in se
...
+This process apply to *all* pull requests, no matter how small.
+Have a look at [our developer documentation](https://docs.twisted.org/en/latest/core/development/dev-process.html) before submitting your Pull Request.
+
+Below is a non-exaustive list (as a reminder):
+
+* The title of the PR should describe the changes and starts with the associated issue number (including the `#` character for GitHub auto-links).
+* A release notes news fragment file was create in src/twisted/newsfragments/ (see: [Release notes fragments docs.](https://docs.twisted.org/en/latest/core/development/dev-process.html#release-notes-management))
+* The automated tests were updated.
+* Once all checks are green, request a review by leaving a comment that contains the `please review` words.
+  This will trigger the review process.
```
exarkun: I'd say what this process looks like so people can evaluate whether it is happening or not (and if not complain to someone).

I'm guessing it is something like "a bot requests a review from some twisted-committers team"?

docs/core/development/review-process.rst
```diff
@@ -104,17 +104,6 @@ Authors: How to merge the change to trunk
...
-This helps automatically close the associated ticket::
-
-    Merge pull request #123 from twisted/FIRST-LINE-AUTOMATICALLY-GENERATED-BY-GITHUB
-
-    Author: comma_separated_github_username
-    Reviewer: comma_separated_github_usernames
-    Fixes #issue number
-
-    Long description (as long as you wish)
-
```
exarkun, line 117: I'd probably leave in a note about how the ticket description or the commit message must say "Fixes #..." to trigger the automation.

It's good that the default ticket description includes this but sometimes people just delete the whole template and put in their own stuff.

exarkun (approved):
Thanks.  I left some comments inline but nothing big.  Please address and then merge.

.github/PULL_REQUEST_TEMPLATE
adiroiban: > I have removed the contribution checklist as I think they are just noise.
> We should use the check list for real tasks associated with a PR.

This was a suggestion 

With the previous checklist, we see this in the reporting, but at least for me is noise.

image

--------

if a ticket has no newsfragment, the CI will fail.

This is why I think that we don't need the check list... is extra work to tick :)

.github/PULL_REQUEST_TEMPLATE
adiroiban: True. Thanks.

.github/workflows/test.yaml
adiroiban: Yes. apidocs is built as part of read the docs.

At the same time, if `apidocs` build fails, the CI is still green.

I am working to fix apidocs as part of https://github.com/twisted/twisted/pull/11578

-----------

the reasoning for my change was that since GHA VMs have 2CPU.. there is not much use to run more than 2 jobs.

----------

And yes. At the same time, reading the output is hard.


But I will refactor this to run in a matrix to reduce the need for copy/paste inside the YAML file.

.github/workflows/test.yaml
adiroiban: OK. I will add a comment.

This is here so that newsfragment check will run, even if the previous fails.

.github/workflows/test.yaml
adiroiban: The thing is that this code is not very useful outside of GHA.

But you are right. I will extract it to make it easier to test it

docs/core/development/review-process.rst
adiroiban: My goal was to reduce duplication... and the probability of getting documentation that is out of sync.

Also, the plan is to have an automated CI check for all this.

If your PR title has no ticket ID , the CI will fail
If the PR description has no `Fixes #1234` the CI will fail.

---------

Here I would keep only the general / high-level review  info.

All the low details can be written in the template and in the CI check.
And the template is just a file that is available.

---------

What we can do, is look into integrating the PR template inside the documentation.
So that we don't have to duplicate the content...and any change to the PR template will also automatically update the documentation.

Maybe this can be done as part of https://github.com/twisted/twisted/pull/11577

adiroiban (commented):
Many thanks for the review.

I will update the branch and merge it.

Not sure why the review was sent earlier.

.github/workflows/test.yaml
adiroiban: In the end I decided to just revert this change.

The plan is to upgrade to the latest version of towncrier , and then we will refactor this to run in a separate job.

PR #11589, opened 2022-07-26
#10276 Correct type hints for IHostnameResolver.resolveHostName

### LINKED ISSUE #10276: Type hints on IHostnameResolver.resolveHostName seem incorrect
|[clokep's avatar](https://github.com/clokep)| @clokep reported|
|-|-|
|Trac ID|trac#10276|
|Type|defect|
|Created|2021-11-16 13:02:18Z|

I believe the type hints on `IHostnameResolver.resolveHostName` are incorrect in two places:

1. `addressTypes` needs to be an iterable of `IAddress` types, not instances. (I don't think it needs to be a sequence, but not 100% sure on that.)
2. The return type might be incorrect. It is declared as `IResolutionReceiver` (one of the input arguments), but at least the `GAIResolver` implementation returns an instance of `IHostResolution`. So something funky is going on.

Found while adding type hints to Synapse: https://github.com/matrix-org/synapse/pull/11328

```
trac-id__10276 10276
type__defect defect
reporter__clokep clokep
priority__low low
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1637067738920446 1637067738920446
changetime__1637067738920446 1637067738920446
version__None None
owner__None None

```
</details>

clokep:
## Scope and purpose

Fixes #10276

Adds missing type hints to `twisted.internet._resolver`, as part of this it updates the type hints for `IHostnameResolver.resolveHostName` to be correct:

* `addressTypes` is a sequence of `IAddress` types, not `IAddress` instances.
* The return value should be `IHostResolution`, not `IResolutionReceiver`.

Ports matrix-org/synapse#11328.

## Contributor Checklist:

* The title of the PR should describe the changes and starts with the associated issue number (including the `#` character for GitHub auto-links).
* Once all checks are green, request a review by leaving a comment that contains the `please review` words.


src/twisted/internet/_resolver.py
```diff
@@ -26,14 +39,25 @@
...
 )
 from twisted.internet.threads import deferToThreadPool
 from twisted.logger import Logger
 from twisted.python.compat import nativeString
+from twisted.python.failure import Failure
+
+if TYPE_CHECKING:
+    from twisted.python.runtime import platform
+
+    if platform.supportsThreads():
```
exarkun: What does it mean to have an executable condition like this inside an `if TYPE_CHECKING` block, where the code will definitely never run?

clokep: I think it works fine, I believe anything inside of `TYPE_CHECKING` is just ignored when CPython executes the code (hence why `ThreadPool` is quoted below). mypy seems to be OK with it, but I'm not sure if that actually executes the module (or how it picks up on what this is doing).

Note that this conditional is copied from `twisted.internet.interfaces`:

https://github.com/twisted/twisted/blob/release-21.2.0-10091/src/twisted/internet/interfaces.py#L56-L59

src/twisted/internet/_resolver.py
```diff
@@ -199,25 +239,25 @@ def resolveHostName(
...
-            hostName = _idnaBytes(hostName)
+            hostName_bytes = _idnaBytes(hostName)
 
         # Make sure it's passed down as a native str, to maintain the interface
-        hostName = nativeString(hostName)
+        hostName = nativeString(hostName_bytes)
 
         resolution = HostResolution(hostName)
         resolutionReceiver.resolutionBegan(resolution)
         onAddress = self._simpleResolver.getHostByName(hostName)
```
clokep, line 250: This line now doesn't match the interface for `IResolverSimple` which states that the `timeout` parameter is required. I'm not sure if I should:

1. Pass the default timeout value here.
2. Update the interface to reflect that a default value exists.

Any thoughts on what to do here?

(There's a similar error related to `IHostnameResolver.resolveHostName` on line 348.)

clokep: We chatted a bit about this in #twisted:

```
12:04:11 PM - clokep: Not sure how to get feedback on an ongoing PR, but I left a question at https://github.com/twisted/twisted/pull/11589#discussion_r961057512 -- the tl;dr is Twisted internally seems to assume that implementations of an interface have a default value which the interface doesn't show. What's the proper solution to this?
12:04:37 PM - clokep: (To add the default to the interface or the provide the defaults in the caller or something else?)
12:07:49 PM - jean-paul[m]: The goal should usually be to fix the mismatch without breaking any code.  Sometimes that's hard, maybe even hard enough that it's not worth doing, but that's the goal that should guide choices.
12:08:35 PM - jean-paul[m]: Since it's basically a point of extension, we should expect developers are implementing the interface as currently defined.
12:09:48 PM - jean-paul[m]: Buuuut if they did that, their implementation wouldn't actually work with Twisted, right?  Because Twisted calls the method without a value for that parameter.
12:10:28 PM - jean-paul[m]: So from that perspective, the interface is wrong and should be updated to have a default - reflecting real usage by Twisted.
12:11:26 PM - jean-paul[m]: You could also look at it from the opposite perspective - someone wrote code that expects implementations of the interface and then uses it according to what the interface says.  Fortunately, this means those implementations would already be passing a value for `timeout` since the interface says you have to.
12:11:40 PM - jean-paul[m]: So this code also wouldn't be broken by adding a default to the interface.
12:12:08 PM - jean-paul[m]: I don't think that's a complete case analysis but it's probably the major points.  And I don't think this interface is an extremely popular extension point anyway.
12:12:21 PM - clokep: Right, my inclination was that adding a default to the interface is the "least" breaking thing to do.
12:12:40 PM - jean-paul[m]: That seems right to me.
12:13:01 PM - clokep: Awesome. Thanks for the help! Will try to finish that up "soon"!
12:13:23 PM - clokep: (I think connectTCP has a similar issue with parameters, will need to check the Synapse codebase again for ignores.)
12:15:00 PM - graingert[m]: I think the policy on type annotations is that they may break at any time right now
12:15:16 PM - graingert[m]: I can't find the docs on that
12:15:17 PM - graingert[m]: But they're currently provisional
12:15:42 PM - clokep: Yeah, I'm not really talking about the type annotations breaking -- I'm talking about type annotations finding bugs in old code. 😁
12:17:13 PM - graingert[m]: Ah right
```

The tl;dr is that option 2 from above seems more reasonable in this case (to avoid breaking code).

src/twisted/internet/endpoints.py
```diff
@@ -747,7 +751,7 @@ def ebLog(error):
         d.addCallback(cbDeliver)
         d.addErrback(ebLog)
         d.addBoth(lambda ignored: resolutionReceiver.resolutionComplete())
-        return resolutionReceiver
+        return resolution
```
clokep, line 755: This changes the return value here, but it is never used:

https://github.com/twisted/twisted/blob/6a8a70b348480cc59e9944f47330ea0f4a6934f7/src/twisted/internet/endpoints.py#L959-L961

src/twisted/internet/base.py
```diff
@@ -1030,7 +1030,9 @@ def runUntilCurrent(self) -> None:
         threadpoolShutdownID = None
 
         def _initThreads(self) -> None:
-            self.installNameResolver(_GAIResolver(self, self.getThreadPool))
+            self.installNameResolver(
+                _GAIResolver(cast(IReactorThreads, self), self.getThreadPool)
+            )
```
clokep, line 1129: I think this is safe -- if you have threads than your reactor likely implements `IReactorThreads`? The underlying code assumes it anyway.

Alternately I can add an assert:

```python
assert isinstance(self, IReactorThreads)
```

clokep:
please review

src/twisted/internet/endpoints.py
adiroiban: Hm... might not be used by Twisted own code, but this looks like public API, so it might be used by someone else.

But I agree that the change makes more sense... since the old code just returns an input argument.

But this needs to follow the deprecations process... or ask for an exception 

-------------

Do we still need `@rtype: L{IResolutionReceiver}` in the docstring ?

same for `@type resolutionReceiver: L{IResolutionReceiver}` ?

src/twisted/internet/interfaces.py
```diff
@@ -189,10 +190,10 @@ class IHostnameResolver(Interface):
...
         resolutionReceiver: IResolutionReceiver,
         hostName: str,
-        portNumber: int,
-        addressTypes: Sequence[IAddress],
-        transportSemantics: str,
-    ) -> IResolutionReceiver:
+        portNumber: int = 0,
+        addressTypes: Optional[Sequence[Type[IAddress]]] = None,
+        transportSemantics: str = "TCP",
+    ) -> IHostResolution:
```
adiroiban, line 196: changing interfaces is complicated ... I think the only way to change an interface, is to add a new one.

---------

so updating the interface to reflect a default value is ok.

----

updating the interface to return a different type, is not ok.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -1770,9 +1770,7 @@ def captureLogs(event):
         globalLogPublisher.addObserver(captureLogs)
         self.addCleanup(lambda: globalLogPublisher.removeObserver(captureLogs))
 
-        receiver = self.resolver.resolveHostName(self.receiver, "example.com")
-
-        self.assertIs(receiver, self.receiver)
+        self.resolver.resolveHostName(self.receiver, "example.com")
```
adiroiban, line 1773: and we have a test to check for the exact return value... so this is part of the public api.

I don't think that is ok to change it.

adiroiban:
> The return value should be IHostResolution, not IResolutionReceiver.

Why?  Just asking.

src/twisted/internet/interfaces.py
clokep: I'm not changing the return value of the interface -- the type hints were incorrect when they were added.

src/twisted/internet/endpoints.py
clokep: > Hm... might not be used by Twisted own code, but this looks like public API, so it might be used by someone else.
> 
> But I agree that the change makes more sense... since the old code just returns an input argument.

I don't think I was clear above -- this implementation (i.e. `_nameResolver` which gets set from a private class also in this file) returns the wrong type, which is never used. I don't see anyway for this to leak out to the public API. This is an internal-only change.

src/twisted/internet/test/test_endpoints.py
clokep: `self.resolver` is instantiated as an private class (`_SimpleHostnameResolver`). It is not part of the public API and not subject to the deprecation policy.

clokep:
> > The return value should be IHostResolution, not IResolutionReceiver.
> 
> Why? Just asking.

Because the type hint was incorrect when added.

The implementation of `GAIResolver`, which was the initial implementation of `IHostNameResolver` has always returned (since 2016) a `IHostResolution` instance, see #548. Note that but the documentation of the return type was incorrect in that pull request (but the actual description was correct):

https://github.com/twisted/twisted/pull/548/files#diff-79b2f1f9960a400f335be8682ed57a4cf5379e47c067c9a320cad9e7c91890e6R186

https://github.com/twisted/twisted/pull/548/files#diff-575bab6dffb171ffba2feff510a95f0aa1105a17f3d7100961bcaeda7b7c5cd1R185-R186

The incorrect documentation than got ported into an incorrect type hint in #1430. (I can't seem to deep link easily into that PR unfortunately.)

When the endpoint resolver was added it seems like it was added by following the (incorrect) documentation:

https://github.com/twisted/twisted/pull/740/files#diff-5f22d09685cbd043da78599cb70caa39bf60b73abf2f94fe1ec72d020ad5ccc5R718

Luckily the `_SimpleHostnameResolver` from the endpoints code is not public *and* the result of calling `resolveHostName` is never used:

https://github.com/twisted/twisted/blob/19a9a4cba206432f0a229b686a07548687ad4c85/src/twisted/internet/endpoints.py#L955-L957

----

In summary, what we have is:

1. Multiple implementations of `IHostNameResolver` which differ in their implementation (`GAIResolver ` and `SimpleResolverComplexifier` return `IHostResolution`; `_SimpleHostnameResolver` returns `IResolutionReceiver`).
2. `IHostNameResolver.resolveHostName` is documented as returning `IResolutionReceiver`.
3. `IHostNameResolver.resolveHostName` has a type hint as returning `IResolverReceiver` (but it has been mentioned that type hints are considered provisional at the moment).

I'm asserting that the best way to *not* break things is to:

1. Consistently implement `IHostNameResolver` between the three implementations, choosing the implementations which are public (yes, this changes `_SimpleHostnameResolver`, but that shouldn't leak outside the APIs at all).
2. Update the type hint on `IHostNameResolver.resolveHostName` since those can change at any point and are not subject to the deprecation policy (that's my understanding -- I might be wrong).

Honestly: this is the only way I can see us *not* breaking code. But it *might* (honestly it likely will) break type hints for anyone that is implementing `IHostNameResolver`. I don't see many folks as having a separate implementation of that though!

As a final thought, I'm fairly certain the initial docstring was wrong (and the implementations are correct) since `IResolutionReceiver` is passed into `resolverHostName`, so returning it is useless. Without returning `IHostResolution`, however, than the caller has no access to it. 🤷

src/twisted/internet/_resolver.py
```diff
@@ -26,14 +39,25 @@
...
 from twisted.python.compat import nativeString
+from twisted.python.failure import Failure
+
+if TYPE_CHECKING:
+    from twisted.python.runtime import platform
+
+    if platform.supportsThreads():
+        from twisted.python.threadpool import ThreadPool
+    else:
+        ThreadPool = object  # type: ignore[misc, assignment]
```
glyph: While this might *work*, it's kind of incoherent and very confusing to read. I think what you want is simply this:

```suggestion
    from twisted.python.threadpool import ThreadPool
```

The code in that other file is a mistake, so let's not repeat it here.

src/twisted/internet/_resolver.py
```diff
@@ -42,13 +66,13 @@ class HostResolution:
...
 
-    def __init__(self, name):
+    def __init__(self, name: str):
         """
         Create a L{HostResolution} with the given name.
         """
         self.name = name
 
-    def cancel(self):
+    def cancel(self) -> NoReturn:
```
glyph, line 70: Hmm. IMHO this should really be `-> None`, to match the interface.

(And this is not this change's problem, but putting "`@implementer` on a superclass that does *not* in fact implement the interface, and just promotes a static type-check error into a dynamic `NotImplementedError` is an antipattern, let's do less of this in the future.)

src/twisted/internet/base.py
glyph: That's not how you do that assertion, it would be something more like `asThreadReactor = IReactorThreads(self)`.

(But also let's not lose *too* much sleep over this, we do not test in an un-threaded configuration and I somewhat doubt that it works any more).

mypy.ini
```diff
@@ -102,10 +102,6 @@ check_untyped_defs = False
 allow_untyped_defs = True
 check_untyped_defs = False
 
-[mypy-twisted.internet._resolver]
-allow_untyped_defs = True
-check_untyped_defs = False
-
 [mypy-twisted.internet._signals]
```
glyph, line 105: ❤️

glyph (approved):
OK. I don't love the `TYPE_CHECKING` block but it you wouldn't mind hitting that in a follow-up, I'll just merge this now.

glyph:
Hopefully that trivial import conflict resolution will not make CI mad, but if I flubbed it anybody else can feel free to fix it up for merge.

src/twisted/internet/_resolver.py
clokep: Good catch. I'm surprised mypy allowed this in the first place.

src/twisted/internet/_resolver.py
clokep: I double checked this change is fine with mypy, so let's commit it! 👍

clokep:
Thanks for the review! 🎉

PR #11611, opened 2022-08-25
#11610 Switch disttrial AMP responders to use decorator syntax

### LINKED ISSUE #11610: The implementation of the AMP protocol used for `trial -j...` could use decorator syntax now
Long ago Python didn't necessarily support decorator syntax.  So, while `Command.responder` was a decorator, it wasn't always used in Twisted using decorator syntax.

Now Python necessarily supports the syntax and we can make the `trial -j ...` support code slightly more idiomatic.

exarkun:
## Scope and purpose

Fixes #11610

### DIFF
--- a/src/twisted/trial/_dist/worker.py
+++ b/src/twisted/trial/_dist/worker.py
@@ -98,14 +98,13 @@ class LocalWorkerAMP(AMP):
     Local implementation of the manager commands.
     """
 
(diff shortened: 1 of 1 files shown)


exarkun:
needs-review

adiroiban (approved):
Thanks for the refactoring. A bit easier to read.

PR #11613, opened 2022-08-25
#11612 Use deferLater for a simpler trial test

### LINKED ISSUE #11612: Some support code in trial's own unit tests could use deferLater to be slightly simpler
`src/twisted/trial/test/erroneous.py` has this:

```
    def test_fail(self):
        """                                                                                                                                                                       
        A test which fails in the callback of the returned L{defer.Deferred}.                                                                                                     
        """
        d = defer.Deferred()
        d.addCallback(self._later)
        reactor.callLater(0, d.callback, None)
        return d

    def _later(self, res):
        self.fail("I fail later")
```

`twisted.internet.task.deferLater` would simplify this a bit.

exarkun:
Fixes #11612

### DIFF
--- a/src/twisted/trial/test/erroneous.py
+++ b/src/twisted/trial/test/erroneous.py
@@ -14,6 +14,7 @@ this code is arranged.
 from unittest import skipIf
 
 from twisted.internet import defer, protocol, reactor
+from twisted.internet.task import deferLater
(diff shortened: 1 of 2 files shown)


exarkun:
needs-review

adiroiban (approved):
Looks  good. Thanks.

PR #11615, opened 2022-08-25
#11614 Annotate raiseException with NoReturn

### LINKED ISSUE #11614: Annotate twisted.python.failure.Failure.raiseException to show it never returns
`raiseException` always raises an exception.  This means that a pattern like:

```
if isinstance(failureOrSomething, Failure):
    failureOrSomething.raiseException()
use(failureOrSomething)
```

is safe.  However, mypy thinks it is unsafe because it doesn't know `raiseException` can't return.

exarkun:
## Scope and purpose

Fixes #11614

### DIFF
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -20,6 +20,7 @@ import linecache
 import sys
 from inspect import getmro
 from io import StringIO
+from typing import NoReturn
 
 import opcode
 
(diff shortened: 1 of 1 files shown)


exarkun:
needs-review

adiroiban (approved):
Thanks. All good.

PR #11617, opened 2022-08-25
#11616 disttrial worker test improvements

### LINKED ISSUE #11616: The test suite for worker result-reporting behavior involved in `trial -j ...` is unnecessarily complicated
The test suite is unnecessarily repetitive, it duplicates the implementation of certain protocol testing idioms, and it encodes implementation details instead of exercising the code under test in a realistic way.

exarkun:
Fixes #11616


src/twisted/trial/_dist/test/test_worker.py
```diff
@@ -86,197 +76,155 @@ class LocalWorkerAMPTests(TestCase):
...
-
-        self.managerAMP.callRemote = fakeCallRemote
 
         class StopTestResult(TestResult):
-            def stopTest(self, test):
-                stopped.append(test)
+            _stopped = False
+
+            def stopTest(self, test: PyUnitTestCase) -> None:
+                self._stopped = True
```
adiroiban:
```suggestion
            _stopped = []

            def stopTest(self, test: PyUnitTestCase) -> None:
                self._stopped.append(True)
```

I am still using append in tests at it can help detect corner cases in which something is called multiple times.

The append might look strange on a class variable, but this class should only be used once, inside this test.

The test code looks uglier ... but I think is best to have as strict of an assertion as possible.

The test code should be easy to maintain, but at the same time is there to prevent unexpected behaviour.

src/twisted/trial/_dist/test/test_worker.py
```diff
@@ -86,197 +76,155 @@ class LocalWorkerAMPTests(TestCase):
...
+                    isTuple(  # type: ignore[arg-type]
+                        equal_to(case),
+                        isFailure(
+                            # AssertionError is the type raised by TestCase.fail
+                            type=equal_to(AssertionError),
+                            value=equal_to(WorkerException("pyunit failure")),
+                        ),
+                    ),
+                ),
+            ),
```
adiroiban, line 148: I am not a big fan of these generic assertions.

we can leave this as it is, since `matched_result` is already in trunk.
I find the name a bit too generic.

Something to my taste.
```
assert_than(result, is_testrun_failure(case, type=AssertionError, value=WorkerException("pyunit failure")
```

and all the extra `isTuple`, `equal_to` and the general structure is handled by the dedicated matcher.

From my experience, with these specialiez markers it was easier to refactor the internal structure of a code as you might only need to update the specialized marker, without updating each test.

src/twisted/trial/_dist/test/test_worker.py
```diff
@@ -86,197 +76,155 @@ class LocalWorkerAMPTests(TestCase):
...
-        errorClass = fullyQualifiedName(ValueError)
-        d = self.worker.callRemote(
-            managercommands.AddError,
-            testName=self.testName,
-            error="error",
-            errorClass=errorClass,
-            frames=["file.py", "invalid code", "3"],
+        case = pyunitcases.PyUnitTest("test_error")
+        result = self.workerRunTest(case)
+        assert_that(
```
adiroiban, line 124: at least for me, this embedded structure is not easier to read ... 
but maybe is just my bias for old-school xunit type assertions.

```diff
- self.assertEqual(failure.type, ValueError)
- self.assertEqual(failure.value, WorkerException("error"))
+ self.assertTestRunFailure(self.result, ValueError, WorkerException("error"))
```

For this type of assertion, you only focus on the end/high-level result, and ignore the internal structure of a test case.

It might be an advantage or disadvantage.

I usually write a test in which the internal structure is validated, and then all the other tests only check the high-level values, ignore the internal structure.


With the generic `matches_result` you need to remember the full structure of the test case result, each time you write the test.

adiroiban (approved):
Thanks for the refactoring.

This is not my style of writing tests... and I am not happy with `matches_result`..and this method is already in trunk.

Overall, there are a few nice cleanups for the tests and I think is best to have it merged.

adiroiban (changes requested):
I now see that this PR was not ready for review.

--------

On a general note, I am -1 for using hypothesis in twisted, especially for general code.
It adds another thing that you must learn and be familiar with, before contributing to the project.

And for people not very familiar  with the library, might be a reason to write invalid tests.

src/twisted/trial/_dist/test/test_worker.py
exarkun: Perhaps an integer counter?

src/twisted/trial/_dist/test/test_worker.py
exarkun: > at least for me, this embedded structure is not easier to read ...

I don't object to that.  I think a lot of it is about familiarity and experience but it is also true that matcher-style tends to have fewer larger expressions rather than more smaller expressions.

Larger expressions in this style probably often take more practice to become comfortable reading.  The benefit that you get is when multiple parts of the value are wrong, you get to learn about all of them after one test run instead of only the one related to the very first assertion.

testtools allows the best of both worlds with an `expectThat` function which can add a test failure _and_ let the test keep running.  But Twisted decided to go with pyhamcrest instead...

Anyhow, I'm open to improving the readability of these tests before merging.

src/twisted/trial/_dist/test/test_worker.py
exarkun: I put the list back because the original test also asserted the right test case was passed as an argument.

src/twisted/trial/_dist/test/test_worker.py
exarkun: > + self.assertTestRunFailure(self.result, ValueError, WorkerException("error"))

Looking at this idea specifically in more detail, I don't see a good way to make it work.  The trouble is that `assertTestRunFailure` isn't reusable.  `test_runError`, `test_runFailure`, `test_runSkip`, etc all look similar but they differ in slight ways.

If we rename the suggested function to `assertTestRunError` (since this case is about test suite "errors" rather than "failures") and we implemented it _however_ (doesn't matter if we use hamcrest or not) then it is usable by exactly one test, `test_runError`.  So we need another one for `test_runFailure`, and another for `test_runSkip`, etc.

In this case I don't think there's any benefit to abstracting the details away from the test.  The purpose of the test method is exactly to encapsulate these details.

I can flatten the hamcrest code so it does repeated calls to `assert_that` instead of a single call with a complex matcher.  I think that change makes it easier for folks without a lot of practice writing and reading these and it doesn't hurt runtime properties _that_ much probably.

I don't think it makes sense to take the details of those assertions out of the test methods though.  Or at least, I don't see a way that doesn't just lead to more work at development time and more work for future readers of the code.

exarkun: I flattened the matcher in https://github.com/twisted/twisted/pull/11617/commits/6deb9f1bd6a237f3792ecc7e39f5e6d01f457fba - I'm curious if you find that to be any better

src/twisted/trial/_dist/test/test_worker.py
exarkun: This overlaps a bit with the comment above and I left a reply there.

exarkun:
needs-review

src/twisted/trial/_dist/test/test_worker.py
adiroiban: thanks for the info and the updated.
Agree.
The flattened version looks easier to read, at a first read.

I am sure that after reading and looking at the code a few more times, the nested structures will look much easier to read.
I just wanted to leave a "raw" feedback.

I guess that other people might want to help contribute the Twisted, but then they found some code that they don't understand and then they give up.

I am searching for the formula for creating an open source project that is easy to contribute to, while still keeping high code quality standards.

src/twisted/trial/_dist/worker.py
```diff
@@ -185,7 +185,7 @@ def addExpectedFailure(
         """
         Add an expected failure to the reporter.
         """
-        _todo = Todo(todo)
+        _todo = Todo("<unknown>" if todo is None else todo)
```
adiroiban, line 188: I guess that this change was done to make it easier to remove `Todo.__init__` while defining `reason: str` and not `Optional[str]`

I am happy to always have a string as the reason.

adiroiban (approved):
I think that this can be merged.

------

I am not a big fan of hamcrest , but I am not  -1 on it.

Jean-Paul, if you think that testtool.assertThat provides a better dev experience in terms of writing matches and diffs, I think we can still consider a switch.

---------

Using simple hypothesis strategies can make the code easier to use and encourage developers to write tests with explicit types.

I saw many bugs (in python 2.7) that were not discovered by a test, because the test was using something like `received = 'binary-data'` ... but that `binary-data` was a valid text that survived an accidental UTF-8 decode on it. 
And in prod you have real binary data and then UTF-8 decoding fails.

-----------

I only did a quick review.

I don't know if anyone will have time to review it.
I think is best to have it merge, than keep this PR in the review queue for many weeks

src/twisted/trial/_dist/worker.py
exarkun: Yep.  I also took a hint from the existing docstring for `Todo` which claimed `reason` was a string and said nothing about None.

exarkun:
> thanks for the info and the updated.
> Agree.
> The flattened version looks easier to read, at a first read.
> 
> I am sure that after reading and looking at the code a few more times, the nested structures will look much easier to read.
> I just wanted to leave a "raw" feedback.
> 
> I guess that other people might want to help contribute the Twisted, but then they found some code that they don't understand and then they give up.
> 
> I am searching for the formula for creating an open source project that is easy to contribute to, while still keeping high code quality standards.
> 

Thanks for the reviews @adiroiban .  I went ahead and flattened the hamcrest usage in the other `test_worker.py` tests to match the example I pushed earlier.  I don't mind trying to keep these expressions more on the simple side.  I think you make a good point about the consequences of such complex expressions on new contributors (or even existing contributors not familiar with this testing style).

I think having a richer matching library can also help us out and right now we're just starting to build one up for trial so we have some more work to do in that department.  Also, I think "it's hard to describe what result I want" is a bit of a code smell for "what you want is not great" so as we find it is hard to write good matchers we might decide we want to introduce better data structures into the implementation to make it easier.

exarkun:
It looks like the conch manhole tests are somewhat reliably broken on Windows now. :/

PR #11624, opened 2022-08-29
Debug windows conch failures

exarkun:
Just for debugging Windows CI failures for #11616

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,6 +51,7 @@ packages = find:
 test =
     cython-test-exception-raiser >= 1.0.2, <2
     PyHamcrest >= 1.9.0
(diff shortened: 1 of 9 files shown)


adiroiban:
Might be about time to move Windows test on GitHub action and enable https://github.com/mxschmitt/action-tmate

i think is the best experience for troubleshooting tests on Windows .

If a test fails on Win or macOs, after the test run fails I connect via SSH and re-run and do debugging directly on the VM 

We have this draft  https://github.com/twisted/twisted/pull/1647 so  maybe continue working on it.

exarkun:
This didn't really help and I'm done with it now anyway.

PR #11629, opened 2022-08-30
#11628 Run Windows CI tests on GitHub Actions.

### LINKED ISSUE #11628: Run Windows tests on GitHub Actions
For legacy reason we have the Windows and macOS tests running on Azure Pipelines 

I think that Azure Pipelines were implemented before GitHub actions.

We already have the Linux tests on GHA.

The scope of this ticket is to move Windows tests.

macOS tests can be moved in a separate ticket.

One big advantage for GHA, is that is easy to debug via the tmate action.

Also, in the future, with Windows an GHA it should be easier to gather test coverage and generate and independent coverage report.

adiroiban:
## Scope and purpose

Fixes #11628

See inline comments.

Besides enabling the Windows tests I have also enabled tmate to allow debugging Windows.

This is the main reason why I have moved the Windows tests to GHA.


.github/workflows/test.yaml
```diff
@@ -1,5 +1,9 @@
 # Try to get a short workflow name and a job name that start with Python
 # version to make it easier to check the status inside GitHub UI.
+#
+# When using external actions check that the extenral repos are permitted via
+# the GitHub configuration at https://github.com/twisted/twisted/settings/actions
+#
```
adiroiban, line 6: I don't know what to say about this.

At work, I fork all the actions that I use to have only reviews/approved code executed.

Another suggestion was to pin a specific commit for the external actions.


I went with configuring GitHub to only allow actions from selected repos.

.github/workflows/test.yaml
```diff
@@ -24,22 +28,44 @@ defaults:
   run:
     shell: bash
 
+env:
+  # Set to 'yes' to open a tunnel to GitHub's VMs through tmate on failures.
+  # Also increase timeout-minutes for the relevant OS when debugging remotely
+  # and you need a longer session.
+  TMATE_DEBUG: 'no'
```
adiroiban, line 42: We only want to enable tmate when we are to start a debug session...and not on any failure.

tmate is only triggered on failure

.github/workflows/test.yaml
```diff
@@ -24,22 +28,44 @@ defaults:
...
 
 jobs:
   testing:
-    runs-on: ubuntu-20.04
+    name: ${{ matrix.python-version }}-${{ matrix.job-name || 'lnx-default-tests' }}
+    # We have Ubuntu as the base for running agains multiple Python versions.
+    runs-on: "${{ matrix.runs-on || 'ubuntu-20.04'}}"
```
adiroiban: the 'ubuntu-20.04' is not explicitly defined in the matrix so that we don't have to repeat it each time we extend it.

We leave it empty and use a default value.

Another option is to define the default values via global environment variables.

```
env:
   DEFAULT_RUNS_ON: 'ubuntu-22.04'
jobs:
  testing:
    runs-on: "${{ matrix.runs-on || 'env.DEFAULT_RUNS_ON'}}"
```

azure-pipelines/run_test_steps.yml
```diff
@@ -65,7 +65,7 @@ steps:
     displayName: 'Run tests'
     continueOnError: ${{ parameters.allowFailure }}
     env:
-      TWISTED_REACTOR: ${{ parameters.windowsReactor }}
+      TRIAL_ARGS: --reactor=${{ parameters.windowsReactor }}
```
adiroiban, line 68: I guess that once we are happy with GHA, we can remote the Windows tests from Azure

tox.ini
```diff
@@ -93,12 +93,12 @@ commands =
     posix: python -c "print('Running on POSIX (no special dependencies)')"
 
     ; Run tests without wrapping them using coverage.
-    nocov: python -m twisted.trial --temp-directory={envtmpdir}/_trial_temp --reactor={env:TWISTED_REACTOR:default} --reporter={env:TRIAL_REPORTER:verbose} {env:TRIAL_ARGS:} {posargs:twisted}
+    nocov: python -m twisted.trial --temp-directory={envtmpdir}/_trial_temp --reporter={env:TRIAL_REPORTER:verbose} {env:TRIAL_ARGS:} {posargs:twisted}
```
adiroiban, line 96: For greater flexibiltiy, if `--reactor` is needed, it is passed via `env:TRIAL_ARGS`

`env:TWISTED_REACTOR` is no longer used.

adiroiban (commented):
I think this is ready for an initial review

A previous run at https://github.com/twisted/twisted/runs/8099083260?check_suite_focus=true

needs-review

adiroiban:
please review

.github/workflows/test.yaml
```diff
@@ -1,5 +1,9 @@
 # Try to get a short workflow name and a job name that start with Python
 # version to make it easier to check the status inside GitHub UI.
+#
+# When using external actions check that the extenral repos are permitted via
```
exarkun: typo in "external" here

.github/workflows/test.yaml
exarkun: I think pinning versions is a good idea.  Is it true that you have to pin a git revision hash to really have a stable version of the action (since tags can always be moved)?  Is pinning a git revision hash even stable (will git/github let you force push a new revision with the same hash but different contents)?

In any case this question seems larger than this PR.  Twisted is already using GitHub Actions and already using various actions in its configuration (even some external actions - from codecov and pypa)  So this issue doesn't need to block this PR.

tox.ini
exarkun: I think you mean `env:TRIAL_REACTOR` is no longer used, right?

azure-pipelines/run_test_steps.yml
exarkun: I thought this PR would remove the Azure pipeline for Windows CI.  I don't know what would separate "happy to merge this PR" from "happy to remove use of Azure for Windows".  They seem like the same thing to me.

Maybe even more strongly than that - if I see some Azure pipeline configuration being removed then I know exactly the minimum requirements for the new GitHub Actions configuration: it must test at least as much as whatever is being removed.  If the Azure pipeline configuration stays then the requirements for the new GitHub Actions configuration is much less clear.

.github/workflows/test.yaml
```diff
@@ -24,22 +28,42 @@ defaults:
...
     shell: bash
 
+env:
+  # Set to 'yes' to open a tunnel to GitHub's VMs through tmate on failures.
+  TMATE_DEBUG: 'no'
 
 jobs:
   testing:
-    runs-on: ubuntu-20.04
+    name: ${{ matrix.python-version }}-${{ matrix.job-name || 'lnx-default-tests' }}
```
exarkun: I think "linux" will be more clear in the UI than "lnx" and I don't think the extra 2 letters will make the names so long it becomes a problem.

.github/workflows/test.yaml
```diff
@@ -48,14 +72,34 @@ jobs:
         tox-wrapper: ['']
         # Tests are executed with the default target which is the full test suite.
         trial-target: ['']
+        # By default the coverate is not skipped.
```
exarkun: Typo in "coverage" here

.github/workflows/test.yaml
```diff
@@ -66,23 +110,27 @@ jobs:
...
           - python-version: pypy-3.7
             tox-env: alldeps-nocov-posix
+            job-name: 'pypy-no-coverage'
+            skip-coverage: yes
           # We still run some tests with coverage,
           # as there are test with specific code branches for pypy.
           - python-version: pypy-3.7
-            tox-env: alldeps-withcov-posix
             trial-target: twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests
+            job-name: 'pypy-coverage'
```
exarkun: We get some redundancy in the name from this -  eg "pypy-3.7-alldeps-nocov-posix".  Maybe `'coverage'` would be better?

.github/workflows/test.yaml
```diff
@@ -139,8 +191,15 @@ jobs:
       run: |
         ${{ matrix.tox-wrapper }} tox ${{ matrix.trial-target }}
 
+    # If one of the above steps fails, fire up tmate for remote debugging.
+    - name: Tmate debug on failure
+      if: ${{ !cancelled() && env.TMATE_DEBUG == 'yes' }}
```
exarkun: The workflow for using this action, then, will be to change `TMATE_DEBUG` at the top of the file to `'yes'` and re-run the build, then change it back to `'no'` again afterwards?

I guess there's a good reason not to have it turned on all the time?  https://github.com/mxschmitt/action-tmate#manually-triggered-debug looks like a slightly nicer workflow, maybe it's not too much more work to do it that way?

.github/workflows/test.yaml
```diff
@@ -139,8 +191,15 @@ jobs:
...
+      uses: mxschmitt/action-tmate@v3
+      with:
+        limit-access-to-actor: true
+
     - name: Prepare coverage
-      if: ${{ !cancelled() && contains(matrix['tox-env'], 'withcov') }}
+      if: ${{ !cancelled() && !matrix.skip-coverage }}
```
exarkun, line 231: I very much like this change to use a structured value (a boolean input parameter to the job) instead of some string scraping to determine the job configuration.  Thank you for this.

.github/workflows/test.yaml
```diff
@@ -99,7 +147,7 @@ jobs:
         echo "::set-output name=dir::$(pip cache dir)"
 
     - name: pip cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
```
exarkun, line 178: I don't know what's better about v3 than v2 but the number is bigger so I guess there must be something. :wink:

exarkun (changes requested):
Thanks (and for talking me through it on IRC too)!  This looks like a great start.  I pointed out a few minor things inline.

.github/workflows/test.yaml
adiroiban: In theory you can have a different commit with the same hash... but sneaking it unobserved is harder, than sneaking a security issue in the latest version that is automatically used by Twisted without any review, any time the tag is changed.

No need to block the PR. Just wanted to raise the idea.

.github/workflows/test.yaml
adiroiban: yes. I went for explicit 'with-coverage'

.github/workflows/test.yaml
adiroiban: While testing, the cache was timing out on Windows.

And I read there were some improvements to reduce timeout...so I gave it a try.
We will see if this is fixed.

If we continue to see cache timeout errors on Windows, then this is not fixed.

azure-pipelines/run_test_steps.yml
adiroiban: OK. I will remove the Windows CI from Azure.

And I will also need to update the branch protection via GitHub UI not to require Azure Windows.

tox.ini
adiroiban: Sorry. A typo. I editted the initial comment.

`env:TWISTED_REACTOR` is no longer used. You can set it via `TRIAL_ARGS`

.github/workflows/test.yaml
adiroiban: For me, editing the text was easier than using the Github action manual workflow trigger.

It is true that you can forget to set it back to `no` afterward.

-------

You don't want it turned on all the time, as then on failure it will "hang" and wait for someone to connect via SSH.

----------

I will push a commit that will also add manual triggers.
Not sure what are the permissions for manual triggers... that is, if you can trigger from a fork.

adiroiban:
I hope I have addressed all the comments.

please review

.github/workflows/test.yaml
```diff
@@ -66,23 +102,63 @@ jobs:
           # `python_requires` from `setup.cfg`.
           - python-version: '3.7.1'
             tox-env: nodeps-withcov-posix
+            job-name: 'nodeps-test'
+
+          # `noipv6` is created to make sure all is OK on an OS which doesn't
+          # have IPv6 available.
+          # Any supported Python version is OK for this job.
+          # We go with latest micro release for the smalles minor relase
```
exarkun:
```suggestion
          # We go with latest micro release for the smallest minor release
```

.github/workflows/test.yaml
```diff
@@ -66,23 +102,63 @@ jobs:
...
+          # that we support.
+          - python-version: '3.7'
+            noipv6: -noipv6
```
exarkun:
```suggestion
            noipv6: '-noipv6'
```

.github/workflows/test.yaml
```diff
@@ -66,23 +102,63 @@ jobs:
...
+            job-name: 'win-default-tests-iocp'
+            # Distributed trial is not yet suported on Windows.
+            trial-args: '--reactor=iocp'
+
           # On PYPY coverage is very slow (5min vs 30min) as there is no C
           # extension.
           # There is very little PYPY specific code so there is not much to
           # gain from reporting coverage.
           - python-version: pypy-3.7
             tox-env: alldeps-nocov-posix
```
exarkun:
```suggestion
          - python-version: 'pypy-3.7'
            tox-env: 'alldeps-nocov-posix'
```

.github/workflows/test.yaml
```diff
@@ -66,23 +102,63 @@ jobs:
...
+            job-name: 'no-coverage'
+            skip-coverage: yes
+
           # We still run some tests with coverage,
           # as there are test with specific code branches for pypy.
           - python-version: pypy-3.7
-            tox-env: alldeps-withcov-posix
             trial-target: twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests
```
exarkun:
```suggestion
          - python-version: 'pypy-3.7'
            trial-target: 'twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests'

File: .github/workflows/test.yaml [hunk 17]
```diff
@@ -149,11 +237,11 @@ jobs:
...
         python -m coverage report --skip-covered
 
-    - uses: codecov/codecov-action@v2
-      if: ${{ !cancelled() && contains(matrix['tox-env'], 'withcov') }}
+    - uses: codecov/codecov-action@v3
+      if: ${{ !cancelled() && !matrix['skip-coverage'] }}
       with:
         files: coverage.xml
-        name: lnx-${{ matrix.python-version }}-${{ matrix.tox-env }}${{ matrix.noipv6 }}
+        name: ${{ matrix.python-version || env.DEFAULT_PYTHON_VERSION }}-${{matrix.job-name || 'lnx-default-tests' }}
```
exarkun: I'm not sure `name` even does anything here (I can't see this name anywhere in the UI except where the config itself is being dumped) but ...

```suggestion
        name: ${{ matrix.python-version || env.DEFAULT_PYTHON_VERSION }}-${{matrix.job-name || 'default-tests' }}
```

enjoy github's delightful decision to render this suggestion as latex or some crazy thing

.github/workflows/test.yaml
```diff
@@ -10,6 +14,13 @@ on:
   pull_request:
     branches: [ trunk ]
 
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
+        type: boolean
+        description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
+        required: false
+        default: false
```
exarkun, line 23: Cool.  No doubt the only way to know if this actually works is to merge the PR and then try clicking buttons on GitHub Actions ... I'll certainly try that out as soon as this lands.

exarkun (approved):
Thanks!  This looks quite good.  I left some minor suggestions inline.  Take them if you like.

One thing I was not able to do was convince myself that `3.7-win-default-tests-select` is actually running the tests against Python 3.7 and `3.10-win-default-tests-iocp` is testing with Python 3.10.  Can you point at anything in the logs that reveals this?  Or maybe add something to the CI jobs somewhere that makes this clear.  Then I'm happy to see this merged.

adiroiban:
Thanks for the review. I will get this merged.

I saw a codecov failure ... so macOS should be next, and then try to run coverage report on GHA.

When tox runs, in any environment, local or remote-ci you will see the python version

See the link https://github.com/twisted/twisted/runs/8102345847?check_suite_focus=true#step:10:36

tox code at 

https://github.com/twisted/twisted/blob/26c3c04e55d0b25532709e1e1dc5d1433dde80f3/tox.ini#L86-L91

Local run is something like this

```diff
$ tox -e alldeps-nocov 
alldeps-nocov wheel-make: cleaning up build directory ...
alldeps-nocov wheel-make: commands[0] | pip wheel /home/adi/chevah/twisted --no-deps --use-pep517 --wheel-dir /home/adi/chevah/twisted/.tox/dist
Processing /home/adi/chevah/twisted
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: Twisted
  Building wheel for Twisted (pyproject.toml) ... done
  Created wheel for Twisted: filename=Twisted-22.4.0.post0-py3-none-any.whl size=3102521 
- [SNIP}
alldeps-nocov run-test-pre: PYTHONHASHSEED='544835508'
alldeps-nocov run-test: commands[0] | python /home/adi/chevah/twisted/admin/dump_all_version_info.py
+::group::Python info
+/home/adi/chevah/twisted/.tox/alldeps-nocov
+/home/adi/chevah/twisted/.tox/alldeps-nocov
+/home/adi/chevah/twisted/.tox/alldeps-nocov/bin/python
+3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
+linux
+::endgroup::
alldeps-nocov run-test: commands[1] | python -m twisted.trial --temp-directory=/home/adi/chevah/twisted/.tox/alldeps-nocov/tmp/_trial_temp --reporter=verbose twisted
```

.github/workflows/test.yaml
adiroiban: Yes. The GitHub UI is updated only after merge.

But you try the `TMATE_DEBUG: 'yes'` version.
That version is also ok for debugging the process itself.
So if the UI trigger doesn't work, you can  force tmate.

.github/workflows/test.yaml
adiroiban: This is for the Codecov reports UI.
I was using it to debug Codecov, as a way to check if the coverage report from a build was received by Codecov.

.github/workflows/test.yaml
```suggestion

          # On PYPY coverage is very slow (5min vs 30min) as there is no C
          # extension.
          # There is very little PYPY specific code so there is not much to
          # gain from reporting coverage.
          - python-version: 'pypy-3.7'
            tox-env: 'alldeps-nocov-posix'
```

adiroiban: Ok. This goes to the "delete-lines" and can't be applied by github.
Will manually apply it.

PR #11630, opened 2022-09-01
Makes the auto-index page HTML5 compliant

Fclem:
use case, for example : 
serving local pip simple index repository where pip clients requires (will be enforced starting at pip 22.2) a valid HTML5 pages as per PEP 503
see https://github.com/pypa/pip/issues/10825


adiroiban:
Thanks for the update.

This needs a newsfragmet to advertise in the next release notes that HTML5 is now supported.

And have at least one test to make sure we don't regress on this feature.

Thanks!

twm:
From the linked issue, I think that the test should verify that the page can be parsed with `html.parser`.

adiroiban:
so this needs-changes in order to have it merged.

PR #11632, opened 2022-09-02
#11631 GitHub Workflows security hardening

### LINKED ISSUE #11631: Remove unnecessary permissions in workflows
This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

sashashura:
Fixes #11631

This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,6 +22,9 @@ on:
         required: false
         default: false
 
+permissions:
+  contents: read
+
 # Only have a run a single parallel for each branch.
 # Runs for trunk are queues.
 # Older runs for non-trunk branches are cancelled and the jobs are executed
@@ -339,6 +342,8 @@ jobs:
   # process late in the release cycle,
   # The files are published only when a tag is created.
   release-publish:
+    permissions:
+      contents: write
     name: Check release and publish on twisted-* tag
     runs-on: 'ubuntu-20.04'
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -408,7 +408,6 @@ jobs:
           else
             echo "Branch not updated for not stable releases: $GITUB_REF"
           fi
-    
 
   # We have this job so that the PR can be blocked on a single job.
   # In this way, each time a job is modified,
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -381,7 +381,7 @@ jobs:
   update-stable-branch:
     permissions:
       contents: write
-    name: Update stable branch - on stable tag
+    name: Update stable branch for RTD - on tag
     runs-on: 'ubuntu-20.04'
     needs: [release-publish]
     if: startsWith(github.ref, 'refs/tags/twisted-')
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -374,6 +374,10 @@ jobs:
       with:
         password: ${{ secrets.PYPI_UPLOAD_TOKEN }}
 
+  # Read the Docs has no support for our "twisted-1.2.3" numbering convention
+  # and can't detect which tag is the stable one.
+  # A workaournd is to manually push to a "stable" branch to inform RTD
+  # that this is what we want at the "/en/stable" link.
   update-stable-branch:
     permissions:
       contents: write
@@ -423,6 +427,7 @@ jobs:
       - apidocs
       - static-checks
       - release-publish
+      - update-stable-branch
     steps:
       - name: Require all successes
         shell: python
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -437,4 +437,4 @@ jobs:
           import os
           import sys
           results = json.loads(os.environ["RESULTS"])
-          sys.exit(0 if all(result == "success" for result in results) else 1)
+          sys.exit(0 if all(result in ["success", "skipped"] for result in results) else 1)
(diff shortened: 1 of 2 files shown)


altendky:
Thanks for making me see this!  I was totally unaware of this feature.  On the off chance, do you know when this was added?  Or did I always just miss it...

sashashura:
Read/write permissions for non pull_request triggers were from very beginning. Explicit control of permissions was introduced about a year ago https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

.github/workflows/test.yaml
```diff
@@ -339,6 +342,8 @@ jobs:
   # process late in the release cycle,
   # The files are published only when a tag is created.
   release-publish:
+    permissions:
+      contents: write
```
adiroiban: ok. so we need this in write mode, only to update the "stable" branch.

The "stable" branch was added as a hack, since Read the docs, doesn't recognizes our `twisted-12.23.45` tag format, and can't determine the latest stable version.

Maybe for better security, the "stable" branch update can be moved into a separate job that is performed after release-publish.

So, this would go further with the principle of least privilege

adiroiban (approved):
Thanks for the update.
I guess that this is ok.
A bit better than what we had before, mostly future-proof.


I think we were always fine, as we make simple usage of GHA.

We don't use `pull_request_target`.

We have recently introduced `workflow_dispatch` but I think this was always safe.
I can't find much info about `workflow_dispatch` security.
I know Jean-Paul has tried using it with forks and it failed.

-------

Also, I hope we can keep GHA usage simple, to reduce our dependency on GHA and at the same time, improve security.

.github/workflows/test.yaml
sashashura: Yes, you got the idea! Restrict as much as possible. I just usually try to minimize the changes to avoid breaking anything. I have split it into a separate job. Please, take a look.

.github/workflows/test.yaml
```diff
@@ -394,7 +408,7 @@ jobs:
           else
             echo "Branch not updated for not stable releases: $GITUB_REF"
           fi
-
+
```
```suggestion

```

.github/workflows/test.yaml
```diff
@@ -371,8 +374,19 @@ jobs:
...
         password: ${{ secrets.PYPI_UPLOAD_TOKEN }}
 
+  # Read the Docs has no support for our "twisted-1.2.3" numbering convention
+  # and can't detect which tag is the stable one.
+  # A workaournd is to manually push to a "stable" branch to inform RTD
+  # that this is what we want at the "/en/stable" link.
+  update-stable-branch:
+    permissions:
+      contents: write
+    name: Update stable branch - on stable tag
```
```suggestion
    name: Update stable branch for RTD - on tag
```

adiroiban (approved):
Many thanks for the changes.

I push a change that adds some extra comment/docs for why we need the extra job.

I have triggered the auto-merge for this. I hope all will be ok.

Not sure how our 'all-success' works with the skip state.
I guess it might fail.

The idea is that we want `all-success` as a single way to monitor the success of the jobs for a commit... but we might need to remove the stable branch update job.

adiroiban:
Thanks, Sasha for the change.

Much appreciated to get help with the supply chain security.
It's a big topic these days and I think it was something overlooked for a long time.

---------

Before merging this I would like to get a second opinion from @altendky regarding https://github.com/twisted/twisted/pull/11632/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R440

I think that not failing on skip is ok.

-------


The previous issue was that on fail, the job was considered skipped, so we didn't had a final failure.

.github/workflows/test.yaml
```diff
@@ -423,4 +437,4 @@ jobs:
           import os
           import sys
           results = json.loads(os.environ["RESULTS"])
-          sys.exit(0 if all(result == "success" for result in results) else 1)
+          sys.exit(0 if all(result in ["success", "skipped"] for result in results) else 1)
```
adiroiban: Make the update stable really optional :)

```suggestion
          sys.exit(0 if all(result == "success" for result in results) else 1)
```

.github/workflows/test.yaml
```diff
@@ -413,6 +426,7 @@ jobs:
       - apidocs
       - static-checks
       - release-publish
+      - update-stable-branch
```
```suggestion
```

PR #11636, opened 2022-09-06
#11635 Allow mixed types in Headers types

### LINKED ISSUE #11635: `t.w.http_headers.Headers` methods `addRawHeader` and `setRawHeaders` are typed `AnyStr`
A call like `headers.addRawHeader(b'foo', 'bar')` is correct at runtime, but fails to typecheck because `AnyStr` is a type variable that constrains both parameters to be the same type.

Similar for `headers.setRawHeaders('foo', [b'ar'])`.

These calls are valid, so the constraint should be removed from `Headers`.

twm:
Fixes #11635.

Purpose is to enable typing of Treq.

### DIFF
--- a/src/twisted/web/http_headers.py
+++ b/src/twisted/web/http_headers.py
@@ -23,7 +23,6 @@ from typing import (
 
 from twisted.python.compat import cmp, comparable
 
-_S = Union[str, bytes]
 _T = TypeVar("_T")
 
 
@@ -232,15 +231,13 @@ class Headers:
                 "bytes or str" % (type(value),)
             )
 
-        _name = _sanitizeLinearWhitespace(self._encodeName(name))
-        if isinstance(value, str):
-            _value = value.encode("utf8")
-        else:
-            _value = value
-        self._rawHeaders[_name] = [
-            *self._rawHeaders.get(_name, ()),
-            _sanitizeLinearWhitespace(_value),
-        ]
+        self._rawHeaders.setdefault(
+            _sanitizeLinearWhitespace(self._encodeName(name)), []
+        ).append(
+            _sanitizeLinearWhitespace(
+                value.encode("utf8") if isinstance(value, str) else value
+            )
+        )
 
     @overload
     def getRawHeaders(self, name: AnyStr) -> Optional[Sequence[AnyStr]]:
--- a/src/twisted/web/http_headers.py
+++ b/src/twisted/web/http_headers.py
@@ -24,6 +24,7 @@ from typing import (
 
 from twisted.python.compat import cmp, comparable
 
+_S = Union[str, bytes]
 _T = TypeVar("_T")
 
 
@@ -84,7 +85,7 @@ class Headers:
 
     def __init__(
         self,
-        rawHeaders: Optional[Mapping[AnyStr, Sequence[AnyStr]]] = None,
+        rawHeaders: Optional[Mapping[_S, Sequence[_S]]] = None,
     ):
         self._rawHeaders: Dict[bytes, List[bytes]] = {}
         if rawHeaders is not None:
@@ -152,7 +153,7 @@ class Headers:
         """
         self._rawHeaders.pop(self._encodeName(name), None)
 
-    def setRawHeaders(self, name: AnyStr, values: Sequence[AnyStr]) -> None:
+    def setRawHeaders(self, name: _S, values: Sequence[_S]) -> None:
         """
         Sets the raw representation of the given header.
 
@@ -200,7 +201,7 @@ class Headers:
 
         self._rawHeaders[_name] = encodedValues
 
-    def addRawHeader(self, name: AnyStr, value: AnyStr) -> None:
+    def addRawHeader(self, name: _S, value: _S) -> None:
         """
         Add a new raw value for the given header.
 
--- a/src/twisted/web/http_headers.py
+++ b/src/twisted/web/http_headers.py
@@ -18,7 +18,6 @@ from typing import (
     Tuple,
     TypeVar,
     Union,
-    cast,
     overload,
 )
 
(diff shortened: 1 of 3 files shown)


twm:
please review

src/twisted/web/test/test_http_headers.py
```diff
@@ -668,3 +668,36 @@ def test_copy(self):
...
+    def test_addRawHeader(self) -> None:
+        """
+        L{Headers.addRawHeader} accepts mixed L{str} and L{bytes}.
+        """
+        h = Headers()
+        h.addRawHeader(b"bytes", "str")
+        h.addRawHeader("str", b"bytes")
+
+        self.assertEqual(h.getRawHeaders(b"Bytes"), [b"str"])
+        self.assertEqual(h.getRawHeaders("Str"), ["bytes"])
```
adiroiban, line 688: should we add a UTF-8 value to  cover the default encoding.

And maybe also update the docstring to confirm that we don't use the `ascii`` decoder.

```python
        h.addRawHeader("str", b"b'big-\xe2\x98\x89'")

        self.assertEqual(h.getRawHeaders("Str"), ["big-\N{sun}"])
```

and similar for str to bytes.

src/twisted/web/test/test_http_headers.py
```diff
@@ -668,3 +668,36 @@ def test_copy(self):
...
+
+    def test_setRawHeaders(self) -> None:
+        """
+        L{Headers.setRawHeaders} accepts mixed L{str} and L{bytes}.
```
adiroiban, line 692: Maybe be explicit about what is the exected behaviour when mixing 

```suggestion
        L{Headers.setRawHeaders} accepts mixed L{str} and L{bytes} keys and value.
        
        The resulting value has the same type as the key.
```

src/twisted/web/test/test_http_headers.py
```diff
@@ -668,3 +668,36 @@ def test_copy(self):
...
+        """
+        h = Headers()
+        h.setRawHeaders(b"bytes", [b"bytes"])
```
adiroiban, line 695: Maybe hypothesis can help here... but I didnt' had time to check hypothesis data generator.

The idea, is that when we have bytes, we should have a value that is never a valid ascii or utf-8. In this way, it should detect bugs caused by accidental decoding attempts.  

```suggestion
        h.setRawHeaders(b'key-with-broken-utf8-\xe2\x98', [b'value-with-broken-utf8-\xe2\x98'])
```

but this is not a blocker, as may other tests in Twisted are written in this way.

It's just someting that maybe we can keep in mind and have in new tests.

src/twisted/web/http_headers.py
```diff
@@ -152,7 +151,21 @@ def removeHeader(self, name: AnyStr) -> None:
         """
         self._rawHeaders.pop(self._encodeName(name), None)
 
-    def setRawHeaders(self, name: AnyStr, values: Sequence[AnyStr]) -> None:
+    @overload
+    def setRawHeaders(self, name: Union[str, bytes], values: Sequence[bytes]) -> None:
```
adiroiban, line 155: I don't understand this... but I don't know much about mypy and typing.

Maybe add a comment to explain why this extra typing definition is required.

I guess that the reason is something like this.

You can set a sequence of either str or bytes, but you can't set a sequence of mixed str and bytes.

adiroiban (approved):
Thanks! Looks good. 
Only minor comments. 
Nothing to block this PR.

src/twisted/web/http_headers.py
exarkun: I'm also not sure I follow.  I don't think @adiroiban 's guess is correct because there's also a version for `Sequence[Union[str, bytes]]` below...

src/twisted/web/test/test_http_headers.py
twm: The goal of these tests is actually the typechecker coverage, not runtime coverage. The values don't matter, and I believe that encoding/decoding is covered elsewhere (and I _know_ the utf-8 misfeature is covered in treq's suite).

src/twisted/web/test/test_http_headers.py
twm: Same &mdash; the typechecker doesn't care.

src/twisted/web/test/test_http_headers.py
exarkun: I don't think this is a blocker for this PR but I wonder if Twisted should try to adopt some other convention for "type tests" - something that doesn't involve running code when the intent of the test is not to verify runtime behavior.

For example, you could have a function like:

```python
def f(s: str, b: bytes, h: Headers) -> None:
    h.addRawHeader(s, s)
    h.addRawHeader(s, b)
    h.addRawHeader(b, s)
    h.addRawHeader(b, b)
    ...
```

... somewhere the typechecker can find it.  Nothing ever needs to actually call it for mypy to report problems with the usage, if there are any.

src/twisted/web/http_headers.py
twm: Yes, the runtime behavior accepts any mix of str and bytes. The overloads were the only way I found to get it to not overconstrain `values`. The issue is that (IIRC, it's been a few weeks) MyPy doesn't accept e.g. `list[str]` as `Sequence[str | bytes]` and you can't express it with variance + a type variable because there is no base type other than object. So I wrote out all the valid cases in the tests and added overloads until MyPy accepted them.

src/twisted/web/test/test_http_headers.py
twm: The suggestion is incorrect, there is no resulting value.

src/twisted/web/test/test_http_headers.py
twm: Yes this would be nice. We'd need to exclude these from coverage as well.

twm: Filed here: https://github.com/twisted/twisted/issues/11757#issue-1449259556

PR #11639, opened 2022-09-06
#11638 manhole vs excepthook

### LINKED ISSUE #11638: twisted.conch.manhole fails to display tracebacks correctly if sys.excepthook has been modified
manhole inherits from `code.InteractiveInterpreter` to handle some of its REPL-y behavior.  `code.InteractiveInterpreter` varies some of its behavior on whether `sys.excepthook` has been fiddled with.  manhole assumes it will _always_ get the behavior you get if `sys.excepthook` hasn't been tampered with so it breaks (and so does one of its unit tests) if it has been tampered with.

This is an obscure hook that's mostly left alone (which is why so many years passed before it was a problem, I suppose).  An example of a library that does stuff with `sys.excepthook` is `exceptiongroup`.  It turns out Hypothesis depends on `exceptiongroup` so if we try to use Hypothesis in Twisted then `sys.excepthook` gets tampered with an manhole tests start failing.

Here's an exceptiongroup bug report - https://github.com/agronholm/exceptiongroup/issues/26 - but whether this is fixable in exceptiongroup or not is unclear (seems like perhaps not).  Probably the stdlib should stop having this wonky conditional but ...

exarkun:
## Scope and purpose

Fixes #11638

This makes manhole itself override sys.excepthook while evaluating code so it can control what happens when an exception is raised.

### DIFF
--- a/src/twisted/conch/test/test_manhole.py
+++ b/src/twisted/conch/test/test_manhole.py
@@ -8,6 +8,7 @@
 Tests for L{twisted.conch.manhole}.
 """
 
+import sys
 import traceback
 from typing import Optional
 
@@ -148,9 +149,6 @@ class WriterTests(unittest.TestCase):
 class ManholeLoopbackMixin:
     serverProtocol = manhole.ColoredManhole
 
-    def wfd(self, d):
-        return defer.waitForDeferred(d)
-
     def test_SimpleExpression(self):
         """
         Evaluate simple expression.
@@ -244,10 +242,21 @@ class ManholeLoopbackMixin:
                     + defaultFunctionName.encode("utf-8"),
                     b"Exception: foo bar baz",
                     b">>> done",
-                ]
+                ],
             )
 
-        return done.addCallback(finished)
+        done.addCallback(finished)
+        return done
+
+    def test_ExceptionWithCustomExcepthook(
+        self,
+    ):
+        """
+        Raised exceptions are handled the same way even if L{sys.excepthook}
+        has been modified from its original value.
+        """
+        self.patch(sys, "excepthook", lambda *args: None)
+        return self.test_Exception()
 
     def test_ControlC(self):
         """
--- a/src/twisted/conch/manhole.py
+++ b/src/twisted/conch/manhole.py
@@ -17,11 +17,15 @@ import code
 import sys
 import tokenize
 from io import BytesIO
+from traceback import format_exception
+from types import TracebackType
+from typing import Type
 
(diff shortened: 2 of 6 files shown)


exarkun:
please review

exarkun:
The codecov/patch failure seems _mostly_ to be due to flaky coverage in unrelated modules though there is one new partially covered line in the _tests_ added by this PR.

The report says

> self.patch(sys, "excepthook", lambda *args: None)

is partially covered.  This is because the lambda function is never called.  This is expected because the point of the PR is to make sure `sys.excepthook` is not called, only manhole's own excepthook is called.  I'm not sure if there's something different Twisted would like in how this coverage situation is handled.

src/twisted/conch/test/test_manhole.py
```diff
@@ -244,10 +242,21 @@ def finished(ign):
...
+        return done
+
+    def test_ExceptionWithCustomExcepthook(
+        self,
+    ):
+        """
+        Raised exceptions are handled the same way even if L{sys.excepthook}
+        has been modified from its original value.
+        """
+        self.patch(sys, "excepthook", lambda *args: None)
```
adiroiban, line 258: Maybe something like this.
This should make coverage happy and should also help with the docs.

There is an exception in the coveragerc config file for `raise NotImplementedError`

This is just a suggetion/idea and might not work,

```suggestion
        def never_called(*args):
           """
           Coverage will complain that this is not called.
           but that is the purpose of this code.
           It is placeholder for the global sys.excepthook
           and the test is here to make sure the global sys.excepthook
           is not called.
           """
           raise NotImplementedError("never_called shouldn't be called")

        self.patch(sys, "excepthook", never_called)
```

src/twisted/conch/test/test_manhole.py
```diff
@@ -244,10 +242,21 @@ def finished(ign):
...
 
-        return done.addCallback(finished)
+        done.addCallback(finished)
+        return done
+
+    def test_ExceptionWithCustomExcepthook(
+        self,
+    ):
+        """
+        Raised exceptions are handled the same way even if L{sys.excepthook}
```
adiroiban, line 255: The same way as .... ?

I understand the purpose of the test in the context of the PR.
But by reading the test code and test docstring, it is not obvious what is tested here.

It also might be just me being tired after work.

adiroiban (approved):
Thanks for the PR.
The code looks good... much better than trunk :)

I left a minor comment.

PR #11643, opened 2022-09-06
#11642 Create a codecov.io report only for the test code itself.

### LINKED ISSUE #11642: Add separate coverage report for the test code
With codecov.io it should be easy to generate a separate report for the test code.

The idea is that the test code should always be 100%

If the test case don't have 100% coverage it means that at least one test is always skipped.

And it means that this is either dead code... or a test that is accidentally skipped.

So I hope that a separate coverage report for the test can help detect tests that are accidentally always skipped.

adiroiban:
## Scope and purpose

Fixes #11642

See ticket description.

This helps to uncover some test code that is always skipped.

There is a new CI check introduced by this PR ...and for now it is red

image

----------

I added "pass" as an exception for coverage.

I think we should not use `pass` but instead use a doctring to document why a `pass` is needed.
And this will make coverage happy and can also serve as documentation.

---------

I plan to remove the tests are the deprecated as part of this PR. 

-----

Not sure if this PR should be merged with the CI being red, as any future PR will have that CI failed and might confuse the developers.

---------

Rather we can use it to check the status of our test code.

And either fix the coverage in this PR, or fix it in other PRs and the merge this.


adiroiban:
The important change of this PR is this new (sub)coverage report

We can see that we only run 96.56% of our test code... this means that some tests might be always skipped

The PR is just to get some feedback for the separate test coverage report.


image


needs-review

glyph (changes requested):
Looks like this still puts a ❌ on the PR's overall status, which excludes it from this query: https://github.com/twisted/twisted/pulls?q=is%3Apr+is%3Aopen+label%3Aneeds-review+status%3Asuccess+

Can the configuration be adjusted so it shows the number but does *not* establish a pass/fail status, and thereby tank the green checkmark on the PR?

adiroiban:
Thanks for checking this PR.

From the PR description ;)

> Not sure if this PR should be merged with the CI being red, as any future PR will have that CI failed and might confuse the developers.

I have created this PR to try to start a conversation around the importance of 100% coverage for our testing code.

If we agree that 100% code coverage for the test code is important,
my plan is to keep this unmerged and have separated PRs to improve the test coverage.
Once trunk is green, this can be merged and the check should protected future PR from introducing testing code that is never executed by our CI.

------------

I don't see much value in adding a check now without failing.

If someone want to see the coverage for the tests, they can go to codecov.io .
For example this URL https://app.codecov.io/gh/twisted/twisted/tree/trunk/src/twisted/test

codecov.yml
```diff
@@ -31,7 +31,16 @@ coverage:
...
+      tests:
+        # We want a separate report just for the test code itself.
+        # If a test code is not executed it means we always skip a test
+        # or a test is accidentally always skipped.
+        # And we see that the report is green, but some bugs might be hidden
+        # due to test not being executed.
+        target: 100%
+        paths:
+          - "src/twisted/*/test/**"
+          - "src/twisted/test/**"
```
adiroiban, line 43: For the initial merge of this functionality it might be easier to pick a sub-set of those tests, make sure those tests have 100% coverage and add them here.

Later we can extend other tests to have 100% coverage and extend the rules

I expect that by looking for 100% covegage for the test, it should be easier to identify some code that is no longer use.

```suggestion
        paths:
          - "src/twisted/conch/test/**"
          - "src/twisted/test/**"
```

adiroiban:
this is for now POC 

we can't merge this as we don't have 100% coverage for the tests.

I plan to update twisted.conch tests with 100% coverage and then enable them here

Later we can look at twisted.web ... and others

PR #11644, opened 2022-09-07
#10314 Better worker-side error handling in disttrial

### LINKED ISSUE #10314: Failure: twisted.protocols.amp.UnknownRemoteError: Code<UNKNOWN>: Unknown Error
|-|-|
|Trac ID|trac#10314|
|Type|enhancement|
|Created|2022-03-08 21:28:12Z|

Sometimes an application-level test failure causes `trial -jN ...` to report this error:

```
Failure: twisted.protocols.amp.UnknownRemoteError: Code<UNKNOWN>: Unknown Error
[FAIL]
```

and then immediately exit with no further details.

The test which encountered the error and the details of the error are all lost, making it extremely difficult to diagnose and fix the problem.

`trial -jN ...` should:

* never fail this way
* always report detailed errors in application code

```
trac-id__10314 10314
type__enhancement enhancement
reporter__exarkun exarkun
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__trial trial
keywords__None None
time__1646774892879232 1646774892879232
changetime__1646774892879232 1646774892879232
version__None None
owner__None None

```
</details>

exarkun:
## Scope and purpose

Fixes #10314 

This adds error handling (any, at all!) to the "workerreporter" that manages communication of test results from the disttrial worker processes back to the parent process so that such errors won't cause the whole trial process to exit without a result report.

It also changes the network protocol between workers and parent to stream error text so that text longer than the AMP limit of 64k makes it back to the parent without error.


exarkun:
please review

src/twisted/trial/_dist/stream.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def write(self, streamId: int, chunk: str) -> None:
+        """
+        Write to an open stream using its unique identifier.
+
+        :raise KeyError: If there is no such open stream.
+        """
+        self._streams[streamId].append(chunk)
+
+    def close(self, streamId: int) -> List[str]:
```
adiroiban: I am a bit confused by the `.close()` as I am not normally expecting it to return the content.

But the method is well documented...so not a blocker.

Maybe consider using:

* open / start
* write / append
* finalize()

I think that finalizing is also used by crypto hash/ciphers to end the process and return something.

src/twisted/trial/_dist/worker.py
```diff
@@ -137,15 +185,17 @@ def _buildFailure(
...
         self,
         testName: str,
-        error: str,
         errorClass: str,
-        frames: List[str],
+        errorStreamId: int,
+        framesStreamId: int,
     ) -> Dict[str, bool]:
         """
         Add an error to the reporter.
```
adiroiban, line 201: Maybe extend the docstring to inform that before adding the error,
the error "content" should be previously sent via the streaming AMP interface.


Same for addFailure

adiroiban (commented):
I did a very quick review.
I am leaving the command as I don't know when I will have time to do a full review.

If no other developer will have time to review this in the next few days, I think that this should be merge.

I don't think is ok to block a merge for a PR that has tests for the new code, is well documented and the tests are green.

src/twisted/trial/_dist/stream.py
exarkun: Maybe `finish` or `popContent`?

adiroiban: I think `finish` is ok, as it should imply that no future actions are valid for the streamID.

`popContent` might lead you to think that you "flush" the current content, but you can continue to write future content.

exarkun: Addressed this in the version on at https://github.com/twisted/twisted/pull/11662

src/twisted/trial/_dist/test/test_workertrial.py
```diff
@@ -101,7 +101,9 @@ def test_forwardCommand(self):
         self.readStream = clientTransport.io
         self.readStream.seek(0, 0)
         main(self.fdopen)
-        self.assertIn(b"No module named 'doesntexist'", self.writeStream.getvalue())
+        # Just brazenly encode irrelevant implementation details here, why
+        # not.
```
adiroiban, line 105: :)

src/twisted/trial/_dist/worker.py
```diff
@@ -68,20 +79,44 @@ class WorkerProtocol(AMP):
...
+                    # already gone wrong we shouldn't be extremely confident
+                    # that this will succeed.  So we will also log it and any
+                    # errors reporting it to our local log.
+                    self.logger.failure(
+                        "Result reporting for {id} failed",
+                        failure=result,
+                        id=testCase,
+                    )
+                    try:
+                        await self._result._addErrorFallible(testCase, result)
```
adiroiban: Maybe have it a public interface

```suggestion
                        await self._result.addErrorFallible(testCase, result)
```

The idea is that maybe in the future we would want to run some code scanners to detect for private access.

src/twisted/trial/_dist/workerreporter.py
```diff
@@ -56,48 +215,85 @@ def _getFrames(self, failure: Failure) -> List[str]:
...
         testName = test.id()
-        self.ampProtocol.callRemote(managercommands.AddSuccess, testName=testName)
+        self._call(
+            lambda: self.ampProtocol.callRemote(  # type: ignore[no-any-return]
+                managercommands.AddSuccess, testName=testName
+            )
+        )
 
-    def addError(self, test, error):
+    async def _addErrorFallible(self, testName: str, errorObj: TrialFailure) -> None:
```
adiroiban: Maybe make it public.

The docs are good enough and I hope people will not abuse it.

```suggestion
    async def addErrorFallible(self, testName: str, errorObj: TrialFailure) -> None:
```

adiroiban (commented):
It looks good. Only minor comments.

-------

src/twisted/trial/test/erroneous.py
```diff
@@ -109,12 +124,26 @@ def test_fail(self) -> defer.Deferred[None]:
...
+        """
+        return deferLater(reactor, 0, self.fail, "I fail later: " + "x" * 2 ** 16)  # type: ignore[arg-type]
+
+    def test_exception(self) -> None:
         """
         A test which raises an exception synchronously.
         """
         raise Exception(self.text)
 
+    def test_exceptionGreaterThan64k(self) -> None:
```
adiroiban, line 142: Maybe add an info that this test is actually executed as part `[test_workerreporter.py](https://github.com/twisted/twisted/pull/11644/files#diff-20c8b908f85e8394203296722e0e1dc7013d9c1fdf8d4d86e7e354634da7f3e0)

the top docstring of this module talks about `L{twisted.trial.test.test_tests}`

My preference is to reduce the size of these external files... and for example have one separate file for each test

Or better, each actual tests will generate a package inside temp and will run the tests from there.

----- 

but I ok not to have the test code auto-generated as part of this PR.

but better docstring would help.

src/twisted/trial/test/erroneous.py
```diff
@@ -109,12 +124,26 @@ def test_fail(self) -> defer.Deferred[None]:
         """
         return deferLater(reactor, 0, self.fail, "I fail later")  # type: ignore[arg-type]
 
-    def test_exception(self):
+    def test_failGreaterThan64k(self) -> defer.Deferred[None]:
+        """
+        A test which fails in the callback of the returned L{defer.Deferred}
+        with a very long string.
+        """
```
adiroiban, line 133:
```suggestion
        """
        A test which returns a L{Failure} in the callback of the returned L{defer.Deferred}
        with a very long string.
        
        Note: there is a separate test for checking the case in which an
        exception is raised.
        
        This is triggered via L{test_addFailureGreaterThan64k}.
        """
```

adiroiban (commented):
I left a few other minor comments.

I still have to review - `_runErrorTest` and `WorkerProtocol.run`

I will try to continue the review later today.

src/twisted/trial/_dist/worker.py
```diff
@@ -68,20 +79,44 @@ class WorkerProtocol(AMP):
...
-        case = self._loader.loadByName(testCase)
-        suite = TrialSuite([case], self._forceGarbageCollection)
-        suite.run(self._result)
-        return {"success": True}
+        with self._result.gatherReportingResults() as results:
+            case = self._loader.loadByName(testCase)
+            suite = TrialSuite([case], self._forceGarbageCollection)
+            suite.run(self._result)
+
+            allSucceeded = True
```
adiroiban: do we need this to be inside the `gatherReportingResults()` context manager?

src/twisted/trial/_dist/worker.py
```diff
@@ -68,20 +79,44 @@ class WorkerProtocol(AMP):
...
+            for (success, result) in await DeferredList(results, consumeErrors=True):
+                if not success:
+                    allSucceeded = False
```
adiroiban: Maybe reduce the indentation and add explicit handling of the success case.

```suggestion
            for (success, result) in await DeferredList(results, consumeErrors=True):
                if success:
                    # Test execution was successful and there is no need for extra
                    # reporting.
                    continue

                # Report the failed test.
                allSucceeded = False
```

src/twisted/trial/_dist/worker.py
```diff
@@ -68,20 +79,44 @@ class WorkerProtocol(AMP):
...
+                    # that this will succeed.  So we will also log it and any
+                    # errors reporting it to our local log.
+                    self.logger.failure(
+                        "Result reporting for {id} failed",
+                        failure=result,
+                        id=testCase,
+                    )
+                    try:
+                        await self._result.addErrorFallible(testCase, result)
+                    except BaseException:
```
adiroiban:
```suggestion
                    except BaseException:
                        # We failed to report the failure to the centralized process.
                        # Just add extra local logging, as there is not much else we can do.
```

adiroiban (approved):
Looks good.

Only a medium comment regarding the usage of context manager.

But I think it can be fixed and merged without another round of reviews.

src/twisted/trial/test/erroneous.py
exarkun: Some of the other tests in this module are also now used by `twisted.trial._dist.test` so I added a note about this to the module docstring.

I think that whatever the eventual factoring, there is still going to have to be sharing between `twisted.trial.test` and `twisted.trial._dist.test` because we want to be sure that trial and disttrial have analogous behaviors for some set of inputs (test cases).  They should share the definition of those inputs so that we know they're really the same.

src/twisted/trial/test/erroneous.py
exarkun: Hmm.  I'm not sure about this.  The last callback on the returned `Deferred` doesn't do `return Failure(...)` but `self.fail(...)`.  It will actually raise an exception.  Of course, this is turned into a `Failure` by `Deferred` so I'm not sure it's possible to tell the difference between these cases (except the raised exception might have more stack frames in its traceback).

src/twisted/trial/_dist/worker.py
exarkun: nope, not at all, and it's better outside, thanks.

PR #11646, opened 2022-09-07
#11645 all rtd formats

### LINKED ISSUE #11645: enable "all" readthedocs formats
HTMLZip is probably the most useful format for people to download, as we don't really QA the PDF or Epub versions of the docs, but people view the HTML site all the time.  However, folks might still like the formats we provide, and might like other formats in the future. It appears that we accidentally disabled this format when creating our `.readthedocs.yml`, so let's just set it back to the default (of `all`).

glyph:
## Scope and purpose

Fixes #11645

## Contributor Checklist:

* Once all checks are green, request a review by leaving a comment that contains exactly the string `needs-review`.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,9 +8,7 @@ sphinx:
   # This can be re-enable once RTD is updated with https://github.com/readthedocs/readthedocs.org/issues/7116
   fail_on_warning: False
 
-formats:
-  - pdf
-  - epub
+formats: all
 
 python:
   version: 3.8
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -8,9 +8,7 @@ sphinx:
   # This can be re-enable once RTD is updated with https://github.com/readthedocs/readthedocs.org/issues/7116
   fail_on_warning: False
 
-formats:
-  - pdf
-  - epub
+formats: all
```
adiroiban, line 20: Maybe add a comment that we have "all" as it should include htmlzip and we want htmlzip for offline docs readers.

Just a help with someone touching this in the future.
Most probably that person will not check the whole history of this line..
and might think that "all" is excessive and html and PDF should be enough.

glyph:
needs-review

adiroiban (approved):
Thanks.

Looks good.
I hope RTD build will pass.

PR #11654, opened 2022-09-10
#6607 - ckeygen should provide a default for the keyfile

### LINKED ISSUE #6607: ckeygen doesn't provide a default for the key file
|[therve's avatar](https://github.com/therve)| @therve reported|
|-|-|
|Trac ID|trac#6607|
|Type|enhancement|
|Created|2013-06-30 12:46:51Z|
|Branch|https://github.com/twisted/twisted/tree/ckeygen-default-file-6607-2|

For example:

```
$ ckeygen -p
Enter file in which the key is (/home/therve/.ssh/id_rsa):
```

It shows a file in parenthesis, but doesn't actually use that as a default value if nothing is specified. It should.

Attachments:

* [6607.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/1c7/1c7eb6b64f78e03c1fdd3d4455e126ef6a71e674/90f10d0cebc07b540378dea2d7c623b7239edea8.patch) (1535 bytes) - added by Saurabh on 2013-10-09 03:47:30Z - 
* [6607_2.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/1c7/1c7eb6b64f78e03c1fdd3d4455e126ef6a71e674/e6fc01ed0871bd67d7731d9188f04cc44261313e.patch) (5052 bytes) - added by Saurabh on 2013-11-07 13:33:47Z - 
* [6607_3.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/1c7/1c7eb6b64f78e03c1fdd3d4455e126ef6a71e674/1dc21ba24ea76dfc2dc75d68e65c2791d282abb4.patch) (5063 bytes) - added by Saurabh on 2013-11-09 04:52:25Z - 

```
trac-id__6607 6607
type__enhancement enhancement
reporter__therve therve
priority__low low
milestone__ 
branch__branches_ckeygen_default_file_6607_2 branches/ckeygen-default-file-6607-2
branch_author__lvh__glyph lvh, glyph
status__new new
resolution__ 
component__conch conch
keywords__easy easy
time__1372596411000000 1372596411000000
changetime__1423033410714556 1423033410714556
version__None None
owner__ioggstream ioggstream
cc__z3p
```
</details>

eevelweezel:
## Scope and purpose

Fixes #6607

ckeygen doesn't provide a default for the key file, even though a default appears in the prompt, https://github.com/twisted/twisted/issues/6607

## Contributor Checklist:


eevelweezel:
needs-review

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,46 @@ def test_changePassphraseSubtypeV1(self):
...
+        options = {"filename": ""}
+        getKeyOrDefault(options)
+        self.assertNotEqual(
+            options["filename"],
+            "",
+        )
+
+    def test_handleBadFileDisplayPublicKey(self):
+        """
+        Ensure FileNotFoundError is handled for an invalid filename.
```
adiroiban: I don't understand the description of this test.
`FileNotFoundError` might be a private implementation details that this test should not care about.

I guess that a description for this test could be.

```suggestion
        When the provided filename via the CLI options is not found
        `displayPublicKey` raises system exits.
        A detail about the error is sent to the standard error.
```

and maybe the name of the tests should be

```diff
-test_handleBadFileDisplayPublicKey
+test_displayPublicKeyHandleFileNotFound
```

Start the name with the target of the test.

Also, it would be nice for the test to assert the condition error.

Maybe SystemExit is raised due to an internal error, and not due to file not found.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,10 +203,18 @@ def _defaultPrivateKeySubtype(keyType):
         return "PEM"
 
 
-def printFingerprint(options):
+def getKeyOrDefault(options):
```
adiroiban: Do we need it as a public API.

----

Public API or private API, I think that this needs a docstring.


---------

I know that `printFingerprint` is public and without documentation, but I think that for the new code we should try to reduce the public interface, and add documentation.

and mabye also add type hinting?

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,10 +203,18 @@ def _defaultPrivateKeySubtype(keyType):
...
-        options["filename"] = input("Enter file in which the key is (%s): " % filename)
+        if platform.system() == "Windows":
+            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
+        options["filename"] = (
+            input("Enter file in which the key is (%s): " % filename) or filename
+        )
+
+
+def printFingerprint(options):
+    getKeyOrDefault(options)
```
adiroiban: this is kind of auto-magic code.

Maybe this will make it easier to read and understand what it does

```suggestion
    options["filename"] = resolveFilename(options)
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -311,6 +322,8 @@ def _saveKey(key, options):
     keyTypeName = KeyTypeMapping[key.type()]
     if not options["filename"]:
         defaultPath = os.path.expanduser(f"~/.ssh/id_{keyTypeName}")
+        if platform.system() == "Windows":
+            defaultPath = os.path.expanduser(fr"~\.ssh\id_{keyTypeName}")
```
adiroiban: Do we have tests for this ?

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,46 @@ def test_changePassphraseSubtypeV1(self):
...
+
+        def mock_input(*args):
+            return ""
+
+        import builtins
+
+        self.patch(builtins, "input", mock_input)
+        options = {"filename": ""}
+        getKeyOrDefault(options)
+        self.assertNotEqual(
```
adiroiban: I am not sure that not being empty is a good enough test...

but I don't know what to suggest.

Can we assume that the path will contain the current username ?

adiroiban:
Thanks for helping with this.

I did just a quick review. Feel free to ignore it and wait for another developer to look over this.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: No, this really doesn't need to be public. Agreed, it could use a docstring. Updated.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: The tests that cover this are platform-independent, and should pass whether run on Linux or Windows. It's not going to cover both branches in any given run, as a result.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: If options["filename"] is empty, this function will insert the default path for  id_rsa in .ssh, expanded based on user's home and Windows/Linux.  I was trying to keep this minimal so it wouldn't fail based on OS, but we could possibly assert on both "not empty" and "contains 'id_rsa'."

eevelweezel: Updated.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: I've amended the test names and added an assert that the error message contains that specific text.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: This should be covered by existing tests when run on a windows image, it just changes the default to a path appropriate for Windows.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: See above, I've added a docstring.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,10 +203,22 @@ def _defaultPrivateKeySubtype(keyType):
...
-        options["filename"] = input("Enter file in which the key is (%s): " % filename)
+        if platform.system() == "Windows":
+            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
+        options["filename"] = (
+            input("Enter file in which the key is (%s): " % filename) or filename
+        )
+
+
+def printFingerprint(options):
+    _getKeyOrDefault(options)
```
exarkun: I suggest that `_getKeyOrDefault` avoids mutating `options` and instead returns the value for `filename`.  Then `printFingerprint` can just use that return value directly rather than fishing around inside the `options` object.

Pure functions are better than functions with side-effects and immutable state is better than mutable state.

Twisted has a _lot_ of mutable state all over so I don't think the current `_getKeyOrDefault` behavior is particularly out of place - but almost all of those places in Twisted that work this way are mistakes that, ideally, would gradually be fixed over time.  We can just make this new API avoid this mistake in the first place. :)

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,59 @@ def test_changePassphraseSubtypeV1(self):
...
+        L{options} will default to "~/.ssh/id_rsa" if the user doesn't
+        specify a key.
+        """
+
+        def mock_input(*args):
+            return ""
+
+        import builtins
+
+        self.patch(builtins, "input", mock_input)
```
exarkun: It's better to make `input` an argument to `_getKeyOrDefault`.  Monkey-patching, especially builtins, is handy for dealing with APIs that don't accept all of their inputs as parameters - but since this API is under our control, we can just make it accept all of its inputs as parameters.

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,59 @@ def test_changePassphraseSubtypeV1(self):
...
+
+    def test_displayPublicKeyHandleFileNotFound(self):
+        """
+        Ensure FileNotFoundError is handled, whether the user has supplied
+        a bad path, or has no key at the default path.
+        """
+        options = {"filename": "/foo/bar"}
+        with self.assertRaises(SystemExit) as e:
+            displayPublicKey(options)
+            self.assertTrue(
```
exarkun: `assertIn` produces better failure behavior for this case

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,59 @@ def test_changePassphraseSubtypeV1(self):
...
+                "could not be opened, please specify a file." in e,
+            )
```
adiroiban: Is the assertion called ?

I don't know why the code coverage is not complaining here.

My suggestion is not to use the context here, and instead make use of the nice feature provided by trial :)

```suggestion
        exception = self.assertRaises(SystemExit, displayPublicKey, options)
        self.assertIn("could not be opened, please specify a file.", e)
```

src/twisted/conch/scripts/ckeygen.py
adiroiban: I think I have also commented about returning a value, instead of mutating the input arguments.

so I guess that this is required for approving this PR

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,10 +203,22 @@ def _defaultPrivateKeySubtype(keyType):
         return "PEM"
 
 
-def printFingerprint(options):
+def _getKeyOrDefault(options):
+    """
+    If C{options["filename"]} is None, prompt the user to enter a path
+    or attempt to set it to .ssh/id_rsa
+    """
```
adiroiban: -----------

And as commented by JP, this function should also allow dependency injection, so that we don't need to path the input, but pass our own custom input method during the tests.
```suggestion
def _getKeyOrDefault(options, input_collector=None):
    """
    If C{options["filename"]} is None, prompt the user to enter a path
    or attempt to set it to .ssh/id_rsa
    """
    if input_collector is None:
        input_collector = input
```

-------

But also it should return the path, instead of mutating the input argument.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,10 +203,22 @@ def _defaultPrivateKeySubtype(keyType):
...
     if not options["filename"]:
         filename = os.path.expanduser("~/.ssh/id_rsa")
-        options["filename"] = input("Enter file in which the key is (%s): " % filename)
+        if platform.system() == "Windows":
+            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
+        options["filename"] = (
+            input("Enter file in which the key is (%s): " % filename) or filename
```
adiroiban:
```suggestion
            input_collector("Enter file in which the key is (%s): " % filename) or filename
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +629,59 @@ def test_changePassphraseSubtypeV1(self):
...
+    def test_printFingerprintHandleFileNotFound(self):
+        """
+        Ensure FileNotFoundError is handled for an invalid filename.
+        """
+        options = {"filename": "/foo/bar", "format": "md5-hex"}
+        with self.assertRaises(SystemExit) as e:
+            printFingerprint(options)
+            self.assertTrue(
+                "could not be opened, please specify a file." in e,
+            )
```
adiroiban: There is a lot of repetition in these tests... but this is just a minor comment.

adiroiban (dismissed):
Thanks for the update.

So I think the main things that need to change

* Don't mutate the input argument
* Allow dependency injection
* Fix the test to have the assertion called

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated so that _getKeyOrDefault doesn't mutate options.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: The problem is that prompting the user before calling _getKeyOrDefault means we don't have a suggested/default path to include in the prompt. We can't just use a string, because that will only work for Windows or Linux, not both.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: The problem is that prompting the user before calling _getKeyOrDefault means we don't have a suggested/default path to include in the prompt. We can't just use a string, because that will only work for Windows or Linux, not both.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: It's not possible to inspect the SystemExit outside of a context manager.  The first line succeeds, but the second is stopped by a TypeError.  
`[ERROR]
  File "/home/eevel/twisted/src/twisted/conch/test/test_ckeygen.py", line 662, in test_displayPublicKeyHandleFileNotFound
    self.assertIn("could not be opened, please specify a file.", e)
  File "/home/eevel/twisted/src/twisted/trial/_synctest.py", line 505, in assertIn
    if containee not in container:
builtins.TypeError: argument of type 'SystemExit' is not iterable
`
Based on stdlib unittest, this test should only succeed if the first line raises a SystemExit, and the second fulfills the assertTrue assertion.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: Agreed, updated.

src/twisted/conch/scripts/ckeygen.py
adiroiban: I don't understand the issue.

Here is the complete implementation, I hope it helps describe how I see this method implemented.

```python
def _getKeyOrDefault(options, input_callback=input):
    """
    If C{options["filename"]} is None, prompt the user to enter a path
    or attempt to set it to .ssh/id_rsa.

    @param input_collector: A callback used to collect user input.
      Default to builting L{input} method.
    """
    filename = options["filename"]
    if not filename:
        filename = os.path.expanduser("~/.ssh/id_rsa")
        if platform.system() == "Windows":
            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
        filename = input_callback("Enter file in which the key is (%s): " % filename) or filename
    return 
```

src/twisted/conch/test/test_ckeygen.py
adiroiban: one very raw suggestion , just an idea of how this can be used.


```python
    def test_useDefaultForKey(self):
        """
        L{options} will default to "~/.ssh/id_rsa" if the user doesn't
        specify a key using the CLI argument nor via the interactive text input.
        """
        input_prompts = []

        def rigged_input(prompt):
            """
            An input that records the for which the user alwasys insert
            an empty text.
            """
            input_prompts.append()
            return ""

        options = {"filename": ""}
        filename = _getKeyOrDefault(options, rigged_input)

        # Resolved path is an RSA key inside .ssh dir.
        self.assertTrue(filename.endswith(os.path.join(".ssh", "id_rsa")))
        # The user is prompted once to enter the path, since no path was
        # provided via CLI.
        self.assertEqual(1, len(input_prompts))
        self.assertTrue(input_prompts.startswith("Enter file in which the key is ("))

```

adiroiban:
Thanks for the updates. I left some comments.

Once this is ready for another review, please let me know

src/twisted/conch/test/test_ckeygen.py
adiroiban: maybe

```
self.assertIn("could not be opened, please specify a file.", e.args[0])
```

exarkun: Just to be clear - the reason it "works" inside the `assertRaises` is that it never runs.  The `displayPublicKey` call raises an exception and execution jumps to the `assertRaises` context manager exit handler.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: Updated to add dependency injection.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: Updated.

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +643,58 @@ def test_changePassphraseSubtypeV1(self):
...
+        self.assertIn("could not be opened, please specify a file.", exc.args[0])
+
+    def test_changePassPhraseHandleFileNotFound(self):
+        """
+        Ensure FileNotFoundError is handled for an invalid filename.
+        """
+        options = {"filename": "/foo/bar"}
+        with self.assertRaises(SystemExit) as e:
+            changePassPhrase(options)
+            self.assertIn(
```
adiroiban: Thanks for the updates and followp.


Just a remider. 
I see these tests calling assertion *after* an exception is raised.
In this case, the code after is never called... so the assertion is never raised.

I think this is verry bad habit 

My suggestion is to try TDD.

Write the test first without updating the implementation.
See that the test fails.
Update your implementation.
See the test pass.

If the test pass from the first run, is harder to tell if the test really does its job.

Cheers

eevelweezel: Updated to remove the context managers.

eevelweezel:
This should address the requested changes, I see that static-checks are also failing in trunk.

eevelweezel:
please review

adiroiban:
I don't think I will have time soon to check this PR... I expect that it looks good and it only needs a quick check.

You might want to do a review exchange with another Twisted  dev.
Review a PR from another dev and then you should get a review back :)

Good luck

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,45 +203,64 @@ def _defaultPrivateKeySubtype(keyType):
         return "PEM"
 
 
-def printFingerprint(options):
-    if not options["filename"]:
+def _getKeyOrDefault(options, input_collector=None):
```
glyph: Can you put some type annotations on here?  New code should have them, but since this is an old file it gets (incorrectly) grandfathered in.

glyph: Another nitpick here: technically according to the Twisted coding standard this should be `inputCollector`.

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,45 +203,64 @@ def _defaultPrivateKeySubtype(keyType):
...
+    """
+    If C{options["filename"]} is None, prompt the user to enter a path
+    or attempt to set it to .ssh/id_rsa
+    """
+    if input_collector is None:
+        input_collector = input
+    filename = options["filename"]
+    if not filename:
```
glyph: Suggestion to implement these comments (1/2):

```suggestion
from typing import Dict, Optional, Callable
def _getKeyOrDefault(options: Dict[str, Optional[str]], inputCollector: Optional[Callable[[str],str]]=None) -> str:
    """
    If C{options["filename"]} is None, prompt the user to enter a path
    or attempt to set it to .ssh/id_rsa
    """
    if inputCollector is None:
        inputCollector = input
    filename = options["filename"]
    if not filename:
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,45 +203,64 @@ def _defaultPrivateKeySubtype(keyType):
...
-        options["filename"] = input("Enter file in which the key is (%s): " % filename)
-    if os.path.exists(options["filename"] + ".pub"):
-        options["filename"] += ".pub"
+        if platform.system() == "Windows":
+            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
+        filename = (
+            input_collector("Enter file in which the key is (%s): " % filename)
+            or filename
+        )
+    return filename
```
glyph: Second half of the suggestion since github has some weird requirements on suggestion boundaries:

```suggestion
        if platform.system() == "Windows":
            filename = os.path.expandvars(R"%HOMEPATH %\.ssh\id_rsa")
        filename = (
            inputCollector("Enter file in which the key is (%s): " % filename)
            or filename
        )
    return filename
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -297,7 +317,7 @@ def _inputSaveFile(prompt: str) -> str:
     return input(prompt)
 
 
-def _saveKey(key, options):
+def _saveKey(key, options, input_collector=None):
```
glyph: Should be `inputCollector` here as well, also some type annotations

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -172,6 +174,19 @@ def test_printFingerprintBadFingerPrintFormat(self):
             "Unsupported fingerprint format: sha-base64", em.exception.args[0]
         )
 
+    def test_printFingerprintSuffixAppended(self):
```
glyph: Let's not accidentally turn off type-checking in the tests

```suggestion
    def test_printFingerprintSuffixAppended(self) -> None:
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -356,6 +371,24 @@ def test_saveKeyNoFilename(self):
         persistedKey = key.fromString(persistedKeyContent, None, b"")
         self.assertEqual(key, persistedKey)
 
+    def test_saveKeyFileExists(self):
```
glyph:
```suggestion
    def test_saveKeyFileExists(self) -> None:
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +661,52 @@ def test_changePassphraseSubtypeV1(self):
         self.assertTrue(
             privateKeyContent.startswith(b"-----BEGIN OPENSSH PRIVATE KEY-----\n")
         )
+
+    def test_useDefaultForKey(self):
```
glyph:
```suggestion
    def test_useDefaultForKey(self) -> None:
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +661,52 @@ def test_changePassphraseSubtypeV1(self):
...
+            "",
+        )
+        # Resolved path is an RSA key inside .ssh dir.
+        self.assertTrue(filename.endswith(os.path.join(".ssh", "id_rsa")))
+        # The user is prompted once to enter the path, since no path was
+        # provided via CLI.
+        self.assertEqual(1, len(input_prompts))
+        self.assertEqual([""], input_prompts)
+
+    def test_displayPublicKeyHandleFileNotFound(self):
```
glyph:
```suggestion
    def test_displayPublicKeyHandleFileNotFound(self) -> None:
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +661,52 @@ def test_changePassphraseSubtypeV1(self):
...
+        """
+        Ensure FileNotFoundError is handled, whether the user has supplied
+        a bad path, or has no key at the default path.
+        """
+        options = {"filename": "/foo/bar"}
+        exc = self.assertRaises(SystemExit, displayPublicKey, options)
+        self.assertIn("could not be opened, please specify a file.", exc.args[0])
+
+    def test_changePassPhraseHandleFileNotFound(self):
```
glyph:
```suggestion
    def test_changePassPhraseHandleFileNotFound(self) -> None:
```

src/twisted/conch/test/test_ckeygen.py
```diff
@@ -628,3 +661,52 @@ def test_changePassphraseSubtypeV1(self):
...
+        """
+        Ensure FileNotFoundError is handled for an invalid filename.
+        """
+        options = {"filename": "/foo/bar"}
+        exc = self.assertRaises(SystemExit, changePassPhrase, options)
+        self.assertIn("could not be opened, please specify a file.", exc.args[0])
+
+    def test_printFingerprintHandleFileNotFound(self):
```
glyph:
```suggestion
    def test_printFingerprintHandleFileNotFound(self) -> None:
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -306,20 +326,29 @@ def _saveKey(key, options):
...
 
-        options["filename"] = newPath.strip() or defaultPath
+        filename = newPath.strip() or defaultPath
 
-    if os.path.exists(options["filename"]):
-        print("{} already exists.".format(options["filename"]))
-        yn = input("Overwrite (y/n)? ")
+    if os.path.exists(filename):
+        print(f"{filename} already exists.")
+        yn = input_collector("Overwrite (y/n)? ")
```
glyph:
```suggestion
        yn = inputCollector("Overwrite (y/n)? ")
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -306,20 +326,29 @@ def _saveKey(key, options):
...
+    if input_collector is None:
+        input_collector = input
     KeyTypeMapping = {"EC": "ecdsa", "Ed25519": "ed25519", "RSA": "rsa", "DSA": "dsa"}
     keyTypeName = KeyTypeMapping[key.type()]
-    if not options["filename"]:
+    filename = options["filename"]
+    if not filename:
         defaultPath = os.path.expanduser(f"~/.ssh/id_{keyTypeName}")
+        if platform.system() == "Windows":
+            defaultPath = os.path.expanduser(fR"%HOMEPATH %\.ssh\id_{keyTypeName}")
```
glyph: It looks like this logic is now expressed twice, both here and in `_getKeyOrDefault`.  Would you mind refactoring it into one place so it's a bit easier to see, rather than hard-coding `_getKeyOrDefault` to RSA keys only?

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -306,20 +326,29 @@ def _saveKey(key, options):
 
     @param options:
     @type options: L{dict}
+
+    @param input_collector: Dependency injection for testing.
+    @type input_collector: L{function} or None.
```
glyph:
```suggestion
    @param inputCollector: Dependency injection for testing.
    @type inputCollector: L{function} or None.
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -339,23 +368,23 @@ def _saveKey(key, options):
 
     comment = f"{getpass.getuser()}@{socket.gethostname()}"
 
-    filepath.FilePath(options["filename"]).setContent(
+    filepath.FilePath(filename).setContent(
```
glyph: Huh. Not your bug, but this is such an odd choice to not use the path object here…

```suggestion
    fp = filepath.FilePath(filename)
    fp.setContent(
```

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -339,23 +368,23 @@ def _saveKey(key, options):
...
         key.toString(
             "openssh",
             subtype=options["private-key-subtype"],
             passphrase=options["pass"],
         )
     )
-    os.chmod(options["filename"], 33152)
+    os.chmod(filename, 33152)
```
glyph: Also this magic number is terrible :)

```suggestion
    fp.chmod(0o100600)
```

glyph: Also we should file a bug here, there's arguably a umask race in this code, we should not be chmod'ing things after the fact

glyph (changes requested):
This looks like a great change and I'm sorry for nit-picking it to death here, but it would be great to give the type-checker a chance to spot a few more bugs before we land it. Would you mind making a few changes? I tried to put in as many of them as suggestions as I could.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated to add typing.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Added typing

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Changed to camelCase.

src/twisted/conch/test/test_ckeygen.py
eevelweezel: Added return types for tests.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated to camelCase.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Refactored and updated tests, "rsa" is now the default param if no key type is supplied.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Definitely makes sense, updated.

src/twisted/conch/scripts/ckeygen.py
eevelweezel: Updated

src/twisted/conch/scripts/ckeygen.py
```diff
@@ -202,45 +205,74 @@ def _defaultPrivateKeySubtype(keyType):
...
+def _getKeyOrDefault(
+    options: Dict[Any, Any],
+    inputCollector: Optional[Callable] = None,
+    keyTypeName: str = "rsa",
+) -> str:
+    """
+    If C{options["filename"]} is None, prompt the user to enter a path
+    or attempt to set it to .ssh/id_rsa
+    @param options: command line options
+    @type options: L{Dict}
```
adiroiban: no need to add `@type` here, as long as we have type hinting.

and I think is best not to have `@type` in the docstring to avoid future conflicts.

eevelweezel: Updated.

eevelweezel:
please review

adiroiban:
I will try to get this reviewed next week, in exchange for your help with the other PRs :)

But if anyone else wants to review it, I am happy to have this merge with a review from another person.

glyph (approved):

glyph:
Looks like all the feedback was addressed!

PR #11658, opened 2022-09-12
#11580 Handle trial logfile with directory separator

### LINKED ISSUE #11580: `trial -j` silently hangs if a logfile with a directory separator in its name is passed
Make a test file like this:

```python3

class NothingTest(TestCase):
    def test_nothing(self) -> None:
        "Do nothing."
```

Then,

```
$ mkdir hello
$ trial -j 1 -l hello/test.log ./test_nothing.py
Running 1 tests.
^CTraceback (most recent call last):
  File "/Users/glyph/.virtualenvs/Twisted/bin/trial", line 33, in <module>
    sys.exit(load_entry_point('Twisted', 'console_scripts', 'trial')())
  File "/Users/glyph/Projects/Twisted/src/twisted/scripts/trial.py", line 654, in run
    test_result = trialRunner.run(suite)
  File "/Users/glyph/Projects/Twisted/src/twisted/trial/_dist/disttrial.py", line 453, in run
    result.raiseException()
  File "/Users/glyph/Projects/Twisted/src/twisted/python/failure.py", line 503, in raiseException
twisted.python.filepath.InsecurePath: 'hello/test.log' contains one or more directory separators
Unhandled error in Deferred:

  File "/Users/glyph/Projects/Twisted/src/twisted/scripts/trial.py", line 654, in run
    test_result = trialRunner.run(suite)
  File "/Users/glyph/Projects/Twisted/src/twisted/trial/_dist/disttrial.py", line 449, in run
    d.addBoth(lambda ignored: self._reactor.stop())
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/defer.py", line 572, in addBoth
    return self.addCallbacks(
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/defer.py", line 511, in addCallbacks
    self._runCallbacks()
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/defer.py", line 891, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/Users/glyph/Projects/Twisted/src/twisted/trial/_dist/disttrial.py", line 449, in <lambda>
    d.addBoth(lambda ignored: self._reactor.stop())
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/base.py", line 708, in stop
    raise error.ReactorNotRunning("Can't stop reactor that isn't running.")
twisted.internet.error.ReactorNotRunning: Can't stop reactor that isn't running.
$ 
```

You'll see the `^C` there, because `trial` just hangs forever.

### LINKED ISSUE #11657: twisted.trial.unittest.SynchronousTestCase.successResultOf is mis-annotated
`successResultOf` is annotated as `Callable[[Deferred[T]], T]`.  However, the implementation uses `ensureDeferred` on the argument so at runtime a coroutine will also work fine.  Only the type checker is unhappy because the annotation on `successResultOf` says it can only work on Deferreds.

The only conceivable reason to use `ensureDeferred` in the implementation is to support other kinds of Deferred-like things, such as coroutines, so the annotation should reflect this.

exarkun:
## Scope and purpose

Fixes #11580
Fixes #11657

### DIFF
--- a/src/twisted/trial/_dist/test/test_disttrial.py
+++ b/src/twisted/trial/_dist/test/test_disttrial.py
@@ -13,8 +13,19 @@ from typing import Callable, List, Set
 
 from zope.interface import implementer, verify
 
-from attrs import Factory, define, field
-from hamcrest import assert_that, contains, equal_to, has_length, none
+from attrs import Factory, assoc, define, field
+from hamcrest import (
+    assert_that,
+    contains,
+    ends_with,
+    equal_to,
+    has_length,
+    none,
+    starts_with,
+)
+from hamcrest.core.core.allof import AllOf
+from hypothesis import given
+from hypothesis.strategies import booleans, sampled_from
 
 from twisted.internet import interfaces
 from twisted.internet.base import ReactorBase
@@ -303,6 +314,40 @@ class WorkerPoolTests(TestCase):
         assert_that(self.successResultOf(joining), none())
         assert_that(started.testLog.closed, equal_to(True))
         assert_that(started.testDirLock.locked, equal_to(False))
+
+    @given(
+        booleans(),
+        sampled_from(
+            [
+                "out.log",
+                "subdir/out.log",
+            ]
+        ),
+    )
+    def test_logFile(self, absolute: bool, logFile: str) -> None:
+        """
+        L{WorkerPool.start} creates a L{StartedWorkerPool} configured with a
+        log file based on the L{WorkerPoolConfig.logFile}.
+        """
+        if absolute:
+            logFile = self.parent.path + "/" + logFile
+
+        config = assoc(self.config, logFile=logFile)
+
+        if absolute:
+            matches = equal_to(logFile)
+        else:
+            matches = AllOf(
+                # This might have a suffix if the configured workingDirectory
+                # was found to be in-use already so we don't add a "/" suffix.
+                starts_with(config.workingDirectory.path),
+                # This should be exactly the suffix so we add a "/" prefix.
+                ends_with("/" + logFile),
+            )
+
+        pool = WorkerPool(config)
+        started = self.successResultOf(pool.start(CountingReactor([])))
+        assert_that(started.testLog.name, matches)
 
 
 class DistTrialRunnerTests(TestCase):
--- a/src/twisted/trial/_dist/test/test_disttrial.py
+++ b/src/twisted/trial/_dist/test/test_disttrial.py
@@ -9,6 +9,7 @@ import os
 import sys
 from functools import partial
 from io import StringIO
+from os.path import sep
 from typing import Callable, List, Set
 
 from zope.interface import implementer, verify
@@ -320,7 +321,7 @@ class WorkerPoolTests(TestCase):
         sampled_from(
             [
                 "out.log",
-                "subdir/out.log",
+                f"subdir{sep}out.log",
             ]
         ),
     )
@@ -330,7 +331,7 @@ class WorkerPoolTests(TestCase):
         log file based on the L{WorkerPoolConfig.logFile}.
         """
         if absolute:
-            logFile = self.parent.path + "/" + logFile
+            logFile = self.parent.path + sep + logFile
 
         config = assoc(self.config, logFile=logFile)
 
@@ -339,10 +340,10 @@ class WorkerPoolTests(TestCase):
         else:
             matches = AllOf(
                 # This might have a suffix if the configured workingDirectory
-                # was found to be in-use already so we don't add a "/" suffix.
+                # was found to be in-use already so we don't add a sep suffix.
                 starts_with(config.workingDirectory.path),
-                # This should be exactly the suffix so we add a "/" prefix.
-                ends_with("/" + logFile),
+                # This should be exactly the suffix so we add a sep prefix.
+                ends_with(sep + logFile),
             )
 
         pool = WorkerPool(config)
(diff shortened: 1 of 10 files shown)


src/twisted/trial/test/matchers.py
```diff
@@ -21,17 +23,46 @@ class _MatchAfter(BaseMatcher[_A]):
...
+            will only work as long as PyHamcrest calls methods in the right
+            order.  The PyHamcrest Matcher interface doesn't seem to allow
+            implementing this functionality in a more reliable way (see the
+            implementation of L{assert_that}).
+        """
+        if self._e is None:
+            super().describe_mismatch(item, mismatch_description)
+        else:
+            mismatch_description.append_text(
+                f"{self.f.__qualname__}({item!r}) raised\n    {self._e}"
```
twm: This should include the exception name, not just the message.

src/twisted/trial/test/test_matchers.py
```diff
@@ -0,0 +1,82 @@
...
+        p = FilePath(self.mktemp())
+
+        description = StringDescription()
+        assert_that(fileContents(anything()).matches(p, description), equal_to(False))
+        assert_that(
+            str(description),
+            AllOf(
+                contains_string("fileContents"),
+                contains_string("No such file or directory"),
+                contains_string(p.path),
```
twm: It looks like you're getting the repr of this in the Windows failure.

src/twisted/trial/_dist/test/test_disttrial.py
```diff
@@ -304,6 +315,40 @@ def test_join(self):
...
+
+        if absolute:
+            matches = equal_to(logFile)
+        else:
+            matches = AllOf(
+                # This might have a suffix if the configured workingDirectory
+                # was found to be in-use already so we don't add a "/" suffix.
+                starts_with(config.workingDirectory.path),
+                # This should be exactly the suffix so we add a "/" prefix.
+                ends_with("/" + logFile),
```
twm: Looks like you need `os.sep` here for Windows.

src/twisted/trial/test/test_script.py
```diff
@@ -33,6 +41,99 @@ def sibpath(filename):
...
+    Run L{logSomething} as a test method using the given configuration.
+    """
+    runner = trial._makeRunner(config)
+    runner.stream = StringIO()
+    suite = TestSuite([pyunit.FunctionTestCase(logSomething)])
+    runner.run(suite)
+    return config
+
+
+runFromArguments = compose(runFromConfig, parseArguments)
```
twm, line 71: _Technically_ I suppose this is a function definition that needs a docstring.

twm (approved):
This looks like a solid fix, thank you!

Hamcrest makes the PR more verbose but the logic is quite easy to review. I'm curious how it impacts runtime.

I believe that the remaining failures will be trivial to address, so feel free to land once green!

exarkun:
> Hamcrest makes the PR more verbose but the logic is quite easy to review. I'm curious how it impacts runtime.

Regarding runtime - I think it's probably negligible?  I'm sure it's slower than `assertEqual`, maybe even measurably slow, but if it were even an extra millisecond on recent hardware I'd be quite shocked.

Hypothesis is much more likely to have a noticable impact on runtime.  That's part of the reason I opted for `sampled_from` with a short list here.  This substantially constrains Hypothesis' example search and limits the number of test iterations to a small handful (eg 4 for `test_logFile`).  This gives up a bit of Hypothesis' power by not letting it search out exciting values from the full input space and I'm not _sure_ it's the right balance to strike but it sticks closer to Twisted's historical testing patterns than the alternative so it seems like a good choice for now (in these early days of Hypothesis-in-Twisted).

exarkun:
> Hamcrest makes the PR more verbose but the logic is quite easy to review. I'm curious how it impacts runtime.

Regarding runtime - I think it's probably negligible?  I'm sure it's slower than `assertEqual`, maybe even measurably slow, but if it were even an extra millisecond on recent hardware I'd be quite shocked.

Hypothesis is much more likely to have a noticable impact on runtime.  That's part of the reason I opted for `sampled_from` with a short list here.  This substantially constrains Hypothesis' example search and limits the number of test iterations to a small handful (eg 4 for `test_logFile`).  This gives up a bit of Hypothesis' power by not letting it search out exciting values from the full input space and I'm not _sure_ it's the right balance to strike but it sticks closer to Twisted's historical testing patterns than the alternative so it seems like a good choice for now (in these early days of Hypothesis-in-Twisted).

src/twisted/trial/test/test_matchers.py
exarkun: Oops.  I guess it's the repr on all platforms but it only makes a difference on Windows because of the platform sep there.  Took me a while. :/

src/twisted/trial/test/matchers.py
exarkun: Done.  I ended up taking the message _out_ of the assertion made by the test (and only using the exception class name) because the message could vary based on locale.

src/twisted/trial/test/test_script.py
exarkun: is there a meaningful way to give it a docstring?  I've seen people write things like:

```
x = ...
"""
Blah blah blah
"""
```

I know that Python itself doesn't do anything constructive with this but maybe one of the doc tools Twisted is using does?

Alternatively, maybe `compose` should synthesize a docstring based on its arguments? :slightly_smiling_face:

exarkun: I filed https://github.com/twisted/twisted/issues/11659

PR #11662, opened 2022-09-13
#11661 Add an AMP stream helper for disttrial

### LINKED ISSUE #11661: disttrial needs some streaming helpers
The fix for #10314 will involve streaming arbitrarily long strings over an AMP connection.  There is no functionality in Twisted yet to support this.  Add some for trial to use.

exarkun:
## Scope and purpose

Fixes #11661

I split this out of the branch for #10314 because it's self-contained and I figured it would be easier to get reviewed this way.

### DIFF
--- a/src/twisted/trial/_dist/stream.py
+++ b/src/twisted/trial/_dist/stream.py
@@ -56,9 +56,10 @@ class StreamReceiver:
         """
         self._streams[streamId].append(chunk)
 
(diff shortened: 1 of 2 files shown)


exarkun:
please review

adiroiban (approved):
Thanks. Looks good and I think it should be merged.

PR #11665, opened 2022-09-14
#11664 Initial run on macOS on GHA.

### LINKED ISSUE #11664: Run macOS CI on GHA
To have all the CI in the same place.

It should make it easier to gather coverage reports and produce an independent coverage report.

It should also be easier to debug using tmate action.

adiroiban:
## Scope and purpose

Fixes #11664 

Run macOS tests on GHA.

Remove Azure pipelines.

Azure pipelines works, but is kind of stucked. The admin account is owned by Amber and no other dev has access to the Azure Devops account.

With GitHub actions, the  whole dev team has access to the CI.

## What to check

* There is a CI job running on macOS hardware
* The macOS job uses the select reactor


.github/workflows/test.yaml
```diff
@@ -122,6 +122,11 @@ jobs:
           # Just Python 3.9 with default settings.
           - python-version: '3.9'
 
+          # newest macOS and newest Python supported version.
+          - python-version: '3.10'
+            runs-on: 'macos-12'
+            job-name: 'macos-default-tests'
```
exarkun: I think it would be good to keep testing an older version of Python on macOS.  On Azure it looks like we tested 3.7 - 3.10.  Perhaps we can at least test 3.7 _and_ 3.10 (skipping the intermediates) on GA.

exarkun (approved):
Thanks.  Looks good to me.  I left one suggestion inline about the Python versions to test against.

adiroiban:
Thanks for the review.
I have added another macOS job for older python and older macos.

I will enable auto-merge for this.

And with this landed, we should be able to generate our independent coverage report.... I will test with towncrier first.

PR #11669, opened 2022-09-16
Debugging twisted.trial._dist.test.test_stream failure on macOS

exarkun:
Not for merge.

I just need GitHub Actions to run with some extra instrumentation on macOS.


exarkun:
done with this

Issue #11671, opened 2022-09-16
Add support for selecting a Hypothesis profile to trial


exarkun:
Hypothesis has a notion of "profiles".  These are named bags of global state which control certain behavior of Hypothesis process-wide.  For example, a profile specifies whether all tests Hypothesis runs must complete in less than some fixed wallclock interval.

Hypothesis profiles may be used by third-party applications in their test suite which is run by trial or they may be used by Twisted itself (#11649).  Either way, it's useful to be able to define a few of these and then choose between them on a run-by-run basis.  For example, you might have a profile that's suitable for developers to use locally, another that's suitable for testing PRs on CI, and a third that's oriented towards expensive searches for previously unknown bugs which is triggered by a separate CI configuration or run in some other dedicated environment.

As far as I know, state of the art in runtime selection of Hypothesis profiles when using trial is to add some code to your test package's `__init__.py` that inspects an environment variable and chooses a profile a profile based on its value, then set that environment variable whenever you run trial.

This is a bit tricky to get right.  First, you have to be aware of the approach at all.  Next, you need to be careful not to allow multiple projects that follow this convention to trample on each other (for example, if two projects use the same environment variable name -- say, `HYPOTHESIS_PROFILE` -- then it becomes ambiguous which of them is being configured and the outcome may be surprising).

Instead of this, `trial` could accept a command line argument which causes trial itself to load a Hypothesis profile.  Applications are still responsible for defining those profiles but it becomes trial's job to translate user intent (from the command line) into action (loading the profile).

adiroiban:
Just a note from here https://github.com/twisted/twisted/pull/11672#discussion_r977950360

If implementing proper support is too much work, and we want this behaviour, we can fallback to ENV vars usage to implement this.

PR #11672, opened 2022-09-16
#11649 Add a deterministic Hypothesis profile

### LINKED ISSUE #11649: Configure Hypothesis for deterministic runs on CI
In the fix for #10314 we introduced a new dependency on Hypothesis.  In [discussion on the mailing list](https://mail.python.org/archives/list/twisted@python.org/thread/TK6TXW3SWYJAFBVFMXJ5G7KTODJEEZI2/) there was some brief discussion of certain Hypothesis configuration options that are desirable for deterministic CI operation.

Quoting:

>  - Turn off Hypothesis' deadline feature, probably everywhere
>   - At least for normal CI, and maybe for development environments, turn
>   off Hypothesis' randomization feature.
>   Perhaps set up a scheduled CI job that runs with randomization enabled.

So, do these things.

exarkun:
## Scope and purpose

Fixes #11649

This is kind of a hack.  It's the hack I've used in other projects with trial-based test suites.  Sometime soon I would like to obsolete this hack with #11671 though.

### DIFF
--- a/src/twisted/trial/test/__init__.py
+++ b/src/twisted/trial/test/__init__.py
@@ -4,3 +4,39 @@
 """
 Unit tests for the Trial unit-testing framework.
 """
+
+from hypothesis import HealthCheck, settings
+
+
+def _activateHypothesisProfile():
+    """
+    Load a Hypothesis profile appropriate for a Twisted test suite.
+    """
+    deterministic = settings(
+        # Disable the deadline.  It is too hard to guarantee that a particular
+        # piece of Python code will always run in less than some fixed amount
+        # of time.  Hardware capabilities, the OS scheduler, the Python
+        # garbage collector, and other factors all combine to make substantial
+        # outliers possible.  Such failures are a distraction from development
+        # and a hassle on continuous integration environments.
+        deadline=None,
+        suppress_health_check=[
+            # With the same reasoning as above, disable the Hypothesis time
+            # limit on data generation by example search strategies.
+            HealthCheck.too_slow,
+        ],
+        # When a developer is working on one set of changes, or continuous
+        # integration system is testing them, it is disruptive for Hypothesis
+        # to discover a bug in pre-existing code.  This is just what
+        # Hypothesis will do by default, by exploring a pseudo-randomly
+        # different set of examples each time.  Such failures are a
+        # distraction from development and a hassle in continuous integration
+        # environments.
+        derandomize=True,
+    )
+
+    settings.register_profile("twisted_trial_test_profile_deterministic", deterministic)
+    settings.load_profile("twisted_trial_test_profile_deterministic")
+
+
+_activateHypothesisProfile()


exarkun:
please review

src/twisted/trial/test/__init__.py
```diff
@@ -4,3 +4,39 @@
...
+            HealthCheck.too_slow,
+        ],
+        # When a developer is working on one set of changes, or continuous
+        # integration system is testing them, it is disruptive for Hypothesis
+        # to discover a bug in pre-existing code.  This is just what
+        # Hypothesis will do by default, by exploring a pseudo-randomly
+        # different set of examples each time.  Such failures are a
+        # distraction from development and a hassle in continuous integration
+        # environments.
+        derandomize=True,
```
adiroiban, line 35: I guess that later we can optionally enable random data injected via env vars?

In this way, we can enable a nightly trunk CI run with random runs.

But right now, I don't think that we have the time to look after nightly CI test results

src/twisted/trial/test/__init__.py
```diff
@@ -4,3 +4,39 @@
...
+        # piece of Python code will always run in less than some fixed amount
+        # of time.  Hardware capabilities, the OS scheduler, the Python
+        # garbage collector, and other factors all combine to make substantial
+        # outliers possible.  Such failures are a distraction from development
+        # and a hassle on continuous integration environments.
+        deadline=None,
+        suppress_health_check=[
+            # With the same reasoning as above, disable the Hypothesis time
+            # limit on data generation by example search strategies.
+            HealthCheck.too_slow,
```
adiroiban, line 26: I guess that this is ok for now.

My understanding is that we will not implement any custom data generation.

But, maybe instead of unlimited time, we can add something like 1 minute.

I guess the limit is big enough for a test input.

The idea is that if something goes wrong on the data generation, you should get a better error message, rather than seeing the whole CI cancelled due to timeout.

adiroiban (approved):
I don't know much about the hypothesis but I think this is ok.

This is what was discussed over email.

Also, Jean-Paul is an active developer and if this change helps him, I think we should have this change.

I don't see any drawback.

At this time, using hypothesis is optional.

Thanks JP for the update.

src/twisted/trial/test/__init__.py
exarkun: Enable random mode somehow, yea.  I filed https://github.com/twisted/twisted/issues/11671 because I think first-class support for profile selection in trial might be a good thing.  In other projects I have just used environment variables so if #11671 doesn't look appealing or is too much work I know we can always fall back to that approach.

I agree that it's not clear right now who would be responsible for _doing_ anything with the nightly random results so it doesn't seem to make sense to run them yet - but hopefully we'll get there.

src/twisted/trial/test/__init__.py
exarkun: I'm still very skeptical of applying any limit.  Basically, our CI providers give us no guarantees whatsoever how much processor time our jobs will get.  Chances are most of the time they'll give us plenty and Hypothesis won't have a problem - but how close to "all of the time" will that be?  I dunno.

Or I guess I could look at it from a different perspective.  _trial_ imposes a 2 minute timeout per Deferred-returning test, by default ... So imposing a 2 minute timeout on Hypothesis stuff is no worse than that.  I guess people probably haven't been complaining lately about the 2 minute trial timeout spuriously engaging on CI?

exarkun: Ah, but there is no way to set a custom amount of time that defines "too slow".  You can either have the "too slow" check or you can turn it off.  The "deadline", which applies to how long the test method itself takes, is configurable, so I could still apply a timeout to that.  However, I don't think this will interrupt a test that takes too long, it will just cause it to fail afterwards.  So for the case of "make CI fail with a better report" I don't think it helps - whether there's a deadline or not, if the test runs for long enough that CI decides to kill it, CI's just gonna kill it and you get no extra info.

So I guess I'll leave this alone for now.  Maybe we'll figure out a better configuration at some future point.

adiroiban: Sure. Sounds right. If we can't get better error reporting, then don't bother here.

I don't remember seeing any CI error due to trial 2 minute limit.

And on my local run, if a test takes more than 10 seconds, I kill it :)

PR #11678, opened 2022-09-21
#11677 Fix test_hiddenException

### LINKED ISSUE #11677: twisted.trial.test.test_reporter.ErrorReportingTests.test_hiddenException is unnecessarily coupled to reactor implementation details
This test asserts a bunch of specific things about the frames in the traceback associated with a logged failure.  The call stack is not part of Twisted's compatibility policy so this is overly specific for any end-user facing behavior.  This test should limit itself to making assertions about the public interface of this functionality.  This makes refactoring the implementation of that public interface easier.

This is specifically motivated by work on #11673 which caused this test to fail because of reactor refactoring.

exarkun:
## Scope and purpose

Fixes #11677

### DIFF
--- a/src/twisted/trial/_dist/workerreporter.py
+++ b/src/twisted/trial/_dist/workerreporter.py
@@ -10,11 +10,11 @@ Test reporter forwarding test results over trial distributed AMP commands.
 """
 
 from types import TracebackType
-from typing import Callable, List, Optional, Sequence, Tuple, Type, TypeVar, Union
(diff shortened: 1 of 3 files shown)


adiroiban (approved):
I guess that this is a test for a deprecated behaviour.

I don't think that is productive to dig deeper into this.

As long as the tests are green, I think this should be merged.

Having it in a separate PR should make it easy to revert if needed.

Thanks.

Issue #5962, opened 2012-09-11
Clock.callLater(0, f) can lead to infinite loop


twisted-trac:
|[<img alt="exarkun's avatar" src="https://avatars.githubusercontent.com/u/254565?s=50" width="50" height="50">](https://github.com/exarkun)| @exarkun reported|
|-|-|
|Trac ID|trac#5962|
|Type|defect|
|Created|2012-09-11 20:52:02Z|
|Branch|https://github.com/twisted/twisted/tree/calllater-infinite-loop-5962|

Consider:

```python
from twisted.internet.task import Clock

def f():
    c.callLater(0, f)

c = Clock()
f()
c.advance(0)
print 'Done'
```

The expected behavior is that `f` runs once and then _Done_ is printed and the program exits.  Instead, `Clock.advance` spends forever running and re-running `f`.  This is similar to a bug that the real `IReactorTime` implementation used to have but that we fixed.

Attachments:

* [clockinfloop.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/149/14925e51a51a2db8fa7f542644cc67360af88436/7a4e253bcfea7baebc10d6eb8a35a34edae83d5c.patch) (2088 bytes) - added by Saurabh on 2013-04-14 10:44:02Z - 
* [clockinfloop_2.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/149/14925e51a51a2db8fa7f542644cc67360af88436/f338f112cf340cc89cf9a39fa1203cd4e7c95ae3.patch) (2818 bytes) - added by Saurabh on 2013-04-16 16:43:00Z - Added topfile, better variable names,comments.
* [callLater-5962-v1.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/149/14925e51a51a2db8fa7f542644cc67360af88436/6791281226e885a77f9e696604c7ad8b9cf692d1.patch) (2619 bytes) - added by mgrazebrook on 2013-12-07 21:34:31Z - 
* [callLater-5962-v2.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/149/14925e51a51a2db8fa7f542644cc67360af88436/5a61d32f2990d6f70d96977bfcc378625217f04b.patch) (3186 bytes) - added by mgrazebrook on 2013-12-07 21:36:35Z - 

[...]

PR #11680, opened 2022-09-21
#11679 LoopingCall vs Clock tests

### LINKED ISSUE #11679: The LoopingCall tests in twisted.test.test_task encode implementation details of Clock
As reported in https://github.com/twisted/twisted/issues/5962, Clock and the real reactor differ in their handling of `callLater(0, ...)`.  Unfortunately the LoopingCall tests in `twisted.test.test_task` encode a dependency on the `Clock` behavior so they only demonstrate that `LoopingCall` has some behavior when used with a broken reactor.

I have a fix for #5962 but I don't want to complicate its PR with a lot of subtle changes to test suites for other units.  So, in preparation for landing that work, let's fix these tests to have the minimum possible dependency on `Clock`'s quirks.

exarkun:
## Scope and purpose

Fixes #11679

From the commit message:

> 
> Change the LoopingCall tests to assert more reasonable behavior                                                                                                                   
> 
> clock.advance(0) will run not only any calls scheduled for clock.seconds() or
> earlier, but it will also run any calls that _those_ calls schedule with
> `clock.callLater(0, ...)`.
> 
> This diverges from the real IReactorTime implementations which force all
> scheduled events to wait at least one reactor iteration before they can run.
> 
> This impacts the LoopingCall test cases for the case of interval == 0 because
> those tests do a lot of scheduling with `clock.callLater(0, ...)`.  In
> particular, the tests often `advance(0)` and expect a `LoopingCall` to execute
> multiple iterations.
> 
> These changes make the tests advance the clock in enough discrete steps that a
> real reactor would have time to run all of the necessary calls.
> 
> After #5962 the old version of these tests will fail.  Also as part of #5962
> one clearly-marked fudge left in these tests will need to be removed because I
> don't see how to make the test work (and remain precise) with both the broken
> Clock behavior and the real reactor behavior.
>

### DIFF
--- a/src/twisted/test/test_task.py
+++ b/src/twisted/test/test_task.py
@@ -506,6 +506,10 @@ class LoopTests(unittest.TestCase):
         loop.clock = clock
         deferred = loop.start(0, now=False)
 
+        # Even though we have a no-delay loop,
+        # a single iteration of the reactor will not trigger the looping call
+        # multiple times.
+        # This is why we explicitly iterate multiple times.
         clock.pump([0] * 5)
         self.successResultOf(deferred)
 
--- a/src/twisted/test/test_task.py
+++ b/src/twisted/test/test_task.py
@@ -491,8 +491,8 @@ class LoopTests(unittest.TestCase):
 
     def test_withCountIntervalZero(self):
         """
-        L{task.LoopingCall.withCount} with interval set to 0
-        will call the countCallable 1.
+        L{task.LoopingCall.withCount} with interval set to 0 calls the
+        countCallable with a count of 1.
         """
         clock = task.Clock()
         accumulator = []
@@ -506,10 +506,10 @@ class LoopTests(unittest.TestCase):
         loop.clock = clock
         deferred = loop.start(0, now=False)
 
-        clock.advance(0)
+        clock.pump([0] * 5)
         self.successResultOf(deferred)
 
-        self.assertEqual([1, 1, 1, 1, 1], accumulator)
+        self.assertEqual([1] * 5, accumulator)
 
     def test_withCountIntervalZeroDelay(self):
         """
@@ -534,20 +534,20 @@ class LoopTests(unittest.TestCase):
         loop.clock = clock
         loop.start(0, now=False)
 
-        clock.advance(0)
+        clock.pump([0] * 2)
         # Loop will block at the third call.
-        self.assertEqual([1, 1], accumulator)
+        self.assertEqual([1] * 2, accumulator)
 
         # Even if more time pass, the loops is not
         # advanced.
-        clock.advance(2)
-        self.assertEqual([1, 1], accumulator)
+        clock.pump([1] * 2)
+        self.assertEqual([1] * 2, accumulator)
 
         # Once the waiting call got a result the loop continues without
         # observing any delay in countCallable.
         deferred.callback(None)
-        clock.advance(0)
-        self.assertEqual([1, 1, 1, 1, 1], accumulator)
+        clock.pump([0] * 4)
+        self.assertEqual([1] * 5, accumulator)
 
     def test_withCountIntervalZeroDelayThenNonZeroInterval(self):
         """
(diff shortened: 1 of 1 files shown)


src/twisted/test/test_task.py
```diff
@@ -506,10 +506,10 @@ def foo(cnt):
         loop.clock = clock
         deferred = loop.start(0, now=False)
 
-        clock.advance(0)
+        clock.pump([0] * 5)
```
adiroiban, line 513: Just to make sure I understood the expected behaviour

```suggestion
        # Even though we have a no-delay loop,
        # a single iteration of the reactor will not trigger the looping call
        # multiple times.
        # This is why we explicitly iterate multiple times.
        clock.pump([0] * 5)
```

adiroiban, line 513: I guess we can also have the test like this

And my understanding is that this is is how the tests will be refactored after the Clock is fixed.

```suggestion
        # Iterating the reactor will only cause the loop to loop once.
        # even if the loop is configured to loop without a delay.
        clock.advance(0)
        
        self.assertEqual([1], accumulator)
```

src/twisted/test/test_task.py
```diff
@@ -534,20 +534,20 @@ def foo(cnt):
...
+        clock.pump([1] * 2)
+        self.assertEqual([1] * 2, accumulator)
 
         # Once the waiting call got a result the loop continues without
         # observing any delay in countCallable.
         deferred.callback(None)
-        clock.advance(0)
-        self.assertEqual([1, 1, 1, 1, 1], accumulator)
+        clock.pump([0] * 4)
+        self.assertEqual([1] * 5, accumulator)
```
adiroiban, line 554: Just asking.

Is there any advantage in this way of writing the test?

I prefer to have the assertions as "dumb" as possible, to reduce the risk of errors...to the extend of having more repetition.

adiroiban (approved):
Changes looks good.

But I am not sure what is the point of this?

I guess that once Clock is fixed, we can have the tests refactored in the same PR.

My understanding is that these tests are wrong.
They work with Clock, but would never work with a real reactor,

Also, once Clock is fixed, we might want to rewrite the tests to explictly define the loop behaviour for `interval=0`

Anyway. If this makes future refactoring easy,  I don't see any harm in merging this.

Thanks again!

Looking forward to the Clock refactoring.

I wasn't aware of this detail, and I think I was bitten / smashehd by this beahviour ...  wondering why the unit test code is different to what I observe with a functional / end to end test.

exarkun:
> Changes looks good.
> 
> But I am not sure what is the point of this?
> 

Thanks for looking.  Sorry I wasn't more clear about the purpose.  This is a somewhat weird PR.

> I guess that once Clock is fixed, we can have the tests refactored in the same PR.
> 

I believe that except for https://github.com/twisted/twisted/pull/11680/files#diff-b3f880003e83657375b9dd7fedfc131fd6208ce3de9b339337853b457e55c305R592 the tests are now correct with respect to both Clock and real reactors.  In an upcoming PR where I fix Clock, I'll just delete that line.  The changes would probably have made more sense as _part_ of the PR that fixes Clock but I didn't expect that to be a small PR so I wanted to try to get as much of it landed separately as possible.  In retrospect, I probably could have done all the hard work of fixing things but left Clock broken, landed that PR, then done the last little bit to apply the fix to Clock and land that change along with all of these test fixes.  Oh well...


> My understanding is that these tests are wrong. They work with Clock, but would never work with a real reactor,

This is correct - that is, in current trunk@HEAD, they are wrong and don't work with a real reactor.  In this branch I think they are now mostly right.

> 
> Also, once Clock is fixed, we might want to rewrite the tests to explictly define the loop behaviour for `interval=0`

Pooossibly.  The tests definitely already define _something_ about that behavior but they're not very systematic about it.  It would probably be good to rewrite them so they cover all of the cases in some logical, coherent fashion.  I'm not sure if I'll commit to taking on that work though. :)

> 
> Anyway. If this makes future refactoring easy, I don't see any harm in merging this.
> 
> Thanks again!

It does, thanks!

> 
> Looking forward to the Clock refactoring.

Coming up soon :)

> 
> I wasn't aware of this detail, and I think I was bitten / smashehd by this beahviour ... wondering why the unit test code is 
different to what I observe with a functional / end to end test.

Indeed!  I had forgotten about it too until these tests hit me in the face...

Thanks again.

src/twisted/test/test_task.py
exarkun: Because of the assertion on line 512 (accumulator == [1] * 5) the clock/reactor will still need to iterate 5 times.  It is the old, broken behavior that makes `clock.advance(0)` run the LoopingCall 5 times.

src/twisted/test/test_task.py
exarkun: I think you're asking about  `[1, 1, 1, 1, 1]` vs `[1] * 5`?  I think I find `[1] * 5` to be the less risky version.  My eyes slide over `[1, 1, 1, 1, 1]` very easily but the exact number of 1s there is absolutely critical to the correctness of the test.

Probably there is some other pattern altogether that is better than either of these, keeping things simpler than list multiplication but also making it easy to read without making a mistake.  I didn't think very hard about finding such a pattern while working on this.

PR #11682, opened 2022-09-22
#11681 procmon vs clock

### LINKED ISSUE #11681: twisted.runner.test.test_procmon.ProcmonTests encode implementation details of Clock
As reported in #5962, Clock and the real reactor differ in their handling of `callLater(0, ...)`. Unfortunately the procmon tests in twisted.runner.test.test_procmon encode a dependency on the `Clock` behavior so they only demonstrate the code under test has some behavior when used with a broken reactor.

I have a fix for #5962 but I don't want to complicate its PR with a lot of subtle changes to test suites for other units. So, in preparation for landing that work, let's fix these tests to have the minimum possible dependency on Clock's quirks.

exarkun:
Fixes #11681

### DIFF
--- a/src/twisted/runner/test/test_procmon.py
+++ b/src/twisted/runner/test/test_procmon.py
@@ -302,6 +302,11 @@ class ProcmonTests(unittest.TestCase):
         # verify that the process restarts
         self.reactor.advance(timeToDie)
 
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Thanks. Ready to merge.

PR #11688, opened 2022-09-24
#10271 t.w.error.Error: Fix handling of `message=None`

### LINKED ISSUE #10271: Problem in web.error str conversion
|[technic's avatar](https://github.com/technic)| @technic reported|
|-|-|
|Trac ID|trac#10271|
|Type|defect|
|Created|2021-10-31 22:06:47Z|

Hi,

This method returns an error

https://github.com/twisted/twisted/blob/3d0be31827585ade5078624e17a1c15240eece02/src/twisted/web/error.py#L92

"can't concat str and None" because self.message is None.

in __init__() message is empty string, and the code is 520, then _codeToMessage() returns None, and finally self.message is None.

I suggest to change the code to something like `self.message = message or "Unknown"`

```
trac-id__10271 10271
type__defect defect
reporter__technic technic
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__web web
keywords__None None
time__1635718007391402 1635718007391402
changetime__1635718007391402 1635718007391402
version__None None
owner__None None

```
</details>

twm:
Also add type annotations and validate HTTP status codes.

## Scope and purpose

Fixes #10271.

## Contributor Checklist:

### DIFF
--- a/src/twisted/web/error.py
+++ b/src/twisted/web/error.py
@@ -22,22 +22,20 @@ __all__ = [
 
 
 from collections.abc import Sequence
-from typing import cast
+from typing import Optional, Union, cast
 
 from twisted.python.compat import nativeString
 from twisted.web._responses import RESPONSES
 
 
-def _codeToMessage(code):
+def _codeToMessage(code: Union[int, bytes]) -> Optional[bytes]:
     """
     Returns the response message corresponding to an HTTP code, or None
     if the code is unknown or unrecognized.
 
-    @type code: L{bytes}
-    @param code: Refers to an HTTP status code, for example C{http.NOT_FOUND}.
+    @param code: HTTP status code, for example C{http.NOT_FOUND}.
 
     @return: A string message or none
-    @rtype: L{bytes}
     """
     try:
         return RESPONSES.get(int(code))
@@ -49,17 +47,23 @@ class Error(Exception):
     """
     A basic HTTP error.
 
-    @type status: L{bytes}
     @ivar status: Refers to an HTTP status code, for example C{http.NOT_FOUND}.
 
-    @type message: L{bytes}
     @param message: A short error message, for example "NOT FOUND".
 
-    @type response: L{bytes}
     @ivar response: A complete HTML document for an error page.
     """
 
-    def __init__(self, code, message=None, response=None):
+    status: bytes
+    message: Optional[bytes]
+    response: Optional[bytes]
+
+    def __init__(
+        self,
+        code: Union[int, bytes],
+        message: Optional[bytes] = None,
+        response: Optional[bytes] = None,
+    ) -> None:
         """
         Initializes a basic exception.
 
@@ -70,11 +74,12 @@ class Error(Exception):
             instead.
 
         @type message: L{bytes}
-        @param message: A short error message, for example "NOT FOUND".
+        @param message: A short error message, for example C{b"NOT FOUND"}.
 
         @type response: L{bytes}
         @param response: A complete HTML document for an error page.
         """
+
         message = message or _codeToMessage(code)
 
         Exception.__init__(self, code, message, response)
@@ -84,24 +89,38 @@ class Error(Exception):
             # downloadPage gives a bytes, Agent gives an int, and it worked by
             # accident previously, so just make it keep working.
             code = b"%d" % (code,)
+        elif len(code) != 3 or not code.isdigit():
+            # Status codes must be 3 digits. See
+            # https://httpwg.org/specs/rfc9110.html#status.code.extensibility
+            raise ValueError(f"Not a valid HTTP status code: {code!r}")
 
         self.status = code
         self.message = message
         self.response = response
 
     def __str__(self) -> str:
-        return nativeString(self.status + b" " + self.message)
+        s = self.status
+        if self.message:
+            s += b" " + self.message
+        return nativeString(s)
 
 
 class PageRedirect(Error):
     """
     A request resulted in an HTTP redirect.
 
-    @type location: L{bytes}
     @ivar location: The location of the redirect which was not followed.
     """
 
-    def __init__(self, code, message=None, response=None, location=None):
+    location: Optional[bytes]
+
+    def __init__(
+        self,
+        code: Union[int, bytes],
+        message: Optional[bytes] = None,
+        response: Optional[bytes] = None,
+        location: Optional[bytes] = None,
+    ) -> None:
         """
         Initializes a page redirect exception.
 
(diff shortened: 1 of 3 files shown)


twm:
Could we stop changing the PR title format? :roll_eyes:

twm:
please review

adiroiban:
The PR title was updates to make it easier to navigate the history.

Now, if you look at the commit history, you get quick links to both the issue and the PR

image


sorry for the trouble.

src/twisted/web/newsfragments/10271.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web.error.Error.__str__ no longer raises an exception when the error's message attribute is None. Additionally, it validates that code is a plausible 3-digit HTTP status code.
```
adiroiban, line 1: The current text is ok.

I am just requestion some feedback for towncrier.

I think there are times when you want 2 independent bugfixes for a single PR 

It would be better to show them in the final release as separate entries.

Rigth now, towcrier wraps everything that it find in the fragment file

Maybe something like this (with en empty line) is better and can instruct towncrier to generate 2 separate event handlers.

```suggestion
twisted.web.error.Error.__str__ no longer raises an exception when the error's message attribute is None.

twisted.web.error.Error validates at initialization that code is a plausible 3-digit HTTP status code.
```

src/twisted/web/test/test_error.py
```diff
@@ -42,19 +42,29 @@ class ErrorTests(unittest.TestCase):
...
         self.assertEqual(e.message, None)
+        self.assertEqual(str(e), "999")
+
+    def test_invalidStatus(self):
+        """
+        If C{code} isn't plausibly an HTTP status code (i.e., composed of
+        digits) it is rejected with L{ValueError}.
+        """
+        with self.assertRaises(ValueError):
+            error.Error(b"InvalidStatus")
```
adiroiban, line 67: Maybe something like this, to also check that we don't get a any ValueError

```suggestion
    error = self.assertRaises(ValueError, error.Error, b"InvalidStatus")
    self.assertEqual("Not a valid HTTP status code: InvalidStatus", error.args[0])
```

adiroiban (approved):
This looks good. Thanks for the cleanup.

But why do we allow message=None in the first place?

I would say that if we pass message=None at init, `web.error.Error` should automatically resolve it into some sort of default message.

That is, call `_codeToMessage` right away.

In this way, later, when you have to deal with the error, you don't need to create all kinds of special cases for message=None

But that should not block this PR..and maybe such a change would be a backward incompatible change.

src/twisted/web/error.py
```diff
@@ -22,22 +22,20 @@
...
 from collections.abc import Sequence
-from typing import cast
+from typing import Optional, Union, cast
 
 from twisted.python.compat import nativeString
 from twisted.web._responses import RESPONSES
 
 
-def _codeToMessage(code):
+def _codeToMessage(code: Union[int, bytes]) -> Optional[bytes]:
```
adiroiban, line 31: this is private method...so we should be fine.

But maybe is better to be strict and only allow bytes as input?

src/twisted/web/error.py
```diff
@@ -84,24 +89,38 @@ def __init__(self, code, message=None, response=None):
...
 
-    def __init__(self, code, message=None, response=None, location=None):
+    location: Optional[bytes]
+
+    def __init__(
+        self,
+        code: Union[int, bytes],
+        message: Optional[bytes] = None,
+        response: Optional[bytes] = None,
+        location: Optional[bytes] = None,
```
adiroiban, line 122: this is already in trunk, but does it make sense to have a redirection without a location ? Is it useful ?

src/twisted/web/test/test_error.py
adiroiban: I am -1 for using the context manager for `assertRaises` as it looks there are people who will add assertions inside the context manager, and will think that all is good, when in fact the assertion code is never executed.

I hope that the code coverage can help identity these types of errors, so I don't think we should enforce this.

is just a nice to have.

src/twisted/web/newsfragments/10271.bugfix
twm: In this case I think that the way to split this is to create a 10271.feature in addition to the .bugfix.

I am generally skeptical of the notion that we should work out these details at the PR level. IMO revising the combined release notes should be done as part of the release process by a human being. There is no way we can extract consistent prose from a bunch of fragment files written by random contributors.

src/twisted/web/error.py
twm: I think we issue this error based on status code without regard to whether there is actually a Location header. Is this good? 🤷‍♂️

src/twisted/web/error.py
twm: I did that in an earlier draft, but it led to awkward control flow to avoid coercing things that aren't already coerced (the `super()` arguments). I didn't think the compatibility implications of this worth the tidings of an internal interface.

src/twisted/web/test/test_error.py
twm: I appreciate this but as the suggested change is not correct and I'm not at my workstation I'm going to lea e it as-is.

src/twisted/web/newsfragments/10271.bugfix
adiroiban: well... we already do just a few releases per year and don't have many people volunteering for doing the release.

This is why I try to make the release process as simple and as quick as possible.

So making sure that random contributors are writing good release notes from the start, is one step in reducing the burden on the release manager :p

src/twisted/web/error.py
adiroiban: I don't know... I guess is ok... but not ideal.

A redirect without location, looks to me like a broken response :)

src/twisted/web/error.py
```diff
@@ -49,17 +47,23 @@ class Error(Exception):
...
-    def __init__(self, code, message=None, response=None):
+    status: bytes
+    message: Optional[bytes]
+    response: Optional[bytes]
+
+    def __init__(
+        self,
+        code: Union[int, bytes],
+        message: Optional[bytes] = None,
+        response: Optional[bytes] = None,
```
DMRobertson, line 65: I believe this change makes `PartialDownloadError.response: Optional[bytes]`. Is that correct? If we've got a partial download, we probably have some bytes? I suppose those might be all header bytes rather than response body bytes though, in which case maybe `response: None` does make sense?

(Noticed by https://github.com/matrix-org/synapse/issues/14008)

DMRobertson: FWIW, The only call site I could find is

https://github.com/twisted/twisted/blob/ab59aae25cd18c2cda96b16c165900cfc57f1f38/src/twisted/web/client.py#L1718-L1720

which gives a non-None `response` object.

twm: Seems likely. You could submit a ticket and PR to refine it in `PartialDownloadError` I guess?

PR #11691, opened 2022-09-26
#8146 #8147 Support systemd named inherited file descriptors

### LINKED ISSUE #8146: Document non-determinism for binding systemd sockets
|[jml's avatar](https://github.com/jml)| @jml reported|
|-|-|
|Trac ID|trac#8146|
|Type|enhancement|
|Created|2015-12-18 12:33:42Z|

The systemd documentation says:

> If multiple socket units activate the same service, the order of the file descriptors passed to its main process is undefined.

(see http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html)

This means that if a Twisted process uses multiple systemd endpoints, it cannot be sure that the right sockets are bound to the right services.

We should document this limitation in Twisted.

c.f. https://clusterhq.atlassian.net/browse/FLOC-3712.

```
trac-id__8146 8146
type__enhancement enhancement
reporter__jml jml
priority__low low
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__documentation documentation
time__1450442022165513 1450442022165513
changetime__1450442022165513 1450442022165513
version__None None
owner__None None

```
</details>

### LINKED ISSUE #8147: Support named systemd sockets
|[jml's avatar](https://github.com/jml)| @jml reported|
|-|-|
|Trac ID|trac#8147|
|Type|enhancement|
|Created|2015-12-18 12:39:15Z|

systemd 227 added support for named sockets: https://github.com/systemd/systemd/blob/master/NEWS#L316-L322

These are documented at http://www.freedesktop.org/software/systemd/man/sd_listen_fds.html

They are exported as the `LISTEN_FDNAMES` environment variables, "which may contain a colon-separated list of names".

This could be a useful thing to support, esp. to mitigate the limitations in [#8146](https://github.com/twisted/twisted/issues/8146).

```
trac-id__8147 8147
type__enhancement enhancement
reporter__jml jml
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1450442355443783 1450442355443783
changetime__1450442355443783 1450442355443783
version__None None
owner__None None

```
</details>

exarkun:
## Scope and purpose

Fixes #8146
Fixes #8147

The unit tests for systemd.py were well-intentioned but sadly misguided (nice try though, jean-paul@2012).  So I kind of rewrote them (a lot of the tests are the same but the infrastructure they fit into is better now).

### DIFF
--- a/docs/core/howto/listings/systemd/www.example.com.socket
+++ b/docs/core/howto/listings/systemd/www.example.com.socket
@@ -1,5 +1,6 @@
 [Socket]
 ListenStream=0.0.0.0:80
+FileDescriptorName=www
 
 [Install]
 WantedBy=sockets.target
--- a/docs/core/howto/listings/systemd/www.example.com.socket
+++ b/docs/core/howto/listings/systemd/www.example.com.socket
@@ -1,6 +1,6 @@
 [Socket]
 ListenStream=0.0.0.0:80
-FileDescriptorName=www
+FileDescriptorName=my-web-port
 
 [Install]
 WantedBy=sockets.target
--- a/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
+++ b/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
@@ -5,7 +5,7 @@ Description=Example Web Server
 ExecStart=/usr/bin/twistd \
     --nodaemon \
     --pidfile= \
-    web --listen systemd:domain=INET:index=0 --path .
+    web --listen systemd:domain=INET:name=www --path .
 
 NonBlocking=true
 
--- a/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
+++ b/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Example Web Server
+Requires=www.example.com.socket
 
 [Service]
 ExecStart=/usr/bin/twistd \
--- a/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
+++ b/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
@@ -5,7 +5,7 @@ Description=Example Web Server
 ExecStart=/usr/bin/twistd \
     --nodaemon \
     --pidfile= \
-    web --listen systemd:domain=INET:name=www --path .
+    web --listen systemd:domain=INET:name=my-web-port --path .
 
 NonBlocking=true
 
--- a/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
+++ b/docs/core/howto/listings/systemd/www.example.com.socketactivated.service
@@ -8,8 +8,6 @@ ExecStart=/usr/bin/twistd \
     --pidfile= \
     web --listen systemd:domain=INET:name=my-web-port --path .
 
-NonBlocking=true
-
 WorkingDirectory=/srv/www/www.example.com/static
 
 User=nobody
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -282,7 +282,7 @@ ExecStart
 
   The ``domain=INET`` endpoint argument makes ``twistd`` treat the inherited file descriptor as an IPv4 socket.
 
-  The ``index=0`` endpoint argument makes ``twistd`` adopt the first file descriptor inherited from ``systemd``\ .
+  The ``name=www`` endpoint argument makes ``twistd`` adopt the file descriptor inherited from ``systemd`` named ``www``.
 
   Socket activation is also technically possible with other socket families and types, but Twisted currently only accepts IPv4 and IPv6 TCP sockets. See :ref:`limitations` below.
 
@@ -353,7 +353,7 @@ You can verify this by using systemctl to report the status of the service. eg
               Active: active (running) since Tue 2013-01-29 15:02:20 GMT; 3s ago
             Main PID: 25605 (twistd)
               CGroup: name=systemd:/system/www.example.com.service
-                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:index=0 --path .
+                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:name=www --path .
 
     Jan 29 15:02:20 zorin.lan systemd[1]: Started Example Web Server.
     Jan 29 15:02:20 zorin.lan twistd[25605]: 2013-01-29 15:02:20+0000 [-] Log opened.
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -291,10 +291,6 @@ ExecStart
 
   Socket activation is also technically possible with other socket families and types, but Twisted currently only accepts IPv4 and IPv6 TCP sockets. See :ref:`limitations` below.
 
-NonBlocking
-
-  This must be set to ``true`` to ensure that ``systemd`` passes non-blocking sockets to Twisted.
-
 Requires
 
   The service no longer knows how to bind the listening port for itself.
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -282,7 +282,7 @@ ExecStart
 
   The ``domain=INET`` endpoint argument makes ``twistd`` treat the inherited file descriptor as an IPv4 socket.
 
-  The ``name=www`` endpoint argument makes ``twistd`` adopt the file descriptor inherited from ``systemd`` named ``www``.
+  The ``name=my-web-port`` endpoint argument makes ``twistd`` adopt the file descriptor inherited from ``systemd`` named ``my-web-port``.
 
   Socket activation is also technically possible with other socket families and types, but Twisted currently only accepts IPv4 and IPv6 TCP sockets. See :ref:`limitations` below.
 
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -295,6 +295,11 @@ NonBlocking
 
   This must be set to ``true`` to ensure that ``systemd`` passes non-blocking sockets to Twisted.
 
+Requires
+
+  The service no longer knows how to bind the listening port for itself.
+  The corresponding socket unit must be started so it can pass the listening port on to the ``twistd`` process.
+
 [Install]
 
   In this example, the ``[Install]`` section has been moved to the socket configuration file.
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -358,7 +358,7 @@ You can verify this by using systemctl to report the status of the service. eg
               Active: active (running) since Tue 2013-01-29 15:02:20 GMT; 3s ago
             Main PID: 25605 (twistd)
               CGroup: name=systemd:/system/www.example.com.service
-                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:name=www --path .
+                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:name=my-web-port --path .
 
     Jan 29 15:02:20 zorin.lan systemd[1]: Started Example Web Server.
     Jan 29 15:02:20 zorin.lan twistd[25605]: 2013-01-29 15:02:20+0000 [-] Log opened.
--- a/docs/core/howto/systemd.rst
+++ b/docs/core/howto/systemd.rst
@@ -65,15 +65,15 @@ Twisted
 
 Basic Systemd Service Configuration
 -----------------------------------
-The essential configuration file for a ``systemd`` service is the `service <http://www.freedesktop.org/software/systemd/man/systemd.service.html>`_ file.
+The essential configuration file for a ``systemd`` service is the `service file <http://www.freedesktop.org/software/systemd/man/systemd.service.html>`_.
 
 Later in this tutorial, you will learn about some other types of configuration file, which are used to control when and how your service is started.
 
 But we will begin by configuring ``systemd`` to start a Twisted web server immediately on system boot.
 
-Create a systemd.service file
+Create a systemd service file
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Create the `service <http://www.freedesktop.org/software/systemd/man/systemd.service.html>`_ file at ``/etc/systemd/system/www.example.com.service`` with the following content:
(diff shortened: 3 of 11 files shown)


docs/core/howto/listings/systemd/www.example.com.socket
```diff
@@ -1,5 +1,6 @@
 [Socket]
 ListenStream=0.0.0.0:80
+FileDescriptorName=www
```
adiroiban: this is what is used in the docs, but I find it a bit confusing / hard to understand.

The `www` value generic.

Maybe call it `custom-socker` or `example-socket-name`  ...

A very minor comment.

docs/core/howto/systemd.rst
```diff
@@ -282,7 +282,7 @@ ExecStart
 
   The ``domain=INET`` endpoint argument makes ``twistd`` treat the inherited file descriptor as an IPv4 socket.
 
-  The ``index=0`` endpoint argument makes ``twistd`` adopt the first file descriptor inherited from ``systemd``\ .
+  The ``name=www`` endpoint argument makes ``twistd`` adopt the file descriptor inherited from ``systemd`` named ``www``.
```
adiroiban: I don't know where to comment.

But at line 261 we have

```
This configuration file contains the following important directives:
```

I think `FileDescriptorName` should be in the list.


----------

Also at line line 272 I think the following diff is requried

```diff
-You also need to modify the ``systemd.service`` file as follows:
+You also need to modify the ``www.example.com.service`` file as follows:
```

----------

And besides `NonBlocking` I think we also make sure the service is configured with `Accept=false`

Maybe `false` is the default value, but I think is important to make sure systemd will not spawn a separate twistd process for each connection, and will let twistd to handle the lower level networking.

adiroiban: I don't know where to comment.

But at line 261 we have

```
This configuration file contains the following important directives:
```

I think `FileDescriptorName` should be in the list.

src/twisted/python/test/strategies.py
```diff
@@ -0,0 +1,33 @@
...
+# See LICENSE for details.
+
+"""
+Hypothesis strategies for values related to L{twisted.python}.
+"""
+
+from hypothesis.strategies import SearchStrategy, characters, text
+
+
+def systemd_descriptor_names() -> SearchStrategy[str]:
```
adiroiban: just a minor comment, and it might be a matter of style, and the current implementation is preferred.

If a strategy is used in single test module / test file , maybe we can have it defined in the same test file, closer to where it is used.

But maybe having it in a separate file from the start, encourages people to reuse existing strategies and code.

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
-        self.assertEqual(
-            sddaemon.inheritedDescriptors(), sddaemon.inheritedDescriptors()
-        )
+from twisted.python.systemd import ListenFDs
+from twisted.trial.unittest import SynchronousTestCase
+from .strategies import systemd_descriptor_names
 
 
-class MemoryOnlyMixin:
+def initializeEnvironment(count: int, pid: object) -> Dict[str, str]:
```
adiroiban: I think we can call it a getter, to make sure no side effects happen here.

And we call it `getEnvironment` as  in the context of theses tests, it shold be clear that we have an systemd environment

```suggestion
def getEnvironment(count: int, pid: object) -> Dict[str, str]:
```

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
     """
-    Mixin for a L{TestCase} subclass which creates creating a fake, in-memory
-    implementation of C{inheritedDescriptors}.  This provides verification that
-    the fake behaves in a compatible way with the real implementation.
+    Create a copy of the process environment and add I{LISTEN_FDS} and
+    I{LISTEN_PID} (the environment variables set by systemd) to it.
```
adiroiban: I think the description can be generic. No need to mention the exact env var names.

I know that this was already in trunk... just a minor comment

```suggestion
    Return a copy of the process environment with I{systemd} specific variables added to it.
```

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
-    """
-
-
-class EnvironmentTests(EnvironmentMixin, InheritedDescriptorsMixin, TestCase):
-    """
-    Apply tests to L{ListenFDs}, constructed based on an environment dictionary.
-    """
+        self.patch(os, "environ", initializeEnvironment(5, os.getpid()))
+        sddaemon = ListenFDs.fromEnvironment()
+        self.assertEqual(list(range(3, 3 + 5)), sddaemon.inheritedDescriptors())
```
adiroiban: I now understand what this test does, but I think the the actual environent variables are important for this tests, and they are hidden in `initializeEnvironment`

Also, for these tests, I prefer less auto-generated expected values.
Should be easier to read and less error prone.

I think that when writing tests is best to be more verbose to make sure
the test is correct and easy to review.

And I think that this can help when you are looking for a big at 2AM, after 10 hours at work

```suggestion
        environ = initializeEnvironment(4, os.getpid())
        environ['LISTEN_FDS'] = 4
        self.patch(os, "environ", )
        sddaemon = ListenFDs.fromEnvironment()
        # The first 3 are standard streams.
        # Our extra streams are added after them.
        self.assertEqual([3, 4, 5, 6], sddaemon.inheritedDescriptors())
```

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
 
-    def test_secondEnvironment(self):
+    def test_secondEnvironment(self) -> None:
         """
-        Only a single L{Environment} can extract inherited file descriptors.
+        L{ListenFDs.fromEnvironment} removes information about the inherited
+        file descriptors from the environment mapping.
```
adiroiban: I see this is existing behaviour... but I am not sure why we should care about this behaviour / feature.

Both the old and the new text description don't help answer this question.

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
         self.assertEqual(list(range(3, 6)), first.inheritedDescriptors())
         self.assertEqual([], second.inheritedDescriptors())
 
-    def test_mismatchedPID(self):
+    def test_mismatchedPID(self) -> None:
         """
-        If the current process PID does not match the PID in the environment, no
-        inherited descriptors are reported.
+        If the current process PID does not match the PID in the environment,
+        no inherited descriptors are reported.
```
adiroiban: Just some extra info about real world use case of this functionality/feature

```suggestion
        If the current process PID does not match the PID in the environment,
        no inherited descriptors are reported.
        
        This is a sign that the systemd sockets are not for us.
        Most probably someone called our code without priour clean of the environment.
```

src/twisted/python/test/test_systemd.py
```diff
@@ -7,158 +7,163 @@
...
+    @given(dictionaries(systemd_descriptor_names(), integers(min_value=0), max_size=10))
+    def test_inheritedNamedDescriptors(self, expected: Mapping[str, int]) -> None:
+        """
+        L{ListenFDs.inheritedNamedDescriptors} returns a mapping from the
+        descriptor names to their integer values, with items formed by
+        pairwise combination of the input descriptors and names.
+        """
+        items = list(expected.items())
+        names = [name for name, _ in items]
+        descriptors = [fd for _, fd in items]
```
adiroiban, line 163: this is ok...but when you are tired, this is harder to read, in comparison with passing a static input data.

just some feedback about hypothesis 

I try to have the test code as simple as possible...as when I am refactoring something, I will see other tests failing, and then I need to read the test to see if the test is valid, or if there was a bug in the previous test itself.

src/twisted/python/systemd.py
```diff
@@ -68,26 +70,70 @@ def fromEnvironment(
...
+    Parse the I{LISTEN_FDNAMES} environment variable supplied by systemd.
+
+    @param environ: The environment variable mapping in which to look for the
+        value to parse.
+
+    @return: The names of the inherited descriptors, in order.
+    """
+    names = environ.get("LISTEN_FDNAMES", "")
+    if len(names) > 0:
+        return tuple(names.split(":"))
```
adiroiban, line 153: I guess that systemd will always pass sane env variables, so we don't need to stip any leading/trailing spaces.

src/twisted/python/systemd.py
```diff
@@ -68,26 +70,70 @@ def fromEnvironment(
...
+
+    @return: The integer values of the inherited file descriptors, in order.
+    """
+    try:
+        count = int(environ["LISTEN_FDS"])
+    except (KeyError, ValueError):
+        return []
+    else:
+        descriptors = list(range(start, start + count))
+        del environ["LISTEN_PID"], environ["LISTEN_FDS"]
```
adiroiban, line 138: I left a comment in the test case.
It is not clear why we are doing this.

I know this is already in `trunk` but looks like a strange side effect.

```suggestion
        # We do THIS, because of THAT.
        del environ["LISTEN_PID"], environ["LISTEN_FDS"]
```

src/twisted/python/systemd.py
```diff
@@ -68,26 +70,70 @@ def fromEnvironment(
...
+
+        return cls(descriptors, names)
+
+    def inheritedDescriptors(self) -> List[int]:
         """
         @return: The configured descriptors.
         """
         return list(self._descriptors)
+
+    def inheritedNamedDescriptors(self) -> Dict[str, int]:
```
adiroiban, line 98: this is ok as I guess is the same "design" as the previous `inheritedDescriptors` API.

But do we need to get the full dict?

I see this is rather used as `inheritedNamedDescriptors(name) -> int`

But this being a low-level, generic API, maybe the current API design is OK.

src/twisted/python/systemd.py
```diff
@@ -68,26 +70,70 @@ def fromEnvironment(
...
-        return cls(descriptors)
-
-    def inheritedDescriptors(self) -> Iterable[int]:
+            descriptors = []
+            names = ()
+
+        # They may both be missing (consistent with not running under systemd
+        # at all) or they may both be present (consistent with running under
+        # systemd 227 or newer).  It is not allowed for only one to be present
+        # or for the values to disagree with each other.
```
adiroiban: If no names are defined for a socket, I guess that systemd will just use a generic `1.socket` name.

It would be nice to have a link with this specification, but is not required.

adiroiban (approved):
I left a few comments.

The code looks good and as long as the tests are green it can be merged.

I am doing these reviews in "emergency" mode, trying not to block the work from @exarkun  ... but I am always tired after work and not in the best shape for review.

Doing proper reviews requires a full time job, not after hours work.

Jean-Paul, I will accept this and put it back for review.

If no other Twisted dev will have time for it, I think is best to merge it.

Thanks

adiroiban:
just asking a review from another team member to make sure I don't miss anything needs-review

adiroiban:
I think that if no other review is made in the next few days, is better to merge this.

I think this is already a bit of improvement over trunk.

Thanks for the docs updates. All the comments that I left are minor and should not block the merge.

src/twisted/python/systemd.py
exarkun: Yea I think that probably would have been alright but I guess consistency in the interface is better than just doing something marginally better for one of the two APIs.  Indeed it is a low-level component that most applications will never touch.

exarkun: (I guess it's also worth mentioning that with the list/dict interfaces you get strictly more power since they let you enumerate the descriptors and you can always index them yourself ... I don't know what usage pattern would actually benefit from this but maybe there is one)

src/twisted/python/systemd.py
exarkun: If it starts doing weird things with formatting I don't think we have any way to make any sense of the weirdness so, yea, I don't think any parsing beyond what's here makes sense.

I wonder if I should drag in a new dependency on `parsec` and do this parsing right, though.... (only half serious)

src/twisted/python/test/strategies.py
exarkun: My thinking here is definitely along the lines of improving re-use.  Having a convenient strategies modules with all kinds of fun stuff in it seems like it might encourage that.  I don't really know.  I haven't worked with any large teams on projects that use Hypothesis so I've only tried these strategies on myself before. :slightly_smiling_face: 

I do know notice that this strategy name violates the Twisted naming convention though...  Hard to keep straight these days :/

exarkun: A little bit more on this...  Perhaps this systemd strategy will only ever be used by `test_systemd`.  I still think it's a good idea to have it out of the test module itself.  The test module's job is not to supply the canonical definition of a systemd "FileDescriptorName" value.  The test module's job is to define some behavior _in terms of_ that.  If I could import this systemd-related strategy from a third-party "systemd" package then I think that would be even better.  Since there is no such thing, I think putting the strategy in a general "strategies" module in Twisted is the next best thing.

docs/core/howto/listings/systemd/www.example.com.socketactivated.service
```diff
@@ -5,7 +5,7 @@ Description=Example Web Server
 ExecStart=/usr/bin/twistd \
     --nodaemon \
     --pidfile= \
-    web --listen systemd:domain=INET:index=0 --path .
+    web --listen systemd:domain=INET:name=www --path .
```
twm: It looks like the prose and socket unit call this `my-web-port`:

```suggestion
    web --listen systemd:domain=INET:name=my-web-port --path .
```

docs/core/howto/systemd.rst
```diff
@@ -353,7 +358,7 @@ You can verify this by using systemctl to report the status of the service. eg
               Active: active (running) since Tue 2013-01-29 15:02:20 GMT; 3s ago
             Main PID: 25605 (twistd)
               CGroup: name=systemd:/system/www.example.com.service
-                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:index=0 --path .
+                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:name=www --path .
```
twm:
```suggestion
                      └─25605 /usr/bin/python /usr/bin/twistd --nodaemon --pidfile= web --port systemd:domain=INET:name=my-web-port --path .
```

twm (approved):
This all looks reasonable to me. I'd echo Adi in the sentiment that Hypothesis seems like overkill here, but no need to change that. Thanks!

Dikuko (approved):

Issue #6016, opened 2012-09-24
Switch trial's profiling code to use twisted.application.app's profiling support instead


twisted-trac:
|[<img alt="itamarst's avatar" src="https://avatars.githubusercontent.com/u/3266662?s=50" width="50" height="50">](https://github.com/itamarst)| @itamarst reported|
|-|-|
|Trac ID|trac#6016|
|Type|enhancement|
|Created|2012-09-24 15:27:30Z|

The profiling code in twisted.application is more extensive, and tested. trial should use it instead of `twisted.trial.util.profiled`.

<details><summary>Searchable metadata</summary>

```
trac-id__6016 6016
type__enhancement enhancement
reporter__itamar itamar
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__ 
time__1348500450000000 1348500450000000
changetime__1348500450000000 1348500450000000
version__None None
owner__ 

```
</details>

PR #11693, opened 2022-09-27
#11692 cleanup trial runner

### LINKED ISSUE #11692: Clean up twisted.trial.runner

twisted.trial.runner has some pretty cruft old stuff in it:

* It has an implicit interface with debuggers
* It manually implements resource allocation/cleanup for log files and the working directory
* `TrialRunner` has a dozen or more attributes manually defined and of poorly document types
* There are few type annotations
* There is loooong-since deprecated support for reporters with no `done` method
* The test suite involves monkey-patching implementation details to observe various side-effects
* The test suite involves monkey-patching implementation details to fix unexplained bugs which apparently no longer exist

All of this stuff gets in the way of trying to implement new features like #11671 .  Let's start to clean it up.

exarkun:
## Scope and purpose

Fixes #11692

### DIFF
--- a/src/twisted/scripts/trial.py
+++ b/src/twisted/scripts/trial.py
@@ -10,7 +10,9 @@ import pdb
 import random
 import sys
 import time
+import trace
 import warnings
+from typing import NoReturn, Optional, Type
 
 from twisted import plugin
 from twisted.application import app
@@ -19,6 +21,8 @@ from twisted.python import failure, reflect, usage
 from twisted.python.filepath import FilePath
 from twisted.python.reflect import namedModule
 from twisted.trial import itrial, reporter, runner
+from twisted.trial._dist.disttrial import DistTrialRunner
+from twisted.trial.unittest import TestSuite
 
 # Yea, this is stupid.  Leave it for command-line compatibility for a
 # while, though.
@@ -232,7 +236,7 @@ class _BasicOptions:
     )
 
     fallbackReporter = reporter.TreeReporter
-    tracer = None
+    tracer: Optional[trace.Trace] = None
 
     def __init__(self):
         self["tests"] = []
@@ -275,8 +279,6 @@ class _BasicOptions:
         Generate coverage information in the coverage file in the
         directory specified by the temp-directory option.
         """
-        import trace
-
         self.tracer = trace.Trace(count=1, trace=0)
         sys.settrace(self.tracer.globaltrace)
         self["coverage"] = True
@@ -476,7 +478,6 @@ class Options(_BasicOptions, usage.Options, app.ReactorSelectionMixin):
 
     fallbackReporter = reporter.TreeReporter
     extra = None
-    tracer = None
 
     def opt_jobs(self, number):
         """
@@ -523,7 +524,7 @@ class Options(_BasicOptions, usage.Options, app.ReactorSelectionMixin):
             failure.DO_POST_MORTEM = False
 
 
-def _initialDebugSetup(config):
+def _initialDebugSetup(config: Options) -> None:
     # do this part of debug setup first for easy debugging of import failures
     if config["debug"]:
         failure.startDebugMode()
@@ -531,13 +532,13 @@ def _initialDebugSetup(config):
         defer.setDebugging(True)
 
 
-def _getSuite(config):
+def _getSuite(config: Options) -> TestSuite:
     loader = _getLoader(config)
     recurse = not config["no-recurse"]
     return loader.loadByNames(config["tests"], recurse=recurse)
 
 
-def _getLoader(config):
+def _getLoader(config: Options) -> runner.TestLoader:
     loader = runner.TestLoader()
     if config["random"]:
         randomer = random.Random()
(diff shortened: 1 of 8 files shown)


src/twisted/scripts/trial.py
```diff
@@ -649,13 +642,13 @@ def run():
 
     suite = _getSuite(config)
     if config["until-failure"]:
-        test_result = trialRunner.runUntilFailure(suite)
+        testResult = trialRunner.runUntilFailure(suite)
```
adiroiban, line 645: ir looks like we have a missing automated tests here.

I think is ok to merge this without adding a new test for this change, as the change is minor.

src/twisted/trial/runner.py
```diff
@@ -258,26 +271,31 @@ def run(self, result):
...
     """
     @param thing: an object from modules (instance of PythonModule,
         PythonAttribute), a TestCase subclass, or an instance of a TestCase.
     """
-    if isTestCase(thing):
+    if isinstance(thing, pyunit.TestCase):
+        return thing.id()
+    elif isinstance(thing, (modules.PythonAttribute, modules.PythonModule)):
+        return thing.name
+    elif isTestCase(thing):
```
adiroiban: very minor style comment ... while reading the `assert False` 


```suggestion
    if isinstance(thing, pyunit.TestCase):
        return thing.id()

    if isinstance(thing, (modules.PythonAttribute, modules.PythonModule)):
        return thing.name

    if isTestCase(thing):
```

src/twisted/trial/runner.py
```diff
@@ -258,26 +271,31 @@ def run(self, result):
...
     else:
-        # thing from trial, or thing from modules.
-        # this monstrosity exists so that modules' objects do not have to
-        # implement id(). -jml
-        try:
-            theName = thing.id()
-        except AttributeError:
-            theName = thing.name
-    return theName
+        assert False
```
adiroiban: I have no idea why we need this.

If this is really needed, I think it needs a comment.

I see there is no coverage for this line.

src/twisted/trial/runner.py
```diff
@@ -765,6 +782,62 @@ def _qualNameWalker(qualName):
...
+    observerFunction = logFileObserver.emit
+    log.startLoggingWithObserver(observerFunction, 0)
+
+    yield
+
+    log.removeObserver(observerFunction)
+    logFile.close()
+
+
+class _Runner(Protocol):
```
adiroiban, line 834: maybe have docstring to describe why we need this.

Do we use Protocol or zope.Interface ?

src/twisted/trial/runner.py
```diff
@@ -797,64 +873,30 @@ def _makeResult(self):
...
-        self._logFileObject = logFile
-        self._logFileObserver = log.FileLogObserver(logFile)
-        log.startLoggingWithObserver(self._logFileObserver.emit, 0)
+    @property
+    def tbformat(self) -> str:
+        return self._tracebackFormat
 
-    def run(self, test):
+    @property
+    def rterrors(self) -> bool:
```
adiroiban, line 921: any idea what real-time errors are? :)

src/twisted/trial/util.py
```diff
@@ -240,8 +241,12 @@ def suppress(action="ignore", **kwarg):
 
 # This should be deleted, and replaced with twisted.application's code; see
 # #6016:
```
adiroiban: very minor comment.

Maybe replace it with the full link https://github.com/twisted/twisted/issues/6016

not sure what is best.

src/twisted/trial/_dist/disttrial.py
```diff
@@ -287,7 +287,7 @@ class DistTrialRunner:
         ``False`` to run through the whole suite and report all of the results
         at the end.
 
-    @ivar _stream: stream which the reporter will use.
+    @ivar stream: stream which the reporter will use.
```
adiroiban, line 290: Do we need it as public member?

I checked the diff, I don't see any usage, other than `self.stream`

adiroiban (approved):
Many thanks for the cleanup

Other than the test case `name` implementation, this looks good.

I think this can be merged.

I don't see anything horrible here that should block a merge (other than the `name` part :)

src/twisted/trial/_dist/disttrial.py
exarkun: Unfortunately it is public on the de-facto "runner" interface defined by `TrialRunner`, which this class is trying to be compatible with.  It was a nice idea to make it private here but that violates the runner interface so it doesn't really work.  As soon as I made the protocol explicit with `_Runner` mypy began to complain that `DistTrialRunner` did not conform.

src/twisted/trial/runner.py
exarkun: The `assert False` in the `else` of this construction makes mypy happy that all branches are covered :/  I don't know why it can't figure it out that the if-elif cases cover everything.  I think `isTestCase(...)` may be tripping it up but I don't understand why.

src/twisted/trial/runner.py
exarkun: I'll try to get rid of it.  If I can't, I'll add a comment.

src/twisted/trial/runner.py
exarkun: I'm not aware of a project policy but I think Protocol is the right default choice.  We can still use zope.interface where it does something better but in the case of a tie, Protocol is supported by mypy natively natively so it seems preferable.

src/twisted/trial/runner.py
exarkun: added the info to the class docstring

PR #11698, opened 2022-10-05
#11685 Share ReactorCore between ReactorBase and _FakeReactor

### LINKED ISSUE #11685: Factor IReactorCore out of twisted.test.test_task._FakeReactor and share the implementation with ReactorBase
This is a step towards #11673.  `_FakeReactor` declares that it implements `IReactorCore` but only offers an incomplete and incorrect implementation.  We have a complete and correct (or at least realistic) implementation in `ReactorBase`.

Share that implementation with `_FakeReactor`.

glyph:
## Scope and purpose

Fixes #11685

This is now has #11703 merged into it so review that first.


src/twisted/internet/_signals.py
```diff
@@ -66,3 +86,356 @@ def isDefaultHandler():
     Determine whether the I{SIGCHLD} handler is the default or not.
     """
     return signal.getsignal(signal.SIGCHLD) == signal.SIG_DFL
+
+
+@implementer(IReadDescriptor)
+class _FDWaker(log.Logger):
```
exarkun, line 94: Note: This is mostly just being moved from posixbase.py

src/twisted/internet/_signals.py
```diff
@@ -66,3 +86,356 @@ def isDefaultHandler():
...
+            return
+        for fd in self.i, self.o:
+            try:
+                os.close(fd)
+            except OSError:
+                pass
+        del self.i, self.o
+
+
+class _SIGCHLDWaker(_FDWaker):
```
exarkun, line 141: Note: This is mostly just being moved from posixbase.py

src/twisted/internet/_signals.py
```diff
@@ -66,3 +86,356 @@ def isDefaultHandler():
...
+        Read some data from my connection and discard it.
+        """
+
+    def connectionLost(self, reason: failure.Failure) -> None:
+        """
+        Called when connection was closed and the pipes.
+        """
+
+
+class _SocketWaker(log.Logger):
```
exarkun, line 365: Note: This is mostly just being moved from posixbase.py

src/twisted/internet/_signals.py
```diff
@@ -66,3 +86,356 @@ def isDefaultHandler():
...
+            self.r.recv(8192)
+        except OSError:
+            pass
+
+    def connectionLost(self, reason):
+        self.r.close()
+        self.w.close()
+
+
+class _UnixWaker(_FDWaker):
```
exarkun, line 415: Note: This is mostly just being moved from posixbase.py

src/twisted/internet/_signals.py
```diff
@@ -66,3 +86,356 @@ def isDefaultHandler():
...
+        """
+
+    def uninstall(self) -> None:
+        """
+        Do nothing because L{install} installed nothing.
+        """
+
+
+@frozen
+class _WithSignalHandling:
```
exarkun, line 212: This is mostly the old logic from `_SignalReactorMixin` which no longer exists.

src/twisted/internet/base.py
```diff
@@ -587,32 +600,307 @@ class ReactorBase(PluggableResolverMixin):
...
+            wakeUp=self.wakeUp,
+        )
+        self.resolve = self._core.resolve  # type: ignore[assignment]
+        self.fireSystemEvent = self._core.fireSystemEvent  # type: ignore[assignment]
+        self.addSystemEventTrigger = self._core.addSystemEventTrigger  # type: ignore[assignment]
+        self.callWhenRunning = self._core.callWhenRunning  # type: ignore[assignment]
+        self.removeSystemEventTrigger = self._core.removeSystemEventTrigger  # type: ignore[assignment]
+        self.stop = self._core.stop  # type: ignore[assignment]
+        self.crash = self._core.crash  # type: ignore[assignment]
+        self.iterate = self._core.iterate  # type: ignore[assignment]
```
exarkun, line 896: I think this whole block is pretty ugly.  We need to keep exposing the `IReactorCore` interface here.  Instead of copying all these attributes over, we could use `proxyForInterface` to automate the delegation.  `proxyForInterface` results in a dynamic base class so mypy will never like it.  I think pydoctor is also a consideration.  All of these `type: ignore` comments are required because there are also stub implementations of these methods below, present only for doc generation.

I'm not sure which of these I like least ... maybe the one that's checked in though.

exarkun:
Superseded by 11685-share-reactorcore-with-fakereactor.4

PR #11701, opened 2022-10-05
#11700 #11699 Remove pyupgrade from our pre-commit tools

### LINKED ISSUE #11699: Run tests with `-j` in the tox environments on Linux
`python -m twisted.trial -j20 twisted` completes successfully on my NixOS development system.  Instead of taking 230 seconds or more, it takes 30 seconds.

The tox environment for running the test suite could provide a `-j...` argument to speed up test suite runs.  This would presumably be a benefit for local development (for those who use tox) as well as on CI (but the number of workers used on CI should be considered and probably tailored to the CI execution environment).

### LINKED ISSUE #11700: Remove pyupgrade from our local dev tools and CI
JP complains that it breaks the code.

Glyph 

> @adiroiban: I think pre-commit is great (perhaps I differ from jean-paul. here) but pyupgrade is just wrong
> it makes changes that cause unit tests to fail
> we should remove it from the pipeline, I'm not sure why it got added in the first place

adiroiban:
## Scope and purpose

Fixes #11700
Fixes #11699

See more info in the ticket.

It looks like pyupgrade has good intentions...but it is broken.

As a drive-by it also enabled the parallel test runs for trial by default.
This should result in much faster tests.

It should also (forcefully :) expose more people to this wonderful feature of `trial`.


glyph (approved):
~~It looks like the attempt to address #11699 at the same time is hitting coverage, probably because of some obscure problem with `coverage.process_startup()` not propagating everywhere it needs to.  If the codecov status resolves feel free to land as-is, otherwise, drop that patch, land the pyupgrade change, and make a new PR so we can investigate separately~~

Looks like consistency eventually'd while I was writing that so SHIP IT

PR #11703, opened 2022-10-05
#11702 move some signal code

### LINKED ISSUE #11702: There is a lot of signal-handling code in `twisted/internet/base.py` when it could be near other signal handling code in `twisted/internet/_signals.py`
If the signal-handling code were all in one place then the various reactor implementation modules would each be a little bit more focused and so easier to understand and maintain.

exarkun:
## Scope and purpose

Fixes #11702

This was glyph's idea while looking at #11685 which refactored a lot of code but left all of the signal-handling pieces inside `base.py` and `posixbase.py`.  I pulled this code movement out because the branch for #11685 is over 1000 lines of diff including these changes.

This is intended to be even _less_ than a refactoring.  It's just mechanical movement of source code from one place to another.

Diff coverage here is not _great_ but all of the uncovered code was also uncovered in its original location and all this does is move it.

### DIFF
--- a/src/twisted/internet/_glibbase.py
+++ b/src/twisted/internet/_glibbase.py
@@ -19,6 +19,7 @@ from zope.interface import implementer
 from twisted.internet import base, posixbase, selectreactor
 from twisted.internet.interfaces import IReactorFDSet
 from twisted.python import log
+from ._signals import _UnixWaker
 
 
 def ensureNotImported(moduleNames, errorMessage, preventImports=[]):
@@ -48,7 +49,7 @@ def ensureNotImported(moduleNames, errorMessage, preventImports=[]):
         sys.modules[name] = None
 
 
-class GlibWaker(posixbase._UnixWaker):
+class GlibWaker(_UnixWaker):
     """
     Run scheduled events after waking up.
     """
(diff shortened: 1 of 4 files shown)


twm (approved):
The code movement and minor changes look reasonable to me. Definitely organizational improvement. Thanks for splitting this out!

From the coverage results I guess we aren't running any CFReactor tests in CI? That seems bad, but oh well, Apple makes it such a pain to their users' detriment.

adiroiban:
on macOS we only run 3.10-macos-12-default-tests , with the default reactor on macOS. I don't know which reactor is the default on macOS... I guess is the posix select. 

I guess this is ok for now. If somebody cares about macOS, the CI infrastructure is ready to run the tests with a non-default reactor.
We do it for Windows, for which the tests are executed on both select and iocp.

Thanks Jean-Paul for the cleanup. Thanks Tom for the review.

exarkun:
On the GitHub Actions macOS job, the CFReactor tests are all skipped like:

> No module named 'CFNetwork'

So maybe it's just a matter of missing some dependencies.

adiroiban:
> So maybe it's just a matter of missing some dependencies.

Thanks. True. Those are the cfreactor targeted tests... and that would be a first step.

At least for  windows, we also try to run the whole tests, not only the `iocp` tests, using the iocp reactor. For a better coverage, and in the case in which the targeted `iocp` have missing coverage.

exarkun:
I filed https://github.com/twisted/twisted/issues/11708

PR #11711, opened 2022-10-09
#11710 more disttrial amp errors

### LINKED ISSUE #11710: `trial -jN ...` fails to report some large test results

```
    test_renew_lease ... Traceback (most recent call last):
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/internet/defer.py", line 1697, in _inlineCallbacks
    result = context.run(gen.send, result)
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/trial/_dist/workerreporter.py", line 261, in addErrorFallible
    await addError(
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/trial/_dist/workerreporter.py", line 48, in addError
    errorStreamId = await stream(amp, chunk(error, 2 ** 16 - 1))
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/trial/_dist/stream.py", line 98, in stream
    await amp.callRemote(StreamWrite, streamId=streamId, data=oneChunk)
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 936, in callRemote
    return co._doCommand(self)
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 1951, in _doCommand
    d = proto._sendBoxCommand(
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 869, in _sendBoxCommand
    box._sendTo(self.boxSender)
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 698, in _sendTo
    proto.sendBox(self)
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 2336, in sendBox
    self.transport.write(box.serialize())
  File "/Users/distiller/project/venv/lib/python3.10/site-packages/twisted/protocols/amp.py", line 677, in serialize
    raise TooLong(False, True, v, k)
twisted.protocols.amp.TooLong:
```

A test that produces more than 2 ** 16 unicode characters of output with any character that encodes to more than 1 byte will trigger the behavior because disttrial measures unicode length but AMP enforces the length limit against bytes.

exarkun:
## Scope and purpose

Fixes #11710

### DIFF
--- a/src/twisted/trial/_dist/workerreporter.py
+++ b/src/twisted/trial/_dist/workerreporter.py
@@ -45,8 +45,8 @@ async def addError(
     :param frames: The lines of the traceback associated with the error.
     """
 
-    errorStreamId = await stream(amp, chunk(error, 2 ** 16 - 1))
-    framesStreamId = await stream(amp, iter(frames))
+    errorStreamId = await stream(amp, chunk(error.encode("utf-8"), 2 ** 16 - 1))
+    framesStreamId = await stream(amp, (frame.encode("utf-8") for frame in frames))
 
     await amp.callRemote(
         managercommands.AddError,
@@ -70,8 +70,8 @@ async def addFailure(
     :param fail: The string representation of the failure.
     :param frames: The lines of the traceback associated with the error.
     """
-    failStreamId = await stream(amp, chunk(fail, 2 ** 16 - 1))
-    framesStreamId = await stream(amp, iter(frames))
+    failStreamId = await stream(amp, chunk(fail.encode("utf-8"), 2 ** 16 - 1))
+    framesStreamId = await stream(amp, (frame.encode("utf-8") for frame in frames))
 
     await amp.callRemote(
         managercommands.AddFailure,
@@ -91,7 +91,7 @@ async def addExpectedFailure(amp: AMP, testName: str, error: str, todo: str) -> 
     :param error: The string representation of the expected failure.
     :param todo: The string description of the expectation.
     """
-    errorStreamId = await stream(amp, chunk(error, 2 ** 16 - 1))
+    errorStreamId = await stream(amp, chunk(error.encode("utf-8"), 2 ** 16 - 1))
 
     await amp.callRemote(
         managercommands.AddExpectedFailure,
--- a/src/twisted/trial/_dist/workerreporter.py
+++ b/src/twisted/trial/_dist/workerreporter.py
@@ -17,7 +17,7 @@ from attrs import Factory, define
 from typing_extensions import Literal
 
 from twisted.internet.defer import Deferred, maybeDeferred
-from twisted.protocols.amp import AMP
+from twisted.protocols.amp import AMP, MAX_VALUE_LENGTH
 from twisted.python.failure import Failure
 from twisted.python.reflect import qual
 from twisted.trial._dist import managercommands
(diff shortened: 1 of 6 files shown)


src/twisted/trial/_dist/workerreporter.py
```diff
@@ -45,8 +45,8 @@ async def addError(
     :param frames: The lines of the traceback associated with the error.
     """
 
-    errorStreamId = await stream(amp, chunk(error, 2 ** 16 - 1))
-    framesStreamId = await stream(amp, iter(frames))
+    errorStreamId = await stream(amp, chunk(error.encode("utf-8"), 2 ** 16 - 1))
+    framesStreamId = await stream(amp, (frame.encode("utf-8") for frame in frames))
```
adiroiban, line 49: I guess that we assume that we don't have super long lines :)

----------

On a second review: the "frames" which stands for "traceback lines" might not be the best name.
But we can live with it.

src/twisted/trial/_dist/workerreporter.py
```diff
@@ -45,8 +45,8 @@ async def addError(
     :param frames: The lines of the traceback associated with the error.
     """
 
-    errorStreamId = await stream(amp, chunk(error, 2 ** 16 - 1))
-    framesStreamId = await stream(amp, iter(frames))
+    errorStreamId = await stream(amp, chunk(error.encode("utf-8"), 2 ** 16 - 1))
```
adiroiban: `2 ** 16 - 1` is fine for now.
I guess that in an ideal world, this would be a constant that can be shared between the producer and consumer.

adiroiban (approved):
All good. Many thanks for the fix. Much appreciated.

src/twisted/trial/_dist/workerreporter.py
exarkun: Yea....... The way traceback lines are chunked is not great :/  I was a bit lazy here and didn't try to improve it much.  I guess time will tell if it is a safe bet or not.

I agree that `frames` is not a great name here.  Even if we ignore that they're strings and not actual frame objects, they're not even all strings representing frames :/  We are somewhat protected by the fact that they're not even exactly the lines of a traceback either - instead it's "code object names", "filenames", and "line numbers" - and I guess the parent process resolves that info to actual source lines when necessary.  These things seem even less likely than source lines to exceed the AMP limit.

Still, someone should probably give this area of the protocol some extra attention.  `frames` is one gross corner but it isn't the only one... :(

src/twisted/trial/_dist/workerreporter.py
exarkun: I'm not sure if I follow.  The consumer can always figure this out by looking at the length of the chunks it receives. :)

adiroiban: True. Thanks for the followup.

I was thinking to have the constant between this code and the AMP protocol.
As this whole chunking work is due to an AMP protocol limitation.

exarkun: Ahh yes I see.  I think AMP does have a constant for this,

Issue #11717, opened 2022-10-22
`twisted.web.resource.IResource` doesn't define `server` attribute used by `Resource`


twm:
**Describe the incorrect behavior you saw**

The `twisted.web.resource.Resource.putChild()` implementation touches the `server` attribute of its `child` argument, which is nominally an `IResource`:

https://github.com/twisted/twisted/blob/4b62da085e6a7b8e4b1231e8cdf16e0d3439591d/src/twisted/web/resource.py#L207-L233

`IResource` doesn't define a `server` attribute, much less a mutable one.

This implies that an `IResource` is associated with a single `server`, which isn't mentioned in `IResource` either.

**Describe how to cause this behavior**

Type annotate the method like `def putChild(self, path: bytes, child: IResource) -> None`.

**Describe the correct behavior you'd like to see**

Options:

1. Define `IResource.server` and describe this behavior in `IResource.putChild()`
2. Annotate `Resource.putChild` as accepting `IResourceAsItExistsInTheRealWorld` (see also: the many attributes that `DummyResource` has to define in order to function).
3. Something more complicated in the name of backward compatibility.
4. Rewrite Twisted Web. Why not give it a try?

adiroiban:
I don't think that rewriting  Twisted web is an option... I remember there was twisted.web2 and it failed... and by now, we are in the web3.0 age :)

I would say that for not option 1 is the first step.

For the future, we can create a new `twisted.web.resource.ModernResource` that should have a better URL traversing and support for 100-continue and receiving the content as a stream.

PR #11718, opened 2022-10-22
#8985: Raise TypeError for `Resource.putChild(str)`

### LINKED ISSUE #8985: twisted.web.Resource.putChild works unexpectedly with Unicode path
|[derwolfe's avatar](https://github.com/derwolfe)| @derwolfe reported|
|-|-|
|Trac ID|trac#8985|
|Type|defect|
|Created|2017-01-10 19:17:12Z|

IResource [https://github.com/twisted/twisted/blob/twisted-16.5.0/src/twisted/web/resource.py#L61-L71] states that the `path` param should be `bytes`. However, if a user passes in a unicode string, no warning is displayed, nor is any exception thrown. On python3, this is especially problematic, as new users might be following the documentation how-to for twisted web [https://twistedmatrix.com/documents/current/web/howto/using-twistedweb.html#resource-trees] and not understand why their unicode path isn't returning the resource that they would expect.

Currently, Resource's implementation [https://github.com/twisted/twisted/blob/twisted-16.5.0/src/twisted/web/resource.py#L209-L220] of `putChild` happily accepts unicode or bytes. 

For example of how this fails, the following twisted site will return `hi` when curl'ed (`curl http://localhost:8888`) and running on python2, but will return a 404 when run with python3.

```
from twisted.web.server import Site
from twisted.web.resource import Resource


class Page(Resource):

        return b"hi"


if __name__ == '__main__':

    root = Resource()
    root.putChild("", Page()) # on python3, "" should be b""

    factory = Site(root)

    reactor.listenTCP(8888, factory)
    reactor.run()
```

```
trac-id__8985 8985
type__defect defect
reporter__derwolfe derwolfe
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__web web
keywords__None None
time__1484075832128825 1484075832128825
changetime__1511669305479914 1511669305479914
version__None None
owner__rodrigc rodrigc
cc__derwolfe
```
</details>

twm:
## Scope and purpose

This makes `twisted.web.resource.Resource.putChild()` raise `TypeError` when called with anything but `bytes`, and adds type annotations indicating the same.

The type annotations snowballed a bit and I had to add more annotations in a few other locations to get MyPy clean. I discovered some buggy tests and a problem in `twisted.web.distrib` analogous to the one I fixed in #1678. There is a new ignore pointing at #11717.

Fixes #8985.

### DIFF
--- a/src/twisted/web/test/test_distrib.py
+++ b/src/twisted/web/test/test_distrib.py
@@ -29,7 +29,7 @@ from twisted.trial.unittest import TestCase
 from twisted.web import client, distrib, resource, server, static
 from twisted.web.http_headers import Headers
 from twisted.web.test._util import _render
-from twisted.web.test.test_web import DummyChannel, DummyRequest
+from twisted.web.test.requesthelper import DummyChannel, DummyRequest
 
 
 class MySite(server.Site):
@@ -395,7 +395,7 @@ class UserDirectoryTests(TestCase):
         """
         self.assertTrue(verifyObject(resource.IResource, self.directory))
 
-    def _404Test(self, name):
+    async def _404Test(self, name: bytes) -> None:
         """
         Verify that requesting the C{name} child of C{self.directory} results
         in a 404 response.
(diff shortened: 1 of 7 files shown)


twm:
please review

twm:
The Codecov diff coverage complaint appears to be nonsense.

.github/scripts/check-pr-text.py
```diff
@@ -41,7 +41,7 @@ def fail(message):
 if not pr_body:
     fail("Body for the PR not found. " "Maybe missing PR_BODY env var.")
 
-title_search = re.search(r"^(#\d+) .+", pr_title)
+title_search = re.search(r"^(#\d+):? .+", pr_title)
```
adiroiban: I guess  that this can be merged and we can look into this as part of the other PR.

src/twisted/web/test/test_distrib.py
```diff
@@ -395,36 +395,32 @@ def test_interface(self):
         """
         self.assertTrue(verifyObject(resource.IResource, self.directory))
 
-    def _404Test(self, name):
+    async def _404Test(self, name: bytes) -> None:
```
adiroiban, line 398: I am not happy with this helper... but is already in trunk, so it should not be a blocker.

The `_404Test` name doesn't disclose too much what this call is all about.

I think that having it start with `check` or `assert` should make it easier to understand what is the purpose of this method.

And all this, without having to guess that `self.directory` is automatically used.


```suggestion
    async def check404Render(self, target, name: bytes) -> None:
```

adiroiban (approved):
Looks good. Thanks!

Only minor comments.

src/twisted/web/test/test_distrib.py
twm: IMO this doesn't matter &mdash; this is not public API, it's used in two tests immediately below. Anyone interested in it can read the docstring (which mentions `self.directory`) or the entire 5-line implementation.

twm:
Thank you for the review @adiroiban! I've rebased and will land this when it goes green.

PR #11722, opened 2022-10-24
#11721: Ignore .hypothesis

### LINKED ISSUE #11721: .hypothesis should be in .gitignore

`git diff` shows me an untracked `.hypothesis` directory.

I ran the tests.

`.hypothesis` [shouldn't be checked in](https://hypothesis.readthedocs.io/en/latest/database.html#the-hypothesis-example-database), so it should be ignored by `git`.

twm:
## Scope and purpose

Fixes #11721. See there for an exhaustive discussion.


twm:
please review

.github/scripts/check-pr-text.py
```diff
@@ -41,7 +41,7 @@ def fail(message):
 if not pr_body:
     fail("Body for the PR not found. " "Maybe missing PR_BODY env var.")
 
-title_search = re.search(r"^(#\d+) .+", pr_title)
+title_search = re.search(r"^(#\d+):? .+", pr_title)
```
adiroiban, line 44: I think that we should also update the PR template to document that multiple title formats are allowed.

AFAIK, the PR template is the only documenting for our PR  naming convention.

adiroiban (approved):
Thanks. All good. 
Only a small change regarding the drive-by change.

.github/scripts/check-pr-text.py
twm: :shrug: I think that a bot should be maintaining this invariant, not human beings.

PR #11724, opened 2022-10-24
#11723 Update coding standard encoding to UTF-8

### LINKED ISSUE #11723: Update coding standard encoding to UTF-8
The coding standard says:

> The encoding must always be ASCII, so no coding cookie is necessary.

This is out of date. Since Python 3 hit (per [PEP 3120](https://peps.python.org/pep-3120/)) UTF-8 has been the default encoding, so no noise at the top of the file is not necessary to use it.

This came up [in a review](https://github.com/twisted/twisted/pull/1699/files#diff-9f8847086d32480a8f708091479f078fb8585de531db01d55e475810be58f0f8R2617).

twm:
Fixes #11723. See the ticket for background.

### DIFF
--- a/docs/development/coding-standard.rst
+++ b/docs/development/coding-standard.rst
@@ -136,7 +136,7 @@ In case of local names conflicts due to import, use the ``as`` syntax, for examp
     from twisted.trial import util as trial_util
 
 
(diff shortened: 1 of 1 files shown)


twm:
please review

adiroiban (approved):
Thanks!

PR #11727, opened 2022-10-25
Revert "#11655 Update package metadata"

glyph:
Reverts twisted/twisted#11656 due to @alex's bug report

Issue #11655, opened 2022-09-11
Update package metadata


ofek:
**Is your feature request related to a problem? Please describe.**

The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). As a result, the execution of `setup.py` files is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

**Describe the solution you'd like**

Use modern packaging standards

**Describe alternatives you've considered**

Doing nothing is an option for now

adiroiban:
Reopening this as the PR will be reverted.

PR #11728, opened 2022-10-25
#11655 Update package metadata take 2

glyph:
Reverts twisted/twisted#11727


glyph:
@ofek if you could stack your commits here that would be great :)

ofek:
https://github.com/twisted/twisted/pull/11730

glyph:
Work continues in #11730.

PR #11730, opened 2022-10-26
#11655 Remove use of pkg_resources

### LINKED ISSUE #11655: Update package metadata

ofek:
## Scope and purpose

Fixes #11655

See https://github.com/twisted/twisted/pull/11728#issuecomment-1291262638


ofek:
https://docs.python.org/3.9/library/importlib.metadata.html

src/twisted/trial/__main__.py
```diff
@@ -4,6 +4,18 @@
...
+    ep_name = "trial"
+    dist = distribution("twisted")
+    for ep in dist.entry_points:
+        if ep.group == "console_scripts" and ep.name == ep_name:
+            trial = ep
+            break
+    else:
+        raise OSError(f"No console_scripts entry point found for: {ep_name}")
+
+    sys.exit(trial.load()())
```
glyph: If we're not going to test the error message, my preference would be for something briefer that won't cause issues with coverage, like this:

```suggestion
    sys.exit(
        next(
            ep
            for ep in distribution("twisted").entry_points
            if (ep.group, ep.name) == ("console_scripts", "trial")
        ).load()()
    )
```

(It's not clear to me how we *would* test this message though, unless we factored it out into a function…)

glyph (approved):
Looks fine, but I'm not sure it's going to pass coverage. I'll switch the merge base to trunk now that it's reviewed, and hopefully that will run through CI

src/twisted/trial/__main__.py
```diff
@@ -4,6 +4,15 @@
 if __name__ == "__main__":
     import sys
 
-    from pkg_resources import load_entry_point
+    if sys.version_info >= (3, 8):
```
glyph: Sigh. Okay. Codecov is too smart for our trickery.

Which actually raises the question: why are we doing this faffing with packaging metadata, anyway?  We used to do this because we wanted to be as similar as possible to the stubs that `pip install` generated, but looking at a modern virtualenv, I see the stubs look like this:

```
#!/Users/glyph/.virtualenvs/Twisted/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from twisted.scripts.trial import run
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(run())
```

So that implies that this should instead simply be:

```
    from twisted.scripts.trial import run
    sys.exit(run())
```

Is there any reason not to do that?

(Suggestion in multiple parts below)

src/twisted/trial/__main__.py
```diff
@@ -4,6 +4,15 @@
...
+        from importlib.metadata import distribution
+    else:
+        from importlib_metadata import distribution
```
glyph: Part 1: no need for this package at all
```suggestion
```

src/twisted/trial/__main__.py
```diff
@@ -4,6 +4,15 @@
...
 
-    sys.exit(load_entry_point("Twisted", "console_scripts", "trial")())
+    sys.exit(
+        next(
+            ep
+            for ep in distribution("twisted").entry_points
+            if (ep.group, ep.name) == ("console_scripts", "trial")
+        ).load()()
+    )
```
glyph: Part 2: let's not overcomplicate things
```suggestion
    from twisted.scripts.trial import run
    sys.exit(run())
```

glyph (changes requested):
Oops. Whatever we do here, `twisted/__main__.py` also needs to be fixed in the same way.

src/twisted/trial/__main__.py
ofek: very true

src/twisted/__main__.py
```diff
@@ -8,7 +8,7 @@
 
 import sys
 
-from pkg_resources import load_entry_point
+from twisted.application.twist._twist import Twist
```
glyph:
```suggestion
```

src/twisted/__main__.py
```diff
@@ -8,7 +8,7 @@
...
 
 if __name__ == "__main__":
-    sys.exit(load_entry_point("Twisted", "console_scripts", "twist")())
+    sys.exit(Twist.main())
```
glyph, line 14:
```suggestion
    from twisted.application.twist._twist import Twist
    sys.exit(Twist.main())
```

glyph (approved):
Made a suggestion to keep the script import inside the `if __name__` block, but otherwise, if this passes tests let's land.

ofek:
rebased

adiroiban:
Many thanks @ofek for the fix and @glyph  for the review.

I have merged this on behalf of @glyph  :)

glyph:
Thanks @adiroiban !

PR #11734, opened 2022-10-28
#10343 - run tests w/ python 3.11

### LINKED ISSUE #10343: Get support for Python 3.11
|-|-|
|Trac ID|trac#10343|
|Type|enhancement|
|Created|2022-05-11 20:27:34Z|

Some info at https://github.com/twisted/twisted/pull/1723#issuecomment-1107654772 from Colin

BTW, I haven't attempted to add test runs for 3.11 yet because there are some other problems.  We need a Cython release with cython/cython[#4667](https://github.com/twisted/twisted/issues/4667), an Automat release with glyph/automat[#7453](https://github.com/twisted/twisted/issues/7453), there are some weird `builtins.TypeError: __dict__ must be set to a dictionary, not a '_Dereference'` test failures that I don't understand, and there are a few fairly trivial test failures due to https://docs.python.org/3.11/whatsnew/3.11.html#enhanced-error-locations-in-tracebacks.  But it's generally looking closer, and this on its own is enough to let me get the testtools test suite passing.

```
trac-id__10343 10343
type__enhancement enhancement
reporter__adiroiban adiroiban
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1652300854237229 1652300854237229
changetime__1652425385864237 1652425385864237
version__None None
owner__None None
cc__mgorny
```
</details>

eevelweezel:
## Scope and purpose

Fixes #10343

Add support for running tests w/ Python 3.11 on Windows, MacOS, and Linux.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -122,11 +122,19 @@ jobs:
           # Just Python 3.9 with default settings.
           - python-version: '3.9'
 
+          # Just Python 3.11 with default settings.
+          - python-version: '3.11'
+
           # Newest macOS and newest Python supported versions.
           - python-version: '3.10'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
 
+          # Newest macOS and newest Python supported versions.
+          - python-version: '3.11'
+            runs-on: 'macos-12'
+            job-name: 'macos-12-default-tests'
+          #
           # Windows, minimum Python version with select reactor.
           - python-version: '3.7'
             runs-on: 'windows-2022'
@@ -139,6 +147,14 @@ jobs:
 
           # Windows, newest Python supported version with iocp reactor.
           - python-version: '3.10'
+            runs-on: 'windows-2022'
+            tox-env: 'alldeps-withcov-windows'
+            job-name: 'win-default-tests-iocp'
+            # Distributed trial is not yet suported on Windows.
+            trial-args: '--reactor=iocp'
+
+          # Windows, newest Python supported version with iocp reactor.
+          - python-version: '3.11'
             runs-on: 'windows-2022'
             tox-env: 'alldeps-withcov-windows'
             job-name: 'win-default-tests-iocp'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -130,11 +130,6 @@ jobs:
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
 
-          # Newest macOS and newest Python supported versions.
-          - python-version: '3.11'
-            runs-on: 'macos-12'
-            job-name: 'macos-12-default-tests'
-          #
           # Windows, minimum Python version with select reactor.
           - python-version: '3.7'
             runs-on: 'windows-2022'
@@ -147,14 +142,6 @@ jobs:
 
           # Windows, newest Python supported version with iocp reactor.
           - python-version: '3.10'
-            runs-on: 'windows-2022'
-            tox-env: 'alldeps-withcov-windows'
-            job-name: 'win-default-tests-iocp'
-            # Distributed trial is not yet suported on Windows.
-            trial-args: '--reactor=iocp'
-
-          # Windows, newest Python supported version with iocp reactor.
-          - python-version: '3.11'
             runs-on: 'windows-2022'
             tox-env: 'alldeps-withcov-windows'
             job-name: 'win-default-tests-iocp'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -126,7 +126,7 @@ jobs:
           - python-version: '3.11'
 
           # Newest macOS and newest Python supported versions.
-          - python-version: '3.10'
+          - python-version: '3.11'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
 
@@ -141,7 +141,7 @@ jobs:
             trial-args: '--reactor=select'
 
           # Windows, newest Python supported version with iocp reactor.
-          - python-version: '3.10'
+          - python-version: '3.11'
             runs-on: 'windows-2022'
             tox-env: 'alldeps-withcov-windows'
             job-name: 'win-default-tests-iocp'
--- a/src/twisted/newsfragments/10343.misc
+++ b/src/twisted/newsfragments/10343.misc
@@ -1,1 +1,0 @@
-Enable Python 3.11 for CI and fix the remaining broken tests.
--- a/src/twisted/newsfragments/10343.misc
+++ b/src/twisted/newsfragments/10343.misc
@@ -1,0 +1,1 @@
+Enable Python 3.11 for CI and fix the remaining broken tests.
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -714,7 +714,7 @@ class FlattenerErrorTests(SynchronousTestCase):
                 flags=re.MULTILINE,
             ),
         )
-        self.assertTrue('RuntimeError: example' in str(failure.value))
+        self.assertTrue("RuntimeError: example" in str(failure.value))
         # The original exception is unmodified and will be logged separately if
         # unhandled.
         self.failureResultOf(failing, RuntimeError)
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -706,9 +706,10 @@ class FlattenerErrorTests(SynchronousTestCase):
                     Exception while flattening:
                       \\[<unrenderable>\\]
                       <unrenderable>
-                      .*
+                      <Deferred at .* current result: <twisted.python.failure.Failure builtins.RuntimeError: example>>
                       File ".*", line \\d*, in _flattenTree
                         element = await element
+                                  .*
                     RuntimeError: example
                     """
                 ),
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -708,9 +708,7 @@ class FlattenerErrorTests(SynchronousTestCase):
                       <unrenderable>
                       <Deferred at .* current result: <twisted.python.failure.Failure builtins.RuntimeError: example>>
                       File ".*", line \\d*, in _flattenTree
-                        element = await element
-                                  .*
-                    RuntimeError: example
+                        element = await element.*
                     """
                 ),
                 flags=re.MULTILINE,
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -714,7 +714,7 @@ class FlattenerErrorTests(SynchronousTestCase):
                 flags=re.MULTILINE,
             ),
         )
-        self.assertTrue("RuntimeError: example" in str(failure.value))
+        self.assertIn("RuntimeError: example", str(failure.value))
         # The original exception is unmodified and will be logged separately if
         # unhandled.
         self.failureResultOf(failing, RuntimeError)
--- a/src/twisted/web/test/test_flatten.py
+++ b/src/twisted/web/test/test_flatten.py
@@ -714,6 +714,7 @@ class FlattenerErrorTests(SynchronousTestCase):
                 flags=re.MULTILINE,
             ),
         )
+        self.assertTrue('RuntimeError: example' in str(failure.value))
         # The original exception is unmodified and will be logged separately if
         # unhandled.
         self.failureResultOf(failing, RuntimeError)
--- a/src/twisted/newsfragments/10343.feature
+++ b/src/twisted/newsfragments/10343.feature
@@ -1,0 +1,1 @@
+Twisted now officially supports Python 3.11.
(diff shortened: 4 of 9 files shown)


eevelweezel:
I'd appreciate some guidance. This enables tests in Python 3.11, but there are multiple places where the test suite fails under that version.

adiroiban:
Many thanks for creating the PR and your help with 3.11

Yes. I think that is best to create separate tickets and try to fix each error type/class in separate PRs.

I think that we will also need to wait for @glyph  to do a new release for automat and then have a PR in twisted that uses the newer version.

glyph:
New automat release is out.

adiroiban:
I thinks is best to start small and don't use 3.11 as default.

I see pydoctor failing...and we might have other development-only dependencies failing.

The first step is to get at least a subset of Twisted test on 3.11

We don't have to move all the build tools to 3.11, yet.

adiroiban:
I see `twisted/persisted/aot.py` and `twisted.spread` failing.

If they need only 2 lines of code to fix 3.11 support, we can have the fix in this PR.

Otherwise, I think is best to fix the changes in separate tickets.

eevelweezel:
> I see `twisted/persisted/aot.py` and `twisted.spread` failing.
> 
> If they need only 2 lines of code to fix 3.11 support, we can have the fix in this PR.
> 
> Otherwise, I think is best to fix the changes in separate tickets.

I've reverted the commit making 3.11 the default python & fixed the tests for twisted.spread and twisted.persisted.

glyph:
Looks like the test fixes have not worked?

adiroiban:
> FAILED (skips=514, failures=3, successes=10598)

Not bad. This is good progress. Thanks!

eevelweezel:
I looked into the 3 remaining failures. Those also needed only 1-2 lines to fix, so I addressed those here as well.

eevelweezel:
please review

glyph:
@eevelweezel Thank you so much for doing this critical maintenance!

src/twisted/newsfragments/10343.misc
```diff
@@ -0,0 +1 @@
+Enable Python 3.11 for CI and fix the remaining broken tests.
```
glyph: This is the user-facing changelog entry, so it should read as a thing that users are interested in.  This should be a `.feature`, and it should be more like this:

```suggestion
Twisted now officially supports Python 3.11.
```

glyph: (Users don't care about the configuration details of our CI or "broken tests"; they care about whether or not we stand behind using Twisted with a particular Python version.)

src/twisted/web/test/test_flatten.py
```diff
@@ -706,10 +706,9 @@ def render(self, request: Optional[IRequest]) -> Flattenable:
                     Exception while flattening:
                       \\[<unrenderable>\\]
                       <unrenderable>
-                      .*
+                      <Deferred at .* current result: <twisted.python.failure.Failure builtins.RuntimeError: example>>
                       File ".*", line \\d*, in _flattenTree
-                        element = await element
-                    RuntimeError: example
+                        element = await element.*
```
glyph, line 711: I wouldn't block on this since it seems fairly minor, but isn't the `RuntimeError` kind of an important part of this report to test? I don't fully appreciate the consequences of this test change.

glyph (approved):
LGTM

I'd really like to move that .`misc` to a `feature` so if anyone has time to do that and land that would be great.

.github/workflows/test.yaml
```diff
@@ -122,11 +122,19 @@ jobs:
           # Just Python 3.9 with default settings.
           - python-version: '3.9'
 
+          # Just Python 3.11 with default settings.
+          - python-version: '3.11'
+
           # Newest macOS and newest Python supported versions.
```
adiroiban, line 128: Please update this comment... 3.10 is no longer the newes Python.

The idea with this comment want to highlight that we only run the tests on a single macOS environemnt.

And that environment is newest macos supported by GitHub  + newest Python supported by Twisted.

In the past we used to run tests on macOS-11 ...but the GitHub environmet was broken , so we stopped running those tests.

Maybe, we can also just remote these tests, and only run python 3.11 on macOS.

I don't think that we need 2 macOS jobs.

.github/workflows/test.yaml
```diff
@@ -145,6 +153,14 @@ jobs:
             # Distributed trial is not yet suported on Windows.
             trial-args: '--reactor=iocp'
 
+          # Windows, newest Python supported version with iocp reactor.
```
adiroiban: Please see the comment on line 148

The comment is no longer valid, as 3.10 is no longer the newest Python.

The idea is that instead of adding a new job for 3.11, we will just update the "newest" job 

So my suggestion is to delete this job, and at line 149, replace 3.10 with 3.11

makes sense?

src/twisted/newsfragments/10343.misc
adiroiban: +1 for having this as a feature and using the text suggested by Glyph.

src/twisted/web/test/test_flatten.py
adiroiban: The description of this test is

When flattening a renderer which raises an exception asynchronously, the error is reported when it occurs.

With this change, it looks like the test is ignoring the "when it occurs" part.

adiroiban (changes requested):
Great job. Thanks.

I am requesting changes due to:

* release notes
* test_flatten

Not adding new jobs for macOS and Windows, would be nice.
we already have a big list of jobs.

src/twisted/web/test/test_flatten.py
adiroiban: I have checked the previous errors, and it looks like the implementation works

The Runtime is also listed last.


Regex didn't match: 'Exception while flattening:\n  \\[<unrenderable>\\]\n  <unrenderable>\n  .*\n  File ".*", line \\d*, in _flattenTree\n    element = await element\nRuntimeError: example\n'

```
 Exception while flattening:
  [<unrenderable>]
  <unrenderable>
  <Deferred at 0x7fded34a2150 current result: <twisted.python.failure.Failure builtins.RuntimeError: example>>
  File "/home/runner/work/twisted/twisted/.tox/alldeps-withcov-posix/lib/python3.11/site-packages/twisted/web/_flatten.py", line 424, in _flattenTree
    element = await element
              ^^^^^^^^^^^^^
RuntimeError: example

```

glyph: Aah, so the thing we're skipping is the `^^^^`. It seems like we could then do something like this:
```suggestion
                        element = await element.*
                    RuntimeError: example
```
which will skip the line with the carets but preserve the exception/message checking.

glyph:
I left off the comments about "newest" just because this is a general cleanup the file needs and nothing to do with 3.11 support.  Anywhere we say something like "newest" or "currently supported", we should really be saying "as of" and a date, but I don't want to put the burden of that kind of a fix onto this PR which is really doing something else.  I realize that this makes things even more out of date but each PR should fix one thing at a time and "3.11 support" and "none of the comments refer to the current accurate state of our CI" are not really related :).

src/twisted/newsfragments/10343.misc
eevelweezel: Updated & moved this to a feature.

src/twisted/web/test/test_flatten.py
eevelweezel: I still couldn't get this to match, but I added an assert that the failure.value contains "RuntimeError: example"

.github/workflows/test.yaml
eevelweezel: I've removed the extra Windows & MacOS python 3.11 jobs, but now I'm confused. Do we want to update this so the "newest" jobs run on 3.11 and make 3.11 the default python?  I had done this previously, but was asked to revert it.

adiroiban: Sorry for the confusion.

I know that at some point you have updated 3.11 as the default Python.
This means that our Sphinx and pydoctor jobs and other non-test releated jobs were exectued on 3.11... and some were failing.
So my suggestion was to go with baby steps, and only use 3.11 for the python tests.

Does it make sense?

.github/workflows/test.yaml
```diff
@@ -122,6 +122,9 @@ jobs:
           # Just Python 3.9 with default settings.
           - python-version: '3.9'
 
+          # Just Python 3.11 with default settings.
+          - python-version: '3.11'
+
           # Newest macOS and newest Python supported versions.
           - python-version: '3.10'
```
adiroiban: My suggestion was to use 3.11 here 

and also use 3.11 for Windows on line 144

```suggestion
          - python-version: '3.11'
```

In this way, we run Windows and macOS python trial tests with latest Python version

src/twisted/web/test/test_flatten.py
```diff
@@ -706,15 +706,15 @@ def render(self, request: Optional[IRequest]) -> Flattenable:
...
```
adiroiban, line 711: Maybe something like this ?

```suggestion
                        element = await element.*
                        RuntimeError: example
```

src/twisted/web/test/test_flatten.py
```diff
@@ -706,15 +706,15 @@ def render(self, request: Optional[IRequest]) -> Flattenable:
...
                     """
                 ),
                 flags=re.MULTILINE,
             ),
         )
+        self.assertTrue("RuntimeError: example" in str(failure.value))
```
adiroiban: I don't think that this is ok.

The previous check already has `builtins.RuntimeError: example>>` which should be the same thing.

So I don't see any extra value / info / assertion here.

I am not 100% what is the expected output, but as mentioned by Glyph ... getting the same expected regex on 3.11 and other python version would be hard.

As it looks like in 3.11 the default expception rending was changed.

src/twisted/web/test/test_flatten.py
glyph: That's exactly what I suggested.  See resolved conversation above.  I resolved the conversation because I thought the slightly less precise `assert...in...` was fine.

src/twisted/web/test/test_flatten.py
glyph: Why, exactly, is this not OK?

What we are trying to test for here is that the exception type is relayed somewhere in the message to the user and we have not, in our custom traceback construction logic, accidentally elided or changed it.  The previous check does not have RuntimeError in it any more, it was removed, due to the traceback formatting changes.  Ostensibly we *could* check it with an individual assertion, sure, but is it worth any additional effort to figure out exactly what precise logic in Python's traceback formatting changed to offset this slightly differently than it was before?  We're testing Twisted's traceback construction, not Python's traceback formatting.

glyph: I guess as long as we're spending so long discussing this line, I could point out that the new assertion loses information, so it would be better to change it to use the unittest idiom for this type of assertion rather than `assertTrue` (which is almost always the wrong choice and gives no information upon failure)

```suggestion
        self.assertIn("RuntimeError: example", str(failure.value))
```

adiroiban: The check is not ok as I think that the previous check already does the same thing.

The assertion can be left, it there is no extra coverage for this assertion.

glyph: Read the diff more carefully.  The previous check does not check that any more.

glyph: Specifically, from the diff view: 
image

eevelweezel: I agree that this should be assertIn vs assertTrue, updated.

.github/workflows/test.yaml
eevelweezel: Ah, makes sense.  Updated.

.github/workflows/test.yaml
eevelweezel: Yes, that makes sense. That should be updated w/ my latest commit.

src/twisted/web/test/test_flatten.py
```diff
@@ -706,15 +706,15 @@ def render(self, request: Optional[IRequest]) -> Flattenable:
...
                       File ".*", line \\d*, in _flattenTree
-                        element = await element
-                    RuntimeError: example
+                        element = await element.*
                     """
                 ),
                 flags=re.MULTILINE,
             ),
         )
+        self.assertIn("RuntimeError: example", str(failure.value))
```
adiroiban, line 717: Just one last call

To me, the previous assertion `assertRegex` already does what the `assertIn` does.

And `assertIn` doesn't assert the order  or the place where the error is raised.

```
self.assertRegex(
    str(failure.value),
    'BLA<twisted.python.failure.Failure builtins.RuntimeError: example>>BLA',
    )
```

adiroiban (approved):
Thanks. This can be merged.

I think that assertIn is not really needed as assertRegex already does the same job.

@glyph feel free to merge

Thanks you for your help.

Happy to see 3.11 up and running :)

src/twisted/web/test/test_flatten.py
glyph: I agree that, ideally, that's what we'd do. But as @eevelweezel put it above:

> I still couldn't get this to match

and I think we should be mindful of the amount of effort it takes to arrive at an ideal solution.

If anyone wants to submit a cleanup in trunk after this is landed I think that would be great.

PR #11743, opened 2022-11-04
#11742 Update GitHub Actions yaml to remove warnings.

### LINKED ISSUE #11742: Fix GitHub Actions warnings
When running the tests GitHub Actions generates 39 warnings 

see the end of the page here https://github.com/twisted/twisted/actions/runs/3360302830

Most of them is using newer actions or inter-job / inter-step communication methods.

Should be easy to fix...but needs time.

adiroiban:
## Scope and purpose

Fixes #11742

This tries to remove all the warnings generated while running GitHub Actions.

--------

As a drive-by the code that was supposed to update the "stable" branch so that we have a "stable" release on Read The docs was removed 

The code doesn't work and I don't think it is worth troubleshooting it.

https://github.com/twisted/twisted/actions/runs/3360096602/jobs/5568908454#step:2:26

## Contributor Checklist:


adiroiban:
Check that no workings are emitted for the CI jobs.

At the end of the page, there should be no warnings

https://github.com/twisted/twisted/actions/runs/3390453477


-------------

On trunk, there are warnings and you can see them here https://github.com/twisted/twisted/actions/runs/3360302830

adiroiban:
needs-review

.github/workflows/test.yaml
```diff
@@ -381,40 +381,6 @@ jobs:
...
-            echo '::endgroup::'
-
-            git log -1
-            git checkout -b $STABLE_BRANCH origin/$STABLE_BRANCH
-            git reset --hard $GITHUB_SHA
-            git push --force
-            echo "Done"
-          else
-            echo "Branch not updated for not stable releases: $GITUB_REF"
-          fi
```
glyph, line 414: I am confused — why isn't this necessary any more?

I don't know RTD intimately, but I don't understand how the other changes could impact this.

Not really a request for changes on the code itself, I don't want a big comment saying why something *isn't* there, but perhaps some missing context in the PR or issue description.

Thanks for these cleanups though, I would love to see less warning noise from GHA.

adiroiban: This code doesn't work.

It is supposed to run only for the latest stable release

Most of the time is skipped, like for this PR.

You can see it run for 22.10.0 release.

It was supposed to update this branch https://github.com/twisted/twisted/tree/stable

But you can see that it just doesn't recognize the stable release, and while it is executed, it skips the update.

https://github.com/twisted/twisted/actions/runs/3360096602/jobs/5568908454#step:2:26

It is hard to troubleshoot it with GitHub Actions

For the last releases, I was just manually updating the branch.

And I don't plan to troubleshoot that code... so is best to remove it.

adiroiban: @glyph what we can do to make Twisted "RTD" friendly , is that instead of using `twisted-1.2.3.` as the name for release tags, we just use `1.2.3`

In this way, RTD can detect which tag is "latest"

adiroiban: Thanks for the review :)

glyph: Ah, okay. If it's already broken then let's not keep dragging it around.

adiroiban: yes... it was broken for the last 2 releases...at least.
And troubleshooting is hard.

I have a ticket to update the release docs... in on my todo... but that is for another weekend.

glyph (approved):
OK. LGTM.

PR #11747, opened 2022-11-06
#11746 remove signalreactormixin

### LINKED ISSUE #11746: Remove `_SignalReactorMixin`
This is a small refactoring of (mostly) `twisted/internet/base.py` which I am pulling out of the branch for #11685.  The goal of #11685 is to make large pieces of reactor functionality available as independently usable pieces with good composition properties.  The branch there has gotten too large to manage.  This is one step towards making it small enough to be able to land.  Ultimately the work in #11685 goes much further than this change but this one step makes it easier to get there (even though this specific factoring will soon be obsolete).

`twisted.internet.base._SignalReactorMixin` implements some parts of `IReactorCore` for `ReactorBase`.  As a mixin, it does this in a complex way which requires doing MRO calculations to understand.  As a mixin, it also lacks first-class access to the `ReactorBase` type and so it is littered with usage of `cast` to convince mypy that the programmer *promises* not to use it wrong.  Also, *all reactors in Twisted* subclass both `ReactorBase` and `_SignalReactorMixin`.  That is, having this as a separate class is providing no utility that Twisted actually makes use of.  Since it is private, no code outside of Twisted can reasonably be using it either  - except good luck having a useful `ReactorBase` subclass _without_ using this mixin (you will have to re-implement signal handling logic yourself or you will go without basics like "ctrl-c stops the reactor").

The functionality `_SignalReactorMixin` offers should just be part of `ReactorBase`.  In the future (#11685) we can offer that functionality with some additional affordances for customization but just removing the mixin is a good step in that direction.

exarkun:
Fixes #11746

See ticket for details.

### DIFF
--- a/src/twisted/internet/base.py
+++ b/src/twisted/internet/base.py
@@ -629,6 +629,52 @@ class ReactorBase(PluggableResolverMixin):
             self._initThreads()
         self.installWaker()
 
+    # Signal handling pieces
+    _installSignalHandlers = False
+
+    def _handleSignals(self) -> None:
+        """
+        Install the signal handlers for the Twisted event loop.
+        """
+        try:
+            import signal
+        except ImportError:
+            log.msg(
+                "Warning: signal module unavailable -- "
+                "not installing signal handlers."
+            )
+            return
+
+        if signal.getsignal(signal.SIGINT) == signal.default_int_handler:
+            # only handle if there isn't already a handler, e.g. for Pdb.
+            signal.signal(signal.SIGINT, self.sigInt)
+        signal.signal(signal.SIGTERM, self.sigTerm)
+
+        # Catch Ctrl-Break in windows
+        SIGBREAK = getattr(signal, "SIGBREAK", None)
+        if SIGBREAK is not None:
+            signal.signal(SIGBREAK, self.sigBreak)
+
+    def run(self, installSignalHandlers: bool = True) -> None:
+        self.startRunning(installSignalHandlers=installSignalHandlers)
+        self.mainLoop()
+
+    def mainLoop(self) -> None:
+        while self._started:
+            try:
+                while self._started:
+                    # Advance simulation time in delayed event
+                    # processors.
+                    self.runUntilCurrent()
+                    t2 = self.timeout()
+                    t = self.running and t2
+                    self.doIteration(t)
+            except BaseException:
+                log.msg("Unexpected error in main loop.")
+                log.err()
+            else:
+                log.msg("Main loop terminated.")  # type:ignore[unreachable]
+
     # override in subclasses
 
     _lock = None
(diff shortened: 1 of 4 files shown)


src/twisted/internet/base.py
```diff
@@ -629,6 +629,52 @@ def __init__(self) -> None:
...
+    _installSignalHandlers = False
+
+    def _handleSignals(self) -> None:
+        """
+        Install the signal handlers for the Twisted event loop.
+        """
+        try:
+            import signal
+        except ImportError:
+            log.msg(
```
glyph, line 642: Is this configuration even possible to achieve any more?

glyph: Also, TODO for a later PR here, it's killing me that we have all these `log.msg` calls everywhere. 😬

src/twisted/internet/base.py
```diff
@@ -629,6 +629,52 @@ def __init__(self) -> None:
...
+                while self._started:
+                    # Advance simulation time in delayed event
+                    # processors.
+                    self.runUntilCurrent()
+                    t2 = self.timeout()
+                    t = self.running and t2
+                    self.doIteration(t)
+            except BaseException:
+                log.msg("Unexpected error in main loop.")
+                log.err()
```
glyph, line 674: Surprised that this didn't have coverage already, it seems like a pretty important case to cover, but given that the code is being moved and not added, let's ignore

src/twisted/internet/base.py
```diff
@@ -629,6 +629,52 @@ def __init__(self) -> None:
...
+            else:
+                log.msg("Main loop terminated.")  # type:ignore[unreachable]
```
glyph, line 676: Should we report this bug to mypy? it seems like a weird oversight.

glyph (approved):
Nice small incremental change here getting rid of an unnecessarily complexifying implementation detail, so I'll just land.

glyph:
New commit is just there to make the static-checks status happy. We really ought to fix it to be more like pre-commit where it just sticks in the '#' for you if forget it.

glyph:
(Or at least fix it so "rebuild" works so we don't have to resort to this. @adiroiban do you know why it doesn't work with re-running jobs, when the title has indeed been updated?)

adiroiban:
@glyph the check for PR title uses the GitHub Actions metadata.
This metadata is a snapshot taken together with the commit.

If you re-run a job, even if the PR description has a new revision, the job use executed using the old PR metadata.

---------

If this is a pain there are 2 options:

* Disable the check
* Update the check to use GitHub API to fetch the PR title and description o each (re)run.

We can disable the check now, and I can look to re-enable it later using GitHub API

glyph:
> If this is a pain there are 2 options:

It does just seem to be the most common type of misconfiguration, which I bump into on almost every code review or submission.

I'm not sure it's providing enough value to be worth the hassle, so perhaps disabling would be best for now.  The "fixes:" line does most of the work.

PR #11749, opened 2022-11-06
#11748 Remove PR text checks.

### LINKED ISSUE #11748: Remove check for PR links to issue IDs
> If this is a pain there are 2 options:

It does just seem to be the most common type of misconfiguration, which I bump into on almost every code review or submission.

I'm not sure it's providing enough value to be worth the hassle, so perhaps disabling would be best for now.  The "fixes:" line does most of the work.

_Originally posted by @glyph in https://github.com/twisted/twisted/issues/11747#issuecomment-1304901054_

adiroiban:
## Scope and purpose

Fixes #11748

This removed the GitHub Actions checks for the PR title and description.

The problem is that if you forget to create a PR with the right text, re-running the checks will not fix the issue....since the check is re-executed using the initial PR metadata.

The fix would be to use GitHub API to retrieve the PR metadata at each (re)run.

But this would link us more with the proprietary GitHub systems

So for now, just remove the check, and reviewers can manually check and ask the PR author to update the PR metadata.


exarkun (approved):
Thank you.

PR #11752, opened 2022-11-07
#11751 Remove ReactorBase._handleSignals

### LINKED ISSUE #11751: Remove ReactorBase._handleSignals
This is a small(ish) refactoring of how `ReactorBase` and related classes hook up their signal handling logic.  The actual signal handling logic is intended to be the same as it was before (as is all other public-facing behavior).  This has been pulled out of the branch for #11685.  The goal of #11685 is to make large pieces of reactor functionality available as independently usable pieces with good composition properties. The branch there has gotten too large to manage. This is one step towards making it small enough to be able to land. Ultimately the work in #11685 goes much further than this change but this one step makes it easier to get there (even though some of this specific factoring will soon be obsolete).

`ReactorBase._handleSignals` is called near `reactor.run` and is meant to install signal handlers for signals like SIGINT (stop the reactor).  `PosixReactorBase` overrides this to add SIGCHLD (notice child exit) handling.  `GlibReactorBase` overrides it to play nicely with glib. Also, wxreactor ... eh.

This refactoring puts all of the "application" logic (how to install the handlers and - to some extent - what to do when the signals arrive) into some new classes which implement a `SignalHandling` protocol.  `ReactorBase` is responsible for make sure one of these is created, installed, and cleaned up at the right times.  Each reactor is responsible for defining a factory that says what its correct signal handling logic is.

Presently the factory is defined as a private method of `ReactorBase` which I think is not amazing.  The other idea I can think of is to make it an `__init__` argument.  However, `PosixReactorBase` wants to compose its logic with that of `ReactorBase` which makes this a bit awkward.

Much of this logic will move around _anyway_ as more of #11685 lands so perhaps this is a fine incremental step and we'll figure out something better later?

exarkun:
Fixes #11751


exarkun:
The missing coverage is in cfreactor - which CI fails to test *at all* - and wxreactor ... which ... I just don't know what to say.

src/twisted/internet/_glibbase.py
```diff
@@ -56,11 +57,33 @@ class GlibWaker(_UnixWaker):
...
+    reactor main loop which makes our signal handling work with glib's signal
+    handling.
+    """
+    from gi import _ossighelper as signalGlue  # type: ignore[import]
+
+    patcher = MonkeyPatcher()
+    patcher.addPatch(signalGlue, "_wakeup_fd_is_active", True)
+    return patcher
+
+
```
glyph, line 86: Beautiful. An excellent refactoring of my attempt. No notes.

src/twisted/internet/_glibbase.py
```diff
@@ -126,14 +150,18 @@ def __init__(self, glib_module: Any, gtk_module: Any, useGtk: bool = False) -> N
         self._crash = _loopQuitter(self._glib.idle_add, self.loop.quit)
         self._run = self.loop.run
 
-    def _handleSignals(self):
+    def _reallyStartRunning(self):
```
glyph, line 168: A docstring would be nice here, given the… somewhat unclear distribution of responsibilities in this squirrely mess we are cleaning up

src/twisted/internet/_glibbase.py
```diff
@@ -126,14 +150,18 @@ def __init__(self, glib_module: Any, gtk_module: Any, useGtk: bool = False) -> N
...
         # First, install SIGINT and friends:
-        super()._handleSignals()
+        super()._reallyStartRunning()
         # Next, since certain versions of gtk will clobber our signal handler,
         # set all signal handlers again after the event loop has started to
-        # ensure they're *really* set. We don't call this twice so we don't
-        # leak file descriptors created in the SIGCHLD initialization:
-        self.callLater(0, posixbase.PosixReactorBase._handleSignals, self)
+        # ensure they're *really* set.
```
glyph, line 178: Hmm. Is this still true? I'd love to get a number on "certain versions". Mainly i'm wondering if this is effectively a repeat of similar logic as set up by `_signalGlue`.

glyph: (Idle curiosity here, not necessary to address before landing.)

src/twisted/internet/_signals.py
```diff
@@ -18,37 +18,45 @@
...
 from twisted.python.runtime import platformType
 
 if platformType == "posix":
     from . import fdesc, process
 
+SignalHandler: TypeAlias = Callable[[int, Optional[FrameType]], None]
+
 
-def installHandler(fd):
+def installHandler(fd: int) -> int:
```
glyph, line 59: love to see a new type annotation

(and I immediately think: should `FileDescriptorInt` be a `NewType`? But I digress)

src/twisted/internet/_signals.py
```diff
@@ -59,7 +67,8 @@ def installHandler(fd):
 
     @param fd: The file descriptor to which to write when I{SIGCHLD} is
         received.
-    @type fd: C{int}
+
+    @return: The file descriptor previously configured for this use.
```
glyph, line 71: 👏

src/twisted/internet/_signals.py
```diff
@@ -80,6 +89,169 @@ def isDefaultHandler():
...
+    """
+    The L{SignalHandling} protocol enables customizable signal-handling
+    behaviors for reactors.
+
+    A value that conforms to L{SignalHandling} has install and uninstall hooks
+    that are called by a reactor at the correct times to have the (typically)
+    process-global effects necessary for dealing with signals.
+    """
+
+    def install(self) -> None:
```
glyph, line 102: This is a state machine, right? Like it should raise an exception if `install` is called twice?

src/twisted/internet/_signals.py
```diff
@@ -80,6 +89,169 @@ def isDefaultHandler():
...
+        SIGBREAK = getattr(signal, "SIGBREAK", None)
+        if SIGBREAK is not None:
+            signal.signal(SIGBREAK, self._sigBreak)
+
+    def uninstall(self) -> None:
+        """
+        Do nothing for historical reasons.
+        """
+        # TODO Make this do something, someday, because cleaning up your state
+        # is a nice idea.
```
glyph: File a ticket, link it here?

src/twisted/internet/_signals.py
```diff
@@ -80,6 +89,169 @@ def isDefaultHandler():
...
+    implementations are factored.
+
+    @ivar _installed: If L{install} has been called but L{uninstall} has not.
+        This is used to avoid double cleanup which otherwise results (at least
+        during test suite runs) because twisted.internet.reactormixins doesn't
+        keep track of whether a reactor has run or not but always invokes its
+        cleanup logic.
+    """
+
+    _delegates: Sequence[SignalHandling]
```
glyph: Minor naming nitpick: `_signalHandlings` would avoid introducing a new bit of jargon here.

src/twisted/internet/_signals.py
```diff
@@ -80,6 +89,169 @@ def isDefaultHandler():
...
+    _addInternalReader: Callable[[IReadDescriptor], object]
+    _removeInternalReader: Callable[[IReadDescriptor], object]
+    _childWaker: Optional[_SIGCHLDWaker] = None
+
+    def install(self) -> None:
+        """
+        Extend the basic signal handling logic to also support handling
+        SIGCHLD to know when to try to reap child processes.
+        """
+        self._childWaker = _SIGCHLDWaker()
```
glyph:
```suggestion
        assert self._childWaker is None
        self._childWaker = _SIGCHLDWaker()
```

src/twisted/internet/_signals.py
```diff
@@ -120,9 +292,8 @@ class _SocketWaker(log.Logger):
 
     disconnected = 0
 
-    def __init__(self, reactor):
+    def __init__(self):
```
glyph: while we're in here… how much damage would this do
```suggestion
    def __init__(self) -> None:
```

src/twisted/internet/_signals.py
```diff
@@ -178,20 +350,19 @@ class _FDWaker(log.Logger):
 
     disconnected = 0
 
-    i = None
-    o = None
+    i: int
+    o: int
 
-    def __init__(self, reactor):
+    def __init__(self):
```
glyph:
```suggestion
    def __init__(self) -> None:
```

src/twisted/internet/base.py
```diff
@@ -632,32 +646,53 @@ def __init__(self) -> None:
     # Signal handling pieces
     _installSignalHandlers = False
```
glyph:
```suggestion
    _installSignalHandlers: bool = False
```

src/twisted/internet/base.py
```diff
@@ -632,32 +646,53 @@ def __init__(self) -> None:
...
 
-    def _handleSignals(self) -> None:
+    def _maybeSignalsFactory(self, installSignalHandlers: bool) -> SignalHandling:
```
glyph: Naming nitpick: does the `maybe` really add much here?  The active nothing of `_WithoutSignalHandling` means you get a `SignalHandling` no matter what.

src/twisted/internet/cfreactor.py
```diff
@@ -74,6 +74,10 @@ class _WakerPlus(_UnixWaker):
     do.
     """
 
+    def __init__(self, reactor):
+        super().__init__()
+        self.reactor = reactor
```
glyph, line 79: just looking at the coverage here: 😬 

can we have a separate PR for installing the `macos-platform` extra?  maybe landed before this one?

src/twisted/internet/wxreactor.py
```diff
@@ -84,7 +84,7 @@ def _installSignalHandlersAgain(self):
             signal.signal(signal.SIGINT, signal.default_int_handler)
         except ImportError:
             return
-        self._handleSignals()
+        self._signals.install()
```
glyph, line 87: I… think I'm willing to live with the coverage gap here. Is anyone still using wxpython 😬

glyph (approved):
No blockers. I'd love to get test coverage of cfreactor but if it hasn't been running, debugging the interaction with CI might be a pain.

glyph (commented):
Apparently I was a bit too optimistic about cfreactor's prospects.  I got some test failures suspiciously close to the reactor core, and then a re-run made them succeed.  Perhaps test_runAfterStop and test_crash have become flaky...?

src/twisted/internet/_glibbase.py
exarkun: Hm.  Yes, that would be good to know.  The test suite passes on my system with this whole method deleted.  I don't know if that means our test suite is insufficient or if this code is only for obsolete versions of glib or something else.

I'm reluctant to actually delete it since I don't know and since you didn't make this a blocker I'll refrain from doing so _or_ further investigating right here.

src/twisted/internet/_signals.py
exarkun: I'm not sure if it's useful to look at it that way - but with the mess in the current implementation of `_WithSignalHandling` it's a bit hard to tell.

However, I think about the actual type signature of this method as `IO ()`.  Its purpose in life is to have a side-effect.  So, maybe you want that side-effect to happen twice because something else happened in between and the second call should really just re-install everything?

src/twisted/internet/wxreactor.py
exarkun: :grimacing:

src/twisted/internet/_signals.py
exarkun: It turns out some conch tests manage to make such an assertion fail.  It looks like I didn't exactly replicate the old logic - which tested to see if there was already a waker and skipped making a new one if so (however it call `install` either way).

I think this suggests that trial doesn't "correctly" start/crash the reactor with respect to `_reallyStartRunning`.  The refactoring doesn't change when signal handlers are installed so I think this is as broken in current trunk as it is in this branch - but in both current trunk and this branch (without your suggestion) we just tolerate the brokenness.  I will change the code in the branch to tolerate it exactly the same way as it is tolerated in trunk - by not creating a new waker if we have one already, and then unconditionally installing it (as long as that passes the test suite).

This should be *slightly* better than what's in the branch now since what's in the branch now will leak the clobbered SIGCHLDWaker's file descriptors.

src/twisted/internet/_glibbase.py
exarkun: https://github.com/twisted/twisted/issues/11762 and linked in the code

src/twisted/internet/_signals.py
exarkun: And as detailed in one of my other comments it turns out something, probably trial, manages to drive the process such that `.install(); ... .install()` really happens.

exarkun: https://github.com/twisted/twisted/issues/11763 and linked in the code

src/twisted/internet/_signals.py
exarkun: https://github.com/twisted/twisted/issues/11763 as mentioned elsewhere

PR #11754, opened 2022-11-08
#11753 Improve the type annotation of `Deferred.fromFuture`

### LINKED ISSUE #11753: Improve type annotation for `twisted.internet.defer.Deferred.fromFuture`

`Deferred.fromFuture` is annotated as returning `Deferred[Any]`. While this is technically correct (since `Any` means "any type"), it can easily be improved. Moreover, it causes projects using strict mypy checks to necessitate additional annotations.

One could improve the type annotation of this method in this way:

```python
_T = TypeVar("T")

...

    @classmethod
    def fromFuture(cls, future: Future[_T]) -> "Deferred[_T]":
```

An alternative would be leaving things as is, but as I've mentioned that requires the project that calls `fromFuture` to add type annotations that could be inferred from `fromFuture`'s.

**Additional context**

[Here](https://github.com/matrix-org/matrix-content-scanner-python/blob/26c6aa0f0525de22b6ad0caf65ac52fa145e6534/matrix_content_scanner/scanner/scanner.py#L161-L163) is an example of code that requires such superfluous annotation: to make mypy happy, `d` needs to be annotated as a `Deferred[MediaDescription]`, even though that can be inferred from the fact that the values in `self._current_scans` are already annotated as `Future[MediaDescription]`.

babolivier:
## Scope and purpose

Fixes #11753

This PR improves the type annotation for `Deferred.fromFuture` so that it's inferred from the `Future`'s return type.

## Contributor Checklist:

* [x] The title of the PR should describe the changes and starts with the associated issue number.
* [x] A release notes news fragment file was create in src/twisted/newsfragments/ (see: [Release notes fragments docs.](https://docs.twisted.org/en/latest/core/development/dev-process.html#release-notes-management))
* ~The automated tests were updated.~ this PR is only changing a type annotation so I'm not sure there's any test to update - please advise if this isn't true.
* [x] Once all checks are green, request a review by leaving a comment that contains exactly the string `please review`.


src/twisted/newsfragments/11753.misc
```diff
@@ -0,0 +1 @@
+Improve the type annotation of Deferred.fromFuture.
```
exarkun:
```suggestion
twisted.internet.defer.Deferred.fromFuture now has a more precise type annotation.
```

exarkun: Note that ".misc" fragments are usually empty because their content will not be included in the release notes.

babolivier: > Note that ".misc" fragments are usually empty because their content will not be included in the release notes.

What would you suggest: should I just empty out this one, or would this fit better in another section?

exarkun: It's public-facing so I'd say go ahead and move it to another section so the details show up.

src/twisted/internet/defer.py
```diff
@@ -1041,7 +1041,7 @@ def maybeSucceed(result: object) -> None:
         return future
 
     @classmethod
-    def fromFuture(cls, future: Future) -> "Deferred[Any]":
+    def fromFuture(cls, future: Future[_T]) -> "Deferred[_T]":
```
exarkun: It looks like `Future` is not generic on Python 3.7, so this fails.

I guess making it a string is the solution?

babolivier: > I guess making it a string is the solution?

Probably yes.

babolivier:
please review

src/twisted/internet/defer.py
```diff
@@ -1041,7 +1041,9 @@ def maybeSucceed(result: object) -> None:
         return future
 
     @classmethod
-    def fromFuture(cls, future: Future) -> "Deferred[Any]":
+    def fromFuture(
+        cls, future: "Future[_DeferredResultT]"
+    ) -> "Deferred[_DeferredResultT]":
```
exarkun, line 1046: Now I'm very confused about `_DeferredResultT` being contravariant... :/

glyph: It's definitely wrong, but having spent some time trying to fix this now, it becomes clear that the real signatures for Deferred's methods are fiendishly complex, and this contravariance papers over some of the complexities (by accepting invalid types).

Trying to see if I can golf my way to a successful replacement that *doesn't* do that.

glyph:
@babolivier FYI, #11770 incorporates these commits, and will automatically close this PR as well if it's merged.

glyph:
(Not blocking on this one, I'd just like to see a more comprehensive change to the annotations. If someone else comes along and merges it I'll happily integrate the commit.)

src/twisted/internet/defer.py
graingert: 3.7 is no longer supported so should be fine now

babolivier:
Just catching up, I see #11770 has been merged, so I'm closing this one.

PR #11756, opened 2022-11-09
#11708 Test cfreactor on macOS on CI

### LINKED ISSUE #11708: Test CFReactor on CI
Twisted has a "CoreFoundation" reactor for doing interesting macOS platform integration.  And there are tests for this reactor in `twisted.internet.test`.  But!  These tests are not by the macOS CI job.  They should be because they don't tell us if something is broken when they're not run.

exarkun:
Fixes #11708


codecov.yml
```diff
@@ -6,6 +6,7 @@
 # the coverage.
 #
 codecov:
+  token: fb52015a-b78e-4d89-a6e0-a4829aea2efa
```
adiroiban: we should not require to add a token for open source project.

I think that the codecov token errors are due to other root causes.

codecov was working without a token for a long time.

src/twisted/internet/cfreactor.py
```diff
@@ -385,6 +385,13 @@ def mainLoop(self):
         finally:
             self._inCFLoop = False
 
+        assert not self.running, (
```
adiroiban: a comment might help us understand why we need this assertion here... or maybe have a different assertion message that can serve to document this code.

adiroiban:
is this ready for review... it was a draft and then marked as not a draft, so the robot automatically tagged it as ready for review.

i see the macOS tests still failing

exarkun:
Sorry, this is still in development.

codecov.yml
glyph: According to the documentation, it's not required.

In practice, there's a race condition on codecov's side, some buggy logic which seems to disproportionately affect projects with larger numbers of github actions statuses that report coverage.

Before we added this coverage was failing something like 40% of builds on this PR; after adding it, I haven't seen one.

src/twisted/internet/cfreactor.py
glyph: This is part of the WIP and has been removed in the final version.

glyph:
please review

src/twisted/internet/cfreactor.py
```diff
@@ -459,7 +517,8 @@ def iterate(self, delay=0):
         by letting the loop run for a little while and then scheduling a timed
         call to exit it.
         """
-        self.callLater(delay, self._stopNow)
+        self._started = True
+        self.callLater(0, self.crash)
```
adiroiban, line 563: is the `delay` ignored on purpose? 
... and why do we call crash on `itearate` ?

I think that this need more comments :)

.github/workflows/test.yaml
```diff
@@ -129,6 +129,7 @@ jobs:
           - python-version: '3.11'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
+            tox-env: 'macos-withcov-alldeps'
```
adiroiban, line 132: with this change, we make sure that cfreactor deps are installed and we don't skip the cfreactor targeted unit/integration tests, while all the other tests are executed with the select reactor.

If we care about cfreactor, I think is best to run the full test suite with the cfreator...
Most probably it will not detect any other failure, but I think is best to have it and increase the probability of detecting regressions or cfreator-related issues.

and I guess that we still want to run select reactor on macOS

So maybe have these CI macOS environments:

* Python 3.7 with select reactor on macOS 12 - the minimum python that we support
* Python 3.11 with cfreactor reactor on macOS 12 - the maximum/latest python that we support

adiroiban:
I left just a few comments as I am not familiar with macOS or cfreactor.
The CI comment is not a blocker.
The  `iterate` part looks strange. I was expecting the tests to fail with such a change.

src/twisted/internet/cfreactor.py
exarkun: 2be0fc7aca

.github/workflows/test.yaml
exarkun: I don't necessarily object but I'm also not going to volunteer for this work.  As-is, this PR adds a lot of CFReactor coverage on CI (it was 0 before, now it's "a lot").  Since you mentioned below that you don't think this is a blocker, I will leave it for another time.

src/twisted/internet/cfreactor.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, runLoop=None, runner=None):
         self._runner = runner
 
         if runLoop is None:
-            runLoop = CFRunLoopGetMain()
+            runLoop = CFRunLoopGetCurrent()
```
exarkun, line 135: `CFRunLoopGetMain` gets "the" main loop.  I suppose this is "the main thread's" event loop?
`CFRunLoopGetCurrent` gets "the current thread's" main loop.

The latter seems more likely to be correct in general.

* https://developer.apple.com/documentation/corefoundation/1542428-cfrunloopgetcurrent 
* https://developer.apple.com/documentation/corefoundation/1542890-cfrunloopgetmain

Both doc pages say the API follows "The Get Rule".  Do the Python bindings deal with that for us or are we supposed to be making `CFRetain` calls from Python (@glyph )?

src/twisted/internet/cfreactor.py
```diff
@@ -372,22 +372,85 @@ def _moveCallLaterSooner(self, tple):
...
 
+    def startRunning(self, installSignalHandlers: bool = True) -> None:
+        """
+        Start running the reactor, then kick off the timer that advances
+        Twisted's clock to keep pace with CFRunLoop's.
+        """
+        super().startRunning(installSignalHandlers)
+        # We must force the simulator to run because in addition to scheduling
+        # timed calls, super().startRunning also fires startup events which may
+        # crash the reactor.
```
exarkun: I don't quite follow this explanation.  I tried to expand some of the implications here but I don't think I got somewhere coherent:

```
        # Prior to this point (maybe because of the super().startRunning)                                                                                   
        # arbitrary application code may have run.  This may include code that                                                                              
        # added event sources (timers or file descriptors) to the reactor.  If                                                                              
        # so, either _scheduleSimulate will already have run and added a Timer                                                                              
        # to the CFRunLoop or _watchFD will have added a Source.  Either way,                                                                               
        # there is work for CFRunLoop to do and things will work well.                                                                                      
        #                                                                                                                                                   
        # If neither of those things happened then there is no work for the                                                                                 
        # CFRunLoop to do.                                                                                                                                  
        #                                                                                                                                                   
        # ... so what?                                                                                                                                      
```

src/twisted/internet/cfreactor.py
```diff
@@ -372,22 +372,85 @@ def _moveCallLaterSooner(self, tple):
...
-    def mainLoop(self):
+    def mainLoop(self) -> None:
         """
         Run the runner (C{CFRunLoopRun} or something that calls it), which runs
         the run loop until C{crash()} is called.
         """
-        self._inCFLoop = True
+        # we were crashed during startup.
+        if not self._started:
+
```
exarkun: I'd put this comment inside the `if` since that's the case it's describing.  Also I tried to expand this one and didn't succeed:

```
            # startRunning set _started to True so between then and now,                                                                                                          
            # something crashed us.  There aren't a lot of candidates for                                                                                                         
            # this.  The most well-known is:                                                                                                                                      
            #                                                                                                                                                                     
            #   reactor.callWhenRunning(reactor.crash)                                                                                                                            
            #                                                                                                                                                                     
            # We *don't* eagerly return here because ...                                                                                                                          
            #                                                                                                                                                                     
            #    why?                                                                                                                                                             
            #                                                                                                                                                                     
            # To avoid that, arrange to let the CF main loop run very briefly.                                                                                                    
```

src/twisted/internet/cfreactor.py
```diff
@@ -372,22 +372,85 @@ def _moveCallLaterSooner(self, tple):
...
+            def docrash() -> None:
+                self.crash()
+
+            # we're running again, just for a moment, while we wait for some
+            # I/O to take place.  We need this to be True for callLater to not
```
exarkun: The "I/O" that we wait for is just the CF Timer to fire and call our crash function right?

src/twisted/internet/cfreactor.py
```diff
@@ -372,22 +372,85 @@ def _moveCallLaterSooner(self, tple):
...
+                    # Sometimes CFRunLoopRun (or its equivalents) may exit
+                    # early. This is really only *supposed* to happen when it
+                    # runs out of sources & timers to process, but it appears
+                    # that in cases where the reactor is being rapidly started
+                    # and crashed repeatedly, and sources are being rapidly
+                    # invalidated both inside and outside the loop (e.g. in the
+                    # Twisted test suite), CoreFoundation (possibly
+                    # incorrectly?) exits the runloop early.  To accommodate
+                    # this stress case, we make extra sure that our scheduled
+                    # timer is re-created on the loop as a CFRunLoopTimer.
```
exarkun:
```suggestion
                    # Twisted test suite), CoreFoundation exits the runloop even
                    # when there *is* some work to do. To accommodate this
                    # stress case, we make extra sure that our scheduled
                    # timer is re-created on the loop as a CFRunLoopTimer.
```

src/twisted/internet/test/test_cfreactor.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        r = self.buildReactor()
+        stopped = []
+
+        def doStop() -> None:
+            r.stop()
+            stopped.append("yes")
+
+        def sleepThenStop() -> None:
+            sleep(0.5)
```
exarkun: What's the point of the sleep?  What order of events is this test trying to exercise?

src/twisted/internet/cfreactor.py
glyph: Correct on both counts. It's pyobjc's job, generally, to deal with retention rules.

src/twisted/internet/cfreactor.py
glyph: What I was trying to get across here is that we need to call into the actual loop, not just run the logic of the timer immediately. To allow some chance for already-ready file descriptors to be invoked. So the "I/O" is explicitly in contrast to this logic.

I'll try to make this clearer

src/twisted/internet/cfreactor.py
glyph: I cannot prove that there is in fact "work to do", at least according to the semantics established by CFRunLoop: there are various points where sources are automatically invalidated (when timers expire, I think), and some of our uses of the API (for example, `CFSocketCreateWithNative` with a pipe file descriptor rather than a socket) are taking some liberties. So it's possible that we just haven't correctly appreciated some nuance of automatic invalidation

src/twisted/internet/test/test_cfreactor.py
glyph: The goal here is to start the loop with no timers, so that scheduleSimulate exercises the "timeout is None" case. If we don't sleep, then the shutdown runs scheduleSimulate with force=True, forcing an immediate shutdown and no further calls to scheduleSimulate. We have to run through startup while fully running the loop, then have some I/O (waker.doRead) wake us up with no timers in sight.

glyph:
> I left just a few comments as I am not familiar with macOS or cfreactor.
> 
> The CI comment is not a blocker.
> 
> The  `iterate` part looks strange. I was expecting the tests to fail with such a change.
> 
> 

Strange... how? It doesn't look strange to me, given that I wrote it that way on purpose, and it does in fact not break the tests :-). (Just not sure what to do with this feedback.)

.github/workflows/test.yaml
glyph: For reasons related to the way global state works in the corefoundation API, this would be *extremely* difficult to implement without completely rewriting trial from scratch. So I am not sure we should even bother tackling this later.

glyph: (In any case, definitely nothing to do on this PR so resolving)

src/twisted/internet/cfreactor.py
glyph: This is … directionally correct, so I'll try to do an update that addresses some of these problems

src/twisted/internet/cfreactor.py
glyph: OK I think that should do it

src/twisted/internet/test/test_cfreactor.py
glyph: Maybe I'm wrong about this. I tried to explain it to myself, and I could not construct the sequence of events that I thought it was there to illustrate.  If codecov agrees then I think we can drop it.

glyph: OK. I was wrong. The coverage gap I saw on that line must have been for unrelated reasons. The sleep is gone!

src/twisted/internet/cfreactor.py
glyph: Much lengthier comment hopefully explains the situation in detail.

src/twisted/internet/cfreactor.py
glyph: Aaand upon improving the verbiage I realized the previous version was wrong anyway. Oops! Hopefully the new version is correct.

glyph:
OK, I think that this is pretty good to land now. Does anyone feel like an approval?

src/twisted/internet/test/test_cfreactor.py
```diff
@@ -0,0 +1,97 @@
...
+    fakeBase = object
+
+
+def noop() -> None:
+    """
+    Do-nothing callable. Stub for testing.
+    """
+
+
+noop()
```
adiroiban: If this is needed, I think that it needs a comment.
It is not obvious why this is needed.

src/twisted/internet/test/test_cfreactor.py
```diff
@@ -0,0 +1,97 @@
...
+        def r2stop() -> None:
+            r2.stop()
+
+        r2.callLater(0, stopBlocking)
+        self.runReactor(r2)
+        self.assertEqual(r.getDelayedCalls(), [delayed])
+
+    def test_whiteboxIterate(self) -> None:
+        """
+        C{.iterate()} starts the main loop, then crashes the reactor.
```
adiroiban: I don't understand this test.
My understanding is that the `crash` part of iterate is a private implementation detail.

For the tests, I was expecting not to have it mentioned in relation to iterate.

----------

So maybe the description could be

`iterate() executes any callLater callbacks and leaves the reactor stopped`

I guess that the purpose of this test is to explain that iterate on cfreactor has a different behaviour than other reactors.

So maybe, in this test we should have `r.iterate(4)` and it should explain that even if requested to run for 4 seconds it will also execute the delayed called scheduled for 100 seconds.

src/twisted/internet/cfreactor.py
```diff
@@ -401,24 +500,28 @@ def _scheduleSimulate(self, force=False):
...
-            CFRunLoopTimerInvalidate(self._currentSimulator)
+        self._stopSimulating()
+        if not self._started:
+            # If the reactor is not running (e.g. we are scheduling callLater
+            # calls before starting the reactor) we should not be scheduling
+            # CFRunLoopTimers against the global CFRunLoop.
+            return
+
+        timeout = 0.0 if force else self.timeout()
+        if timeout is None:
```
adiroiban, line 511: I know that this was already in trunk.
I is not clear what is the semantic of `timeout = None`

adiroiban (approved):
Thanks for your work !

I left a few minor comments.
I see that the changes from this PR are done by both Jean Paul and Glyph.

If you are happy, I think that this should be merged.

I have never used the cfreactor... so I am not the best person to approve, but I am not sure if anyone else will have  time to review it soon.

And if  anyone cares about cfreactor, I think we can always follow up with a new PR.

My understanding is that the code from trunk is broken anyway.

Regards

src/twisted/internet/test/test_cfreactor.py
glyph: The exiting, black-box tests already test this, but I'll update the docstring to explain the internal state being tested.

src/twisted/internet/cfreactor.py
glyph: This is a common feature of pretty much every reactor; it just means "there are no scheduled calls". 

This ought to be documented in `twisted.internet.base.ReactorBase.timeout` but that's a separate change :)

glyph:
> My understanding is that the code from trunk is broken anyway.

Technically this is accurate, but "broken" as a blanket term is misleading. The reactor before this change works fine for applications, and even *mostly* passes the Twisted suite itself.  There are just some edge case race conditions that specifically get triggered in the very unusual context of Twisted's tests, where thousands of instances of the CF reactor are being instantiated and started and stopped.

While the implementation is now more correct, I do not expect that its correctness will be visible to applications, which is why I decided to go with a `.misc` change and not include the robustness improvements in the changelog; the robustness is mostly just for our own usage in our own testing.

The problem here is that CoreFoundation provides tons of global, un-inspectable state, which we cannot test against and we cannot isolate to individual instances, only to threads.

PR #11767, opened 2022-11-21
#11766 upgrade mypy

### LINKED ISSUE #11766: upgrade mypy
it's been a minute, let's get up to date

glyph:
## Scope and purpose

Fixes #11766

## Contributor Checklist:

### DIFF
--- a/src/twisted/logger/test/test_util.py
+++ b/src/twisted/logger/test/test_util.py
@@ -85,40 +85,19 @@ class UtilTests(unittest.TestCase):
         def o5(e: LogEvent) -> None:
             pass
 
-        o1.name = "root/o1"  # type: ignore[attr-defined]
+        o1.name = "root/o1"
         o2.name = "root/p1/o2"
         o3.name = "root/p1/o3"
         o4.name = "root/p1/p2/o4"
         o5.name = "root/o5"
 
+        expectedTrace: str  # populated below
+
         @implementer(ILogObserver)
         def testObserver(e: LogEvent) -> None:
             self.assertIs(e, event)
             trace = formatTrace(e["log_trace"])
-            self.assertEqual(
-                trace,
-                (
-                    "{root} ({root.name})\n"
-                    "  -> {o1} ({o1.name})\n"
-                    "  -> {p1} ({p1.name})\n"
-                    "    -> {o2} ({o2.name})\n"
-                    "    -> {o3} ({o3.name})\n"
-                    "    -> {p2} ({p2.name})\n"
-                    "      -> {o4} ({o4.name})\n"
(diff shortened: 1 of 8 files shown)


src/twisted/web/_flatten.py
```diff
@@ -272,7 +272,7 @@ def keepGoing(
         dataEscaper: Callable[[Union[bytes, str]], bytes] = dataEscaper,
         renderFactory: Optional[IRenderable] = renderFactory,
         write: Callable[[bytes], object] = write,
-    ) -> Generator[Union[Generator, Deferred[Generator]], None, None]:
+    ) -> Generator[Union[Flattenable, Deferred[Flattenable]], None, None]:
```
glyph, line 275: Kinda weird that this worked before!

src/twisted/logger/test/test_util.py
```diff
@@ -130,4 +109,31 @@ def testObserver(e: LogEvent) -> None:
...
+            root=root,
+            o1=o1,
+            o2=o2,
+            o3=o3,
+            o4=o4,
+            o5=o5,
+            p1=p1,
+            p2=p2,
+            oTest=oTest,
+        )
```
rockstar, line 138: Are said "shenanigans" a thing that would prevent us from using f-strings here?

src/twisted/web/_flatten.py
rockstar: I just spent 20 minutes trying to figure out why it worked before, and my best guess now is "mypy bug."

rockstar (approved):
Thanks for doing this. This kinda stuff often gets overlooked but goes a long way in producing quality over time. I had one small question but, honestly, f-strings v `.format` is six of one and half a dozen of the other; if you don't want to address it, don't even worry about it.

src/twisted/logger/test/test_util.py
glyph: Yeah. The shenanigan is that we want to grab this "name" attribute that is not properly predeclared, and mypy (rightly) complains if it has visibility into the code that's being executed in the string.

glyph: (fstrings would obviously improve the visibility even more, which would make it worse.)

adiroiban (approved):
Thanks Paul for the review and Glyph for the mypy updates.

I am doing a rubber-stamp approval here to get this PR cleared.
It looks like Paul is not a member of Twisted.

@glyph feel free to ask a review from the Twisted team if you think that the review from Paul is not enough.

glyph:
@adiroiban Thanks, much appreciated! We need to make the robot allow members accepting reviews from external contributors, so the robot will be credited with the approval, if someone external does an approval.

PR #11770, opened 2022-11-22
#11772 Correct deferred annotations

### LINKED ISSUE #11753: Improve type annotation for `twisted.internet.defer.Deferred.fromFuture`

### LINKED ISSUE #11772: Deferred's type annotations are generally wrong
The current type annotations have a number of issues:

- `_DeferredResultT` is contravariant, which is confusing and inaccurate; it should be invariant
- `addCallback`, `addErrback`, and `addBoth` don't check their callbacks' signatures at all; you don't have to accept a `Failure` for the latter two, to say nothing of the rest of the signature.
- `Deferred.fromFuture` uses an unspecialized generic (see #11753)
- `maybeDeferred` incorrectly returns a `Deferred[Deferred[X]]` if its function returns a deferred

Some of these are due to new features in Mypy that have come out since the type hints were added, some are simply errors in the initial implementation, but we should give it a new pass to address both and have more correct stubs.  (Some issues are not possible to fix, such as `addCallabcks` having multiple `ParamSpec` signatures in a way that mypy does not yet allow expressing.)

glyph:
## Scope and purpose

Fixes #11772
Fixes #11753 

## Contributor Checklist:


glyph:
Note that this is stacked on top of #11767 since it requires some newer mypy features.

glyph:
It's also stacked on top of #11754 but it supersedes that PR since there are a lot of follow-on changes from there

src/twisted/internet/defer.py
```diff
@@ -174,8 +175,22 @@ def execute(
...
+
+@overload
 def maybeDeferred(
     f: Callable[_P, _T], *args: _P.args, **kwargs: _P.kwargs
+) -> "Deferred[_T]":
+    ...
+
+
+def maybeDeferred(
+    f: Callable[_P, Union[_T, Deferred[_T]]], *args: _P.args, **kwargs: _P.kwargs
```
exarkun: This doesn't annotate the coroutine case correctly (but it also wasn't annotated correctly before so maybe this is unrelated).

src/twisted/internet/defer.py
```diff
@@ -466,16 +481,27 @@ def __init__(
...
-            ...,
-            "Union[_NextDeferredResultT, Deferred[_NextDeferredResultT]]",
+        callback: Union[
+            Callable[..., _NextDeferredResultT],
+            Callable[..., Deferred[_NextDeferredResultT]],
+            Callable[..., Failure],
+            Callable[
+                ...,
+                Union[_NextDeferredResultT, Deferred[_NextDeferredResultT], Failure],
+            ],
```
exarkun, line 504: The repetition of the return two in unions in two different places looks silly.  Is this really required to make the type checker happy?  If so it might be worth a comment.

src/twisted/internet/defer.py
```diff
@@ -466,16 +481,27 @@ def __init__(
...
+            None,
+        ] = None,
+        callbackArgs: Tuple[Any, ...] = (),
```
exarkun, line 516: is it a bug in mypy that you can't use ParamSpec here?  Or a missing feature from the ParamSpec PEP?  A comment / issue link here would probably be good.

src/twisted/internet/defer.py
```diff
@@ -515,54 +542,151 @@ def addCallbacks(
         #     the caller should treat the result as the new type, consistently.
         return cast(Deferred[_NextDeferredResultT], self)
 
+    @overload
```
exarkun, line 575: Was there a reason not to use `@overload` for `addCallbacks` but to use it here?

edit - maybe because there could be as many as ~20 overloads for `addCallbacks`?

src/twisted/internet/defer.py
```diff
@@ -515,54 +542,151 @@ def addCallbacks(
...
+        **kwargs: _P.kwargs,
+    ) -> Deferred[_NextDeferredResultT]:
+        ...
+
+    def addCallback(
+        self,
+        callback: Any,
+        *args: Any,
+        **kwargs: Any,
+    ) -> "Deferred[Any]":
```
exarkun: Errrr :/  Those `Any` annotations look so sad.  PEP 484 and the Python library documentation both imply you can leave annotations _off_ of the final definition of an `@overload`ed function ... but I see that mypy complains about this function if we do so.

Is that because of something we set in our mypy config?  If it's possible to keep these `Any` annotations out it seems like that would be less confusing.  As far as I can tell, mypy does ignore these annotations (apart from requiring that they exist _at all_) so this is more about confusion to future maintainers and other humans.

src/twisted/internet/defer.py
```diff
@@ -515,54 +542,151 @@ def addCallbacks(
...
+        self,
+        errback: Callable[Concatenate[Failure, _P], _NextDeferredResultT],
+        *args: _P.args,
+        **kwargs: _P.kwargs,
+    ) -> "Deferred[Union[_DeferredResultT, _NextDeferredResultT]]":
+        ...
+
+    def addErrback(
+        self,
+        errback: Callable[..., Any],
```
exarkun: If we have to annotate this, it seems desirable for the annotation of `errback` to `addErrback` match the annotation of `callable` to `addCallback`.

src/twisted/internet/defer.py
```diff
@@ -515,54 +542,151 @@ def addCallbacks(
...
+    ) -> Deferred[_NextDeferredResultT]:
+        ...
+
+    @overload
+    def addBoth(
+        self,
+        callback: Callable[Concatenate[_T, _P], _T],
+        *args: _P.args,
+        **kwargs: _P.kwargs,
+    ) -> Deferred[_DeferredResultT]:
```
exarkun: Yikes?  This seems like a very unprincipled overload.  I commented it out and mypy didn't give me any errors.  I can't see a reason to _want_ this signature if it isn't being jammed down our throats.

src/twisted/internet/defer.py
```diff
@@ -515,54 +542,151 @@ def addCallbacks(
...
-            ...,
-            "Union[_NextDeferredResultT, Deferred[_NextDeferredResultT]]",
+            Concatenate[Union[_DeferredResultT, Failure], _P],
+            Union[_DeferredResultT, Failure],
         ],
-        *args: object,
-        **kwargs: object,
+        *args: _P.args,
+        **kwargs: _P.kwargs,
+    ) -> Deferred[_NextDeferredResultT]:
```
exarkun: I don't understand this annotation.  The combination of Union and `@overload` (which are both basically different ways to enlarge the type of the function) probably doesn't help.

This says "Give me a `Deferred[_DeferredResultT]` and a function that returns either `_DeferredResultT` or `Failure` and I will give you a `Deferred[_NextDeferredResultT]`".  `_NextDeferredResultT` isn't even bound in this statement so there's no way this could work.

I don't think this `@overload` is covering a meaningful case at all and should just be removed.

src/twisted/internet/defer.py
```diff
@@ -583,8 +707,13 @@ def addTimeout(
         self,
         timeout: float,
         clock: IReactorTime,
-        onTimeoutCancel: Optional[Callable[[object, float], object]] = None,
-    ) -> "Deferred[_DeferredResultT]":
+        onTimeoutCancel: Optional[
+            Callable[
+                [Union[_DeferredResultT, Failure], float],
+                Union[_NextDeferredResultT, Failure],
```
exarkun: Maybe this is related to the `addBoth` that I commented about on line 662?

src/twisted/internet/defer.py
```diff
@@ -1629,7 +1757,7 @@ def returnValue(val: object) -> NoReturn:
 
 
 @attr.s(auto_attribs=True)
-class _CancellationStatus:
+class _CancellationStatus(Generic[_DeferredResultT]):
```
exarkun: At least `_inlineCallbacks` uses `_CancellationStatus` unparameterized.

src/twisted/internet/defer.py
```diff
@@ -1824,16 +1952,16 @@ def _cancellableInlineCallbacks(
     @return: L{Deferred} for the C{@}L{inlineCallbacks} that is cancellable.
     """
 
-    def cancel(it: Deferred[object]) -> None:
+    def cancel(it: Deferred[_T]) -> None:
         it.callbacks, tmp = [], it.callbacks
         it.addErrback(handleCancel)
         it.callbacks.extend(tmp)
         it.errback(_InternalInlineCallbacksCancelledError())
```
exarkun, line 2128: well, I'm not especially glad I read these lines...

src/twisted/internet/defer.py
```diff
@@ -1824,16 +1952,16 @@ def _cancellableInlineCallbacks(
...
 
-    deferred: Deferred[object] = Deferred(cancel)
+    deferred: Deferred[_T] = Deferred(cancel)
     status = _CancellationStatus(deferred)
 
-    def handleCancel(result: Failure) -> Deferred[object]:
+    def handleCancel(result: Failure) -> Union[Failure, _T, Deferred[_T]]:
```
exarkun: Looks like this never returns anything except `Deferred[_T]`

exarkun (commented):
Thanks.  I left some comments inline.

glyph:
Thanks @exarkun, the attention to detail is appreciated. As you may imagine there was a lot of thrashing around here, and I think you spotted a few cases where I was commenting/uncommenting and experimentally trying to induce errors, and left things in in an invalid state.

src/twisted/internet/defer.py
glyph: So, the problem here is the undecidability of `_NextDeferredResultT` vs. `_NextDeferredResultT  | Failure`.  If we don't put a `Failure` into an explicit union in the return type, then we get `Deferred[foo | Failure]` as a return type sometimes, which is bogus.  In fact, you called my attention to the fact here that I didn't put *enough* gross overloads onto `addCallback` for this to work properly!

In the absence of a better way to add tests for type coverage, I'll probably write some unit tests that will fail to typecheck under the current signatures, but it would probably be good to start adding some "this should be a static type error" tests in there too…

glyph: OK after some brief further experimentation, I need to go back to the drawing board on this and take notes to make this more minimal.

glyph: (And probably report some bugs in Mypy. Happy thanksgiving 🦃 )

src/twisted/internet/defer.py
glyph: It's a deliberately missing feature from the `ParamSpec` PEP. And yeah, I thrashed around for like an hour on this, trying to understand why it wasn't working because *surely* it works in a context other than `*args`? but no, it doesn't, on purpose. I can link to it for clarity.

src/twisted/internet/defer.py
glyph: The fact that you can't specify the argument signatures to the callbacks and errbacks makes it pointless to even attempt overloads there.

src/twisted/internet/defer.py
glyph: Leaving the annotations off of an overload makes everything an implicit any, and then the lack of an annotation turns off type checking for everything that *isn't* a parameter in the function body besides.  So yes, we intentionally disable this in our mypy configuration to avoid that behavior.

src/twisted/internet/defer.py
moshez: The PR is marked "needs review", but I'm not sure what the conclusion of this discussion was.
Do we need a review or is the drawing board still the place?

src/twisted/internet/defer.py
glyph: OK I think I got this one

src/twisted/internet/defer.py
glyph: It looks like exarkun did not feel this review was complete :)

src/twisted/internet/defer.py
glyph: Yes, good catch.

src/twisted/internet/defer.py
glyph: This was definitely a mistake. Thank you for catching. I am seriously starting to think about writing a program to emit all these overloads because that might be easier to check for errors than this manual repetition :-|

src/twisted/internet/defer.py
glyph: OK, I think I have eliminated all the "callback callable returns `_DeferredResultT`" overloads. In order to make this easier to read, I've renamed the variables to `_SelfResultT` and `_NextResultT`, to have a bit less confusing repetition and hopefully make this type of mistake easier to spot.

src/twisted/internet/defer.py
glyph: Yeeaahh. With that fixed, I get `src/twisted/internet/defer.py:809:55: error: Argument 1 to "addBoth" of "Deferred" has incompatible type "Callable[[_T], _T]"; expected "Callable[[object], Failure]"` and I think I am missing one of the other overloads from `addCallback`

glyph: OK the type inferencer just kind of gives up here unless you give it this unnecessary help that breaks it in other circumstances. I don't think this is the sort of thing that application code is likely to be doing, so I'll just add the annotation here and move on.  Definitely a mypy bug but any attempt to minify it makes it go away, so we will have to figure out exactly what it is later.

src/twisted/internet/defer.py
glyph: Looks like that's the only place in the codebase that it's used, so fixed :)

glyph:
OK, I think that's everything

src/twisted/internet/defer.py
glyph: Comment added.

glyph:
please review

src/twisted/conch/unix.py
```diff
@@ -51,10 +52,12 @@
 class UnixSSHRealm:
     def requestAvatar(
         self,
-        username: bytes,
+        username: bytes | Tuple[()],
         mind: object,
         *interfaces: portal._InterfaceItself,
     ) -> Tuple[portal._InterfaceItself, UnixConchUser, Callable[[], None]]:
+        if not isinstance(username, bytes):
```
adiroiban, line 59: I guess that this is fine, since this is the current behaviour.

----------

Just a comment on design.

I did take a look at twisted.cred a long time ago, and start using it, but then I switched to a custom implementation.

For anonymous I ended up implementing it using different credential type ( `IAnonymous` ) and then the realm doesn't have to care about anonymous.

The anonymous cred checker will reject or accept the anonymous user and return the avatar that will handle the anonymous access.

Not sure which is best.
I still think that the idea of pluggable auth method is nice... but things get more complicated when for example you implement OIDC + extra OAuth2 API access for Realm

src/twisted/internet/_resolver.py
```diff
@@ -247,26 +246,25 @@ def resolveHostName(
...
         resolutionReceiver.resolutionBegan(resolution)
-        onAddress = self._simpleResolver.getHostByName(hostName)
-
-        def addressReceived(address: str) -> None:
-            resolutionReceiver.addressResolved(IPv4Address("TCP", address, portNumber))
-
-        def errorReceived(error: Failure) -> None:
-            if not error.check(DNSLookupError):
-                self._log.failure(
+        (
```
adiroiban, line 249: maybe it's best to have this as inline callbacks?
I am not a fan of LISP and big nested parenthesis and chained functions.

glyph:
The test failure is a result of #11843 and should be addressed by its merge.

glyph:
please review

glyph:
'comment' reviews really should not take it out of the review queue :)

src/twisted/conch/unix.py
glyph: > I did take a look at twisted.cred a long time ago, and start using it, but then I switched to a custom implementation.

This is interesting for discussion but not really relevant to this ticket so I am resolving :)

src/twisted/internet/_resolver.py
glyph: There's a teetering jenga tower of type inference happening here and I'd prefer to keep it as-is for the moment rather than `async def`-ifying it just for aesthetics.  Happy to do a follow-up PR about this though.

glyph:
> The test failure is a result of #11843 and should be addressed by its merge.

And, indeed, it was.

adiroiban:
> 'comment' reviews really should not take it out of the review queue :)

True... but this is how GitHub works ... if you request a review from a team, as soon as a team member makes a comment on the PR, the team is removed.

This is one reason why we have the "needs-review" label, to have better control.

In the future I will try to manually re-request a review from the team.

And in the future, maybe the robot can do this for us... if a only a "comment" review is added, re-add the team in the list of reviewers.

ihaywood3:
this is a big PR but I can't see any problems with it. I note that two core developers have already reviewed most of the changes.
My only question is about `_resolver.py` - I assume these changes are to please mypy. Maybe a comment in the code or in the newsfragment?

glyph:
> this is a big PR but I can't see any problems with it. I note that two core developers have already reviewed most of the changes.

I will interpret this as an "approve" review, make the changes and land.

glyph:
> My only question is about `_resolver.py` - I assume these changes are to please mypy. Maybe a comment in the code or in the newsfragment?

Ah, reviewing this now I can see why you might be confused, but the change doesn't modify behavior (and thus doesn't belong in the news fragment) and this style isn't sufficiently interesting in its own context to warrant an explanation.  The *change* could perhaps use one though, so I'll explain here:

Deferreds are an unusual kind of object because their type changes when you call methods on them.  Technically speaking, this is impossible for a type-checker like Mypy to actually validate.  Like, consider this:

```python3
def mutate(d: Deferred[int]) -> Deferred[str]:
    d.addCallback(str)
    return d
...

x: Deferred[int] = deferredIntReturner()
mutate(x)
mutate(x)
```
There's no way that mypy could see that the second call to `mutate` is invalid. It thinks that `x` is, and forever remains, `Deferred[int]`.  If we were designing `Deferred` today, it would almost certainly return a new value from `addCallback` for this reason.

However, *idiomatically*, it's reasonable to work this way and callers sharing `Deferred` objects is not a problem that practically arises very often.  So `addCallback((int) -> str)` can literally `return self` but be *annotated* to say it returns `Deferred[str]`, and everyone can participate in this fiction.

In order to make the fiction of enforced-single-use-`Deferred` and Mypy's implementation match up, we have to *use the return values* of the `addCallback` family of methods rather than simply call methods repeatedly on the same object.  The change to `resolver.py` is adopting this style to allow for type checking where previously none was happening because we were getting implicit `Any`s sneaking in to various places in the signature.  More correct annotations now result in the need to actually make this check properly.

PR #11780, opened 2022-12-02
fix ReactorBase.crash to re-add startup event trigger at the right time

### LINKED ISSUE #11779: repeated calls to `ReactorBase.crash` can add duplicate startup event triggers
`crash` calls `self.addSystemEventTrigger("during", "startup", self._reallyStartRunning)` because event triggers are cleared out when they are run; once startup has happened, anything you want to happen the *next* time it fires needs to be added back.  The idea here being that once we have `crash()`ed, we might start up again.

This is wrong because you *could* call `crash()` multiple times outside the loop; effectively it's a broken implementation of a state machine that will corrupt the internal state of the reactor.

This is a very weird stress case that should not occur in normal operation (you call crash between zero and one times in a normal twisted program after all) but should nevertheless be fixed.

(One acceptable fix might just be making calls to `crash` outside the loop raise an exception because seriously what are you doing?)

glyph:
## Scope and purpose

Fixes: #11779

## Contributor Checklist:


glyph:
@exarkun your refactorings might obviate the need for this so feel free to close, I just filed/open this so we don't forget about it

PR #11789, opened 2022-12-12
#11788 Pin tox to latest version 3 for GitHub Actions.

### LINKED ISSUE #11788: GitHub Actions CI broken due to latest tox
```
ROOT: will run in automatically provisioned tox, host /opt/hostedtoolcache/Python/3.10.8/x64/bin/python is missing [requires (has)]: tox-wheel>=0.6.0
ROOT: provision> .tox/.tox/bin/python -m tox
/home/runner/work/twisted/twisted/.tox/.tox/bin/python: No module named tox
Error: Process completed with exit code 1.
```

Just run any test on our CI.

See a build run here 

https://github.com/twisted/twisted/actions/runs/3574757910/jobs/6212002907#step:10:12

**Additional context**

I guess that this is due to tox 4.

I previous run of the same commit, was green and was using "Using cached tox-3.27.1-py2.py3-none-any.whl (86 kB)"

The failing CI using "Downloading tox-4.0.8-py3-none-any.whl"

adiroiban:
## Scope and purpose

Fixes #11788

This tries to get our CI green again with the least effort.

In a separate PR we can look at removing our dependence on tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,6 +20,7 @@ minversion=3.24.1
 requires=
     virtualenv>=20.7.2
     tox-wheel>=0.6.0
+    tox < 4
 skip_missing_interpreters=True
 envlist=lint, mypy,
     apidocs, narrativedocs, newsfragment,
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -213,7 +213,7 @@ jobs:
     - uses: twisted/python-info-action@v1
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip tox coverage
+        python -m pip install --upgrade pip tox=~3.0 coverage
 
     # Make sure the matrix is defined in such a way that this is triggered
     # only on Linux.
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -20,6 +20,7 @@ minversion=3.24.1
 requires=
     virtualenv>=20.7.2
     tox-wheel>=0.6.0
+    tox < 4
```
adiroiban, line 23: Even if you run tox v3, when it creates ~/.tox/.tox venv, it will create it with tox v4.

This looks like it can fix it.

adiroiban:
This is ready for review. Thanks!

needs-review

glyph (approved):
We should really develop a comprehensive strategy for pinning *everything*, and not just with inequalities; we should use exact versions everywhere (except for our package metadata) and let dependabot send upgrade PRs.

But, let's get trunk back to passing ASAP.

glyph:
> In a separate PR we can look at removing our dependence on tox.

What would you suggest replacing it with? `nox`?

PR #11741, opened 2022-11-03
#11740 Update minimum dependecies.

adiroiban:
## Scope and purpose

Fixes #11740
Fixes #11737

it looks like our package metadata has wrong minium deps.

Try to add a CI that can detect these errors.

Update the dependencies to actual minimum versions.

---------

I have reused the noIP6 jobs for doing mindeps.
We can have a separate job, but I think we already have a lot of jobs.

I have added the mindesp in tox.

## Contributor Checklist:

This process applies to *all* pull requests - no matter how small.
Have a look at [our developer documentation](https://docs.twisted.org/en/latest/core/development/dev-process.html) before submitting your Pull Request.

Below is a non-exhaustive list (as a reminder):

* The title of the PR should describe the changes and starts with the associated issue number, like “#1234 Brief description”.
* A release notes news fragment file was create in src/twisted/newsfragments/ (see: [Release notes fragments docs.](https://docs.twisted.org/en/latest/core/development/dev-process.html#release-notes-management))
* The automated tests were updated.
* Once all checks are green, request a review by leaving a comment that contains exactly the string `please review`.
  Our bot will trigger the review process, by applying the pending review label
  and requesting a review from the Twisted dev team.

Issue #11737, opened 2022-11-01
Dep on attrs should be >= 21.3.0


twm:
`import attrs` (vs. `import attr`) was introduced in [attrs 21.3.0](https://www.attrs.org/en/stable/changelog.html#id17).

We currently depend on attrs >= 19.2.0:

https://github.com/twisted/twisted/blob/8e7857d08355273179ca330c55f41713c65254c2/pyproject.toml#L38

We import from the `attrs` module in many places:

```
src/twisted/python/systemd.py:from attrs import Factory, define
src/twisted/trial/_synctest.py:from attrs import frozen
src/twisted/trial/runner.py:from attrs import define
docs/web/howto/listings/render_slots_attrs.py:from slots_attributes_1 import ExampleElement
docs/web/howto/listings/render_slots_attrs.py:from twisted.web.template import flattenString
src/twisted/application/runner/_runner.py:from attr import Factory, attrib, attrs
src/twisted/trial/_dist/disttrial.py:from attrs import define, field, frozen
src/twisted/trial/_dist/disttrial.py:from attrs.converters import default_if_none
src/twisted/trial/_dist/stream.py:from attrs import Factory, define
src/twisted/trial/_dist/worker.py:from attrs import frozen
src/twisted/trial/_dist/workerreporter.py:from attrs import Factory, define
src/twisted/application/runner/test/test_runner.py:from attr import Factory, attrib, attrs
src/twisted/trial/_dist/test/test_disttrial.py:from attrs import Factory, assoc, define, field
```

We should update our minimum dep.

adiroiban:
A similar issue with our invalid minimum deps at #11740

I will create a PR and see how we can detect such errors

adiroiban:
Thanks for the suggestion. I have applied the updated attr version here https://github.com/twisted/twisted/pull/11741

adiroiban:
Thanks for the suggestion. I have applied the updated attr version here https://github.com/twisted/twisted/pull/11741

PR #11790, opened 2022-12-19
Fixed missing import of attrs

Collert:
File included an import of attrs, incorrect import of which would throw `ModuleNotFoundError: No module named 'attrs'`. Spelling it this way seems to resolve the issue.


exarkun:
Thanks.  The "attrs" import is correct.  This looks like a duplicate of #11737 which was resolved in https://github.com/twisted/twisted/pull/11741

PR #11794, opened 2023-01-05
#11793 deprecated legacy ciphers for conch

### LINKED ISSUE #11793: some ciphers for conch triggered deprecation warning with `cryptography>=37`
Some ciphers triggered deprecation warning with `cryptography>=37` when using conch. They are
* `CAST5`
* `SEED`
* `IDEA`
* `Blowfish`
from `cryptography.hazmat.primitives.ciphers.algorithms`.

`cryptography>=37` has deprecated these ciphers which will be removed in a future version.
More details https://github.com/pyca/cryptography/pull/6922.

Deprecation warning should be gone.

**Testing environment**
`Twisted==22.10.0` on Mac 12.6 (21G115).

llouislu:
Fixes #11793

This PR deprecates legacy ciphers (CAST5, Blowfish) used in conch and updates documentation.


src/twisted/conch/ssh/transport.py
```diff
@@ -94,17 +94,13 @@ class SSHCiphers:
 
     cipherMap = {
         b"3des-cbc": (algorithms.TripleDES, 24, modes.CBC),
-        b"blowfish-cbc": (algorithms.Blowfish, 16, modes.CBC),
```
adiroiban, line 97: Just a question.

Can we have them still supported, but removed by default?

src/twisted/conch/ssh/transport.py
```diff
@@ -284,10 +280,6 @@ def _getSupportedCiphers():
         b"aes192-cbc",
         b"aes128-ctr",
         b"aes128-cbc",
-        b"cast128-ctr",
```
adiroiban, line 287: I don't think that this change is ok.

The Twisted backward compatibility state that before removing something we first start by raising a dedicated deprecation warning, informing the users that this feature will be removed.
A new Twisted version should be released with this version.

Then, after 2 other releases the functionality can be removed.

More info here https://docs.twisted.org/en/stable/development/compatibility-policy.html

adiroiban (changes requested):
Many thanks for the changes and your help!

I think that before removal, we should first deprecate the SSH ciphers in Twisted Conch.

Even though there is a warning in cryptography, this is not a Twisted warning, and you might be using an older cryptography version in which the warning is not raised.

This is why I think it is required to raise a dedicated warning in Twisted.

The Twisted deprecation and removal policy is here https://docs.twisted.org/en/stable/development/compatibility-policy.html

There is the option of removing code without a prior warning, but this will need an explicit policy exception request and approval from the team.

-----------

Triggering a valid warning is not easy.
I assume that many Conch user might have blowfish enabled, but this is done via the default configuration ann not via an explicit configuration.

And we don't have a standard explicit configuration for ciphers in conch... so is hard  to tell if someone is using a cipher because it is needed or because it was enabled by default.

So I guess we can just go with a generic warning.

But we need to raise a warning for Twisted, and not rely only on the warning from cryptography.

The backward policy is an important part of Twisted and the Twisted dev team aim is to create a Twisted library that is stable and that will not crash at the next release due to unknown reasons.

------------

I hope my feedback makes sense.

If you think that my judgment is not correct, please resubmit for review.

Thanks again for the PR!

src/twisted/conch/ssh/transport.py
llouislu: Yes. Deprecated ciphers are in another list/dict while the original cipher list is kept.

src/twisted/conch/ssh/transport.py
llouislu: I have refactored this to a deprecation warning with a test case.

llouislu:
@adiroiban I've refactored it to a deprecation. Please review again.

src/twisted/conch/ssh/transport.py
```diff
@@ -70,6 +71,15 @@ class _MACParams(tuple):
     """
 
 
+warnings.warn(
+    "Legacy SSH ciphers 'CAST5', 'Blowfish' "
+    "were deprecated in Twisted 22.11. "
```
adiroiban: I don't know if raising a warning at import time is the best option...

but I guess that if this is imported it means that SSH is also used.

My concern is to avoid raising the warning if you just have Twisted but for example you use it only for HTTPS.

But I think this is fine.

The warning here, should have the advantage that it is only raised once.

--------------

This needs to use a placeholder.

Version 22.11 is wrong.

```suggestion
    "were deprecated in Twisted NEXT. "
```

Also, maybe move it a bit at the top.

I am not sure what to update.

If you use those ciphers, you should not update :)

The message is a bit confusing.

And what is a "legacy" cipher ? 

Maybe just keept is short.

```
The ciphers based on CAST5 and Blowfish were deprecated in Twisted NEXT and will be removed in a future release.
```

src/twisted/conch/ssh/transport.py
```diff
@@ -92,21 +102,28 @@ class SSHCiphers:
     @ivar inMAc: see outMAC, but for the incoming MAC.
     """
 
+    deprecatedCipherMap = {
```
adiroiban: This is a new public API that needs documentation.

I don't understand why these dicts were created.

I would prefer not to introduce new API for the purpose of the deprecation PR.

src/twisted/conch/test/test_transport.py
```diff
@@ -3013,6 +3015,34 @@ def test_makeMAC(self):
...
+        self.assertEqual(warningsShown[0]["category"], DeprecationWarning)
+        self.assertEqual(
+            warningsShown[0]["message"],
+            "Legacy SSH ciphers 'CAST5', 'Blowfish' "
+            "were deprecated in Twisted 22.11. "
+            "Please check and update if any is in use.",
+        )
+        majorVersion = int(cryptography.__version__.split(".")[0])
+        expectedNumberOfWarnings = 1
+        if majorVersion >= 37:
```
adiroiban, line 3042: I think that we need a comment to explain why we have this separate branch,

src/twisted/conch/test/test_transport.py
```diff
@@ -3013,6 +3015,34 @@ def test_makeMAC(self):
                 f"Failed HMAC test vector; key={key!r} data={data!r}",
             )
 
+    def test_deprecation(self):
+        """
+        Test deprecation warning for legacy ciphers `CAST5`, `Blowfish`.
+
+        A deprecation warning is emitted when directly importing the
+        SSHCiphers class.
+        """
```
adiroiban: Maybe the descript can be brief.

I don't think that SSHCiphers is relevant.
From the code, it looks like the warning is raised even if SSHCiphers is not used.

```suggestion
    def test_deprecationCAST5Blowfish(self):
        """
        A deprecation warning is raised when importing the L{twisted.conch.transport} module.
        """
```

src/twisted/newsfragments/11793.removal
```diff
@@ -0,0 +1 @@
+Twisted Conch has deprecated but still supports legacy ciphers (CAST5, SEED, IDEA, Blowfish) as the upstream library "cryptography" has deprecated them since version 37.
```
adiroiban: Are we deprecating SEED and IDEA? From the code I see only CAST5 and Blowfish.

Maybe the release notes can be:

twisted.conch.ssh.transport has the CAST5 and Blowfish based cipher deprecated. This was done since the same ciphers were deprecated in the `cryptography` library since version 37.

adiroiban (changes requested):
Many thanks for the changes and the test.

I think the changes are in the right direction.

Is important to use the `NEXT` placeholder to denote the version.

Also, it would help not to introduce new public API.

If the public API is required, the it needs documentation.

Thanks again!

src/twisted/conch/ssh/transport.py
llouislu: These dict/lists are to group deprecated ciphers and this makes it outstanding and easy to find when they need to be removed in the future.

I have updated them as private.

src/twisted/newsfragments/11793.removal
llouislu: > Are we deprecating SEED and IDEA? From the code I see only CAST5 and Blowfish.

`cryptography` deprecates all 4 of these ciphers but I only found use of `blowfish` and `cast5` in twisted source code.

I've updated with your notes.

src/twisted/conch/test/test_transport.py
llouislu: i've added it.

llouislu:
@adiroiban Please review.

src/twisted/conch/ssh/transport.py
exarkun: > I don't know if raising a warning at import time is the best option...

Isn't reporting a deprecation warning about these ciphers _exactly_ the pre-existing behavior - because that's just what python-cryptography does?

I don't see who benefits from having all of these deprecations reported twice.

Just as bad, like the python-cryptography warnings, the Twisted warnings are 100% inactionable.  There is nothing you can do while using Twisted Conch for SSH to avoid these new warnings.  It doesn't matter if your code is affected or not, you will be shown the warnings.  This dilutes the value of warnings and encourages people to ignore them.

It would really be much better if the warnings were emitted when someone actually _used_ something that's deprecated.  This lets them get rid of the deprecation by removing that use.

src/twisted/conch/ssh/transport.py
```diff
@@ -92,21 +102,28 @@ class SSHCiphers:
...
     """
 
+    _deprecatedCipherMap = {
+        b"cast128-cbc": (algorithms.CAST5, 16, modes.CBC),
+        b"cast128-ctr": (algorithms.CAST5, 16, modes.CTR),
+        b"blowfish-ctr": (algorithms.Blowfish, 16, modes.CTR),
+        b"blowfish-cbc": (algorithms.Blowfish, 16, modes.CBC),
+    }
     cipherMap = {
+        # 3DES, ARC4 are also legacy (and ARC4 has serious security issues),
```
adiroiban, line 112: Why should we care here about ARC4 ?

And these days many automated SSH scanner will complain about any CBC mode...

I am not sure that we should add comment about these things here.

My suggestion is to just remove this comment.

adiroiban (commented):
I am only leaving a comment, since I don't know what to suggest here.

For example, I am using the twisted code with a custom list of ciphers

Basically I have something like this.

```diff
-supportedCiphers = _getSupportedCiphers()
+supportedCiphers = _getCustomCiphers()
```

I am not using blowfish (or 3DES or CBC), but with this change, I will still receive the deprecation warning.

And I guess that might be people using `supportedCiphers` as a property... so for example if they receive a connection from an IP range dedicate to legacy deviced they return a set of ciphers, while for everything else, they go with secure ciphers.

---------

It would be nice to have better support for custom ciphers list in Twisted ... but this should be done in a separate ticket/PR.

----------

At the same time, instead of removing a cipher, I would prefer to just have that cipher disabled by default.

For blowfish and IDEA this is not a big issue, as many embedded devices support AES-128.

--------

For example in a separate PR twisted removed a DH KEX Group... and I had to hack it back, since an old switch was only supporting that KEX...and using SSH is still better  than telnet :)

so I am ok to disable the support for something by default, but it should not be removed. 

-----------

so maybe the deprecation should only be raised in `SSHTransportBase.sendKexInit` if `supportedCiphers` by that time still includes the deprecated ciphers.

----------

But this is not a complete rejection.

I don't really mind the extra deprecation warning.

Is just that maybe this is  good opportunity to improve the design and the extensibility of Twisted conch ssh.

But maybe, I am the only one with strange requirements for SSH... so we should not bother and just merge this code.

adiroiban (changes requested):
Thanks for the changes.

I have re-read Jean Paul's comment

> It would really be much better if the warnings were emitted when someone actually used something that's deprecated. This lets them get rid of the deprecation by removing that use.

I agree with him.

So I am requesting changes.

@llouislu I am happy to help with the code for a better designed conch code.

So for this PR, we can discuss your use case, try to understand what problem it tries to solve and see how we can improve Twisted conch.

Thanks!

Issue #5275, opened 2011-09-26
t.w.template does not always properly escape comments


twisted-trac:
|[<img alt="bitglue's avatar" src="https://avatars.githubusercontent.com/u/1232472?s=50" width="50" height="50">](https://github.com/bitglue)| @bitglue reported|
|-|-|
|Trac ID|trac#5275|
|Type|defect|
|Created|2011-09-26 17:43:15Z|
|Branch|https://github.com/twisted/twisted/tree/template-comment-fix-5275|

```
>>> from twisted.web.template import flattenString, Comment
>>> from twisted.python.util import println
>>> flattenString(None, Comment('the quick---brown fox')).addCallback(println)
<!--the quick- --brown fox-->
<Deferred at 0x1020d6a28 current result: None>
```

Which of course is malformed HTML, because HTML is SGML, and comments in SGML begin and end with '--', and are between '<!' and '>'. Or something like that.

What this generates is a comment declaration with one comment, 'the quick- ', and some illegal characters 'brown fox--'.

http://htmlhelp.com/reference/wilbur/misc/comment.html

Attachments:

* [template-comments-5275.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/451/451e7470b4f5da501cd95543e8a874c4fb1aa542/39c1449b406e0b1ffaa4c2c9144b90783281e9d0.patch) (2065 bytes) - added by indigo on 2011-09-26 18:13:56Z - proposed fix
* [template-comments-5275.2.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/451/451e7470b4f5da501cd95543e8a874c4fb1aa542/ce4df89315ef0c5e13f42e5ed45f6692179fa1ac.patch) (2342 bytes) - added by indigo on 2011-09-26 18:18:03Z - 
* [template-comments-5275.3.patch](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/451/451e7470b4f5da501cd95543e8a874c4fb1aa542/7f03e3e4645be4f12bcee6fd5cbda030a43ab58d.patch) (2654 bytes) - added by indigo on 2011-09-26 18:37:31Z - 

<details><summary>Searchable metadata</summary>

```
trac-id__5275 5275
type__defect defect
reporter__indigo indigo
priority__normal normal
milestone__ 
[...]

PR #11805, opened 2023-02-01
#11804 Remove > character replacement in escaped comments

### LINKED ISSUE #11804: The > sign should not be escaped within conditional CSS
Email templates for MS Outlook require conditional CSS. A line of code for that looks like this:  `<!--[if (gte mso 9)|(IE)]>`
That is an if block that will be called if the version of MSOutlook receiving the email is greater than 9 or if it’s any version of Internet Explorer.

Emails sent to MSO users are displaying the code for the conditional CSS:
Screenshot 2023-02-01 at 1 24 43 PM

If you look closely you’ll see that the line of code that I used as an example looks like this: ` <!--[if (gte mso 9)|(IE)]&gt;` 
The transformation of `>` into it's HTML counterpart, `&gt;` is not allowing the if block to adequately close.

The source of that transformation comes from: `twisted.web._flatten.escapedComment` where this line of code exists: ` data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")` 

That was introduced in [this PR](https://github.com/twisted/twisted/commit/478617aab2edc7f0ac55360ab4d0105374a62170)

In the docstring of `test_commentEscaping` this statement was made: "Comments start with  and never contain -- or >." That is not correct given that we close if statements with a `>`
i.e. <!--[if (gte mso 9)|(IE)]`>`

Also, the referenced XML spec defines comment as:
```
Char	   ::=   	#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]	/* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */
Comment	   ::=   	'<!--' ((Char - '-') | ('-' (Char - '-')))* '-->'
```
And `>` would be  `#x3E`.

To solve this issue we patched `twisted.web._flatten.escapeComment` at runtime, removing `.replace(b">", b"&gt;")`

I am creating this issue to introduce the problem and subsequently I will submit a PR that aims to solve the problem. 

**Testing environment**
- Python version 3.10.4
- Twisted version 22.10.0
- ProductName:         macOS
   ProductVersion:      13.1
   BuildVersion:           22C65
   
$ pip freeze
alabaster==0.7.13
appdirs==1.4.4
astor==0.8.1
astroid==2.14.1
attrs==22.2.0
Automat==22.10.0
Babel==2.11.0
CacheControl==0.12.11
cachetools==5.3.0
certifi==2022.12.7
cfgv==3.3.1
chardet==5.1.0
charset-normalizer==3.0.1
click==8.1.3
click-default-group==1.2.2
colorama==0.4.6
ConfigArgParse==1.5.3
constantly==15.1.0
coverage==6.6.0b1
dill==0.3.6
distlib==0.3.6
docutils==0.19
exceptiongroup==1.1.0
filelock==3.9.0
hyperlink==21.0.0
hypothesis==6.65.2
identify==2.5.17
idna==3.4
imagesize==1.4.1
incremental==22.10.0
isort==5.12.0
Jinja2==3.1.2
lazy-object-proxy==1.9.0
lockfile==0.12.2
lunr==0.6.2
MarkupSafe==2.1.2
mccabe==0.7.0
msgpack==1.0.4
nodeenv==1.7.0
packaging==23.0
platformdirs==2.6.2
pluggy==1.0.0
pre-commit==3.0.2
pydoctor==22.9.1
pyflakes==2.5.0
Pygments==2.14.0
PyHamcrest==2.0.4
pylint==2.16.0
pyproject_api==1.5.0
pytz==2022.7.1
PyYAML==6.0
readthedocs-sphinx-ext==2.2.0
requests==2.28.2
six==1.16.0
snowballstemmer==2.2.0
sortedcontainers==2.4.0
Sphinx==5.3.0
sphinx-rtd-theme==0.5.1
sphinxcontrib-applehelp==1.0.4
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==2.0.1
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.5
toml==0.10.2
tomli==2.0.1
tomlkit==0.11.6
towncrier==22.12.0
tox==4.4.4
-e git+https://github.com/twisted/twisted.git@59783c7c2e40868283395b2a88125618b254631c#egg=Twisted
twistedchecker==0.7.4
typing_extensions==4.4.0
urllib3==1.26.14
virtualenv==20.17.1
wrapt==1.14.1
zope.interface==5.5.2

ghost:
Fixes #11804

Email templates for MS Outlook require conditional CSS. The closing tag for those is a `>` character, and `_flatten.escapedComment` was replacing that character for its HTML counterpart. That was in turn preventing the conditional CSS blocks from closing causing the ghost code to appear superimposed in the body of Outlook emails. 

Emails sent to MSO users are displaying the code for the conditional CSS:
Screenshot 2023-02-01 at 1 24 43 PM

If you look closely you’ll see that the conditional `if` block looks like this: ` <!--[if (gte mso 9)|(IE)]&gt;`, and that the closing bracket, `>`, has been converted to `&gt;`.
As a result, the transformation of `>` into it's HTML counterpart, `&gt;` is not allowing the if block to adequately close.

The source of that transformation was introduced in [this PR](https://github.com/twisted/twisted/commit/478617aab2edc7f0ac55360ab4d0105374a62170)

In the docstring of `test_commentEscaping` this statement was made: "Comments start with  and never contain -- or >." That is not correct given that we close if statements with a `>`
i.e. <!--[if (gte mso 9)|(IE)]`>`

[Outlook conditional CSS overview](https://stackoverflow.design/email/base/mso/)

--------------------------

To solve this issue, this PR adds a helper function that detects if the code being parsed contains Outlook conditional CSS, and then uses that to determine whether the `>` character should be escaped or not when parsing comments. It also adds tests to make sure the behavior is correct and does not break/alter anything else.

### DIFF
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,1 +1,1 @@
-twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments.
+The (>) character within comments is no longer escaped by twisted.web.template.flattenString. This update allows the correct parsing of Microsoft Outlook conditional comments.
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,1 +1,1 @@
-twisted.web.template.flattenString now recognizes and correctly parses MSO content.
+twisted.web.template.flattenString now distinguishes and correctly parses MSO content. 
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,4 +1,4 @@
 twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments. 
 Outlook conditional comments allow us to add bits of HTML that are only read by the Word-based versions of Outlook.
-This fix is necessary for those creating email templates that include recipients that use Windows Outlook 2003 and above because that system uses Microsoft Word as a rendering engine, which can lead to some weird rendering issues.
+This fix is necessary for those creating email templates for recipients that use Windows Outlook 2003 and above because that system uses Microsoft Word as a rendering engine, which can lead to some weird rendering issues.
 In sum, we can now use MSO (Microsoft Office) tags to add HTML / CSS anywhere in an email template. 
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,0 +1,1 @@
+twisted.web.template.flattenString now recognizes and correctly parses MSO content.
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,1 +1,4 @@
-twisted.web.template.flattenString now distinguishes and correctly parses MSO content. 
+twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments. 
+Outlook conditional comments allow us to add bits of HTML that are only read by the Word-based versions of Outlook.
+This fix is necessary for those creating email templates that include recipients that use Windows Outlook 2003 and above because that system uses Microsoft Word as a rendering engine, which can lead to some weird rendering issues.
+In sum, we can now use MSO (Microsoft Office) tags to add HTML / CSS anywhere in an email template. 
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,4 +1,1 @@
-twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments. 
-Outlook conditional comments allow us to add bits of HTML that are only read by the Word-based versions of Outlook.
-This fix is necessary for those creating email templates for recipients that use Windows Outlook 2003 and above because that system uses Microsoft Word as a rendering engine, which can lead to some weird rendering issues.
-In sum, we can now use MSO (Microsoft Office) tags to add HTML / CSS anywhere in an email template. 
+twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments.
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,1 +1,1 @@
-To prevent parsing errors and ensure validity when serializing HTML comments, twisted.web.template.flattenString has been updated to escape the --> sequence within comments
+To prevent parsing errors and ensure validity when serializing HTML comments, twisted.web.template.flattenString has been updated to escape the --> sequence within comments.
--- a/src/twisted/newsfragments/11804.bugfix
+++ b/src/twisted/newsfragments/11804.bugfix
@@ -1,1 +1,1 @@
-The (>) character within comments is no longer escaped by twisted.web.template.flattenString. This update allows the correct parsing of Microsoft Outlook conditional comments.
+To prevent parsing errors and ensure validity when serializing HTML comments, twisted.web.template.flattenString has been updated to escape the --> sequence within comments
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -166,32 +166,42 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
-def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
-    if isinstance(html_string, bytes):
-        html_string = html_string.decode("utf-8")
+def _hasMSOComments(data: bytes) -> bool:
+    """
+    Check if the input C{data} contains Microsoft Office (MSO) comments.
 
-    if not html_string.startswith("[if "):
+    MSO comments are conditional comments that are only interpreted by
+        Microsoft Office applications, and not other email clients.
+
+    The format of MSO conditional comments is as follows:
+        "<!--[if condition]> ... content ... <![endif]-->"
+
+    @param data: The data (bytes) to check for MSO comments.
+
+    @return: A boolean indicating whether the C{data} has MSO comments or not.
+    """
+    if not data.startswith(b"[if "):
         return False
 
-    start_index = html_string.find("[if ")
-    section_to_check = html_string[start_index:]
+    start_index = data.find(b"[if ")
+    section_to_check = data[start_index:]
 
     mso_operator = section_to_check[4:]
     if any(
         mso_operator.startswith(prefix)
         for prefix in [
-            "gt mso",
-            "lt mso",
-            "gte mso",
-            "lte mso",
-            "mso",
-            "!mso",
-            "(!mso",
-            "IE",
-            "!IE",
-            "(!IE",
-            "(mso",
-            "(IE",
+            b"mso",
+            b"gt mso",
+            b"lt mso",
+            b"gte mso",
+            b"lte mso",
+            b"!mso",
+            b"(mso",
+            b"(!mso",
+            b"IE",
+            b"!IE",
+            b"(!IE",
+            b"(IE",
         ]
     ):
         return True
@@ -201,7 +211,7 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
 
 def escapedComment(data: Union[bytes, str]) -> bytes:
     """
-    Escape a comment for inclusion in a document.
+    Escape a comment for inclusion in a document, unless it is an MSO conditional comment.
 
     @param data: The string to escape.
 
@@ -210,9 +220,7 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    if _hasMSOComments(data):
-        data = data.replace(b"--", b"- - ")
-    else:
+    if not _hasMSOComments(data):
         data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
     if data and data[-1:] == b"-":
         data += b" "
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -178,25 +178,9 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     section_to_check = html_string[start_index:]
 
     mso_operator = section_to_check[4:]
-    if mso_operator.startswith("gt mso"):
-        return True
-    elif mso_operator.startswith("lt mso"):
-        return True
-    elif mso_operator.startswith("gte mso"):
-        return True
-    elif mso_operator.startswith("lte mso"):
-        return True
-    elif mso_operator.startswith("mso"):
-        return True
-    elif mso_operator.startswith("!mso"):
-        return True
-    elif mso_operator.startswith("IE"):
-        return True
-    elif mso_operator.startswith("!IE"):
-        return True
-    elif mso_operator.startswith("(mso"):
-        return True
-    elif mso_operator.startswith("(IE"):
+    if any(mso_operator.startswith(prefix) for prefix in [
+        "gt mso", "lt mso", "gte mso", "lte mso", "mso", "!mso", "IE", "!IE", "(mso", "(IE"
+    ]):
         return True
     else:
         return False
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -186,8 +186,10 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
             "lte mso",
             "mso",
             "!mso",
+            "(!mso",
             "IE",
             "!IE",
+            "(!IE",
             "(mso",
             "(IE",
         ]
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -7,7 +7,6 @@ Context-free flattener/serializer for rendering Python objects, possibly
 complex or arbitrarily nested, as strings.
 """
 
-import re
 from inspect import iscoroutine
 from io import BytesIO
 from sys import exc_info
@@ -177,6 +176,7 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     start_index = html_string.find("[if ")
     section_to_check = html_string[start_index:]
 
+    mso_operator = section_to_check[4:]
     mso_operator = section_to_check[4:]
     if any(
         mso_operator.startswith(prefix)
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -177,7 +177,7 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    data = data.replace(b"--", b"- - ")
+    data = data.replace(b"-->", b"--&gt;")
     if data and data[-1:] == b"-":
         data += b" "
     return data
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -167,7 +167,7 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
-def has_mso_comments(html_string: Union[str, bytes]) -> bool:
+def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     """
     MSO tags contain either 'mso' or 'IE' within brackets that start with '[if '.
     """
@@ -187,7 +187,7 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    if has_mso_comments(data):
+    if _hasMSOComments(data):
         data = data.replace(b"--", b"- - ")
     else:
         data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -177,7 +177,6 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     section_to_check = html_string[start_index:]
 
     mso_operator = section_to_check[4:]
-    mso_operator = section_to_check[4:]
     if any(
         mso_operator.startswith(prefix)
         for prefix in [
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -167,7 +167,7 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
-def has_mso_comments(html_string):
+def has_mso_comments(html_string: Union[str, bytes]) -> bool:
     """
     MSO tags contain either 'mso' or 'IE' within brackets that start with '[if '.
     """
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -178,7 +178,7 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
+    data = data.replace(b"--", b"- - ")
     if data and data[-1:] == b"-":
         data += b" "
     return data
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -178,9 +178,21 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     section_to_check = html_string[start_index:]
 
     mso_operator = section_to_check[4:]
-    if any(mso_operator.startswith(prefix) for prefix in [
-        "gt mso", "lt mso", "gte mso", "lte mso", "mso", "!mso", "IE", "!IE", "(mso", "(IE"
-    ]):
+    if any(
+        mso_operator.startswith(prefix)
+        for prefix in [
+            "gt mso",
+            "lt mso",
+            "gte mso",
+            "lte mso",
+            "mso",
+            "!mso",
+            "IE",
+            "!IE",
+            "(mso",
+            "(IE",
+        ]
+    ):
         return True
     else:
         return False
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -168,12 +168,54 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
 
 
 def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
-    """
-    MSO tags contain either 'mso' or 'IE' within brackets that start with '[if '.
-    """
     if isinstance(html_string, bytes):
         html_string = html_string.decode("utf-8")
-    return re.search(r"\[if .*(mso|IE)", html_string) is not None
+
+    if "[if " not in html_string:
+        return False
+
+    start_index = html_string.find("[if ")
+    section_to_check = html_string[start_index:]
+
+    mso_operator = section_to_check[4:]
+    if mso_operator.startswith("gt "):
+        mso_operator = mso_operator[3:]
+        if mso_operator.startswith("mso"):
+            return True
+        else:
+            return False
+    elif mso_operator.startswith("lt "):
+        mso_operator = mso_operator[3:]
+        if mso_operator.startswith("mso"):
+            return True
+        else:
+            return False
+    elif mso_operator.startswith("gte "):
+        mso_operator = mso_operator[4:]
+        if mso_operator.startswith("mso"):
+            return True
+        else:
+            return False
+    elif mso_operator.startswith("lte "):
+        mso_operator = mso_operator[4:]
+        if mso_operator.startswith("mso"):
+            return True
+        else:
+            return False
+    elif mso_operator.startswith("mso"):
+        return True
+    elif mso_operator.startswith("!mso"):
+        return True
+    elif mso_operator.startswith("IE"):
+        return True
+    elif mso_operator.startswith("!IE"):
+        return True
+    elif mso_operator.startswith("(mso"):
+        return True
+    elif mso_operator.startswith("(IE"):
+        return True
+    else:
+        return False
 
 
 def escapedComment(data: Union[bytes, str]) -> bytes:
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -170,7 +170,7 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     if isinstance(html_string, bytes):
         html_string = html_string.decode("utf-8")
 
-    if "[if " not in html_string:
+    if not html_string.startswith("[if "):
         return False
 
     start_index = html_string.find("[if ")
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -178,30 +178,14 @@ def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
     section_to_check = html_string[start_index:]
 
     mso_operator = section_to_check[4:]
-    if mso_operator.startswith("gt "):
-        mso_operator = mso_operator[3:]
-        if mso_operator.startswith("mso"):
-            return True
-        else:
-            return False
-    elif mso_operator.startswith("lt "):
-        mso_operator = mso_operator[3:]
-        if mso_operator.startswith("mso"):
-            return True
-        else:
-            return False
-    elif mso_operator.startswith("gte "):
-        mso_operator = mso_operator[4:]
-        if mso_operator.startswith("mso"):
-            return True
-        else:
-            return False
-    elif mso_operator.startswith("lte "):
-        mso_operator = mso_operator[4:]
-        if mso_operator.startswith("mso"):
-            return True
-        else:
-            return False
+    if mso_operator.startswith("gt mso"):
+        return True
+    elif mso_operator.startswith("lt mso"):
+        return True
+    elif mso_operator.startswith("gte mso"):
+        return True
+    elif mso_operator.startswith("lte mso"):
+        return True
     elif mso_operator.startswith("mso"):
         return True
     elif mso_operator.startswith("!mso"):
--- a/src/twisted/web/_flatten.py
+++ b/src/twisted/web/_flatten.py
@@ -166,52 +166,9 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
-def _hasMSOComments(data: bytes) -> bool:
-    """
-    Check if the input C{data} contains Microsoft Office (MSO) comments.
-
-    MSO comments are conditional comments that are only interpreted by
-        Microsoft Office applications, and not other email clients.
-
-    The format of MSO conditional comments is as follows:
-        "<!--[if condition]> ... content ... <![endif]-->"
-
-    @param data: The data (bytes) to check for MSO comments.
-
-    @return: A boolean indicating whether the C{data} has MSO comments or not.
-    """
-    if not data.startswith(b"[if "):
-        return False
-
-    start_index = data.find(b"[if ")
-    section_to_check = data[start_index:]
-
-    mso_operator = section_to_check[4:]
-    if any(
-        mso_operator.startswith(prefix)
-        for prefix in [
-            b"mso",
-            b"gt mso",
-            b"lt mso",
-            b"gte mso",
-            b"lte mso",
-            b"!mso",
-            b"(mso",
-            b"(!mso",
-            b"IE",
-            b"!IE",
-            b"(!IE",
-            b"(IE",
-        ]
-    ):
-        return True
-    else:
-        return False
-
-
 def escapedComment(data: Union[bytes, str]) -> bytes:
     """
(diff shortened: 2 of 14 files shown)


my-logic:
The "stderr" section mentions a "RuntimeError" with a message that the "Poetry configuration is invalid". This seems to be related to a conflict with the "pip-shims" package. 

This issue can be solved via: https://github.com/PyCQA/isort/releases/tag/5.12.0

my-logic:
please review

adiroiban:
Many thanks for the PR. Don't forget to remote the "Draft marker" once this is ready for review.

Also, this needs a release notes fragment, in the form of a file like `src/twisted/newsfragmetns/11804.bugfix` that will describe the change library's to end users.

I am not sure that this change is ok.

From that I see in the previous PR, the comments were transformed into normal content in order to keep the XML structure valid.

I checked this SO question https://stackoverflow.com/questions/2640453/single-line-comment-in-html  ... but the focus there is mostly on HTML, while the first answer is generic SGML

and reviewed the initial issue report https://github.com/twisted/twisted/issues/5275

-------

The change might be ok, but it need better testing.

It needs at least one new test that explains the new behaviour

-------------

My undestanding of this issue is this.

The current twisted code tried to make "invalid SGML" comments into valid "SGML content"
But from what I see, it converts any comment (valid or invalid) into SGML content.

So. One way to fix this is to recognizie invalid comments and convert them into content.
While, anything that looks like valid SGML comment should be kept in place.

I am not an expert into XML/HTML markup, so we might need to wait for another team member to take a look

Thanks again

felipemelendez:
Thank you for your prompt response, I will include the release notes fragment. Also, we are parsing XHTML and solved this issue by monkey patching `_flatten.escapedComment`, but following your arguments, it seems that I might have disregarded the fact that this method is also used to escape HTML comments? If that is the case the method needs to be further developed to take this into consideration, and conditionally escape the `>` depending on whether the content is HTML or XML, and a test needs to follow suit.

adiroiban (changes requested):
Based on latest conversation this needs changes.

Let me know when this is ready again for a new revew.

Thanks again

felipemelendez:
After careful consideration, the problem to address was the detection and proper handling of MSO (Microsoft Office) conditional CSS, and not XML code as a whole. 
To that end I have created a helper function that detects if the data parsed by `escapedComment` contains MSO tags so that it may correctly escape the appropriate characters. 
I read several docs and this one summarizes the requirements: [Outlook Conditional CSS](https://stackoverflow.design/email/base/mso/) rules of engagement 
In addition, I revamped the test for `escapedComment`
I also tested my work with production.

felipemelendez:
please review

src/twisted/web/test/test_flatten.py
```diff
@@ -264,6 +264,32 @@ def test_serializeComment(self) -> None:
         """
         self.assertFlattensImmediately(Comment("foo bar"), b"<!--foo bar-->")
 
+    def test_hasMSOComments(self):
```
my-logic, line 267: hmm, I think I got a little carried away with testing - this might be redundant given the test coverage I added to `test_commentEscaping`, but I don't want to keep making modifications after I requested a review (sorry about that). Reviewer, please let me know if you think this test is overkill.

felipemelendez, line 267: hmm, I think I got a little carried away with testing - this might be redundant given the test coverage I added to `test_commentEscaping`, but I don't want to keep making modifications after I requested a review (sorry about that). Reviewer, please let me know if you think this test is overkill.

src/twisted/web/test/test_flatten.py
```diff
@@ -272,41 +298,50 @@ def test_commentEscaping(self) -> Deferred[List[bytes]]:
...
+        whereas Outlook conditional comments contain their logic within brackets 
+        that follow this format <!-- [if foo]>
 
         Also by XML syntax, a comment may not end with '-'.
 
         @see: U{http://www.w3.org/TR/REC-xml/#sec-comments}
         """
 
-        def verifyComment(c: bytes) -> None:
+        def verifyComment(z: bytes) -> None:
```
felipemelendez: because `c` was used as a variable name, whenever I tested the output with a breakpoint and pressed `c` to check the content of the variable "continue" was activated. Perhaps that is why I mishandled my revision of the test cases? Not sure, but for some reason I woke up with that realization, checked, and found an issue with the test - this now works as expected.

felipemelendez:
#11809 will fix the `isort` errors that are causing the CI checks to break

felipemelendez:
Well @adiroiban, sorry for all the changes and adds that I made after I submitted for review, I realize that this PR is a little chaotic but as you can see my intentions were good. Now I am really done making changes, at least until after the PR gets reviewed. Hope all is well

adiroiban:
Many thanks for the changes.

I have never used Twisted XML handling, so this is just a very generic review.

------------

Just an overall comment.

I am not very happy with the design of this change.

Instead of adding to the core Twisted core library all kind of exceptions for all kind of applications, I would love to see an implementation that would allow plugging my own exceptions.

I see that `escapedComment` is called by `_flattenElement` without too much context... which is called by `_flattenTree`, which is called by `flatten` 

Since I have never used this code, is hard to make any valid suggestion.

The only design change that I could think of is trying to implement some sort of "raw" element.

```python
    elif isinstance(root, Raw):
       write(root.data)
    elif isinstance(root, Comment):
        write(b"<!--")
        write(escapedComment(root.data))
        write(b"-->")
```

But maybe that will not work.

```python
    elif isinstance(root, Tag) and root.tagName == '__TWISTED_RAW__':
        write(root.data)
    elif isinstance(root, Comment):
        write(b"<!--")
        write(escapedComment(root.data))
        write(b"-->")
    elif isinstance(root, Tag):
        slotData.append(root.slotData)
```

but maybe all those suggestions make no sense.
As I mentions at the start, I have never used this code and I have no idea how you are using it :)

so ... I think that as long as this change doesn't introduce any regression and the code is well-tested and documented we can have it merged.

I see the test coverage is not 100% ... so this is a sign that the code is not too well tested :)

adiroiban:
The general code coverage is ok, but branch coverage is not complete

Here is the link

https://app.codecov.io/gh/twisted/twisted/pull/11805#f153fa5c03b7f5fb161078326a24ce23-R170


and in the case the link is not valid, a screenshot

image

src/twisted/web/test/test_flatten.py
adiroiban: why not rename it to "comment" or "targetedComment" or "expectedComment" or something that will explain what this variable is all about.

I am not a mathematician, and I don't like single letter variables :)

src/twisted/web/test/test_flatten.py
```diff
@@ -272,41 +272,57 @@ def test_commentEscaping(self) -> Deferred[List[bytes]]:
...
         results = []
-        for c in [
+        for z in [
             "",
             "foo---bar",
             "foo---bar-",
-            "foo>bar",
-            "foo-->bar",
             "----------------",
+            "foo > bar",
```
adiroiban, line 299: Instead of extending an existing test, I would love to see a new test dedicated especially to this use case.

The docstring of the new test can explain what is tested, why it tests and what future developers should take into consideration to make sure no regressions are added.

Also, with a separate case dedicated to this functionalty, i can just run and debug this single test and don't be bother by other use cases...so in the debugger, I will not have to press `continue` that many times :p

src/twisted/web/_flatten.py
```diff
@@ -167,6 +166,39 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
+def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
+    if isinstance(html_string, bytes):
```
adiroiban: this branch is not covered... it looks like html_string is always bytes in the tests.

src/twisted/newsfragments/11804.bugfix
```diff
@@ -0,0 +1,4 @@
+twisted.web.template.flattenString now recognizes and correctly parses Microsoft Outlook conditional comments.
```
adiroiban: I think that we should keep the release notes brief and describe the change at a very high level.

The information from here is very useful, but I think that it should be moved into our main "narative" documentation.

Something like here https://docs.twisted.org/en/stable/web/howto/twisted-templates.html#quoting

or add a separate small example to describe how the comments are handled.

In this way, if in the future someone has any question about how Twisted XML comments are escaped, it should be easy to discover, without having to read the whole code.... and this code is in a private part of Twisted.

src/twisted/web/_flatten.py
```diff
@@ -167,6 +166,39 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
     return data.replace(b"]]>", b"]]]]><![CDATA[>")
 
 
+def _hasMSOComments(html_string: Union[str, bytes]) -> bool:
```
adiroiban: even if this is a private method, it needs a docstring.

This is a good place to let other developers know about why this method exists.
Maybe add some more info about how MS Office comments work or a link to the MS Office specs.

src/twisted/web/_flatten.py
```diff
@@ -178,7 +210,10 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
+    if _hasMSOComments(data):
+        data = data.replace(b"--", b"- - ")
```
adiroiban:
```suggestion
    if _hasMSOComments(data):
        # Don't force escaping the closing tag angle bracket
        # for MS office, as we want them verbatim.
        data = data.replace(b"--", b"- - ")
```

Since I have not used this code, is hard to come with a good suggestion, but a comment would help.

adiroiban (changes requested):
Many thanks for the changes.

I think that the code is close to be merged.

But this feature is not easy to understand, for non MS Office hardcore users :)

This is why is important to have good code documentation, functionality documentation and automated tests.

Otherwise, we will merge the code now, but in the future there is a high risk or regressions.

In 1 year one developer might hit a different issue with this part of the code, will look at the code and will not understand it, and might change the code in such a way that MS Office logic will no longer work.

src/twisted/web/_flatten.py
felipemelendez: good point, if `data` is a string then it will be encoded to bytes in `escapedComment` using the utf-8 encoding before `_hasMSOComments` is called, that means that the data passed to `_hasMSOComments` is always of type bytes. I will fix it.

src/twisted/newsfragments/11804.bugfix
felipemelendez: got it, I will make the release notes brief, add docstrings where needed (will make a separate test with a proper docstring), and overall explain the work better, but not sure about the narrative documentation part, is that something the community also contributes to, or who handles that?

felipemelendez: [fe9c8ad](https://github.com/twisted/twisted/pull/11805/commits/fe9c8ad7af50c0085089a5b9d82c357dc58ad1ec)

src/twisted/web/_flatten.py
felipemelendez: [4810253](https://github.com/twisted/twisted/pull/11805/commits/481025321d7b95b19ed9213517e821a6a118128b)

src/twisted/web/_flatten.py
felipemelendez: thank you for pointing that out, that's an oversight from my part, see [4810253](https://github.com/twisted/twisted/pull/11805/commits/481025321d7b95b19ed9213517e821a6a118128b)

I fixed it to be:
```
if not _hasMSOComments(data):
    data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
```

src/twisted/web/test/test_flatten.py
felipemelendez: you're right, I was cutting corners, it's a bad habit that I will break.
I created a test for this particular change and added a proper docstring: [e909111](https://github.com/twisted/twisted/pull/11805/commits/e909111956155cf3dd3d4a09ea845e4647dc2548)

I also don't like single letter variables so I wrote descriptive variable names for my test function, but I left the other one as it was, didn't feel comfortable making that sort of change elsewhere

felipemelendez:
well @adiroiban, I have addressed the changes requested in your previous comments:
* I have written a docstring that explains the purpose of `_hasMSOComments`
* I added to the docstring of `escapedComment` to explain the presence of `_hasMSOComments`
* I fixed the issue with the handling of the call to `_hasMSOComments`, particularly in that because `data` is converted to bytes before `_hasMSOComments` the type of its argument is always bytes. 
* I created a test function that specifically validates the new functionality added by `_hasMSOComments`
* the new test function has a proper docstring that will help developers understand its intent
* with these changes we are now at 100%

Please let me know if there's anything else that needs to be addressed before we can merge this pull request, I would be happy to address it. But first let's enjoy the rest of our Sunday :sunglasses:

src/twisted/web/_flatten.py
```diff
@@ -178,7 +220,8 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
+    if not _hasMSOComments(data):
+        data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
```
exarkun: Why isn't the whole change:

```
     data = data.replace(b"-->", b"--&gt;")
```

felipemelendez: Fair question, this is done so that we may precisely follow the recommendations set by w3 regarding the Extensible Markup Language (XML). There is a lot to this, and that's why we find ourselves fine-tuning from time to time, so that we may adhere to these rules and thereby ensure language compatibility. 

* The part about the `>` can be found in the [Character Data and Markup section](https://www.w3.org/TR/REC-xml/#syntax). There we find this rule: 
"The right angle bracket (>) may be represented using the string `&gt;`, and must, for compatibility, be escaped using either `&gt;` or a character reference when it appears in the string `]]>` in content, when that string is not marking the end of a CDATA section."

* In regards to the `--`, the rules for that can be found in the [Comments section](https://www.w3.org/TR/REC-xml/#sec-comments). There we find this:
"For compatibility, the string `--` (double-hyphen) must not occur within comments"

So, if we were to just replace `-->` we would miss cases where we encounter standalone instances of `--` or `>`

exarkun: Okay.  Indeed, there is a complex and partially contradictory set of rules that apply - the rules of XML, the rules of SGML, and whatever rules MSIE is following.  Additionally, `twisted.web.template` doesn't even clearly state what its output is meant to be (in some places it talks about HTML, in some places it talks about XML, in other places it implies HTML5).

Still, the actual problem for MSIE is that ">" gets escaped inside comments.  The change I suggested would make MSIE happy but it would leave the output potentially in violation of the SGML comment rule that comments may not include "--".

This would make SGML and MSIE happy:

```
    data = data.replace(b"--", b"- - ")
```

That is, drop the ">" quoting from the current behavior but leave the other part alone.  This fixes the MSIE behavior by leaving ">" alone.  I don't think it breaks XML.  XML doesn't actually dictate that you quote this character inside a comment.  Instead, it says that the contents of comments don't even have to be parsed by an XML parser (https://www.w3.org/TR/REC-xml/#dt-comment) - and the associate example even includes unquoted "<", ">", and "&".  

I don't know what SGML makes of this behavior but I also wonder if it actually matters.  In practice `twisted.web.template` is not for emitting SGML except for the special case of HTML (which may or may not sometimes be an SGML subset).  The only reference to SGML in all of Twisted is, in fact, the test for this existing behavior.  It at least avoids the most serious issue SGML has - it gets rid of any "--" inside comments.

felipemelendez: > I don't know what SGML makes of this behavior but I also wonder if it actually matters.

Sorry, I am not trying to be facetious but I sincerely can't tell if you're suggesting a refactor. I want the best for Twisted, and I too don't know how removing the replacement of the `>` character will affect things elsewhere, all I know is that it has not been a problem until now. It seems to me that we'd rather avoid any regressions so if I have any say in this I'd rather keep the replacement of the character where applicable.

felipemelendez: However, to be fair, that is exactly what we did to monkey patch the issue we were having:
```
def escapedComment(data: bytes | str) -> bytes:
    """
    Same as _flatten.escapedComment but with .replace(b">", b"&gt;") removed
    Twisted issue #11804: https://github.com/twisted/twisted/issues/11804
    """
    if isinstance(data, str):
        data = data.encode("utf-8")
    data = data.replace(b"--", b"- - ")
    if data and data[-1:] == b"-":
        data += b" "
    return data
 ```

Comment (felipemelendez) [reply src/twisted/web/_flatten.py]: Point well taken, in fact, here is the snippet that we used to monkey patch around this issue:
```
def escapedComment(data: bytes | str) -> bytes:
    """
    Same as _flatten.escapedComment but with .replace(b">", b"&gt;") removed
    Twisted issue #11804: https://github.com/twisted/twisted/issues/11804
    """
    if isinstance(data, str):
        data = data.encode("utf-8")
    data = data.replace(b"--", b"- - ")
    if data and data[-1:] == b"-":
        data += b" "
    return data
 ````
 
But, to be fair, our use case is, in the grand scheme of things, relatively narrow. The problem here is that in XML, the > character has a special meaning as the closing bracket for elements. If the > character is used inside an XML element as text, it could be misinterpreted by an XML parser as the end of the element, causing the parser to generate an error. By using character references, we can avoid any potential issues with the parser misinterpreting the characters.

The HTML specification document summarizes these rules in a way that is a little more succinct: https://html.spec.whatwg.org/multipage/syntax.html#comments

There, you'll find that:
text must not start with the string `>`, nor start with the string `->`, nor contain the strings `<!--`, `-->`, or `--!>`, nor end with the string `<!-`.

I reference the HTML specification document because XML has a stricter syntax that requires well-formedness, while HTML has a more relaxed syntax, and still the HTML doc makes this point. While there are some differences between HTML and XML, both use similar rules for certain elements such as comments, and the use of the `>` character is one such example. The HTML specification document provides a concise and clear summary of the rules for comments in HTML, which is helpful in understanding the potential issues with using the `>` character in comments.

exarkun: > There, you'll find that: text must not start with the string >, nor start with the string ->, nor contain the strings , or --!>, nor end with the string <!-.

Perhaps this is an important rule to implement?  I notice that at least one these cases ("nor end with the string `<!-`") is handled by neither the existing code nor the proposed change.

More generally, the idea I'm wondering about is why we don't just implement the XML and HTML rules for comments instead of adding a couple hundred lines of code specific to MSIE.

Correct me if I'm wrong here: these MSIE comments _aren't_ illegal according to any of the relevant specs.

If that's true then the problem here is that the comment mangling code `Comment` uses to try to avoid the _illegal_ cases incorrectly mangles some _legal_ content.  The solution that makes sense to me is to stop mangling any legal content and only mangle illegal content.  The solution in this PR seems like it is still mangling a mix of legal and illegal cases (for example, `x>` is a legal comment but before and after this PR, `Comment` mangles it to `x&gt;`).

Thanks for reading this far.  The code in question is not great and I appreciate your efforts on it.  All of the rules that apply here are complex and I haven't worked directly with XML or HTML encoding in a long time so I don't have ready-made answers - hence the questions.  I do have a hunch that this code could be both simpler and closer correct and my reading of the current PR hasn't shown me that this hunch is wrong.  It _could_ be wrong and I don't mind if that's the outcome, but I'd like the reasons it is wrong to show up clearly somewhere - either in the PR or in these comments.

felipemelendez: I know there's a life lesson in here somewhere. 
I think that your hunch is well founded and I am hard pressed to think of reasons why you would be wrong, particularly because the snippet that I shared in my previous comment is working fine. See [my first commit](https://github.com/twisted/twisted/pull/11805/commits/ab006310f79316b55408f7b010cafdc0dcf0e33e) that was the path that I intended to take...
What is more, before I made the suggestion to my team on how to solve the MSIE issue I did a lot of research and found that `data = data.replace(b"--", b"- - ")` was enough, and it has been.
But then I decided to jump in here and try to help address this issue but perhaps got carried away with trying too hard or too much? In retrospect I think that I have been walking on egg shells making sure that we keep in mind every potential user and use case, and weary that I can't possibly test for all of them. 

By the way, in regards to the `<!-`, I can't think of another instance where that would appear within a comment besides this: `<!--<![endif]-->`, and that line has not been a problem. We have tested our email templates using Litmus (and send them in production to a large number of users), which tests email templates across a wide range of email clients, browsers, and devices. All together it supports over 90 email clients, including popular ones such as Gmail, Yahoo Mail, Outlook, Apple Mail, and more., so I can confidently say that there is not need to touch that. And what is more, that range of coverage is so wide that it serves as a good indicator that we no longer need to escape the `>` character.

felipemelendez: So I shifted my research criteria to history and here is what I found.
The XML 1.0 specification, which was released in February 1998, originally allowed the use of the ">" character within comments, as long as it was not part of the "-->" sequence. However, security concerns were later raised about the potential for malicious code to be injected into an XML document through comments. I believe this type of attack is known as cross-site scripting (XSS).
In the early days of HTML and XML, some web browsers and other software used to interpret them may not have handled the ">" character correctly, causing parsing errors or incorrect interpretation of the comment boundaries. This could potentially lead to security vulnerabilities or other issues if an attacker was able to inject malicious code or unexpected input into a comment.
However, modern web browsers and other software are generally designed to follow the HTML and XML specifications closely and fully support all standard features, including comments with the ">" character. So, while older systems or outdated software may have had issues with this, it is less likely to be a problem with current parsers and more recent technology.

While doing this research I learned that one example of a tool that did not support the ">" character within comments in the past was Internet Explorer 6 and earlier versions – that's the Microsoft's MSXML parser (specifically versions 3.0 and earlier). These older versions of Microsoft's web browser had a bug where they would not properly handle comments that contained the ">" character. This meant that any XML document that contained this character within a comment would not be rendered correctly in Internet Explorer. However, this bug was fixed in later versions of Internet Explorer, and most modern web browsers and XML tools should have no problem handling the ">" character within comments in XML documents. Similarly, and I add this to show that my research has been exhaustive, some older versions of the Apache Xerces C++ parser had a bug where they would not properly handle comments that contained the ">" character, which could cause parsing errors. 

Some examples of software that had issues with the > character within comments include: 
- Internet Explorer 6 and older versions of the browser 
- Netscape Navigator 4 and older versions of the browser 
- Some older versions of the libxml2 library, particularly versions prior to 2.7.8 

And these are all old versions of these parsers that have since been updated and support the use of > within comments

felipemelendez: > * The part about the `>` can be found in the [Character Data and Markup section](https://www.w3.org/TR/REC-xml/#syntax).

On a final note, I now realize that I made a mistake when I referenced this. I figure that I was trying to align with the statements made within `test_commentEscaping`. Where in fact:
`
Char	   ::=   	#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
Comment	   ::=   	''
`
The definition for Char includes all characters in the range from #x1 to #xD7FF, the range from #xE000 to #xFFFD, and the range from #x10000 to #x10FFFF.
The range [#x1-#xD7FF] includes all Unicode code points from #x0001 to #xD7FF, which means that #x3E (which is a hexadecimal representation of the greater-than character '>') falls within this range. 

Sadly, I made this statement when I drafted the issue

felipemelendez:
Esteemed @exarkun, thank you very much for providing your valuable input and insight on this PR. I have a lot of respect for your work and was hesitant to make changes that could compromise it. Instead of following my initial approach, I decided to work around it. I have now come full circle and made the appropriate changes. I also conducted additional research to support the reasoning behind this change, and I hope it will help inform your final decision. Whatever verdict you reach, I am fully supportive. Best regards.

src/twisted/web/test/test_flatten.py
```diff
@@ -269,14 +269,17 @@ def test_commentEscaping(self) -> Deferred[List[bytes]]:
         The data in a L{Comment} is escaped and mangled in the flattened output
         so that the result is a legal SGML and XML comment.
 
-        SGML comment syntax is complicated and hard to use. This rule is more
-        restrictive, and more compatible:
+        BNF (Backus-Naur Form) production rule for comments:
+        Comment	 ::=  '<!--' ((Char - '-') | ('-' (Char - '-')))* '-->'
```
glyph: Just a suggestion, I think it would be helpful to present the production rule as code, rather than flowed in paragraph text:

```suggestion
        BNF (Backus-Naur Form) production rule for comments::

            Comment	 ::=  '<!--' ((Char - '-') | ('-' (Char - '-')))* '-->'
```

src/twisted/web/_flatten.py
```diff
@@ -178,7 +177,7 @@ def escapedComment(data: Union[bytes, str]) -> bytes:
     """
     if isinstance(data, str):
         data = data.encode("utf-8")
-    data = data.replace(b"--", b"- - ").replace(b">", b"&gt;")
+    data = data.replace(b"--", b"- - ")
```
glyph: This is still mangling, not quoting (it looks like quoting is not possible here) but it is a less legible form of quoting.

What do you think about this, instead?

```suggestion
    data = data.replace(b"-->", b"--&gt;")
```

I parsed a `--` in a comment with `html5lib` 1.1 and despite the spec's prohibition on `--`, this document:

```html
<!DOCTYPE html>
<html>
 <head>
  <title>hi</title>
 </head>
 <body>
  <!-- -- -->
 </body>
</html>
```

responds to this program:
```python3
from html5lib import parse
from xml.etree.ElementTree import tostring

with open("comment.html") as f:
    d = parse(f, namespaceHTMLElements=False)
    c = list(d.iter())[-1]
    print(c.text)
```
by printing out `--`, and can happily re-serialize it without any warnings I can detect.

Given that we need to mangle it, it would seem to me that we should mangle as minimally as possible, which appears to be what browsers and reference specs do. For example, that HTML file in a browser looks like this as well:

```
> document.body.childNodes[1].textContent
< " -- "
```

Most of all, adding in a space to mangle like this seems really arbitrary.

glyph:
Hopefully I can do a little bit of gordian knot-cutting here by suggesting what the "right" behavior for twisted.web.template is, in general:

First and foremost, any parsed document should round-trip as closely as possible. We do not need to commit to *perfect* byte-for-byte fidelity; there are some cases (i.e. dealing with certain subtle encoding errors) where we might do something different, but we should perturb as little as possible.  The current behavior here violates that constraint and that's the most important thing to fix.  Our current behavior of quoting `>` is wrong, but quoting `--` would be equally wrong, and is unnecessary.

Then there's the question of what to do with potentially invalid *synthetic* documents; i.e. what should we do if we didn't *parse* something but a user constructed it. We don't want to suddenly start raising exceptions on people constructing nodes with strings that would previously have serialized something correctly, so constructing a comment with a `--` in it shouldn't raise an error.  The question about what to do here is what specs do we look to in order to figure out how any quoting behavior should be exposed to applications.  Is it an HTML processor, an SGML processor, an XML processor or what? If these specs disagree, what's the tiebreaker?

The history of SGML, HTML (3, 4), and XML is *interesting*, but in practice twisted.web.template is used to serialize HTML documents, usually in an HTML5 context.  At the time of its creation, compatibility with XHTML was really pressing, and thus we should try to maintain compatibility with XML (and XHTML in particular) as much as possible, but HTML5 is so dominant, and what we are serializing *is* HTML here, that when we need a tie broken on what behavior to have, the hierarchy of applicable specs is:

1. if browsers & the reference implementation (html5lib) agree on something, do that (as long as it doesn't result in output that results in quoting bugs that will corrupt the document or potentially create escaping problems with untrusted data (more on that in a moment))
2. if the implementation behavior doesn't agree, try to follow the HTML5 spec
3. if possible, follow any relevant XML specs as well, but HTML5 wins if there's a hard conflict where something just won't work

So, applying those principles in this case:

The current implementation will mangle valid documents it does not need to. That is a bug. If you read `>` in a comment, which is allowed per most of these specs, then you'll get back `&gt;` if you reserialize it, which is unnecessary. We should fix that.

Mangling `--` instead will mangle *invalid* documents according to the spec, but the spec is more of a tiebreaker in the case that the implementations don't agree, so what do the implementations do? As I explained above, they actually allow `--` just fine.

But then there's that caveat about quoting / injection attacks; i.e. if a document can be made to fail to round-trip through other implementations, we should do something that is at least safe against that.  And here, we can see that html5lib, at least, is not safe, when we try to create a synthetic comment that contains `-->` by doing this to the aforementioned document:

```python3
with open("comment.html") as f:
    d = parse(f, namespaceHTMLElements=False)
    c = list(d.iter())[-1]
    c.text = "-->"
    print(c.text)
    print(tostring(d))
```

and you get this:

```
b'<html><head>\n  <title>hi</title>\n </head>\n <body>\n  <!---->-->\n \n\n</body></html>'
```

which is just wrong, because that will produce a document that will be parsed completely differently than the document model that serialized to it.  Here's where we still need to do some kind of mangling, because bailing entirely is not acceptable (that leads to an attack where serializing some untrusted data in a comment can blow up your rendering) and quoting is not possible in this context.  This will produce something which is still XML, still HTML, and unfortunately is slightly mangled but only in the case that a document contains text that *cannot be parsed*, thus cannot break the round-trip rule.

src/twisted/web/_flatten.py
felipemelendez: Awesome @glyph, thank you for your explanation, it was incredibly helpful and improved my understanding. I also appreciate the invocation of Greek mythology to help break the impasse. Alexander would be proud.   

Also, although it's a small matter, I followed your lead and conducted similar tests as you and noticed that ending a comment with a dash does not appear to have any effect. Following the ideal to perturb as little as possible, I would be tempted to remove the following code block, but I don’t want to be petty and make any unwarranted changes, so I won't touch it unless directed to do so.
``` python
if data and data[-1:] == b"-": 
    data += b" " 
``` 

I'll jump on this as soon as I get a chance. Once again I appreciate your input on this.

glyph: Oh, that's a great catch. Thank you for pointing it out :-).

glyph: (And yes, I agree that adding whitespace to break the connection of the ending `-` with the closing `-->` is probably the least intrusive thing I can think of.)

felipemelendez:
Gentlemen, I am thrilled to have the opportunity to contribute to this fantastic project. I realize that my induction was chaotic, but in retrospect the best adventures are those that come with a little drama.  

@adiroiban , thank you for making sure that my work was done right. 

@exarkun, you jumped into this PR with the right idea and I veered you off course, my apologies. I owe you a cold one for that.  

@glyph, I follow your blueprint at work daily, so I was thrilled to see you peek into this PR. I unfortunately missed the chance to work with you at Pilot, but this is a pretty cool consolation prize.  

If there is anything else I need to do to make sure this contribution is up to par please let me know. Enjoy the weekend everyone :sunglasses:

src/twisted/web/_flatten.py
```diff
@@ -169,7 +168,10 @@ def escapedCDATA(data: Union[bytes, str]) -> bytes:
 
 def escapedComment(data: Union[bytes, str]) -> bytes:
     """
-    Escape a comment for inclusion in a document.
+    Within comments the sequence C{-->} can be mistaken as the end of the comment.
+    To ensure consistent parsing and valid output the sequence is replaced with C{--&gt;}.
+    Furthermore, whitespace is added when a comment ends in a dash. This is done to break
```
adiroiban, line 173: Is space still added?

src/twisted/newsfragments/11804.bugfix
```diff
@@ -0,0 +1 @@
+To prevent parsing errors and ensure validity when serializing HTML comments, twisted.web.template.flattenString has been updated to escape the --> sequence within comments
```
adiroiban:
```suggestion
To prevent parsing errors and ensure validity when serializing HTML comments, twisted.web.template.flattenString has been updated to escape the --> sequence within comments.
```

Just a full top at the end, to make sure that this was not left unfinished.

adiroiban (commented):
I think the changes looks much better.

I still have a question regarding docstring, before this can be merged.

Thanks again for all your work on this.

src/twisted/web/_flatten.py
felipemelendez: yeah:
```python
if data and data[-1:] == b"-":
        data += b" "
```        
I also questioned this to be certain that it should remain and got the green light.
In this regard Glyph said:  "Yes, I agree that adding whitespace to break the connection of the ending - with the closing --> is probably the least intrusive thing I can think of."

src/twisted/newsfragments/11804.bugfix
felipemelendez: [b754b4a](https://github.com/twisted/twisted/pull/11805/commits/b754b4a84dac22373d1e60acce4511610f752dc9)

felipemelendez:
Gentlemen, I think the fact that my last commit was to add a period is pretty symbolic, if that is the final touch needed to complete this PR that would be metaphorically awesome. But if there is anything else that needs to be done to make sure that this is up to par please let me know. I am thrilled to have the opportunity to contribute to this fantastic project. I realize that my induction has been a little chaotic, but in retrospect the best adventures are those that come with a little drama.

@adiroiban, thank you for making sure that my work was complete and done right.

@exarkun, you jumped into this PR with the right idea and I veered you off course, my apologies. I owe you a cold one for that.

@glyph, I follow your blueprint at work daily, so I was thrilled to see you peek into this PR. I unfortunately missed the chance to work with you at Pilot, but this is a pretty cool consolation prize.

adiroiban (approved):
All good then. Sorry for the trouble.

PR #11809, opened 2023-02-09
[#11806] Update isort deps in pre-commit config to fix CI failures.

### LINKED ISSUE #11806: isort is broken
a transitive dependency broke isort (and thus pre-commit and CI)

5.12.0 fixes a bug that was breaking CI everywhere.

https://github.com/PyCQA/isort/releases/tag/5.12.0

felipemelendez:
## Scope and purpose

Fixes #11806

Specified isort to be 5.12 in pre commit config

5.12.0 fixes a bug that was breaking CI everywhere.

https://github.com/PyCQA/isort/releases/tag/5.12.0


felipemelendez:
please review

src/twisted/newsfragments/11806.bugfix
```diff
@@ -0,0 +1,2 @@
+.pre-commit-config.yaml now references isort revision 5.12.0
```
adiroiban: This should be a `.misc` file.

The .bugfix is only for fixes to the twisted changes that affect end users.

this is a CI / dev only change.

No need to spam Twisted users with internal changes.
We have the git history for that :)

adiroiban (approved):
Thanks. Awesome work.

Even if this is a small change, this change is critical.

I will wailt for CI to be green and then I will merge it.

adiroiban:
Many thanks @felipemelendez for this PR.

I have merged this ASAP to get trunk green again.

Having a more robust dependency definition would be nice, but I think this can be done in a separate ticket/PR.

PR #11810, opened 2023-02-14
#11626 fix `twist conch --auth=sshkey`

### LINKED ISSUE #11626: `twist conch --auth=sshkey`  doesn't work due to passing bytes to getpwnampwd.
Just saw this on 22.8.0.rc1, with the equivalent of `twist conch --auth=sshkey` .

```
2022-08-29T07:31:45+0000 [twisted.conch.ssh.userauth.SSHUserAuthServer#critical] Error checking auth for user b'user'
	  File "/code/env/lib/pypy3.8/site-packages/twisted/conch/ssh/userauth.py", line 285, in auth_publickey
	    return self.portal.login(c, None, interfaces.IConchUser)
	  File "/code/env/lib/pypy3.8/site-packages/twisted/cred/portal.py", line 120, in login
	    ).addCallback(self.realm.requestAvatar, mind, *interfaces)
	  File "/code/env/lib/pypy3.8/site-packages/twisted/internet/defer.py", line 531, in addCallback
	    return self.addCallbacks(callback, callbackArgs=args, callbackKeywords=kwargs)
	  File "/code/env/lib/pypy3.8/site-packages/twisted/internet/defer.py", line 511, in addCallbacks
	    self._runCallbacks()
	  File "/code/env/lib/pypy3.8/site-packages/twisted/internet/defer.py", line 892, in _runCallbacks
	    current.result, *args, **kwargs
	  File "/code/env/lib/pypy3.8/site-packages/twisted/conch/ssh/userauth.py", line 208, in _ebMaybeBadAuth
	    reason.trap(error.NotEnoughAuthentication)
	  File "/code/env/lib/pypy3.8/site-packages/twisted/python/failure.py", line 480, in trap
	    self.raiseException()
	  File "/code/env/lib/pypy3.8/site-packages/twisted/python/failure.py", line 504, in raiseException
	  File "/code/env/lib/pypy3.8/site-packages/twisted/internet/defer.py", line 892, in _runCallbacks
	    current.result, *args, **kwargs
	  File "/code/env/lib/pypy3.8/site-packages/twisted/conch/unix.py", line 49, in requestAvatar
	    user = UnixConchUser(username)
	  File "/code/env/lib/pypy3.8/site-packages/twisted/conch/unix.py", line 57, in __init__
	    self.pwdData = pwd.getpwnam(self.username)
	builtins.TypeError: expected str, got bytes object
```

_Originally posted by @glyph in https://github.com/twisted/twisted/issues/10286#issuecomment-1229896032_

Refer to the patch for #10286 for `UserDB` typing and mocks that will be helpful.

glyph:
## Scope and purpose

Fixes #11626 

## Contributor Checklist:

### DIFF
--- a/src/twisted/conch/test/test_tap.py
+++ b/src/twisted/conch/test/test_tap.py
@@ -5,9 +5,11 @@
 Tests for L{twisted.conch.tap}.
 """
 
+from typing import Any
+
 from twisted.application.internet import StreamServerEndpointService
 from twisted.cred import error
-from twisted.cred.checkers import ICredentialsChecker
+from twisted.cred.checkers import FilePasswordDB, ICredentialsChecker
 from twisted.cred.credentials import ISSHPrivateKey, IUsernamePassword, UsernamePassword
 from twisted.internet.defer import Deferred
 from twisted.python.reflect import requireModule
@@ -39,7 +41,7 @@ class MakeServiceTests(TestCase):
 
     usernamePassword = (b"iamuser", b"thisispassword")
 
-    def setUp(self):
+    def setUp(self) -> None:
         """
         Create a file with two users.
         """
@@ -48,7 +50,7 @@ class MakeServiceTests(TestCase):
             f.write(b":".join(self.usernamePassword))
         self.options = tap.Options()
 
-    def test_basic(self):
+    def test_basic(self) -> None:
         """
         L{tap.makeService} returns a L{StreamServerEndpointService} instance
         running on TCP port 22, and the linked protocol factory is an instance
@@ -60,7 +62,7 @@ class MakeServiceTests(TestCase):
         self.assertEqual(service.endpoint._port, 22)
         self.assertIsInstance(service.factory, OpenSSHFactory)
 
-    def test_defaultAuths(self):
+    def test_defaultAuths(self) -> None:
         """
         Make sure that if the C{--auth} command-line option is not passed,
         the default checkers are (for backwards compatibility): SSH and UNIX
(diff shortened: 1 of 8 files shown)


glyph:
please review

glyph:
It would be nice to address the coverage gap here, but the types are sufficient in this case (I've tested this interactively, and ssh clients can indeed authenticate to conch again)

adiroiban:
@glyph is it ok to merge a change without a test?
In the past we were rejecting PR is they don't have tests for the changes done in that particular PR.

Regards

glyph:
@adiroiban I suppose it's up to you, as the reviewer, if you want to say "no, we must have test coverage" ;-).

As a matter of policy we require tests. As a matter of principle I'd say "I just added type annotations, no behavior was changed" is an acceptable reason to that requirement. But of course, this change *doesn't* adhere to that principle; I do encoding where there wasn't any encoding before, to address the bug.

My reasoning for this fairly unprincipled exception is:

- a *real* test of this behavior would be exceedingly expensive, and require like… an entire docker container with a `twist` running as root since `pwd` is otherwise inaccessible, not to mention the ability to create an arbitrary home directory, so the best we could do would be a fairly unconvincing fake, which strikes me as not worth it
- this code was previously changed (and regressed) somehow, since pubkey auth used to work on python3 in this configuration; I know because I run a server whose SSH access broke
- the code just obviously doesn't work at all right now and this is a quick fix

but, you could make a similar case for any other too-long-regressed bug, so 🤷 I'll have to leave it up to a reviewer

glyph:
> * a _real_ test of this behavior would be exceedingly expensive, and require like… an entire docker container with a `twist` running as root since `pwd` is otherwise inaccessible, not to mention the ability to create an arbitrary home directory, so the best we could do would be a fairly unconvincing fake, which strikes me as not worth it

well I guess the fake version is okay if everything is typechecked everywhere, so I added a bunch more types

glyph:
OK, tests cover it. @adiroiban I think this might be more of a rubber stamp now :)

adiroiban (approved):
Many thanks for this great update.
Should be ready to merge.

glyph:
Thanks for the prompting on the test!

PR #11823, opened 2023-03-13
#11822 filepath annotations

### LINKED ISSUE #11822: FilePath and friends should have type annotations
Particularly given the text/bytes-mode distinction of how they can work, we should really include type hints with them.

glyph:
## Scope and purpose

Fixes #11822 

## Contributor Checklist:


glyph:
review please

glyph:
haven't run the tests yet myself, but the type-check should be clean

src/twisted/python/filepath.py
```diff
@@ -38,7 +39,24 @@
     S_IXOTH,
     S_IXUSR,
 )
-from typing import IO, Union, cast
+from typing import (
+    IO,
+    TYPE_CHECKING,
+    AnyStr,
+    Callable,
+    Dict,
```
exarkun, line 48: Why `__future__.annotations` _and_ `Dict`, `List`, etc?

glyph: `__future__.annotations` lets you forward-declare stuff, it doesn't affect builtin types.  Which means that APIs like [`inspect.get_annotations(..., eval_str=True)`](https://docs.python.org/3.11/library/inspect.html#inspect.get_annotations) (and its more ad-hoc predecessors) will work just fine in the presence of `__future__.annotations` on older Python versions, but will *not* work if you start using `list[...]`, `dict[...]`, et. al.

glyph: (When we drop <3.9, [we can ask pyupgrade to make this transformation for us mechanically](https://github.com/asottile/pyupgrade/tree/67e006e96f7fab6533ab5f831872e72c109a7b28#pep-585-typing-rewrites), and I look forward to doing so.)

glyph:
So, those test failures were embarrassing. Let's try that again…

glyph:
Looks like I've got a few minor coverage gaps which should be addressed but otherwise I think we're good now

glyph:
OK. Coverage gaps all addressed.

adiroiban (approved):
I don't have much experience with Python typing.
I did a very quick look and it looks ok
I hope @graingert can also take another look to check all is ok.

Since this is waiting in the queue since March, as long as the tests are green, I think that this can be  merged.

Thanks Glyph for the update.

glyph:
thanks @adiroiban ! I'm happy to take any post-hoc feedback from @graingert (or anyone else for that matter) but as you note, this *has* been waiting since March, so let's get it out of the queue and move things along, there are plenty of other PRs that need review attention…

src/twisted/python/zippath.py
```diff
@@ -206,11 +270,13 @@ class ZipArchive(ZipPath):
     converting if required.
     """
 
+    _zipfileFilename: AnyStr
+
     @property
     def archive(self):
         return self
```
graingert, line 277: this is the only method not annotated, it should probably be  `-> Self: # type: ignore[override]`

src/twisted/python/filepath.py
```diff
@@ -273,17 +283,87 @@ def _secureEnoughString(path):
     return _coerceToFilesystemEncoding(path, secureishString)
```
graingert, line 283: this is missing a type annotation - should be `def _secureEnoughString(path: AnyStr) -> AnyStr: ...`

PR #11829, opened 2023-03-21
11828 synchronous test pbfailure

### LINKED ISSUE #11828: test_pbfailure uses real TCP sockets and fails intermittently
We should be using an in-memory transport.

(Probably we should get rid of PB entirely, but, one thing at a time)

glyph:
## Scope and purpose

Fixes #11828


glyph:
please review

alex (approved):

Issue #3008, opened 2008-01-23
twisted.conch.ssh.keys.Key._fromString_PRIVATE_OPENSSH fails on RSA keys


twisted-trac:
|[<img alt="exarkun's avatar" src="https://avatars.githubusercontent.com/u/254565?s=50" width="50" height="50">](https://github.com/exarkun)| @exarkun reported|
|-|-|
|Trac ID|trac#3008|
|Type|enhancement|
|Created|2008-01-23 15:16:57Z|
|Branch|https://github.com/twisted/twisted/tree/broken-private-key-3008|

```
          File "/home/exarkun/Projects/Twisted/trunk/twisted/conch/ssh/keys.py", line 186, in _fromString_PRIVATE_OPENSSH
            n, e, d, p, q = decodedKey[1:6]
        exceptions.ValueError: need more than 1 value to unpack
```

```decodedKey``` appears to be a list of one element containing the desired values.

This is a regression from the previous release.

Attachments:

* [temp](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/2ec/2ec22303caffe3c4a2a675a85aa5bde87da43c8a/d969831eb8a99cff8c02e681f43289e5d3d69664) (1675 bytes) - added by exarkun on 2008-02-20 17:50:53Z - rsa private key

<details><summary>Searchable metadata</summary>

```
trac-id__3008 3008
type__enhancement enhancement
reporter__exarkun exarkun
priority__highest highest
milestone__twisted_8_0 twisted-8.0
branch__branches_broken_private_key_3008 branches/broken-private-key-3008
branch_author__ 
status__closed closed
resolution__fixed fixed
component__conch conch
keywords__ 
time__1201101417000000 1201101417000000
changetime__1230850780000000 1230850780000000
version__None None
owner__ 
cc__exarkun
```
</details>

twisted-trac:
|<img alt="z3p's avatar" src="https://avatars.githubusercontent.com/u/0?s=50" width="50" height="50"><a name="note_1"></a>|z3p commented|
|-|-|

(In [22635]) Branching to 'broken-private-key-3008'

twisted-trac:
|<img alt="z3p's avatar" src="https://avatars.githubusercontent.com/u/0?s=50" width="50" height="50"><a name="note_2"></a>|z3p set status to `assigned`|
|-|-|

[...]

PR #11843, opened 2023-04-21
remove pyasn1 from conch and everywhere else

reaperhulk:
This PR removes pyasn1 from the dependencies of conch. However, there is one consequence: broken OpenSSH RSA keys from https://github.com/twisted/twisted/issues/3008 are no longer supported.

### DIFF
--- a/src/twisted/conch/newsfragments/11843.removal
+++ b/src/twisted/conch/newsfragments/11843.removal
@@ -1,1 +1,5 @@
 PyAsn1 has been removed as a conch dependency.
+
+twisted.conch.ssh.keys.Key no longer supports loading "alternate" OpenSSH private keys.
+These are some private keys that at some point were handled by OpenSSH but for which no specification exists.
+For more info about these OpenSSH keys see https://github.com/twisted/twisted/issues/3008.
--- a/src/twisted/conch/newsfragments/11843.removal
+++ b/src/twisted/conch/newsfragments/11843.removal
@@ -1,0 +1,1 @@
+PyAsn1 has been removed as a conch dependency.
--- a/src/twisted/conch/ssh/keys.py
+++ b/src/twisted/conch/ssh/keys.py
@@ -8,7 +8,6 @@ Handling of RSA, DSA, ECDSA, and Ed25519 keys.
 
 
 import binascii
-import itertools
 import struct
 import unicodedata
 import warnings
@@ -27,12 +26,6 @@ from cryptography.hazmat.primitives.serialization import (
     load_pem_private_key,
     load_ssh_public_key,
 )
-from pyasn1.codec.ber import (  # type: ignore[import]
-    decoder as berDecoder,
-    encoder as berEncoder,
-)
-from pyasn1.error import PyAsn1Error  # type: ignore[import]
-from pyasn1.type import univ  # type: ignore[import]
 
 from twisted.conch.ssh import common, sexpy
 from twisted.conch.ssh.common import int_to_bytes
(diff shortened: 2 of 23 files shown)


adiroiban:
Many thanks, Paul for the PR.

I think this is a good start.

I don't think this can be merged as it is.

There is a bit of red tape, in that I don't see an associated ticket for this PR, but I think that is not a big problem.

-------------

The biggest problem is the backward compatibility support.

A deprecation warning should be first raised for anyone still using the RSA keys that are no longer supported.

This procedure is documented here https://docs.twisted.org/en/stable/development/compatibility-policy.html#procedure-for-exceptions-to-this-policy

Since these RSA keys are most probably a bug I think that this can also qualify for Gross Violation of Specifications

https://docs.twisted.org/en/stable/development/compatibility-policy.html#bug-fixes-and-gross-violation-of-specifications


-------------

> TODO: why not just support serialization here

I think that this to-do needs a follow-up GitHub issue, not just a comment.

----------

I am +1 for using the `cryptography` API for the serialization and deserialization of SSH keys.

I think that twisted.conch.ssh.keys should only handle load-blod/sign/verify operations.
All the OpenSSH/PuTTY/Tectia key format handling should be done outside of Twisted (it already does sign and verify).
In this way, the code can be shared with `paramiko` and other Python SSH implementations.

--------------

I think that we can also go with `Gross Violation of Specifications` but we should also update the release notes to mention that some RSA keys will no longer work.

I can look into updating the existing twisted.conch.ssh.Keys code to raise a deprecation warning when a "broken" SSH key is loaded. I can trigger a release with this change.

And after that, we can just merge this.

Thanks again for the PR.

I am happy to see code removed from Twisted and moved into libraries that can be shared by other projects.

adiroiban:
At the same time, I see that the current Twisted `trunk` fails to load this OpenSSH key.

To implement the deprecation warning, I have moved the key into a separate test, and I can see that the tests fails if  executed on top of 525da4a0be  (this should be the current trunk)

```diff
diff --git a/src/twisted/conch/test/test_keys.py b/src/twisted/conch/test/test_keys.py
index 22bd1ac691..d1c4c5a4ca 100644
--- a/src/twisted/conch/test/test_keys.py
+++ b/src/twisted/conch/test/test_keys.py
@@ -281,10 +281,7 @@ class KeyTests(unittest.TestCase):
             ),
             keys.Key.fromString(keydata.privateRSA_openssh),
         )
-        self.assertEqual(
-            keys.Key.fromString(keydata.privateRSA_openssh_alternate),
-            keys.Key.fromString(keydata.privateRSA_openssh),
-        )
+
         self._testPublicPrivateFromString(
             keydata.publicDSA_openssh,
             keydata.privateDSA_openssh,
@@ -300,6 +297,18 @@ class KeyTests(unittest.TestCase):
                 keydata.Ed25519Data,
             )
 
+    def test_fromOpenSSH_alternate_deprecate(self):
+        """
+        Some versions of OpenSSH generate these (slightly different keys)
+        It is not any standard key format and was probably a bug in OpenSSH at
+        some point.
+
+        When loading such a key a deprecation warning is raised.
+        """
+        referenceKey = keys.Key.fromString(keydata.privateRSA_openssh)
+        alternateKey = keys.Key.fromString(keydata.privateRSA_openssh_alternate)
+        self.assertEqual(referenceKey, alternateKey)
+
```

```
[ERROR]
  File "/home/adi/chevah/twisted/src/twisted/conch/test/test_keys.py", line 307, in test_fromOpenSSH_alternate
    alternateKey = keys.Key.fromString(keydata.privateRSA_openssh_alternate)
  File "/home/adi/chevah/twisted/src/twisted/conch/ssh/keys.py", line 236, in fromString
    return method(data, passphrase)
  File "/home/adi/chevah/twisted/src/twisted/conch/ssh/keys.py", line 633, in _fromString_PRIVATE_OPENSSH
    return cls._fromPrivateOpenSSH_PEM(data, passphrase)
  File "/home/adi/chevah/twisted/src/twisted/conch/ssh/keys.py", line 566, in _fromPrivateOpenSSH_PEM
    decodedKey = berDecoder.decode(keyData)[0]
  File "/home/adi/chevah/twisted/venv/lib/python3.10/site-packages/pyasn1/codec/ber/decoder.py", line 2003, in __call__
    for asn1Object in streamingDecoder:
  File "/home/adi/chevah/twisted/venv/lib/python3.10/site-packages/pyasn1/codec/ber/decoder.py", line 1918, in __iter__
    for asn1Object in self._singleItemDecoder(
  File "/home/adi/chevah/twisted/venv/lib/python3.10/site-packages/pyasn1/codec/ber/decoder.py", line 1778, in __call__
    for value in concreteDecoder.valueDecoder(
  File "/home/adi/chevah/twisted/venv/lib/python3.10/site-packages/pyasn1/codec/ber/decoder.py", line 660, in valueDecoder
    for asn1Object in self._decodeComponentsSchemaless(
  File "/home/adi/chevah/twisted/venv/lib/python3.10/site-packages/pyasn1/codec/ber/decoder.py", line 604, in _decodeComponentsSchemaless
    componentTypes.add(component.tagSet)
builtins.AttributeError: 'NoneType' object has no attribute 'tagSet'

twisted.conch.test.test_keys.KeyTests.test_fromOpenSSH_alternate
Ran 170 tests in 8.888s
```

adiroiban:
So... the test is green with `pyasn1==0.4.8` but fails with `pyasn1==0.5.0`

I can see this was raised by someone upstream https://github.com/pyasn1/pyasn1/issues/29

alex:
FWIW, the twisted tests breaking on this weeks pyasn1 release was the
proximate motivation to look to remove pyasn1 entirely from twisted.

On Sun, Apr 23, 2023 at 1:09 PM Adi Roiban ***@***.***> wrote:
>
> So... the test is green with pyasn1==0.4.8 but fails with pyasn1==0.5.0
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>



-- 
All that is necessary for evil to succeed is for good people to do nothing.

adiroiban:
Thanks, Alex for the info. 

I guess that a quick fix is to update Twisted deps to depend on an older pyans1.

I am trying this solution here https://github.com/twisted/twisted/pull/11845

----

I think that for Twisted Conch, it's ok to remove pyasn1 and use only `cryptography`

At the same time, I am happy to see that the devs from pyasn1 are considering the support for streaming.

adiroiban:
The PR for raising a deprecation warning is here https://github.com/twisted/twisted/pull/11847

I think that it's best to make a release of Twisted that will raise a warning if these OpenSSH keys are used.

And then we can remove support for them in the following Twisted release.

glyph:
> A deprecation warning should be first raised for anyone still using the RSA keys that are no longer supported.

I disagree. Dealing with old, broken, rarely-used data formats whose compatibility is not explicitly documented is a not covered by any of the requirements [here](https://docs.twisted.org/en/stable/development/compatibility-policy.html#changes-covered-by-the-compatibility-policy).  It hews pretty closely to "gross violation of specifications", particularly when it comes to cryptographic formats, which we have to be considerably more aggressive about deprecating for security reasons.

glyph:
I *absolutely* do not want to pin to older pyasn1 because that is preventing security updates to a security-critical library.

src/twisted/conch/test/test_keys.py
```diff
@@ -1191,7 +1183,7 @@ def test_toOpenSSHDSA(self):
         L{keys.Key.toString} serializes a DSA key in OpenSSH format.
         """
         key = keys.Key.fromString(keydata.privateDSA_lsh)
-        self.assertEqual(key.toString("openssh"), keydata.privateDSA_openssh)
+        self.assertEqual(key.toString("openssh").strip(), keydata.privateDSA_openssh)
```
adiroiban, line 1186: So... with these changes, toString will generate some extra newlines.

Maybe we call strip() inside `toString()` so that the result is exactly the same in any Twisted version

src/twisted/conch/ssh/keys.py
```diff
@@ -1563,75 +1487,24 @@ def _toPrivateOpenSSH_PEM(self, passphrase=None):
...
+            encryptor = serialization.BestAvailableEncryption(passphrase)
+        if self.type() != "Ed25519":
             return self._keyObject.private_bytes(
                 serialization.Encoding.PEM,
                 serialization.PrivateFormat.TraditionalOpenSSL,
                 encryptor,
             )
-        elif self.type() == "Ed25519":
+        else:
+            # TODO: why not just support serialization here
```
adiroiban, line 1502: I don't understand this TODO here.
It this a TODO for this  branch or we should have a follow up ticket for it?

src/twisted/conch/ssh/keys.py
```diff
@@ -1563,75 +1487,24 @@ def _toPrivateOpenSSH_PEM(self, passphrase=None):
...
+        else:
+            encryptor = serialization.BestAvailableEncryption(passphrase)
+        if self.type() != "Ed25519":
             return self._keyObject.private_bytes(
                 serialization.Encoding.PEM,
                 serialization.PrivateFormat.TraditionalOpenSSL,
                 encryptor,
             )
-        elif self.type() == "Ed25519":
+        else:
```
adiroiban, line 1501: Very minor comment.
is this else needed?

Can we just remove it and reduce the indentation?

src/twisted/conch/newsfragments/11843.removal
```diff
@@ -0,0 +1 @@
+PyAsn1 has been removed as a conch dependency.
```
adiroiban, line 1: I think it is worth adding a note about the OpenSSH keys that used to work, but are now no longer supported

```suggestion
PyAsn1 has been removed as a conch dependency.

twisted.conch.ssh.keys.Key no longer supports loading "alternate" OpenSSH private keys.
These are some private keys that at some point were handled by OpenSSH but for which no specification exists.
For more info about these OpenSSH keys see https://github.com/twisted/twisted/issues/3008.
```

adiroiban (dismissed):
Thanks for the PR.

I left a few comments inline.

I think they are not controversial/hard to fix.

Let me know if you need help with this branch.

Happy to help get this merge.

glyph:
Thanks for the re-review @adiroiban !

src/twisted/conch/ssh/keys.py
glyph: I find the "else" easier to read so I think we should leave it :)

src/twisted/conch/test/test_keys.py
glyph: The newlines are semantically insignificant and small variances like this between upgrades seem fine to me.

src/twisted/conch/ssh/keys.py
glyph: @alex @reaperhulk I'm going to merge this as-is since I don't want to block "CI working at all" on this minor TODO detail, but I'd really love an issue that followed up with this and removed the TODO to put it in a more actionable place (i.e. an issue that explains the serialization issue)

glyph (approved):
LGTM, I think this works as-is and the remainder can be addressed in a follow-up.

PR #11845, opened 2023-04-23
#11844 Pin pyasn1 dependency for Twisted Conch to version below 0.5

### LINKED ISSUE #11844: twisted.conch.test.test_keys are failing with latest pyasn1 0.5.0
There was an updated pyasn1 that breaks our tests in trunk 

There is a PR to remove pyasn1 completely https://github.com/twisted/twisted/pull/11843

The only issue is that it removes support for some buggy OpenSSH keys.

I guess that a quick fix is to update Twisted deps to depend on pyasn1 0.4.8 only

This should be easy to merge and later we can remove pyasn1

adiroiban:
## Scope and purpose

Fixes #11844

The tests are failing with the latest pyans1 version.

Update our deps to depend on older pyasn1.

## Changes

only update the deps declaration.

### DIFF
--- a/src/twisted/conch/newsfragments/11844.bugfix
+++ b/src/twisted/conch/newsfragments/11844.bugfix
@@ -1,3 +1,2 @@
-The `pyasn1` dependency was updated to depend on version below 0.5.0.
-`pyasn1` intorduce a few changes that so far we know that break the Twisted
-test suite.
+The `pyasn1` dependency was updated to depend on a version below 0.5.0.
+`pyasn1` introduce changes that break the Twisted test suite.
--- a/src/twisted/conch/newsfragments/11844.bugfix
+++ b/src/twisted/conch/newsfragments/11844.bugfix
@@ -1,0 +1,3 @@
+The `pyasn1` dependency was updated to depend on version below 0.5.0.
+`pyasn1` intorduce a few changes that so far we know that break the Twisted
+test suite.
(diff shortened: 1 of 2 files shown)


adiroiban:
This should be ready for review.

Codecov.io coverage upload failed, and I triggered a retry

needs-review

src/twisted/conch/newsfragments/11844.bugfix
```diff
@@ -0,0 +1,3 @@
+The `pyasn1` dependency was updated to depend on version below 0.5.0.
+`pyasn1` intorduce a few changes that so far we know that break the Twisted
```
exarkun: Typo - `intorduce`

Why "so far as we know"?  Don't we just know it?

src/twisted/conch/newsfragments/11844.bugfix
```diff
@@ -0,0 +1,3 @@
...
+test suite.
```
adiroiban:
```suggestion
The `pyasn1` dependency was updated to depend on a version below 0.5.0.
`pyasn1` introduce changes that break the Twisted test suite.
```

Thanks for the review. Hope this is better :)

glyph (changes requested):
Pinning to an old version of a security-critical dependency is not an option that we should ever consider.

The only acceptable version of this would be to add a `[old_broken_ssh_keys]` extra that would add the pin, that is *not* part of `all` of `all_non_platform`.

adiroiban:
Thanks Glyph for the review.
I think that this can be closed as won't fix.

The effort should be done on merging https://github.com/twisted/twisted/pull/11843

PR #11847, opened 2023-04-23
#11846 Deprecate 'alternate' OpenSSH private keys in Twisted Conch.

### LINKED ISSUE #11846: Raise a warning if non-standard OpenSSH private keys are used with Twisted Conch SSH.
This is just preparing the removal of pyasn1 and using only `cryptography`

The pyans1 removal is done here https://github.com/twisted/twisted/pull/11843 but with that PR, some OpenSSH keys that used to work with Conch, will no longer work.

The plan is to deprecated these OpenSSH private keys format.

adiroiban:
## Scope and purpose

Fixes #11846

Depends #11844  since we need and older pyasn1.

It looks like at some point OpenSSH was handling non-standard PEM keys.
This was done at https://github.com/twisted/twisted/issues/3008

The plan is to remove support for these private keys, so that we can remove our custom PEM handling code and reuse the `cryptography` OpenSSH code for handling OpenSSH PEM files.

### DIFF
--- a/src/twisted/conch/newsfragments/11846.removal
+++ b/src/twisted/conch/newsfragments/11846.removal
@@ -1,0 +1,3 @@
+twisted.conch.ssh.keys.Key support for loading "alternate" OpenSSH private keys is now deprecated.
+These are some private keys that at some point were handled by OpenSSH but for which no specification exists.
+For more info about these OpenSSH keys see https://github.com/twisted/twisted/issues/3008.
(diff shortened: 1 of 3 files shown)


adiroiban:
This is ready for review... but https://github.com/twisted/twisted/issues/11844 should be merged first

needs-review

adiroiban:
The plan is to just remove support for pyasn1, without any deprecation process.

Issue #11590, opened 2022-07-29
Fix pydoctor errors generated after the pydoctor upgrade


adiroiban:
In https://github.com/twisted/twisted/pull/11578 we have updated the pydoctor version and cleand the code.

Some errors were generated after the upgrade.

They should be fixed here.

While on this, maybe also look to move the pydoctor template from `src/twisted/python/_pydoctortemplates/` maybe to `/docs/pydoctor-templates` or something like that. 

To make it easier to discover it.

adiroiban:
Here is the list of errors generated by latest `pydoctor ~= 23.4.0` version
```
apidocs run-test: commands[0] | - pydoctor src/twisted
/twisted/src/twisted/trial/_dist/test/matchers.py:126: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/trial/_dist/workerreporter.py:117: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/trial/_dist/workerreporter.py:120: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/test/test_adbapi.py:362: Assignment to constant "DB_DIR" inside an instance is ignored, this value will not be part of the docs.
/twisted/src/twisted/trial/test/test_assertions.py:1610: Invalid call to twisted.python.deprecate.deprecated(), first argument should be a call to incremental.Version()
/twisted/src/twisted/trial/test/test_assertions.py:1618: Invalid call to twisted.python.deprecate.deprecated(), first argument should be a call to incremental.Version()
/twisted/src/twisted/trial/_dist/functional.py:34: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/trial/_dist/workerreporter.py:69: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/trial/_dist/workerreporter.py:134: bad docstring: Improper paragraph indentation.
/twisted/src/twisted/conch/checkers.py:117: Cannot find link target for "twisted.conch.checkers.pwd.struct_passwd", resolved from "pwd.struct_passwd"
/twisted/src/twisted/conch/checkers.py:136: Cannot find link target for "spwd.struct_spwd" (you can link to external docs with --intersphinx)
/twisted/src/twisted/conch/checkers.py:41: Cannot find link target for "twisted.conch.checkers.pwd.struct_passwd", resolved from "pwd.struct_passwd"
[...]

Issue #11583, opened 2022-07-15
Update towncrier


adiroiban:
We use on old version of towncrier with a custom script to detect release branch.

The latest towncrier should have support to checking a release branch.

We should update to towncrier and remove our custom script.

Any good stuff used in twisted/twisted custom towncrier script should be moved upstream.

PR #11860, opened 2023-05-06
#11859 Fix Read The Docs build process and update documentation deps.

### LINKED ISSUE #11859: Fix Read The Docs build
We a are buidling the docs with python 3.8 and some deps are no longer working  inside the RTD build environment.

adiroiban:
## Scope and purpose

Fixes #11859

We are building our docs with Python 3.8 and somehow the Read The Docs build environment no longer support this.

This PR tries to fix the RTD build.

It also tries to update the libraries that we use to build the documentation.

The goal is to build the docs using Sphinx 6 version.

The deps allow Sphinx v5 to support Python 3.7

This updated towncrier just to see that all is ok.
There is a separate ticket to use `towncrier` CLI for our CI, so that we don't have to reinvent the wheel https://github.com/twisted/twisted/issues/11583

This is also a separate ticket to cleanup pydoctor errors, so I think it's ok to just use latest pydocotor here  -  https://github.com/twisted/twisted/issues/11590


adiroiban:
needs-review

adiroiban:
The RTD looks ok

* https://twisted--11860.org.readthedocs.build/en/11860/
* API https://twisted--11860.org.readthedocs.build/en/11860/api/index.html

I see the checks are green, but there are failures

* sphinx doc build - https://github.com/twisted/twisted/actions/runs/4902054289/jobs/8753675388?pr=11860#step:5:165
* pydoctor apidic build - https://github.com/twisted/twisted/actions/runs/4902054289/jobs/8753674134?pr=11860#step:5:37


needs-changes

src/twisted/newsfragments/11859.bugfix
```diff
@@ -0,0 +1 @@
+The Read The Docs documentation build process was updated to latest Python and Sphinx version to fix a build error.
```
twm: I don't think that we really need this sort of build maintenance in the changelog.

twm (approved):
LGTM &mdash; I just made the same change in treq. Thank you @adiroiban!

adiroiban:
Thanks, Tom for the review.

The Sphinx build now looks ok.

I have enabled auto-merge to get trunk green.

I hope there will be a new version of `cachecontrol` released soon ... but it might take some time for urllib3 support to be added. I see that `cachecontrol` is vendored in [pip](https://github.com/pypa/pip/tree/main/src/pip/_vendor) so I hope it will be fixed at some point soon

@glyph with this change, we pin urllib to an older version... but only for the `dev-release` extra.
Since this is not used for production deps, I hope this is ok

glyph:
> @glyph with this change, we pin urllib to an older version... but only for the `dev-release` extra.
> Since this is not used for production deps, I hope this is ok

any pins to older versions are fine as long as we have a clear plan for getting out of them and documenting why they happen. even in the docs build, this will eventually break or have some version conflict with something else :).  Can you file a ticket on urllib, or cachecontrol, or sphinx, or wherever the real problem is, and link to it from an issue on Twisted so we will at least know when we can remove it?

pyproject.toml
```diff
@@ -59,18 +59,22 @@ test = [
...
-    "readthedocs-sphinx-ext ~= 2.1",
-    "sphinx >= 4.1.2, <6",
+    "towncrier ~= 22.12",
+    "pydoctor ~= 23.4.0",
+    "sphinx-rtd-theme ~= 1.2",
+    "readthedocs-sphinx-ext ~= 2.2",
+    # We can move to v6 once we no longer need to support Python 3.7.
+    "sphinx >= 5, <7",
+    # This is due to cachecontrol not being compatible with urllib3>=2.0
+    # See https://github.com/ionrock/cachecontrol/issues/293
```
glyph, line 69: 👍 thanks, love to see it

PR #11861, opened 2023-05-09
[#9885] add `tls:` server endpoint that supports SNI, PEM parsing, and supporting code that changes OpenSSL connection/context building

### LINKED ISSUE #9885: Server TLS endpoint
|[twm's avatar](https://github.com/twm)| @twm reported|
|-|-|
|Trac ID|trac#9885|
|Type|enhancement|
|Created|2020-07-01 05:08:13Z|

Add a server-side counterpart to `wrapClientTLS`, as described here: https://twistedmatrix.com/pipermail/twisted-web/2020-March/005369.html

Don't forget the strports plugin.

```
trac-id__9885 9885
type__enhancement enhancement
reporter__twm twm
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1593580093447844 1593580093447844
changetime__1648438453436296 1648438453436296
version__None None
owner__None None

```
</details>

glyph:
## Scope and purpose

Fixes #9885 

This also makes it possible to do `twist web --path whatever --listen=tls:path/to/certbot/config/live` if `path/to/certbot/config/live` is a directory containing `.pem` files contining private keys, certificates, and chains in any layout (as, for example, the [certbot](https://certbot.eff.org)  live config dir does.


glyph:
please review

glyph:
okay this clearly needs slightly more test coverage, but given that it's kind of a big refactoring and introduces a new interface, I'd love to get some feedback before I pour more effort into it

glyph:
(although i will probably be working on that coverage gap regardless, just want to let people chime in if anyone has time)

src/twisted/internet/_sslverify.py
```diff
@@ -1123,7 +1140,26 @@ def __init__(self, hostname, ctx):
...
+        self._configureContext = configureContext
+        self._configureConnection = configureConnection
+
+    def createClientCreator(
+        self,
+        connectionSetupHook: Callable[[SSL.Connection], None],
+        contextSetupHook: Callable[[SSL.Context], None],
+    ) -> IOpenSSLClientConnectionCreator:
+        """
+        Create a client creator.
```
adiroiban: This is private ... so not a big name

but maybe we can have a better docstring.

To me, this docstring doesn't add any more info that I alreay got by reading the function signature `def createClientCreator`

Why not call it "factory" ? like `getClientOptionsFactory` ?

just asking :)

src/twisted/internet/_sslverify.py
```diff
@@ -1141,9 +1177,19 @@ def clientConnectionForTLS(self, tlsProtocol):
         @return: the configured client connection.
         @rtype: L{OpenSSL.SSL.Connection}
         """
+        if self._ctx is None:
+            self._ctx = self._createContext()
+            # Note: remember client options for testing.  See
+            # twisted.internet.test.test_endpoints.tlsHostnameFromEndpoint
+            self._ctx._clientOptions = self
```
adiroiban: Maybe something like this 

"remember client for testing" seemls like a task for the developers 

```suggestion
            # Note: To help with testing, we keep a reference to the TLS options.
            $ See
            # twisted.internet.test.test_endpoints.tlsHostnameFromEndpoint
            self._ctx._clientOptions = self
```

adiroiban (commented):
I left a few commnents.

I am not very familiar with the serverFromString API.

I was expecting to see more documentation in this branch, covering how or why to use it.

------

I don't know when I will have time to fully review this.

I think that as long as it has good documentation and code coverage, we can merge it.

If anyone else is interested in this API, we can update it later.

Thanks again.

adiroiban:
I had to deal with Twisted context factory about 10 years ago.
I got it working somehow, and since then I forgot about it.

@glyph if you have time, maybe we can do a 15-30 minutes voice/screen sharing meeting and look into giving some feedack.

I am a bit lost into all the SSL context wrapping API.
I have only worked with IOpenSSLContextFactory and I have a custom implementation, without using much code from Twisted.

src/twisted/internet/_sslverify.py
glyph: This code really suffers from a bunch of iterations of evolving an interface, where they all kind of do the same thing but progressively more correctly. `OptionsFactory2`, `OptionsFactory3` looks ugly and doesn't express the differences between them, so I tried to find synonyms ("creator" rather than "factory") to help draw the distinction

glyph:
> @glyph if you have time, maybe we can do a 15-30 minutes voice/screen sharing meeting and look into giving some feedack.

That would be great.

adiroiban (dismissed):
Thanks for working on this.

I am requesting a change for this PR.


I think that the biggest issue here is the lack of documentation for `serverFromString` and the new TLS option.

How a developer should know that the `tls:` server string is available and what are the available options ?


------------

`serverFromString` API is [introduced here as](https://docs.twistedmatrix.com/en/latest/core/howto/tutorial/intro.html#do-nothing): 

```
fingerEndpoint = endpoints.serverFromString(reactor, "tcp:1079")
fingerEndpoint.listen(FingerFactory())
reactor.run()
```

Glyph mention that this is not the use case for which `serverFromString` was designed for. 

-----------

We also have the [documentation page dedicated to endpoints](https://docs.twistedmatrix.com/en/latest/core/howto/endpoints.html#endpoint-types-included-with-twisted) but this PR is not updating that page.

I think that this PR should add an information that the SSL endpoint should not be used, and add a few words about the new TLS server endpoint.

glyph:
Attempting to address the coverage gaps in the legacy compatibility code, I wrote these notes, which lead to me filing #12411 :

--------

both ALPN and NPN are asymmetric.  In NPN, the callback is client-side and the static advertised list is server-side.  In ALPN, the callback is server-side and the static advertised list is client-side.

OpenSSL allows per-Connection customization of the static ALPN list, but the callback *must* be configured on the connection

or to put it more precisely, SSL_set_alpn_protos exists but SSL_set_alpn_select_cb does not, only SSL_CTX_set_alpn_select_cb.

as of this branch, twisted compat has 3 generations

- oldest: implicit getContext
- older: explicit IOpenSSLContextFactory
- old: IOpenSSLClientConnectionCreator/IOpenSSLServerConnectionCreator
- new (in this branch): IOpenSSLClientCreationCreatorFactory/IOpenSSLServerConnectionCreatorFactory

if application code supplies an "old" connection creator, the application does not supply the Context to the framework; we can only get it from the returned Connection object.

thus, IF:

1. the application code has supplied an "old" connection creator,
2. AND the application has supplied a factory providing IProtocolNegotiationFactory to TLSMemoryBIOFactory,
3. AND the application is making a *server*-side connection,

it is no longer (and was never!) even *possible* to negotiate the set of ALPN protocols that the application wants.

the new connection creator interface could techncially make this work, because it passes in the context hook to allow Twisted to twiddle the context, but that's ... really just bad, as it confuses the responsibility for selecting a list of ALPN protocols.  The application should do that in one place, when it configures the connection creator object.  Twisted should never look at an OpenSSL context to configure it for you after the fact.

we also need to file some follow-ups here to clean up the previous two generations of compatibility and make sure the docs focus on the new one

glyph:
Full coverage punch list, described in terms of actual interesting semantics to test rather than lines:

- [x] NPN advertise callback (might just need to skip that since NPN isn't available any more)
- [x] NPN modified test (`test_NPNAndALPNNoOverlap`), same deal
- [ ] SNI is sent for a non-None name, but lookup returns None, and needs to recurse looking for wildcards
- [ ] SNI lookup fully fails, connection is dropped, KeyError is logged
- [ ] maybe just … sending SNI at all? the SNI tlsext_servername_callback is just never invoked, so maybe `_lookupContext` is only called by the default-context lookup???
- [ ] key-pair matching failed (directory contains public cert but not private key); need to implement a little logging (in `PEMObjects.inferDomainMapping`
- [ ] `IOpenSSLServerConnectionCreator` compatibility path for legacy compat (i.e. "new" path before this branch existed, medium-"old" path now)

glyph:
> I think that the biggest issue here is the lack of documentation for `serverFromString` and the new TLS option.

Yep, this is for sure necessary to merge this.

src/twisted/internet/_sslverify.py
glyph: Thankfully I was able to just eliminate this.

src/twisted/internet/_sslverify.py
glyph: yay, also gone

glyph:
please review

glyph:
This may still need some work but it's definitely time for another look now that all coverage gaps are fixed.

adiroiban:
Thanks for moving this forward .

I am still not sure how people are expected to learn about this new Twisted functionality.
It this intended to be uses as API/code with documentation provided as `apidocs` or this is an turn-key functionality with `narrative`  docs?

I am not very familiar with the serverFromString API.

I was expecting to see narrative documentation in this branch, covering how or why to use it... and also describe the available options 

This PR is huge so maybe I am missing something

glyph:
> I am still not sure how people are expected to learn about this new Twisted functionality. It this intended to be uses as API/code with documentation provided as `apidocs` or this is an turn-key functionality with `narrative` docs?

This is a good point. The narrative documentation does still need to be updated. It's really simple to use, but the documentation in the "servers" section here https://docs.twisted.org/en/stable/core/howto/endpoints.html#servers needs to be updated with a similar note to the "clients" section here https://docs.twisted.org/en/stable/core/howto/endpoints.html#:~:text=SSL%20(Deprecated)

docs/core/examples/tls_alpn_server.py
```diff
@@ -30,7 +30,7 @@
     openssl s_client -connect localhost:8080 -alpn h2,http/1.1
     openssl s_client -connect localhost:8080 -nextprotoneg h2,http/1.1
```
twm: You should remove mention of `-nextprotoneg` as well, no?

docs/core/howto/ssl.rst
twm, line 1: The [overview section of this document](https://docs.twisted.org/en/stable/core/howto/ssl.html#overview) needs an update to mention the `tls:` endpoint, rather than `SSL4ServerEndpoint`.

Consider moving the ALPN section up. It seems more relevant these days than `startTLS`.

The [conclusion section of this document](https://docs.twisted.org/en/stable/core/howto/ssl.html#overview) should mention ALPN.

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+
+log = Logger()
+
+
+@implementer(IOpenSSLServerConnectionCreator)
+@dataclass
+class SNIConnectionCreator(object):
+    _contextLookup: Callable[[bytes | None], Context | None]
+
+    def _lookupContext(self, name: bytes | None) -> Context:
```
twm: Could you add a docstring here (or on the class) that explains the double-lookup?

The "coverage" comments here are rather cryptic. Do they mean "I wrote this on many lines rather than one expression so that coverage would ensure that all the branches are hit"?

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+        1. too much re-scanning; re-reading full file contents for every single
+           certificate even if only one has changed.  a mtime/length cache
+           would be a good place to start with this
+
+        2. too trusting; we get a network request for a billion certificate
+           names per second, we go ahead and do a bunch of work every single
+           time (and, see point 1, re-scan and re-parse every single file)
+
+        3. not *enough* re-scanning on the happy path; if certificates go
+           stale, we just let them sit there until we get an unknown hostname
```
twm, line 185: Could we at least, like, check the mtime of the cert directory to address some of these issues?

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+
+        4. we don't look at notAfter/notBefore, so if we find multiple certs,
+           we may end up using the wrong one
```
twm, line 188: Is this a practical concern with certbot, or purely theoretical?

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+@dataclass
+class PEMObjects:
+    """
+    A collection of objects loaded from a PEM encoded file.
+    """
+
+    certificates: List[tuple[FilePath[str], Certificate]]
+    keyPairs: List[tuple[FilePath[str], KeyPair]]
```
twm: These need docs.

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+                with fp.open() as f:
+                    subself = cls.fromLines(fp, f)
+                    self.certificates.extend(subself.certificates)
+                    self.keyPairs.extend(subself.keyPairs)
+        return self
+
+    @classmethod
+    def fromLines(cls, fp: FilePath[str], pemlines: Iterable[bytes]) -> PEMObjects:
+        """
+        Load some objects from the lines of a PEM binary file.
```
twm: I wouldn't call PEM _binary_:

```suggestion
        Load some objects from the lines of a PEM-encoded file.
```

Also, the parameters need docs. Would a function that takes `fp` alone be a better public interface?

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,281 @@
...
+    A collection of objects loaded from a PEM encoded file.
+    """
+
+    certificates: List[tuple[FilePath[str], Certificate]]
+    keyPairs: List[tuple[FilePath[str], KeyPair]]
+
+    @classmethod
+    def fromDirectory(cls, directory: FilePath[str]) -> PEMObjects:
+        """
+        Load a single PEMObjects from all the PEMs in a big directory.
```
twm: Should say something like "recursively walk a directory looking for files with a `.pem` extension".

Also, should this ignore hidden files? Right now it would load a file with a basename of `.pem`.

src/twisted/internet/endpoints.py
```diff
@@ -2419,3 +2427,38 @@ def parseStreamClient(reactor, *args, **kwargs):
         @rtype: L{IStreamClientEndpoint}
         """
         return _parseClientTLS(reactor, *args, **kwargs)
+
+
+@implementer(IPlugin, IStreamServerEndpointStringParser)
+class _TLSServerEndpointParser:
```
twm, line 2458: This needs to be documented in `docs/core/howto/endpoints.rst`

src/twisted/internet/interfaces.py
```diff
@@ -2373,20 +2373,19 @@ def getPeerCertificate() -> object:
...
-        completed this property may incorrectly return L{None}: wait until data
-        has been received before trusting it (see
-        https://twistedmatrix.com/trac/ticket/6024).
+        The protocol selected to be spoken using ALPN.  If the remote peer does
+        not support ALPN, or ALPN is not available via local TLS libraries,
+        C{negotiatedProtocol} will be L{None}.  Otherwise,
+        C{negotiatedProtocol} will be the name of the selected protocol as
+        C{bytes}.  Until the TLS handshake has completed, this property may
+        incorrectly return L{None}: wait until data has been received before
+        trusting it.  See U{https://twistedmatrix.com/trac/ticket/6024}.
```
twm:
```suggestion
        trusting it.  See U{https://github.com/twisted/twisted/issues/6024}.
```

src/twisted/internet/endpoints.py
```diff
@@ -2419,3 +2427,38 @@ def parseStreamClient(reactor, *args, **kwargs):
...
+        return TLSServerEndpoint(
+            TCP6ServerEndpoint(reactor, int(port), int(backlog), interface),
+            SNIConnectionCreator(autoReloadingDirectoryOfPEMs(p)),
+        )
+
+    def parseStreamServer(
+        self, reactor: IReactorCore, *args: Any, **kwargs: Any
+    ) -> IStreamServerEndpoint:
+        """
+        Parse a TLS stream server.
```
twm:
```suggestion
        Parse a TLS stream server endpoint string.
```

src/twisted/internet/_sslverify.py
```diff
@@ -1080,42 +1021,56 @@ class ClientTLSOptions:
...
         encoding.  ASCII values, however, will always work.
-    @type _hostnameASCII: L{unicode}
 
     @ivar _hostnameIsDnsName: Whether or not the C{_hostname} is a DNSName.
         Will be L{False} if C{_hostname} is an IP address or L{True} if
         C{_hostname} is a DNSName
-    @type _hostnameIsDnsName: L{bool}
+
+    @ivar _sendServerName: Whether the hostname will be sent via the TLS
+        ServerName Indicatino extension.
```
twm: Typo:

```suggestion
        Server Name Indication extension.
```

src/twisted/internet/_sslverify.py
```diff
@@ -1126,69 +1081,78 @@ def __init__(self, hostname, ctx):
...
-                    verifyIPAddress(connection, self._hostnameASCII)
+                verify_service_identity(extract_patterns(cert), [svcid], [])
             except VerificationError:
+                ourVerifyResult = False
                 f = Failure()
-                transport = connection.get_app_data()
-                transport.failVerification(f)
+                assert weakProtoRef is not None
+                weakProtoRef.failVerification(f)
+        except BaseException:
```
twm, line 1094: I take it bad stuff happens if you raise in an OpenSSL callback?

src/twisted/internet/_sslverify.py
```diff
@@ -87,52 +113,6 @@ def _getExcludedTLSProtocols(oldest, newest):
     return excludedVersions
 
 
-class SimpleVerificationError(Exception):
-    """
-    Not a very useful verification error.
-    """
-
-
-def simpleVerifyHostname(connection, hostname):
```
twm: :tada: I'm very happy to see this go!

src/twisted/web/test/test_http2.py
```diff
@@ -2938,6 +2937,8 @@ async def test_realRequest(self) -> None:
         """
         Twisted's HTTP/2 server can talk to a third-party HTTP/2 client.
         """
+        from twisted.test.test_sslverify import certificatesForAuthorityAndServer
```
twm, line 2940: Circular import? It feels like this belongs in `twisted.internet.testing` or `twisted.test.ssl_helpers`.

twm (commented):
It's awesome to see this functionality! Thank you for taking this on, @glyph.

I do see a few coding standard issues:

* There are a bunch of missing docstrings
* The new files need a copyright header (it feels like we should have tooling for this?)

I also flagged some more specific issues inline. I'm not sure when I'll have time to come back to this, so just leaving this review as a comment.

docs/core/examples/tls_alpn_server.py
glyph: Yup.

src/twisted/protocols/_sni.py
glyph: > The "coverage" comments here are rather cryptic. Do they mean "I wrote this on many lines rather than one expression so that coverage would ensure that all the branches are hit"?

Sorry, that was just a `todo` comment, meant to mark out some uncovered lines. It's just complete now, so I'll remove them.

src/twisted/protocols/_sni.py
glyph: If you point at the `config/live` directory, it's theoretical. All the certificates in there should be current.  But `config/archive` will have lots of expired certs in it.

src/twisted/protocols/_sni.py
glyph: I think if we are going to add more code here it would just be easier to check the cert itself.

src/twisted/protocols/_sni.py
glyph: > Also, the parameters need docs. Would a function that takes `fp` alone be a better public interface?

Heh, had this thought even before getting to this review point. Yes.

src/twisted/internet/_sslverify.py
glyph: Yeah. Will add a comment for that.

src/twisted/internet/_sslverify.py
glyph: It was a relief to delete it :)

src/twisted/web/test/test_http2.py
glyph: I don't want to make it public, and I'm not confident in its factoring, so I'd rather just keep this one janky import rather than do a big project of refactoring it :)

src/twisted/protocols/_sni.py
glyph: > Should say something like "recursively walk a directory looking for files with a `.pem` extension".

yeah, good call.  Worth noting that this is a private object, though, not a public docstring.

> Also, should this ignore hidden files? Right now it would load a file with a basename of `.pem`.

I don't see why it should?

src/twisted/internet/endpoints.py
glyph: yep, for sure, this is the most important missing piece

src/twisted/protocols/_sni.py
glyph: I think these can be deferred for future work, given the broadly theoretical nature of the concern and the ease with which one can work around it (you shouldn't be throwing a bunch of certs you *don't* want to use at this endpoint, regardless)

glyph:
acking the perf regressions because they're showing up as regressions in an avx-specialized memcpy and the internals of fcntl, so this seems more likely to be a test environment change, or noise, than anything related to the branch

src/twisted/protocols/_sni.py
twm: That seems reasonable to me.

src/twisted/protocols/_sni.py
glyph: Various docstrings added, and coverage comments removed.

glyph:
please review

glyph:
@adiroiban @twm if either of you want to take a look at this, I think we are ready for a second round!

Still a bit too big, but nowhere near as overwhelming now that I've deleted some irrelevant stuff and merged the initial SNI support separately.

docs/core/examples/echoclient_ssl.py
```diff
@@ -8,18 +8,17 @@
...
-    echoClient = yield endpoint.connect(factory)
-
+    options = ssl.optionsForClientTLS("example.com", trustRoot=authority)
+    tcpEndpoint = endpoints.HostnameEndpoint(reactor, "localhost", 8000)
+    endpoint = endpoints.wrapClientTLS(options, tcpEndpoint, reactor)
     done = defer.Deferred()
-    echoClient.connectionLost = lambda reason: done.callback(None)
-    yield done
+    echoClient = await endpoint.connect(factory)
+    echoClient.dataReceived = lambda data: done.callback(data)
```
adiroiban, line 20: I guess that this works. I know what is going on here... but this might be a bit complicated for new users... and I guess that this documentation is targeted to new users.

Not sure what is the best way to demonstrate this.

I was expecting to have a separate deferred on the echo protocol like `firstDataReceived` that is automatically called first time data is received

docs/core/examples/server.pem
```diff
@@ -1,31 +1,123 @@
...
-A1UEBhMCVVMxIjAgBgkqhkiG9w0BCQEWE2V4YW1wbGVAZXhhbXBsZS5jb20xEDAO
-BgNVBAcTB2V4YW1wbGUwgZ8wDQYJKoZIhvcNAQEBBQADgY0AMIGJAoGBAKkRahIc
-Fp0V44QYpO9ue7mjZNbZYPAC8caoQC1jUgL42CT40PcoOiZWLgRk+Qw6P7PoJzO/
-T4ufK0qoPUJm1jErDRWy9eWlGLE0grPECM+jxFfLXxJLKdPtuwMA8Ip72JMirFN5
-Y/JTBZOR3j5a/mbY5tcRqgffKxm4QQegnhiBAgMBAAEwDQYJKoZIhvcNAQEEBQAD
-gYEAWANPpp985nXMoIwHlsSMm8ijkk7XQU3oioCYDcM6pLT+mvBDe1mZc8mUlrWy
-Zo/lT6HF44SHIZ0zCgPYwTpWV6C0K+/kKlYBERZ3ajrzGf5ACfUTNyk5P81C68mc
-9fQ7lhq1iuNKzVh8b746Z4ufn6iI1VygnyOQ1hZ/lOX56TA=
------END CERTIFICATE-----
+# coding: utf-8
```
adiroiban, line 1: Why it this file called `server.pem` ?

Maybe it would help to have a top docstring describing the purpose of this file.... I guess that this information is found in the main docs... but it would help to have some info , if you get the file outside of the docs.

adiroiban:
Thanks Glyph for the update. I am looking into this.. I hope to get this done by the end of this week.

glyph:
> Thanks Glyph for the update. I am looking into this.. I hope to get this done by the end of this week.

Thanks! It's been 30 months, a few more days is no big deal ;)

docs/core/examples/echoserv_ssl.py
```diff
@@ -16,7 +17,9 @@ def main(reactor):
     certData = getModule(__name__).filePath.sibling("server.pem").getContent()
     certificate = ssl.PrivateCertificate.loadPEM(certData)
     factory = protocol.Factory.forProtocol(echoserv.Echo)
-    reactor.listenSSL(8000, factory, certificate.options())
+    tcpEndpoint = TCP6ServerEndpoint(reactor, 8000)
+    endpoint = wrapServerTLS(certificate.options(), tcpEndpoint, reactor)
+    endpoint.listen(factory)
     return defer.Deferred()
```
adiroiban, line 23: Not sure if this can explain better what is going on here

I feel that for new users, there is a lot of "magic" happening here.

```suggestion
    # This will be fired when the port is bound,
    # but we return a deferred that is never fired to keep the
    # reactor loop always active.
    deferred = endpoint.listen(factory)
    deferred.addCallback(lambda _: defer.Deferred())
    return deferred
```

Maybe also rewrite this example to `async def main(reactor)` similar to docs/core/examples/ssl_clientauth_client.py to encourage new users to start with "best practices"

docs/core/examples/ssl_clientauth_client.py
```diff
@@ -8,20 +8,23 @@
...
+        clientCertificate=clientCertificate,
+    )
+    tcpEndpoint = endpoints.HostnameEndpoint(reactor, "localhost", 8000)
+    endpoint = endpoints.wrapClientTLS(options, tcpEndpoint, reactor)
     done = defer.Deferred()
-    echoClient.connectionLost = lambda reason: done.callback(None)
-    yield done
+    echoClient = await endpoint.connect(factory)
+    echoClient.dataReceived = lambda data: done.callback(data)
+    print(await done)
```
adiroiban, line 27: Maybe just a small comment to explain what is going on here

```suggestion
    # Wait to receive the first data and then end the execution.
    # For this example, we don't do any cleanup or TLS shutdown request.
    print(await done)
```

docs/core/howto/endpoints.rst
```diff
@@ -354,18 +354,14 @@ TCP (IPv6)
...
-   Please refer to `OpenSSL's documentation on dhparam <http://www.openssl.org/docs/apps/dhparam.html>`_ for further details.
+TLS
+    The ``tls:`` string endpoint is a wrapper, that wraps a TCP endpoint.
+    By default, it wraps ``tcp:443``.
+    Its first argument is a path to a directory where you keep your PEM-format certificate and private key files.
+    It will automatically match keys, certificates, and intermediate certificate bundles up with each other for you, and determine which ones to use based on the server name identification sent by the client.
+    The remaining arguments are all arguments to the ``tcp`` endpoint type.
 
-   For example, ``ssl:port=443:privateKey=/etc/ssl/server.pem:extraCertChain=/etc/ssl/chain.pem:sslmethod=SSLv3_METHOD:dhParameters=dh_param_1024.pem``.
+    For example, ``tls:/etc/ssl/private:443:interface=1.2.3.4``.
```
adiroiban, line 364: Maybe use a better name and describe the structure of the directory ... `/etc/ssl/private` might make you think that you have a single file, to the private key....

I know that the docs above mention that you need to pass the path to a directory

Is this directory structure compatible with `certbot` ?
In that case, the path should be `/etc/letsencrypt/live/` ... so we configured with the root path for all certificate names ... and not a specific certificate.
 

```suggestion
    For example, ``tls:/etc/ssl/server-name:443:interface=1.2.3.4``.
    
    The `/etc/ssl/server-certs` will contain the following files,
    which are compatible with the directory structure used by `certbot`
    to store the files for a single certificate name.
    
    * /etc/ssl/server-name/cert.pem
    * /etc/ssl/server-name/privkey.pem
    * /etc/ssl/server-name/chain.pem 
```

docs/core/howto/ssl.rst
```diff
@@ -28,8 +27,13 @@ To ensure that you have the required additional libraries installed, please ``pi
 TLS Quick Start
 ---------------
 
-To set up a TLS server, use the ``ssl:`` string endpoint prefix.
-If you're using a command-line tool with a ``--listen`` argument, such as ``twist web``, you can create an ``ssl:`` endpoint that serves your certificate on port 443.
+To set up a TLS server, use the ``tls:`` string endpoint prefix.
+If you're using a command-line tool with a ``--listen`` argument, such as ``twist web``, you can create an ``tls:`` endpoint that serves your certificate on port 443, by doing ``twist web --listen tls:/path/to/certbot/config/live:443``.
```
adiroiban, line 31: Maybe give a more "exact" path used by certbot ?

I am not sure if you should pass `/etc/letsencrypt/live/certname` or `/etc/letsencrypt/live` 

I guess that SNI is supported... so you pass `/etc/letsencrypt/live` 

```suggestion
If you're using a command-line tool with a ``--listen`` argument, such as ``twist web``, you can create an ``tls:`` endpoint that serves your certificate on port 443, by doing ``twist web --listen tls:/etc/letsencrypt/live:443``.
```

src/twisted/protocols/_sni.py
```diff
@@ -0,0 +1,351 @@
...
+        log.warn(msg, name=name, path=path)
+        doReload()
+        return lookup(name, False)
+
+    doReload()
+    return lookup
+
+
+@dataclass
+class PEMObjects:
```
adiroiban, line 215: I think that this code is ok.
I don't think that we need to reuse the code from https://pypi.org/project/pem/


But I feel that this is use case is quite commmon.

Maybe it make sense to try to get these type of helpers into `cryptography` ?

src/twisted/internet/test/test_endpoints.py
```diff
@@ -2954,6 +2987,376 @@ def createClientEndpoint(self, reactor, clientFactory, **connectArgs):
...
+        we will see a legible logged error.
+        """
+        logObserver = EventLoggingObserver.createWithCleanup(self, globalLogPublisher)
+        self.connectionTest(ConnectionLost)
+        self.assertEqual(
+            formatEvent(logObserver[-1]),
+            "no server certificate for name b'endpoint-test.example.com'",
+        )
+
+    setattr(test_noCertsAtAll, "noCertsAtAll", True)
```
adiroiban, line 3331: Just asking... maybe I am missing something.

Why we can't pass the `noCertsAtAll=True` as an argument to `self.connectionTest(ConnectionLost)` ?

This comment is valid for all the other tests in which we use setattr

src/twisted/internet/test/test_endpoints.py
```diff
@@ -2954,6 +2987,376 @@ def createClientEndpoint(self, reactor, clientFactory, **connectArgs):
...
+        """
+        return {"timeout": 10, "bindAddress": ("localhost", 49595)}
+
+    def listenArgs(self):
+        """
+        @return: C{dict} of keyword arguments to pass to listen
+        """
+        return {"backlog": 100, "interface": "127.0.0.1"}
+
+    def setUp(self):
```
adiroiban, line 3098: instead of implict call to `setUp`, I think that we can have an explicit  `createClientContext(serverServiceIdentity, clientServiceIdentity... plus more)` that is called at the start of each test.

I think that this should make each test "self contained" and make it easier to read the test.

All the function attributes are kind of arbitrary and it's hard to tell which affects the result of the test

adiroiban (commented):
Thanks for the changes.

For the documentation I tried to use the "new user hat" ... and review it as it was read by someone new to Twisted...
I might be wrong with some comments as maybe even new users are already familiar with some concept and we don't need to go into the implementation details.

... you can alwasy read the source code

But I guess that these days many software developers are too busy with various tasks and they just try to find the fastest way to get the job done, without doing a full documentation read, reading a book about Twisted or reading the source code. 

------

I don't know if we have enought tests and documentation for the certificates directory structur supported for Twisted

I guess that we support anything created by `certbot` ... but does Twisted support other directory structures ? or maybe it also support flat directories with random certifcate names.

glyph:
> I guess that we support anything created by `certbot` ... but does Twisted support other directory structures ? or maybe it also support flat directories with random certifcate names.

Yes, it should support just about any structure, it hoovers up all the certificates, matches them with their keys, and looks them up by hostname.

docs/core/examples/echoclient_ssl.py
glyph: Do you think that we can file a follow-up issue for this? I feel like this is an improvement on what was there, and I don't want to block this for another 3 months while we try to come up with a whole new way to document this.  (This, in particular, is a tradeoff to be slightly more confusing about a not-particularly-interesting detail of the example in order to make the example much shorter. It might not be the best tradeoff but coming up with something better might be a significant project.)

docs/core/examples/server.pem
glyph: The intention with this file is to be opaque for users; you can use it as a certificate for the examples, and not look at it.  Hence, server.pem. The code here is just a helper for maintainers to regenerate the key & certificate when the example stops working due to key size or certificate property constraints that may emerge in the future.

docs/core/examples/echoserv_ssl.py
glyph: This is example code; the prose around it ought to be explaining what is happening. I would really like to keep it at zero comments. Expanding the example with lots of inline comments makes the narrative doc harder to read because you're hit with a wall of code that makes it impossible to even see the prose above and below without scrolling back and forth. Inline end-of-line comments to explain some detail might be okay but a whole extra paragraph is too distracting.

docs/core/howto/endpoints.rst
glyph: > Maybe use a better name and describe the structure of the directory ... `/etc/ssl/private` might make you think that you have a single file, to the private key....

`/etc/ssl/private` is the standard location for private certificate data in many linux distributions. I don't know if it's in LFS, but e.g. Ubuntu documents it officially: https://documentation.ubuntu.com/server/explanation/security/certificates/

I've never seen `/etc/ssl/server-name` as an idiom before.

docs/core/howto/ssl.rst
glyph: > I am not sure if you should pass `/etc/letsencrypt/live/certname` or `/etc/letsencrypt/live`

I put `/live` on the end of the path to illustrate that you should be pointing at the live directory. Pointing at `…/certname` would only let you serve a single cert which would defeat the point of all this logic I added to collect SANs and map them using SNI :)

glyph: I had not seen "/etc/letsencrypt" as a standard location before. I can find various forum posts indicating it's a place that some people put stuff in their config files but I haven't seen where it's documented as a default.

src/twisted/protocols/_sni.py
glyph: > I don't think that we need to reuse the code from https://pypi.org/project/pem/

I do love @hynek but it does seem like too small a micro-dependency to take, particularly given the fact that the dependency would become circular (it has Twisted support code in it).

glyph: > Maybe it make sense to try to get these type of helpers into `cryptography` ?

It's a bit of a weird / niche edge case; I kept this as private API for us because I don't want to deal with all the edge cases around leading and trailing data, encodings, etc. It would be great if `cryptography` handled it but I don't know that it's worth the effort?

src/twisted/internet/test/test_endpoints.py
glyph: Definitely a valid criticism and not the way I would set up these tests if I had to create them new today.

And speaking of how I would not set them up new, the reason for these attributes being the way they are is that a bunch of test cases come from `EndpointTestCaseMixin`, by way of `ServerEndpointTestCaseMixin` and `ClientEndpointTestCaseMixin`. Those tests are going to *implicitly* depend on `setUp` being called, which is why we can't have every test explicitly call a specific setUp.

If I were doing this today I'd definitely arrange everything in terms of a composable fixture object and a protocol, rather than inheritance and action-at-a-distance attributes.  But fixing this in-line would require probably another thousand lines of diff to restructure the entire massive endpoints suite.  `setattr` is ugly but it's worth it to avoid blowing up all the existing parametric test setup here.

src/twisted/internet/test/test_endpoints.py
glyph: Nope, we can't. As explained above. In addition to the other problems already mentioned, attempting to do this would mean copying `test_interface`, `test_endpointConnectSuccess`, `test_endpointConnectFailure`, `test_endpointConnectingCancelled`, `test_endpointConnectNonDefaultArgs`, `test_endpointListenSuccess`, `test_endpointListenFailure`, `test_endpointListenNonDefaultArgs`, and every subsequent new cross-cutting endpoint test to `TCP4EndpointsTests`, `TCP6EndpointsTests`, `TCP6EndpointNameResolutionTests`, `HostnameEndpointMemoryIPv4ReactorTests`, `HostnameEndpointMemoryIPv6ReactorTests`, `HostnameEndpointsOneIPv4Tests`, `HostnameEndpointsOneIPv6Tests`, `SSL4EndpointsTests`, `TLSEndpointsTests`, `UNIXEndpointsTests` and `AdoptedStreamServerEndpointTests`.

Assuming each copied test requires a def line, a 3-line docstring, and a minimum of 1 trailing blank, that is a minimum of 440 lines of additional diff right now, plus 40 new lines for any new endpoint test we want to add.  And that's before we count the additional 22 lines of new setup code spread across all these copy/pasted tests.

We could, of course, write some code to automate this, with `setattr` on the class or something, but then we'd be re-implementing our own version of inheritance and setUp and that strikes me as doubly pointless.

docs/core/examples/echoserv_ssl.py
glyph: I had a look at the relevant prose and I was going to try to add something there, but I felt like I already explained it, so I don't think I'm going to change anything here.

We really need a bigger project on this too, to get some new users in to read through the docs and tell us what is wrong (I suspect they will give us a list with a million items on it, and this one will be down in the high hundreds of thousands ;))

glyph:
Having given some explanations for why things are the way they are I don't feel like I have anything left to reply to. It looks like the workflow bot has a bug because it doesn't notice "comment" reviews remove the "review request"; this is still in "needs review" without showing up on the twisted-contributors review dashboard.

glyph:
please review

glyph:
Nice to see that it still can respond by re-adding the request though!

adiroiban (approved):
Thanks for the info.

I think that we should merge this... looks into Python 3.14 support and then make a new release.

Once this is available, people can start using it and if there are any issues we can fix them later.

Thanks again!

glyph:
@adiroiban thanks for all the reviews, glad to finally land this! Given that this is pretty big and complicated, I'll probably ask for more feedback to get any quick fixes or polishes into the release.

PR #11870, opened 2023-06-05
line 537 is a noop in python3

Zectbumo:
using map in python3 without iterating does nothing failing to register checkers.

## Scope and purpose

Fixes #

## Contributor Checklist:


adiroiban:
Thanks for looking into this. It looks like the current CI errors are  not related to the changes from this PR.

The last CI run on `trunk` was green... it looks like this is another case of a test failure generated by un-pinned dependencies

adiroiban:
Thanks for the fix. 
I have never used the Twisted  MailService code  before, but by reading the code, it makes sense and the fix looks good.

I can see that the test are now green.

From what I can see, the tests were green even before this fix... so it means that we don't have automated tests for this part.

Would it be possible to also update the automated tests to make use of the fix from here?


Thanks

Zectbumo:
> From what I can see, the tests were green even before this fix... so it means that we don't have automated tests for this part.

Right, that's because most of the tests have been disabled due to python 3. 

> Would it be possible to also update the automated tests to make use of the fix from here?

I have made a larger set of modifications to make the mail portion python 3 compatible and tried to re enable several of the tests again. I was working through a good number of tests up until I ran out of time and couldn't revisit the code again. I think I was close to finishing except it was a bit of a can of worms. The more I opened the tests up the more worms that crawled out.

adiroiban:
Thanks for the follow-up.

If you managed to get a single test on Py3, I think that it is worth sending a patch for it.

It is much better than not running any test at all.

We don't need to get all the tests running from a single PR.

------------

If  the current mail service is not yet py3 ready and doesn't work on Py3, then I am not sure if this change will help in any way :)

--------

I encourage you to push your current code/patch as it is.
Even if it's not 100% ready
If someone else would want to run the mail service on py3 , they can use it as a starting point.

---------

I have never used the Twisted mail functionality before and I am not familiar with this part of Twisted.
I am just providing high level feedback

Zectbumo:
Good point. Yes, I have activated and passed several py3 tests already. I will commit as is then.

Zectbumo:
> push your current code/patch as it is

done #11898

glyph:
please review

src/twisted/mail/mail.py
```diff
@@ -533,8 +533,7 @@ def addDomain(self, name, domain):
         @type domain: L{IDomain} provider
         @param domain: A domain object.
         """
-        portal = Portal(domain)
-        map(portal.registerChecker, domain.getCredentialsCheckers())
+        portal = Portal(domain, domain.getCredentialsCheckers())
```
graingert, line 536: probably this code is (was) important

```suggestion
        portal = Portal(domain, domain.getCredentialsCheckers())
        for credentialsChecker in domain.getCredentialsCheckers():
            portal.registerChecker(credentialsChecker)
```

graingert: this test would fail without this - if it were not skipped

https://github.com/twisted/twisted/blob/0315f0b05740597d9efbac4bde4cdcce779f8794/src/twisted/mail/test/test_mail.py#L983-L987

glyph: Can we unskip it in *this* branch though? If so, great, if not, then we might just need to land #11898

glyph:
Since this PR does not have a news fragment, or a ticket, or a unit test, I am going to close it and open an issue describing the problem.

glyph:
Superseded by #12432

PR #11872, opened 2023-06-05
#11871 Use Ubuntu 22.04 for GitHub Action jobs.

### LINKED ISSUE #11871: Update test environment to have all test pass
This looks like a `pynacl` related issue, but `1.5.0`  is used in both the initial successful run and the new failed re-run.

There were some changes in `cryptography==41.0.0` 

This run was green https://github.com/twisted/twisted/actions/runs/4921915478/jobs/8792209998

And the re-run is red - https://github.com/twisted/twisted/actions/runs/4921915478

See what option we have to get `trunk` CI green again.

so this might be due to our detection of ED25519

``
if default_backend().ed25519_supported():
```

and looking at the code, I don't know why pynacl is installed in the first place.
`couch-pynacl` should be an option dependency and act as manual fallback.

adiroiban:
## Scope and purpose

Fixes #11871

Try to get the GHA tests for Linux from 20.04 to 22.04


adiroiban:
@cjwatson  Only if you have time. Can you take a look and suggest how we might get this fixed.

It looks like there is an incompatibility between `pynacl` (always at 1.5.0) and `cryptography`

* With `cryptography==41.0.1` tests are failing
* With `cryptography-40.0.2` all is ok.

In the cryptography change I don't see anything that might suggest this failure https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst

The changes might be ok, and is just the test code which is using private method that needs to be updated.

I have to go now... but I will try to continue to troubleshoot this

alex:
My suggestion would be to delete https://github.com/twisted/twisted/blob/trunk/src/twisted/conch/ssh/_keys_pynacl.py. ed25519 is supported on effectively any modern openssl.

adiroiban:
The goal of this PR is to get our tests on Ubuntu 22.04

I think that is best as this is closest to what most people are using in PROD these days

needs-review

graingert (approved):

PR #11873, opened 2023-06-09
#11871 Make tests on trunk pass again by removing PyNaCl optional dependency

### LINKED ISSUE #11871: Update test environment to have all test pass

glyph:
## Scope and purpose

Fixes #11871


glyph:
Per @alex 's advice https://github.com/twisted/twisted/pull/11872#issuecomment-1578624106

worth noting that while 1.1.1 is not technically completely EOL, it will be dead (not security supported upstream) in 3 months

https://endoflife.date/openssl

alex:
1.1.1 has ed25519

glyph:
> 1.1.1 has ed25519

oh. looks like I misread the "<1.1.1" message.

I don't know why we even added this so recently, the release adding it was in 2022 but 1.1.0 "Ended 3 years and 9 months ago" according to endoflife.date

glyph:
please review

alex (approved):

adiroiban:
I think that this should be merged.

Just some context.

This was added quite recently by @cjwatson to support Ubuntu 18.04  in https://github.com/twisted/twisted/pull/1607 - released in 22.4.0 

I think Ubuntu 18.04 is now EOL, so it's safe to remove it.

Thanks @glyph and @alex for the fix and feedback

adiroiban (approved):
All good. thanks

adiroiban:
I have merged this to fix `trunk`

I will use the other PR to look into upgrading our CI to Ubuntu 22.04

---------

GitHub Actions no longer supports Ubuntu 18.04 ... so another reason to remove support for Ubuntu 18.04 from  Twisted.

alex:
All supported versions of both Debian and Ubuntu have versions of
OpenSSL with ed25519 support.

On Fri, Jun 9, 2023 at 10:20 AM Adi Roiban ***@***.***> wrote:
>
> I think that this should be merged.
>
> Just some context.
>
> This was added quite recently by @cjwatson to support Ubuntu 18.04 in #1607 - released in 22.4.0
>
> I think Ubuntu 18.04 is now EOL, so it's safe to remove it.
>
> Thanks @glyph and @alex for the fix and feedback
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.Message ID: ***@***.***>



-- 
All that is necessary for evil to succeed is for good people to do nothing.

PR #11876, opened 2023-06-14
#11875 Expose new method seek in _InputStream class

### LINKED ISSUE #11875: _InputStream should expose seek method so that we can read environ["wsgi.input"] more than once
In our application we do `environ["wsgi.input"].read()` to parse the request payload. In some case we then rewrite the request and on some others we leave the request as is. When we leave it as-is we will need to access `environ["wsgi.input"]` later down the line but since the cursor is at the end of the stream when we read it again it's empty. We need to be able to place the cursor back at the beginning like this: `environ["wsgi.input"].seek(0)`.

I already implemented the solution and it works (PR incoming). It should not have any impact since it's a new method that is exposed.

Wolfcast:
Fixes #11875

In our application we do `environ["wsgi.input"].read()` to parse the request payload. In some case we then rewrite the request and on some others we leave the request as is. When we leave it as-is we will need to access `environ["wsgi.input"]` later down the line but since the cursor is at the end of the stream when we read it again it's empty. We need to be able to place the cursor back at the beginning like this: `environ["wsgi.input"].seek(0)`.

### DIFF
--- a/src/twisted/web/newsfragments/11875.feature
+++ b/src/twisted/web/newsfragments/11875.feature
@@ -1,0 +1,3 @@
+twisted.web.wsgi now exposes new seek metod in class _InputStream
+
+The new seek method in class _InputStream (twisted.web.wsgi) allows you to move the position in the stream allowing you to skip part of the stream or read the stream or (parts of) more than once.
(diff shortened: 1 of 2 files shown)


glyph:
@Wolfcast Hi! Just checking to make sure you deleted/closed this on purpose and it was not an artifact of some other Git operation.

Wolfcast:
Yeah fixed this issue by doing a copy of the object instead of reusing it (which needed the "seek" fix as stated above). I still think this could be useful to some people but I needed a fast fix for our production app so I did the "copy" fix. Thank you.

PR #11884, opened 2023-06-25
#10310 Avoid circular reference with `Deferred._canceller`

### LINKED ISSUE #10310: There is a persistent reference cycle in deferLater().
|[JCoder3's avatar](https://github.com/JCoder3)| @JCoder3 reported|
|-|-|
|Trac ID|trac#10310|
|Type|defect|
|Created|2022-02-16 06:26:36Z|

A TCP server port is used to listen for connections.

When port.stopListening() (in twisted\internet\tcp.py) is called, there is memory leak.

It's due to a cyclic reference in deferLater() (in twisted\internet\task.py).

It could be solved by enforcing
**deferred._canceller = None** 
where it is returned.

Maybe it's better to have a method for clearing deferred._canceller so as to break the reference loop.

Attachments:

* [cyclic_reference.JPG](https://raw.githubusercontent.com/twisted/twistedmatrix.com-trac-attachments/trunk/ticket/5c2/5c202c1f703b4df85bd3f0123eb1c4038ade3e14/6ad32adb0c7bac520c28ce51c5eb60b7c7501d25.JPG) (80191 bytes) - added by JCoder3 on 2022-02-16 06:29:09Z - the reference cycle

```
trac-id__10310 10310
type__defect defect
reporter__JCoder3 JCoder3
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1644992796628301 1644992796628301
changetime__1646083036271745 1646083036271745
version__None None
owner__JCoder3 JCoder3
cc__TBBle
```
</details>

erikjohnston:
Fixes #10310

This is to avoid circular references where the canceller holds a reference to the deferred (which is often the case).

It's safe to clear the `_canceller` after the Deferred has finished since we only call `_canceller` if the Deferred hasn't finished. Note that https://github.com/twisted/twisted/issues/10310#issuecomment-1168428766 suggests that its not safe to make this change, however the only place I can see that `_canceller` is used is behind a check against `.called`:

https://github.com/twisted/twisted/blob/c462c5f7d940260ea3ca553c8653d0d227bbf468/src/twisted/internet/defer.py#L768-L771

I can't see anywhere where we reset `called`, and the docs fairly clearly state it will remain `True`.

### DIFF
--- a/src/twisted/newsfragments/10310.misc
+++ b/src/twisted/newsfragments/10310.misc
@@ -1,1 +1,1 @@
-Fix circular references when using a `Deferred` with a canceller that references the `Deferred`, e.g. `DelayedCall` and `inlineCallbacks`.
+When using a `Deferred` with a canceller that references the `Deferred` (ex. `DelayedCall` and `inlineCallbacks`), a canceller reference from the `Derefferd` is removed a soon as the `Deferred` is called. In this way, the canceller should have a higher change of being garbage collected.
--- a/src/twisted/newsfragments/10310.misc
+++ b/src/twisted/newsfragments/10310.misc
@@ -1,0 +1,1 @@
+Fix circular references when using a `Deferred` with a canceller that references the `Deferred`, e.g. `DelayedCall` and `inlineCallbacks`.
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1700,6 +1700,8 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
 
         deferred.callback(None)
 
+        # We manually remove the reference from the local
+        # function scope, to avoid interfering with the test.
         del deferred
         del canceller
 
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1684,8 +1684,8 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
     @pyunit.skipIf(_PYPY, "GC works differently on PyPy.")
     def test_canceller_circular_reference(self) -> None:
         """
-        Test that a circular reference between a `Deferred` and its canceller
-        is broken when the deferred is resolved.
+        A circular reference between a `Deferred` and its canceller
+        is broken when the deferred is called.
         """
 
         # Create a canceller and weak reference to track if its been freed.
@@ -1707,15 +1707,19 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         self.assertIsNone(weakCanceller())
 
 
-class _TestCircularCanceller:
+class DummyCanceller:
     """
-    c.f. L{DeferredTest.test_canceller_circular_reference}
+    A callable that does nothing.
+    It is intended to be used just to get an object reference
+    to support GC related tests.
     """
 
     deferred: Optional[Deferred[Any]] = None
 
-    def __call__(self, deferred: Deferred[Any]) -> None:
-        pass
+    def __call__(self, deferred: Deferred[Any]) -> None:  # pragma: no cover
+        """
+        This is not expected to be called as part of the current test suite.
+        """
 
 
 def _setupRaceState(numDeferreds: int) -> tuple[list[int], list[Deferred[object]]]:
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1757,9 +1757,8 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         deferred: Deferred[Any] = Deferred(canceller)
         canceller.deferred = deferred
 
-        deferred.callback(None)
-
         deferred.addCallback(lambda _: Deferred())
+        deferred.callback(None)
 
         del deferred
         del canceller
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1712,6 +1712,8 @@ class _TestCircularCanceller:
     c.f. L{DeferredTest.test_canceller_circular_reference}
     """
 
+    deferred: Optional[Deferred[Any]] = None
+
     def __call__(self, deferred: Deferred[Any]) -> None:
         pass
 
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -1682,14 +1682,14 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
 
     @pyunit.skipIf(_PYPY, "GC works differently on PyPy.")
-    def test_canceller_circular_reference(self) -> None:
+    def test_canceller_circular_reference_callback(self) -> None:
         """
         A circular reference between a `Deferred` and its canceller
         is broken when the deferred is called.
         """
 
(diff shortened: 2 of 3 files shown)


src/twisted/newsfragments/10310.misc
```diff
@@ -0,0 +1 @@
+Fix circular references when using a `Deferred` with a canceller that references the `Deferred`, e.g. `DelayedCall` and `inlineCallbacks`.
```
adiroiban: Maybe this is a bugfix

just trying to understand the scope of this change 

```suggestion
When using a `Deferred` with a canceller that references the `Deferred` (ex. `DelayedCall` and `inlineCallbacks`), a canceller reference from the `Derefferd` is removed a soon as the `Deferred` is called. In this way, the canceller should have a higher change of being garbage collected.
```

adiroiban (changes requested):
Thanks for the changes and creating this PR.

The change looks good and the comment make sense.

I am not sure how to write a test for this :)

I am also reading the comment from @exarkun  here https://github.com/twisted/twisted/issues/10310#issuecomment-1168428763

It is true that you can have a deferred, for which you call `.callback(result)` , and then later, that `result` will trigger one of the callbacks to return another deferred.

But my understanding is that the `canceller` is there only for the only direct `callback` invocation.
If the deferred, was previously called and is waiting for a second deferred, when cancel is called, the `cancellere of the second` deferred is called.

So I would say that this is fine... based on my reading of the code :)
I might be wrong.

As a test, I think  that even just a simple set of unit tests which checks that `deferred._canceller` is None after errback or callback should be good enough.

erikjohnston:
@adiroiban Looks like the `macos` failure is a flakey test? Otherwise, the codecov seems to be complaining that the new test helper class is untested.

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1681,42 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+        # Once all local references have been dropped, the canceller should have
+        # been freed.
+        self.assertIsNone(weakCanceller())
+
+
+class _TestCircularCanceller:
+    """
+    c.f. L{DeferredTest.test_canceller_circular_reference}
+    """
```
adiroiban:
```suggestion
class DummyCanceller:
    """
    A callable that does nothing.
    It is intended to be used just to get an object reference
    to support GC related tests.
    """
```

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1681,42 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+    deferred: Optional[Deferred[Any]] = None
+
+    def __call__(self, deferred: Deferred[Any]) -> None:
+        pass
```
adiroiban: I think it's important to add the `pragma: no cover` marker if we don't expect this line to be called.

```suggestion
    def __call__(self, deferred: Deferred[Any]) -> None:  # pragma: no cover
        """
        This is not expected to be called as part of the current test suite.
        """
```

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1681,42 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+        # Create a canceller and weak reference to track if its been freed.
+        canceller = _TestCircularCanceller()
+        weakCanceller = weakref.ref(canceller)
+
+        # Create a Deferred with the above canceller, adding a reference to the
+        # deferred to the canceller to create the circular reference.
+        deferred: Deferred[Any] = Deferred(canceller)
+        canceller.deferred = deferred
+
+        deferred.callback(None)
```
adiroiban, line 1701: Maybe it is worth extending the test to check that the canceller is "free" even when the deferred doesn't have a "final" result yet.

```suggestion
        pendingResult = Deferred()
        deferred.callback(pendingResult)
```

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1681,42 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
         for thing in thingsThatAreNotCoroutines:
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
 
+    @pyunit.skipIf(_PYPY, "GC works differently on PyPy.")
+    def test_canceller_circular_reference(self) -> None:
+        """
+        Test that a circular reference between a `Deferred` and its canceller
+        is broken when the deferred is resolved.
```
adiroiban: This should target the "callback" function.
It's important to have a similar test for `errback()`

```suggestion
        A circular reference between a `Deferred` and its canceller
        is broken when the deferred is called.
```

See https://jml.io/test-docstrings

adiroiban (changes requested):
Thanks for the changes.

I re-run the macOS tests...but I guess is a flaky test :(

`_startRunCallbacks` is a private method that is called by both the `callback` and `errback` public API.
I think it's important to test the behavior of these both public APIs

I left a few minor inline comments.

src/twisted/test/test_defer.py
erikjohnston: Ah, didn't know about `# pragma: no cover`

erikjohnston:
@adiroiban Thanks, I've added a couple of more tests as you suggested.

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1681,107 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+        # Create a Deferred with the above canceller, adding a reference to the
+        # deferred to the canceller to create the circular reference.
+        deferred: Deferred[Any] = Deferred(canceller)
+        canceller.deferred = deferred
+        weakDeferred = weakref.ref(deferred)
+
+        deferred.callback(None)
+
+        del deferred
```
adiroiban, line 1705:
```suggestion
        # We manually remove the reference from the local
        # function scope, to avoid interfering with the test.
        del deferred
```

adiroiban (approved):
Changes look good.

The Deferred._canceller is a private member set by `Deferred.__init__` and I think it's safe to assume that any class from `Deferred` can set it to  None at any time...since it's something private.

I prefer this fix, as at least for me, it's easy to understand.

I think that we should merge this.

If anything breaks, we can revert this.

Thanks for the work on this.

Thanks

erikjohnston:
Thanks for the reviews ❤️

src/twisted/newsfragments/10310.misc
```diff
@@ -0,0 +1 @@
+When using a `Deferred` with a canceller that references the `Deferred` (ex. `DelayedCall` and `inlineCallbacks`), a canceller reference from the `Derefferd` is removed a soon as the `Deferred` is called. In this way, the canceller should have a higher change of being garbage collected.
```
adiroiban, line 1: @erikjohnston 

Is this a .misc, like ... don't bother letting other people know about this change.

`.bugfix` -- like things were broken in previous Twisted version , see if you might be affected and consider an update

`.feature` -- only consider an update if you plan make to use of this improvement

erikjohnston: I think the only people that care about this are applications with large and non standard memory usage. You could argue for any of the three tbh, but I think few people will care?

For Synapse I think we'd consider this a misc change (but we still include those in the changelogs)

exarkun: Twisted doesn't include the text of misc fragments in the news file.

Issue #11830, opened 2023-03-21
process tests are flaky


glyph:
Tests which call spawnProcess, particularly recently `twisted.python.test.test_sendmsg`, have become flaky enough that they're interfering with the downstream CI of the Cryptography project.

This has recently become worse with the introduction of posix_spawn, although it hints at a similar bug underlying our own file-descriptor logic.

This is a fix for the former since I couldn't find anything as obviously wrong with the latter.

Issue #10332, opened 2022-04-12
Fix flaky test generated by warnings


twisted-trac:
|[<img alt="adiroiban's avatar" src="https://avatars.githubusercontent.com/u/204609?s=50" width="50" height="50">](https://github.com/adiroiban)| @adiroiban reported|
|-|-|
|Trac ID|trac#10332|
|Type|defect|
|Created|2022-04-12 12:09:21Z|

I think that we need to run a warning check at the end of each test to detect which test generates a warning

On Windows, the tests are generating warnings that are not checked.

inside t.i.test.test_posixbase on Windows 

There is also t.t.teste_twisted test_setUidSameAsCurrentUid

If one test wants to check for the warnings, it will receive all the previous "unuexpected" warnings.

See a test run here https://dev.azure.com/twistedmatrix/twisted/_build/results?buildId=4668&view=logs&j=a9f5be88-9f44-53f4-896b-687d3085a015&t=50e42789-57d9-5d74-e8a4-3a4e15f86668&l=16003

Note that test fails at `test_delayedCall` but most probably the warning was generated by another test.

Not sure if this is a bug with the way the tests are written and not waiting for cleanup, or there is a bug on Windows

```
[ERROR]
Traceback (most recent call last):
  File "D:\a\1\s\.tox\alldeps-withcov-windows\lib\site-packages\twisted\internet\test\test_posixbase.py", line 55, in tearDown
    len(warnings), 0, f"Warnings found at the end of the test:\n{warnings}"
  File "D:\a\1\s\.tox\alldeps-withcov-windows\lib\site-packages\twisted\trial\_synctest.py", line 422, in assertEqual
    super().assertEqual(first, second, msg)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\unittest\case.py", line 852, in assertEqual
    assertion_func(first, second, msg=msg)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\unittest\case.py", line 845, in _baseAssertEqual
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: 2 != 0 : Warnings found at the end of the test:
[...]

Issue #6635, opened 2013-07-20
inlineCallbacks causes local variables to have very long lifetime.


twisted-trac:
|<img alt="blaisegassend's avatar" src="https://avatars.githubusercontent.com/u/0?s=50" width="50" height="50">| blaisegassend reported|
|-|-|
|Trac ID|trac#6635|
|Type|defect|
|Created|2013-07-20 00:35:29Z|

Currently returnValue raises an exception causing the stack frame, include all the caller's local variables to be stored by sys. This is annoying because it increases object lifetime making tests that check for memory leaks difficult to write. 

Worse, it stores parts of the traceback as local variables creating a memory loop: _inlineCallbacks is in the traceback created when raising _DefGen_Return and that traceback is now in _inlineCallbacks's local variables. This means that the locals from the caller to returnValue will persist until the next garbage collection.

The following patch against 13.1.0 fixes the problem:

```
--- defer.py    2013-06-12 15:27:40.000000000 -0700
+++ defer_modified.py   2013-07-19 17:33:50.330581820 -0700
@@ -21,7 +21,7 @@
 import traceback
 import types
 import warnings
-from sys import exc_info
+from sys import exc_info, exc_clear
 from functools import wraps
 
 # Twisted imports
@@ -1110,6 +1110,9 @@
                         ultimateTrace.tb_frame.f_code.co_name,
                         appCodeTrace.tb_frame.f_code.co_name),
                     DeprecationWarning, filename, lineno)
+                ultimateTrace = None
+            appCodeTrace = None
+            exc_clear()
             deferred.callback(e.value)
             return deferred
         except:
```

<details><summary>Searchable metadata</summary>

```
trac-id__6635 6635
type__defect defect
reporter__blaisegassend blaisegassend
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__ 
time__1374280529000000 1374280529000000
changetime__1374282224000000 1374282224000000
version__None None
owner__ 

```
</details>

twisted-trac:
|[<img alt="itamarst's avatar" src="https://avatars.githubusercontent.com/u/3266662?s=50" width="50" height="50">](https://github.com/itamarst)<a name="note_2"></a>|@itamarst commented|
|-|-|

Thanks for the patch! In the future, could you attach it as a file? Also, a unit test would be great if you could manage it. I think weakrefs are the way to test things don't have extra references lying around.

PR #11886, opened 2023-06-25
Avoid circular reference with `defer.inlineCallbacks`

### LINKED ISSUE #11885: `_inlineCallbacks` machinery has a circular reference
The `_inlineCallbacks` machinery currently causes a circular reference to be made, which stops Python from freeing the deferreds etc immediately. 

This is caused by usages of inline functions. The fix is to move those functions out into standalone ones. (PR incoming).

Similar to https://github.com/twisted/twisted/issues/10310.

erikjohnston:
## Scope and purpose

Fixes #11885

~~Requires #11884 to work, and so is based on that (otherwise the test fails).~~

This PR does make the marginally code style worse, in that we now have a bunch of free functions that really should be bundled together. However, given how frequently `_inlineCallbacks` is used (and not just for `@inlineCallbacks`) I think this trade-off is worth it to for the memory performance.


erikjohnston:
I've merged in https://github.com/twisted/twisted/issues/11885, so this should now be ready to be properly looked at.

I'm not sure why the code coverage is failing, I tried to add a test to hit the partial line, but I don't fully grok the machinery going on in there.

I also cheekily tried to make `twisted.test.test_twistd.UnixApplicationRunnerStartApplicationTests.test_setUidSameAsCurrentUid` test less flakey (c.f. [this test run](https://github.com/twisted/twisted/actions/runs/5435987737/jobs/9885527570))

erikjohnston:
> I'm not sure why the code coverage is failing, I tried to add a test to hit the partial line, but I don't fully grok the machinery going on in there.

Ah, I managed to work that out!

src/twisted/test/test_twistd.py
```diff
@@ -1014,6 +1014,10 @@ def test_setUidSameAsCurrentUid(self):
         If the specified UID is the same as the current UID of the process,
         then a warning is displayed.
         """
+
+        # Flush any existing warnings.
+        self.flushWarnings()
```
adiroiban: I don't this that this is a good idea.

I guess that the warning is generated by another test and that test is not checking/clearing the warnings, and it breaks here.

I think it's best to see which tests generated the warning and fix that one here.

But there is one thing that can be done here to help troubleshoot this.

Instead of checking the length first, we can check the warning content first and this might give an idea of which test is generating the other warning.

```python
        warningsShown = self.flushWarnings()
        expectedWarning = (
            "tried to drop privileges and setuid {} but uid is already {}; "
            "should we be root? Continuing.".format(currentUid, currentUid)
        )
        self.assertEqual(expectedWarning, warningsShown[0]["message"])
        # Make sure we don't have more warnings.
        self.assertEqual(1, len(warningsShown))
```
The best option is to have a dedicated warning assertion, which will also show the traceback of the warning, in the case of a warning.

I think that fixing flaky test is very important... but just ignoring the error will not help... as next time it will be another test that will fail

src/twisted/test/test_defer.py
```diff
@@ -1681,6 +1683,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
 
+    def test_inlineCallbacksCancelCaptured(self) -> None:
+        """
+        Cancelling an `inlineCallbacks` correctly handles the function catching
+        the `CancelledError`.
+        """
+        cancellerMock1 = Mock()
+        cancellerMock2 = Mock()
+        d1: Deferred[Any] = Deferred(cancellerMock1)
+        d2: Deferred[Any] = Deferred(cancellerMock2)
```
adiroiban: I am not a fan of Mock, expecilly one without a specification 

Maybe something like this

```suggestion
        canceller1Calls = []
        canceller2Calls = []
        d1: Deferred[Any] = Deferred(lambda d: canceller1Calls.append(d))
        d2: Deferred[Any] = Deferred(lambda d: canceller2Calls.append(d))
```


My biggest issue with Mock assertion is that for example if you call

```diff
-cancellerMock1.assert_called_once()
+cancellerMock1.asert_called_once()
```

it will still pass

src/twisted/test/test_twistd.py
erikjohnston: It's not obvious to me that we want to assert that no warnings (including ones from dependencies) were generated at any point before this test. If we want to ensure that no warnings (except the ones we expect) to be raised then I think we might want to explicitly add that to the pre/post test hooks?

Though I think it's a fair idea to try and figure out what the warning is, so I've pushed a commit that will fail if there were any existing warnings.

adiroiban: Yes... a post test hook would be great... but this needs to be done in a separate ticket/PR.

I have created this ticket a long time ago https://github.com/twisted/twisted/issues/10332 ... but didn't had time to look into it.

src/twisted/test/test_twistd.py
```diff
@@ -1014,6 +1013,11 @@ def test_setUidSameAsCurrentUid(self):
         If the specified UID is the same as the current UID of the process,
         then a warning is displayed.
         """
+
+        # Assert that there were no existing warnings.
+        existing_warnings = self.flushWarnings()
+        self.assertEqual([], existing_warnings)
+
```
adiroiban, line 1021: Maybe add a link to the issue so that we will review this test once the test suite has some hooks for warnings.

```suggestion
        # FIXME:https://github.com/twisted/twisted/issues/10332
        # Assert that there were no existing warnings.
        existing_warnings = self.flushWarnings()
        self.assertEqual([], existing_warnings)

```

adiroiban:
I have accidentally found this ticket https://github.com/twisted/twisted/issues/6635

not sure if it's related... or if this will fix it.

erikjohnston:
Is `twisted.python.test.test_sendmsg.SendmsgTests.test_sendSubProcessFD` a known flakey test? I'm failing to a) reproduce it, and b) see how the changes would affect it.

```
Error: 
Failure: twisted.internet.defer.TimeoutError: <twisted.python.test.test_sendmsg.SendmsgTests testMethod=test_sendSubProcessFD> (test_sendSubProcessFD) still running at 120.0 secs

twisted.python.test.test_sendmsg.SendmsgTests.test_sendSubProcessFD
```

glyph:
> Is `twisted.python.test.test_sendmsg.SendmsgTests.test_sendSubProcessFD` a known flakey test?

Yes, although I think we already tried to address this? https://github.com/twisted/twisted/issues/11830

I've re-run the tests for you, because yes, this seems unrelated.

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
         for thing in thingsThatAreNotCoroutines:
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
 
+    def test_inlineCallbacksCancelCaptured(self) -> None:
+        """
+        Cancelling an `inlineCallbacks` correctly handles the function catching
```
adiroiban: what does "correctly" mean?

I guess that we are writing any test to make sure the code "works correctly".

Can you please add a few more details about what is expected to happen in this test.

What does correctly means?

If this test would fail in the future, how can you tell which behaviour is correct and which is not?

src/twisted/test/test_defer.py
```diff
@@ -1823,6 +1855,66 @@ def test_canceller_circular_reference_non_final(self) -> None:
         # been freed.
         self.assertIsNone(weakCanceller())
 
+    @pyunit.skipIf(_PYPY, "GC works differently on PyPy.")
+    def test_inlineCallbacksNoCircularReference(self) -> None:
+        """
+        Tests that there is no circular dependency when using
+        L{defer.inlineCallbacks}, so that the machinery gets cleaned up
+        immediately rather than waiting for a GC.
+        """
```
adiroiban: Maybe something like this, focusing on the behaviour that is tested here.

We don't have any code in the test that looks into the GC part.

```suggestion
        """
        When using L{defer.inlineCallbacks},
        after the function exits,
        it will not keep references to the function itself or the arguments.
        """
```

src/twisted/test/test_defer.py
```diff
@@ -1823,6 +1855,66 @@ def test_canceller_circular_reference_non_final(self) -> None:
...
+        # Create an object and weak reference to track if its gotten freed.
+        obj: Set[Any] = set()
+        objWeakRef = weakref.ref(obj)
+
+        async def func(a: Any) -> Any:
+            return a
+
+        # Run the function
+        funcD = Deferred.fromCoroutine(func(obj))
+        assert funcD.called
```
adiroiban: Maybe this assertion is better.

It makes sure that the code returns the expected result... not only that it is called.

It can be called with a different result... or called with  failure.

```suggestion
        self.assertEqual(obj, self.successResultOf(funcD))
```

src/twisted/test/test_defer.py
```diff
@@ -1823,6 +1855,66 @@ def test_canceller_circular_reference_non_final(self) -> None:
         # been freed.
         self.assertIsNone(weakCanceller())
 
+    @pyunit.skipIf(_PYPY, "GC works differently on PyPy.")
```
adiroiban: can you add a comment about why this tests fails on PyPy?
I guess that it's expected for PyPy to work differently.
Maybe maybe something like:

`del instance` will not immediately free the object in PyPy.

-------

Maybe for people using PyPy this is common sense.
I have never used PyPy, so feel free to ignore this comment

src/twisted/internet/defer.py
```diff
@@ -2085,6 +2105,48 @@ def gotResult(r: object) -> None:
             waiting[1] = None
 
 
+def _handleCancelInlineCallbacks(
```
adiroiban, line 2124: minor comment. just for readability.
Maybe move this after `def _addCancelCallbackToDeferred` as this looks like a help that is used only there

can it be nested in _addCancelCallbackToDeferred ? or this will not help with the circular reference fix?

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+        funcD = testFunc()
+        self.assertFalse(funcD.called)
+
+        funcD.cancel()
+
+        self.assertEqual(canceller1Calls, [d1])
+
+        funcD.cancel()
+        self.failureResultOf(funcD)
```
adiroiban: Maybe it would help to also check which failure we have.

Maybe there is a bug and we get a TypeError or ValueError and I don't this this is what we are expecting.
```suggestion
        failure = self.failureResultOf(funcD)
        # Add a check for the type of failure.
```

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+                yield d1
+            except Exception:
+                pass
+
+            yield d2
+
+        funcD = testFunc()
+        self.assertFalse(funcD.called)
+
+        funcD.cancel()
```
adiroiban: Maybe add a comment explaining the correct behaviour
```suggestion
        # On the first cancel it just handled by the
        # function by ignoring the exception.
        # The function still has no results as the
        # second deferred is not yet called.
```

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+        self.assertEqual(canceller1Calls, [d1])
```
adiroiban: Maybe also add a check for canceller2Calls ?
```suggestion
        self.assertEqual(canceller1Calls, [d1])
        self.assertEqual(canceller2Calls, [])
        # Still waiting for the second deferred.
        self.assertNoResult(funcD)
```

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,37 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+
+        funcD.cancel()
```
adiroiban:
```suggestion
        # Canceling again, will trigger the
        # cancellation errors as the second deferred
        # has no handling for cancellation.
        funcD.cancel()
```

adiroiban (changes requested):
Many thanks for the cleanup. It looks good.

I only have a minor comment on the test part as I find it hard to understand what is going on there on the first read.

The code over here is a bit complex, so I think it's important to have good test to explain what is the expected behaviour...so that in the future it's easier to decide which behaviour is a bug and which is expected

Kind regards

src/twisted/internet/defer.py
erikjohnston: I have moved this function down.

> can it be nested in _addCancelCallbackToDeferred ? or this will not help with the circular reference fix?

Generally having nested functions can cause circular references, so I'm loathed to nest them. That said, nesting them doesn't cause the tests to fail, but that is probably because the objects caught by any circular reference here is limited to a small set of object we can't test for.

src/twisted/test/test_defer.py
erikjohnston: Good point. I've tried to clarify, let me know if it needs some correction.

src/twisted/test/test_defer.py
erikjohnston: I've added a note to the docstring.

erikjohnston:
@adiroiban Thank you for the review! Hopefully the tests are a bit clearer now, but let me know if they're still a bit confusing as I'm happy to continue fixing them up (it is important).

And as always, thanks everyone for maintaining Twisted :slightly_smiling_face:

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,61 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
...
+            3. Cancelling the deferred again cancels any deferred the function
+               is waiting on, and the exception is raised.
+        """
+        canceller1Calls: List[Deferred[Any]] = []
+        canceller2Calls: List[Deferred[Any]] = []
+        d1: Deferred[Any] = Deferred(canceller1Calls.append)
+        d2: Deferred[Any] = Deferred(canceller2Calls.append)
+
+        @defer.inlineCallbacks
+        def testFunc() -> Any:
```
graingert: does this work with Deferred[object] isntead of Deferred[Any]?

```suggestion
        canceller1Calls: List[Deferred[object]] = []
        canceller2Calls: List[Deferred[object]] = []
        d1: Deferred[object] = Deferred(canceller1Calls.append)
        d2: Deferred[object] = Deferred(canceller2Calls.append)

        @defer.inlineCallbacks
        def testFunc() -> object:
```

erikjohnston: Good point, yes it does. Albeit with a more complicated return type for `testFunc`

src/twisted/newsfragments/11885.misc
```diff
@@ -0,0 +1 @@
+Fix circular references when using `defer.inlineCallbacks`.
```
adiroiban: Maybe we can call it a feature for CPython ... and name this file 11885.feature

```suggestion
`twisted.internet.defer.inlineCallbacks`, when using CPython`, after the deferred is called, it no longer keeps a reference to the wrapped function and arguments.
```

src/twisted/test/test_defer.py
```diff
@@ -1735,6 +1736,61 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
         for thing in thingsThatAreNotCoroutines:
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
 
+    def test_inlineCallbacksCancelCaptured(self) -> None:
```
adiroiban: Just asking

Is this test related to circular references?

Was this test created as there was regression that you discovered while refactoring the code, and the bug was not detected by the previous set of tests?

--------

I know that the existing code is already not ideal.

Maybe, as part of this PR, we can create a new `InlineCallbacksTests` class that will group all the tests targeting `inlineCallbacsk` usage

The current `DeferredTests` is a mess, as there are so many tests here and is hard to check for duplicate tests.

adiroiban (approved):
I think that it looks good.

The only minor comment is moving the tests targeting `inlineCallbacks` into a separate class, to make it easier to check the scope and coverage of the tests... and detect if we write duplicate tests for the same functionality/behaviour

src/twisted/test/test_defer.py
erikjohnston: The test isn't related to circular references, nor really a regression test. It's a new test that covers this particular edge case of cancellations that weren't covered by any other tests. It's a fiddly bit of logic, so didn't want to refactor it without more coverage.

---

Happy to move the tests into their own class. I'll do that now.

erikjohnston: Done.

src/twisted/newsfragments/11885.misc
erikjohnston: I've changed it to be a feature. I've tried to word it so that library users can better understand the impact, but I'm not sure if it's just too wordy. 

I'm happy for you to just change it if you want to. Words aren't my forte.

erikjohnston:
There's a test failure on

> twisted.test.test_twistd.UnixApplicationRunnerStartApplicationTests.test_setUidSameAsCurrentUid

I think that must be unrelated?

glyph:
Given that another test has been stuck in a pre-start state for 4 hours, I'm guessing that there was some blip in the CI runners.  Nothing on githubstatus.com, but, it could have been just on one host…

glyph:
Kicking off a failed re-run and hopefully we can fix those flakes.

erikjohnston:
Thanks!

> Given that another test has been stuck in a pre-start state for 4 hours, I'm guessing that there was some blip in the CI runners. Nothing on githubstatus.com, but, it could have been just on one host…

We've been seeing Github having issues all over the shop over the past few hours.

PR #11890, opened 2023-07-01
#11841 fix pending futures

### LINKED ISSUE #11841: asyncio.ensure_future not working in reactor with inlineCallbacks
If you use `await asyncio.ensure_future` inside a `Deferred`-awaiting coroutine you will produce an error: `builtins.RuntimeError: await wasn't used with future`
Example to reproduce:
```python
# import uvloop
import asyncio
# uvloop.install()

from twisted.web.resource import Resource
from twisted.web.server import NOT_DONE_YET


class ClockPage(Resource):
    isLeaf = True

    async def _some_background_task(self, request):
        print("hello!")

    async def _more_test(self, request):
        request.write(
            b"<!DOCTYPE html><html><head><meta charset='utf-8'>"
            b"<title></title></head><body>TEST</body></html>"
        )
        print("start")
        await asyncio.ensure_future(self._some_background_task(request))
        # request.finish()
        print("finish")

    @defer.inlineCallbacks
    def _real(self, request):
        yield defer.Deferred.fromCoroutine(self._more_test(request))
        request.finish()

        reactor.callLater(0, self._real, request)
        return NOT_DONE_YET

resource = ClockPage()

```

Run such s project like `twistd -r asyncio  -n web --path .` and refresh page a few time, after a small delay you will see such error. 

```
If you have a traceback, please paste it here; otherwise, delete this.
```
```
2023-04-17T19:08:32+0900 [twisted.internet.defer#critical] Unhandled error in Deferred:
2023-04-17T19:08:32+0900 [twisted.internet.defer#critical] 
          File "/test-twisted/env/lib/python3.11/site-packages/twisted/internet/defer.py", line 1697, in _inlineCallbacks
            result = context.run(gen.send, result)
          File "/test-twisted/test1.rpy", line 31, in _real
            yield defer.Deferred.fromCoroutine(self._more_test(request))
          File "/test-twisted/env/lib/python3.11/site-packages/twisted/internet/defer.py", line 1138, in fromCoroutine
            return _cancellableInlineCallbacks(coro)
          File "/test-twisted/env/lib/python3.11/site-packages/twisted/internet/defer.py", line 1857, in _cancellableInlineCallbacks
            _inlineCallbacks(None, gen, status, _copy_context())
          File "/test-twisted/test1.rpy", line 31, in _real
            yield defer.Deferred.fromCoroutine(self._more_test(request))
          File "/test-twisted/env/lib/python3.11/site-packages/twisted/internet/defer.py", line 1697, in _inlineCallbacks
            result = context.run(gen.send, result)
          File "/test-twisted/test1.rpy", line 25, in _more_test
            await asyncio.ensure_future(self._some_background_task(request))
        builtins.RuntimeError: await wasn't used with future
```

`asyncio.ensure_future` should work as other async/await functions. 

**Testing environment**
 - Operating System and Version; paste the output of these commands:
  `Linux rainbow 6.2.5-gentoo #1 SMP PREEMPT_DYNAMIC Mon Mar 13 00:22:30 JST 2023 x86_64 AMD Ryzen 9 5900X 12-Core Processor AuthenticAMD GNU/Linux`
but I can confirm on latest Debian as well. 
 - Twisted version 22.10.0
 - Reactor: asyncio
 - Python 3.11.4

**Other:**

This issue blocks to use aiohttp inside twisted.

stalkerg:
It should fix #11841. 

The main problem if `corouting.send` returns future in pending status which means we must back to the event loop and wait until it will be resolved. Otherwise, we will see the error "await wasn't used with future" which comes from the `send` code: https://github.com/python/cpython/blob/3.11/Modules/_asynciomodule.c#L1609 .

We are doing the same thing if we receive `Deferred` but for some reason ignored asyncio futures. This PR does the same thing for pending futures as for `Deferred`. 

Most problems come from `asyncio.ensure_future` function, which used by aiohttp. Without these changes not possible to use aiohttp inside Twisted even if you are using asyncio reactor and it reduces compatibility with tons of libraries.

### DIFF
--- a/src/twisted/newsfragments/11841.bugfix
+++ b/src/twisted/newsfragments/11841.bugfix
@@ -1,0 +1,1 @@
+`defer.Deferred.fromCoroutine` now correctly processes pending futures.
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -18,6 +18,7 @@ from asyncio import (
     CancelledError,
     Future,
     new_event_loop as _new_event_loop,
+    ensure_future,
 )
 from typing import (
     TYPE_CHECKING,
@@ -1677,6 +1678,26 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
 
         for thing in thingsThatAreNotCoroutines:
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
+
+    @ensuringDeferred
+    async def test_fromCoroutineWithEnsureFuture(self) -> None:
+        """
+        L{Deferred.fromCoroutine} should properly process pending futures
+        """
+        result = object()
+        async def test() -> object:
+            return result
+
+        async def ensure_wrapper() -> object:
+            return await ensure_future(test())
+
+        d = await Deferred.fromCoroutine(ensure_wrapper())
+
+        assert_that(
+            self.successResultOf(d),
+            is_(result),
+        )
+
 
 
 def _setupRaceState(numDeferreds: int) -> tuple[list[int], list[Deferred[object]]]:
--- a/src/twisted/test/test_defer.py
+++ b/src/twisted/test/test_defer.py
@@ -17,8 +17,8 @@ from asyncio import (
     AbstractEventLoop,
     CancelledError,
     Future,
-    new_event_loop as _new_event_loop,
     ensure_future,
+    new_event_loop as _new_event_loop,
 )
 from typing import (
     TYPE_CHECKING,
@@ -1685,6 +1685,7 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
         L{Deferred.fromCoroutine} should properly process pending futures
         """
         result = object()
+
         async def test() -> object:
             return result
 
@@ -1697,7 +1698,6 @@ class DeferredTests(unittest.SynchronousTestCase, ImmediateFailureMixin):
             self.successResultOf(d),
             is_(result),
         )
-
 
 
 def _setupRaceState(numDeferreds: int) -> tuple[list[int], list[Deferred[object]]]:
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 import traceback
 import warnings
 from abc import ABC, abstractmethod
-from asyncio import AbstractEventLoop, Future, iscoroutine
+from asyncio import AbstractEventLoop, Future, iscoroutine, isfuture
 from enum import Enum
 from functools import wraps
 from sys import exc_info
@@ -1874,7 +1874,8 @@ def _inlineCallbacks(
             status.deferred.callback(callbackValue)
             return
 
-        if isinstance(result, Deferred):
+        is_not_done_future = isfuture(result) and not result.done()
+        if isinstance(result, Deferred) or is_not_done_future:
             # a deferred was yielded, get the result.
             def gotResult(r: object) -> None:
                 if waiting[0]:
@@ -1883,6 +1884,8 @@ def _inlineCallbacks(
                 else:
                     _inlineCallbacks(r, gen, status, context)
 
+            if is_not_done_future:
+                result = Deferred.fromFuture(result)
             result.addBoth(gotResult)
             if waiting[0]:
                 # Haven't called back yet, set flag so that we get reinvoked
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1874,8 +1874,10 @@ def _inlineCallbacks(
             status.deferred.callback(callbackValue)
             return
 
-        is_not_done_future = isfuture(result) and not result.done()
-        if isinstance(result, Deferred) or is_not_done_future:
+        if isfuture(result) and not result.done():
+            result = Deferred.fromFuture(result)
+
+        if isinstance(result, Deferred):
             # a deferred was yielded, get the result.
             def gotResult(r: object) -> None:
                 if waiting[0]:
@@ -1884,8 +1886,6 @@ def _inlineCallbacks(
                 else:
                     _inlineCallbacks(r, gen, status, context)
 
-            if is_not_done_future:
-                result = Deferred.fromFuture(result)
             result.addBoth(gotResult)
             if waiting[0]:
                 # Haven't called back yet, set flag so that we get reinvoked


stalkerg:
please review

src/twisted/test/test_defer.py
```diff
@@ -1678,6 +1679,26 @@ def test_fromCoroutineRequiresCoroutine(self) -> None:
         for thing in thingsThatAreNotCoroutines:
             self.assertRaises(defer.NotACoroutineError, Deferred.fromCoroutine, thing)
 
+    @ensuringDeferred
+    async def test_fromCoroutineWithEnsureFuture(self) -> None:
+        """
+        L{Deferred.fromCoroutine} should properly process pending futures
```
adiroiban, line 1685: What does "properly" means ?  

I don't know why this has no coverage.
Is this test executed?

src/twisted/newsfragments/11841.bugfix
```diff
@@ -0,0 +1 @@
+`defer.Deferred.fromCoroutine` now correctly processes pending futures.
```
adiroiban, line 1: By reading the code, it looks like any `@defer.inlineCallbacks` is affected by this change.

I don't have any experience with futures and coroutines... so maybe "correctly" makes sense for someone who has experience.

But I think that it would help add more information about what is considered "correctly"

adiroiban (commented):
Many thanks for the  changes.

I don't have experience with `future` and it looks like the automated tests is not executed... so I can't do a proper review.

I left a few general comments regarding the PR.

I  think the main change looks good... but I think that this PR needs improvement on documenting the changes and the tests

Thanks again

src/twisted/test/test_defer.py
stalkerg: > What does "properly" means ?

It means we should back control the event loop before we can call `send` again. Not sure how to describe it. 
And yes, this test is executed, at least locally, and without these changes is failed.

src/twisted/newsfragments/11841.bugfix
stalkerg: Okey, I will explain it in more detail.

stalkerg:
@adiroiban, thanks for the review!

> I don't have experience with future and it looks like the automated tests is not executed... so I can't do a proper review.

It's triggered locally, do you have any idea why it's not working in CI?

> I think the main change looks good... but I think that this PR needs improvement on documenting the changes and the tests

Sure, I will try to improve it. Also, who can look into this PR with futures and coroutines experience?

stalkerg:
A little more background for this issue. 
`Deferred.fromCoroutine` is not entirely correct. It's even described in the documentation:
> Coroutine functions return a coroutine object, similar to how generators work.

`courutine.send` in the asyncio it return awaitable object, which includes `__await__`. It's not a coroutine automatically and can be Future which is close to Derefer and should be scheduled separately on the event loop if it's not already resolved(done). 

Basically, the original `Deferred.fromCoroutine` supported only trivial coroutines without `Future` or `Task` inside.

stalkerg:
>  I don't have experience with future and it looks like the automated tests is not executed... so I can't do a proper review.

I can reproduce it locally, working on it.

graingert:
Deferred.fromCoroutine intentionally does not support asyncio coroutines they have different cancellation semantics etc. To wait on a future you should explicitly call `fromFuture(asyncio.create_task)`

graingert:
Deferred.fromCoroutine intentionally does not support asyncio coroutines they have different cancellation semantics etc. To wait on asyncio coroutines you should explicitly call `fromFuture(asyncio.create_task)`

stalkerg:
@graingert 

> Deferred.fromCoroutine intentionally does not support asyncio coroutines

I see, you support async/await but without asyncio. But it's very confusing in docs, https://docs.twisted.org/en/stable/core/howto/defer-intro.html#inline-callbacks-using-yield here we have the phrase:
> Coroutines are supported by dedicated Python syntax, are compatible with asyncio, and provide higher performance.

also, the documentation includes examples only with `fromCoroutine` and has no mention `fromFuture`.
PS I think nowadays async/await in Python is associated with asyncio (or trio).

exarkun:
> PS I think nowadays async/await in Python is associated with asyncio (or trio).

This doesn't mean that only asyncio and trio are allowed to use them, nor that everyone using them is using asyncio or trio.

stalkerg:
> This doesn't mean that only asyncio and trio are allowed to use them, nor that everyone using them is using asyncio or trio.

Sure, but it's just in the context of confusion. Also, because we are using `asyncio` reactor, developers can have expectations of compatibility. 

Maybe we should raise an exception if during follow coroutines we found Future? Because current errors like `await wasn't used with future` doesn't explain anything (and it's come from C code). Also because sometimes not easy to check what exactly the async function uses deep inside. 

but still, my PR is working, I know it seems like will not be correct during the cancel case, but maybe is it possible to handle it?

src/twisted/test/test_defer.py
glyph: > It means we should back control the event loop before we can call send again.

Then the docstring should say those words, rather than "properly" :-) 

See https://jml.io/test-docstrings/

glyph:
> Without these changes not possible to use aiohttp inside Twisted even if you are using asyncio reactor and it reduces compatibility with tons of libraries.

That is not accurate. `Deferred.fromFuture` and `Deferred.asFuture` should do any conversion from/to aiohttp's async primitives. If we can do this automatically and losslessly for the user then by all means we should, but I'd much rather require manual conversion all the time than have some programs randomly fail to run if you selected the wrong reactor.

I haven't evaluated the code here yet though, maybe it's great :-)

stalkerg:
@glyph 
> That is not accurate. Deferred.fromFuture and Deferred.asFuture should do any conversion from/to aiohttp's async primitives. 

`Deferred.fromFuture` is working, but I wrongly expected `Deferred.fromCoroutine` also should work (because coroutine deep inside can use Future or Task)

src/twisted/test/test_defer.py
stalkerg: @glyph thanks! Good page.

glyph:
Ah. Yes, the naming here is unfortunate. Clearly at least the documentation could use an improvement here.

stalkerg:
@exarkun, if I will try to solve `cancel` semantics difference, does this PR have sense for you? If not, I will close it.

exarkun:
> @exarkun, if I will try to solve `cancel` semantics difference, does this PR have sense for you? If not, I will close it.

I'll leave this to someone else to answer, I think.

graingert:
It's also `asyncio.current_task()` that needs to work too

glyph (changes requested):
@stalkerg can you update this branch so that it passes tests and resolve any conflicts?

stalkerg:
@glyph I can if you think this PR makes sense and has value.

glyph:
@stalkerg 

> I can if you think this PR makes sense and has value.

I definitely think it has value. I would love it if we could automatically support futures.

I am not totally sure that this would cover all the edge cases and correctly propagate cancellation semantics and whatnot, but getting full code coverage to ensure that we have tests that cover the whole implementation is like step 0 to even figuring that out :).

I would also like to see a current codspeed result, since I think this may have nasty performance implications.

Issue #11888, opened 2023-06-28
Python 3.12 test failures


stratakis:
Currently in Fedora we are rebuilding twisted with Python 3.12 beta 3 and there are various test failures that we've observed.

First of all test_log hangs on test_startLoggingOverridesWarning.

By removing test_log I was able to reach 5 test failures and 41 test errors. Full build log is attached.

Sample:
```
[FAIL]
Traceback (most recent call last):
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/trial/test/test_warning.py", line 180, in test_warningsConfiguredAsErrors
    self.assertEqual(len(result.errors), 1)
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/trial/_synctest.py", line 441, in assertEqual
    super().assertEqual(first, second, msg)
  File "/usr/lib64/python3.12/unittest/case.py", line 885, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/python3.12/unittest/case.py", line 878, in _baseAssertEqual
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: 5 != 1

[ERROR]
Traceback (most recent call last):
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/test/test_unix.py", line 345, in test_cannotListen
    s = reactor.listenUNIXDatagram(addr, p)
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/internet/posixbase.py", line 285, in listenUNIXDatagram
    p.startListening()
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/internet/udp.py", line 184, in startListening
    self._bindSocket()
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/internet/unix.py", line 516, in _bindSocket
    raise error.CannotListenError(None, self.port, le)
twisted.internet.error.CannotListenError: Couldn't listen on any:/builddir/build/BUILD/twisted-twisted-22.10.0/_trial_temp/twisted.test.test_unix/DatagramUnixSocketTests/test_cannotListen/us0ykj97/temp: AF_UNIX path too long.

[ERROR]
Traceback (most recent call last):
  File "/builddir/build/BUILD/twisted-twisted-22.10.0/build/lib/twisted/test/test_sob.py", line 38, in testStyles
[...]

PR #11910, opened 2023-08-16
#11857 test on py 3.12rc

### LINKED ISSUE #11857: Enable python 3.12 rc on GitHub Actions
GHA has 3.12 alpha 

I think that we can give it a try and prepare Twisted for the future Python version.

@hrnciar has already started some cleanup for deprecated and now removed functions.

I think that more will be removed ... crypto / spwd 

List of modules here https://peps.python.org/pep-0594/#deprecated-modules

For 3.12 I think that we are ok, so I expect this to be easy.

3.13 will be more complicated with the removal of `cgi` and `spwd`

graingert:
## Scope and purpose

Fixes https://github.com/twisted/twisted/issues/11857 and https://github.com/twisted/twisted/issues/11888


src/twisted/internet/defer.py
```diff
@@ -1341,8 +1342,9 @@ def main(reactor):
 
         @raise ValueError: If C{coro} is not a coroutine or generator.
         """
-        # asyncio.iscoroutine identifies generators as coroutines, too.
-        if iscoroutine(coro):
+        # asyncio.iscoroutine <3.12 identifies generators as coroutines, too.
+        # for >=3.12 we need to check isgenerator also
```
graingert, line 1346:
```suggestion
        # for >=3.12 we need to check isgenerator also
        # see https://github.com/python/cpython/issues/102748
```

src/twisted/persisted/aot.py
```diff
@@ -231,13 +231,17 @@ def indentify(s):
     out = []
     stack = []
     l = ["", s]
+
+    def readline():
+        return l.pop().replace("\0", "")
```
graingert:
```suggestion
        # tokenize no longer supports null bytes in source code since 3.12
        # see https://github.com/python/cpython/issues/102856
        return l.pop().replace("\0", "")
```

graingert:
relevant cpython issues:

https://github.com/python/cpython/issues/51574
https://github.com/python/cpython/issues/100425
https://github.com/python/cpython/issues/102856
https://github.com/python/cpython/issues/102748

src/twisted/trial/_synctest.py
```diff
@@ -1328,7 +1328,8 @@ def mktemp(self):
         )
         if not os.path.exists(base):
             os.makedirs(base)
-        dirname = tempfile.mkdtemp("", "", base)
+        # workaround https://github.com/python/cpython/issues/51574
+        dirname = os.path.relpath(tempfile.mkdtemp("", "", base))
```
adiroiban: Maybe add a comment with 3.11

When we will remove support for 3.11 we can do a search in the code and refactor this.

```suggestion
        # With 3.11 or older mkdtemp returns a relative path.
        # With newer it is absolute.
        # Here we make sure we always handle a relative path.
        # See https://github.com/python/cpython/issues/51574
        dirname = os.path.relpath(tempfile.mkdtemp("", "", base))
```

src/twisted/test/test_task.py
```diff
@@ -467,6 +467,13 @@ def test_withCountFloatingPointBoundary(self):
         for x in range(count):
             clock.advance(interval)
 
+        # work around https://github.com/python/cpython/issues/100425 on py312
+        def sum(items):
+            total = 0.0
```
adiroiban: can we have a different name for this method?

And explain what the function does

```suggestion
        def sum_compat(items):
            """
            Make sure the result is more precise.
            On Python 3.11 or older this can be a float with ~ 0.00001
            in precision difference.
            See: https://github.com/python/cpython/issues/100425
            """
            total = 0.0
```

src/twisted/persisted/aot.py
```diff
@@ -231,13 +231,19 @@ def indentify(s):
     out = []
     stack = []
     l = ["", s]
+
+    def readline():
+        # tokenize no longer supports null bytes in source code since 3.12
+        # see https://github.com/python/cpython/issues/102856
```
adiroiban: Maybe the the info as doctring.

I would include 3.11 in the docstrin so that once we remove the support for 3.11 we can refactor this code... or update the comment.

```suggestion
    def pop_line():
        """
        In Python 3.11 or older, tokenize no longer supports null bytes in source code
        # See https://github.com/python/cpython/issues/102856
        """
```

src/twisted/internet/defer.py
```diff
@@ -1341,8 +1342,10 @@ def main(reactor):
...
+        # see https://github.com/python/cpython/issues/102748
+        if iscoroutine(coro) or inspect.isgenerator(coro):
```
adiroiban, line 1348: I don't have experince with coroutine.

I don't understand this comment. Sorry

adiroiban (commented):
Great work. thanks!

Changes looks good. I hope someone can take a look and review the defered coroutine part.
Or maybe an updated comment can help.

adiroiban:
I see that the current tests are with 3.12.0-rc.1 ... but the setup is to test with any latest 3.12 . I think that's fine.

And I think that the scope of this PR is not only to get tests on py3.12 , but also a general update of twisted codebase for 3.12 ... for example the deprecationwarning fix

src/twisted/trial/_synctest.py
graingert: the workaround can't be removed when 3.11 support is removed

src/twisted/persisted/aot.py
graingert: This breaks indentify because the null bytes are used as placeholders for indention levels, this is relying on broken behaviour from py3.11. I've vendored the tokenize py3.11 module to fix it

src/twisted/test/test_task.py
graingert: you still get a float back with 3.12 sum, it's just 1.0 instead of 0.9999999999

src/twisted/persisted/aot.py
```diff
@@ -13,11 +13,11 @@
 import copyreg as copy_reg
 import re
 import types
-from tokenize import generate_tokens as tokenize
 
 from twisted.persisted import crefutil
 from twisted.python import log, reflect
 from twisted.python.compat import _constructMethod
+from ._tokenize import generate_tokens as tokenize
```
graingert, line 23:
```suggestion
# tokenize from py3.11 is vendored to work around https://github.com/python/cpython/issues/105238
# on 3.12 
from ._tokenize import generate_tokens as tokenize
```

src/twisted/test/test_persisted.py
```diff
@@ -446,6 +446,12 @@ def test_circularTuple(self):
         self.assertIs(oj[0][0], oj)
         self.assertEqual(oj[0][1], 4321)
 
+    def testIndentify(self):
+        self.assertEqual(
+            aot.jellyToSource({"hello": {"world": []}}),
+            "app={\n  'hello':{\n    'world':[],\n    },\n  }",
+        )
+
```
adiroiban: Maybe is easier to read the test... but as long as the test has a docstring, it should be good to merge.

```suggestion
    def testIndentify(self):
        """
        The generated serialization is indented.
        """
        self.assertEqual(
            aot.jellyToSource({"hello": {"world": []}}),
            """app={
  'hello':{
      'world':[],
     },
   }""")

```

src/twisted/persisted/_token.py
```diff
@@ -0,0 +1,146 @@
+"""
+Token constants.
+vendored from https://github.com/python/cpython/blob/6b825c1b8a14460641ca6f1647d83005c68199aa/Lib/token.py
+Licence: https://docs.python.org/3/license.html
+"""
```
adiroiban: just a warning

```suggestion
"""
FIXME:https://github.com/twisted/twisted/issues/3843
This can be removed once t.persisted.aot is removed.
New code should not make use of this.

Token constants.
vendored from https://github.com/python/cpython/blob/6b825c1b8a14460641ca6f1647d83005c68199aa/Lib/token.py
Licence: https://docs.python.org/3/license.html
"""
```

src/twisted/persisted/_tokenize.py
```diff
@@ -0,0 +1,901 @@
+"""Tokenization help for Python programs.
+vendored from https://github.com/python/cpython/blob/6b825c1b8a14460641ca6f1647d83005c68199aa/Lib/tokenize.py
```
adiroiban:
```suggestion
"""
FIXME:https://github.com/twisted/twisted/issues/3843
This can be removed once t.persisted.aot is removed.
New code should not make use of this.

Tokenization help for Python programs.
vendored from https://github.com/python/cpython/blob/6b825c1b8a14460641ca6f1647d83005c68199aa/Lib/tokenize.py
```

adiroiban (commented):
I am not happy with vendoring...but if the plan is to remove aot, it should be fine.

src/twisted/test/test_task.py
adiroiban: True. I tried to update my suggested comment.

src/twisted/trial/_synctest.py
adiroiban: True. I have updated  the comment.

adiroiban (approved):
Thanks. I am happy with the changes.

Not sure about the corouting part, as I don't have experince  with that.
but as long as the tests are green, it should be fine.

Many thanks!

PR #11914, opened 2023-08-16
#11913 drop support for python 3.7

### LINKED ISSUE #11913: drop support for eol python 3.7
3.7 is eol we should no longer support it

drop support for 3.7

graingert:
## Scope and purpose

Fixes #11913

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,14 +17,13 @@ dynamic = [
 description = "An asynchronous networking framework written in Python"
 license = "MIT"
 # When updating this value, make sure our CI matrix includes a matching minimum version.
-requires-python = ">=3.7.1"
+requires-python = ">=3.8.0"
 authors = [
     { name = "Twisted Matrix Laboratories", email = "twisted-python@twistedmatrix.com" },
 ]
 classifiers = [
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -106,12 +104,8 @@ http2 = [
     "priority >= 1.1.0, < 2.0",
 ]
 
-contextvars = [
```
adiroiban, line 109: maybe we should still keep this extra dependency ... but leave it empty for compatibility.

Maybe we can add a note about this in the release notes for 2023.8.0

-----------

If we remove this, if someone tries to `pip install twisted[contextvars]`  with this change  it will fail..

But maybe is ok to have these type of breaking changes... I don't know how to best handle this.

src/twisted/web/http.py
```diff
@@ -983,41 +983,29 @@ def requestReceived(self, command, path, version):
...
-                        )
-                    else:
-                        # The parse_multipart function on Python 3
-                        # decodes the header bytes as iso-8859-1 and
-                        # returns a str key -- we want bytes so encode
-                        # it back
-                        self.args.update(
-                            {x.encode("iso-8859-1"): y for x, y in cgiArgs.items()}
-                        )
+                    cgiArgs = cgi.parse_multipart(
```
adiroiban, line 986: thanks for the cleanup... I think that for python 3.13 we need to plan for a replacment for cgi... I think that cgi is deprecated and will be removed.

adiroiban (approved):
Looks good. Thanks for the cleanup.

With the 2023.8.0 release branch created, this can be merged.

src/twisted/newsfragments/11913.misc
```diff
@@ -0,0 +1 @@
+Drop support for Python 3.7.
```
adiroiban: Maybe this file should be `.removal` and maybe it should also mention that the twisted[contextvar] extra was also removed.

Thanks

pyproject.toml
graingert: It's not a breaking change, if you attempt to install an extra that doesn't exist you just get a warning, and the install completes successfully:
```bash
$ pip install twisted[banana]
Collecting twisted[banana]
  Using cached Twisted-22.10.0-py3-none-any.whl (3.1 MB)
WARNING: twisted 22.10.0 does not provide the extra 'banana'
```

src/twisted/web/http.py
graingert: @adiroiban https://github.com/twisted/twisted/pull/11918

pyproject.toml
adiroiban: Thanks for the info. all good.

PR #11917, opened 2023-08-18
#11583 Remove custom code for towncrier check

### LINKED ISSUE #11583: Update towncrier

adiroiban:
## Scope and purpose

Fixes #11583

This removes the custom code that we had to run towncrier check as part of the CI.

Since towncrier 22.8 there is support to run `towncrier check` on a release branch

This also updated to latest towncrier ... we don't any specific bugfix or feature... but just in case.

A build with failure due to missing newsfragment https://github.com/twisted/twisted/actions/runs/5901669253/job/16008257362?pr=11917

A build that pass after newsfragment was added https://github.com/twisted/twisted/actions/runs/5901731330/job/16008426758?pr=11917


graingert (dismissed):

adiroiban:
Thanks for looking into this

The code from here is still broken.. and the first CI is not yet ready.
This is not ready for review or approval

I will do a first CI run to see the checks fails as the PR doesn't have a news fragment,
then I will push a commit with the newsfragment and see that it pass

needs-changes

tox.ini
```diff
@@ -132,7 +133,7 @@ commands =
     # FIXME:https://github.com/twisted/twisted/issues/11583
     # With latest towncrier we should be amble to just use towncrier included
     # `check` tool.
```
graingert:
```suggestion
```

adiroiban: Thanks. Well spotted.
I forgot to check about this.

For my other projects I use the ticket it in the branch name and have an automated check to make sure there are no FIXME comments left in the code reference the same ticket ID.

adiroiban:
I think this is ready for review.

project coverage might fail since we removed code.. but it should be fine.

adiroiban:
needs-review

graingert (approved):

adiroiban:
Many thanks for  the review. Auto-merge is enable for this PR.

PR #11918, opened 2023-08-18
#11919 drop dep on deprecated cgi (removed in 3.13)

### LINKED ISSUE #11919: drop dep on deprecated cgi (removed in 3.13)
cgi is deprecated since python 3.11 and removed in 3.12
cgi.escape got removed and replaced with html.escape a long time ago (the .tac files are in need of some love)

vendor bits of cgi needed

depend on legacy-cgi

graingert:
## Scope and purpose

Fixes #11919


docs/core/howto/tutorial/listings/finger/finger16.tac
```diff
@@ -1,5 +1,5 @@
 # Read from file, announce on the web, irc
-import cgi
+import html
 
 from twisted.application import internet, service, strports
```
graingert:
```suggestion
from twisted.application import internet, service, strports
```

docs/core/howto/tutorial/listings/finger/finger17.tac
```diff
@@ -1,5 +1,5 @@
 # Read from file, announce on the web, irc, xml-rpc
-import cgi
+import html
 
 from twisted.application import internet, service, strports
```
graingert:
```suggestion
from twisted.application import internet, service, strports
```

docs/core/howto/tutorial/listings/finger/finger20.tac
```diff
@@ -1,5 +1,5 @@
 # Do everything properly, and componentize
-import cgi
+import html
 
 from zope.interface import Interface, implementer
```
graingert:
```suggestion
from zope.interface import Interface, implementer
```

docs/core/howto/tutorial/listings/finger/finger21.tac
```diff
@@ -1,5 +1,5 @@
 # Do everything properly, and componentize
-import cgi
+import html
 
 from zope.interface import Interface, implementer
```
graingert:
```suggestion
from zope.interface import Interface, implementer
```

adiroiban:
Many thanks for the PR. I think that vendoring is the only way to keep backward compatibilty.

We can replace parse_multipart with `email.message_from_bytes` and use Message.walk() but we will not keep the FieldStorage API.

I think that it's ok to merge this.

We should find a better API for twisted form parsing...and once we have that new API we can deprecate this.

src/twisted/web/_cgi.py
```diff
@@ -0,0 +1,721 @@
+"""
+cgi.parse_header, cgi.parse_multipart and cgi.FieldStorage backported from
+https://github.com/python/cpython/blob/60edc70a9374f1cc6ecff5974e438d58fec29985/Lib/cgi.py
+Licence: https://docs.python.org/3/license.html
+"""
+import locale
```
graingert, line 8:
```suggestion
from __future__ import annotations

import locale
```

src/twisted/web/_cgi.py
```diff
@@ -0,0 +1,838 @@
...
+        pass
+    fs = FieldStorage(
+        fp,
+        headers=headers,
+        encoding=encoding,
+        errors=errors,
+        environ={"REQUEST_METHOD": "POST"},
+        separator=separator,
+    )
+    return {k: fs.getlist(k) for k in fs}
```
twm, line 837: I think that we should use the version of `cgi.parse_multipart` from before [this commit](https://github.com/python/cpython/commit/cc3fa204d357be5fafc10eb8c2a80fe0bca998f1). That commit regressed on the type of headers (from `bytes` to `str`) and introduced the encoding stuff, which is (a) a questionable design choice at this layer and (b) actively counterproductive for Twisted, since we need `bytes`.

src/twisted/web/test/test_cgi.py
```diff
@@ -87,8 +87,8 @@
 """
 
 URL_PARAMETER_CGI = """\
-import cgi
-fs = cgi.FieldStorage()
+from twisted.web import _cgi
+fs = _cgi.FieldStorage()
```
twm, line 91: If you take my suggestion above I think that we can avoid vendoring `FieldStorage` entirely by replace this with something that echoes `os.environ['QUERY_STRING']` ([RFC](https://www.rfc-editor.org/rfc/rfc3875#section-4.1.7)). `FieldStorage` [appears to have been used purely for convenience](https://github.com/twisted/twisted/commit/e90bfd4c09da8e6168e2b4c711099a8ce927a675), and as the `cgi` module is going away I don't think there's any sense verifying compatibility with it.

`HEADER_OUTPUT_CGI` looks like it'd work for this.

twm (commented):
I am hesitant to vendor so much code without also vendoring its tests. It looks like using the older `parse_multipart` implementation would lower the burden in both ways.

PR #11927, opened 2023-08-19
#11934 Add more types

### LINKED ISSUE #11934: Move a few more modules to the mypy disallow_untyped_defs config section

graingert:
Fixes #11934 disallowing untyped defs in more places


src/twisted/python/threadpool.py
```diff
@@ -10,14 +10,30 @@
 """
 
 from threading import Thread, current_thread
```
graingert, line 14:
```suggestion
from __future__ import annotations

from threading import Thread, current_thread
```

src/twisted/_threads/_pool.py
```diff
@@ -10,13 +10,24 @@
...
 from twisted.python.log import err
+from ._ithreads import IWorker
 from ._team import Team
 from ._threadworker import LockWorker, ThreadWorker
 
 
-def pool(currentLimit, threadFactory=Thread):
+class _ThreadFactory(Protocol):
+    def __call__(self, *, target: Callable[..., object]) -> Thread:
+        pass
```
adiroiban:
```suggestion
class _ThreadFactory(Protocol):
    """
    This is here only to define the interface.
    """
    def __call__(self, *, target: Callable[..., object]) -> Thread:   # pragma: no cover
        ...
```

src/twisted/python/randbytes.py
```diff
@@ -109,11 +109,11 @@ def insecureRandom(self, nbytes):
         @return: a string of random bytes.
         @rtype: C{str}
         """
-        for src in ("_randBits", "_randModule"):
-            try:
-                return getattr(self, src)(nbytes)
-            except SourceNotAvailable:
-                pass
+        try:
```
adiroiban, line 112: I guess this is a drive-by and not related to type checking.

src/twisted/python/threadpool.py
```diff
@@ -9,15 +9,33 @@
...
 from twisted.python.deprecate import deprecated
 from twisted.python.failure import Failure
 from twisted.python.versions import Version
 
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+class _SupportsQsize(Protocol):
+    def qsize(self) -> int:
```
adiroiban, line 30: Can we use the zope.Interface here ?

I don't have experience with Python typing...so maybe this is ok.

```suggestion
class _SupportsQsize(Protocol):
    """
    This is only used to define the interface.
    """
    def qsize(self) -> int:  # pragma: no cover
```

src/twisted/python/threadpool.py
```diff
@@ -9,15 +9,33 @@
...
+        pass
+
+
+class _State(TypedDict):
```
adiroiban, line 34: Is it expected to define types without docstrings?

Is this the convention for python typing?

adiroiban (approved):
I don't have much experience with typing.

I left a few comments inline.

I think  that this is ok and can be merged.

@graingert do you have time to take a look and review https://github.com/twisted/twisted/pull/11823 

Thanks!

src/twisted/python/threadpool.py
glyph: I think there's a soft preference for protocols unless Zope.Interface's features (adaptation, explicit opt-in) are helpful in a certain situation. See https://blog.glyph.im/2021/03/interfaces-and-protocols.html for lots of details

src/twisted/python/threadpool.py
glyph: Types should really have docstrings just like anything else, but sometimes you need to describe a very simple object like "has a `qsize` method" and bloating it up with multiple lines of docstrings does not facilitate any understanding.

I'd prefer if we used `...` in these situations instead of `pass` though, since that's the convention for stubs.

adiroiban: thanks for the info.

src/twisted/python/randbytes.py
graingert: This is required for the method to type check

src/twisted/_threads/_pool.py
graingert:
```suggestion
        ...
```

src/twisted/python/threadpool.py
```diff
@@ -9,15 +9,33 @@
...
 from twisted.python.failure import Failure
 from twisted.python.versions import Version
 
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+
+class _SupportsQsize(Protocol):
+    def qsize(self) -> int:
+        pass
```
graingert:
```suggestion
        ...
```

src/twisted/_threads/_pool.py
graingert: "This is here only to define the interface." is a somewhat useless docstring, `Protocol.__call__` are sufficiently self-documenting

graingert:
replacing `pass` with `...` mostly resolves the coverage I added a test to cover the remaining ThreadPool._queue. Do we want to Deprecate ThreadPool.q and ThreadPool._queue ?

graingert:
woops itchy fingers on that merge, happy to revert if there's anything wrong in the unreviewed test

adiroiban:
> Do we want to Deprecate ThreadPool.q and ThreadPool._queue ?

I think that it would be nice to start the depreciation for this.

PR #11943, opened 2023-09-01
#11942 require type annotations in t.python.filepath and t.python.zippath

### LINKED ISSUE #11942: require type annotations in zippath and filepath
There are two methods that were missed in the filepath/zippath annotations they should be annotated

graingert:
## Scope and purpose

Fixes #11942

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "Automat >= 0.8.0",
     "hyperlink >= 17.1.1",
     "attrs >= 21.3.0",
-    "typing_extensions >= 3.10.0",
+    "typing_extensions >= 4.0.0",
     "twisted-iocpsupport >= 1.0.2, <2; platform_system == 'Windows'",
 ]
 
@@ -446,7 +446,6 @@ module = [
     'twisted.protocols.test.test_basic',
     'twisted.protocols.test.test_tls',
     'twisted.python.failure',
-    'twisted.python.filepath',
     'twisted.python.formmethod',
     'twisted.python.logfile',
     'twisted.python.modules',
@@ -597,7 +596,6 @@ module = [
     'twisted.python.roots',
     'twisted.python.shortcut',
     'twisted.python.syslog',
-    'twisted.python.zippath',
     'twisted.runner.inetdtap',
     'twisted.runner.inetdconf',
 ]
@@ -646,4 +644,6 @@ module = [
     'twisted.python.test.test_htmlizer',
     'twisted.application.test.test_service',
     'twisted.python.randbytes',
+    'twisted.python.zippath',
+    'twisted.python.filepath',
 ]
--- a/src/twisted/python/filepath.py
+++ b/src/twisted/python/filepath.py
@@ -269,7 +269,7 @@ class UnlistableError(OSError):
         self.originalException = originalException
 
 
-def _secureEnoughString(path):
+def _secureEnoughString(path: AnyStr) -> AnyStr:
     """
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     "Automat >= 0.8.0",
     "hyperlink >= 17.1.1",
     "attrs >= 21.3.0",
-    "typing_extensions >= 3.10.0",
+    "typing_extensions >= 4.0.0",
```
graingert, line 40: https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#added-in-version-400 for `typing_extensions.Self`

src/twisted/python/zippath.py
```diff
@@ -273,7 +273,7 @@ class ZipArchive(ZipPath[AnyStr, AnyStr]):
     _zipfileFilename: AnyStr
 
     @property
-    def archive(self):
+    def archive(self) -> Self:  # type: ignore[override]
```
graingert, line 276: the parent type is a mutable `ZipArchive[_ArchiveStr]` hence the ignore override

pyproject.toml
```diff
@@ -646,4 +644,6 @@ module = [
     'twisted.python.test.test_htmlizer',
     'twisted.application.test.test_service',
     'twisted.python.randbytes',
+    'twisted.python.zippath',
+    'twisted.python.filepath',
```
glyph, line 648: hmm, this feels like a place we should be strictly removing stuff from a list, rather than adding to one. do you know if there's a reason it was set up this way?

src/twisted/python/filepath.py
```diff
@@ -269,7 +269,7 @@ def __init__(self, originalException: OSError):
         self.originalException = originalException
 
 
-def _secureEnoughString(path):
+def _secureEnoughString(path: AnyStr) -> AnyStr:
```
glyph, line 272: love to see it

src/twisted/python/zippath.py
glyph: Hmmmmmm. Ignore overrides typically mean a breakage of LSP. Is that happening here? I think the idea is that we want to conform to a more abstract FilePath protocol (which isn't specified because [`IFilePath` can't be generic](https://github.com/Shoobx/mypy-zope/issues/95) and `ZipArchive` and `ZipPath` are actually derivatives of that as-yet-unspecified thing?

If I'm interpreting this correctly can you file a separate issue, and maybe drop a comment on this line

tox.ini
```diff
@@ -78,7 +78,7 @@ deps =
     mindeps: Automat == 0.8.0
     mindeps: hyperlink == 17.1.1
     mindeps: attrs == 21.3.0
-    mindeps: typing_extensions == 3.10.0
+    mindeps: typing_extensions == 4.0.0
```
glyph, line 81: A thought for future work here; `typing_extensions` really ought to be an optional dependency that we don't use on my recent Pythons where it's not required. 3.11 has `typing.Self` for example.

glyph (approved):
address comments as you see fit and land at your leisure

pyproject.toml
graingert: Yeah I think we should make disallow untyped defs the default

tox.ini
graingert: Yeah but then it's horrible if this else that import

src/twisted/python/zippath.py
graingert: I don't know, I made a new issue to investigate

PR #11947, opened 2023-09-01
#11948 disallow_untyped_decorators = true 

### LINKED ISSUE #11948: mypy disallow_untyped_decorators = true
configure disallow_untyped_decorators = true and fix the type errors

graingert:
## Scope and purpose

Fixes #11948


src/twisted/protocols/amp.py
```diff
@@ -291,6 +291,9 @@ def trapZero(result):
 ]
 
 
+_T_Callable = TypeVar("_T_Callable", bound=Callable[..., object])
```
graingert, line 294: some places use `_Tc` we should be consistent

src/twisted/test/test_defer.py
```diff
@@ -3566,7 +3566,7 @@ def test_deprecatedTimeout(self) -> None:
         """
         L{twisted.internet.defer.timeout} is deprecated.
         """
-        deferred: Deferred[None] = Deferred()
+        deferred: Deferred[object] = Deferred()
         defer.timeout(deferred)
```
graingert, line 3570: defer.timeout should probably take a `Deferred[None]` rather than a `Deferred[object]`

src/twisted/trial/_synctest.py
```diff
@@ -1071,7 +1074,9 @@ def run(self, result):
 
         result.stopTest(self)
 
-    def addCleanup(self, f, *args, **kwargs):
+    def addCleanup(
+        self, f: Callable[_P, object], /, *args: _P.args, **kwargs: _P.kwargs
```
graingert: this changes the defintion of addCleanup to match the TestCase.addCleanup - this is a breaking change where we should follow python or add `# type: ignore[override]`

src/twisted/trial/_asynctest.py
```diff
@@ -295,7 +299,9 @@ def _runFixturesAndTest(self, result):
         finally:
             self._undeprecateReactor(reactor)
 
-    def addCleanup(self, f, *args, **kwargs):
+    def addCleanup(
```
graingert: this changes the defintion of addCleanup to match the TestCase.addCleanup - this is a breaking change where we should follow python or add `# type: ignore[override]`

src/twisted/internet/defer.py
```diff
@@ -2129,7 +2129,11 @@ def handleCancel(result: Failure) -> Deferred[_T]:
 
         return status.deferred
 
-    _inlineCallbacks(None, gen, status, _copy_context())
+    # error: Argument 2 to "_inlineCallbacks" has incompatible type
+    # "Generator[Deferred[_T], object, _T] | Coroutine[Deferred[_T], object, _T]";
+    # expected
+    # "Generator[Deferred[_T], object, None] | Coroutine[Deferred[_T], object, None]"
+    _inlineCallbacks(None, gen, status, _copy_context())  # type: ignore[arg-type]
```
graingert: @glyph any idea what the fix should be here?

`def _inlineCallbacks` was marked `@_extraneous` meaning it lost its arg-type making this call valid, fixing `@_extraneous` causes this arg-type error.

src/twisted/internet/base.py
```diff
@@ -345,7 +346,9 @@ def getHostByName(
             timeoutDelay, self._cleanup, name, lookupDeferred
         )
         self._runningQueries[lookupDeferred] = (userDeferred, cancelCall)
-        lookupDeferred.addBoth(self._checkTimeout, name, lookupDeferred)
+        _: Deferred[None] = lookupDeferred.addBoth(
```
adiroiban, line 349: is this  bug in mypy or standard practice for type annotation ?

src/twisted/trial/_asynctest.py
glyph: I can't see what's breaking about this change? It looks like it's just the annotation.

graingert: It makes f a positional only argument

graingert: It changes the behaviour of `self.addCleanup(f=cleanup)`

src/twisted/internet/defer.py
glyph: YieldType (the first type arg) here is wrong. It should be `Deferred[object]`.  Saying that YieldType is `Deferred[_T]` and that matches ReturnType is saying that `async () -> int` using `Deferred`s can *only ever await upon `Deferred[int]`s* which is obviously silly.

I'm pretty sure we're wrong about this in a bunch of places and I probably propagated this error further in my recent type-hints upgrade.

graingert: yep that fixed it for _inlineCallbacks

src/twisted/protocols/amp.py
adiroiban: Just asking.
Can we have these defined inside something like `twisted.python.typing` and import the types from there ?

I hope this can help with getting consistent type helpers.

src/twisted/test/test_defer.py
adiroiban: sorry. I don't understand the comment.

src/twisted/trial/_asynctest.py
adiroiban: For reference. stdlib has `def addCleanup(self, f, /, *args, **kwargs):`

I think that is OK to make it positional only... to get it in sync with stdlib.

I think that this can qualify for an API exception.

if @glyph also agree with the breaking change, I think that we should have enough votes to pass the API exception.

src/twisted/trial/_synctest.py
adiroiban: I prefer to follow Python stdlib API here, as commented above.

adiroiban (commented):
Thanks for the changes.
Leaving only a comment.


I think the changes looks good but it needs a API exception

Here it is defines as "and at least three committers agree that the change is worthwhile"

I agree with such an exception

https://docs.twisted.org/en/stable/development/compatibility-policy.html#procedure-for-exceptions-to-this-policy

src/twisted/internet/base.py
graingert: This is to coax mypy to pick the correct override definition

I think a new pep is coming so we can call `addBoth[None](...)` instead

src/twisted/protocols/amp.py
graingert: Yes you can do that, but it's best practice to define type variables each time you use them, this will be enforced with https://peps.python.org/pep-0695/

adiroiban: Thanks for the info. https://peps.python.org/pep-0695/ looks good :)

we can live with the current code

graingert:
@glyph I'm happy to change addCleanup to `# type: ignore[override]` and remove the breaking change in this PR

adiroiban:
I have created this ticket https://github.com/twisted/twisted/issues/11967

I think that you can use ignore for this PR and  we can look at addCleanup in a separate PR.
Please add the link to the ticket as a comment, so that we will not forget about the old ticket and avoid creating a duplicate ticket.

thanks

adiroiban (approved):
looks good. thanks. Only minor comments.

src/twisted/_threads/_ithreads.py
```diff
@@ -23,7 +25,7 @@ class IWorker(Interface):
     All methods on this interface must be thread-safe.
     """
 
-    def do(task):
+    def do(task: Callable[..., object]) -> None:
```
graingert: This should be a Callable[[], None]

src/twisted/trial/_synctest.py
graingert: I now have these methods marked as `type: ignore[override]`

src/twisted/test/test_defer.py
graingert: It's deprecated anyway so it doesn't really matter but `defer.timeout` calls `deferred.callback(None)` not `deferred.callback(object())`

adiroiban (approved):
Looks good. Thanks!

PR #11950, opened 2023-09-03
#11949 trial: Pass parallel trial worker index as env variable

### LINKED ISSUE #11949: Parallel Twisted trial workers should know an index to differentiate between each other
Currently tests run under parallel `trial` do not know about each other. This makes it difficult to coordinate access to shared resources (e.g. database) that take a long time to construct. 

For example, consider a case of 10000 tests that access a database. Currently tests can only be run in sequence, each using the same database and cleaning up after itself. Running tests in parallel would involve constructing 10000 separate databases which is infeasible. However, if each test knew the index of the parallel Twisted trial worker it runs under, then only small number of databases would need to be created and tests would use them like in serial case. E.g. when running tests using `trial -j16`, only 16 databases would need to be created.

Expose the index of the parallel trial worker as an environment variable e.g. `TWISTED_TRIAL_PARALLEL_INDEX`.

An alternative is to inspect `os.getcwd()` by the tests. It contains the index of the parallel trial worker as the last path component. E.g. `_trial_temp/21`. However, this is an implementation detail of trial which is problematic to rely on. Twisted should provide a proper way to detect this.

p12tic:
## Scope and purpose

Fixes #11949

Currently tests run under parallel trial do not know about each other. This makes it difficult to coordinate access to shared resources (e.g. database) that take a long time to construct.

For example, consider a case of 10000 tests that access a database. Currently tests can only be run in sequence, each using the same database and cleaning up after itself. Running tests in parallel would involve constructing 10000 separate databases which is infeasible. However, if each test knew the index of the parallel Twisted trial worker it runs under, then only small number of databases would need to be created and tests would use them like in serial case. E.g. when running tests using trial -j16, only 16 databases would need to be created.

This is done by exposing environment variable
TWISTED_TRIAL_PARALLEL_INDEX to the tests run under the parallel runner.


p12tic:
please review

src/twisted/trial/_dist/test/test_disttrial.py
```diff
@@ -263,7 +263,9 @@ def fakeSpawnProcess(
         # The child process runs with PYTHONPATH set to exactly the parent's
         # import search path so that the child has a good chance of finding
         # the same source files the parent would have found.
-        self.assertEqual(os.pathsep.join(sys.path), environment["PYTHONPATH"])
+        self.assertEqual(os.pathsep.join(sys.path), environments[0]["PYTHONPATH"])
+        parallel_indexes = set(e["TWISTED_TRIAL_PARALLEL_INDEX"] for e in environments)
+        self.assertEqual(set(["0", "1", "2", "3"]), parallel_indexes)
```
adiroiban, line 268: This is just a small comment 

parrallel_indexes vs parallelIndexes for pedantic code standard

and also check that PYTHONPATH is present in all environments

```suggestion
        pythonpathEnvs = [e["PYTHONPATH" for e in environments]
        self.assertEqual([os.pathsep.join(sys.path)] * 4, pythonpathEnvs)
        parallelIndexes = set(e["TWISTED_TRIAL_PARALLEL_INDEX"] for e in environments)
        self.assertEqual(set(["0", "1", "2", "3"]), parallelIndexes)
```

docs/core/howto/testing.rst
```diff
@@ -246,3 +246,7 @@ then deletes that schema in the tearDown function, your tests will behave in an
 unpredictable fashion as they tromp upon each other if they have their own
 schema.  And this won't actually indicate a real error in your code, merely a
 testing-specific race-condition.
+
+Trial provides `TWISTED_TRIAL_PARALLEL_INDEX` environment variable to the tests when run in parallel.
```
adiroiban, line 250: I think that this documentation is good enough.

Just a curiosity and for future reference,

Do you know how other  testing frameworks are dealing with this issue?

Do they also set an env variable?
Is there some convention for the variable name?

Are there any other things that can help with shared resources?

adiroiban (dismissed):
Many thanks for the  changes.

They look good. Great PR.

I left only a few minor comments.

adiroiban:
I have approved the changes and the PR looks good to me.

I see in #11949 that Glyph and Jean Paul are not happy with the way this is implemented.

I will leave this unmerged waiting for a consensus on how to get this done.

Regards

glyph (changes requested):
As discussed on #11949 , I don't think that this is a good API surface for us to support, and I think the documentation is insufficient.  I'd like to see more explanation of whether this is actually necessary or not yet.

adiroiban:
@glyph 

> There was already some discussion on the ticket I don't see addressed in the review.

Sorry about that.
The first notification was for the PR ... and I did the review without checking the ticket.
My bad. I will check the ticket discussion before doing a review.

thanks for the follow up here.

glyph:
> The first notification was for the PR ... and I did the review without checking the ticket. My bad. I will check the ticket discussion before doing a review.

No worries. We *should* discuss stuff on tickets more, but so many are just perfunctory CI / types / test housekeeping these days that it is an understandable habit to fall in to :)

PR #11956, opened 2023-09-04
#11955 fix warn_return_any

### LINKED ISSUE #11955: mypy warn_return_any = True
configure warn_return_any  = True and fix the issues

graingert:
## Scope and purpose

Fixes #11955


src/twisted/names/dns.py
```diff
@@ -378,7 +379,36 @@ def _isSubdomainOf(descendantName, ancestorName):
...
+        raise ValueError("Invalid time interval specifier: " + s)
+
+
+@overload
+def str2time(s: Union[str, bytes, int]) -> int:
+    ...
+
+
+@overload
+def str2time(s: None) -> None:
```
graingert, line 410: str2time can technically pass through anything but int and None are the only usefull ttl passthrough

src/twisted/names/dns.py
```diff
@@ -395,26 +425,12 @@ def str2time(s):
...
-    suffixes = (
-        ("S", 1),
-        ("M", 60),
-        ("H", 60 * 60),
-        ("D", 60 * 60 * 24),
-        ("W", 60 * 60 * 24 * 7),
-        ("Y", 60 * 60 * 24 * 365),
-    )
     if isinstance(s, bytes):
-        s = s.decode("ascii")
```
graingert, line 407: mypy doesn't like type-punning this s: str | bytes | int | None into a str so I used a new function

src/twisted/conch/ssh/keys.py
```diff
@@ -1093,59 +1094,59 @@ def size(self):
...
 
-    def data(self):
+    def data(self) -> dict:
         """
         Return the values of the public key as a dictionary.
 
         @rtype: L{dict}
         """
         if isinstance(self._keyObject, rsa.RSAPublicKey):
-            numbers = self._keyObject.public_numbers()
```
graingert, line 1103: mypy is really annoying here demanding a new numbers variable for every different type of public/private numbers

src/twisted/words/protocols/jabber/jid.py
```diff
@@ -27,7 +27,7 @@ class InvalidFormat(Exception):
     """
 
 
-def parse(jidstring):
+def parse(jidstring: str) -> Tuple[Union[str | None], str, Union[str, None]]:
```
graingert:
```suggestion
def parse(jidstring: str) -> Tuple[Union[str, None], str, Union[str, None]]:
```

src/twisted/names/dns.py
```diff
@@ -378,7 +379,36 @@ def _isSubdomainOf(descendantName, ancestorName):
     return descendantLabels[-len(ancestorLabels) :] == ancestorLabels
 
 
-def str2time(s):
+def _str2time(s: str) -> int:
```
adiroiban, line 382: Maybe add a comment with the reason why we need  this separate/helper method.

I see you left a comment in the PR... but I think that any comment that you fell is needed to be added to the PR, is best to have it as a code comment.

src/twisted/words/protocols/jabber/jid.py
```diff
@@ -74,7 +74,9 @@ def parse(jidstring):
     return prep(user, host, resource)
 
 
-def prep(user, host, resource):
+def prep(
+    user: Union[str, None], host: str, resource: Union[str, None]
+) -> Tuple[Union[str | None], str, Union[str, None]]:
```
graingert:
```suggestion
) -> Tuple[Union[str, None], str, Union[str, None]]:
```

src/twisted/words/protocols/jabber/jid.py
```diff
@@ -141,16 +143,19 @@ class JID:
...
+        self,
+        str: Union[str, None] = None,
+        tuple: Union[Tuple[str, str, str], None] = None,
+    ):
         if str:
             user, host, res = parse(str)
-        else:
+        elif tuple:
             user, host, res = prep(*tuple)
+        else:
```
adiroiban, line 155: I think that this  change is ok.

Is Twisted jabber still used by people in PROD ?

adiroiban (dismissed):
Looks good. Many thanks for the cleanup.

Only small comments.

Thanks again

src/twisted/names/dns.py
graingert, line 382:
```suggestion
# helper function for str2time, needed to appease mypy see GH-11956 
def _str2time(s: str) -> int:
```

src/twisted/words/protocols/jabber/jid.py
graingert: change was required to appease mypy for the `*tuple` otherwise it was optional in the `if not str:` branch

src/twisted/names/test/test_dns.py
```diff
@@ -163,6 +162,12 @@ def test_invalidPrefix(self):
         """
         self.assertRaises(ValueError, dns.str2time, "fooS")
 
+    def test_invalidSuffix(self) -> None:
+        """
+        If an invalid suffix is given, L{dns.str2time} raises L{ValueError}.
+        """
+        self.assertRaises(ValueError, dns.str2time, "Q")
```
graingert:
```suggestion
        self.assertRaises(ValueError, dns.str2time, "1Q")
```

src/twisted/internet/iocpreactor/tcp.py
```diff
@@ -359,7 +359,7 @@ def __init__(self, sock, protocol, clientAddr, serverAddr, sessionno, reactor):
         self.sessionno = sessionno
         logPrefix = self._getLogPrefix(self.protocol)
         self.logstr = f"{logPrefix},{sessionno},{self.clientAddr.host}"
-        self.repstr = "<{} #{} on {}>".format(
+        self.repstr: str = "<{} #{} on {}>".format(
```
graingert, line 362: worrying that this wasn't hit by tests :(

graingert: oh it was hit just took a while for the coverage check to integrate the 3.11 iocp windows job

src/twisted/conch/ssh/keys.py
adiroiban: I can understand why it requires separate names.

Does it still contain, even when reqriting  to code not to use `elif`... like ?

```
if condition:
  return

if other-condition:
  return
```

src/twisted/names/dns.py
```diff
@@ -378,7 +379,37 @@ def _isSubdomainOf(descendantName, ancestorName):
     return descendantLabels[-len(ancestorLabels) :] == ancestorLabels
 
 
-def str2time(s):
+# helper function for str2time, needed to appease mypy see GH-11956
+def _str2time(s: str) -> int:
```
adiroiban: Thanks for the comment.
It still don't make sense ... according to my poor brain.

Instead of `gh-1245` I think is best add the full url.
I wasn't sure if this a mypy bug reference of twisted bug reference.

maybe have something  like this.

```suggestion
def _str2time(s: str) -> int:
    """
    mypy doesn't like type-punning str | bytes | int | None into a str so we have this helper function.
    """
```

src/twisted/conch/ssh/keys.py
graingert: yep I tried without elif first

adiroiban (approved):
Thanks. It looks good.
I am not a mypy / typing  expert but I think we can merge this.

the problem is that we currently don't have too many people available for review.

@glyph if you have time, you can do a post review of this and we can update anything that is needed.

thanks

graingert:
this introduces a bug fixed in [`a9a6345` (#11958)](https://github.com/twisted/twisted/pull/11958/commits/a9a634589aaadff285098d5592b4ffadde8da50e)

PR #11958, opened 2023-09-04
#11957 expand mypy star overrides

### LINKED ISSUE #11957: expand mypy .* module overrides
we'd like to be able to delete a module from the pyproject.toml to mark it as fully type annotated, however having .* overrides with weaker type hinting prevents this

expand mypy .* module overrides

graingert:
## Scope and purpose

Fixes #11957


pyproject.toml
```diff
@@ -599,7 +889,60 @@ module = [
...
+    'twisted.words.iwords',
+    'twisted.words.protocols.jabber.ijabber',
+    'twisted.words.protocols.jabber.jid',
+    'twisted.words.protocols.jabber.sasl_mechanisms',
+    'twisted.words.protocols.jabber.xmpp_stringprep',
+    'twisted.words.tap',
+    'twisted.words.test.test_basechat',
+    'twisted.words.test.test_ircsupport',
+    'twisted.words.test.test_jabbererror',
+    'twisted.words.test.test_jabberjid',
```
graingert, line 939: this makes test_jabberjid a check_untyped_defs test suite! Previously the whole of twisted.words.* was ignored!

this means that the types in jabber.jid are now tested against the test suite - which were broken!

graingert:
I used these snippets to help generate the config: https://gist.github.com/graingert/a2dbd999c9eadd0d6c7ec3a71a2a9d9f

adiroiban (approved):
Many thanks for the update.

I am not an experting in Python typing best practices, but this looks good.

I guess that once this is merged, we will no longer need to run the helper scripts or update the list of  exceptions.

If a new module is added to Twisted, it should be fully `typing` friendly from the start and should not require any exceptions.

While existing modules are updated they can be removed from these lists.

And with these lists, it should also be easier to accept smaller PR.

You can work on updating  the typing for a single module ... or for a single "exception" class.

I think this can be merged. Thanks again.

PR #11960, opened 2023-09-05
#11959 apply autotyping and commit test modules that pass mypy

### LINKED ISSUE #11959: apply autotyping to tests and commit those that pass mypy
most test modules consist of hundreds of `-> None:` test functions that could be automatically typed with https://github.com/JelleZijlstra/autotyping/ in --safe mode

apply https://github.com/JelleZijlstra/autotyping/ and commit any test modules that pass mypy

graingert:
## Scope and purpose

Fixes #11959


src/twisted/web/test/test_flatten.py
```diff
@@ -632,6 +632,7 @@ def test_traceback(self) -> None:
         If a L{FlattenerError} is created with traceback frames, they are
         included in the string representation of the exception.
         """
+
```
adiroiban, line 635: We can leave this change.
Was this change also generated by auto-typing ?

Maybe a bug in auto-typing and upstream might want to know about it.

src/twisted/web/test/test_template.py
```diff
@@ -217,7 +217,6 @@ def test_file(self) -> None:
 
 
 class XMLLoaderTestsMixin:
-
```
adiroiban, line 220: Was this change also generated by auto-typing ?

Maybe a bug in auto-typing and upstream might want to know about it.

adiroiban (approved):
Looks good. Many thanks.

The missing coverage is ok.

Only a small comment.

Feel free to merge when you have time. Thanks again.

src/twisted/web/test/test_flatten.py
graingert: autotyping runs via libcst which automatically applies black, this change is from that

src/twisted/web/test/test_template.py
graingert: autotyping runs via libcst which automatically applies black, this change is from that

PR #11962, opened 2023-09-05
#11961 run pipx run pre-commit autoupdate

### LINKED ISSUE #11961: run pre-commit autoupdate then apply pre-commit run --all-files
black is very outdated causing problems when contributors have the latest black in their IDE

run pre-commit autoupdate then apply pre-commit run --all-files

graingert:
## Scope and purpose

Fixes #11961


setup.cfg
```diff
@@ -6,6 +6,11 @@ extend-ignore =
    E203,
 
 per-file-ignores =
+    src/twisted/internet/defer.py:
+        # redefinition of unused 'DeferredList' from line 1437
```
adiroiban: Wouldn't a `  # noqa: F811` comment be best here?

Here we ignore all errors inside the file.
With `  # noqa` we should be able to ignore only that line.

adiroiban: I do see that we have `disable-noqa = True` in config :)

adiroiban (approved):
Changes looks good. Many thanks.

Only one major request.

Can you please use enable `noqa` in flake8 and then use `  #noqa F811` to ignore a single line.

Maybe pyflakes already supports `@overload` and we no longer need `if TYPE_CHECKING` ... but this can be done in a separate PR

I think that is best to reduce the manual changes.

Approving it but please check my comments before merge.

Many thanks again

glyph:
This seems likely to cause massive conflicts across all open PRs. Do we have a merge driver that is Black-aware, or a plan to update all ~100-ish open PRs so that contributors don't get discouraged? We already have enough review latency as it is.

(Not saying that this shouldn't land, the thing with IDEs is definitely annoying, and we will need a newer Black for newer Pythons at some point regardless, just that we should be careful about blowing up hundreds of in-progress PRs for a minor code-format update.)

adiroiban:
Thanks @glyph for the input.

I don't think there is a `black merge driver` https://github.com/psf/black/issues/896

I guess that since we chose the `black` path we are ready to put the extra effort required to fix all the possible merge conflicts, in exchange for not having worry about style or add style related comments for each review.

As @glyph is not -1 on this PR I think that is ok to land it and deal with merge conflicts later.

glyph:
> As @glyph is not -1 on this PR I think that is ok to land it and deal with merge conflicts later.

As long as someone involved in the merging actually does the work to deal with the merge conflicts, and doesn't shove it off to new contributors who will immediately bounce off :)

adiroiban:
There are already many changes done related to code annotation, I guess they can also generate merge conflicts in existing PRs.

Merge conflicts can happen at any time.

I think that if we decided to use black, we should go all in, by trying to use the latest version

glyph:
> Merge conflicts can happen at any time.

Yes, and usually the way we manage the load is that small numbers of conflicts arise and therefore remediating them in one or two PRs is no big deal.

If we are going to provoke thousands at once I think we should be mindful.

adiroiban:
> If we are going to provoke thousands at once I think we should be mindful.

I think that I am missing some information here, as I am not very familiar with managing a source code updated by `black`

I don't see how the changes from this PR, are any different for example for the changes that add typing to function definitions.


Also, if we decided to use `black` does it automatically imply that we must pin `black` version to the version used when the initial migration was done and never upgrade ?

glyph:
> I don't see how the changes from this PR, are any different for example for the changes that add typing to function definitions.

If we had a PR that added type annotations to >150 files in a single commit, I'd have similar concerns.

Which does bring to mind one potential solution that is less disruptive: one way to mitigate the issue is to commit each file individually, so that the scope of the conflict is less overwhelming and the tools break down less egregiously.

adiroiban:
Sound like a good plan :)

I will try to get the non-black changes merged in a separate PR, and then we can have the black changes merged with one commit per file.

adiroiban:
I got a branch with this script... for each file, pre-commit is executed...so it takes some time ... so it's very slow.

But it should do the job.

I will push the changes in a separate branch.

```
#!/bin/bash
# This is a helper for creating a separate commit for every file that was
# automatically updated by black.
#
# This is designed to be used after black was updated and it needs to
# change many files at once.
#
# To use this script:
# * Update the black version in .pre-commit-config.yaml
# * Commit the .pre-commit-config.yaml change
# * Run `pre-commit run --all`
# * Run this script.
#
for file in $(git diff --name-only)
do
    git commit -m "Black auto-update: $file." $file
done

```

adiroiban (approved):
Based on the feedback from @glyph we should have separate commits for black.

I now think that is best to have this fixed by this PR https://github.com/twisted/twisted/pull/11984

adiroiban (changes requested):
Should be done in https://github.com/twisted/twisted/pull/11984

adiroiban (approved):
Over gitter/matrix, Thomas commented that he doesn't think that the version in which each file is in a separate commit is better that a single commit.

I am approving this again.

For me, any option is ok.

Feel free to merge this version or the PR with separate commits.

Thanks

glyph:
Yep, superseded by https://github.com/twisted/twisted/pull/11984 .

The 'files changed' tab is empty as we would expect now that we've updated this branch.

PR #11963, opened 2023-09-08
#11964 Type annotate more tests

### LINKED ISSUE #11964: type annotate more tests
tests are pretty easy to type annotate just lots of manual `-> None:`

type annotate some more tests

graingert:
## Scope and purpose

Fixes #11964


src/twisted/web/test/_util.py
```diff
@@ -13,10 +13,13 @@
...
 from twisted.web._flatten import flattenString
 from twisted.web.error import FlattenerError
+from twisted.web.http import Request
+from twisted.web.resource import IResource
 from twisted.web.template import Flattenable
+from .requesthelper import DummyRequest
 
 
-def _render(resource, request):
+def _render(resource: IResource, request: Request | DummyRequest) -> Deferred[None]:
```
graingert, line 22: should there be an Interface that both http.Request and DummyRequest implement? Currently there's no interface with `notifyFinish()`

src/twisted/words/test/test_tap.py
```diff
@@ -24,37 +28,37 @@ def setUp(self):
...
         self._loginTest(opt)
 
-    def test_auth(self):
+    def test_auth(self) -> None:
         """
         Tests that the --auth command generates a checker.
         """
         opt = tap.Options()
         opt.parseOptions(["--auth", "file:" + self.file.name])
         self._loginTest(opt)
```
graingert, line 59: should this be?
```suggestion
        return self._loginTest(opt)
```

how is the assertion error in self._loginTest collected if the Deferred isn't returned? I verified that the AssertionError is collected if I change the callback to d.addCallback to raise

src/twisted/trial/test/test_doctest.py
```diff
@@ -26,15 +26,15 @@ def test_id(self):
...
+    def test_basicTrialIntegration(self) -> None:
         """
         L{loadDoctests} loads all of the doctests in the given module.
         """
         loader = runner.TestLoader()
         suite = loader.loadDoctests(mockdoctest)
         self.assertEqual(7, suite.countTestCases())
 
-    def _testRun(self, suite):
+    def _testRun(self, suite: pyunit.TestSuite) -> None:
```
adiroiban, line 37: just a drive-by comment.

I think that pyunit.TestSuite is ok

I am not sure we still need `twisted.trial.unittest.TestSuite`

src/twisted/web/test/_util.py
adiroiban: the HTTP interfaces are not the best .

I think the current annotation is ok. thanks

What I did, I have created my own DummyRequest that  inherits from http.Request and does the minimum overrwriting possible.

src/twisted/words/test/test_tap.py
adiroiban: It  should return a deferred... or inlineCallbacks :)

If you have time, please update these tests. thanks

adiroiban (approved):
Looks good. Many thanks.

I think we can have this merged, and then we can have separate PR to look at the "controversial" code.

Can we have macOS tests with 3.8 enabled in this branch ?

I left a few minor comment inline.

Thanks again.

src/twisted/trial/test/test_doctest.py
graingert: I checked at runtime and it's a regular pyunit.UnitTest

graingert:
> Can we have macOS tests with 3.8 enabled in this branch ?

I think that still makes sense in a separate PR

graingert:
@adiroiban 

> "controversial" code.

What do you mean by that?

adiroiban:
> > "controversial" code.
> What do you mean by that?

Changes like `request: Request | DummyRequest` or returning a deferred in the test_tap.py

graingert:
Ah I see

PR #11969, opened 2023-09-08
[#11968] Deprecate twisted.words.

### LINKED ISSUE #11968: Deprecate twisted.words
twisted.words is not really maintained.

I am not sure if the jabber or IRC implementation is production-ready.

I think continuing to keep the code just adds unnecessarily extra work to maintainers ... for example when adding annotations

If that code is still useful, I think we can move it into something like twisted/txirc and twisted/txjabber 

I am +1 for the deprecation and I think I am not the  only one :)

adiroiban:
## Scope and purpose

Fixes #11968

This introduces the deprecation warning for twisted.words

It also updates the documentation to covers how to deprecate a whole module

https://twisted--11969.org.readthedocs.build/en/11969/development/compatibility-policy.html#modules

### DIFF
--- a/docs/development/compatibility-policy.rst
+++ b/docs/development/compatibility-policy.rst
@@ -457,6 +457,22 @@ but is not raised from `from twisted.old_code.sub.module import SomeClass`.
 
 This is why it's recommended to just use `warnings.warn()` at the top level of the module.
 
+.. code-block:: python
+
+    """
+    The normal docstring of the top-level package.
+
+    This package is now deprecated.
+    """
+    import warnings
+
+    from incremental import Version, getVersionString
+
+    warningString = "twisted.old_code was deprecated at {}".format(
+        getVersionString(Version("Twisted", "NEXT", 0, 0))
+    )
+    warnings.warn(warningString, DeprecationWarning, stacklevel=3)
+
 
 Testing Deprecation Code
 ------------------------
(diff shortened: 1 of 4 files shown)


docs/development/compatibility-policy.rst
```diff
@@ -573,3 +594,30 @@ Tests which need to use deprecated classes should use the :py:meth:`getDeprecate
...
+
+.. code-block:: python
+
+    from importlib import reload
+
+    import twisted.words
+    from twisted.trial import unittest
+
+
+    class WordsDeprecationTests(unittest.TestCase):
```
glyph: this should probably be OldCodeDeprecationTests

glyph (changes requested):
Let's split the uncontroversial documentation upgrades into a separate branch, and give it at least a week on the mailing list before landing an actual deprecation.

docs/development/compatibility-policy.rst
```diff
@@ -573,3 +594,30 @@ Tests which need to use deprecated classes should use the :py:meth:`getDeprecate
...
+        """
+        Ensures that importing twisted.words directly or as a
+        module of twisted raises a deprecation warning.
+        """
+
+        def test_deprecationDirect(self):
```
adiroiban: With typing

```suggestion
        def test_deprecationDirect(self) -> None:
```

adiroiban:
Thanks for the review and the feedback.

Agree with the move. Docs changes were moved here https://github.com/twisted/twisted/pull/11971

I know that we are still talking about whether or not to deprecate twisted.words

I think this is ready for review... in case we decide to deprecate the code.

needs-review

adiroiban:
we can keep twisted.words for now. @ralphm will try to keep an eye over this code.

Issue #7943, opened 2015-06-17
Remove usage of microdom


twisted-trac:
|[<img alt="adiroiban's avatar" src="https://avatars.githubusercontent.com/u/204609?s=50" width="50" height="50">](https://github.com/adiroiban)| @adiroiban reported|
|-|-|
|Trac ID|trac#7943|
|Type|enhancement|
|Created|2015-06-17 09:57:51Z|
|Branch|https://github.com/twisted/twisted/tree/remove-microdom-usage-7943|

In order to deprecate microdom as requested in [#3561](https://github.com/twisted/twisted/issues/3561) we first need to remove all usage of microdom from Twisted code.

<details><summary>Searchable metadata</summary>

```
trac-id__7943 7943
type__enhancement enhancement
reporter__adiroiban adiroiban
priority__normal normal
milestone__ 
branch__branches_remove_microdom_usage_7943 branches/remove-microdom-usage-7943
branch_author__adiroiban adiroiban
status__closed closed
resolution__fixed fixed
component__core core
keywords__None None
time__1434535071015376 1434535071015376
changetime__1452337674102803 1452337674102803
version__None None
owner__hawkowl hawkowl

```
</details>

twisted-trac:
|[<img alt="glyph's avatar" src="https://avatars.githubusercontent.com/u/716529?s=50" width="50" height="50">](https://github.com/glyph)<a name="note_1"></a>|@glyph commented|
|-|-|

Luckily there's not a whole lot at this point.

  - domhelpers, which should be deprecated right along with `microdom`,
  - some unused imports (!?!?!) in the finger tutorial,
  - a disused example, `lj.rpy.py`,
  - one use, which could probably just be replaced with an "is this file empty" check, that is just making sure Sphinx produced some HTML, in `test_release`.

I think that's it.

twisted-trac:
|[<img alt="adiroiban's avatar" src="https://avatars.githubusercontent.com/u/204609?s=50" width="50" height="50">](https://github.com/adiroiban)<a name="note_2"></a>|@adiroiban commented|
|-|-|

(In [46567]) Branching to remove-microdom-usage-7943.

twisted-trac:
|[<img alt="adiroiban's avatar" src="https://avatars.githubusercontent.com/u/204609?s=50" width="50" height="50">](https://github.com/adiroiban)<a name="note_3"></a>|@adiroiban commented|
|-|-|

Thanks glyph for the checklist :)

[...]

PR #11970, opened 2023-09-08
Deprecate twisted.web.microdom and twisted.web.domhelpers

### LINKED ISSUE #3561: deprecate microdom
|-|-|
|Trac ID|trac#3561|
|Type|task|
|Created|2008-12-01 17:44:45Z|

In the olden days, there were lots of XML parsers, and they were bad.  Now there's even more XML parsers, and they're not _so_ bad.  But microdom is still bad.  Now it's probably worse than most other XML libraries.  Plus, we're not going to fix the obvious bugs in it (eg [#8049](https://github.com/twisted/twisted/issues/8049)).  People shouldn't use it, so it should be deprecated.

`twisted.web.html` uses microdom, so it should also be deprecated probably (it's not an intensely useful module anyway).  Anything else in Twisted which uses microdom (hopefully, not many things) should stop or also be deprecated (some things in domhelpers use it, for example).

```
trac-id__3561 3561
type__task task
reporter__exarkun exarkun
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__new new
resolution__ 
component__web web
keywords__ 
time__1228153485000000 1228153485000000
changetime__1434535103308074 1434535103308074
version__None None
owner__ 
cc__glyph
```
</details>

adiroiban:
## Scope and purpose

Fixes #3561

The usage of these modules in Twisted was removed in https://github.com/twisted/twisted/issues/7943 - in 2016

I think it's time to raise deprecation warnings ... and at some point remove this code.

### DIFF
--- a/src/twisted/web/domhelpers.py
+++ b/src/twisted/web/domhelpers.py
@@ -4,12 +4,22 @@
 
 """
 A library for performing interesting tasks with DOM objects.
-"""
 
+This module is now deprecated.
+"""
+import warnings
 from io import StringIO
+
+from incremental import Version, getVersionString
 
 from twisted.web import microdom
 from twisted.web.microdom import escape, getElementsByTagName, unescape
+
+warningString = "twisted.web.domhelpers was deprecated at {}".format(
+    getVersionString(Version("Twisted", "NEXT", 0, 0))
+)
+warnings.warn(warningString, DeprecationWarning, stacklevel=3)
+
 
 # These modules are imported here as a shortcut.
 escape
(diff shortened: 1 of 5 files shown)


adiroiban:
needs-review

glyph (approved):
This looks great. Although I would love it if we had better CI coverage to fail on warnings so that we don't have to trawl through logs to make sure that we have properly removed all of these.  

Anyway, `domhelpers` was never our proudest moment :).  Glad to see it on its way out.

adiroiban:
Thanks for the review.
I agreee that it would be nice to have better support for deprecation warnings in `trial`

PR #11974, opened 2023-09-09
#11973 test on py3.8 macos

### LINKED ISSUE #11973: test on py3.8 macos
twisted.internet.cfreactor regressed (https://github.com/twisted/twisted/issues/11965) because of missing test coverage for macos and older python versions

 we should test on macos py3.8

**Additional context**

graingert:
## Scope and purpose

Fixes #11973

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -134,6 +134,12 @@ jobs:
           # Just Python 3.12 with default settings.
           - python-version: '3.12'
 
+          # Newest macOS and oldest Python (major) supported versions.
+          - python-version: '3.8'
+            runs-on: 'macos-12'
+            job-name: 'macos-12-default-tests'
+            tox-env: 'macos-withcov-alldeps'
+
           # Newest macOS and newest Python supported versions.
           - python-version: '3.11'
             runs-on: 'macos-12'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -135,7 +135,7 @@ jobs:
           - python-version: '3.12'
 
           # Newest macOS and oldest Python (major) supported versions.
-          - python-version: '3.8.0'
+          - python-version: '3.8.10'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
             tox-env: 'macos-withcov-alldeps'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -135,7 +135,7 @@ jobs:
           - python-version: '3.12'
 
           # Newest macOS and oldest Python (major) supported versions.
-          - python-version: '3.8.10'
+          - python-version: '3.8'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
             tox-env: 'macos-withcov-alldeps'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -135,7 +135,7 @@ jobs:
           - python-version: '3.12'
 
           # Newest macOS and oldest Python (major) supported versions.
-          - python-version: '3.8'
+          - python-version: '3.8.0'
             runs-on: 'macos-12'
             job-name: 'macos-12-default-tests'
             tox-env: 'macos-withcov-alldeps'
--- a/src/twisted/conch/test/test_scripts.py
+++ b/src/twisted/conch/test/test_scripts.py
@@ -30,7 +30,10 @@ except ImportError:
 else:
     try:
         tkinter.Tk().destroy()
-    except tkinter.TclError as e:
+    except (tkinter.TclError, RuntimeError) as e:
+        # On GitHub Action the macOS Python might not support the version of TK
+        # provided by the OS and it will raise a RuntimeError
+        # See: https://github.com/actions/setup-python/issues/649
         doSkip = True
         skipReason = "Can't test Tkinter: " + str(e)
 
--- a/src/twisted/newsfragments/11973.misc
+++ b/src/twisted/newsfragments/11973.misc
@@ -1,0 +1,1 @@
+test on macos python 3.8


graingert:
Looks like we're hit by https://github.com/actions/setup-python/issues/402

graingert:
Which is supposed to be fixed

adiroiban:
Can we go with 3.9 ? I think for `__future__ anotations` 3.9 is still valid, as annotations are only by default in 3.10.
I might be wrong.

I remember seeing this bug on 3.7 and I think that this is why we didn't had macos on 3.7 in the first place.

graingert:
3.9 makes dict[x, y] work and 3.10 makes T | None work so we do need 3.8

adiroiban:
I think that latest version is https://github.com/actions/setup-python/issues/649

.github/workflows/test.yaml
```diff
@@ -134,6 +134,12 @@ jobs:
           # Just Python 3.12 with default settings.
           - python-version: '3.12'
 
+          # Newest macOS and oldest Python (major) supported versions.
+          - python-version: '3.8'
```
adiroiban, line 138: maybe try a very old 3.8

```suggestion
          - python-version: '3.8.0'
```

.github/workflows/test.yaml
```diff
@@ -134,6 +134,12 @@ jobs:
           # Just Python 3.12 with default settings.
           - python-version: '3.12'
 
+          # Newest macOS and oldest Python (major) supported versions.
+          - python-version: '3.8.0'
```
adiroiban: 3.8.0 has a different failue - https://github.com/twisted/twisted/actions/runs/6131008336/job/16640617797?pr=11974#step:10:13916

try something newer

```suggestion
          - python-version: '3.8.10'
```

adiroiban:
@graingert do you use Twisted on macOS with python 3.8 ?
If yes, then it's ok to spend more time with this PR.

Otherwise, I think we can leave this for someone who cares a lot about Twisted on macOS with python 3.8

It might be that all macOS users are on latest Python.... so we are just waisting time here.

.github/workflows/test.yaml
```diff
@@ -134,6 +134,12 @@ jobs:
           # Just Python 3.12 with default settings.
           - python-version: '3.12'
 
+          # Newest macOS and oldest Python (major) supported versions.
+          - python-version: '3.8.10'
```
```suggestion
          - python-version: '3.8'
```

adiroiban:
Maybe just skip `twisted.conch.test.test_scripts` on macOS with python 3.8 ?

It should be better than nothing.

And we can have a .feature bugfix release note with somehing like

```
Limited support was added for Python 3.8 on macOS.
If you plan to use Twisted on macOS with Python 3.8 check the discussion at
https://github.com/twisted/twisted/issues/11973
```

adiroiban:
@graingert I think that we can get this ready for merge.

Is there anything left to do to get this PR ready for review?

Thanks

graingert:
Just to double check that cfreactor is covered

src/twisted/conch/test/test_scripts.py
```diff
@@ -30,7 +30,10 @@
 else:
     try:
         tkinter.Tk().destroy()
-    except tkinter.TclError as e:
+    except (tkinter.TclError, RuntimeError) as e:
+        # On GitHub Action the macOS Python might not support the version of TK
+        # provided by the OS and it will raise a RuntimeError
+        # See: https://github.com/actions/setup-python/issues/649
```
graingert, line 36: It seems that setup-python is prone to releasing broken tkinter releases I think it would be better to reraise the exception if not on known bad sys.version/sys.platform so we're alerted to it if eg 3.11 starts failing suddenly

adiroiban: The change was done along the lines of "should Twisted continue to ship Tkinter-based UIs?"

I think that any extra effort should be done for deprecating tkinter support in Twisted.

adiroiban (approved):
I think that this is good enough, and much better than what we already have in `trunk`, which is not runny any tests on macOS with 3.8

I would prefer to spend any other extra time deprecating tk.

src/twisted/conch/test/test_scripts.py
glyph: There are definitely users of twisted with Tkinter applications. It mostly juts works for them.

exarkun: I think the point adi meant to raise here is that supporting the Tk application that ships with *conch* may be worth re-considering at this point - not that Twisted as a whole should desupport Tk and make it impossible to write applications that use Tk and Twisted.

adiroiban: Sorry for the misunderstanding.

yes. The idea is to deprecate the tk support in `twisted.conch`

We can continue to have the tk reactor.

glyph: Oh. Yeah… difficult to make the case that tkconch is holding up any critical infrastructure these days :)

I'd love to ship a tk demo that does something interesting with Twisted but the only thing I can get tkconch to do right now is to spit out this traceback, crash, and exit when I hit 'connect':

```
  File "/Users/glyph/Projects/Twisted/src/twisted/python/log.py", line 96, in callWithLogger
  File "/Users/glyph/Projects/Twisted/src/twisted/python/log.py", line 80, in callWithContext
  File "/Users/glyph/Projects/Twisted/src/twisted/python/context.py", line 117, in callWithContext
  File "/Users/glyph/Projects/Twisted/src/twisted/python/context.py", line 82, in callWithContext
    return func(*args, **kw)
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/selectreactor.py", line 148, in _doReadOrWrite
    why = getattr(selectable, method)()
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/tcp.py", line 248, in doRead
  File "/Users/glyph/Projects/Twisted/src/twisted/internet/tcp.py", line 253, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/Users/glyph/Projects/Twisted/src/twisted/conch/ssh/transport.py", line 764, in dataReceived
    self.dispatchMessage(messageNum, packet[1:])
  File "/Users/glyph/Projects/Twisted/src/twisted/conch/ssh/transport.py", line 781, in dispatchMessage
    f(payload)
  File "/Users/glyph/Projects/Twisted/src/twisted/conch/ssh/transport.py", line 1968, in ssh_KEX_DH_GEX_GROUP
    return self._ssh_KEX_ECDH_REPLY(packet)
  File "/Users/glyph/Projects/Twisted/src/twisted/conch/ssh/transport.py", line 1903, in _ssh_KEX_ECDH_REPLY
    d = self.verifyHostKey(hostKey, fingerprint)
  File "/Users/glyph/Projects/Twisted/src/twisted/conch/scripts/tkconch.py", line 471, in verifyHostKey
    {b"ssh-dss": "DSA", b"ssh-rsa": "RSA"}[keyType],
builtins.KeyError: b'ecdsa-sha2-nistp256'
```

Not exactly the best demonstration of our capabilities :-|

PR #11978, opened 2023-09-10
#11977 mypy disallow_any_generic = true

### LINKED ISSUE #11977: disallow_any_generic = true
It looks like some type annotations have been unintentionally using implicit Any

configure disallow_any_generic = true and fix the errors, use explicit Any or the correct type where possible

graingert:
## Scope and purpose

Fixes #11977


src/twisted/internet/defer.py
```diff
@@ -2226,7 +2227,7 @@ def unwindGenerator(*args: _P.args, **kwargs: _P.kwargs) -> Deferred[_T]:
 
 
 _ConcurrencyPrimitiveT = TypeVar(
-    "_ConcurrencyPrimitiveT", bound="_ConcurrencyPrimitive"
+    "_ConcurrencyPrimitiveT", bound="_ConcurrencyPrimitive[Any]"
```
graingert, line 2232: Why is _ConcurrencyPrimitive even generic? ` _SelfResultT` is only used in `def run` where it would be valid without a class level Generic

The genericism isn't exposed publicly because DeferredLock and DeferredSemaphore subclass `_ConcurrencyPrimitive[Any]`

src/twisted/web/_element.py
```diff
@@ -31,7 +41,7 @@ class Expose:
     on the class object of which they are methods.
     """
 
-    def __call__(self, *funcObjs: Callable) -> Callable:
+    def __call__(self, f: _Tc, /, *funcObjs: Callable[..., object]) -> _Tc:
```
graingert, line 44: if we don't like the new error message for the TypeError - we can use an `@override` instead

src/twisted/protocols/postfix.py
```diff
@@ -105,7 +107,13 @@ def do_put(self, keyAndValue):
...
 
 
-class PostfixTCPMapDictServerFactory(UserDict, protocol.ServerFactory):
+if TYPE_CHECKING or sys.version_info >= (3, 9):
+    _PostfixTCPMapDict = UserDict[bytes, Union[str, bytes]]
+else:
+    _PostfixTCPMapDict = UserDict
+
+
+class PostfixTCPMapDictServerFactory(_PostfixTCPMapDict, protocol.ServerFactory):
```
graingert, line 116: a bit horrible, maybe this should be `typing.Dict` instead of `collections.UserDict` one day

src/twisted/test/test_defer.py
```diff
@@ -387,15 +387,15 @@ def testDeferredListDontConsumeErrors(self) -> None:
...
+        resultLists: List[_DeferredListResultListT[None]] = []
         dl.addCallback(resultLists.append)
 
         d1.errback(GenericError("Bang"))
         self.assertEqual("Bang", errorTrap[0].value.args[0])
         self.assertEqual(1, len(resultLists))
         firstResult = resultLists[0][0]
         assert firstResult is not None
-        self.assertEqual("Bang", firstResult[1].value.args[0])
+        self.assertEqual("Bang", firstResult[1].value.args[0])  # type: ignore[attr-defined]
```
graingert, line 398: this shows the type annotation for DeferredList is working now, but the type annotation says the result can never be a Failure.

This lie works for Deferred but cause attr-defined errors for DeferredList

src/twisted/trial/_dist/disttrial.py
```diff
@@ -165,7 +166,7 @@ class WorkerPool:
     def _createLocalWorkers(
         self,
         protocols: Iterable[LocalWorkerAMP],
-        workingDirectory: FilePath,
+        workingDirectory: FilePath[Any],
```
graingert, line 169:
```suggestion
        workingDirectory: FilePath[str],
```

called `workingDirectory.child(str(x))`

graingert: ah no this should still be `FilePath[Any]` child is defined as`def child(self, name: OtherAnyStr) -> FilePath[OtherAnyStr]: ...`

src/twisted/test/test_defer.py
adiroiban: thanks. we can keep the ignore for now.

src/twisted/web/_element.py
adiroiban: I think that this should be food enough for this PR.

src/twisted/protocols/postfix.py
adiroiban: it should be ok for this PR

src/twisted/internet/defer.py
adiroiban: Just go with whatever you think is best. Thanks :)

adiroiban (approved):
Looks good. Thanks.

I think that we can merge this, as otherwise the PR might stay for a long time without a review.

If later someone has anything to say/comment/suggest changes for a particular type definition, we can make that change in a separate PR.

Thanks again

src/twisted/internet/defer.py
graingert: @adiroiban I have a PR removing the Generic and fixing up the type here https://github.com/twisted/twisted/pull/11986

PR #11979, opened 2023-09-11
11976 stop processing pipelined HTTP/1.1 requests that are received together

### LINKED ISSUE #11976: twisted.web.server doesn't support HTTP pipelining
twisted.web.server responds in wrong order

https://gist.github.com/mukeran/01874b38334bb7618415695d666a3f29

When running the above code with `xxx.xxx.xxx.xxx:10780` as the target reverse proxy site and sending the following HTTP pipeline request to twisted, it seems that twisted processes the request asynchronously.

The `/hello` route responds first, then the `/reverse` route responds. The pipeline response is out-of-order.

```http
GET /reverse HTTP/1.1
Host: a.com
Connection: keep-alive

GET /hello HTTP/1.1
Host: a.com
Connection: close

```

Twisted should send the response of `/reverse` first instead of `/hello`'s.

**Testing environment**

 - Operating System and Version:
     ```
     Linux *** 6.2.0-26-generic #26~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jul 13 16:27:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
     DISTRIB_ID=Ubuntu
     DISTRIB_RELEASE=22.04
     DISTRIB_CODENAME=jammy
     DISTRIB_DESCRIPTION="Ubuntu 22.04.2 LTS"
    ```
 - Twisted version: 23.8.0

glyph:
## Scope and purpose

Fixes #11976


src/twisted/web/http.py
```diff
@@ -2250,6 +2250,9 @@ def lineReceived(self, line):
         Called for each line from request until the end of headers when
         it enters binary mode.
         """
+        assert (
+            not self._handlingRequest
+        ), "when handling a request, we MUST be in raw mode to buffer the incoming data without parsing it"
```
glyph: I wanted the tests to run with this once but we can probably remove it if CI is happy.

src/twisted/web/test/test_web.py
```diff
@@ -1849,3 +1852,39 @@ def test_defaultReactor(self):
 
         factory = http.HTTPFactory()
         self.assertIs(factory.reactor, reactor)
+
+
+class QueueResource(Resource):
+    isLeaf = True
```
adiroiban:
```suggestion
class QueueResource(Resource):
    """
    Add all requests to an internal queue,
    without responding to the requests.
    You can access the requests from the queue and handle their response.
    """
    isLeaf = True
```

src/twisted/web/test/test_web.py
```diff
@@ -1849,3 +1852,39 @@ def test_defaultReactor(self):
...
+        proto = site.buildProtocol(None)
+        serverTransport = StringTransport()
+        proto.makeConnection(serverTransport)
+        proto.dataReceived(
+            b"GET /first HTTP/1.1\r\nHost: a\r\n\r\n"
+            b"GET /second HTTP/1.1\r\nHost: a\r\n\r\n"
+        )
+        self.assertEqual(len(qr.queue), 1)
+        qr.queue[0].finish()
+        self.assertEqual(len(qr.queue), 2)
```
adiroiban:
```suggestion
        # The TCP data contains 2 requests,
        # but only 1 request was dispatched,
        # as the first request was not yet finalized.
        self.assertEqual(len(qr.queue), 1)
        # The first request is finalized and the
        # second request is dispatched right away.
        qr.queue[0].finish()
        self.assertEqual(len(qr.queue), 2)
```

src/twisted/web/test/test_web.py
```diff
@@ -1849,3 +1852,39 @@ def test_defaultReactor(self):
...
         factory = http.HTTPFactory()
         self.assertIs(factory.reactor, reactor)
+
+
+class QueueResource(Resource):
+    isLeaf = True
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.queue: List[Request] = []
```
adiroiban: I had to read the test more than once to understand how the tests is designed.

Maybe this is a better name.

```suggestion
        self.dispatchedRequests: List[Request] = []
```

src/twisted/web/http.py
```diff
@@ -2431,14 +2434,17 @@ def allContentReceived(self):
...
-    def dataReceived(self, data):
-        """
-        Data was received from the network.  Process it.
-        """
-        # If we're currently handling a request, buffer this data.
+    def rawDataReceived(self, data):
+        # If we're currently handling a request, we'll be in raw mode. Buffer
+        # any data.
         if self._handlingRequest:
             self._dataBuffer.append(data)
```
adiroiban, line 2467:
```suggestion
        if self._handlingRequest:
            # We're currently handling a request.
            # Buffer any data for the next pending request.
            self._dataBuffer.append(data)
```

src/twisted/web/http.py
```diff
@@ -2431,14 +2434,17 @@ def allContentReceived(self):
...
         req.requestReceived(command, path, version)
 
-    def dataReceived(self, data):
-        """
-        Data was received from the network.  Process it.
-        """
-        # If we're currently handling a request, buffer this data.
+    def rawDataReceived(self, data):
+        # If we're currently handling a request, we'll be in raw mode. Buffer
+        # any data.
```
adiroiban:
```suggestion
    def rawDataReceived(self, data):
        """
        This is called when data is received after all the headers for the current requests are received.
        
        This can be the body of the current request.
        This data can also be a new request, received while we still process the current request.
        """
```

src/twisted/web/http.py
adiroiban: Maybe something like this... to check these type of errors in the wild.

I might have hit this bug in the wild, without knowing why.

I think that some .NET http client implementation have some HTTP pipelining and other IIS optimization and will send more requests in the same TCP package.


```suggestion
        # This is here since September 2023 to make sure we don't accidentally handle requests out of order.
        # If no bugs are reported in the next two year,
        # this line can be removed.
        assert (
            not self._handlingRequest
        ), "when handling a request, we MUST be in raw mode to buffer the incoming data without parsing it"
```

src/twisted/web/http.py
```diff
@@ -2431,14 +2434,17 @@ def allContentReceived(self):
 
         self._handlingRequest = True
 
+        # We go into raw mode here even though we will be receiving lines next
+        # in the protocol; however, this data will be buffered and then passed
+        # back to line mode in the setLineMode call in requestDone.
+        self.setRawMode()
```
adiroiban, line 2437: Maybe  add some note about pipelining to help other understand what is going on here.

```suggestion
        # back to line mode in the setLineMode call in requestDone.
        # This is done to comply with the HTTP pipelining specification.
        # https://en.wikipedia.org/wiki/HTTP_pipelining
        self.setRawMode()
```

adiroiban (commented):
Not sure if this is ready for review.
I left some comments as I had a bit of time to look into this.

Thanks for the quick fix.

----

I think the macOS failure is flaky ...  I did a manual retry

glyph:
please review

glyph:
sorry I forgot I need to summon the robot even for the first PR :)

src/twisted/web/http.py
glyph: This is a comment that reads back the names of the variables and I think is noisier than it is helpful.

src/twisted/web/http.py
glyph: This docstring is inaccurate, but I'll add one if that's the feedback.

src/twisted/web/http.py
glyph: This code should be impossible to hit in the wild and we have tests to verify that this is the case, I don't think random asserts like this are helpful.

src/twisted/web/http.py
glyph: I don't find this reference particularly illuminating, since we specifically *aren't* doing pipelining here, just buffering

src/twisted/web/http.py
adiroiban: We receive "text content" while in raw mode.
It's not 100%  "obvious" what is going on here.

This is why I think that some sort of documentation to explain the code design could help.

src/twisted/web/http.py
```diff
@@ -2250,6 +2250,9 @@ def lineReceived(self, line):
         Called for each line from request until the end of headers when
         it enters binary mode.
         """
+        assert (
```
adiroiban: Just a reminder, not to forget to remove this code before the merge.

src/twisted/web/test/test_web.py
```diff
@@ -1849,3 +1852,50 @@ def test_defaultReactor(self):
...
+            b"GET /second HTTP/1.1\r\nHost: a\r\n\r\n"
+        )
+        # The TCP data contains 2 requests,
+        # but only 1 request was dispatched,
+        # as the first request was not yet finalized.
+        self.assertEqual(len(qr.dispatchedRequests), 1)
+        # The first request is finalized and the
+        # second request is dispatched right away.
+        qr.dispatchedRequests[0].finish()
+        self.assertEqual(len(qr.dispatchedRequests), 2)
```
adiroiban, line 1902: I think that is best to have the new test added to `t.w.test.test_http.PipeliningBodyTests`

But this tests is still kept in test_web,
maybe it is worth adding the test in which the request is received in a separate package.

I know that this test is green even on `trunk`, but maybe it can help with future refactoring.

```suggestion
        self.assertEqual(len(qr.dispatchedRequests), 2)

    def test_multipleRequestsInDifferentSegments(self) -> None:
        """
        Twisted MUST NOT respond to a second HTTP/1.1 request while the first
        is still pending, even if the second request is received in a separate
        TCP package.
        """
        qr = QueueResource()
        site = Site(qr)
        proto = site.buildProtocol(None)
        serverTransport = StringTransport()
        proto.makeConnection(serverTransport)
        raw_data = (
            b"GET /first HTTP/1.1\r\nHost: a\r\n\r\n"
            b"GET /second HTTP/1.1\r\nHost: a\r\n\r\n"
        )
        # Just go byte by byte for the extreme case in which each byte is
        # received in a separate TCP package.
        for chunk in iterbytes(raw_data):
            proto.dataReceived(chunk)
        # The TCP data contains 2 requests,
        # but only 1 request was dispatched,
        # as the first request was not yet finalized.
        self.assertEqual(len(qr.dispatchedRequests), 1)
        # The first request is finalized and the
        # second request is dispatched right away.
        qr.dispatchedRequests[0].finish()
        self.assertEqual(len(qr.dispatchedRequests), 2)

```

src/twisted/web/test/test_web.py
```diff
@@ -1849,3 +1852,50 @@ def test_defaultReactor(self):
 
         factory = http.HTTPFactory()
         self.assertIs(factory.reactor, reactor)
+
+
+class QueueResource(Resource):
```
adiroiban, line 1858: why is this test added to test_web.py and not to test_http.py ?

Just asking.

adiroiban (approved):
The changes look good. Thanks for the quick fix.

I would prefer to have the new test added to `t.w.test.test_http.PipeliningBodyTests` , rather than test_web

This is the only major comment.

The others are minor.

Thanks again

src/twisted/web/test/test_web.py
glyph: `test_http` already has a test for this which doesn't cover the full stack top-to-bottom.

The boundaries are fuzzy, and this fix obviously has its foot in both worlds, but HTTP is for the HTTP protocol parsing, whereas web is for request handling.  This is for the boundary where a parsed HTTP protocol message *becomes* an invocation of the request handling logic.

At first I felt like the problem might be at a higher level because there's already a test for this in `test_http`, before I realized that it was a TCP segment colocality issue.  But, problems *might* be at a higher level in the future, so having the test live at this later, both in `test_web` and with `server.Site` being the layer of the public API getting tested, still makes sense to me.

adiroiban: thanks for the info

Maybe test_http.py needs a docstring that is more than `Test HTTP support.` same for `test_web.py` with `Tests for various parts of L{twisted.web}.` 

I will keep this in mind and will try to update the docstring, next time I will touch these files

adiroiban:
there was a flaky test failure on macOS . I manually triggered a retry.
I saw this test failing on macOS quite often.

PR #11983, opened 2023-09-11
#11982 Update pre-commit tools, with the exception of black.

### LINKED ISSUE #11982: Update pre-commit tools
The plan is to update the pre-commit tools, with the exception of `black`.

We can update black as part of https://github.com/twisted/twisted/issues/11961 or https://github.com/twisted/twisted/pull/11962

`black` is a separate beast since updating it will cause 170+ files to also be updated.

adiroiban:
## Scope and purpose

Fixes #11982

This is basically the work done by @graingert  in https://github.com/twisted/twisted/pull/11962 but without the `black` changes.

I hope we can get this merged and then we can focus on the `black` upgrade

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: 23.7.0
+    rev: 21.7b0
     hooks:
       - id: black
         additional_dependencies: ['click<8.1']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,23 +6,23 @@ repos:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: 21.7b0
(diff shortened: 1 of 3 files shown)


adiroiban:
I think that this is ready for review 

@graingert this is 99% your work.

I went with inline `#noqa`

needs-review

adiroiban:
this needs an update for mypy

```
src/twisted/internet/defer.py:1442:1: error: Name "DeferredList" already defined on line 1439  [no-redef]
src/twisted/internet/defer.py:1444: error: Unused "type: ignore" comment  [unused-ignore]
```

graingert (approved):

adiroiban:
Many thanks for the  review.

I have moved the inline  comment for mypy ... black was reformating the code.

I have enabled auto-merge. Hope all goes well.

glyph:
Thanks for splitting this stuff out so we can move forward faster and not block on Black! :)

PR #11986, opened 2023-09-12
#11985 fix DeferredLock.run/Semaphore.run/maybeDeferred/ensureDeferred/inlineCallbacks and fromCoroutine type annotation

### LINKED ISSUE #11985: DeferredLock and DeferredSemaphore run method is incorrectly type annotated
given

```python

def bar() -> defer.Deferred[int]:
    return defer.succeed(1)


l = defer.DeferredLock()
d = l.run(bar)
reveal_type(d)  # Revealed type is "twisted.internet.defer.Deferred[Any]"
```

the revealed type should be  "twisted.internet.defer.Deferred[int]"

graingert:
## Scope and purpose

Fixes #11985


src/twisted/internet/defer.py
```diff
@@ -2262,12 +2283,16 @@ def run(
...
         """
 
-        def execute(ignoredResult: object) -> Deferred[_SelfResultT]:
-            return maybeDeferred(f, *args, **kwargs).addBoth(self._releaseAndReturn)
+        def execute(ignoredResult: object) -> Deferred[_T]:
+            # maybeDeferred arg type requires one of the possible union members
+            # and won't accept all possible union members
+            return maybeDeferred(f, *args, **kwargs).addBoth(
+                self._releaseAndReturn
+            )  # type: ignore[return-value]
```
graingert, line 2324: I tried to get this to pass type checking but I can't work out how: https://mypy-play.net/?mypy=latest&python=3.12&gist=7a0b28937ca2f2137a55f29610da1d16&flags=strict

src/twisted/newsfragments/11985.bugfix
```diff
@@ -0,0 +1 @@
+Fix the type annotation of DeferredLock.run and DeferredSemaphore.run
```
graingert: could do with a hand writing the changelog entry for this

src/twisted/test/test_defer.py
```diff
@@ -846,6 +847,7 @@ def plusFive(x: int) -> int:
         results: List[int] = []
         errors: List[Failure] = []
         d = defer.maybeDeferred(plusFive, 10)
+        assert_type(d, Deferred[int])
```
graingert, line 851: this was correct

src/twisted/test/test_defer.py
```diff
@@ -929,6 +931,7 @@ def typedCallback(result: int) -> int:
             return result + 1
 
         coroutineDeferred = defer.maybeDeferred(g)
+        assert_type(coroutineDeferred, Deferred[int])
```
graingert, line 934: this was `Deferred[Any]` before

src/twisted/test/test_defer.py
```diff
@@ -915,6 +917,20 @@ def test_maybeDeferredCoroutineSuccess(self) -> None:
         by the function.
         """
 
+        async def f() -> int:
+            return 7
+
+        coroutineDeferred = defer.maybeDeferred(f)
+        assert_type(coroutineDeferred, Deferred[int])
```
graingert, line 925: this was `Deferred[Any]` before

src/twisted/test/test_defer.py
```diff
@@ -2701,6 +2718,15 @@ def helper(resultValue: object, returnValue: object = None) -> object:
...
         self.assertFalse(lock.locked)
 
+        def returnsInt() -> Deferred[int]:
+            return defer.succeed(2)
+
+        async def returnsCoroInt() -> int:
+            return 1
+
+        assert_type(lock.run(returnsInt), Deferred[int])
+        assert_type(lock.run(returnsCoroInt), Deferred[int])
```
graingert, line 2749: these were both `Deferred[Any]`

src/twisted/newsfragments/11985.bugfix
graingert:
```suggestion
Fix the type annotations of DeferredLock.run, DeferredSemaphore.run, maybeDeferred, ensureDeferred, inlineCallbacks and fromCoroutine that used to return Deferred[Any] to return the result of the passed Coroutine/Coroutine function
```

src/twisted/test/test_defer.py
```diff
@@ -1712,6 +1729,14 @@ def ic(d: object) -> Generator[Any, Any, None]:
         self.assertEqual("test_inlineCallbacksTracebacks", tb[0][2])
         self.assertEqual("f.raiseException()", tb[0][3])
 
+    def test_fromCoroutine(self) -> None:
```
adiroiban, line 1741: Can you please add a docstring explaining the scope/purpose of this test?

Thanks

Maybe 
```suggestion
    def test_fromCoroutine(self) -> None:
        """
        It can discover the return type of a first level coroutine.
        """
```

Now ... what `first-level` means :)
I can see that the coroutines are not chained ... not sure if the behaviour is different if you have a coroutine returning another coroutine.

adiroiban (approved):
Thanks for the changes.

I only did a brief review.

I am not a fan of typing...hence, not a fan of `assert_type` 
But, I guess that if someone want to add them, there is no reason to remove then.

As long as they are not mandatory, it should be fine.

My understanding is that you don't need `assert_type`.

Just write the tests as regular, and mypy will automatically assert the type for you.

We use just:

```
self.assertEqual(3, someResult)
```

We don't write the tests as

```
self.assertEqual(3, someResult)
self.assertIsInstance(3, (int,))
```

-----

But if it helps troubleshooting the tests, I think we can keep the `assert_type` part.

------------

Please also apply your suggestions for the release notes.

------------

I am going to approve this, but I will put it back for review.

If in 1 week, no other developer will look into this, I think it's safe to merge it.

Thanks

graingert:
Ideally I'd like a review from @glyph on this for touching the types here most recently

glyph:
> Ideally I'd like a review from @glyph on this for touching the types here most recently

I put this on the ol' todo list, will try to look soon

adiroiban:
@graingert I think that this can be merged. If Glyph or someone else wants to review and suggest changes we can have them done in a separate PR.

Thanks

graingert:
@adiroiban I needed to add more `Generator[Deferred[Any], Any, None]` due to the code introduced in https://github.com/twisted/twisted/pull/11886 using the old style

src/twisted/test/test_defer.py
```diff
@@ -1685,6 +1702,69 @@ def raiseError(ignored: object) -> None:
...
+        """
+        f = getDivisionFailure()
+        d: Deferred[None] = Deferred()
+        try:
+            f.raiseException()
+        except BaseException:
+            d.errback()
+
+        def ic(d: object) -> Generator[Any, Any, None]:
+            yield d
```
adiroiban: It is not clear how this test is constructured.

Also, it is not clear why we don't have coverage here.

I expect the missing coverage is expected, as I can see that the function is only wrapped and the resulting decorated function is not called.
 
```suggestion
        def ic(d: object) -> Generator[Any, Any, None]:  # pragma: no cover
            """
            This is never called.
            It is only used as the decorated function.
            The resulting function is never called in this test.
            This is used to make sure that if we wrap
            an already failed deferred, inlineCallbacks
            will not add any extra traceback frames. 
            """
            yield d
```

adiroiban (approved):
Thanks for the changes.

I left a somehow import comment in terms of missing coverage and test explanation.

Please consider updating the docstring and adding the nocoverage marker before merging.

Thanks

Issue #11989, opened 2023-09-13
Lots of small writes to TLS transport use a lot of CPU


itamarst:
**Describe the incorrect behavior you saw**

If you do lots of small writes to a TLS transport, it uses a really significant amount of CPU and bandwidth compared to larger writes. The impact seems far more significant in TLS than it would be in TCP, e.g. TCP would combine the writes due to Nagle and therefore usually won't have 23× bandwidth increase!

**Describe how to cause this behavior**

See upstream pyOpenSSL issue for reproducer and numbers (10× CPU and 23× bandwidth): https://github.com/pyca/pyopenssl/issues/1250

I discovered this in the wild in Foolscap, which writes many things one byte at a time... with code basically copy/pasted from Twisted: https://github.com/twisted/twisted/blob/trunk/src/twisted/spread/banana.py#L31 and later code.

**Describe the correct behavior you'd like to see**

Fixing this in pyOpenSSL is likely difficult; there is the upstream bug but it's not clear to me an extra buffering layer there makes sense. In contrast, `twisted/protocols/tls.py` already has a buffering layer, albeit for other use cases (wantread/wantwrite errors from OpenSSL).

The solution I'm imagining is having the TLS protocol register as a pull producer with the underlying TCP transport and aggregate small writes until it gets a pull message.

Another alternative is to just document this behavior, and fix upstream protocol implementations to be less chatty.

**Testing environment**
- Twisted checkout as of Sep 12, 2023
- Ubuntu 22.04

glyph:
Thanks for the detailed report! It seems like we could easily batch writes once per reactor tick at the very least, and then do something nagle-ish if we want to be more sophisticated about it?

We could overload `ITransport.setTcpNoDelay` or have a separate method to control the buffering behavior…

itamarst:
Oh I guess that's simpler than a producer yeah. I'll try implementing that. If it's every reactor tick, it won't really add much latency (certainly much less than Nagle) so not sure if `setTcpNoDelay()` is a requirement.

itamarst:
I read `TLSMemBIOProtocol` and... adding this is going to be complicated: it has an implicit state machine already that relies on 4-5 different attributes, and this adds a bunch more state. So possibly I will need to refactor the class first to have an explicit state machine.

itamarst:
Another alternative is having current code be a `LowLevelTLSMemBioProtocol` and then there's a new `TLSMemBIOProtocol` that wraps it and adds buffering.

[...]

PR #11991, opened 2023-09-18
#5206 When using ReactorBuilder, twisted.internet.reactor points to the temporary reactor

### LINKED ISSUE #5206: ReactorBuilder test infrastructure breaks code that uses the global reactor
|[itamarst's avatar](https://github.com/itamarst)| @itamarst reported|
|-|-|
|Trac ID|trac#5206|
|Type|defect|
|Created|2011-07-27 11:52:15Z|

Code that uses the global reactor, e.g. `twisted.internet.task.cooperate`, will not work correctly with tests that use the ReactorBuilder infrastructure. While such code should be updated to have an explicit reactor passed in, doing so can be non-trivial (e.g. `cooperate` needs a registry, or it's a 3rd party library that cannot be modified).

Providing some workaround to this would both be helpful for testing, but also (depending on the implementation) could later be expanded to help catch code that relies on the global reactor instance.

```
trac-id__5206 5206
type__defect defect
reporter__itamar itamar
priority__normal normal
milestone__ 
branch__ 
branch_author__ 
status__closed closed
resolution__wontfix wontfix
component__core core
keywords__ 
time__1311767535000000 1311767535000000
changetime__1321499609000000 1321499609000000
version__None None
owner__itamar itamar

```
</details>

itamarst:
## Scope and purpose

Fixes #5206

Working on #11989, I end up fighting the fact that the TLS classes don't have a reference to the reactor, so I need to rely on the global reactor, which then hits the limitation of ReactorBuilder that it doesn't affect the global reactor.

This PR makes ReactorBuilder affect the global reactor. It's horrible, I admit, but the alternative is lots of patching.

## Contributor Checklist:


src/twisted/python/test/strategies.py
```diff
@@ -26,7 +26,7 @@ def systemdDescriptorNames() -> SearchStrategy[str]:
             min_codepoint=0,
             max_codepoint=127,
             # This one excludes control characters.
-            blacklist_categories=("Cc",),
+            blacklist_categories=["Cc"],
```
itamarst: This is unrelated fix that makes `mypy` happy with newer Hypothesis.

src/twisted/internet/test/test_tls.py
```diff
@@ -390,16 +389,6 @@ class AbortSSLConnectionTests(
     requiredInterfaces = (IReactorSSL,)
     endpoints = SSLCreator()
 
-    def buildReactor(self):
```
itamarst, line 393: No longer necessary since we're now making the global reactor match the temporary one.

glyph:
Could the TLS protocol classes be adapted, instead, to use something like `.transport.reactor`, to avoid doing this?  ReactorBuilder not supporting patching the global reactor is one of the few subtle pressures we've got to encourage maintenance to slowly remove global reactor references.

itamarst:
Is `.transport.reactor` guaranteed to work on all transports and/or all reactor implementations? It's not part of the `ITransport` interface, so there's no reason to think it'll be there.

glyph:
> Is `.transport.reactor` guaranteed to work on all transports and/or all reactor implementations? It's not part of the `ITransport` interface, so there's no reason to think it'll be there.

It absolutely isn't! But is `__dict__` assignment guaranteed to work across all reactors? ;-)

itamarst:
The difference is that `__dict__` assignment only happens in the test suite, but `.transport.reactor` would have to happen in non-test code.

itamarst:
How about I do the other PR without relying on this, and then it'll be clearer why this is useful.

src/twisted/internet/test/reactormixins.py
```diff
@@ -247,6 +247,13 @@ def unbuildReactor(self, reactor):
         for c in calls:
             c.cancel()
 
+        if global_reactor_was_munged:
```
adiroiban: Just a minor comment.
Maybe it would be better to have this as a separate function called `unbuildGlobalReactor`

I think that having a call like

```
self.addCleanup(self.unbuildReactor, reactor)
self.addCleanup(self.unbuildGlobalReactor, reactor)
```

is easier to read than the "magic"

```
self.addCleanup(self.unbuildReactor, reactor)
```

----

Basically, this is the reverse, with `global_reactor_was_munged=False` by default

src/twisted/internet/test/test_gireactor.py
```diff
@@ -104,7 +104,7 @@ def test_gApplicationActivate(self) -> None:
         L{Gio.Application} instances can be registered with a gireactor.
         """
         reactor = gireactor.GIReactor(useGtk=False)
-        self.addCleanup(self.unbuildReactor, reactor)
+        self.addCleanup(self.unbuildReactor, reactor, False)
```
adiroiban: I can see that the test is executed on linux and the coverage was published for that run... maybe a codecov.io bug

image

adiroiban (approved):
Thanks for the cleanup.

Only minor comment regarding using the flag argument vs a separate function. 

As this is just test code, I think we can merge it.

If anyone find something wrong with this code, we can change it later.

src/twisted/internet/test/reactormixins.py
```diff
@@ -216,7 +216,7 @@ def tearDown(self):
                         % (process.reapProcessHandlers,)
                     )
 
-    def unbuildReactor(self, reactor):
+    def unbuildReactor(self, reactor, global_reactor_was_munged=True):
```
glyph:
```suggestion
    def unbuildReactor(self, reactor, globalReactorWasMunged=True):
```

src/twisted/internet/test/reactormixins.py
glyph:
```suggestion
        if globalReactorWasMunged:
```

glyph (commented):
(Just a couple of coding standard inline suggestions for your convenience. I don't disagree with Adi's feedback, if you want to go that route.)

src/twisted/internet/test/reactormixins.py
itamarst: Hm. It is a little confusing as is. Perhaps a better solution is making `test_gireactor.py` work better, since that's the only use case for the flag.

itamarst: I ended making `unbuildReactor` private.

src/twisted/internet/test/test_gireactor.py
```diff
@@ -103,8 +103,8 @@ def test_gApplicationActivate(self) -> None:
         """
         L{Gio.Application} instances can be registered with a gireactor.
         """
-        reactor = gireactor.GIReactor(useGtk=False)
-        self.addCleanup(self.unbuildReactor, reactor)
+        self.reactorFactory = lambda: gireactor.GIReactor(useGtk=False)
+        reactor = self.buildReactor()
```
adiroiban, line 107: The refactoring looks very good.

Just a minor suggestion:
```suggestion
        reactor = self.buildReactor(reactorFactory=lambda: gireactor.GIReactor(useGtk=False))
```

PR #11992, opened 2023-09-19
#11993 type annotate more tests

### LINKED ISSUE #11993: type annotate more tests
tests are generally safe to type annotate without breaking anyone lets add some more -> Nones

graingert:
## Scope and purpose

Fixes #11993

### DIFF
--- a/src/twisted/test/test_log.py
+++ b/src/twisted/test/test_log.py
@@ -505,7 +505,7 @@ class FileObserverTests(LogPublisherTestCaseMixin, unittest.SynchronousTestCase)
         when = calendar.timegm((2001, 2, 3, 4, 5, 6, 7, 8, 0))
 
         # Pretend to be in US/Eastern for a moment
-        self.flo.getTimezoneOffset = lambda when: 1800  # type: ignore[method-assign]
+        self.flo.getTimezoneOffset = lambda when: 18000  # type: ignore[method-assign]
         self.assertEqual(self.flo.formatTime(when), "2001-02-02 23:05:06-0500")
 
         # Okay now we're in Eastern Europe somewhere
--- a/src/twisted/test/test_log.py
+++ b/src/twisted/test/test_log.py
@@ -717,7 +717,7 @@ class FileObserverTests(LogPublisherTestCaseMixin, unittest.SynchronousTestCase)
         class PreStartObserver:
             def __call__(self, eventDict: log.EventDict) -> None:
                 if "pre-start" in eventDict.keys():
-                    # later we assert that received is empty so this is not covered 
+                    # later we assert that received is empty so this is not covered
                     received.append(eventDict)  # pragma: no cover
 
         newPublisher(evt)
--- a/src/twisted/test/test_log.py
+++ b/src/twisted/test/test_log.py
@@ -329,7 +329,7 @@ class _LogPublisherTestCaseMixinBase(Protocol):
     def assertIsInstance(
         self, instance: object, classOrTuple: type, message: str
     ) -> None:
-        pass
+        ...
 
 
 class LogPublisherTestCaseMixin:
--- a/src/twisted/test/test_log.py
+++ b/src/twisted/test/test_log.py
@@ -13,9 +13,14 @@ import sys
 import time
 import warnings
 from io import IOBase, StringIO
-from typing import List
+from typing import Callable, List
+
+from zope.interface import implementer
+
+from typing_extensions import Protocol
 
 from twisted.logger import (
+    ILogObserver,
     LogBeginner,
     LoggingFile,
     LogLevel as NewLogLevel,
@@ -39,7 +44,7 @@ class TextFromEventDictTests(unittest.SynchronousTestCase):
     Tests for L{textFromEventDict}.
     """
 
-    def test_message(self):
+    def test_message(self) -> None:
         """
         The C{"message"} value, when specified, is concatenated to generate the
         message.
@@ -48,7 +53,7 @@ class TextFromEventDictTests(unittest.SynchronousTestCase):
         text = log.textFromEventDict(eventDict)
         self.assertEqual(text, "a b c")
 
-    def test_format(self):
+    def test_format(self) -> None:
         """
         The C{"format"} value, when specified, is used to format the message.
         """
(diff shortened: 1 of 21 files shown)


glyph (approved):
love to see it

src/twisted/test/test_log.py
```diff
@@ -490,40 +497,40 @@ def test_getTimezoneOffsetWithoutDaylightSavingTime(self):
...
+    def test_timeFormatting(self) -> None:
         """
         Test the method of L{FileLogObserver} which turns a timestamp into a
         human-readable string.
         """
         when = calendar.timegm((2001, 2, 3, 4, 5, 6, 7, 8, 0))
 
         # Pretend to be in US/Eastern for a moment
-        self.flo.getTimezoneOffset = lambda when: 18000
+        self.flo.getTimezoneOffset = lambda when: 1800  # type: ignore[method-assign]
```
graingert:
```suggestion
        self.flo.getTimezoneOffset = lambda when: 18000  # type: ignore[method-assign]
```

src/twisted/test/test_log.py
```diff
@@ -316,8 +323,17 @@ class EvilReprStr(EvilStr, EvilRepr):
     pass
 
 
+class _LogPublisherTestCaseMixinBase(Protocol):
+    out: FakeFile
+
+    def assertIsInstance(
+        self, instance: object, classOrTuple: type, message: str
+    ) -> None:
+        pass
```
adiroiban: maybe use 3 dots to auto-skip the coverage?

src/twisted/test/test_log.py
```diff
@@ -329,14 +345,7 @@ def setUp(self):
         self.flo = log.FileLogObserver(self.out)
         self.lp.addObserver(self.flo.emit)
 
-        try:
-            "\N{VULGAR FRACTION ONE HALF}"
-        except UnicodeEncodeError:
-            pass
-        # This is the behavior we want - don't change anything.
```
adiroiban, line 336: I guess this was valid only for py2?

I am not sure why the previous code was needed

adiroiban (approved):
It looks good.
I only did the review of the changes done after the review done by Glyph.

Many thanks for the cleanup!

src/twisted/test/test_log.py
graingert:
```suggestion
        ...
```

src/twisted/test/test_log.py
```diff
@@ -698,28 +705,30 @@ class SysModule:
...
         received = []
 
-        def preStartObserver(x):
-            if "pre-start" in x.keys():
-                received.append(x)
+        @implementer(ILogObserver)
+        class PreStartObserver:
+            def __call__(self, eventDict: log.EventDict) -> None:
+                if "pre-start" in eventDict.keys():
+                    received.append(eventDict)
```
graingert:
```suggestion
                    received.append(eventDict)  # pragma: no cover
```

src/twisted/test/test_log.py
```diff
@@ -698,28 +705,30 @@ class SysModule:
...
         received = []
 
-        def preStartObserver(x):
-            if "pre-start" in x.keys():
-                received.append(x)
+        @implementer(ILogObserver)
+        class PreStartObserver:
+            def __call__(self, eventDict: log.EventDict) -> None:
+                if "pre-start" in eventDict.keys():
+                    received.append(eventDict)  # pragma: no cover
```
adiroiban, line 721: maybe add a comment why this `no cover` is here.

Under normal usage, we should just delete any test code that has no coverage, as this code is never used.

Maybe replace this code branch with raising an AssertionError, to communicate that pre-start should never be called in the scope of this test.
thanks

src/twisted/test/test_plugin.py
```diff
@@ -570,9 +583,11 @@ def test_newPluginsOnReadOnlyPath(self):
             errno.EPERM,
         )
         for event in events:
-            if expected in textFromEventDict(event):
+            maybeText = textFromEventDict(event)
+            assert maybeText is not None
+            if expected in maybeText:
```
graingert:
```suggestion
            if expected in maybeText:  # pragma: no branch
```

src/twisted/test/test_log.py
graingert: I don't think adding an assertion will work because exceptions get caught by the log observation machinery

src/twisted/test/test_log.py
graingert: yeah checking git blame, previously this called `if not _PY3: reload(sys).setdefaultencoding("ascii")` which isn't needed anymore because setdefaultencoding doesn't exist anymore

src/twisted/test/test_log.py
graingert: this whole section of the test is weird, it makes a new LogPublisher submits an event to it then adds an observer which can never receive an event

graingert: ah no we're attaching it to "theLogPublisher" using `tempLogPublisher` and then asserting that a warning doesn't result in any pre-start events

graingert, line 721:
```suggestion
                    # later we assert that received is empty so this is not covered 
                    received.append(eventDict)  # pragma: no cover
```

adiroiban: we can use pragma then, and just add a comment that this code is not covered on purpose.

an maybe rename `received` to `preStartUnexpectedEvents` or something like that.

to inform that we are expecting it to be empty

but I think the current comment is also ok

thanks

adiroiban (approved):
Thanks.

I think that we can merge this.

src/twisted/test/test_log.py
glyph: It's not a big deal, but this sort of lack of coverage is an anti-pattern.  Better is to have a fixture that gets re-used in tests that hit both code paths, so that both the positive and negative paths are caught.  Not worth holding up a merge if it's a complicated refactoring, but in general we should try to take these kinds of gaps seriously when writing new code.

As @graingert says though this test is weird, some tests are just weird and it's not worth the effort to fix them ;)

PR #11996, opened 2023-09-19
#11989 Lots of small writes to the TLS transport use a lot of cpu

### LINKED ISSUE #11989: Lots of small writes to TLS transport use a lot of CPU

itamarst:
## Scope and purpose

Fixes #11989

This makes TLS much faster when it involves many small writes ([see here for a before/after benchmark](https://github.com/twisted/twisted/issues/11989#issuecomment-1723776842) - actual results are slightly worse but it's still a big improvement).

As a consequence of relying on the passage of time, some tests needed to be updated to advance time with a `Clock`.

## Contributor Checklist:


src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
...
+    def write(self, data: bytes) -> None:
+        if isinstance(data, str):  # type: ignore[unreachable]
+            raise TypeError("Must write bytes to a TLS transport, not str.")
+        self._aggregator.write(data)
+
+    def writeSequence(self, sequence: Iterable[bytes]) -> None:
+        self._aggregator.write(b"".join(sequence))
+
+    def loseConnection(self) -> None:
+        self._aggregator.flush()
```
itamarst, line 800: Commenting this line out causes test failures in `twisted.protocols.test.test_tls`, so I feel it has sufficient test coverage even without new tests.

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
...
+        factory: TLSMemoryBIOFactory,
+        wrappedProtocol: IProtocol,
+        _connectWrapped: bool = True,
+    ):
+        super().__init__(factory, wrappedProtocol, _connectWrapped)
+        actual_write = super().write
+        self._aggregator = AggregateSmallWrites(actual_write, factory._clock)
+
+    def write(self, data: bytes) -> None:
+        if isinstance(data, str):  # type: ignore[unreachable]
```
itamarst, line 792: Personally I feel this check should be deleted, it adds performance overhead and type checking means these bugs are easier to catch. But I left it in given there are tests that make sure this check exists.

hacklschorsch:
🚀!

src/twisted/test/test_sslverify.py
```diff
@@ -288,17 +289,27 @@ def connectionLost(self, reason):
...
     )
 
     sProto, cProto, pump = connectedServerAndClient(
         lambda: serverFactory.buildProtocol(None),
         lambda: clientFactory.buildProtocol(None),
     )
+
+    # Need time to pass for flushing to work:
+    def flush(pump_flush=pump.flush):
+        clock.advance(0.001)
```
twm: Does advancing 0.0 not work? ~It'd be preferable to avoid disrupting tests that rely on advancing time in precise ways (I've written such tests in the past to cover timeout/retry logic).~

itamarst: Hm. Looking at the implementation, it might actually.

src/twisted/protocols/tls.py
itamarst: @twm thoughts on this? If you don't want to think about changing this in this PR, I can do it as a separate issue + PR.

twm: I think it would be totally reasonable to remove this check &mdash; it was introduced as a Python 3 porting aid, I assume. It does merit a note in the release notes IMO, so a separate PR would be nice.

itamarst: OK, follow-up PR it'll be, then.

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
         return self._connectionForTLS(protocol)
 
 
+class AggregateSmallWrites:  # TODO make this private for now?
```
adiroiban: I think is best to have this private.

src/twisted/test/test_sslverify.py
```diff
@@ -288,17 +289,27 @@ def connectionLost(self, reason):
...
-        serverOpts, isClient=False, wrappedFactory=plainClientFactory
+        serverOpts, isClient=False, wrappedFactory=plainClientFactory, clock=clock
     )
 
     sProto, cProto, pump = connectedServerAndClient(
         lambda: serverFactory.buildProtocol(None),
         lambda: clientFactory.buildProtocol(None),
     )
+
+    # Need time to pass for flushing to work:
```
adiroiban: can we have this as the default behaviour for `pump` so that we don't have to make this change here.

I can see that things are complicated, since the previous code is not dealing with the clock.

I expect that with this change, many other projects for which the TLS tests are written using iosim might need to be updated.

If iosim `pump` will automatically handle the new buffering behaviour, no changes should be required in the test code of 3rd party projects.

so maybe just call something like `reactor.doIteration(0.000001)` inside `pump.flush()`

Maybe `reactor.doIteration(0)` will also work.

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
...
+    def write(self, data: bytes) -> None:
+        """
+        Buffer the data, or write it immediately if we've accumulated enough to
+        make it worth it.
+
+        Accumulating too much data can result in higher memory usage.
+        """
+        self._buffer.append(data)
+        self._bufferLen += len(data)
+        if self._bufferLen > 64_000:
```
adiroiban: Minor comment.

Maybe extract this value into a class member, and have it documented at the start of the class definition

```suggestion
        if self._bufferLen > self._MAX_CHUNK:
```

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
...
+            self.flush()
```
adiroiban, line 727: maybe just return after this flush.

It should make it easier to read as it's a hint that nothing should happen after the flush.

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,100 @@ def clientConnectionForTLS(self, protocol):
...
+        if self._scheduled is None:
+            self._scheduled = self._clock.callLater(0, self._scheduledFlush)
```
adiroiban: Just minor comment that I hope will make the code easier to read

```suggestion
        if self._scheduled:
            # We already have a scheduled send,
            # so with the data in the buffer,
            # there is nothing more to do here.
            return

        # Schedule the write of the accumulated buffer
        # for the next reactor iteration.
        self._scheduled = self._clock.callLater(0, self._scheduledFlush)
```

src/twisted/protocols/test/test_tls.py
```diff
@@ -1819,3 +1833,53 @@ def acceptableProtocols(self):
...
+                st.integers(min_value=1, max_value=100_000).map(
+                    lambda length: (b"0123456789ABCDEFGHIJ" * ((length // 20) + 1))[
+                        :length
+                    ]
+                ),
+            ),
+            max_size=1_000,
+        )
+    )
+    def test_writes_get_aggregated(self, writes: list[Union[bytes, None]]) -> None:
```
adiroiban, line 1886: I am not a big fan of writing tests in this way.
It makes the test code harder to read ...and reading the test code is the only way to check the validity of the tests.

also, by writing multiple tests you can write separate tests for each scenario.

adiroiban (commented):
Thanks for the PR.

I think that this is an important change and would be happy to see this merged.


I left some comments.
I hope that the comment will make it easier for the next person to review this PR.


I am -1 on style in which the tests are written... but if other Twisted devs are happy with the style, I don't block the merge.

Maybe the current tests are ok, as unit tests for AggregateSmallWrites, but I prefer to have `AggregateSmallWrites` as a private implementation and write unit and integration tests for the new public `BufferingTLSTransport`


In terms of test cases for BufferingTLSTransport:

* Receive a new small chunk to write ... see that it is not written right away... it is written at the next iteration... on following iteration, nothing happens
* Receive a new small chunk and then another small chunk... still not written right way, but sent together on next reactor iteration
* Receive a new small chunk and then a big chunk, check that it's written right away


I hope it helps. Thanks again for the PR.

src/twisted/protocols/test/test_tls.py
exarkun: What do you mean "this way"?

src/twisted/protocols/test/test_tls.py
```diff
@@ -1819,3 +1833,53 @@ def acceptableProtocols(self):
...
+                st.none(),
+                st.integers(min_value=1, max_value=100_000).map(
+                    lambda length: (b"0123456789ABCDEFGHIJ" * ((length // 20) + 1))[
+                        :length
+                    ]
+                ),
+            ),
+            max_size=1_000,
+        )
+    )
```
meejah, line 1885: Adding a descriptive comment to help people unfamiliar with Hypothesis might be nice. e.g.:

`# "writes" is a list of proposed writes, each bytestrings 1 to 100k long or a None`

Also I _believe_ that `None` is used here as a sentinel to flush the output (and not to test some feature of the interface) -- so that could be clarified in a comment too.

src/twisted/protocols/tls.py
meejah: +1

meejah (commented):
Overall, LGTM (some inline comments).

I am interested in the "downstream" mentions and implications for such projects. I looked quickly in `txtorcon` (and all my tests pass against this branch) and this leads me to an additional question for downstream:

If one is using `TLSMemoryBIOFactory` do we see the benefits of this buffering, or not?

src/twisted/protocols/test/test_tls.py
adiroiban: I mean writing test code that has its logic flow, with nested if/else, without describing the use cases that are covered.

I prefer tests that just setup the system under tests for a specific scenario, the system is executed and then the result is asserted.

Also, if there is a bug for a particular use case, you should see that particular tests failing.

--------

I understand the need for parameters on tests, but I think it makes the tests harder to read.

-----

For example, is not easy to understand the scenario in which `None` is used.

adiroiban:
> If one is using TLSMemoryBIOFactory do we see the benefits of this buffering, or not?

If you use `TLSMemoryBIOFactory` with the default protocol, you should automatically get these benefits.

The default protocol for `TLSMemoryBIOFactory` is now the buffering TLS protocol.

src/twisted/protocols/test/test_tls.py
itamarst: I tried refactoring it to be clearer.

adiroiban:
@itamarst I see that you have requested a new review, but I don't see any code changes, other than a merge.

For example, I see that you marked the comment for the `class AggregateSmallWrites: # TODO make this private for now?` line as resolved, but the TODO comment is still there, and the class is still public.

Is that on purpose ?

THanks

itamarst:
Conflict meant my push failed :facepalm:, fixed.

src/twisted/protocols/test/test_tls.py
adiroiban: This still looks complicated to me and if there is a bug, I am not sure how I am supposed to troubleshot this test.

It's not clear how you do TDD with these style of tests. I tried to search hypothesis docs for TDD but there are no results.


It's not clear why the decorators are needed.

We have `st.none()` defined in the decorator, and the explanation is in the docstring.

To me ,it seems like the decorator can be replaced with simple function calls.

```
@given(st.integers(), st.integers())
def test_ints_are_commutative(x, y):
    assert x + y == y + x

def test_ints_are_commutative():
    x = st.integers()
    y = st.integers()
    assert x + y == y + x
```

I understand the need for things like `st.integers()` 
but I think that instead of decorators, is better to have a dedicated generator function.

I understand that for complex generators, it can lead to some boilerplace code... but this code is harder to read

src/twisted/protocols/test/test_tls.py
```diff
@@ -1819,3 +1833,77 @@ def acceptableProtocols(self):
...
+
+
+class _AggregateSmallWritesTests(SynchronousTestCase):
+    """Tests for ``_AggregateSmallWrites``."""
+
+    @given(
+        st.lists(
+            st.one_of(
+                st.none(),
+                st.integers(min_value=1, max_value=100_000).map(
```
adiroiban, line 1877: This is more of a philosophical comment in the context of Twisted dev process.

We don't have much time dedicated to doing code reviews.

I had to read this code a few times to understand what is going on here and how the test is structured.

I think that this can discourage a review.

Someone might not have much to do a review, will read the code and will just stop doing the review.

I think that a lot of the contribution to Twisted are done as a 2nd job, without too much time dedicated to doing code reviews.

Now you can argue that if you don't have time to do top quality code review, is better not do do it ... but then we end up with nobody doing reviews.


-------

I understand that the current test input generate is comprehensive and define in a brief way.
I find it hard to read.

I think is better to have comment close this definition.

For example, I don't understand why we need this whole integer to bytes mapping here.
Why not use `st.binary()` ?

src/twisted/test/iosim.py
```diff
@@ -309,10 +313,11 @@ def flush(self, debug=False):
 
     def pump(self, debug=False):
         """
-        Move data back and forth.
+        Move data back and forth, and increase clock slightly.
 
         Returns whether any data was moved.
         """
+        self.clock.advance(0.000001)
```
adiroiban: I remember hitting the same issue while writing tests for my code.

Looking at this, I guess that this is a bug in `twisted.internet.task.Clock`

if you do `clock.callLater(0, lambda: None)` I am expecting  to have it called on `clock.advance(0)`

I see this ticket https://github.com/twisted/twisted/issues/5962

Maybe the current behaviour is caused by this bug... but it doesn't look like this is the cause.

So maybe is best to create a separate ticket and add a comment to that ticket here, so that once the ticket is fixed, we can update this.

This can be a low priority.

But if use write tests with iosim, the fact that the time auto-magically is advanced, might end up with tests failing for apparent unknown reasons

src/twisted/test/test_iosim.py
```diff
@@ -299,3 +300,15 @@ def test_clientStreamingProducer(self) -> None:
         (stream producer) registered with the client transport.
         """
         self._testStreamingProducer(mode="client")
+
+    def test_timeAdvances(self) -> None:
+        """
+        L{IOPump.pump} advances time in the given L{Clock}.
```
adiroiban, line 306: maybe add a comment here why advancing the time is a feature or temporary workaround.

src/twisted/protocols/tls.py
```diff
@@ -682,6 +690,116 @@ def clientConnectionForTLS(self, protocol):
...
+            self._bufferLen = 0
+            self._write(b"".join(self._buffer))
+            del self._buffer[:]
+
+
+def _get_default_clock() -> IReactorTime:
+    """
+    Return the default reactor.
+
+    This is a function so it can be monkey-patched in tests; see #5206.
```
adiroiban: Can we have https://github.com/twisted/twisted/pull/11991 merged so that we will not need this ?

adiroiban (approved):
Changes looks good. Many thanks for the updated.

I left a few minor comments.

----------

I also left a few philosophical questions, but they should not block this PR.

I am +1 for using test input generators, instead of static/hardcoded values.

I am using custom generators in my private projects and I think that is good to have a 3rd party library that can be used to share these type of generators. 

But I am not sure if `st.binary()` is good enough and will not lead to flaky tests.

For my bytes generator helper, I try to always generate bytes that will fail UTF-8 or UTF-16 decoding. In this way, it should help identity cases in which binary data is accidentally handled as text.

src/twisted/protocols/test/test_tls.py
itamarst: I didn't do `st.binary()` because my first though is it means it spends time searching the input space random bytes, and this test isn't about the contents of the bytes, it's about the lengths.

src/twisted/test/iosim.py
itamarst: The only reason I did this is because you requested it in a previous review comment. Perhaps I misunderstood what you wanted?

itamarst: Specifically, you asked for it here: https://github.com/twisted/twisted/pull/11996#discussion_r1345570974

src/twisted/protocols/test/test_tls.py
itamarst: You need the decorator to ensure you run multiple versions of the inputs.

src/twisted/test/iosim.py
itamarst: Ohhhh, you just don't want 0.000001. That's easy to fix, it's not actually necessary.

src/twisted/protocols/tls.py
itamarst: I'll try.

itamarst:
I added a minimal end-to-end test of small-write buffering, just to show it's hooked up.

src/twisted/protocols/test/test_tls.py
meejah: Another way to say why there are decorators is: that's just the way Hypothesis works.

I do get @adiroiban hesitations; Hypothesis isn't always very intuitive at first (is my recollection). That said, though, it is pretty powerful and my (somewhat limited) use of it _has_ revealed surprising (to me) bugs. It's also decent at "narrowing" so it doesn't just dump out the first random failing thing (it tries to make a shorter example). I don't think Hypothesis is appropriate for every test, but it is a powerful tool and can get a lot of mileage out of a single test.

All that said, I do agree that this composite strategy is hard to read. The "none" part is hard to understand (at least a quick comment as to why -- "indicates a flush", I think? -- would help a lot).

Another way to factor this code would be to make a custom strategy -- then the docstring for that custom strategy can explain what it is generating and how/why. I would probably use `binary()` since that's at least familiar for Hypothesis users (and does something extremely similar to this code).

(Maybe I'll try to state this suggestion as a diff....)

itamarst: I pushed some updates so if you do, work off of that.

src/twisted/protocols/tls.py
itamarst: I fixed one part, but this particular one is still necessary for `test_agent.py`, unfortunately.

adiroiban:
Thanks for the update.

> Ohhhh, you just don't want 0.000001. That's easy to fix, it's not actually necessary.

yes. My issue was with 0.00001 vs just `0` .

As long as the time is not changed by iosim, it should be ok.

-----------

Thanks for the info about hypothesis decorators. It makes sense.

So maybe we just need a helper like `st.binaryLengthVariation(min, max)` that will generate bytes of various lengths.

But this should not be a blocker for this PR.

----------

Also thanks for the new integration tests :)

As soon as Github Actions is happy, I think that this should be merged.

Thanks again for the PR.

hacklschorsch:
🎉 🥳

Issue #8570, opened 2016-07-07
Incorrect (too long) header from server causes Twisted Agent to fail


twisted-trac:
|[<img alt="pawelmhm's avatar" src="https://avatars.githubusercontent.com/u/2700942?s=50" width="50" height="50">](https://github.com/pawelmhm)| @pawelmhm reported|
|-|-|
|Trac ID|trac#8570|
|Type|enhancement|
|Created|2016-07-07 12:58:22Z|

If Twisted Agent receives HTTP header with value that exceeds size limits defined in protocol it will fail and crash. 

Expected behavior: if server sends incorrect header client should not fail but simply ignore this header.

To reproduce, create following Twisted server sending header with 100 000 characters in value:

```
#
import sys
from twisted.python import log
from twisted.web import server, resource
from twisted.internet import reactor

class Simple(resource.Resource):
    isLeaf = True

    def render_GET(self, request):
        request.setHeader("Set-Cookie", "a" * 100000)
        return "<html>Hello, world!</html>"

site = server.Site(Simple())
reactor.listenTCP(8080, site)
log.startLogging(sys.stdout)
reactor.run()
```

Twisted client trying to reach above server:

```
from twisted.internet import reactor
from twisted.web.client import Agent
from twisted.web.http_headers import Headers

agent = Agent(reactor)

d = agent.request(
    'GET',
    'http://localhost:8080',
    Headers({'User-Agent': ['Twisted Web Client Example']}),
    None)


def cbResponse(failure):
    print(failure)


d.addBoth(cbResponse)


def cbShutdown(ignored):
    reactor.stop()

d.addBoth(cbShutdown)

reactor.run()
```

Output is:

```
python client.py 
[Failure instance: Traceback (failure with no frames): <class 'twisted.web._newclient.ResponseFailed'>: [<twisted.python.failure.Failure exceptions.AttributeError: 'TransportProxyProducer' object has no attribute 'loseConnection'>]
```

<details><summary>Searchable metadata</summary>

```
trac-id__8570 8570
type__enhancement enhancement
reporter__pawelmhm pawelmhm
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__http__Agent__webclient http, Agent, webclient
time__1467896302835926 1467896302835926
changetime__1511740897981113 1511740897981113
version__None None
owner__None None

```
</details>

twisted-trac:
[...]

PR #11998, opened 2023-09-20
#11997 Fix for Twisted web client support of trailer Server-Timing

### LINKED ISSUE #11997: Twisted web client doesn't support trailer Server-Timing
Twisted web client doesn't support trailer Server-Timing.
Example: https://w3c.github.io/server-timing/#the-server-timing-header-field

Use console program (simple http client) from this gist: https://gist.github.com/taroved/06922ceeb872094304cc889dacae099f 

Always you run the code:
```shell
python httpclient.py https://www.esquire.com/
```
You see such exception:
```
Failure: twisted.web._newclient.ResponseFailed: [<twisted.python.failure.Failure twisted.web.http._MalformedChunkedDataError: Chunk did not end with CRLF>, <twisted.python.failure.Failure twisted.web.http._DataLoss: Chunked decoder in 'TRAILER' state, still expecting more data to get to 'FINISHED' state.>]
```

Such pages as https://www.esquire.com/ should be downloaded without exceptions.

**Testing environment**
 - Linux 939ff8bae3fe 5.15.49-linuxkit-pr
 - Twisted version: 23.8.0

taroved:
## Scope and purpose

Fixes #11997

Replaces simple check of trailer chars (CRLF) with additional check for trailer Server-Timing header like data.
Trailer Server-Timing should be handled without excepion. CURL or Browsers handle such data without errors.

Read more about trailer Server-Timing here: https://w3c.github.io/server-timing/

This fix will not have affect on efficiency.

## Contributor Checklist:


taroved:
please review

src/twisted/web/test/test_http.py
```diff
@@ -1469,6 +1469,21 @@ def finished(extra):
         self.assertEqual(errors, [])
         self.assertEqual(successes, [True])
 
+    def test_tailerServerTiming(self):
+        """
+        L{_ChunkedTransferDecoder.dataReceived} decodes chunked-encoded data
+        and correctly handle trailer Server-Timing data.
```
adiroiban: Can you please add more details about what correctly means?

We alwasy write tests to make sure the implementation is correct :) 

So it helps to skip the "correctly handles" part and add a few more details about the scope / purpose / expected behaviour in this test.

src/twisted/newsfragments/11997.bugfix
```diff
@@ -0,0 +1 @@
+fix for Twisted web client support of trailer Server-Timing
```
adiroiban: We are using this text as part of our public release notes.

Can you please write it as a full sentence and add a few info about how was  the previous behaviour and how it behaves how?

src/twisted/web/http.py
```diff
@@ -1998,10 +1999,14 @@ def _dataReceived_TRAILER(self) -> bool:
         if len(self._buffer) < 2:
             return False
 
-        if not self._buffer.startswith(b"\r\n"):
+        if self._buffer.startswith(b"\r\n"):
+            data = memoryview(self._buffer)[2:].tobytes()
+        # starts with case-insensitive trailer header prefix "Server-Timing:"
+        elif _startsWithServerTiming.match(self._buffer):
+            data = memoryview(self._buffer)[:].tobytes()
```
adiroiban: is this comment about the `elif _startsWithServerTiming.match(self._buffer):` branch?

If yes, I think is best to move it inside the branch and add a comment about why we have this special handling.

`_startsWithServerTiming` already tells everyting that is in the current comment 
 
So here it would help to have some info about why the code from this branch is different from the handling in the other branches

```suggestion
        elif _startsWithServerTiming.match(self._buffer):
            # For "Server-Timing" headers we DO AND THIS
            # BECAUSE OF THIS. 
            data = memoryview(self._buffer)[:].tobytes()
```

src/twisted/web/test/test_http.py
```diff
@@ -1469,6 +1469,21 @@ def finished(extra):
...
+        """
+        L = []
+        finished = []
+        p = http._ChunkedTransferDecoder(L.append, finished.append)
+        p.dataReceived(b"3\r\nabc\r\n5\r\n12345\r\n")
```
adiroiban, line 1468: To help with test, it might help to also run a separate test in which each byte is received in a separate TCP package.

So, as a testing "strategy"
run a test in which all the data is received in one TCP packet, by using a single `dataReceived` call and another test in which we iterate over the data and call `dataReceive` for each byte.

Here, I can see the data is received in 2 TCP packets.

adiroiban (commented):
Many thanks for your PR.
It looks good.

I left a few comments.
Most probable because I am not familiar with the spec.

I only did a quick read about this header at 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing

I haven't yet read the whole specification.

Also, I haven't executed a manual test yet to see the before and after part. I saw you mention a manual test in the initial ticket.

I will try to continue the review a bit later,
but I hope the current comments will help to get this ready for merge.

Thanks again

src/twisted/web/test/test_http.py
taroved: Server-Timing data cannot be split to parts because we don't have any information about the data size. So Server-Timing data should be in a single packet.

exarkun: This isn't how TCP works, though.  There is no guarantee about where packet boundaries will fall.  @adiroiban 's suggestion to write a test that delivers the data one byte at a time represents one extreme end of the *legal* TCP behavior - to split the stream into single byte packets and deliver them separately.

So there must be some framing information *somewhere* to use.  It's just part of the trailer isn't it?  So I would expect that normal trailer framing rules apply.

taroved: I find out that the topic is a bit wider. It is required to implement "trailing headers" (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer). I thought that trailing header can be only Server-Timing, but it can be almost any type. So I'm going to implement handling of any type of trailing header. And don't know should I give users access to such trailing headers. Or I can just ignore them?

adiroiban: Yes... so we have "HTTP chunks" and "TCP packets payload"

The trailing headers can only be delivered via the last HTTP chunk.

But in theory, the single HTTP chunk can be delivered via multiple TCP packets.

----------

In terms of API. `finishCallback` init argument of the `_ChunkedTransferDecoder` is used as a private API.
The value passed by `finishCallback` is handled by `HTTPChannel._finishRequestBody`  which is also a private API.

If I understand correctly, `finishCallback` returns bytes, which might be part of the next response.

Maybe we can update `finishCallback` to call with bytes + trailing headers.

Then `Response.deliverBody` takes an `IProtocol` for which on success the `IProtocol.connectionLost` is called with `t.internet._newclient.ResponseDone`
Maybe we can update `ResponseDone` to have `ResponseDone.trailerHeaders`

So you can read the trailer headers after the response was finalized.

This is just the low-level API... which might be ugly... but it's an idea.

----------

I guess that many implementations don't expect `ResponseDone` to have the trailier headers... so in practice, the headers are ignored.

But someone intersted in ready them, can add special handling.

Note that I'm not an expert in HTTP or HTTP client, so this is just a sugestion that I hope can help to move forward with this PR.

----------

At the same time, to reduce the scope of this PR,
I guess we can also have this PR updates to ignore any trailing headers, and just make sure the response doesn't fails.

We can them consider a separate PR to parse and expose the headers in some way.

Regards

src/twisted/newsfragments/11997.bugfix
```diff
@@ -0,0 +1 @@
+add web client support for trailer headers. Previous behaviour with throwing exception is replaced with collecting of trailer headers
```
twm:
```suggestion
twisted.web now ignores trailer headers rather than raising an exception.
```

src/twisted/web/test/test_http.py
```diff
@@ -1469,6 +1455,37 @@ def finished(extra):
         self.assertEqual(errors, [])
         self.assertEqual(successes, [True])
 
+    def test_tailerHeaders(self):
```
twm: Typo:

```suggestion
    def test_trailerHeaders(self):
```

src/twisted/web/http.py
```diff
@@ -1909,6 +1910,7 @@ def __init__(
         self.finishCallback = finishCallback
         self._buffer = bytearray()
         self._start = 0
+        self._trailerHeaders = []
```
twm: It looks like MyPy is [failing](
https://github.com/twisted/twisted/actions/runs/6276966953/job/17068965951?pr=11998) for want of an annotation here:

```
src/twisted/web/http.py:1913:9: error: Need type annotation for "_trailerHeaders" (hint: "_trailerHeaders: List[<type>] = ...")  [var-annotated]
```

twm (changes requested):
Awesome, thanks for diving in here!

Before we can merge you'll have to fix the missing type annotation. As you're not doing anything with the trailers yet perhaps drop the ivar for now? (If you want to keep it, please add a test that inspects its value!)

Otherwise this looks good to me! Please re-request review when everything's green.

src/twisted/web/test/test_http.py
taroved: changed in another commit

src/twisted/newsfragments/11997.bugfix
```diff
@@ -0,0 +1 @@
+twisted.web now ignores trailer headers rather than raising an exception.
```
adiroiban: I am trying to narrow the scope of this change.

Rather than the whole twisted.web , it would help to have a closer definition of what was changed.

I think that for once, this affects twisted.web.client.Agent

```suggestion
twisted.web.http.HTTPChannel now ignores the trailer headers provided in the last chunk of a chunked encoded response, rather than raising an exception.
```

src/twisted/web/http.py
```diff
@@ -1890,12 +1889,14 @@ class _ChunkedTransferDecoder:
...
-        arrives over many calls to C{dataReceived}.
+    @ivar _start: While in the C{'CHUNK_LENGTH'} and C{'TRAILER'} states,
+        tracks the index into the buffer at which search for CRLF should resume.
+        Resuming the search at this position avoids doing quadratic work if the
+        chunk length line arrives over many calls to C{dataReceived}. Not used
+        in any other state.
 
-        Not used in any other state.
+    @ivar _trailerHeaders: Accumulates trailer headers. Not passed outside the
+        decoder class for a while.
```
adiroiban: Maybe we can just document what it does.

This is private, so people should not use it.

I have created a follow ticket to look into exposing them.
and added a link.

So if anyone that wants to use this member, should look into defining the public API

```suggestion
    @ivar _trailerHeaders: Accumulates raw/unparsed trailer headers.
        See https://github.com/twisted/twisted/issues/12014
```

src/twisted/web/http.py
```diff
@@ -1985,23 +1987,31 @@ def _dataReceived_CRLF(self) -> bool:
...
+            # Restart the search upon receipt of more data at the start of the
+            # new data, minus one in case the last character of the buffer is
+            # CR.
+            self._start = len(self._buffer) - 1
             return False
 
-        if not self._buffer.startswith(b"\r\n"):
-            raise _MalformedChunkedDataError("Chunk did not end with CRLF")
+        if eolIndex > 0:
+            self._trailerHeaders.append(self._buffer[0:eolIndex])
```
adiroiban, line 2018: will this append happen without limits?

What would happen if a server returns an infinite stream of headers

```
\r\n0\r\n A LOT OF DATA WITHOUT the CRLFCRLF marker.
```

I guess that we will use all the memory

I think that we should add a limit ... maybe 64K of headers ?

I know that we have a similar limit for the HTTP headers.
So maybe use the same limit for trailing headers as for the normal headers.

adiroiban (changes requested):
Changes look good.

I am not very familiar with how Twisted web chunk encoding works, but I think that with this implementation we can cause a remote deny of sevice error, by consuming all the available memory

this can be done by a malicious HTTP client or HTTP server

adiroiban:
sorry for the late review. crazy days over here.

feel free to ping me or check over Gitter to request a review

Is my concern for accepting an infinite amount of trailing headers valid ?

src/twisted/web/http.py
taroved: Current headers limit is 16K. I used the same limit for trailer headers.

src/twisted/web/http.py
```diff
@@ -1985,23 +1996,38 @@ def _dataReceived_CRLF(self) -> bool:
...
```
adiroiban, line 2018: The current implementation is ok.

Maybe we can just check for size after this append.
In this way, we don't need the extra `self._receivedTrailerHeadersSize`

src/twisted/web/http.py
```diff
@@ -1890,12 +1889,21 @@ class _ChunkedTransferDecoder:
...
+    @ivar _start: While in the C{'CHUNK_LENGTH'} and C{'TRAILER'} states,
+        tracks the index into the buffer at which search for CRLF should resume.
+        Resuming the search at this position avoids doing quadratic work if the
+        chunk length line arrives over many calls to C{dataReceived}. Not used
+        in any other state.
+
+    @ivar _trailerHeaders: Accumulates raw/unparsed trailer headers.
+        See https://github.com/twisted/twisted/issues/12014
+
+    @ivar _totalTrailerHeadersSize: Maximum bytes for trailer header from the
```
adiroiban: Minor comment

Maybe call it "maximum" if this is about max limit :)

Otherwise, you might thing this is the current lenght.

```suggestion
    @ivar _maximumTrailerHeadersSize: Maximum bytes for trailer header from the
```

src/twisted/web/http.py
taroved: If we move the check and if the server response with infinite trailer header without CRLF then memory will be filled completely. So eolIndex will always equal to -1.

adiroiban: Thanks.

Can we have an autoamted test for this case?

It looks like the current automated tests is not covering this  case.

The current implementation might be ok, but the future, someone might refactor the code and the existing tests will still pass, while a bug is introduced.

------

I will have to check the code.
I still don't know why we can't just check the size of `self._buffer` and why we need an extra variable...
but maybe there is a good reason.

Note that the goal of raisin an exception is not to be matematically correct, but only to prevent a memory attack.

src/twisted/web/http.py
```diff
@@ -1985,23 +1996,38 @@ def _dataReceived_CRLF(self) -> bool:
...
-        if not self._buffer.startswith(b"\r\n"):
-            raise _MalformedChunkedDataError("Chunk did not end with CRLF")
+        if eolIndex > 0:
+            self._trailerHeaders.append(self._buffer[0:eolIndex])
+            del self._buffer[0 : eolIndex + 2]
+            self._start = 0
+            self._receivedTrailerHeadersSize += eolIndex + 2
+            return True
+
+        # eolIndex == 0
```
adiroiban: why is this line commented?

src/twisted/web/http.py
taroved: I added new unit test for unfinished trailer headers.

About current logic: in this extra variable we accumulate number of consumed bytes. When we meet CRLF characters we shrink buffer and increase this extra variable. How we would know how many characters we consumed without extra variable? Buffer self._buffer shrinks on every trailer header (every met CRLF chars)

src/twisted/web/http.py
taroved: This line just signals that eolIndex in this part of code is equal to 0. I think that it is not obviously. So I added this comment

src/twisted/web/http.py
taroved: So the extra variable solves 2 problems:
1. Monitor number of chars which was sent before CRLF. And if there is too many such characters algorithm throws exception.
2. If there is a lot of trailer headers the extra variable accumulates the lengths of this headers. `self._buffer` shrinks on every trailer header so we can't use `self._buffer` for detection of length of received data.

src/twisted/web/http.py
adiroiban: I think we need better comments.
You explanations from the PR comments are ok, but we need to have this as comment in the code.

Maybe someone else will read the code and will have the same questions.

src/twisted/web/http.py
```diff
@@ -1895,12 +1894,21 @@ class _ChunkedTransferDecoder:
...
 
-    @ivar _start: While in the C{'CHUNK_LENGTH'} state, tracks the index into
-        the buffer at which search for CRLF should resume. Resuming the search
-        at this position avoids doing quadratic work if the chunk length line
-        arrives over many calls to C{dataReceived}.
+    @ivar _start: While in the C{'CHUNK_LENGTH'} and C{'TRAILER'} states,
+        tracks the index into the buffer at which search for CRLF should resume.
+        Resuming the search at this position avoids doing quadratic work if the
+        chunk length line arrives over many calls to C{dataReceived}. Not used
+        in any other state.
```
adiroiban: Very minor comment.

Maybe other states will use it and forget to update the docstring.

So maybe this is better :) 
```suggestion
        chunk length line arrives over many calls to C{dataReceived}.
```

src/twisted/web/http.py
adiroiban: Thanks for the explanation and sorry for the delay.
The code logic seems complicated to me, and the current comments don't help much...but I didn't had the time to look into this and suggest an alternative.

It looks to me like the `self._start` thing is an optimization, not to search again over the part that was already checked.

I think that at this point we expect to have small buffers, so it should not have a big performance impact. 

--------

But we have automated tests, so it should be ok.
I am bit worried that `self._start` might be used somewhere else and end up with more trouble.

This is why I prefer not to optimize, if we don't have a big performance issue.

adiroiban (approved):
Sorry for the late reply.

I left a few minor comments.

I think that this is better than what we have in trunk and can be merge.

Thanks again.

src/twisted/web/http.py
```diff
@@ -1910,6 +1917,9 @@ def __init__(
         self.finishCallback = finishCallback
         self._buffer = bytearray()
         self._start = 0
+        self._trailerHeaders: List[bytearray] = []
+        self._maxTrailerHeadersSize = 16384
```
adiroiban: See https://github.com/twisted/twisted/issues/8570#issuecomment-1864682301

No need to update the value in this PR.

But I think that it would help to define this value as a private constant for now.

Later we can look at sharing the same constant across multiple places.

maybe something like this ?

```suggestion
        self._maxTrailerHeadersSize = 2 ** 16
```

src/twisted/web/http.py
```diff
@@ -1986,23 +1996,34 @@ def _dataReceived_CRLF(self) -> bool:
...
-        if len(self._buffer) < 2:
+        if (
+            self._receivedTrailerHeadersSize + len(self._buffer)
+            > self._maxTrailerHeadersSize
+        ):
+            raise _MalformedChunkedDataError("Trailer headers data is too long")
+
+        eolIndex = self._buffer.find(b"\r\n", self._start)
+
+        if eolIndex == -1:
```
adiroiban, line 2011: Maybe a comment.
```suggestion
        if eolIndex == -1:
            # Still no end of network line marker found.
            # Continue processing more data.
```

src/twisted/web/http.py
```diff
@@ -1986,23 +1996,34 @@ def _dataReceived_CRLF(self) -> bool:
...
-
-        @raises _MalformedChunkedDataError: when anything other than CRLF is
-            received.
         """
-        if len(self._buffer) < 2:
+        if (
+            self._receivedTrailerHeadersSize + len(self._buffer)
+            > self._maxTrailerHeadersSize
+        ):
+            raise _MalformedChunkedDataError("Trailer headers data is too long")
```
adiroiban: Minor comment. I think is best to have a full stop at the end of the sentence.

```suggestion
            raise _MalformedChunkedDataError("Trailer headers data is too long.")
```

src/twisted/web/http.py
```diff
@@ -1986,23 +1996,34 @@ def _dataReceived_CRLF(self) -> bool:
...
+
+        eolIndex = self._buffer.find(b"\r\n", self._start)
+
+        if eolIndex == -1:
             return False
 
-        if not self._buffer.startswith(b"\r\n"):
-            raise _MalformedChunkedDataError("Chunk did not end with CRLF")
+        if eolIndex > 0:
```
adiroiban, line 2016:
```suggestion
        if eolIndex > 0:
            # A trailer header was detected.
```

adiroiban (approved):
Sorry for the late review. It looks good. Only minor comments.

adiroiban:
The merge of this PR is blocked as @twm has requested changed. Tom, do you have time to check the latest changes and see your are happy with this PR? Thanks!

adiroiban:
Thanks Alexandr for the PR.
I plan to have a release soon ... if we don't get another review until then, I think it's ok to have this merged and released.

We can then get more feedback for this feature as people are using it.

I have tagged it as a "Release-blocker" as a reminder.

Regards

adiroiban:
@glyph I would like to do a new Twisted release soon .

I think that this PR is in good shape.

The blocking comment from Tom was about type annotation. As long as mypy is green, I think that we are file.

I will apply my suggestion, as they are only code comments and I will merge.

adiroiban:
I am merging this as I am preparing a new release.
Pypy pass, so I think that Tom's request was addressed.

Thanks for your help with this.

PR #12003, opened 2023-09-26
#12004 type annotate more tests

### LINKED ISSUE #12004: type annotate more tests
another batch of tests type annotated please

graingert:
## Scope and purpose

Fixes #12004

### DIFF
--- a/src/twisted/names/test/test_hosts.py
+++ b/src/twisted/names/test/test_hosts.py
@@ -4,7 +4,9 @@
 """
 Tests for the I{hosts(5)}-based resolver, L{twisted.names.hosts}.
 """
+from __future__ import annotations
 
+from typing_extensions import Protocol
 
 from twisted.internet.defer import gatherResults
 from twisted.names.dns import (
@@ -22,8 +24,13 @@ from twisted.python.filepath import FilePath
 from twisted.trial.unittest import SynchronousTestCase
 
 
+class _SupportsMktemp(Protocol):
+    def mktemp(self) -> str:
+        ...
+
+
 class GoodTempPathMixin:
-    def path(self):
+    def path(self: _SupportsMktemp) -> FilePath[bytes]:
         return FilePath(self.mktemp().encode("utf-8"))
 
 
--- a/src/twisted/test/test_error.py
+++ b/src/twisted/test/test_error.py
@@ -1,45 +1,62 @@
 # Copyright (c) Twisted Matrix Laboratories.
 # See LICENSE for details.
-
+from __future__ import annotations
 
 import errno
 import socket
+import sys
+from typing import Sequence
 
 from twisted.internet import error
-from twisted.python.runtime import platformType
 from twisted.trial import unittest
 
 
 class StringificationTests(unittest.SynchronousTestCase):
     """Test that the exceptions have useful stringifications."""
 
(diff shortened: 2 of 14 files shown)


src/twisted/names/test/test_hosts.py
```diff
@@ -22,8 +24,13 @@
 from twisted.trial.unittest import SynchronousTestCase
 
 
+class _SupportsMktemp(Protocol):
```
adiroiban, line 27: Just asking.

Is there a convention for naming single-method interfaces?

I think that for golang, this would be something like `_TempDirCreator`

I am ok with `_SupportsMktemp` no need to spend more time finding a better name here.

-------

same question for `class _HasException(Protocol):`

src/twisted/test/test_error.py
```diff
@@ -1,45 +1,62 @@
...
 from twisted.internet import error
-from twisted.python.runtime import platformType
 from twisted.trial import unittest
 
 
 class StringificationTests(unittest.SynchronousTestCase):
     """Test that the exceptions have useful stringifications."""
 
-    listOfTests = [
+    listOfTests: list[
```
adiroiban, line 17: just asking. Do we get any benefits by defining the type for this list? 
I would say that `list[object]` should be good enough.
This is just a testing helper.

What I am trying to do here, is see if we can have some sort of rule for relaxing the type annotation in tests.

src/twisted/test/test_error.py
```diff
@@ -216,28 +259,28 @@ def assertCorrectException(self, errno, message, result, expectedClass):
...
-    def test_errno(self):
+    def test_errno(self) -> None:
         """
         L{error.getConnectError} converts based on errno for C{socket.error}.
         """
         self.assertErrnoException(errno.ENETUNREACH, error.NoRouteError)
         self.assertErrnoException(errno.ECONNREFUSED, error.ConnectionRefusedError)
         self.assertErrnoException(errno.ETIMEDOUT, error.TCPTimedOutError)
-        if platformType == "win32":
+        if sys.platform == "win32":
```
adiroiban, line 269: Maybe we need type definition for platformType ?

Or if `t.python.runntime.Platform` is no longer relavant , maybe we can deprecate it and just use sys.platform.

adiroiban (approved):
Thanks. It looks good.

Only minor and general questions.

I have not looked into the details of why a specific `# type: ignore` is used.
I think is best to merge this as it is.
If someone wants better checks, we can have a separate PR.

Thanks again.

src/twisted/test/test_error.py
exarkun: String comparisons for platform detection are an anti-pattern.  `twisted.python.runtime.platform.isWindows()` is likely preferable here.

graingert: This needs to be sys.platform for the type checker to know we mean win32

src/twisted/names/test/test_hosts.py
graingert: I'm copying the naming convention from typeshed

src/twisted/test/test_error.py
graingert: the full type annotation is needed for def testThemAll to typecheck

PR #12013, opened 2023-10-11
#12012 Fix loading JQuery for Read The Docs theme.

### LINKED ISSUE #12012: Read The Docs documentation website search not working
You can test this link

https://docs.twisted.org/en/twisted-23.8.0/search.html?q=threads&check_keywords=yes&area=default#

it looks like jquery is not included

adiroiban:
## Scope and purpose

Fixes #12012

This tries to fix search page and RTD version navigation issue that was introduced with 23.8.0

## Issue description

You can see that search works for 22.10.0 https://docs.twisted.org/en/twisted-22.10.0/search.html?q=deferLater&check_keywords=yes&area=default

But this fails https://docs.twisted.org/en/twisted-23.8.0/search.html?q=deferLater&check_keywords=yes&area=default

Also clicking on the RTD version, does nothing

image

## Changes

Explicit use of the jquery extension.

I have updated the RTD sphinx theme and pydoctor versions... while on it.

As a drive-by, I have removed the Gitter embeeded chat box, as it's no longer available.
I have also removed the PDF code, since we don't build the PDF version of the docs.

## How to test

Check the search and RTD version navigation for the docs build generated by this PR

https://twisted--12013.org.readthedocs.build/en/12013/search.html?q=deferLater&check_keywords=yes&area=default

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,6 +20,8 @@ from pprint import pprint
 
 import sphinx_rtd_theme
 
+from twisted import version as twisted_version_object
+
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
@@ -37,17 +39,9 @@ sys.path.insert(0, os.path.abspath(".."))
 # ones.
 extensions = [
     "sphinx.ext.intersphinx",
+    "sphinxcontrib.jquery",
     "pydoctor.sphinx_ext.build_apidocs",
 ]
-
-try:
-    import rst2pdf.pdfbuilder
-
-    extensions.append("rst2pdf.pdfbuilder")
-except ImportError:
-    pass
-
-from twisted import version as twisted_version_object
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
@@ -105,14 +99,6 @@ html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
 
-
-# These paths are either relative to html_static_path
-# or fully qualified paths (eg. https://...)
-html_js_files = [
-    "js/custom.js",
-    # Here we have a Sphinx HTML injection hack to make the JS script load without blocking.
-    'https://sidecar.gitter.im/dist/sidecar.v1.js" defer hack="',
-]
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = "Twisteddoc"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,9 +61,8 @@ test = [
 # release scripts and process.
 dev-release = [
     "towncrier ~= 23.6",
-    "pydoctor ~= 23.4.0",
-    "sphinx-rtd-theme ~= 1.2",
-    "readthedocs-sphinx-ext ~= 2.2",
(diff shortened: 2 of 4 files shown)


adiroiban:
the test failure, looks like a flaky test

I think this is ready for review

needs-review

docs/conf.py
```diff
@@ -106,14 +100,6 @@
 html_static_path = ["_static"]
 
 
-# These paths are either relative to html_static_path
```
adiroiban, line 109: this is the gitter chat box that is no longer available... now that gitter was migrated to matrix

pyproject.toml
```diff
@@ -61,9 +61,8 @@ test = [
 # release scripts and process.
 dev-release = [
     "towncrier ~= 23.6",
-    "pydoctor ~= 23.4.0",
-    "sphinx-rtd-theme ~= 1.2",
-    "readthedocs-sphinx-ext ~= 2.2",
```
adiroiban, line 66: I am not sure why the sphinx-ext was pinned here.
We only need the `sphinx-rtd-theme` and it should pull the dependencies

docs/conf.py
glyph: Probably best to avoid this anyway

docs/conf.py
```diff
@@ -37,18 +39,10 @@
...
 extensions = [
     "sphinx.ext.intersphinx",
+    "sphinxcontrib.jquery",
     "pydoctor.sphinx_ext.build_apidocs",
 ]
 
-try:
-    import rst2pdf.pdfbuilder
-
-    extensions.append("rst2pdf.pdfbuilder")
```
glyph, line 46: huh, weird. what does this have to do with jquery?

pyproject.toml
glyph: why are we pinning *anything* here? we should have a requirements.txt with exact pins in it for CI; these should be pretty loose.

glyph (approved):
LGTM

docs/conf.py
adiroiban: I left the comment about this inside the PR description.

This was a drive-by cleanup.

> As a drive-by, I have removed the Gitter embeeded chat box, as it's no longer available.
I have also removed the PDF code, since we don't build the PDF version of the docs.

pyproject.toml
adiroiban: These are used for both CI and local dev.

I think it's important to pin the direct dependencies.

Also, I prefer to have all the dependendencies in the same pyproject.toml file, rather than having requirements.txt and then other deps inside readthedocs.yaml or GitHub Actions yaml

adiroiban:
Thanks for the review and the merge.

pyproject.toml
glyph: > These are used for both CI and local dev.
> 
> I think it's important to pin the direct dependencies.
> 
> Also, I prefer to have all the dependendencies in the same pyproject.toml file, rather than having requirements.txt and then other deps inside readthedocs.yaml or GitHub Actions yaml

I think that we should be using `pip-compile` to produce the pinned dependencies *from* the `pyproject.toml`, not manually duplicating them :)

adiroiban: I am not sure I got this.
These changes are done in `pyproject.toml`
Also, with the latest changes, we should now have only top-level dependencies declared.

-------

I have never used `pip-compile` . I am not against it.
But so far, I think that we are ok, and we other other issues, bigger than pinning the deps for the dev tools.

Issue #11750, opened 2022-11-07
[Task] Heads up for PEP 594 and support Python 3.13


nandgator:
## Ref: [PEP 594 – Removing dead batteries from the standard library](https://peps.python.org/pep-0594/)

**Is your feature request related to a problem? Please describe.**

PEP 594 is accepted. Deprecation warning is rolling out from Python 3.11.

```
====================================================================================== test session starts =======================================================================================
platform linux -- Python 3.11.0, pytest-7.2.0, pluggy-1.0.0
rootdir: ~/my-project
plugins: anyio-3.6.2, Faker-15.2.0
collected 5 items                                                                                                                                                                                

tests/test_utils.py .....                                                                                                                                                                  [100%]

======================================================================================== warnings summary ========================================================================================
../../../../../.cache/pypoetry/virtualenvs/venv-6o_5W4s2-py3.11/lib/python3.11/site-packages/twisted/web/http.py:103
  ~/.cache/pypoetry/virtualenvs/venv-6o_5W4s2-py3.11/lib/python3.11/site-packages/twisted/web/http.py:103: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    import cgi

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================== 5 passed, 1 warning in 2.16s ==================================================================================
```

**Describe the solution you'd like**

Twisted is using [`crypt`](https://github.com/twisted/twisted/search?q=%22import+crypt%22) [`spwd`](https://github.com/twisted/twisted/search?q=%22import+spwd%22) and [`cgi`](https://github.com/twisted/twisted/search?p=2&q=%22import+cgi%22). Create a task to clean up these libraries.

---------

Python 3.13.0 beta 1 is scheduled for 2024-05-07 with final release 3.13.0 2024-10-01

Todo for 3.13, listed in priority order

* [x] Remove cgi and html #12015
* [x] Remove cgi multipart parsing #12058
* [x] util.getPassword #12060
* [x] builtins.TypeError: '<' not supported between instances of 'NoneType' and 'int' - https://github.com/twisted/twisted/issues/12056
* [x] Update docstring formatting https://github.com/twisted/twisted/issues/12063
* [x] defer.inlineCallbacks exception traceback formating https://github.com/twisted/twisted/issues/12061
* [x] twisted.internet.utils https://github.com/twisted/twisted/issues/12062
[...]

PR #12016, opened 2023-10-12
#12015 Replace cgi.escape and cgi.parse_header

### LINKED ISSUE #12015: Replace cgi.escape with html.escape
the idea is to take the non controversial changes from https://github.com/twisted/twisted/issues/11919 and merge them in a separate PR

we can then also merge the controlversial changes... if we need to revert the merge for the controversial change, we can do that

adiroiban:
## Scope and purpose

Fixes #12015

This is the code  done by @graingert in https://github.com/twisted/twisted/pull/11918

but without the multipart parsing.

I think that the multipart part can be done in the separate PR and maybe coordinate with https://github.com/twisted/treq/issues/355

These changes from here are supposed to be non-controversion and be easy to merge.


adiroiban:
@twm and @graingert  If you have time, please take a look and see if this can be merged.

-------

The only thing left from CGI is multipart parsing.
And basically, this is the same questions as https://github.com/twisted/treq/issues/355

I would say that we can try to use stdlib Message to parse the multipart.

Colin suggested using `multipart` , but at this point, I would prefer to use stdlib and if needed send fixes upstream to Python

I hope we can take a decision as part of https://github.com/twisted/twisted/pull/11918

needs-review

nandgator:
Can someone please link issue #11750.

Closes #11750

graingert (approved):

graingert:
Lgtm

adiroiban:
Thanks. I have enabled auto-merge for this PR.

glyph:
amazing, thank you @adiroiban and @graingert .  glad we are finally free of this legacy dependency

glyph:
Also, just for posterity: any use of `html.escape` should *eventually* be replaced with `twisted.web.template`.  We have a whole templating system, we should use it :)

PR #12024, opened 2023-10-25
#12023 Add GitHub security policy

### LINKED ISSUE #12023: Update security policy link in GItHub.com
In the past, GitHub was happy to consider our `/docs/security.rst` as the security policy.

It looks like now, it wants `/SECURITY.md`

Since  we use GitHub for security bug reporting and fixing via private .

Also, add a note in the release process to make sure we alwasy have a CVE for any security bug.

adiroiban:
## Scope and purpose

Fixes #12023

Right now the main "Security" page for twisted/twisted looks like

image

We would like have the security info right away. something like electron

image

----------

In order to actual see the result of this PR, it needs to be merged

You can view the content of the file at https://github.com/twisted/twisted/blob/12023-github-security-policy/.SECURITY.md


adiroiban:
I left the file as /SECURITY.md to be as visible as possible 

For now SECURITY.md has the same content as docs/security.rst

Maybe /SECURITY.md can be brief, just an introduction and then link to our full security reporting page

What do you think

needs-review

docs/security.rst
```diff
@@ -1,11 +1,16 @@
...
 
+..
+  ATTENTION: For GitHub UI we also have /SECURITY.md file, which is a Markdown conversion of this file.
+  When making changes to this make, make sure to also update /SECURITY.md.
+
 We take security very seriously.
 Your input and feedback on our security is always appreciated.
 
-You can send urgent or sensitive reports directly to security@twistedmatrix.com
+You can send urgent or sensitive reports via `GitHub Security Advisory <https://github.com/twisted/twisted/security/advisories/new>`_.
```
adiroiban, line 11: I think that we should also add a big warning at the start of this document, to mention that any security PR should be done via private repos using the GitHub functionalty.


Otherwise, by creating a normal PR the bug is disclosed before the fix being ready :)

docs/security.rst
```diff
@@ -1,11 +1,16 @@
 Report a security issue
 =======================
 
+..
+  ATTENTION: For GitHub UI we also have /SECURITY.md file, which is a Markdown conversion of this file.
+  When making changes to this make, make sure to also update /SECURITY.md.
```
glyph:
```suggestion
  ATTENTION: This is the reStructuredText version of /.SECURITY.md. Any changes should also be made in the Markdown version.
```

glyph: This just corrects some grammatical errors and makes the construction parallel to the comment in `.SECURITY.md`

.SECURITY.md
```diff
@@ -0,0 +1,466 @@
+# Report a security issue
```
glyph: Should this file be `.SECURITY.md` or should it be `SECURITY.md`? The verbiage seems inconsistent with the filename.

adiroiban:
The file should be `SECURITY.md` 
I will try to get this fixed.

adiroiban:
The rename is done.

My suggetion is to keep SECURITY.MD brief.

Just inform that security reports should be done by clicking the button from top right.

Inform that no public ticket or public PR should be made for this issue, and that the GitHub process will handle the privacy.

Also provide a link to the security.rst full docs.

What do you think?

adiroiban:
As we try to receive money from Tidelift, for each CVE we will need to answer to these questions:

* Does this vulnerability still apply when the package is only used as a build tool or dev/test dependency? YES | NO
* Does this vulnerability apply only if certain methods, classes, or functionality are in use? YES | NO
* Please list the affected classes, methods, functionality: TEXTARE
* Are there any other conditions that users should check for to determine if they're vulnerable? YES|NO
* Please describe the conditions users should check for: TEXTAREA
* For users who are unable to upgrade to a supported release, is there a workaround available? YES | NO
* When using the package as intended, how likely is it that users are affected by this vulnerability? Scale 0 - 10 (0 not likely - 10 highly likely)
* Please explain: TEXTAREA
* Please select the releases affected by this vulnerability. CHECKBOXES

---------

I see that for previous CVEs (Twisted releases), someone from Tidelift has already answered these questions.

-----------

I think that the questions are reasonable, and we might want include them in the release notes for a security issue.

glyph:
> My suggetion is to keep SECURITY.MD brief.
> 
> Just inform that security reports should be done by clicking the button from top right.
> 
> Inform that no public ticket or public PR should be made for this issue, and that the GitHub process will handle the privacy.
> 
> Also provide a link to the security.rst full docs.
> 
> What do you think?

I'd say let's just land this since it seems mostly fine, and possibly do the clean-up you're suggesting in a follow-up.

glyph (approved):

adiroiban:
Thanks for the review. I have enabled the auto-merge.
There was a new flaky macos tests and I opened a new ticket to track it

PR #12027, opened 2023-10-28
#12026 Adjust to deprecation of 3-arg generator.throw()

### LINKED ISSUE #12026: twisted uses deprecated three-argument form of generator.throw
`src/twisted/python/failure.py` does `return g.throw(self.type, self.value, self.tb)`. This is deprecated in Python 3.12: https://github.com/python/cpython/issues/96348 , https://docs.python.org/3.12/whatsnew/3.12.html#deprecated . This is a particular problem for https://github.com/fedora-infra/fedora-messaging consumers - they print a deprecation warning about once a second.

Try using twisted with Python 3.12, in a way that hits this codepath (I'm not sure precisely how to do that other than "use a fedora-messaging consumer", sorry). You get a deprecation warning:

```
Oct 27 23:11:08 openqa-lab01.iad2.fedoraproject.org fedora-messaging[172628]: /usr/lib/python3.12/site-packages/twisted/python/failure.py:518: DeprecationWarning: the (type, exc, tb) signature of throw() is deprecated, use the single-arg signature instead.
Oct 27 23:11:08 openqa-lab01.iad2.fedoraproject.org fedora-messaging[172628]:   return g.throw(self.type, self.value, self.tb)
```

No deprecation warning.

**Testing environment**
Python 3.12 on Fedora 39, don't think anything else is relevant. The code is still using the deprecated form on current git trunk.

**Additional context**
I would send a pull request changing this, but I'm unclear on [whether it's safe to just use the modern one-argument form on Python 3.8](https://github.com/python/cpython/issues/96348#issuecomment-1783626480).

AdamWill:
In Python 3.12, the 3-arg signature of generator.throw() is deprecated, you're only supposed to use the 1-arg signature where you pass only an exception instance. I *think* this is the right thing to pass in this case.

## Scope and purpose

Fixes #12026 (use of a deprecated form of `generator.throw()` causes deprecation warnings on Python 3.12). Note we are not totally sure yet whether this is safe on Python 3.8: see [here](https://github.com/python/cpython/issues/96348#issuecomment-1783626480).

## Contributor Checklist: (done)

### DIFF
--- a/src/twisted/newsfragments/12026.bugfix
+++ b/src/twisted/newsfragments/12026.bugfix
@@ -1,0 +1,1 @@
+twisted.python.failure.Failure now throws exception for generators without triggering a deprecation warnings on Python 3.12.
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -516,7 +516,7 @@ class Failure(BaseException):
         """
         # Note that the actual magic to find the traceback information
         # is done in _findFailure.
-        return g.throw(self.type, self.value, self.tb)
+        return g.throw(self.value.with_traceback(self.tb))
 
     @classmethod
     def _findFailure(cls):


AdamWill:
gah. stupid comment cleanup. just a sec.

AdamWill:
https://github.com/twisted/twisted/pull/12028 is an alternative that just suppresses the warning.

AdamWill:
oh, note I *think* this is the right thing to pass, but *possibly* it should just be `self.value` - I wasn't 100% sure.

src/twisted/newsfragments/12026.bugfix
```diff
@@ -0,0 +1 @@
+adjust generator `throw()` call to avoid triggering deprecation warnings on Python 3.12
```
adiroiban: Maybe something like this ?

```suggestion
twisted.python.failure.Failure now throws exception for generators without triggering a deprecation warnings on Python 3.12.
```

src/twisted/python/failure.py
```diff
@@ -516,7 +516,7 @@ def throwExceptionIntoGenerator(self, g):
         """
         # Note that the actual magic to find the traceback information
         # is done in _findFailure.
-        return g.throw(self.type, self.value, self.tb)
+        return g.throw(self.value.with_traceback(self.tb))
```
adiroiban, line 519: We were already using this code in CI with Python 3.12

I guess that the warning was raised, but there was no check for it.

It would be nice to have an explicit test that no warnings are raised, on any Python version.

adiroiban (approved):
Changes looks good.

There were some CI failures, but I don't think they are related.

Retrying the CI.

It would be nice to have an automated test, to make sure we don't regress.


Thanks for the PR.

adiroiban:
@AdamWill based on the comment from #12028 does it mean that this PR lacks sufficient test coverage?

I can see that it uses the single argument version and the tests pass on 3.8 and I guess that this code is exercise in the python 3.8

Do we need #12028?

Regards

AdamWill:
well, I'm not an expert on twisted, just a distro guy who noticed this problem, so I don't know for sure how extensive the test suite is. if you're happy with the test suite's coverage, then I guess this one should be fine.

glyph:
> well, I'm not an expert on twisted, just a distro guy who noticed this problem, so I don't know for sure how extensive the test suite is. if you're happy with the test suite's coverage, then I guess this one should be fine.

The suite is unfortunately kinda hit-and-miss in some areas, but in core abstractions like `Deferred` it is fairly exhaustive.  I'd be confident that if the tests are passing for this, then it probably works fine.

Given how close to the core this change is, it might not be amiss to ping the mailing list and see if anyone could run their own test suites, just to be sure we haven't missed something before the next release. But that's really just belt-and-suspenders checking.

adiroiban:
Email sent https://mail.python.org/archives/list/twisted@python.org/thread/MJFZT7AC4CTA54U5KNSWP2CZQDBIFQZK/

We can wait 1 week and after that we can merge this.

Thanks Adam and Glyph for your feedback

glyph:
> We can wait 1 week and after that we can merge this.

I didn't see any complaints, shall we merge?

adiroiban:
My only minor complain about the merge is the text if the release notes.
The current text reads more like a commit message, rather than a public release notes.

AdamWill:
Rebased and updated the changelog text as requested by @adiroiban .

PR #12040, opened 2023-11-27
#9347 log method parens

### LINKED ISSUE #9347: twisted.logger doesn't execute methods despite documentation
|ElementalAlchemist's avatar| ElementalAlchemist reported|
|-|-|
|Trac ID|trac#9347|
|Type|defect|
|Created|2017-12-06 02:56:36Z|

It may come up that people want to log things based on data gotten from object methods, e.g.
```
self.ircd.log.debug("Disconnecting user {user.uuid} ({user.hostmask()}): {reason}", user=self, reason=reason)
```
where `user.hostmask()` is some sort of useful data.

According to the documentation (https://twistedmatrix.com/documents/16.1.1/core/howto/logger.html#format-strings), not only is this allowed, but there is an example showing that this is explicitly allowed.

However, attempting to log the above message results in Twisted logging something like this:
```
2017-12-05 20:52:26-0600 [-] Unable to format event {'log_namespace': 'txircd', 'log_level': <LogLevel=debug>, 'format': '%(log_legacy)s', 'log_logger': <Logger 'txircd'>, 'log_source': None, 'system': '-', 'reason': 'Registration timeout', 'user': <txircd.user.IRCUser instance at 0x7f0c092e8050>, 'time': 1512528746.643236, 'log_format': 'Disconnecting user {user.uuid} ({user.hostmask()}): {reason}', 'message': (), 'log_time': 1512528746.643236}: IRCUser instance has no attribute 'hostmask()'
```

Instead, in order to get the desired logging, you need to modify it so that the functions aren't run on objects in the string:
```
self.ircd.log.debug("Disconnecting user {user.uuid} ({hostmask()}): {reason}", user=self, hostmask=self.hostmask, reason=reason)
```

This logs as expected, but the need to change to this doesn't match up with what the documentation says.

Either object methods need to be allowed, or the documentation needs to be changed to disallow it (including modifying the example where an object method is explicitly used in this way).

```
trac-id__9347 9347
type__defect defect
reporter__ElementalAlchemist ElementalAlchemist
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__logger logger
keywords__None None
time__1512528996358368 1512528996358368
changetime__1513653408400714 1513653408400714
version__None None
owner__None None

```
</details>

glyph:
## Scope and purpose

Fixes #9347

### DIFF
--- a/src/twisted/logger/_format.py
+++ b/src/twisted/logger/_format.py
@@ -166,7 +166,7 @@ def formatEventAsClassicLogText(
     return eventText + "\n"
 
 
-def keycall(key: str, get: Callable[[str], Any]) -> PotentialCallWrapper:
+def keycall(key: str, getter: Callable[[str], Any]) -> PotentialCallWrapper:
     """
     Check to see if C{key} ends with parentheses ("C{()}"); if not, wrap up the
     result of C{get} in a L{PotentialCallWrapper}.  Otherwise, call the result
@@ -175,15 +175,15 @@ def keycall(key: str, get: Callable[[str], Any]) -> PotentialCallWrapper:
     @param key: The last dotted segment of a formatting key, as parsed by
         L{Formatter.vformat}, which may end in C{()}.
 
-    @param get: A function which takes a string and returns some other object,
-        to be formatted and stringified for a log.
+    @param getter: A function which takes a string and returns some other
+        object, to be formatted and stringified for a log.
 
     @return: A L{PotentialCallWrapper} that will wrap up the result to allow
         for subsequent usages of parens to defer execution to log-format time.
     """
     callit = key.endswith("()")
     realKey = key[:-2] if callit else key
-    value = get(realKey)
+    value = getter(realKey)
     if callit:
         value = value()
     return PotentialCallWrapper(value)
--- a/src/twisted/logger/_format.py
+++ b/src/twisted/logger/_format.py
@@ -6,6 +6,8 @@
 Tools for formatting logging events.
 """
 
+from __future__ import annotations
+
 from datetime import datetime as DateTime
 from typing import Any, Callable, Iterator, Mapping, Optional, Union, cast
 
@@ -164,6 +166,51 @@ def formatEventAsClassicLogText(
     return eventText + "\n"
 
 
+def keycall(key: str, get: Callable[[str], Any]) -> PotentialCallWrapper:
+    """
+    Check to see if C{key} ends with parentheses ("C{()}"); if not, wrap up the
+    result of C{get} in a L{PotentialCallWrapper}.  Otherwise, call the result
+    of C{get} first, before wrapping it up.
+
+    @param key: The last dotted segment of a formatting key, as parsed by
+        L{Formatter.vformat}, which may end in C{()}.
+
+    @param get: A function which takes a string and returns some other object,
+        to be formatted and stringified for a log.
+
+    @return: A L{PotentialCallWrapper} that will wrap up the result to allow
+        for subsequent usages of parens to defer execution to log-format time.
+    """
+    callit = key.endswith("()")
+    realKey = key[:-2] if callit else key
+    value = get(realKey)
+    if callit:
+        value = value()
+    return PotentialCallWrapper(value)
+
+
+class PotentialCallWrapper(object):
+    """
+    Object wrapper that wraps C{getattr()} so as to process call-parentheses
+    C{"()"} after a dotted attribute access.
+    """
+
+    def __init__(self, wrapped: object) -> None:
+        self._wrapped = wrapped
+
+    def __getattr__(self, name: str) -> object:
+        return keycall(name, self._wrapped.__getattribute__)
+
+    def __format__(self, format_spec: str) -> str:
+        return self._wrapped.__format__(format_spec)
+
+    def __repr__(self) -> str:
+        return self._wrapped.__repr__()
+
+    def __str__(self) -> str:
+        return self._wrapped.__str__()
+
+
 class CallMapping(Mapping[str, Any]):
     """
     Read-only mapping that turns a C{()}-suffix in key names into an invocation
@@ -190,12 +237,7 @@ class CallMapping(Mapping[str, Any]):
         Look up an item in the submapping for this L{CallMapping}, calling it
         if C{key} ends with C{"()"}.
         """
-        callit = key.endswith("()")
-        realKey = key[:-2] if callit else key
-        value = self._submapping[realKey]
-        if callit:
-            value = value()
-        return value
+        return keycall(key, self._submapping.__getitem__)
 
 
 def formatWithCall(formatString: str, mapping: Mapping[str, Any]) -> str:
--- a/src/twisted/logger/test/test_format.py
+++ b/src/twisted/logger/test/test_format.py
@@ -36,6 +36,12 @@ class FormattingTests(unittest.TestCase):
     Tests for basic event formatting functions.
     """
 
+    def format(self, logFormat: AnyStr, **event: object) -> str:
+        event["log_format"] = logFormat
+        result = formatEvent(event)
+        self.assertIs(type(result), str)
+        return result
+
     def test_formatEvent(self) -> None:
         """
         L{formatEvent} will format an event according to several rules:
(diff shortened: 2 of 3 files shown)


glyph:
please review

src/twisted/logger/test/test_format.py
```diff
@@ -36,6 +36,12 @@ class FormattingTests(unittest.TestCase):
     Tests for basic event formatting functions.
     """
 
+    def format(self, logFormat: AnyStr, **event: object) -> str:
+        event["log_format"] = logFormat
+        result = formatEvent(event)
+        self.assertIs(type(result), str)
+        return result
```
adiroiban: Maybe add a short docstring with the purpose of this function.


Also, minor comment, since we are here, maybe we can have `assertion(EXPECTED, ACTUAL)` as a convention for the assertions.

```suggestion
    def format(self, logFormat: AnyStr, **event: object) -> str:
        """
        Return the text rendering of C{event} based on the
        C{logFormat} template.
        
        This is a wrapper for L{format} to reduce code
        duplication in these tests.
        """
        event["log_format"] = logFormat
        result = formatEvent(event)
        self.assertIs(str, type(result))
        return result
```

src/twisted/logger/_format.py
```diff
@@ -164,6 +166,51 @@ def formatEventAsClassicLogText(
...
+
+    @param get: A function which takes a string and returns some other object,
+        to be formatted and stringified for a log.
+
+    @return: A L{PotentialCallWrapper} that will wrap up the result to allow
+        for subsequent usages of parens to defer execution to log-format time.
+    """
+    callit = key.endswith("()")
+    realKey = key[:-2] if callit else key
+    value = get(realKey)
```
adiroiban: can we have this renames as `getter` ?

src/twisted/logger/test/test_format.py
```diff
@@ -53,27 +59,36 @@ def test_formatEvent(self) -> None:
...
+        C{()} as methods.
+        """
+
+        class World:
+            def where(self) -> str:
+                return "world"
+
+        self.assertEqual(
+            "hello world", self.format("hello {what.where()}", what=World())
+        )
```
adiroiban, line 96: I don't see a separate test for this.

Do we want to allow acces to private stuff ?

this is simiar to this discussion https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

```suggestion
        class World:
            def where(self) -> str:
                return "world"
            def _private(self) -> str:
                return "private"


        self.assertEqual(
            "hello world",
            self.format("hello {what.where() what.__class__ what._private()}", what=World())
        )
```

src/twisted/logger/_format.py
```diff
@@ -164,6 +166,51 @@ def formatEventAsClassicLogText(
     return eventText + "\n"
 
 
+def keycall(key: str, get: Callable[[str], Any]) -> PotentialCallWrapper:
```
adiroiban: Not sure if this is a better name

```suggestion
def keyGetter(key: str, getter: Callable[[str], Any]) -> PotentialCallWrapper:
```

adiroiban (approved):
Thanks for the fix.
I think that this can be merged, as it's better than `trunk`.

The comments about naming stuff are minor.

I did a quick check of the current automated tests and I don't see any tests for accessing private members.

Not sure if we should allow or deny access to private members.

I would say that formatting should only allow acces to public members.

adiroiban:
It looks like there is no protection for private access

with the following diff

```
diff --git a/src/twisted/logger/test/test_format.py b/src/twisted/logger/test/test_format.py
index 209ccca7fd..dce73291b6 100644
--- a/src/twisted/logger/test/test_format.py
+++ b/src/twisted/logger/test/test_format.py
@@ -83,11 +83,15 @@ class FormattingTests(unittest.TestCase):
         """
 
         class World:
+            _priv = 'priv'
             def where(self) -> str:
                 return "world"
+            def _private(self) -> str:
+                return "private"
+
 
         self.assertEqual(
-            "hello world", self.format("hello {what.where()}", what=World())
+            "hello world", self.format("hello {what.where()} {what._priv.upper()} {what._private()}", what=World())
         )
```

I get this test fialure... and it looks like Twisted is happy to format the private members.

```diff
===============================================================================
[FAIL]
Traceback (most recent call last):
  File "/home/adi/chevah/twisted/src/twisted/logger/test/test_format.py", line 93, in test_formatMethod
    self.assertEqual(
  File "/home/adi/chevah/twisted/src/twisted/trial/_synctest.py", line 444, in assertEqual
    super().assertEqual(first, second, msg)
  File "/usr/lib/python3.11/unittest/case.py", line 873, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.11/unittest/case.py", line 1253, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
twisted.trial.unittest.FailTest: 'hello world' != 'hello world PRIV private'
- hello world
+ hello world PRIV private


twisted.logger.test.test_format.FormattingTests.test_formatMethod
```


Twisted log format is also happy to format `{what._priv.upper().upper()}` 

I think the formating should be less permisive.

As a start, don't allow access to private members.

src/twisted/logger/_format.py
glyph: It doesn't get the result though, it calls the result :)

glyph:
> It looks like there is no protection for private access

> I think the formating should be less permisive.
> 
> As a start, don't allow access to private members.

… why? There are at least four big problems with doing that:

1. We don't *enforce* private-member restrictions anywhere else in the code; it's a convention. So this would be inconsistent.
2. Private member restrictions would only make sense if we knew that the objects that you're logging from are not owned by the same person authoring the log format string. Quite possibly the object comes from the application that is logging about it, which is fine.
3. These format strings are designed to emulate the behavior of `f""` strings, or the built in `format()` or `str.format` operators, neither of which enforce private member restrictions. So it would be inconsistent and surprising.
4. Private members were not previously restricted so this would be a huge compatibility break. This only fixes our `()` formatting extension, private members were happily loggable before.

adiroiban:
Thanks Glyph for the response.

I am ok to have this merged.
I am not using the Twisted log format function for any user configurable log templates.

Also, maybe there aren't that many people using this code for user facing configuration.

adiroiban:
My concern about stdlib `.format` and also Twisted log format, is that it's too permisive.
and you end up with security issues like these https://github.com/lovasoa/pyformat-challenge

As long as you control the template, it's fine... but I guess that there might me people trying to use all kind of smart tricks with `.format()`

glyph:
> I am not using the Twisted log format function for any user configurable log templates.
> 
> Also, maybe there aren't that many people using this code for user facing configuration.

Yeah, it is explicitly a security failure for allowing "user-facing configuration" in log strings.  The idea is that they're supposed to be static and constant, always.

om26er:
This PR broke one of the existing ways of formatting
```
l = Logger()

class Shape(object):
    sides = 4
    name = "bamboozle"
    config = dict(foo='bar')

l.critical("{what.config[foo]} {what.sides} {what.name}", what=Shape())
```
ERROR is
```
Unable to format event {'what': <__main__.Shape object at 0x7f2d685d2410>, 'log_logger': <Logger '__main__'>, 'log_level': <LogLevel=critical>, 'log_namespace': '__main__', 'log_source': None, 'log_format': '{what.config[foo]} {what.sides} {what.name}', 'log_time': 1705018367.486725}: 'PotentialCallWrapper' object is not subscriptable
```

glyph:
@om26er ~~can you open a new bug, ideally a release blocker, if this hasn't been in a release yet?~~

(Did it myself, no need)

oberstet:
chiming in, thank you @om26er for early warning and @glyph for your clear statement and reaction! cool, IMO this is OSS as it "should be", working together, being good OSS citizens - because, yes, fallout (will) hit txaio/autobahn/crossbar, but I just recently argued for keeping Twisted _trunk_ (as well as multiple releases) in our CI since we do want to know early on about anything coming down the road with Twisted. We are (as you likely know) intensive users of Twisted and (still) love it;) It did break pretty new ground decades ago in Python and in general, paved the way, and is a broad spectrum, reliable weapon. I am still convinced (within the context of Python) even with aio and so present nowerdays .. anyways, cool=)

glyph:
Thanks for running trunk and catching issues like this before they make it to release!

PR #12045, opened 2023-11-30
#12044 no more urllib2

### LINKED ISSUE #12044: remove all references to `urllib2` and `cookielib`
It's been a decade or so since these modules were deprecated, and it would be good to clean up both our docs and our implementation to stop referring to them.

glyph:
## Scope and purpose

Fixes #12044

### DIFF
--- a/docs/core/howto/tutorial/intro.rst
+++ b/docs/core/howto/tutorial/intro.rst
@@ -395,28 +395,17 @@ server does.
 Read Status from the Web
 ------------------------
 
-
-
-The web. That invention which has infiltrated homes around the
-world finally gets through to our invention. In this case we use the
-built-in Twisted web client
-via ``twisted.web.client.getPage`` , a non-blocking version of
-Python's :func:`urllib2.urlopen(URL).read <urllib2.urlopen>` .
-Like ``getProcessOutput`` it returns a Deferred which will be
-called back with a string, and can thus be used as a drop-in
-replacement.
-
-
-
+The web. That invention which has infiltrated homes around the world finally
+gets through to our invention. In this case we use the built-in Twisted web
+client via ``twisted.web.client.getPage`` , a non-blocking version of Python's
+:func:`urllib.urlopen(URL).read <urllib.request.urlopen>` .  Like
+``getProcessOutput`` it returns a Deferred which will be called back with a
+string, and can thus be used as a drop-in replacement.
 
 Thus, we have examples of three different database back-ends, none of which
 change the protocol class. In fact, we will not have to change the protocol
 again until the end of this tutorial: we have achieved, here, one truly usable
 class.
-
-
-
-
 
 :download:`finger10.py <listings/finger/finger10.py>`
 
--- a/docs/web/howto/listings/client/cookies.py
+++ b/docs/web/howto/listings/client/cookies.py
@@ -1,5 +1,7 @@
+from http.cookiejar import CookieJar
+
 from twisted.internet import reactor
-from twisted.python import compat, log
+from twisted.python import log
 from twisted.web.client import Agent, CookieAgent
 
 
@@ -12,7 +14,7 @@ def displayCookies(response, cookieJar):
 
 
 def main():
-    cookieJar = compat.cookielib.CookieJar()
+    cookieJar = CookieJar()
     agent = CookieAgent(Agent(reactor), cookieJar)
 
     d = agent.request(b"GET", b"http://httpbin.org/cookies/set?some=data")
--- a/src/twisted/web/client.py
+++ b/src/twisted/web/client.py
@@ -1239,6 +1239,7 @@ class _FakeStdlibRequest(_RequestBase):
     uri: str
     type: str
     host: str
+    # The received headers managed using Twisted API.
     _twistedHeaders: Headers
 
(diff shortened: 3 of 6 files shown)


glyph:
please review

docs/core/howto/tutorial/intro.rst
```diff
@@ -395,29 +395,18 @@ server does.
...
-replacement.
-
-
-
+The web. That invention which has infiltrated homes around the world finally
+gets through to our invention. In this case we use the built-in Twisted web
+client via ``twisted.web.client.getPage`` , a non-blocking version of Python's
+:func:`urllib.urlopen(URL).read <urllib.request.urlopen>` .  Like
+``getProcessOutput`` it returns a Deferred which will be called back with a
+string, and can thus be used as a drop-in replacement.
```
adiroiban, line 403: This is ok. Just a minor comment.
Maybe when we reformat the lines, we can also update to
[semantic line breaks](https://sembr.org/).

docs/web/howto/listings/client/cookies.py
```diff
@@ -1,5 +1,7 @@
+from http.cookiejar import CookieJar
+
 from twisted.internet import reactor
-from twisted.python import compat, log
```
adiroiban, line 2: can this PR also remove the long deprecated compat alias?

https://github.com/twisted/twisted/blob/311d8dbaf40ddeb887cd3edab1333a3180e9d4d9/src/twisted/python/compat.py#L597-L602

I think that we are aready past 1 year and 2 releases

src/twisted/web/client.py
```diff
@@ -43,6 +47,17 @@
     IResponse,
 )
 
+# Allow for type-checking of our faked-out types.
```
adiroiban: Maybe something like this.

```suggestion
# For the purpose of type-checking we want our faked-out types to be identical to the types they are replacing.
# For the purpose of the impementation, we want to start
# with a blank slate so that we don't accidentally use
# any of the real implementation.
```

src/twisted/web/client.py
```diff
@@ -1200,36 +1215,37 @@ def request(self, method, uri, headers=None, bodyProducer=None):
...
     @ivar host: The host[:port] of the URI.
 
     @since: 11.1
     """
 
-    def __init__(self, uri):
+    uri: str
+    type: str
+    host: str
+    _twistedHeaders: Headers
```
adiroiban, line 1243: I think it would help to have a comment for the purpose of this variable.

```suggestion
    # The received headers managed using Twisted API.
    _twistedHeaders: Headers
```

src/twisted/web/client.py
```diff
@@ -1240,19 +1256,19 @@ def __init__(self, uri):
             self.host += ":" + str(_uri.port)
 
         self.origin_req_host = nativeString(_uri.host)
-        self.unverifiable = lambda _: False
```
adiroiban, line 1243: I guess that this was a bug in the previous implementation.

adiroiban (approved):
Many thanks for the cleanup. It looks good.
Only minor comments.

-------

I see some typing changes in twisted.web.client and twisted.web.test.test_agent.

Can we now remove the typing exceptions for these modules from `pyproject.toml` ?

-------

I think it is worth removing the deprecated compat alias in this PR, rather than creating another ticket and another PR and doing another review.


--------

In terms of coverage. [There is a missing branch coverage](https://app.codecov.io/gh/twisted/twisted/pull/12045?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=twisted), that was also missing in trunk

image

but I think we are fine


Thanks again

glyph:
> I see some typing changes in twisted.web.client and twisted.web.test.test_agent.
> 
> Can we now remove the typing exceptions for these modules from `pyproject.toml` ?

This will reduce the amount of work required to do that, but we're not quite finished with the necessary annotations, I don't think.

docs/core/howto/tutorial/intro.rst
glyph: I am going to skip that in this PR because I cannot figure out how to get my editor's paragaph wrapping (Emacs's fill-paragraph) to properly respect these line breaks, or have any kind of linting ensure them.  I would like to figure out a systematic way to do this though.

docs/web/howto/listings/client/cookies.py
glyph: I would like to separate out that work, in case it needs to be reverted for some reason, it's cleaner to have it as a separate single trunk merge commit.

src/twisted/web/client.py
glyph: Yeah, I'm not even really sure what the bug *was*. It seems like previously it would have been always unverifiable, even though it would be never-unverifiable based on runtime semantics.  Having fixed it to do what it was intending to do… no tests fail? Plus, `is_unverifiable` looks like it's never used?

adiroiban:
@glyph. thanks for the cleanup. I will try to make another release by the end of December.
For now, the goal is to have a new release every 2 months.

If I don't forget, I will try to create some cleanup PR before the release to remove any API that is deprecated since a long time

Issue #8214, opened 2016-02-19
Twisted should move to a src/ directory layout


twisted-trac:
|[<img alt="hawkowl's avatar" src="https://avatars.githubusercontent.com/u/3307100?s=50" width="50" height="50">](https://github.com/hawkowl)| @hawkowl reported|
|-|-|
|Trac ID|trac#8214|
|Type|enhancement|
|Created|2016-02-19 11:52:09Z|
|Branch|https://github.com/twisted/twisted/tree/src-move-8214-5|

This is current best practice (https://hynek.me/articles/testing-packaging/) and fixes quite a few issues:

- tox having to cd to a different directory to run the tests without running against the checkout accidentally (all of the ```toxrundir``` lines)
- reduce confusion as scripts will put stuff in the local checkout dir, instead of sometimes random dirs in ```build/``` (like ```.coverage``` files, or apidocs)
- allow us to have different packages in the repo that work in lockstep (see [#7945](https://github.com/twisted/twisted/issues/7945))

<details><summary>Searchable metadata</summary>

```
trac-id__8214 8214
type__enhancement enhancement
reporter__hawkowl hawkowl
priority__normal normal
milestone__None None
branch__src_move_8214_5 src-move-8214-5
branch_author__hawkowl hawkowl
status__closed closed
resolution__fixed fixed
component__core core
keywords__None None
time__1455882729084077 1455882729084077
changetime__1471480431447809 1471480431447809
version__None None
owner__hawkowl hawkowl

```
</details>

twisted-trac:
|[<img alt="hawkowl's avatar" src="https://avatars.githubusercontent.com/u/3307100?s=50" width="50" height="50">](https://github.com/hawkowl)<a name="note_1"></a>|@hawkowl commented|
|-|-|

(In [46887]) Branching to src-move-8214.

twisted-trac:
|[<img alt="hawkowl's avatar" src="https://avatars.githubusercontent.com/u/3307100?s=50" width="50" height="50">](https://github.com/hawkowl)<a name="note_2"></a>|@hawkowl commented|
|-|-|

This depends on [#7985](https://github.com/twisted/twisted/issues/7985) to make the py3 tests pass on an installed Twisted.

twisted-trac:
[...]

PR #648, opened 2016-12-27
Fix ssh examples on Python 3

rodrigc:
https://twistedmatrix.com/trac/ticket/8968

PR #401, opened 2016-07-24
[#8214] Move Twisted to a src layout

hawkowl:
https://twistedmatrix.com/trac/ticket/8214

PR #12047, opened 2023-12-01
#12046 Fix handling of function keys in conch.insults.window.Widget object

### LINKED ISSUE #12046: Conch widgets are ignoring function keys
`conch.insults.window.Widget` objects are ignoring function keys

Happens with any `Widget`, most probably caused in 2016 with these two commits:

6659c6b4f0290b83d5978e93ed8aff3d4878f649 changed how `FUNCTION_KEYS` are defined (from `const` instances to `bytes`) in `twisted.conch.insults.insults`
5d2ff6938cd1fcfce9e0861ae2c3a562c21b8156 changed how `Widget.functionKeyReceived` handles them in `twisted.conch.insults.window`

This is not covered by tests, so it slipped under radar. Looks like I'm first user in years. Even demos included in docs are bit rusty, so here is somewhat stripped gist -> [`window.tac`](https://gist.github.com/Menyadar/09e989c56b23f14386f3c8224b551c3d). Focus "Selection" widgets (there are two of them - columns with numbers, jump to them with `tab` key) and try to change the selection using arrow keys or page up/down.

Movement should just work

**Testing environment**
 - Operating System and Version; paste the output of these commands:
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"
```
 - Twisted version [e.g. 22.2.0]
```
$ twist --version
23.10.0.post0
```

**Additional context**
Fix is trivial, PR on the way

Menyadar:
## Scope and purpose

Fixes #12046

Included test is a bit naïve.

## Contributor Checklist:


Menyadar:
please review

src/twisted/conch/insults/window.py
```diff
@@ -109,7 +109,7 @@ def functionKeyReceived(self, keyID, modifier):
         name = keyID
         if not isinstance(keyID, str):
             name = name.decode("utf-8")
-        func = getattr(self, "func_" + name, None)
+        func = getattr(self, "func_" + name[1:-1], None)
```
adiroiban: So... this was introduced in https://github.com/twisted/twisted/pull/648

Before that PR, we had `func = getattr(self, 'func_' + keyID.name, None)`

--------

Why remove the `F` prefix?

This looks like breaking backward compatibility.

Previously the functions that were dispached were named `func_f12` ... with this change the need to be named `func_12` 

-----

I might be missing something here.

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +67,49 @@ def test_parent(self) -> None:
...
         self.assertIs(scrolled._viewport.parent, scrolled)
+
+
+class SelectionTests(TestCase):
+    """
+    Tests for L{Selection}, a widget which allows to select item from
+    list of items.
+    """
+
+    seq = [f"{x}".encode("ascii") for x in range(10)]
```
adiroiban: Can this be moved inside `test_selection` ?

Do they need to be class members ?

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +67,49 @@ def test_parent(self) -> None:
...
+    keys = {
```
adiroiban: Also, I think that this can be moved inside the test function?

Is this used in multiple tests ?

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +67,49 @@ def test_parent(self) -> None:
...
+
+    def test_defined_keynames(self) -> None:
+        """
+        Test if expected key names are still defined in t.c.i.i.FUNCTION_KEYS
+        """
+        self.assertTrue(set(self.keys.values()).issubset(set(FUNCTION_KEYS)))
+
+    def test_selection(self) -> None:
+        """
+        Test if sending function key codes actually changes focus
```
adiroiban: I think that this needs to be split into multiple tests, each test with a separate docstring.

The current test docstring doesn't contain much information about the scope and the expected behaviour.

If you have time, please take a look at  this page, it should be a quick read

https://jml.io/test-docstrings/

src/twisted/conch/newsfragments/12046.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conch.insults.window.Widget objects now correctly respond to function keys
```
adiroiban: Maybe it's to be specific.

I tried to write my understanding of this change.

But I think that the change is wrong, as this breaks backward compatibility.



```suggestion
twisted.conch.insults.window.Widget objects now will dispatch the function keys press events to function having the name prefixed by `func_` followed by the function number, without the `F` part.
```

adiroiban (changes requested):
Thanks for the PR.

I might be missing something here.

I don't understand the changes from `functionKeyReceived` and I don't see an automated test for `functionKeyReceived` 

Please check may inline comments.

Thanks again

src/twisted/conch/insults/window.py
Menyadar: It's not removing "prefix".  Before 6659c6b4f0290b83d5978e93ed8aff3d4878f649 `FUNCTION_KEYS` were defined as:
```python
_KEY_NAMES = ('UP_ARROW', 'DOWN_ARROW', 'RIGHT_ARROW', 'LEFT_ARROW',
              'HOME', 'INSERT', 'DELETE', 'END', 'PGUP', 'PGDN', 'NUMPAD_MIDDLE',
              'F1', 'F2', 'F3', 'F4', 'F5', 'F6', 'F7', 'F8', 'F9',
              'F10', 'F11', 'F12',
              'ALT', 'SHIFT', 'CONTROL')

FUNCTION_KEYS = [_const(_name) for _name in _KEY_NAMES]
```
these `const` instances were then assigned using their names to `insults.ServerProtocol` using
```python
for name, const in zip(_KEY_NAMES, FUNCTION_KEYS):
    setattr(ServerProtocol, name, const)
```

and `const` instance was passed down to `keystrokeReceived` when appropriate key was received, so (before 5d2ff6938cd1fcfce9e0861ae2c3a562c21b8156)  you could use
```python
func = getattr(self, "func_" + keyID.name, None)
```
and you ended up with proper function reference.

However AFTER 6659c6b4f0290b83d5978e93ed8aff3d4878f649 `FUNCTION_KEYS` are defined as
```python
FUNCTION_KEYS = [_const(_name).__bytes__() for _name in _KEY_NAMES]
```
and this means you have something like:
```python
[b'[UP_ARROW]', b'[DOWN_ARROW]', b'[RIGHT_ARROW]', b'[LEFT_ARROW]', b'[HOME]', ...]
```
this end up in `ServerProtocol`:
```python
>>> ServerProtocol.DOWN_ARROW
b'[DOWN_ARROW]'
```
and is passed down to `keystrokeReceived` (added prints to `ButtonDemo.keystrokeReceived` in `window.tac` to show this)
```
2023-12-01T13:51:55+0100 [stdout#info] keystrokeReceived: b'f' None
2023-12-01T13:51:57+0100 [stdout#info] keystrokeReceived: b'\t' None
2023-12-01T13:51:58+0100 [stdout#info] keystrokeReceived: b'\t' None
2023-12-01T13:51:59+0100 [stdout#info] keystrokeReceived: b'[DOWN_ARROW]' None
2023-12-01T13:52:00+0100 [stdout#info] keystrokeReceived: b'[UP_ARROW]' None
```

This all means now (after 5d2ff6938cd1fcfce9e0861ae2c3a562c21b8156) when `functionKeyReceived` is called with e.g. `b'[DOWN_ARROW]'`...
```python
def functionKeyReceived(self, keyID, modifier):
    name = keyID
    if not isinstance(keyID, str):
        name = name.decode("utf-8")
    func = getattr(self, "func_" + name, None)
    if func is not None:
        func(modifier)
```
... it won't (obviously) find any `func_[DOWN_ARROW]` method.

TL;DR: if something broke backwards compatibility, it was patch 6659c6b4f0290b83d5978e93ed8aff3d4878f649, as it changed the way `FUNCTION_KEYS` are defined, followed by 5d2ff6938cd1fcfce9e0861ae2c3a562c21b8156, where it was not noticed `keyID` is now in form of `b'[<whatever>]'`. My patch simply strips those brackets to get valid identifier.

src/twisted/conch/test/test_window.py
Menyadar: I've just used other docstrings / comments in the same file as a guide, will fix this

src/twisted/conch/test/test_window.py
Menyadar: `keys` is used in both `test_defined_keynames` and `test_selection`, `seq` just in `test_selection`

src/twisted/conch/newsfragments/12046.bugfix
Menyadar: Nope. No change in logic, just fix of state broken for 7 years. Now it works exactly how it was supposed to work (and how it did those 7 years ago, only using `bytes` now instead of `const`). You're free to reword it, though, as english isn't my native language.

src/twisted/conch/insults/window.py
Menyadar: > [...] I don't see an automated test for `functionKeyReceived`

Ok, i'll add this

src/twisted/conch/newsfragments/12046.bugfix
adiroiban: Thanks. My bad. Sorry for the noise.
Also, sorry for the old review for https://github.com/twisted/twisted/pull/401
It looks like I was the one approving the changes there.

For the release notes maybe

> twisted.conch.insults.window.Widget.functionKeyReceived fix a regression introduced with #8214 in Twisted 16.5.0

src/twisted/conch/test/test_window.py
Menyadar: > I think that this needs to be split into multiple tests, each test with a separate docstring.

Really? Most tests I see in `conch.test` are written in the same manner (testing functinality multiple values in single test)

src/twisted/conch/insults/window.py
Menyadar: Anyway this all 
```python
class _const:
    """
    @ivar name: A string naming this constant
    """

    def __init__(self, name: str) -> None:
        self.name = name

    def __repr__(self) -> str:
        return "[" + self.name + "]"

    def __bytes__(self) -> bytes:
        return ("[" + self.name + "]").encode("ascii")


FUNCTION_KEYS = [_const(_name).__bytes__() for _name in _KEY_NAMES]
```
in [twisted.conch.insults.insults.py#L435](https://github.com/twisted/twisted/blob/311d8dbaf40ddeb887cd3edab1333a3180e9d4d9/src/twisted/conch/insults/insults.py#L435) is just superfluous, as the only use of it now is to convert `"THIS"` to `b"[THIS]"` and could be replaced with:

```python
FUNCTION_KEYS = [f"[{x}]".encode("ascii") for x in _KEY_NAMES]
```
I'll extend this PR to include this change.

src/twisted/conch/test/test_window.py
adiroiban: I don't remember where I read or where it is document... I guess this is common TDD procedure.

I think the best practice is to have a single use case / functionaly per test.

That is only test a single feature for the system under test.

That is, each test case should have a single call to `functionKeyReceived` and should have just a few assertions that are related.

And the docstring of each test can explain the purpose of the test.

Menyadar:
please review

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,123 @@ def test_parent(self) -> None:
...
+    def test_selection_pgup(self) -> None:
+        """
+        Send PGUP to select element one page up (here: first element)
+        """
+        self.widget.keystrokeReceived(ServerProtocol.PGUP, None)  # type: ignore[attr-defined]
+        self.assertIs(self.widget.focusedIndex, 0)
+
+
+class TestWidget(Widget):
+    triggered: dict[str, bool] = dict()
```
adiroiban: Minor comment
```suggestion
class RecordingWidget(Widget):
    """
    A dummy Widget implementation to support the testing
    by recording the the triggered functional keys.
    """
    triggered: dict[str, bool] = dict()
```

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,123 @@ def test_parent(self) -> None:
...
+
+    def func_F1(self, modifier) -> None:  # type: ignore[no-untyped-def]
+        self.triggered["F1"] = True
```
adiroiban: I think that is important to record each call... in the case there is a bug and the same function is called more than once.
```suggestion
        self.triggered["F1"].append(True)
```

src/twisted/conch/test/test_window.py
adiroiban: To detect the case in which the function is accidentally triggered multiple times for a single event.

```suggestion
    triggered: dict[str, list[bool]] = {
       "F1": [],
       "HOME": [],
       "DOWN_ARROW": [],
       ... etc ..
    }
```

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,123 @@ def test_parent(self) -> None:
...
+        Make dummy widget to receive function keys
+        """
+        self.widget = TestWidget()
+
+    def test_widget_function_key_f1(self) -> None:
+        """
+        Widget receives F1 key
+        """
+        self.widget.functionKeyReceived(ServerProtocol.F1, None)  # type: ignore[attr-defined]
+        self.assertTrue(self.widget.triggered["F1"])
```
adiroiban: Minor comment. Not sure if setUp is needed here.
And we can have and independent test in arrange, act, assertion format.

Also, full stop at the end of the docstring sentence might help to confirm that the docstring is complete, and that it was not accidentally left unfishished.

```suggestion
    def test_widget_function_key_f1(self) -> None:
        """
        Widget receives F1 key.
        """
        widget = TestWidget()

        widget.functionKeyReceived(ServerProtocol.F1, None)  # type: ignore[attr-defined]

        self.assertEqual([True], self.widget.triggered["F1"])
```

src/twisted/conch/newsfragments/12046.bugfix
```diff
@@ -0,0 +1 @@
+twisted.conch.insults.window.Widget.functionKeyReceived fix a regression introduced with #8214 in Twisted 16.5.0
```
adiroiban: Maybe this summarize better this change

```suggestion
twisted.conch.insults.window.Widget.functionKeyReceived now dispatched functional key events to method with the name `func_KEYNAME`, where `KEYNAME` can be `F1`, `F2`, or `UP_ARROW`. This is a regression introduced with #8214 in Twisted 16.5.0, in which `KEYNAME` was handled with square brackets like `[F1]`.
```

adiroiban (approved):
Many thanks for the changes.

I think that this is much better than what we have in `trunk`.

I left a few minor comments.
It would be nice to have them before merge, but if you don't have time to work on this, that's ok.

Thanks again.

src/twisted/conch/test/test_window.py
exarkun: All `TestWidget` instances share the same `triggered` dictionary, reducing isolation between tests and probably generally being confusing.  Each should have its own.

src/twisted/conch/newsfragments/12046.bugfix
Menyadar: TL;DR: I think this begs for deeper refactoring and this PR fixes just the regression

I have the rest of updates prepared, however reading this made me to re-read whole part in `insults` to check how it is done. Now i really think moving from `const` objects to plain `bytes` in Twisted 16.5.0 when handling special keys was really a step back and proper way to fix this would be to reintroduce `const` objects back. It just feels wrong when i have to blindly "unpack" some "constant" from string value.

1. List of function keys "names" (what was once a source for name part of the constant) is considered implementation detail in `insults.insults._KEY_NAMES` and even there it was considered as a bad placement with following comment:

   ```python
   # XXX TODO - These attributes are really part of the
   # ITerminalTransport interface, I think.
   ```
2. My test now "randomly" uses some keys, because it'd be lot of copy&paste to test them all. Even then, you can't be sure if that's all of them, because there is no single source of truth for them.
3. Setting keys as attributes directly on `insults.ServerProtocol` is also strange, counterintuitive and potential source of weird bugs. Should be in it's own namespace and then you could list them easily from anywhere.
4. Same list of key names is used on two other places — `insults.helper` and `conch.recvline`; these places are identical and weird on their own, as bytes are concatenated to form a string that's converted to plain `str` and then put through `exec()`, just to introduce new variable based on the name and having plain `object()` as value. Original `exec()` code dates back 19 years (shouldn't be excuse to use `exec` even then), however useless `bytes` → `str` conversion is far newer. Simple `vars()["F1"] = object()` would be enough (or `locals()` / `globals()`).

adiroiban: Aggree. The keyname  part is strange and confusing. Thanks for the info.
With this bug, it looks like since 2016 nobody is using this code.
Otherwise I was expecting to see this bug report sooner.

So I think that it safe to make backward incompatible changes, if you find it easier to use this code.

I am not using this code and don't plan to use it. Just trying to help with a review here.

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,134 @@ def test_parent(self) -> None:
...
+        self.assertIs(self.widget.focusedIndex, 0)
+
+
+class RecordingWidget(Widget):
+    """
+    A dummy Widget implementation to test handling of function keys by
+    recording keyReceived events.
+    """
+
+    triggered: dict[str, list[bool]] = {
```
adiroiban: Jean-Paul is right(tm) :) This needs to be an instance member, not a class member.

Please create it in `__init__`

adiroiban (dismissed):
Sorry for blocking this.

I think that the test needs to be fixed before the merge.

Other than that, we can merge this and I am looking forward to your future PRs :)

Thanks again

src/twisted/conch/test/test_window.py
Menyadar: it is instance of plain list now, as methods append their respective key value

src/twisted/conch/newsfragments/12046.bugfix
Menyadar: This code has so few examples and practically zero docs i suspect there is nobody using it (here 'it' = insults). I just tried is as i knew it was there (first experimenting with it when it first came out ~19 years ago), but i will probably use something else in the end.

Menyadar:
please review

src/twisted/conch/insults/insults.py
```diff
@@ -431,23 +431,7 @@ def log(s):
...
-
-    def __repr__(self) -> str:
-        return "[" + self.name + "]"
-
-    def __bytes__(self) -> bytes:
-        return ("[" + self.name + "]").encode("ascii")
-
-
-FUNCTION_KEYS = [_const(_name).__bytes__() for _name in _KEY_NAMES]
+FUNCTION_KEYS = [f"[{_name}]".encode("ascii") for _name in _KEY_NAMES]
```
glyph, line 434: I am going to skip over the compatibility concern here since I'm assuming the other reviewers did some due diligence here (and I see there was already a regression, so we are in uncomfortable territory).  In general though this type of type-change on a public attribute would need to be accompanied by some kind of transition period.

src/twisted/conch/insults/window.py
glyph: A comment here saying "peel off square brackets added by `twisted.conch.insults.insults.FUNCTION_KEYS`" would help explain what's going on here.

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,129 @@ def test_parent(self) -> None:
...
+    def setUp(self) -> None:
+        """
+        Create L{ScrolledArea} widget with 10 elements and position selection to 5th element.
+        """
+        seq: list[bytes] = [f"{_num}".encode("ascii") for _num in range(10)]
+        self.widget = Selection(seq, None)
+        self.widget.height = 10  # type: ignore[assignment]
+        self.widget.focusedIndex = 5
+
+    def test_selection_down_arrow(self) -> None:
```
glyph: This is a violation of the Twisted test naming convention, which is `test_selectionDownArrow`. I realize that this departs from PEP-8 so it's a common mistake :)

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,129 @@ def test_parent(self) -> None:
...
+    """
+    A dummy Widget implementation to test handling of function keys by
+    recording keyReceived events.
+    """
+
+    def __init__(self) -> None:
+        Widget.__init__(self)
+        self.triggered: list[str] = []
+
+    def func_F1(self, modifier) -> None:  # type: ignore[no-untyped-def]
```
glyph: this… does not seem to be a valid use of `no-untyped-def`. We know that `modifier` must be a `str`?

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,129 @@ def test_parent(self) -> None:
...
+    Change focused entry in L{Selection} using function keys.
+    """
+
+    def setUp(self) -> None:
+        """
+        Create L{ScrolledArea} widget with 10 elements and position selection to 5th element.
+        """
+        seq: list[bytes] = [f"{_num}".encode("ascii") for _num in range(10)]
+        self.widget = Selection(seq, None)
+        self.widget.height = 10  # type: ignore[assignment]
```
glyph: we really ought to specify the height attribute, if it's part of the public API like this, rather than ignoring it out

src/twisted/conch/test/test_window.py
```diff
@@ -66,3 +73,129 @@ def test_parent(self) -> None:
...
+        self.widget.focusedIndex = 5
+
+    def test_selection_down_arrow(self) -> None:
+        """
+        Send DOWN_ARROW to select element just below the current one.
+        """
+        self.widget.keystrokeReceived(ServerProtocol.DOWN_ARROW, None)  # type: ignore[attr-defined]
```
glyph, line 96: So, for adding a single attribute like `height` I think it's worth fixing up within the PR, but... for addressing an architectural issue like the absence of these attributes from the type checker's visibility, let's not block, but let's file a new issue.

glyph (approved):
Thank you for your tenacity on this bugfix. Since the feedback I have is fairly trivial, I'm going to address it myself and land.  I've written it up here just for future reference.

src/twisted/conch/test/test_window.py
glyph: Filed https://github.com/twisted/twisted/issues/12110

glyph:
Thanks again @Menyadar, and congrats on getting this all the way through the process!

PR #12049, opened 2023-12-02
#12049 Use valid "clock" in TLSMemoryBIOFactory

### LINKED ISSUE #12048: After upgrade from 22.10.0 to 23.10.0 HTTP server does not reply when using with SSL and keep-alive
Client sent request, server receive, process the request, but respose is not sent to the client.

Our app work fine with Twisted 22.10.0 but does not work after upgrade to 23.10.0,
it is needed to have HTTPS with keep-alive then it happen all the time.

Preferable a [Short, Self Contained, Correct (Compilable), Example](http://www.sscce.org/) on a branch or on [a gist](https://gist.github.com).

Not ready yet, I will try to prepare minimal test to expose the problem (probably on Monday).

Response, even if it's short, should be sent to client.

**Testing environment**
 - Operating System and Version:
   - on any Windows version that our tests are running: Win2012, Win2016 and Win2019
 - Twisted version: 23.10.0
 - Reactor iocp

**Additional context**
My investigation pointed to introduced buffering for TLS, disable it cause our app is working again:
```
 if is_windows:
     from twisted.internet.iocpreactor.reactor import IOCPReactor as Reactor
+    from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
+
+    # disable buffering protocol as it does not work with SSL and keep-alive
+    TLSMemoryBIOFactory.protocol = TLSMemoryBIOProtocol  # type: ignore
```

ddzialak:
## Scope and purpose

Fixes #12048

Without this change, using IOCPReactor with TLS and Keep-Alive
will cause that response from server will be cached and not sent to the client.

## Contributor Checklist:

### DIFF
--- a/src/twisted/newsfragments/12049.bugfix
+++ b/src/twisted/newsfragments/12049.bugfix
@@ -1,0 +1,1 @@
+Fixed IOCPReactor when using with TLS and Keep-Alive option.
--- a/src/twisted/internet/iocpreactor/reactor.py
+++ b/src/twisted/internet/iocpreactor/reactor.py
@@ -168,7 +168,7 @@ class IOCPReactor(base.ReactorBase, _ThreadedWin32EventsMixin):
             """
             port = self.listenTCP(
                 port,
-                TLSMemoryBIOFactory(contextFactory, False, factory),
+                TLSMemoryBIOFactory(contextFactory, False, factory, clock=self),
                 backlog,
                 interface,
             )
@@ -184,7 +184,7 @@ class IOCPReactor(base.ReactorBase, _ThreadedWin32EventsMixin):
             return self.connectTCP(
                 host,
                 port,
-                TLSMemoryBIOFactory(contextFactory, True, factory),
+                TLSMemoryBIOFactory(contextFactory, True, factory, clock=self),
                 timeout,
                 bindAddress,
             )


ddzialak:
please review

adiroiban:
Thanks for the PR.

I think that we have a bigger issue here.
If no clock is passed to the  TLS factory, it should just use the [default reactor](https://github.com/twisted/twisted/pull/11996/files#diff-ecb5e92aca8ada72d70a04a7cdbe5743ff813c92dc24f69efae23c33bbc7a780R752).

```
```

-------

Do you have multiple reactors installed on your application ?

If somewhere later from your code you call `from twisted.internet import reactor` do you get the IOCP reactor of the select reactor?

src/twisted/newsfragments/12049.bugfix
```diff
@@ -0,0 +1 @@
+Fixed IOCPReactor when using with TLS and Keep-Alive option.
```
adiroiban, line 1: I think that the fix and the root cause is more generic. This not only about TLS and keep-alive.

I still trying to understand what is going in here... but it looks like in the `trunk` branch `TLSMemoryBIOFactory` is using the default reactor... and somehow, that reactor is not the same reactor that you are using for your HTTP application.

I am not sure.

I self contained exemple would help a lot to find the root cause of this issue.

Thanks

ddzialak: Probably it is a consequence of that default: 
https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/default.py#L49

In our code we import only one reactor: 
```
if is_windows:
    from twisted.internet.iocpreactor.reactor import IOCPReactor as Reactor
elif is_cygwin:
    # Beware, select is currently limited to 512 sockets, see:
    # https://bugs.python.org/issue28708
    from twisted.internet.selectreactor import SelectReactor as Reactor
else:
    from twisted.internet.epollreactor import EPollReactor as Reactor  # type: ignore

```
BUT we have defined our class:
```
from twisted.web.server import NOT_DONE_YET, Site

class _SfSite(Site):
```
and adding raising exception in selectreactor give traceback:
```
  File "c:\tmp\darek-testing\repo\starfish\starfish\src\sfutils\service\twisted_app_server.py", line 230, in handle_requests
    site = _SfSite(
           ^^^^^^^^
  File "c:\tmp\darek-testing\repo\starfish\starfish\src\sfutils\service\twisted_app_server.py", line 153, in __init__
    super().__init__(resource, **kwargs)
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\web\server.py", line 814, in __init__
    super().__init__(*args, **kwargs)
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\web\http.py", line 3226, in __init__
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\internet\reactor.py", line 38, in <module>
    from twisted.internet import default
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\internet\default.py", line 55, in <module>
    install = _getInstallFunction(platform)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\internet\default.py", line 49, in _getInstallFunction
    from twisted.internet.selectreactor import install
  File "C:\tmp\darek-testing\venv\Lib\site-packages\twisted\internet\selectreactor.py", line 199, in <module>
    raise AssertionError("OPS")

```

adiroiban: it's ok to have a custom site.

Just to make sure
So, for the reactor, I guess that you have something like this at the very start of your app

```python
from twisted.internet import iocpreactor
iocpreactor.install()

# Then everywhere else:
```

The documentation is here ... not sure if the docs are good enough

https://docs.twisted.org/en/stable/core/howto/choosing-reactor.html#core-howto-choosing-reactor-win32-iocp

----------

We also need an automated test to make sure we will never reintroduce this bug, ever again :)

ddzialak: We don't have `iocpreactor.install()` because we have our custom Reactor that is a subclass of `IOCPReactor` and it wasn't clear that "install" is needed. Anyway, as I see these "install()" functions just create reactor instance and invoke `main.installReactor(reactor)`.
I've verified that after invoking "main.installReactor(reactor)" problem also disappeared so I will go for that, closing that issue.

adiroiban: Happy to hear this was fixed.

Intall part is required, even for custome reactors.

This is what makes a singleton for the default reactor.

PR #12058, opened 2023-12-20
#11848 Remove the usage of cgi.parse_multipart and replace with email module.

### LINKED ISSUE #11848: DeprecationWarning: "'cgi' is deprecated and slated for removal in Python 3.13"
The currently used package "cgi" will be removed in Python 3.13.

```
> python -W always
Python 3.11.0 (main, Feb  3 2023, 12:28:31) [MSC v.1929 64 bit (AMD64)] on win32
>>> from twisted.web import http
...\site-packages\twisted\web\http.py:103: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
  import cgi
```

No warning.

**Testing environment**
 - Windows: 11
 - Python: 3.11.0
 - Twisted: 22.10.0

adiroiban:
## Scope and purpose

Fixes #11848

This should be the last usage of `cgi` module in Twisted.

It uses the stdlib `email` module.

I with with stdlib to reduce the need for an extra dependency or reinvent the wheel.

I think that if there are any issued with stdlib, they should be fixed upstream.

See https://github.com/twisted/treq/issues/355 for other options.

### DIFF
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -101,13 +101,13 @@ __all__ = [
 import base64
 import binascii
 import calendar
-import cgi
 import math
 import os
 import re
 import tempfile
 import time
 import warnings
+from email import message_from_bytes
 from email.message import EmailMessage
 from io import BytesIO
 from typing import AnyStr, Callable, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


src/twisted/web/http.py
```diff
@@ -272,6 +294,9 @@ def parse_qs(qs, keep_blank_values=0, strict_parsing=0):
     """
     Like C{cgi.parse_qs}, but with support for parsing byte strings on Python 3.
 
+    This was created to help with Python 2 to Python 3 migration.
```
adiroiban, line 297: I have added this comment to trigger a conversation.

Maybe this function should be deprecated (in a separate PR), and just use stdlib.

adiroiban:
I did a quick fix. The tests pass... but I am not sure this is the right fix.

needs-review

src/twisted/web/http.py
glyph: Could we just replace it with a deprecated alias of `parse_qs`, or does its behavior differ significantly?

In any case yeah eventually deprecating it sounds good.

glyph (approved):
really glad to see the process working and advance builds with alphas causing us to catch up with deprecations

PR #12064, opened 2023-12-21
#12063 Deprecation helper function docstring handling consistent with Python 3.13

### LINKED ISSUE #12063: Update docstring handling for Python 3.13
In Python 3.13 the docstring are trimmed by default.

We need to update some of the deprecation helpers to work with this.

adiroiban:
## Scope and purpose

Fixes #12063

In Python 3.13 the docstrings are stripped.

This causes some errors for our deprecation helpers.

Update the code to have the same behaviour to always strip and handle like Python 3.13.

---------

From 3.13 release notes

> Compiler now strip indents from docstrings. This will reduce the size of [bytecode cache](https://docs.python.org/3.13/glossary.html#term-bytecode) (e.g. .pyc file). For example, cache file size for sqlalchemy.orm.session in SQLAlchemy 2.0 is reduced by about 5%. This change will affect tools using docstrings, like [doctest](https://docs.python.org/3.13/library/doctest.html#module-doctest). (Contributed by Inada Naoki in [gh-81283](https://github.com/python/cpython/issues/81283).)


adiroiban:
I am sending this for review as an initial feedback.

Not sure if for 3.13 we should handle the docstring stripped, while for older Python version keep the non-strip version.

I went with always strip for "future" compatibility.

needs-review

src/twisted/newsfragments/12063.removal
```diff
@@ -0,0 +1,2 @@
+twisted.python.deprecate helper function will now always strip the docstrings.
```
itamarst: Maybe "strip whitespace from the docstrings".

src/twisted/python/deprecate.py
```diff
@@ -258,8 +258,14 @@ def _appendToDocstring(thingWithDoc, textToAppend):
     elif len(docstringLines) == 1:
         docstringLines.extend(["", textToAppend, ""])
     else:
-        spaces = docstringLines.pop()
+        trailer = docstringLines[-1]
+        spaces = ""
+        if not trailer.strip():
+            # On Python 3.13 the docstring is already stripped.
+            # For older Python version we keep the trailer.
```
itamarst: I'm not sure this sentence is quite accurate?

itamarst (approved):

itamarst:
@adiroiban Same as other PR; minor changes but you can feel free to merge.

src/twisted/newsfragments/12063.removal
adiroiban:
```suggestion
twisted.python.deprecate helper function will now always strip whitespaces from the docstrings.
```

src/twisted/python/deprecate.py
adiroiban: I am also not sure what was going in there :( 
This was done 2 months ago, and now I no longer remember what was the issue here.

I will try to review and update the docstring

adiroiban:
Many thanks for the review. Happy to see that Twisted on 3.13 gets some attention :)

Python 3.13.0 beta 1 is scheduled for 2024-05-07 with final release 3.13.0 2024-10-01

So we should have enough time to prepare Twisted for 3.13

adiroiban:
Thanks for the merge.

PR #12069, opened 2023-12-26
#12068: Fix incorrect env handling in spawnProcess on Posix when posix_spawnp is used.

### LINKED ISSUE #12068: spawnProcess() passes incorrect environment to subprocess when env=None and posix_spawnp() is used
[Documentation on reactor.spawnProcess](https://docs.twisted.org/en/stable/api/twisted.internet.interfaces.IReactorProcess.html) says the following about env parameter:

```env is None:  On POSIX: pass os.environ```

However, twisted has [this code](https://github.com/twisted/twisted/blob/68f112f1eecb4613a3b678314a5479464c184ab4/src/twisted/internet/process.py#L881) in the code path leading to a call to posix_spawnp().

```
if environment is None:
    environment = {}
```

This leads to a subprocess being initialized with empty environment even though `os.environ` was expected.
 
There's a PR with automated tests added to Twisted.

Subprocess having parent process environment when invoked via `reactor.spawnProcess(..., env=None)`.

**Testing environment**

 - Operating System and Version; 
   - Debian 12
 - Twisted version: 23.10.0
 - Reactor: default on Linux

**Additional context**
Probably a regression since 23.8.0 when posix_spawnp was enabled.

### PR DESCRIPTION
**Pull request by p12tic**
Fixes #12068. This is a regression starting at Twisted 23.8.0.

[Documentation on reactor.spawnProcess](https://docs.twisted.org/en/stable/api/twisted.internet.interfaces.IReactorProcess.html) says the following about env parameter:

```env is None:  On POSIX: pass os.environ```

This was not correctly handled in `os.posix_spawnp()` code path.

## Scope and purpose

This PR is just a simple one line fix to address #12068. Tests for environment passing on Posix have been added.

### DIFF
--- a/src/twisted/newsfragments/12068.bugfix
+++ b/src/twisted/newsfragments/12068.bugfix
@@ -1,0 +1,1 @@
+twisted.internet.process.Process, used by ``reactor.spawnProcess``, now copies the parent environment when the `env=None` argument is passed on Posix systems and ``os.posix_spawnp`` is used internally.
--- a/src/twisted/internet/process.py
+++ b/src/twisted/internet/process.py
@@ -879,7 +879,7 @@ class Process(_BaseProcess):
             else:
                 fdState.append((eachFD, isCloseOnExec))
         if environment is None:
-            environment = {}
+            environment = os.environ
 
         setSigDef = [
             everySignal
--- a/src/twisted/internet/test/test_process.py
+++ b/src/twisted/internet/test/test_process.py
@@ -29,6 +29,7 @@ from twisted.python.compat import networkString
 from twisted.python.filepath import FilePath, _asFilesystemBytes
 from twisted.python.log import err, msg
 from twisted.python.runtime import platform
+from twisted.test.test_process import Accumulator
 from twisted.trial.unittest import SynchronousTestCase, TestCase
 
 # Get the current Python executable as a bytestring.
@@ -999,6 +1000,132 @@ class ProcessTestsBuilder(ProcessTestsBuilderBase):
         hamcrest.assert_that(
             results,
             hamcrest.equal_to(["process already removed as desired"]),
+        )
+
+    def checkSpawnProcessEnvironment(self, spawnKwargs, expectedEnv, usePosixSpawnp):
+        """
+        Shared code for testing the environment variables
+        present in the spawned process.
+
+        The spawned process serializes its environ to stderr or stdout (depending on usePTY)
+        which is checked against os.environ of the calling process.
+        """
+        p = Accumulator()
+        d = p.endedDeferred = Deferred()
+
+        reactor = self.buildReactor()
+        reactor._neverUseSpawn = not usePosixSpawnp
+
+        reactor.callWhenRunning(
+            reactor.spawnProcess,
+            p,
+            pyExe,
+            [
+                pyExe,
+                b"-c",
+                networkString(
+                    "import os, sys; "
+                    "env = dict(os.environ); "
+                    # LC_CTYPE is set by python, see https://peps.python.org/pep-0538/
+                    'env.pop("LC_CTYPE", None); '
+                    'env.pop("__CF_USER_TEXT_ENCODING", None); '
+                    "sys.stderr.write(str(sorted(env.items())))"
+                ),
+            ],
+            usePTY=self.usePTY,
+            **spawnKwargs,
+        )
+
+        def shutdown(ign):
+            reactor.stop()
+
+        d.addBoth(shutdown)
+
+        self.runReactor(reactor)
+
+        expectedEnv.pop("LC_CTYPE", None)
+        expectedEnv.pop("__CF_USER_TEXT_ENCODING", None)
+        self.assertEqual(
+            bytes(str(sorted(expectedEnv.items())), "utf-8"),
+            p.outF.getvalue() if self.usePTY else p.errF.getvalue(),
+        )
+
+    def checkSpawnProcessEnvironmentWithPosixSpawnp(self, spawnKwargs, expectedEnv):
+        return self.checkSpawnProcessEnvironment(
+            spawnKwargs, expectedEnv, usePosixSpawnp=True
+        )
+
+    def checkSpawnProcessEnvironmentWithFork(self, spawnKwargs, expectedEnv):
+        return self.checkSpawnProcessEnvironment(
+            spawnKwargs, expectedEnv, usePosixSpawnp=False
+        )
+
+    @onlyOnPOSIX
+    def test_environmentPosixSpawnpEnvNotSet(self):
+        """
+        An empty environment is passed to the spawned process, when the default value of the C{env}
+        is used. That is, when the C{env} argument is not explicitly set.
+
+        In this case posix_spawnp is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithPosixSpawnp({}, {})
+
+    @onlyOnPOSIX
+    def test_environmentForkEnvNotSet(self):
+        """
+        An empty environment is passed to the spawned process, when the default value of the C{env}
+        is used. That is, when the C{env} argument is not explicitly set.
+
+        In this case fork+execvpe is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithFork({}, {})
+
+    @onlyOnPOSIX
+    def test_environmentPosixSpawnpEnvNone(self):
+        """
+        The parent process environment is passed to the spawned process, when C{env} is set to
+        C{None}.
+
+        In this case posix_spawnp is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithPosixSpawnp(
+            {"env": None}, os.environ
+        )
+
+    @onlyOnPOSIX
+    def test_environmentForkEnvNone(self):
+        """
+        The parent process environment is passed to the spawned process, when C{env} is set to
+        C{None}.
+
+        In this case fork+execvpe is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithFork({"env": None}, os.environ)
+
+    @onlyOnPOSIX
+    def test_environmentPosixSpawnpEnvCustom(self):
+        """
+        The user-specified environment without extra variables from parent process is passed to the
+        spawned process, when C{env} is set to a dictionary.
+
+        In this case posix_spawnp is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithPosixSpawnp(
+            {"env": {"MYENV1": "myvalue1"}},
+            {"MYENV1": "myvalue1"},
+        )
+
+    @onlyOnPOSIX
+    def test_environmentForkEnvCustom(self):
+        """
+        The user-specified environment without extra variables from parent process is passed to the
+        spawned process, when C{env} is set to a dictionary.
+
+        In this case fork+execvpe is used as the backend for spawning processes.
+        """
+        return self.checkSpawnProcessEnvironmentWithFork(
+            {"env": {"MYENV1": "myvalue1"}},
+            {"MYENV1": "myvalue1"},
         )
 
 

### CODE REVIEW

**Comment by p12tic** [comment]
please review

**Comment by adiroiban** [comment]
Thanks for the PR!

Not sure what to do here.

I think that we should fix the documentation instead.

That is, the bug lies in the documentation and not in the code :)

This change was introduced 2 years ago and is present in a public release.

By doing this change, we will break backward compatibility.

The initial change from 16 years ago was using `os.execvpe`

https://github.com/twisted/twisted/blame/dee676b040dd38b847ea6fb112a712cb5e119490/src/twisted/internet/process.py#L468

My understanding is that [execvpe](https://linux.die.net/man/3/execvpe) always requires an explicit environment, and it will not copy the environment from the parent process.

Regards

**Comment by glyph** [comment]
> Not sure what to do here.
> 
> I think that we should fix the documentation instead.
> 
> That is, the bug lies in the documentation and not in the code :)
> 
> This change was introduced 2 years ago and is present in a public release.
> 
> By doing this change, we will break backward compatibility.
> 
> The initial change from 16 years ago was using `os.execvpe`

Normally I would agree, except the circumstances under which `posix_spawn` is used are not entirely under control of the user, and this introduces an inconsistency *within* twisted's behaviors.

In this specific case, I think the least impactful thing is to fix the code, rather than the docs.

But perhaps we should shoot this over to the mailing list for a tiebreaker?

**Comment by p12tic** [comment]
> This change was introduced 2 years ago and is present in a public release.

This change of behavior was actually released in 23.8.0 by enabling use of `posix_spawnp`. From release notes: "reactor.spawnProcess() now uses posix_spawnp when possible, making it much more efficient (https://github.com/twisted/twisted/issues/5710)"

The regression is actually pretty bad. It silently drops environment variables only under certain conditions (e.g. when `path` argument of `spawnProcess` is not specified). Which means that most of the uses of `spawnProcess` in a certain codebase will work, except a few which depend on environment variables and satisfy the conditions for using `posix_spawnp`. Additionally, `spawnProcess` is something that's usually mocked in unit tests. Which means that unit tests wouldn't catch the failure. Only full end to end or manual testing could possibly catch it. These naturally have much worse code coverage, which means that failures are likely to be unnoticed for quite some time.

This is how the bug was discovered in the first place - by investigating a regression in a downstream project caused by updating Twisted.

> In this specific case, I think the least impactful thing is to fix the code, rather than the docs.

The code fix is needed in either case, because if we fix the docs, we need to adjust the behavior of how `spawnProcess` works in the case when regular `fork` is used.

**Comment by p12tic** [comment]
> My understanding is that [execvpe](https://linux.die.net/man/3/execvpe) always requires an explicit environment, and it will not copy the environment from the parent process.

execvpe uses os.environ when env argument is None. See here: https://github.com/python/cpython/blob/3.12/Lib/os.py#L594

File: src/twisted/test/test_process.py [hunk 1]
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
 
         return ended
 
+    def _testEnvironment(self, spawnKwargs, expectedEnv):
```
adiroiban: Maybe this is a better name and the docstring exlpain when it's ok to reuse this code in other tests.

This is testing code so I don't think that we need to worry about private methods.
```suggestion
    def checkSpawnProcessEnvironment(self, spawnKwargs, expectedEnv):
        """
        Shared code for testing the environment variables
        present in the spawned process.

        The spawned process serializes its environ to stderr which is checked against
        os.environ of the calling process.
        """
```

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+                bytes(str(sorted(expectedEnv.items())), "utf-8"),
+                p.outF.getvalue() if self.usePTY else p.errF.getvalue(),
+            )
+
+        return d.addCallback(processEnded)
+
+    def test_environmentPosixSpawnpEnvNotSet(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
+        os.environ of the calling process.
```
adiroiban: I am not sure if this is the right docstring.
This looks like the docstring for `_testEnvironment`

I was expecting for the docstring to describe the specific usage scenario for the API.

```suggestion
        An empty environment is passed to the spawned process, when the default value of the C{env} is used.
        That is, when the C{env} argument is not explicitly set.
```

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+        """
+        return self._testEnvironment({}, {})
+
+    def test_environmentForkEnvNotSet(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
+        os.environ of the calling process.
```
adiroiban: This is my undestanding of this test.

It's not clear why `os.fork` and not `os.posix_spawn` is expected to be used in this test. 

```suggestion
        An empty environment is passed to the spawned process, when the default value of the C{env} is used.
        Passing other argument to the spawned process will
        not affect the environment.
```

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+        """
+        return self._testEnvironment({"path": "/tmp"}, {})
+
+    def test_environmentPosixSpawnpEnvNone(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
+        os.environ of the calling process.
```
adiroiban:
```suggestion
        The parent environment is passed to the spawned process, when the C{env} argument is set to C{None}.
```

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+        """
+        return self._testEnvironment({"path": "/tmp", "env": None}, os.environ)
+
+    def test_environmentPosixSpawnpEnvCustom(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
+        os.environ of the calling process.
```
adiroiban:
```suggestion
        The explicit C{env} environment is passed to the spawned process.
```

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+    def test_environmentPosixSpawnpEnvNone(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
+        os.environ of the calling process.
+        """
+        return self._testEnvironment({"env": None}, os.environ)
+
+    def test_environmentForkEnvNone(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
```
adiroiban: I am not sure why `fork` is expected to be triggered by this test.

I think that a dedicated docstring could help identity what is tested here and answer the question: "Why we need this test?"

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,86 @@ def checkTerminated(ignored):
...
+        os.environ of the calling process.
+        """
+        return self._testEnvironment(
+            {"env": {"MYENV1": "myvalue1"}},
+            {"MYENV1": "myvalue1"},
+        )
+
+    def test_environmentForkEnvCustom(self):
+        """
+        Spawned process serializes its environ to stderr which is checked against
```
adiroiban: Same. Not sure why `fork` is expected to be used here.

Please consider adding a dedicated docstring to describe this test.

src/twisted/newsfragments/12068.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed invalid handling of ``env=None`` in ``reactor.spawnProcess`` on Posix
+when os.posix_spawnp() is being used.
```
adiroiban:
```suggestion
twisted.internet.process.Process, used by ``reactor.spawnProcess``, now copies the parent environment when the `env=None` argument is passed on Posix systems.
```

Is there a case of a supported OS when os.posix_spawnp is not used on Posix (Linux and macOS) ?

I guess that the fork code can be removed / deprecated.
Are there still posix OSes without `posix_spawnp` ?

adiroiban (changes requested):
Many thanks, Povilas for the PR and the explanations and investigation.

With the latest info, I also think that is best to fix the code.

I left a few minor comments regarding the scope/purpose of each test.

I tried to read the code, but I can't find why fork is expected to be used in those tests.

Please check my commend and add your feedback.

Other than that, this can be merged.

Once merged, I will try to trigger a new release.

Thanks again.

src/twisted/test/test_process.py
p12tic: Makes sense, thanks.

src/twisted/test/test_process.py
p12tic: I've now reused reactor._neverUseSpawn which is used in other tests. This should make the code much clearer compared to setting `path` argument to spawnProcess to something that is not supported by `posix_spawnp`.

src/twisted/test/test_process.py
p12tic: Fixed.

src/twisted/test/test_process.py
p12tic: Fixed.

src/twisted/test/test_process.py
p12tic: Fixed.

src/twisted/test/test_process.py
p12tic: Fixed.

src/twisted/newsfragments/12068.bugfix
p12tic: Unfortunately posix_spawnp does not support changing the current directory, uid and gid. Setting current directory has been accepted to POSIX (https://www.austingroupbugs.net/view.php?id=1208), but for uid and gid we'll need to use fork+execvpe for a long time it seems.

p12tic:
@adiroiban I addressed your comments.

p12tic:
please review

src/twisted/newsfragments/12068.bugfix
adiroiban: Many thanks for the info.
Just a question.

Do you think it is worth updating the current documentation to explain that posix_spawnp is used when possible, and if you use path and uid/gid the good old fork + execvpe is used ?

glyph: > Do you think it is worth updating the current documentation to explain that posix_spawnp is used when possible, and if you use path and uid/gid the good old fork + execvpe is used ?

If so, not in this PR :)

p12tic: > Do you think it is worth updating the current documentation to explain that posix_spawnp is used when possible, and if you use path and uid/gid the good old fork + execvpe is used ?

I think this is useful addition. Usage of os.fork is user visible because new python throws a warning, so end users may want to adjust their code to not use it (e.g. invoke spawnProcess differently).

I've opened https://github.com/twisted/twisted/pull/12072.

p12tic: @adiroiban I think it's important to mention that only posix_spawnp code path is being affected in the release note. I've used your wording as baseline and adjusted it to include posix_spawnp note.

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,115 @@ def checkTerminated(ignored):
...
+        Shared code for testing the environment variables
+        present in the spawned process.
+
+        The spawned process serializes its environ to stderr or stdout (depending on usePTY)
+        which is checked against os.environ of the calling process.
+        """
+        p = Accumulator()
+        d = p.endedDeferred = defer.Deferred()
+
+        reactor._neverUseSpawn = not usePosixSpawnp
```
adiroiban: Instead of using the global reactor,
maybe we can instantiate a separate reactor and then we will not have to care about the cleanup.

The suggestion is just an idea... the import should be at the start of the file.

```suggestion
        from twisted.internet.selectreactor import SelectReactor()
        supportingReactor = SelectReactor()
        supportingReactor._neverUseSpawn = not usePosixSpawnp
```

I think that is important not to modify the global / singleton instanced during a test.

src/twisted/test/test_process.py
```diff
@@ -1199,6 +1199,115 @@ def checkTerminated(ignored):
 
         return ended
 
+    def checkSpawnProcessEnvironment(self, usePosixSpawnp, spawnKwargs, expectedEnv):
```
adiroiban: I find it hard to read this code and know that the `False` means "use fork"

```python
return self.checkSpawnProcessEnvironment(
            False,
            {"env": {"MYENV1": "myvalue1"}},
            {"MYENV1": "myvalue1"},
        )
```

Just a minor suggestion.

We can also merge the current test code as it is 

The suggestion is just an idea, not the actual suggestion to commit.

```suggestion
    def checkSpawnProcessEnvironmentWithFork():
        return self._checkSpawnProcessEnvironment(self, spawnKwargs, expectedEnv, usePosixSpawnp=False)
    def checkSpawnProcessEnvironmentWithPosixSpawnp():
        return self._checkSpawnProcessEnvironment(self, spawnKwargs, expectedEnv, usePosixSpawnp=True)
    def _checkSpawnProcessEnvironment(self, spawnKwargs, expectedEnv, usePosixSpawnp):
```

adiroiban (dismissed):
Many thanks for the PR.

It looks good.

My only blocker for this PR is trying to use a dedicated reactor instance, rather than using the global reactor.

Maybe it's a bad suggestion :)

See my inline comments. Thanks!

src/twisted/test/test_process.py
exarkun: This isn't a safe way to do this.  Instead, there's a whole framework in twisted/internet/test/ for using per-test reactors (including many tests for `IReactorProcess`).

adiroiban: Thanks for the info.

My suggestion is not ok as it only guarantees that the `spawnProcess` will work for the `SelectReactor` 

The test is not executed for all the other reactors.

I guess that these tests needs to be moved to https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/test/test_process.py#L106-L113

and use `reactor = self.buildReactor()`

In this way, the tests are wrote once, but executed for any supported reactor.

glyph: > I guess that these tests needs to be moved to https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/test/test_process.py#L106-L113

Yes, exactly.

`twisted.test` is largely a trap for contributors at this point. We need to more clearly signpost the good vs. the bad ways to write tests and try to direct people to copy from the good examples (or at least, the better ones) and avoid these antipatterns.

p12tic: Thanks for suggestions. I've applied them all.

src/twisted/test/test_process.py
p12tic: Thanks, applied.

p12tic:
please review

p12tic:
@adiroiban 

Thanks a lot for your suggestions. I think I addressed all comments.

glyph (approved):
Looks great. Thanks for integrating all that feedback.

PR #12072, opened 2024-01-03
#12071: Document when spawnProcess uses os.posix_spawnp

### LINKED ISSUE #12071: Documentation on spawnProcess should document when posix_spawnp is used.
posix_spawnp is not identical to fork+execve. New python versions throw a deprecation warning on `os.fork` usage. Downstream users may want to adjust usages of spawnProcess to not use fork+execve.

p12tic:
posix_spawnp is not identical to fork+execve. New python versions throw a deprecation warning on os.fork usage. Downstream users may want to adjust usages of spawnProcess to not use fork+execve.

## Scope and purpose

Fixes #12071.

### DIFF
--- a/src/twisted/internet/interfaces.py
+++ b/src/twisted/internet/interfaces.py
@@ -1090,6 +1090,16 @@ class IReactorProcess(Interface):
         arguments will be decoded up to L{unicode} using the encoding given by
         L{sys.getfilesystemencoding} (C{utf8}) and given to Windows's native "wide" APIs.
 
+        On POSIX, C{spawnProcess} may use deprecated C{os.fork} under the following
+        conditions:
+
+         - C{path} being not C{None} or a path that resolves to something different
+            than the current directory.
+         - C{uid} being not C{None}.
+         - C{gid} being not C{None}.
+
+        If none of the above conditions hold, deprecated C{os.fork} will not be used.
+
         @param processProtocol: An object which will be notified of all events
             related to the created process.
 
--- a/src/twisted/newsfragments/12071.doc
+++ b/src/twisted/newsfragments/12071.doc
@@ -1,0 +1,1 @@
+twisted.internet.interfaces.IReactorProcess now documents certain details of the POSIX-based implementation included with Twisted.


p12tic:
please review

src/twisted/newsfragments/12071.doc
```diff
@@ -0,0 +1 @@
+Added documentation on when ``spawnProcess`` uses ``os.posix_spawnp``.
```
exarkun: This should be like `twisted.internet.interfaces.IReactorProcess now documents certain details of the POSIX-based implementation included with Twisted.`

Or something similar if the docs move elsewhere.

src/twisted/internet/interfaces.py
```diff
@@ -1090,6 +1090,16 @@ def spawnProcess(
...
 
+        On POSIX, C{spawnProcess} will use C{os.posix_spawnp} if it is available
+        and the arguments passed to C{spawnProcess} are supported. Otherwise,
+        a combination of C{os.fork} and C{os.execvpe} is used. The following
+        arguments cause C{spawnProcess} to not use C{os.posix_spawnp}:
+
+         - C{path} being not C{None} or a path that resolves to something different
+            than the current directory.
+         - C{uid} being not C{None}.
+         - C{gid} being not C{None}.
```
exarkun, line 1099: These seem more like implementation details than the public interface we want to promise (consider: spawnProcess never used to use posix_spawnp at all - and it was changed to do so without introducing a new interface).

I don't know if that means these words belong somewhere else (the interface is certainly an obvious and tempting place to put them).  There is long-form documentation about using processes with Twisted which might be a good place (but it might be less visible, if that matters).  At the very least, if this info stays here, I'd like to see a note with it saying these are implementation details and subject to change without invoking the usual backwards incompatibility process.

adiroiban: I think that the narrative / non-API docs are here https://docs.twisted.org/en/stable/core/howto/process.html#running-another-process


I am +1 for moving this into the implementation API documention rather than having them in the interface.

I think one options is moving this here https://docs.twistedmatrix.com/en/stable/api/twisted.internet.process.Process.html

adiroiban (changes requested):
Many thanks for the PR. Much appreciate!

I think that this is 99% ok :)

I think is best to have the interface documentation as generic as possible.

adiroiban (changes requested):
Many thanks for the PR. Much appreciate!

I think that this is 99% ok :)

I think is best to have the interface documentation as generic as possible.

src/twisted/internet/interfaces.py
p12tic: @exarkun @adiroiban I agree with your comments. The current addition to documentation is badly worded and its place is in implementation docs in the current form.

However, I think that there still needs to be update to the API docs. The fact that certain uses of `spawnProcess` may invoke `os.fork` under the hood became important after `os.fork` became deprecated.

For the cases where `spawnProcess` uses `os.fork` right now, `os.fork` is the only alternative. Deprecation of `os.fork` implies that it may be removed in the future. This means that the arguments of `spawnProcess` that currently require `os.fork` are sort of softly deprecated too.

Accordingly, it makes sense to encode the fact that certain uses of `spawnProcess` are guaranteed not to invoke `os.fork` into the API contract. It probably makes sense to deprecate the arguments of `spawnProcess` that cause `os.fork` to be invoked too, but I think this is a step too far right now. Maybe `os.fork` will be deprecated but not removed for a decade and in the meantime `posix_spawnp` will get the missing functionality on the platforms we care about.

src/twisted/newsfragments/12071.doc
p12tic: Applied.

src/twisted/internet/interfaces.py
p12tic: I've updated the wording to be more generic. It only specifies that under certain conditions `os.fork` *may* be used. If these conditions do not hold, then `os.fork` *will not* be used. I think this is generic enough and does not constrain the implementation in any way as at this point increasing the usage of `os.fork` should probably be considered a bug.

p12tic:
please review

p12tic:
I've added several comments explaining the PR into this thread https://github.com/twisted/twisted/pull/12072#discussion_r1440584211

PR #12077, opened 2024-01-11
fix trunk failing build by pinning PyPy to 3.7.13, as 3.7.14 is failing currently, and add some type annotations to deal with new annotations in pyOpenSSL

### LINKED ISSUE #12076: trunk build failing
It seems like pypy-3.9 and mypy have broken on trunk between commits, so some dependency has shifted in a way that breaks a bunch of Deferred tests (???).

See the build history here: https://github.com/twisted/twisted/actions/runs/7356600638

image

We really need to pin mor ethings.

glyph:
Fixes #12076

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -168,14 +168,14 @@ jobs:
           # extension.
           # There is very little PYPY specific code so there is not much to
           # gain from reporting coverage.
-          - python-version: 'pypy-3.9'
+          - python-version: 'pypy-3.10'
             tox-env: 'alldeps-nocov-posix'
             job-name: 'no-coverage'
             skip-coverage: yes
 
           # We still run some tests with coverage,
           # as there are test with specific code branches for pypy.
-          - python-version: 'pypy-3.9'
+          - python-version: 'pypy-3.10'
             trial-target: 'twisted.test.test_compat twisted.test.test_defer twisted.internet.test.test_socket twisted.trial.test.test_tests'
             job-name: 'with-coverage'
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -168,14 +168,14 @@ jobs:
           # extension.
           # There is very little PYPY specific code so there is not much to
           # gain from reporting coverage.
-          - python-version: 'pypy-3.10'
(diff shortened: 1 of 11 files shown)


glyph:
please review

.github/workflows/test.yaml
```diff
@@ -168,14 +168,14 @@ jobs:
...
-          - python-version: 'pypy-3.9'
+          - python-version: 'pypy3.10-v7.3.13'
             tox-env: 'alldeps-nocov-posix'
             job-name: 'no-coverage'
             skip-coverage: yes
 
           # We still run some tests with coverage,
           # as there are test with specific code branches for pypy.
-          - python-version: 'pypy-3.9'
+          - python-version: 'pypy3.10-v7.3.13'
```
adiroiban, line 178: Maybe mention in the releas notes that PyPy3.9 is no longer supported, and we now support PyPy 3.10?

adiroiban (approved):
Looks good. Thanks.

I don't know why the iocp reactor code is not reported as coverted by automated test.

We run the whole test suite on Windows with iocp reactor and coverage enabled.

But this can be looked into a separate PR.

The important part is to fix trunk.

---------

@glyph if you have time to review, I can work on the mypy upgrade.

Regards

adiroiban:
Instead of "hopefully fix trunk" maybe the title of this PR should be "Use PyPy 3.10 in testing to fix CI failures caused by latest changes to PyPy 3.9."

.github/workflows/test.yaml
glyph: The interesting thing that is happening here is that we are not supporting the latest PyPy, 7.3.14; 7.3.13 works with both 3.9 and 3.10, and the CPython coverage for syntax differences seems sufficient, given that those two "versions" are substantially similar. So I wouldn't say it's unsupported, and if we have PyPy-3.9 specific issues we could add another buildbot. I bumped the Python version since if we are going to make space in the test matrix, it seems like 3.10 would probably be more used at this point than 3.9.

glyph:
> We run the whole test suite on Windows with iocp reactor and coverage enabled.

It must be a codecov bug. Note that the codecov status check says that codecov/patch hits 100% of the diff target, and some of the changed lines are in those files, so it's not even self-consistent.

PR #12083, opened 2024-01-14
12084 PyPy stack depth check change for 7.3.14

### LINKED ISSUE #12084: twisted's test suite fails on pypy 7.3.14 due to stack depth changes
This was reported in https://github.com/pypy/pypy/issues/4850 but is really a bug in Twisted, not PyPy, since PyPy is *fixing* a bug here to get closer to CPython's behavior.

mattip:
## Scope and purpose

Fixes #12084

PyPy 7.3.14 made stack crawling more compatible with CPython, this broke tests in twisted. Since this is a testing-only fix, does it need a release note?

## Contributor Checklist:

### DIFF
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -58,6 +58,9 @@ log = Logger()
 _T = TypeVar("_T")
 _P = ParamSpec("_P")
 
+# See use in _inlineCallbacks for explanation and removal timeline.
+_oldPypyStack = _PYPY and implementation.version < (7, 3, 14)
+
 
 class AlreadyCalledError(Exception):
     """
@@ -2022,14 +2025,10 @@ def _inlineCallbacks(
             appCodeTrace = traceback.tb_next
             assert appCodeTrace is not None
 
-            _oldPypyStackCompatibility = _PYPY and implementation.version < (
-                7,
-                3,
-                14,
-            )
-            if _oldPypyStackCompatibility:
-                # PyPy before 7.3.14 adds an extra frame.
-                # This code can be removed once we no longer need to support PyPy 7.3.13 or older.
+            if _oldPypyStack:
+                # PyPy versions through 7.3.13 add an extra frame; 7.3.14 fixed
+                # this discrepancy with CPython.  This code can be removed once
+                # we no longer need to support PyPy 7.3.13 or older.
                 appCodeTrace = appCodeTrace.tb_next
                 assert appCodeTrace is not None
 
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -17,7 +17,7 @@ from asyncio import AbstractEventLoop, Future, iscoroutine
 from contextvars import Context as _Context, copy_context as _copy_context
 from enum import Enum
 from functools import wraps
-from sys import exc_info
+from sys import exc_info, implementation
 from types import CoroutineType, GeneratorType, MappingProxyType, TracebackType
 from typing import (
     TYPE_CHECKING,
@@ -53,6 +53,8 @@ from twisted.python.deprecate import deprecated, warnAboutFunction
 from twisted.python.failure import Failure, _extraneous
 
 log = Logger()
+
+_STACK_LEVEL_INCOMPATIBLE = _PYPY and implementation.version < (7, 3, 14)
 
 
 _T = TypeVar("_T")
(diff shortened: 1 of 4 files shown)


mattip:
There was a timeout in the 3.8-macos-12-default-tests run, but this seems to be happening in other PRs.

mattip:
please review

adiroiban:
Thanks Matti for the PR.

So we had #12077 in which we updated the code to work with pypy3.10-v7.3.13

I am not sure what is our policy for pypy support, but I guess that with our limited resourcers we should try to support the latest pypy version.

With that in mind, I think that we can consider this PR as something like `Add support for PyPy 7.3.14 on Python 3.10`

I will push a commit on this branch to run our tests with 7.3.14

src/twisted/newsfragments/12083.bugfix
```diff
@@ -0,0 +1 @@
+The stack depth check in twisted.internet.defer was adjusted for PyPy > 7.3.14
```
adiroiban: I think that we have `>=` - not only releases newer than 7.3.14, but also 7.3.14

```suggestion
twisted.internet.defer inline deferred stack introspection was adjusted for latest PyPy 7.3.14 release.
```

src/twisted/internet/defer.py
```diff
@@ -54,6 +54,8 @@
 
 log = Logger()
 
+_STACK_LEVEL_INCOMPATIBLE = _PYPY and implementation.version < (7, 3, 14)
```
adiroiban: It seems like this is used in a single place...so maybe we can move it inline.

In terms on name, I would prefer something like `_OLD_PYPY_STACK_COMPATIBILITY`

It looks like older PyPy were not compatible with cpython... but latest PyPy release is.

```suggestion
```

src/twisted/internet/defer.py
```diff
@@ -2022,8 +2024,8 @@ def _inlineCallbacks(
             appCodeTrace = traceback.tb_next
             assert appCodeTrace is not None
 
-            if _PYPY:
-                # PyPy as of 3.7 adds an extra frame.
+            if _STACK_LEVEL_INCOMPATIBLE:
+                # PyPy as of 3.7 and before 7.3.14 adds an extra frame.
```
adiroiban: Maybe something like this, with a comment that in the future this can be removed.

```suggestion
            _old_pypy_stack_compatibility = _PYPY and implementation.version < (7, 3, 14)
            if _old_pypy_stack_compatibility:
                # PyPy before 7.3.14 adds an extra frame.
                # This code can be removed once we no longer need to support PyPy 7.3.13 or older.
```

adiroiban (commented):
Many thanks for the fix. Much appreciated.

I only have minor comments.

I left some inline suggestion.  I hope @glyph  can review and we can have this merged soon.

I can see PyPy test running with [ Successfully set up PyPy 7.3.14 with Python (3.10.13)
](https://github.com/twisted/twisted/actions/runs/7517950179/job/20464718892?pr=12083#step:3:15) . So the changes in this PR ar tested.

mattip:
> I am not sure what is our policy for pypy support

Please feel free to ping me directly in this repo whenever PyPy problems crop up. I appreciate the efforts you all do to support PyPy, and will try to make it a priority to help out where I can.

src/twisted/internet/defer.py
mattip: Sure, these changes look fine to me. I will let the twisted maintainers commit these suggestions.

adiroiban:
> There was a timeout in the 3.8-macos-12-default-tests run, but this seems to be happening in other PRs.

MacOS tests are flaky.

We need to fix that soon as they can be a big annoyance and confusion for contributors.

adiroiban (approved):
All good. Thanks, Matti.

We now run Twisted tests on CPython 3.10 with whatever GitHub Action thinks it's the latest PyPy release.

So we should be able to observe any future changes and we will get in touch.

-------

We don't run tests with unreleased PyPy versions.

For reference, if someone would like to run Twisted tests with main PyPy dev version , there is subset of Twisted tests that can be executed and which should provide good enough coverage

https://github.com/twisted/twisted/actions/runs/7518115938/job/20465098025?pr=12083#step:10:689

On our GitHub Actions, the test run takes 1 minutes...but it mostly tox venv setup. The tests run is about 10 seconds.

adiroiban:
@glyph looks like codecov.io is flaky again.


Note that we have this [PR for review](https://github.com/twisted/twisted/pull/11745) in which the diff coverage is generated using only GitHub Actions.

-------

Please double check that we are ok with the coverage and feel free to merge. Thanks!

I have flagged PR so that we will not forget to merge this before the next release.

glyph:
I am going to make a few minor tweaks here (getting the additional arithmetic / comparison out of a *very* hot loop, moving to a single constant, and re-pinning pypy to .14; we should figure out another way to get notified of pypy upgrades that doesn't break CI for unrelated issues)

Thanks a *ton* @mattip for figuring this out; great to work with you on it :).

glyph:
> @glyph looks like codecov.io is flaky again.

This is a real coverage issue. If we drop .13 from the matrix entirely, then we don't cover the old branch. I'm going to run .13 with coverage, and .14 without, since we already have 2 pypy jobs, and that should make codecov happy.

PR #12090, opened 2024-01-22
#12011 Speed up small TLS writes

### LINKED ISSUE #12011: Performance issue in TLS: assert that each written chunk of a data is not a str
This check is in the hot path, and can slow down writing when there are lots of small writes: https://github.com/twisted/twisted/blob/88172f74de390f148ec2a44829090718cdfd43c5/src/twisted/protocols/tls.py#L452

Fixing it will be done as followup to #11989

itamarst:
Fixes #12011.

Benchmark results with Python 3.11:

`trunk`:

```
$ time python tls_benchmark.py small

real    0m2.425s
user    0m2.302s
sys     0m0.069s
```

This branch:

```
$ time python tls_benchmark.py small

real    0m1.705s
user    0m1.562s
sys     0m0.088s
```

Benchmark:

```python
import sys

from twisted.internet.protocol import Protocol
from twisted.internet.ssl import CertificateOptions, PrivateCertificate
from twisted.protocols.loopback import loopbackTCP
from twisted.internet.task import react

class Client(Protocol):
    def connectionMade(self):
        self.count = 10_000
        self.transport.startTLS(CertificateOptions(verify=False, requireCertificate=False))
        assert sys.argv[1] in ("small", "big")
        self.next_iteration()

    def next_iteration(self):
        self.count -= 1
        if sys.argv[1] == "small":
            for i in range(1_000):
                self.transport.write(b"X")
        else:
            self.transport.write(b"X" * 1000)

        if self.count == 0:
        else:
            reactor.callLater(0, self.next_iteration)


class Server(Protocol):
    received = 0

    def connectionMade(self):
        with open("server.pem", "rb") as f:
            options = PrivateCertificate.loadPEM(f.read()).options()
        self.transport.startTLS(options)

    def dataReceived(self, data):
        self.received += len(data)

    def connectionLost(self, reason):
        assert self.received == 10_000_000, self.received


assert len(sys.argv) == 2
react(lambda reactor: loopbackTCP(Server(), Client(), noisy=True))
```


adiroiban:
Hi Itamar, Thanks for looking into this. Is this PR ready for review? Regards

itamarst:
Yes, it's ready for review.

src/twisted/protocols/tls.py
```diff
@@ -788,11 +788,10 @@ def __init__(
...
         self._aggregator = _AggregateSmallWrites(actual_write, factory._clock)
-
-    def write(self, data: bytes) -> None:
-        if isinstance(data, str):  # type: ignore[unreachable]
-            raise TypeError("Must write bytes to a TLS transport, not str.")
-        self._aggregator.write(data)
+        # This is kinda ugly, but speeds things up a lot in a hot path with
+        # lots of small TLS writes. May become unnecessary in Python 3.13 or
+        # later if JIT and/or inlining becomes a thing.
+        self.write = self._aggregator.write  # type: ignore[method-assign]
```
adiroiban, line 794: It's not pretty, but I think is the best option.

I think that, in the end, having the `_AggregateSmallWrites` code using composition is nicer.

src/twisted/protocols/test/test_tls.py
```diff
@@ -687,25 +687,6 @@ def connectionMade(self):
 
         return self.writeBeforeHandshakeTest(SimpleSendingProtocol, data)
 
-    def test_writeUnicodeRaisesTypeError(self):
```
adiroiban: Maybe leave this test as a reminder that passing bytes should never be supported again.

But I am also ok to have this removed.

src/twisted/newsfragments/12011.bugfix
```diff
@@ -0,0 +1 @@
+Improved performance when doing lots of small writes to TLS transports.
```
adiroiban: One suggestion to follow the coding convention

```suggestion
twisted.protocols.tls.BufferingTLSTransport, used by default by twisted.protocols.tls.TLSMemoryBIOFactory, was refactored for improved performance when doing a high number of small writed.
```

adiroiban (approved):
Looks good. Thanks. Only minor comments.

src/twisted/protocols/test/test_tls.py
itamarst: The problem is it doesn't fail in a useful way now.

adiroiban: I think the important part is to make sure that it fails... in any way. Rather than ending up with a case in which the data is ignored... or something like that.

So even if the error is not explicit. I think that it still important to make sure that it fails.

The previous explicit error was a huge help for my in the past... but I am ok to have something less helful, in the name of performance.

I hope in the future mypy can help here.

itamarst: OK I restored the test.

PR #12094, opened 2024-01-26
#8570 Increase HTTP protocol length limits

### LINKED ISSUE #355: Default downloader fails to get page
'http://autos.msn.com/research/userreviews/reviewlist.aspx?ModelID=14749'

Looks like the default downloader implemented with twisted lib can't fetch the above url.  I ran 'scrapy shell  http://autos.msn.com/research/userreviews/reviewlist.aspx?ModelID=14749', and got  the following output. 

```
  File "/usr/local/bin/scrapy", line 5, in <module>
    pkg_resources.run_script('Scrapy==0.17.0', 'scrapy')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 489, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 1207, in run_script
    execfile(script_filename, namespace, namespace)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/EGG-INFO/scripts/scrapy", line 4, in <module>
    execute()
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/cmdline.py", line 143, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/cmdline.py", line 88, in _run_print_help
    func(*a, **kw)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/cmdline.py", line 150, in _run_command
    cmd.run(args, opts)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/commands/shell.py", line 47, in run
    shell.start(url=url, spider=spider)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/shell.py", line 43, in start
    self.fetch(url, spider)
  File "/Library/Python/2.7/site-packages/Scrapy-0.17.0-py2.7.egg/scrapy/shell.py", line 85, in fetch
    reactor, self._schedule, request, spider)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/twisted/internet/threads.py", line 118, in blockingCallFromThread
    result.raiseException()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/twisted/python/failure.py", line 370, in raiseException
    raise self.type, self.value, self.tb
twisted.internet.error.ConnectionDone: Connection was closed cleanly.
```

But both urlopen of urllib2 and requests.get can download the page smoothly.

### LINKED ISSUE #8570: Incorrect (too long) header from server causes Twisted Agent to fail

Gallaecio:
## Scope and purpose

Fixes #8570, fixes https://github.com/scrapy/scrapy/issues/355, closes https://github.com/scrapy/scrapy/pull/5911.

Increase the length limit affecting the HTTP client parser to support scenarios seen in the wild, such as https://www.vapestore.co.uk/ returning a header with a 28k+ char  value.

It is an arbitrary increase mimicking CPython’s, though, so there is a chance that in the future some new scenario for which these new limits are not enough warrants either increasing the limits further or making it easier to increase the limits from user code, which [is not straightforward at the moment without relying on private APIs](https://github.com/scrapy/scrapy/pull/5911).

I also made a minor change on the contributor docs, switching `pip install -e .[dev]` to `pip install -e .[dev,test]` based on my experience. Not sure if that’s OK to do here, or if that’s the right change (maybe you prefer `pip install -e .[test]` to be suggested on the following section about running tests).

~Regarding tests, there are no existing tests that check the specific values, which I think may make sense given these are trivially-selected limits. Also, there are already tests for the limits themselves (those tests use custom values, for performance reasons I imagine).~

## Post-merge to-do

- [ ] Create an issue about https://github.com/twisted/twisted/pull/12094/files/3ade9422f18248993f2921a1439649bd39440c15#r1473705634


Gallaecio:
please review

docs/development/dev-process.rst
```diff
@@ -23,7 +23,7 @@ see in particular `Git and GitHub learning resources <https://help.github.com/ar
     $ git clone https://github.com/twisted/twisted twisted
     $ python3 -m venv ./venv
     $ . venv/bin/activate
-    $ pip install -e .[dev]
+    $ pip install -e .[dev,test]
```
adiroiban, line 26: Thanks. We can have this for now.

In an ideal case, dev will also install test.
But for another PR.

Free free to send a PR if you have time to help with this.

src/twisted/protocols/basic.py
```diff
@@ -403,6 +403,9 @@ def _handleParseError(self):
         self.brokenPeer = 1
 
 
+_MAX_LINE_LENGTH = 32768
```
adiroiban: Even if this is a private variable, I think that it helps to add a comment about the scope/purpose of this variable.

it should help other know if they can reuse it, or define a new one.

src/twisted/protocols/newsfragments/8570.bugfix
```diff
@@ -0,0 +1,3 @@
+Doubled the MAX_LENGTH of twisted.protocols.basic.LineOnlyReceiver and
+twisted.protocols.basic.LineReceiver, from 16384 bytes to 32768 bytes, to
+support scenarios seen in the wild.
```
adiroiban: I think that this needs to go to src/twistd/newsfragments

we don't have a separate section for protocol.

To comply with our standard

```suggestion
twisted.protocols.basic.LineOnlyReceiver and
twisted.protocols.basic.LineReceiver, can now handle lines up to 32768 bytes long. In previous versions, the limit was 16384 bytes.
```

src/twisted/web/newsfragments/8570.bugfix
```diff
@@ -0,0 +1,2 @@
+Multiplied the totalHeadersSize of twisted.web.http.HTTPChannel by 4, from
+16384 bytes to 65536 bytes, to support scenarios seen in the wild.
```
adiroiban: To comply with the project release notes convetion.

The notes can be a summary of the change.
People can find out more about the change by checking the associated issue in GitHub.

```suggestion
twisted.web.http.HTTPChannel.totalHeadersSize now has the value of 65536 bytes. In previous releases it was 16384 bytes.
```

src/twisted/web/http.py
```diff
@@ -2208,7 +2208,7 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
     """
 
     maxHeaders = 500
-    totalHeadersSize = 16384
+    totalHeadersSize = 65536
```
adiroiban: Minor comment.
Why not extract this into a variable ?

-------

And it looks like we have missing automates tests.

With `totalHeadersSize = 16384` in `trunk` branch, we should have a test that will check that a header size of `16384 + 1 ` will trigger a failure.

And with the change from this branch, I was expecting to see that test fail now.

Can you please look into writing a test to make sure that Twisted now accepts headers of size `65536` and rejects headers at `65536 + 1` ?

Thanks

adiroiban (changes requested):
Many thanks for the changes.

I left more comments inline.

I think they are an improvement over what we have in `trunk`

it is very important to have automated test for these values.

Let me know if you need help with the autoamted tests.

src/twisted/protocols/basic.py
```diff
@@ -419,7 +422,7 @@ class LineOnlyReceiver(protocol.Protocol):
 
     _buffer = b""
     delimiter = b"\r\n"
-    MAX_LENGTH = 16384
+    MAX_LENGTH = _MAX_LINE_LENGTH
```
exarkun: Making this as a change to *all* `LineOnlyReceiver` and `LineReceiver` uses seems pretty excessive.  The issue is with certain HTTP servers.  Assuming the right fix is to make Twisted more liberal in what it accepts from HTTP servers, this doesn't mean *all* line-based protocols implemented with Twisted should have the same change applied.

I'd move this fix into the HTTP implementation and leave `LineReceiver` / `LineOnlyReceiver` alone.

src/twisted/protocols/basic.py
graingert: Why not up it to 65536 to match cpython's http.client?

https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L111

src/twisted/protocols/basic.py
Gallaecio: Agreed. I have now limited the scope of the changes to `HTTPClientParser`.

Gallaecio:
please review

src/twisted/web/test/test_newclient.py
```diff
@@ -962,6 +961,32 @@ def test_ignored1XXResponseCausesLog(self):
         self.assertEquals(event["log_format"], "Ignoring unexpected {code} response")
         self.assertEquals(event["code"], 103)
 
+    def test_headerShortEnough(self):
+        """Test that HTTP responses with header lines up to 65536 bytes long
+        are allowed."""
```
adiroiban: Maybe keep it simple :)

based on https://jml.io/test-docstrings/

```suggestion
        """
        It allows receiving HTTP header lines up to 65536 bytes long.
        """
```

src/twisted/web/test/test_newclient.py
```diff
@@ -962,6 +961,32 @@ def test_ignored1XXResponseCausesLog(self):
...
+        prefix = b"a: "
+        line = prefix + b"a" * (65536 - len(prefix)) + b"\r\n"
+        response = b"HTTP/1.1 200 OK\r\n" + line + b"\r\n"
+        protocol = HTTPClientParser(
+            Request(b"GET", b"/", _boringHeaders, None), lambda ign: None
+        )
+        protocol.makeConnection(StringTransport())
+        protocol.dataReceived(response)
+        self.assertEqual(protocol.transport.disconnecting, False)
```
adiroiban, line 1060: Can we check that the header was received in some way... instead of being somehow discarded ?

src/twisted/web/test/test_newclient.py
```diff
@@ -962,6 +961,32 @@ def test_ignored1XXResponseCausesLog(self):
...
+
+    def test_headerTooLong(self):
+        """Test that HTTP responses with header lines up to 65536 bytes long
+        are allowed."""
```
adiroiban: I think that this is a copy/paste error.

Maybe something like this?

```suggestion
        """
        When the remote server returns an HTTP header that has a long line, will disconnected the connection.
        """
```

Is disconnection everything that it does?

I was expecting to also see an error and the reason for disconnection.

Is it possible to also test the error message in this test?
In yes, it wouldhelp to update the docstring.

src/twisted/web/test/test_newclient.py
```diff
@@ -962,6 +961,32 @@ def test_ignored1XXResponseCausesLog(self):
...
+        prefix = b"a: "
+        line = prefix + b"a" * (65536 - len(prefix) + 1) + b"\r\n"
```
adiroiban, line 1067: For this tests, does it matter the format of the header?

Maybe just write random long data.

src/twisted/web/test/test_newclient.py
```diff
@@ -962,6 +961,32 @@ def test_ignored1XXResponseCausesLog(self):
...
+        response = b"HTTP/1.1 200 OK\r\n" + line + b"\r\n"
+        protocol = HTTPClientParser(
+            Request(b"GET", b"/", _boringHeaders, None), lambda ign: None
+        )
+        protocol.makeConnection(StringTransport())
+        protocol.dataReceived(response)
+        self.assertEqual(protocol.transport.disconnecting, True)
```
adiroiban, line 1075: Can we also check the disconnection error / reason message?

I think that is important to know the reason.
If it's not possible to get the reason, I think that is important to add a FIXME here and create an follow up issue

Once we can get the error, we can update this test.

src/twisted/web/newsfragments/8570.bugfix
```diff
@@ -0,0 +1,2 @@
+twisted.web._newclient.HTTPClientParser.MAX_LENGTH now has the value of 65537
+bytes. In previous releases it was 16384 bytes.
```
adiroiban: Just a very minor comment.

```suggestion
twisted.web._newclient.HTTPClientParser.MAX_LENGTH now has the value of 65537
bytes. In previous releases it was 16384 bytes. It now matches the value from CPython HTTP client.
```

adiroiban (changes requested):
It looks good. Many thanks!

I think that this is very close to being ready to merge.

I left a few minor comments, but that needs to be fixed before merge.

src/twisted/web/test/test_newclient.py
Gallaecio: I was actually originally going to ask about that. I was first checking by testing the resulting headers, which I will probably restore in combination with the disconnect check based on https://github.com/twisted/twisted/pull/12094#discussion_r1473704052, and was going to ask for a better way to check that things failed, as I don’t know. The relevant code [simply calls “loseConnection” on the transport](https://github.com/twisted/twisted/blob/234f3788b60f21ccdc51e7b865d244b9343877bf/src/twisted/protocols/basic.py#L541) as far as I can tell.

src/twisted/web/test/test_newclient.py
Gallaecio: Given https://github.com/twisted/twisted/pull/12094/files/3ade9422f18248993f2921a1439649bd39440c15#r1473704052, maybe it makes sense to keep this a valid header, so that checking that the parsed header list is empty has more value (i.e. it is not empty because it was not a valid header, it is empty due to the length limit).

src/twisted/web/test/test_newclient.py
Gallaecio: [I don’t think there’s any error](https://github.com/twisted/twisted/blob/234f3788b60f21ccdc51e7b865d244b9343877bf/src/twisted/protocols/basic.py#L609-L619). [The original Scrapy issue traceback only mentions the disconnection](https://github.com/scrapy/scrapy/issues/355).

PR #1487, opened 2020-11-29
[10061] increase size of FFDH keys for conch testing

reaperhulk:
This matters because cryptography will break you shortly because OpenSSL is going to break us.

## Contributor Checklist:

* [x] The associated ticket in Trac is here: https://twistedmatrix.com/trac/ticket/10061
* [x] I ran `tox -e lint` to format my patch to meet the [Twisted Coding Standard](https://twistedmatrix.com/documents/current/core/development/policy/coding-standard.html)
* [x] I ran additional checks listed at [Getting Your Patch Accepted](https://twistedmatrix.com/trac/wiki/TwistedDevelopment#GettingYourPatchAccepted)
* [x] I have created a newsfragment in src/twisted/newsfragments/ (see: [News files](https://twistedmatrix.com/trac/wiki/ReviewProcess#Newsfiles))
* [x] I have updated the automated tests.
* [x] I have submitted the associated Trac ticket for review by adding the word `review` to the keywords field in Trac, and putting a link to this PR in the comment; it shows up in https://twisted.reviews/ now.

PR #12095, opened 2024-01-28
#12095 Increase RSA keys generated in tests

alex:
This will allow pyca/cryptography to increase the minimum key size. 512-bit keys are wildly insecure, so we want to disallow them.

### DIFF
--- a/src/twisted/conch/test/test_keys.py
+++ b/src/twisted/conch/test/test_keys.py
@@ -1675,8 +1675,8 @@ class PersistentRSAKeyTests(unittest.TestCase):
         tempDir = FilePath(self.mktemp())
         keyFile = tempDir.child("mykey.pem")
 
-        key = keys._getPersistentRSAKey(keyFile, keySize=512)
-        self.assertEqual(key.size(), 512)
+        key = keys._getPersistentRSAKey(keyFile, keySize=1024)
+        self.assertEqual(key.size(), 1024)
         self.assertTrue(keyFile.exists())
 
     def test_noRegeneration(self):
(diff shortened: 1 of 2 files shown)


adiroiban:
Thanks Alex. Let me know when this is ready for review.

I guess that this is a change needes as in the future cryptography will no longer allow keys below 1k.

Cheers

alex:
This is ready for review now (unless you think it's missing something?)

And yes, this is to keep twisted tests green on a future release -- twisted itself is fine, just the tests are impacted

adiroiban (approved):
I am happy to have this merged.
Much appreciated for including twisted into the `cryptography` test suite.

-------

Is there a GitHub issue, or discussion related to this change?

I guess that this is a follow up of PR https://github.com/twisted/twisted/pull/1487  ... this updates KEX part.

--------

I know that in the past @glyph has insisted on created a separate/dedicated issue for each change.
But I think that for these code cleanup, it's ok to only have a PR, as long as the PR has a descriptions.

This PR has no description, so I would prefer to merge this only after we add a few words about these changes from here.

thanks again!

PS: Many thanks for the cryptography 42 release :p

alex:
Ooops, there was some extra text in teh commit message that I acciddentally removed from the PR description.

Let me know if that looks good -- if so I'll merge.

adiroiban:
Many thanks. I have merged this.

exarkun:
> I know that in the past @glyph has insisted on created a separate/dedicated issue for each change. But I think that for these code cleanup, it's ok to only have a PR, as long as the PR has a descriptions.

This isn't just glyph's insistence.  This is project policy.  It's part of the responsibility of committers to adhere to project policy.  If you think there should be a variance, take it up on the list to get an exception or have the overall policy changed.

adiroiban:
True. Thanks for the heads up.
I start a discussion to change the policy at https://mail.python.org/archives/list/twisted@python.org/thread/L6GEKQYSWUX4NBIHYEML3AKB2PXDD56N/

PR #12101, opened 2024-02-01
#11902 - Fix type for twisted.web.server.Request.defaultContentType

### LINKED ISSUE #11902: twisted.web.server.Request.defaultContentType is missing type annotation
MyPy assumes it's always `bytes`, but docs explain it can also be `None`, so really it should be `Optional[bytes]`.

eevelweezel:
## Scope and purpose

Fixes #11902

Set type for twisted.web.server.Request.defaultContentType to allow bytes or None.

### DIFF
--- a/src/twisted/newsfragments/11902.misc
+++ b/src/twisted/newsfragments/11902.misc
@@ -1,0 +1,1 @@
+Change typing for twisted.web.server.Request.defaultContentType to allow None.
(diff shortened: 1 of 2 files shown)


eevelweezel:
needs review

eevelweezel:
needs-review

adiroiban (approved):
Thanks for the fix. All good! I will merge.

PR #12109, opened 2024-03-06
#12108 Speed up the twisted.web HTTP client

### LINKED ISSUE #12108: Speed up twisted.web HTTP client
The Twisted HTTP client has some code paths that aren't as optimized as one might hope, such that it should be possible to make it faster.

An initial experiment yielded at least one meaningful speedup, so I'm going to see how many more I can do.

### LINKED ISSUE #12112: Failure.__getstate__ is ignored by pickle
1. `Failure` subsclasses `BaseException`.
2. `BaseException` implements the reduce pickle API, in C.
3. As a result, you need to override `__reduce__` to impact pickling behavior.

Demonstration—notice that `Failure.__getstate__` does `c["pickled"] = 1` in its implementation in the dictionary it returns.

```python
Python 3.11.8 (main, Feb 25 2024, 16:39:33) [GCC 11.4.0] on linux
>>> import pickle
>>> from twisted.python.failure import Failure
>>> f = Failure(ZeroDivisionError())
>>> pickle.loads(pickle.dumps(f)).pickled
0
```

itamarst:
## Scope and purpose

Fixes #12108 
Fixes #12112 

Why this is faster:

* Minor reductions in general effort.
* Only calculate `Failure.parents` on demand, since it's expensive and often not used.
* `Failure.__init__` does a huge amount of unnecessary work for simple cases, so avoided it with alternative constructor. The constructor results in different pickling behavior, because of subclassing `BaseException` it seems!!!! See https://gist.github.com/itamarst/4e984bddef23a96619b37bfb929e5b24

I run my benchmark script with `hyperfine` to reduce noise from dictionary ordering randomness and the like.

Before:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      3.992 s ±  0.056 s    [User: 3.979 s, System: 0.012 s]
  Range (min … max):    3.917 s …  4.118 s    10 runs
```

After:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      3.266 s ±  0.032 s    [User: 3.253 s, System: 0.012 s]
  Range (min … max):    3.222 s …  3.336 s    10 runs
```

## ORIGINAL APPROACH, NOW OBSOLETE

**Important:** This relies on `Failure` objects being effectively immutable, which I _think_ is the case but should be something to consider when reviewing.

I run my benchmark script with `hyperfine` to reduce noise from dictionary ordering randomness and the like.

Before:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      3.992 s ±  0.056 s    [User: 3.979 s, System: 0.012 s]
  Range (min … max):    3.917 s …  4.118 s    10 runs
```

After:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      3.022 s ±  0.025 s    [User: 3.005 s, System: 0.016 s]
  Range (min … max):    2.982 s …  3.051 s    10 runs
```

Test script:

```python
from twisted.web._newclient import HTTP11ClientProtocol, Request
from twisted.internet.testing import StringTransport
import time

RESPONSE = """HTTP/1.1 200 OK
Host: blah
Foo: bar
Gaz: baz
Content-length: 3

abc""".replace("\n", "\r\n").encode("utf-8")

start = time.time()
for i in range(100_000):
    protocol = HTTP11ClientProtocol()
    protocol.makeConnection(StringTransport())
    request = Request(b"GET", b"/foo/bar", Headers({b"Host": [b"example.com"]}), None)
    d = protocol.request(request)
    protocol.dataReceived(RESPONSE)
    l = []
    d.addCallback(l.append)
    assert l
print("requests/sec:", 100_000 / (time.time() - start))
``` 

Follow-up PRs will:

- Optimize http_headers.py, which does a lot of repetitive and unnecessary work.
- Possibly split up the identity transfer encoding into two variants, one with a fixed length and one without.

## Contributor Checklist:

### DIFF
--- a/src/twisted/logger/_json.py
+++ b/src/twisted/logger/_json.py
@@ -55,7 +55,7 @@ def failureFromJSON(failureDict: JSONDict) -> Failure:
     f = Failure.__new__(Failure)
     typeInfo = failureDict["type"]
     failureDict["type"] = type(typeInfo["__name__"], (), typeInfo)
-    f.__dict__ = failureDict
+    f.__setstate__(failureDict)
     return f
 
 
--- a/src/twisted/newsfragments/12112.bugfix
+++ b/src/twisted/newsfragments/12112.bugfix
@@ -1,1 +1,1 @@
-Fix bug where ``twisted.python.failure.Failure``'s custom pickling code wasn't being used.
+twisted.python.failure.Failure once again utilizes the custom pickling logic it used to in the past.
--- a/src/twisted/newsfragments/12112.bugfix
+++ b/src/twisted/newsfragments/12112.bugfix
@@ -1,0 +1,1 @@
+Fix bug where ``twisted.python.failure.Failure``'s custom pickling code wasn't being used.
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -250,7 +250,7 @@ class Failure(BaseException):
 
     pickled = 0
     stack = None
-    __parents = None
+    _parents = None
 
     # The opcode of "yield" in Python bytecode. We need this in
     # _findFailure in order to identify whether an exception was
@@ -419,15 +419,15 @@ class Failure(BaseException):
 
     @property
     def parents(self):
-        if self.__parents is not None:
-            return self.__parents
+        if self._parents is not None:
+            return self._parents
 
         if inspect.isclass(self.type) and issubclass(self.type, Exception):
             parentCs = getmro(self.type)
-            self.__parents = list(map(reflect.qual, parentCs))
+            self._parents = list(map(reflect.qual, parentCs))
         else:
-            self.__parents = [self.type]
-        return self.__parents
+            self._parents = [self.type]
+        return self._parents
 
     def _extrapolate(self, otherFailure):
         """
@@ -613,11 +613,25 @@ class Failure(BaseException):
     def __str__(self) -> str:
         return "[Failure instance: %s]" % self.getBriefTraceback()
 
+    def __setstate__(self, state):
+        state["_parents"] = state.pop("parents")
+        self.__dict__.update(state)
+
     def __getstate__(self):
-        """Avoid pickling objects in the traceback."""
-        if self.pickled:
-            return self.__dict__
+        """
+        Avoid pickling objects in the traceback.
+
+        This is not called direclty by pickle, since C{BaseException}
+        implements reduce; instead, pickle calls C{Failure.__reduce__} which
+        then calls this API.
+        """
+        # Make sure _parents field is populated:
+        _ = self.parents
+
         c = self.__dict__.copy()
+
+        # Backwards compatibility with old code, e.g. for Perspective Broker:
+        c["parents"] = c.pop("_parents")
 
         c["frames"] = [
             [
@@ -649,6 +663,11 @@ class Failure(BaseException):
 
         c["pickled"] = 1
         return c
+
+    def __reduce__(self):
+        # BaseException implements a __reduce__ (in C, technically), so TODO
+        from functools import partial
+        return (partial(Failure.__new__, Failure), (), self.__getstate__())
 
     def cleanFailure(self):
         """
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -16,10 +16,10 @@ from __future__ import annotations
 # System Imports
 import builtins
 import copy
-from functools import partial
 import inspect
 import linecache
 import sys
+from functools import partial
 from inspect import getmro
 from io import StringIO
 from typing import Callable, NoReturn, TypeVar
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -459,7 +459,7 @@ class Failure(BaseException):
         self.frames = frames
 
     @staticmethod
-    def without_traceback(value: BaseException) -> Failure:
+    def _without_traceback(value: BaseException) -> Failure:
         """
         Create a L{Failure} for an exception without a traceback.
 
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -472,7 +472,7 @@ class Failure(BaseException):
         result.captureVars = False
         result.count = count
         result.frames = []
-        result.stack = []
+        result.stack = []  # type: ignore
         result.value = value
         result.type = value.__class__
         result.tb = None
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -249,6 +249,7 @@ class Failure(BaseException):
 
     pickled = 0
     stack = None
+    __parents = None
 
     # The opcode of "yield" in Python bytecode. We need this in
     # _findFailure in order to identify whether an exception was
@@ -417,11 +418,18 @@ class Failure(BaseException):
                 )
             )
             tb = tb.tb_next
+
+    @property
+    def parents(self):
+        if self.__parents is not None:
+            return self.__parents
+
         if inspect.isclass(self.type) and issubclass(self.type, Exception):
             parentCs = getmro(self.type)
-            self.parents = list(map(reflect.qual, parentCs))
+            self.__parents = list(map(reflect.qual, parentCs))
         else:
-            self.parents = [self.type]
+            self.__parents = [self.type]
+        return self.__parents
 
     def _extrapolate(self, otherFailure):
         """
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -289,9 +289,6 @@ class Failure(BaseException):
         self.type = self.value = tb = None
         self.captureVars = captureVars
 
-        if isinstance(exc_value, str) and exc_type is None:
-            raise TypeError("Strings are not supported by Failure")
-
         stackOffset = 0
 
         if exc_value is None:
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -11,6 +11,7 @@ Asynchronous-friendly error mechanism.
 See L{Failure}.
 """
 
+from __future__ import annotations
 
 # System Imports
 import builtins
@@ -456,6 +457,26 @@ class Failure(BaseException):
         # Merging current stack with stack stored in the Failure.
         frames.extend(self.frames)
         self.frames = frames
+
+    @staticmethod
+    def without_traceback(value: Exception) -> Failure:
+        """
+        Create a L{Failure} for an exception without a traceback.
+
+        By restricting the inputs significantly, this constructor runs much
+        faster.
+        """
+        result = Failure.__new__(Failure)
+        global count
+        count += 1
+        result.captureVars = False
+        result.count = count
+        result.frames = []
+        result.stack = []
+        result.value = value
+        result.type = value.__class__
+        result.tb = None
+        return result
 
     def trap(self, *errorTypes):
         """
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -667,6 +667,7 @@ class Failure(BaseException):
     def __reduce__(self):
         # BaseException implements a __reduce__ (in C, technically), so TODO
         from functools import partial
+
         return (partial(Failure.__new__, Failure), (), self.__getstate__())
 
     def cleanFailure(self):
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 # System Imports
 import builtins
 import copy
+from functools import partial
 import inspect
 import linecache
 import sys
@@ -665,9 +666,8 @@ class Failure(BaseException):
         return c
 
     def __reduce__(self):
-        # BaseException implements a __reduce__ (in C, technically), so TODO
-        from functools import partial
-
+        # BaseException implements a __reduce__ (in C, technically), so we need
+        # to override this to get pickling working.
         return (partial(Failure.__new__, Failure), (), self.__getstate__())
 
     def cleanFailure(self):
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -754,6 +754,35 @@ class Failure(BaseException):
         """
         self.printTraceback(file, elideFrameworkCode, detail="verbose")
 
+    def _freeze(self) -> "_FrozenFailure":
+        """
+        Create a frozen instance, consuming this one.
+        """
+        result = _FrozenFailure.__new__(_FrozenFailure)
+        result.__dict__ = self.__dict__
+        # Destroy this failure, just to be sure:
+        self.__dict__ = {}
+        return result
+
+
+class _FrozenFailure(Failure):
+    """
+    A L{Failure} that shouldn't be mutated.
+    """
+
+    def __setattr__(self, key, value):
+        if key == "__dict__":
+            object.__setattr__(self, key, value)
+            return
+        raise RuntimeError(
+            f"Tried to set {key} to {value}, but this object should not be mutated"
+        )
+
+    def __delattr__(self, key):
+        raise RuntimeError(
+            f"Tried to delete attribute {key}, but this object should not be mutated"
+        )
+
 
 def _safeReprVars(varsDictItems):
     """
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -459,7 +459,7 @@ class Failure(BaseException):
         self.frames = frames
 
     @staticmethod
-    def without_traceback(value: Exception) -> Failure:
+    def without_traceback(value: BaseException) -> Failure:
         """
         Create a L{Failure} for an exception without a traceback.
 
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -460,7 +460,7 @@ class Failure(BaseException):
         self.frames = frames
 
     @staticmethod
-    def _without_traceback(value: BaseException) -> Failure:
+    def _withoutTraceback(value: BaseException) -> Failure:
         """
         Create a L{Failure} for an exception without a traceback.
 
--- a/src/twisted/python/failure.py
+++ b/src/twisted/python/failure.py
@@ -754,35 +754,6 @@ class Failure(BaseException):
         """
         self.printTraceback(file, elideFrameworkCode, detail="verbose")
 
-    def _freeze(self) -> "_FrozenFailure":
-        """
-        Create a frozen instance, consuming this one.
-        """
-        result = _FrozenFailure.__new__(_FrozenFailure)
-        result.__dict__ = self.__dict__
-        # Destroy this failure, just to be sure:
-        self.__dict__ = {}
-        return result
-
-
-class _FrozenFailure(Failure):
-    """
-    A L{Failure} that shouldn't be mutated.
-    """
-
-    def __setattr__(self, key, value):
-        if key == "__dict__":
-            object.__setattr__(self, key, value)
-            return
-        raise RuntimeError(
-            f"Tried to set {key} to {value}, but this object should not be mutated"
-        )
-
-    def __delattr__(self, key):
-        raise RuntimeError(
-            f"Tried to delete attribute {key}, but this object should not be mutated"
-        )
-
 
 def _safeReprVars(varsDictItems):
     """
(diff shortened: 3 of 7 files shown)


itamarst:
please review

src/twisted/web/_newclient.py
```diff
@@ -929,6 +928,7 @@ def writeTo(self, transport):
                 self._writeToEmptyBodyContentLength(transport)
             else:
                 self._writeHeaders(transport, None)
+            return succeed(None)
         elif self.bodyProducer.length is UNKNOWN_LENGTH:
```
adiroiban, line 932: Can we conver this to an `if` ?

Not sure if there is a performance implicatin, but it should make the code easier to read.


```suggestion

        if self.bodyProducer.length is UNKNOWN_LENGTH:
```

src/twisted/web/_newclient.py
```diff
@@ -929,6 +928,7 @@ def writeTo(self, transport):
...
             return self._writeToBodyProducerChunked(transport)
         else:
```
adiroiban, line 934: Can we aslo remove this else?
Not sure if it has any performance impact. Should make the code easier to read

src/twisted/web/_newclient.py
```diff
@@ -1270,15 +1273,15 @@ def _bodyDataFinished_INITIAL(self, reason=None):
         """
         self._state = "DEFERRED_CLOSE"
         if reason is None:
-            reason = Failure(ResponseDone("Response body fully received"))
+            reason = _RESPONSE_DONE_FAILURE
```
adiroiban: I guess that in this way we loose the traceback from the `Failure`. I think that each Failure, at initialization, will keep a copy of the  current traceback.

I guess that the main issue here is `Failure.__init__` doing all the traceback/parent search.

One suggestion.

Since the traceback is no longer usable with this refactor, would it help to have something like `twisted.python.failure.FailureWithoutTraceback` implementation, that is still a mutable failure, but has no initial traceback ?

We would still have an object initialized, but that should be faster.
I think that `Failure` without tracebacks is a valid instance.

-----------

I think that for the purpose of pefromance, it's ok not to worry about tracebacks.

But maybe we should have 2 separate  pre-created failures here.

Maybe we can have something like this, just to help with future troubleshooting and logs:

* Response body fully received in first packet.
* Response body fully received

The suggested text might not be 100% accurate. 

Since this is not really a faiulure, I think it's ok not to worry about the traceback.

src/twisted/web/_newclient.py
```diff
@@ -1605,7 +1615,7 @@ def _finishResponse_WAITING(self, rest):
                 # keeping persistent connections around is an optimisation:
                 self._log.failure("")
                 self.transport.loseConnection()
-            self._disconnectParser(reason)
+            self._disconnectParser(_DONE)
```
adiroiban: I am not sure why we need this change.

adiroiban (commented):
Many thanks for the changes.

For now, I just left this is a comment, and put back the team for review.

Happy to see performance improvements in the HTTP client.

I left a few comments inline.

As you commented in the description of the PR, my biggest concern is related to mutating the failure instance.

----------


I remember seing instances in which a failure is mutated, but I don't remember the exact code or the ticket/issue number.

I think that for the purpose of the current code, it's ok to this change.

I expect that at least some downstream application/library that relies on Twisted is mutating failures

I think I have some places in my  applications in which I mutate `Failure.value`. This was done for convenience...but I think the right thing to do is to have someting like `newFailure = oldFailure.copy(Exception('new exception'))` in which the traceback is kept.

I am doing this type of things mostly in high level application code.
The error messsage text generated by Twisted is not good enought for end users... but I still want to keep the traceback for troubleshoting 

------------

Maybe we can create a new `FrozenFailure` class and use it here.
The `FrozenFailure` will raise an error is someone tries to mutate it.

itamarst:
It's not clear to me there _is_ a traceback in this case, since it's directly creating exceptions. Will validate tomorrow.

adiroiban:
> It's not clear to me there is a traceback in this case, since it's directly creating exceptions. Will validate tomorrow.

You are right.
There is no traceback here, so traceback is not an issue.

Still, maybe it's a good idea to have different text messages, depending on the place from where this failure is returned.

src/twisted/web/_newclient.py
itamarst: Because there's no `reason` variable anymore.

src/twisted/web/_newclient.py
itamarst: I don't see why it'd make things more readable?

itamarst: Oh I see, based on the previous suggestion.

src/twisted/web/_newclient.py
itamarst: I tried it, I think it makes the code a little less clear.

itamarst:
So, the main issue is mutable `Failure`. `FrozenFailure` is certainly possible, but... it'd have to make the exception frozen somehow, when it's not clear how. So a generic version is hard. And any mutable exception attributes can be mutated even if `__setattr__` is disabled... so a generic version seems difficult.

On the other hand, here it's just `ResponseDone` and `ConnectionDone` which don't have mutable attributes, really, and oughtn't be mutated. So might be doable in this case.

itamarst:
Actually, there's a number of places that just assume immutable exceptions already, e.g. the `CONNECTION_DONE` constant, so maybe just read-only `Failure` is sufficient.

itamarst:
OK, added a Failure subclass.

glyph:
> OK, added a Failure subclass.

This is *probably* fine, but thinking about it does send a chill down my spine.

I don't think that an immutable failure *subclass* is actually substitutable, or thereby compliant with the letter of the compatibility policy, as it stands. I'd kinda rather make *all* Failures be immutable if that is practical, and initially enforce it at type-time, which is not covered by the compatibility policy?

exarkun:
> > OK, added a Failure subclass.
> 
> This is _probably_ fine, but thinking about it does send a chill down my spine.
> 

*cough* CopyableFailure *cough*

> I don't think that an immutable failure subclass is actually substitutable, or thereby compliant with the letter of the compatibility policy, as it stands. I'd kinda rather make all Failures be immutable if that is practical, and initially enforce it at type-time, which is not covered by the compatibility policy?

It's certainly not.  I think _only_ enforcing it at type-time is probably good enough.  There will never be a 100% immutable Python object.  If the type checker tells someone they're not allowed to change something and they go ahead and run code that changes it anyway, that seems like a choice that's up to them.

Ideally, there'd be a generic like `typing.Readonly` - as in, `Readonly[Failure]`.  I don't think there is, but a `ReadonlyFailure` protocol that exposes all of `Failure`'s public attributes, but only for reading is still simple, just more typing.  Then I believe the definition of `freeze` is:

```
    def freeze(self) -> ReadonlyFailure: return self
```

exarkun:
(This PR also seems like a weird place to be solving this 24 year old design mistake)

itamarst:
https://github.com/python/typing/discussions/1123 has an example of how to define a Protocol with read-only attributes. But actually these days you can just use `Final`, which is simpler.

itamarst:
Given the Failure object is going to end up in places, e.g. `Protocol.connectionLost` for a body protocol of a `Response`, using typing to enforce read-only-ness might end up going pretty far in terms of changing the API.

glyph:
> _cough_ CopyableFailure _cough_

Okay, perhaps a comparable amount of chills, but it *is* a proper Liskov substitute ;)

glyph:
> Then I believe the definition of `freeze` is:

My suggestion would be to just have `Failure` itself be frozen by default, because:

1. Mutating a failure is a pretty weird operation that I would guess rarely happens?
2. Any code doing the mutating would continue to work at runtime, unless it mutates one of these shared instances (but what mutation would such code be making, I wonder?) in which case behavior would change, but not in a technically incompatible way.
3. Any code doing the mutating would become a type error, which seems like a reasonable way for users to catch it and fix things.

If my first assumption is wrong (this would not affect much code) then we should probably do another approach. Although then I'd really want to know what the common failure-mutation patterns are.

src/twisted/web/_newclient.py
adiroiban: We can leave it as it is. No problem. Thanks for looking into this.

src/twisted/web/newsfragments/12108.feature
```diff
@@ -0,0 +1 @@
+The HTTP client now runs faster by using a little less CPU.
```
adiroiban: Maybe something like this to comply with the convension for release notes

I guess that this is the main public API

```suggestion
twisted.web.client.HTTPConnectionPool used by HTTP clients now runs faster by using a little less CPU.
```

adiroiban (approved):
Thanks for the changes.

I only have a minor comment regarding the release notes wording.

Based on the latest discussions, I think that we can merge this PR as it is.

Also, I am also ok with merging this code with a standard `Failure`, just like it was in the first review request.

I have approved this PR, but maybe wait one day to receive more feedback from Jean-Paul and Glyph.

----------

> (This PR also seems like a weird place to be solving this 24 year old design mistake)

We don't have to make big API changes in this PR.
I think it is still usefull to discuss this issue.

I was not able to find an existing ticket/GitHub Issue discussing the read only design for Failure.

I guess we should create a new one now :)

> I'd kinda rather make all Failures be immutable if that is practical,

I think that such a change should be done in a separate PR.

So my suggestion for this PR:

* Create a new GitHub issue dedicated to the Failure read-only design (or find the existing one, if such an issue was already raised)
* Commit this PR using using `_FrozenFailure` and add a note to remove it once the Failure read-only changes are made

Thanks again!

itamarst:
There is one existing Failure API that mutates it, https://docs.twisted.org/en/stable/api/twisted.python.failure.Failure.html#cleanFailure

itamarst:
Based on above discussion, before proceeding on merging this PR I'm going to:

1. Open a separate issue about make `Failure` (and relevant exceptions) immutable.
2. Make a sketch on a new branch.
3. Maybe even get it to review stage, we'll see.

If it still seems plausible then we'll see what order of merging separate PRs makes sense.

itamarst:
I opened #12111 for the immutable `Failure` case, would appreciate if people on this thread check it to see if I missed any important comments from the discussion here.

itamarst:
I gave up on reusing `Failure` instances and went with a different approach.

itamarst:
please review

itamarst:
please review

itamarst:
MyPy failures don't appear to have anything to do with this branch...

glyph:
> MyPy failures don't appear to have anything to do with this branch...

Confirmed, fresh trunk checkout gives similar errors here. Investigating…

glyph:
> > MyPy failures don't appear to have anything to do with this branch...
> 
> Confirmed, fresh trunk checkout gives similar errors here. Investigating…

filed https://github.com/twisted/twisted/issues/12113 to track and discuss

glyph:
@itamarst an attempted fix is here https://github.com/twisted/twisted/pull/12115

glyph:
OK, fix attempt seems to have worked, hopefully that can unblock this

glyph:
Merging in the approved fix, so maybe we can get a successful run.

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
             "<twisted.python.failure.Failure " "%s: division by zero>" % (typeName,),
         )
 
+    def test_failure_without_traceback(self) -> None:
```
glyph:
```suggestion
    def test_failureWithoutTraceback(self) -> None:
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
...
+        self.assertIsInstance(failure2, failure.Failure)
+        dict2 = failure2.__dict__.copy()
+
+        # count increments with each new Failure constructed:
+        self.assertEqual(dict1.pop("count") + 1, dict2.pop("count"))
+
+        # The rest of the attributes should be identical:
+        self.assertEqual(dict1, dict2)
+
+    def test_failure_pickling(self) -> None:
```
glyph:
```suggestion
    def test_failurePickling(self) -> None:
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
...
+        """
+        The failure can be pickled and unpickled, and the C{parents} attribute
+        is included in the pickle.
+        """
+        failure2 = pickle.loads(pickle.dumps(original_failure))
+        expected = original_failure.__dict__.copy()
+        expected["pickled"] = 1
+        self.assertEqual(expected, failure2.__dict__)
+
+    def test_failure_pickling_includes_parents(self) -> None:
```
glyph:
```suggestion
    def test_failurePicklingIncludesParents(self) -> None:
```

src/twisted/python/failure.py
```diff
@@ -452,6 +459,26 @@ def _extrapolate(self, otherFailure):
         frames.extend(self.frames)
         self.frames = frames
 
+    @staticmethod
+    def _without_traceback(value: BaseException) -> Failure:
```
glyph:
```suggestion
    def _withoutTraceback(value: BaseException) -> Failure:
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
             "<twisted.python.failure.Failure " "%s: division by zero>" % (typeName,),
         )
 
+    def test_failure_without_traceback(self) -> None:
+        """
+        C{Failure._without_traceback(exc)} gives the same result as
```
glyph:
```suggestion
        C{Failure._withoutTraceback(exc)} gives the same result as
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
...
+        C{Failure(exc)}.
+        """
+        exc = ZeroDivisionError("hello")
+        dict1 = failure.Failure(exc).__dict__.copy()
+        failure2 = failure.Failure._without_traceback(exc)
```
glyph:
```suggestion
        failure2 = failure.Failure._withoutTraceback(exc)
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
...
+
+        # count increments with each new Failure constructed:
+        self.assertEqual(dict1.pop("count") + 1, dict2.pop("count"))
+
+        # The rest of the attributes should be identical:
+        self.assertEqual(dict1, dict2)
+
+    def test_failure_pickling(self) -> None:
+        """
+        C{Failure(exc)} and C{Failure._without_traceback(exc)} can be pickled
```
glyph:
```suggestion
        C{Failure(exc)} and C{Failure._withoutTraceback(exc)} can be pickled
```

src/twisted/test/test_failure.py
```diff
@@ -517,6 +522,55 @@ def test_repr(self) -> None:
...
+        and unpickled.
+        """
+        exc = ComparableException("hello")
+        failure1 = failure.Failure(exc)
+        self.assertPicklingRoundtrips(failure1)
+
+        # You would think this test is unnecessary, since it's just a
+        # C{Failure}, but actually the behavior of pickling can sometimes be
+        # different because of the way the constructor works!
+        failure2 = failure.Failure._without_traceback(exc)
```
glyph:
```suggestion
        failure2 = failure.Failure._withoutTraceback(exc)
```

src/twisted/web/_newclient.py
```diff
@@ -1270,15 +1270,19 @@ def _bodyDataFinished_INITIAL(self, reason=None):
         """
         self._state = "DEFERRED_CLOSE"
         if reason is None:
-            reason = Failure(ResponseDone("Response body fully received"))
+            reason = Failure._without_traceback(
```
glyph:
```suggestion
            reason = Failure._withoutTraceback(
```

src/twisted/web/_newclient.py
```diff
@@ -1270,15 +1270,19 @@ def _bodyDataFinished_INITIAL(self, reason=None):
...
+            )
         self._reason = reason
 
     def _bodyDataFinished_CONNECTED(self, reason=None):
         """
         Disconnect the protocol and move to the C{'FINISHED'} state.
         """
         if reason is None:
-            reason = Failure(ResponseDone("Response body fully received"))
+            reason = Failure._without_traceback(
```
glyph:
```suggestion
            reason = Failure._withoutTraceback(
```

src/twisted/web/_newclient.py
```diff
@@ -1590,7 +1597,7 @@ def _finishResponse_WAITING(self, rest):
             or self._state != "QUIESCENT"
             or not self._currentRequest.persistent
         ):
-            self._giveUp(Failure(reason))
+            self._giveUp(Failure._without_traceback(reason))
```
glyph:
```suggestion
            self._giveUp(Failure._withoutTraceback(reason))
```

glyph (commented):
just some coding-standard compliance suggestions

src/twisted/logger/_json.py
```diff
@@ -55,7 +55,7 @@ def failureFromJSON(failureDict: JSONDict) -> Failure:
     f = Failure.__new__(Failure)
     typeInfo = failureDict["type"]
     failureDict["type"] = type(typeInfo["__name__"], (), typeInfo)
-    f.__dict__ = failureDict
+    f.__setstate__(failureDict)
```
adiroiban, line 58: Is this API used ?

In `src/twisted/logger/__init__.py` I can see only these as public API

```
from ._json import (
    eventAsJSON,
    eventFromJSON,
    jsonFileLogObserver,
    eventsFromJSONLogFile,
)

```

and I don't see any test for this code.

src/twisted/python/failure.py
```diff
@@ -417,11 +417,18 @@ def __init__(self, exc_value=None, exc_type=None, exc_tb=None, captureVars=False
                 )
             )
             tb = tb.tb_next
+
+    @property
+    def parents(self):
```
adiroiban, line 422: I know that in `trunk` we alrady have `Failure.parents` as public API and without documentation.

Can we add a few words about `Failure.parents` in this PR?

The documentation should go in `Failure.__doc__` and not here... but I can't leave a comment there.

src/twisted/newsfragments/12112.bugfix
```diff
@@ -0,0 +1 @@
+Fix bug where ``twisted.python.failure.Failure``'s custom pickling code wasn't being used.
```
adiroiban: Just a minor suggest to comply with the release notes format for bugfix

```suggestion
twisted.python.failure.Failure can now be serialized using pickling.
```

src/twisted/test/test_failure.py
```diff
@@ -24,6 +25,18 @@
...
+class ComparableException(Exception):
+    """An exception that can be compared by value."""
+
+    def __eq__(self, other: object) -> bool:
+        return (self.__class__ == other.__class__) and (
+            self.args == cast(ComparableException, other).args
+        )
+
+    def __ne__(self, other: object) -> bool:
+        return not self.__eq__(other)
```
adiroiban: It looks like this code is not used. Can we removed it ?

src/twisted/logger/_json.py
itamarst: The tests were failing without this, so I think so.

src/twisted/python/failure.py
itamarst: It's an implementation detail, it's better if people _don't_ use it. The fact it's public is a historic quirk...

src/twisted/test/test_failure.py
itamarst: My worry is someone tweaks the test logic slightly and then it starts acting weirdly because of the lack of `__ne__`. But I just learned that they're optional in Python 3, lacking one it will DTRT, so will do.

src/twisted/python/failure.py
adiroiban: Thanks for the info.

If this should be private, then maybe we need a followup issue to mark it private and raise a deprecation warning for any public usage of the API ?

PR #12117, opened 2024-03-15
#12116 Speed up http_headers operations

### LINKED ISSUE #12116: Speed up http_headers operations
Handling HTTP headers is a significant percentage of time spent in the HTTP client (and probably the HTTP server as well). In particular, all the normalization (which happens both incoming and outgoing!) adds up.

itamarst:
## Scope and purpose

Fixes #12116

The most significant changes are:

1. Change the internal representation of headers, so they're canonicalized once instead of lowercased incoming and canonicalized outgoing.
2. Cache canonicalization, with limits to prevent high memory usage by attackers.
3. Get rid of runtime type checks, given static type checking is now available in Python.

Same benchmark as #12108, measuring sending and parsing a HTTP client request. This will likely speed up the HTTP server as well.

Before:

```
  Time (mean ± σ):      3.262 s ±  0.027 s    [User: 3.250 s, System: 0.012 s]
  Range (min … max):    3.223 s …  3.307 s    10 runs
```

After:

```
  Time (mean ± σ):      2.929 s ±  0.036 s    [User: 2.918 s, System: 0.011 s]
  Range (min … max):    2.877 s …  3.016 s    10 runs
```

### DIFF
--- a/src/twisted/web/http_headers.py
+++ b/src/twisted/web/http_headers.py
@@ -54,8 +54,11 @@ class Headers:
     and values as opaque byte strings.
 
     @cvar _caseMappings: A L{dict} that maps lowercase header names
-        to their canonicalized representation. Additional entries may be added,
-        as it is used as a cache.
+        to their canonicalized representation, for headers with unconventional
+        capitalization.
+
+    @cvar _canonicalHeaderCache: A L{dict} that maps header names to their
+        canonicalized representation.
 
     @ivar _rawHeaders: A L{dict} mapping header names as L{bytes} to L{list}s of
         header values as L{bytes}.
@@ -70,6 +73,8 @@ class Headers:
         b"www-authenticate": b"WWW-Authenticate",
         b"x-xss-protection": b"X-XSS-Protection",
     }
+
+    _canonicalHeaderCache: Dict[Union[bytes, str], bytes] = {}
 
     __slots__ = ["_rawHeaders"]
 
@@ -104,16 +109,39 @@ class Headers:
 
     def _encodeName(self, name: Union[str, bytes]) -> bytes:
         """
-        Encode the name of a header (eg 'Content-Type') to an ISO-8859-1 encoded
-        bytestring if required.
+        Encode the name of a header (eg 'Content-Type') to an ISO-8859-1
+        encoded bytestring if required.  It will be canonicalized and
+        whitespace-sanitized.
 
         @param name: A HTTP header name
 
         @return: C{name}, encoded if required, lowercased
         """
+        if canonicalName := self._canonicalHeaderCache.get(name, None):
+            return canonicalName
+
         if isinstance(name, str):
-            return name.lower().encode("iso-8859-1")
-        return name.lower()
+            bytes_name = name.encode("iso-8859-1")
+        else:
+            bytes_name = name
+
+        if bytes_name.lower() in self._caseMappings:
+            # Some headers have special capitalization:
+            result = self._caseMappings[bytes_name.lower()]
+        else:
+            result = _sanitizeLinearWhitespace(
+                b"-".join([word.capitalize() for word in bytes_name.split(b"-")])
+            )
+
+        # In general, we should only see a very small number of header
+        # variations in the real world, so caching them is fine. However, an
+        # attacker could generate infinite header variations to fill up RAM, so
+        # we cap how many we cache. The performance degradation from lack of
+        # caching won't be that bad, and legit traffic won't hit it.
+        if len(self._canonicalHeaderCache) < 10_000:
+            self._canonicalHeaderCache[name] = result
+
+        return result
 
     def copy(self):
         """
@@ -171,7 +199,7 @@ class Headers:
 
         @return: L{None}
         """
-        _name = _sanitizeLinearWhitespace(self._encodeName(name))
+        _name = self._encodeName(name)
         encodedValues: List[bytes] = []
         for v in values:
             if isinstance(v, str):
@@ -190,9 +218,7 @@ class Headers:
 
         @param value: The value to set for the named header.
         """
-        self._rawHeaders.setdefault(
-            _sanitizeLinearWhitespace(self._encodeName(name)), []
-        ).append(
+        self._rawHeaders.setdefault(self._encodeName(name), []).append(
             _sanitizeLinearWhitespace(
                 value.encode("utf8") if isinstance(value, str) else value
             )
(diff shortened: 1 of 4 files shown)


itamarst:
please review

src/twisted/web/http_headers.py
```diff
@@ -112,16 +108,39 @@ def __cmp__(self, other):
...
+            result = _sanitizeLinearWhitespace(
+                b"-".join([word.capitalize() for word in bytes_name.split(b"-")])
+            )
+
+        # In general, we should only see a very small number of header
+        # variations in the real world, so caching them is fine. However, an
+        # attacker could generate infinite header variations to fill up RAM, so
+        # we cap how many we cache. The performance degradation from lack of
+        # caching won't be that bad, and legit traffic won't hit it.
+        if len(self._canonicalHeaderCache) < 10_000:
```
glyph: can you make this a parameter, mostly so that we can get the test coverage for the negative branch?  (But also, threading it through so that users can actually customize the value from the client or server would be nice, we tend to be pretty bad about presenting tunables properly.)

src/twisted/web/http_headers.py
```diff
@@ -81,6 +73,10 @@ class Headers:
         b"x-xss-protection": b"X-XSS-Protection",
     }
 
+    _canonicalHeaderCache: Dict[Union[bytes, str], bytes] = {}
+
+    __slots__ = ["_rawHeaders"]
```
glyph, line 81: If we can compatibly go from non-slotted to slotted (can we?) could we just go full-blown `@dataclass` and save ourselves some headaches here, getting rid of the custom `@comparable`?

src/twisted/web/http_headers.py
```diff
@@ -81,6 +73,10 @@ class Headers:
         b"x-xss-protection": b"X-XSS-Protection",
     }
 
+    _canonicalHeaderCache: Dict[Union[bytes, str], bytes] = {}
```
glyph: If we are going to have a class-level (i.e., global & mutable) variable here, let's make sure to be explicit about it.

```suggestion
    _canonicalHeaderCache: ClassVar[Dict[Union[bytes, str], bytes]] = {}
```

src/twisted/web/http_headers.py
```diff
@@ -81,6 +73,10 @@ class Headers:
         b"x-xss-protection": b"X-XSS-Protection",
```
glyph, line 74: Hmm. Slightly tangential, but it strikes me that `_caseMappings` should probably have a `ClassVar[Mapping[...]]` annotation here, we don't ever want to mutate this one.

src/twisted/web/http_headers.py
```diff
@@ -112,16 +108,39 @@ def __cmp__(self, other):
...
+        Encode the name of a header (eg 'Content-Type') to an ISO-8859-1
+        encoded bytestring if required.  It will be canonicalized and
+        whitespace-sanitized.
 
         @param name: A HTTP header name
 
         @return: C{name}, encoded if required, lowercased
         """
+        if canonicalName := self._canonicalHeaderCache.get(name, None):
+            return canonicalName
```
glyph, line 123: love a good walrus

src/twisted/web/http_headers.py
```diff
@@ -112,16 +108,39 @@ def __cmp__(self, other):
...
+
         if isinstance(name, str):
-            return name.lower().encode("iso-8859-1")
-        return name.lower()
+            bytes_name = name.encode("iso-8859-1")
```
glyph: I find ternaries easier to read for this sort of thing, and I think that `charmap` better communicates the role of the encoding here, so, if it's all the same to you,

(can't do this as a "suggestion" because of the deleted line")

```python3
bytes_name = name.encode("charmap") if isinstance(name, str) else name
```

src/twisted/web/newsfragments/12116.feature
```diff
@@ -0,0 +1 @@
+twisted.web.http_headers now uses less CPU.
```
glyph: Can you put a rough number on this? "Making a small HTTP request X% faster on average?"

glyph (approved):
The two major things to make sure we have sorted out:

- is `__slots__` a compatibility issue? we don't need to have a big discussion, I leave it to you to make the judgement, I just want to make sure you've considered it, and
- the coverage gap

Other various suggestions inline but otherwise I'm quite happy to see this.

src/twisted/web/http_headers.py
itamarst: I am not sure it needs to be tunable? There's only so many HTTP headers in the real world. So just need it to be large enough that any real-world usage never runs out, and small enough that memory usage isn't a big deal. Could make it 100K by default just as as easily.

I'll start with a private class attribute and if anyone ever complains it can be made public.

adiroiban:
Just a FYI. We have this PR https://github.com/twisted/twisted/pull/12094 to look into increasing the limit for headers

I didn't had time to follow up with it.

@itamarst  or @glyph  maybe you can take a look and see if we can merge it

I am +1 for Itamar
I am ok to have hardcoded limits, as long as they can handle all real world usage.

src/twisted/web/http_headers.py
itamarst: The `@dataclass` made it slower (assuming you mean the attrs one).

The backwards incompatibility is you can't add random attributes to the class, which ... you really shouldn't. Subclasses can still add whatever attributes they want. So it seems fine to me.

itamarst: Actually `@dataclass` is fine, never mind.

itamarst: But also it has issues with `__init__`, and its the deprecated attrs API, and I don't really want to spend the time figuring this out.

src/twisted/web/http_headers.py
itamarst: > Historically, HTTP has allowed field content with text in the
>   ISO-8859-1 charset [[ISO-8859-1](https://datatracker.ietf.org/doc/html/rfc7230#ref-ISO-8859-1)], supporting other charsets only
 >  through use of [[RFC2047](https://datatracker.ietf.org/doc/html/rfc2047)] encoding.  In practice, most HTTP header
 >  field values use only a subset of the US-ASCII charset [[USASCII](https://datatracker.ietf.org/doc/html/rfc7230#ref-USASCII)].
>   Newly defined header fields SHOULD limit their field values to
>   US-ASCII octets.  A recipient SHOULD treat other octets in field
>  content (obs-text) as opaque data.

So I'll stick to iso-8859-1.

PR #12123, opened 2024-03-27
#12122 Speed up Deferred by removing assertions

### LINKED ISSUE #12122: Remove isinstance assertions in Deferred
The Deferred implementation has a bunch of assertions about inputs. These assertions slow down code. We could remove them, relying on type checking to catch these problems.

itamarst:
## Scope and purpose

Fixes #12122

This meaningfully impacts the HTTP client benchmark I've been using. Before:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      2.942 s ±  0.026 s    [User: 2.931 s, System: 0.011 s]
  Range (min … max):    2.902 s …  2.989 s    10 runs
```

After:

```
$ hyperfine --warmup 1 "python http_client_benchmark.py"
  Time (mean ± σ):      2.847 s ±  0.023 s    [User: 2.834 s, System: 0.012 s]
  Range (min … max):    2.807 s …  2.893 s    10 runs
```

## Some reasons not to merge this!

1. ~Some of these issues are caught by type checking, but not all. In particular, the assertion that the result is not a Deferred isn't caught by type checking, I think.~
2. ~Users could just run with PYTHONOPTIMIZE=2. In practice, that slows down execution(!), not sure why. And I suspect no one runs this way.~
3. Users who don't use type checking will have harder time.

### DIFF
--- a/src/twisted/newsfragments/12122.feature
+++ b/src/twisted/newsfragments/12122.feature
@@ -1,0 +1,1 @@
+``twisted.internet.defer.Deferred`` operations use less CPU.
--- a/src/twisted/newsfragments/12122.feature
+++ b/src/twisted/newsfragments/12122.feature
@@ -1,1 +1,1 @@
-``twisted.internet.defer.Deferred`` operations use less CPU.
+twisted.internet.defer.Deferred.callback() and twisted.internet.defer.Deferred.addCallbacks() no longer use `assert` to check the type of the arguments. You should now use type checking to validate your code. These changes were done to reduce the CPU usage.
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -522,9 +522,6 @@ class Deferred(Awaitable[_SelfResultT]):
         if errbackKeywords is None:
             errbackKeywords = {}  # type: ignore[unreachable]
 
-        assert callable(callback)
-        assert callable(errback)
-
         self.callbacks.append(
             (
                 (callback, callbackArgs, callbackKeywords),
@@ -873,7 +870,6 @@ class Deferred(Awaitable[_SelfResultT]):
         @raise AlreadyCalledError: If L{callback} or L{errback} has already been
             called on this L{Deferred}.
         """
-        assert not isinstance(result, Deferred)
         self._startRunCallbacks(result)
 
     def errback(self, fail: Optional[Union[Failure, BaseException]] = None) -> None:
@@ -1552,7 +1548,6 @@ class DeferredList(  # type: ignore[no-redef] # noqa:F811
             if succeeded == SUCCESS and self.fireOnOneCallback:
                 self.callback((result, index))  # type: ignore[arg-type]
             elif succeeded == FAILURE and self.fireOnOneErrback:
-                assert isinstance(result, Failure)
                 self.errback(Failure(FirstError(result, index)))
             elif self.finishedCount == len(self.resultList):
                 # At this point, None values in self.resultList have been
--- a/src/twisted/internet/defer.py
+++ b/src/twisted/internet/defer.py
@@ -1548,6 +1548,7 @@ class DeferredList(  # type: ignore[no-redef] # noqa:F811
             if succeeded == SUCCESS and self.fireOnOneCallback:
                 self.callback((result, index))  # type: ignore[arg-type]
             elif succeeded == FAILURE and self.fireOnOneErrback:
+                assert isinstance(result, Failure)
                 self.errback(Failure(FirstError(result, index)))
             elif self.finishedCount == len(self.resultList):
                 # At this point, None values in self.resultList have been


itamarst:
please review

adiroiban:
Many thanks for the PR

I don't think `PYTHONOPTIMIZE=2` is an option

I think `PYTHONOPTIMIZE` also removes docstrings, and I think that in Twisted we have some exceptions classes that get the exception message from the class docstring.

That is, Twisted is not supported with `PYTHONOPTIMIZE=2`

----------

```diff
-       assert not isinstance(result, Deferred)
        self._startRunCallbacks(result)
```

For the change over here. I don't know what to say.
I don't remember trying to pass a deferred to `Deferred.callback()`, but I might have done it at the start of using Twisted. I can image some new developers might try to do that...and end up with unexpected results


For use cases in which this is an important performance issue, maybe you can do something like:

```
Deferred.callback = Deferred._startRunCallbacks
```

and maybe we can have something like that can be enabled for users at bootstrap:

```
@classmethod
def tuneForPerformance(cls):
    """
    Change the deferred to improve performance.
    Certain safety nets are removed, so you should make sure you code is valid.
    """
    cls.callback = cls._startRunCallbacks
```


we can write a test and make sure this functionality is supported.

In the future, we might add other performance improvements

It might be a stupid idea... but I don't know how else to fix this performance issue in a "standard" way.

So as a general rule...we might have Twisted runnig with "training wheels" by default.
Once you are happy with your code, you can remove the training wheels and expert a small performance improvement.

glyph:
> I don't remember trying to pass a deferred to `Deferred.callback()`, but I might have done it at the start of using Twisted. I can image some new developers might try to do that...and end up with unexpected results

Previously we had no other way to enforce this. Now, the answer to such confusion would be "you should run mypy on your code". It will catch mismatches like this.

glyph:
> For use cases in which this is an important performance issue,

Deferreds are in the hot loop of pretty much every Twisted application, so I think we can say that it's OK to assume that this is always a performance issue.

> maybe you can do something like:

Performance should be good by default. Users will not discover a hidden performance tunable like this, so I am -1 on such a change.

glyph:
> So as a general rule...we might have Twisted runnig with "training wheels" by default.

Mypy is the training wheels :)

glyph:
I really wish we could have repeatable performance benchmarks in CI though. I miss speed.twistedmatrix.com :-(

adiroiban:
> Previously we had no other way to enforce this. Now, the answer to such confusion would be "you should run mypy on your code". It will catch mismatches like this.

I don't know if mypy can fail if we pass a Defered instance.
The current signature is just `object`

I read something about `TypeGuard` but I don't know if it can help here... it looks like runtime code  https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards  


--------

> I really wish we could have repeatable performance benchmarks in CI though. I miss speed.twistedmatrix.com :-(

This is complicated... I don't think the GitHub VMs have concsitent performance, so I don't know how we can measure it...

Then there is the issue of persisting the results

adiroiban:
> I really wish we could have repeatable performance benchmarks in CI though. I miss speed.twistedmatrix.com :-(

With Datadog you can ingess general xunit XML report files and see the execution time for each test... and in theory you can generate a graph with the evolution over time.

I didn't had time to follow up with that...  and I don't recommend Datadog as it is expensive.

Datadog is not a turn key solution.
But Datadog can provide persistance for test results and has some tools to create graphs / reports / dashboards

glyph:
> I don't know if mypy can fail if we pass a Defered instance.

It does fail.  Consider:

```
from twisted.internet.defer import Deferred
d = Deferred[int]()
d.callback(Deferred[int]())
```
produces:

```
src/twisted/x.py:3:12: error: Argument 1 to "callback" of "Deferred" has incompatible type "Deferred[int]"; expected "int | Failure"  [arg-type]
src/twisted/x.py:3:12: note: Maybe you forgot to use "await"?
Found 1 error in 1 file (checked 1 source file)
```

glyph:
`_SelfResultT` is not the same as `object`.

adiroiban:
I see. Thanks. I wasn't aware of the mypy_zope wizardry :)

This means you need to add type checking to any existing code

```
# test.py
from twisted.internet.defer import Deferred
trigger = Deferred[bytes]()
source = Deferred[int]()
source.callback(trigger)

$ mypy test.py 
test.py:4:17: error: Argument 1 to "callback" of "Deferred" has incompatible type "Deferred[bytes]"; expected "int | Failure"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

---------

As long as others are happy with this change, in the name of performance, I am fine with this change.

If needed I can create a custom `add_training_wheels()` functions that adds back type checking.

glyph:
> If needed I can create a custom `add_training_wheels()` functions that adds back type checking.

There are actually third-party libraries that can translate type hints into assertions, like https://github.com/beartype/beartype .  Not sure how well they work, but I think a general solution is better than a few manually-added hacks like this.

itamarst:
Actually I may have been wrong and type checking _will_ catch passing a Deferred to `callback()`.

In particular, the signature of `callback()` is `callback(self, result: Union[_SelfResultT, Failure]) -> None:`. And mypy at least forces you to define the generic type of the `Deferred`, which means it will get caught.

Assuming, of course, you're using type checking.

itamarst:
Oh and apparently this exact discussion happened while I was playing around with it myself, oops.

src/twisted/newsfragments/12122.feature
```diff
@@ -0,0 +1 @@
+``twisted.internet.defer.Deferred`` operations use less CPU.
```
adiroiban: Maybe something like this to highlight which methods are affected and provide a workaround.

```suggestion
twisted.internet.defer.Deferred.callback() and twisted.internet.defer.Deferred.addCallbacks() no longer use `assert` to check the type of the arguments. You should now use type checking to validate your code. These changes were done to reduce the CPU usage.
```

adiroiban (approved):
Looks good. Thanks! Only a minor comment.

PR #12136, opened 2024-04-24
Update default DSA key length to 2048 bits

### LINKED ISSUE #12135: Continued Use of Insecure Default Parameters
Hi authors, I've noticed that some files, like [this one](https://github.com/twisted/twisted/blob/02a2b658cd1ade5d7f41f97d898913686313e615/src/twisted/conch/scripts/ckeygen.py#L152), still use insecure default values such as 1024-bit DSA keys and MD5 hashing.
These are not recommended by modern cryptographic standards due to security concerns. 
Are there plans to update to more secure alternatives?

EvanXiaa:
A simple modification that updates default DSA key length to 2048 bits. Fix #12135  .

### DIFF
--- a/src/twisted/conch/scripts/ckeygen.py
+++ b/src/twisted/conch/scripts/ckeygen.py
@@ -154,7 +154,7 @@ def generateDSAkey(options):
     from cryptography.hazmat.primitives.asymmetric import dsa
 
     if not options["bits"]:
-        options["bits"] = 1024
+        options["bits"] = 2048
     keyPrimitive = dsa.generate_private_key(
         key_size=int(options["bits"]),
         backend=default_backend(),


glyph:
"Update default DSA key length to 2048 bits" should be a bit more specific here when writing the news fragment, because this is only updating the command-line `ckeygen` default key length for generating DSA keys.  Also, as written, this doesn't really fix #12135 since that gestures broadly a "some files", and just cites this one as a specific example.

If you want to update that issue to be more specific to `ckeygen`'s default DSA key size that would be adequate, but it would be even nicer to keep a tracking issue open and link a series of more specific issues as you work through them.

Finally, a citation for an actual authority on current key size practices would be good.  2048 is fine I guess, but it would be more secure to use 4096, so why not that?   Or, you know, just future-proof it a little bit and make the key size `0x10000` and tack a new zero on every year?  Clearly there's a tradeoff there and I don't want Twisted to be in the business of becoming an authority on that, we don't have any cryptographers around qualified to keep that number up to date, so an authority we can point to that we can update along with would be much better.

(Normally, delegating to the popular thing, i.e. OpenSSH, would be my preference, but it looks like OpenSSH currently has the same default you are calling "insecure" so maybe not)

EvanXiaa:
Learned a lot. 
I would explain that my concerns are partly based on [NIST SP 800-57](https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-57pt3r1.pdf), where it says:

```
In addition, this revision adds a new section for Secure Shell (SSH). Specific changes in this revision include the following.
...
1024-bit RSA and DSA are no longer approved for generating digital signatures after 2013.
...
```
Thanks for replying. I will close this PR and consider better solutions.

glyph:
> Thanks for replying. I will close this PR and consider better solutions.

Thanks, looking forward to seeing your other ideas here! NIST is definitely not a bad place to start.

glyph:
(Weird that this is 2015 and OpenSSH did not update their default yet in response; maybe nobody is really using it to generate DSA keys any more so there wasn't interest.)

PR #12137, opened 2024-04-24
Adding cipher tag to the serverFromString

### LINKED ISSUE #12134: Adding custom cipher list to serverFromString
I have to limit the cipher from which requests could be accepted even in Tls1.2 and I was able to do it by getting the context from 'ssl.DefaultOpenSslContextFactory'
And setting the option using set_cipher_list 

I would like to set that while using serverFromString or after or before that before running the reactor 
Since Daphne is using serverFromString this would be a good addition 

I have considered making my own server to serve it and limit the cipher list but will have to migrate from Daphne to twisted
And not sure if any functionality would break doing that

KaviHarjani:
## Scope and purpose

Fixes #12134

Customising serverFromString further to handle ciphers as well, 
Daphne extends the use case for twisted by the "-e" flag, this make 
Daphe and Twisted more flexible, helping it add a static cipher to accept from 

[Futher detail here](https://github.com/twisted/twisted/issues/12134)

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3873,7 +3873,7 @@ Twisted Core 15.4.0 (2015-09-04)
 ================================
 
 This is the last Twisted release where Python 2.6 is supported, on any
-platform. 
+platform.
 
 Features
 --------
@@ -4417,6 +4417,7 @@ Twisted Core 15.0.0 (2015-01-24)
 
 Features
 --------
+ - twisted.internet.endpoints adding support for cipher in `serverFromString`(#12134)
  - twisted.internet.protocol.ClientFactory (and subclasses) may now
    return None from buildProtocol to immediately close the connection.
    (#710)
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3873,7 +3873,7 @@ Twisted Core 15.4.0 (2015-09-04)
 ================================
 
 This is the last Twisted release where Python 2.6 is supported, on any
-platform.
+platform. 
 
 Features
 --------
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -4417,7 +4417,6 @@ Twisted Core 15.0.0 (2015-01-24)
 
 Features
 --------
- - twisted.internet.endpoints adding support for cipher in `serverFromString`(#12134)
  - twisted.internet.protocol.ClientFactory (and subclasses) may now
    return None from buildProtocol to immediately close the connection.
    (#710)
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1472,8 +1472,8 @@ def _parseSSL(
     )
 
     if cipher:
-        cipher_bytes = cipher.replace(",", ":").encode("ascii")
-        cf.getContext().set_cipher_list(cipher_bytes)
+        cipherBytes = cipher.replace(",", ":").encode("ascii")
+        cf.getContext().set_cipher_list(cipherBytes)
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
 
 
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -22,6 +22,8 @@ from unicodedata import normalize
 
 from zope.interface import directlyProvides, implementer, provider
 
+import OpenSSL
+
 from constantly import NamedConstant, Names
 from incremental import Version
 
@@ -1470,10 +1472,13 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
-
     if cipher:
-        cipherBytes = cipher.replace(",", ":").encode("ascii")
-        cf.getContext().set_cipher_list(cipherBytes)
+        try:
+            cipherBytes = cipher.replace(",", ":").encode("ascii")
+            cf.getContext().set_cipher_list(cipherBytes)
+        except OpenSSL.SSL.Error:
+            raise Exception("Invalid cipher list passed")
+
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
 
 
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1415,6 +1415,9 @@ def _parseSSL(
         constant in C{OpenSSL.SSL}.
     @type sslmethod: C{str}
 
+    @param cipher: A string containing a list of cipher values, formatted as comma-separated values.
+    @type cipher: C{str}
+
     @param extraCertChain: The path of a file containing one or more
         certificates in PEM format that establish the chain from a root CA to
         the CA that signed your C{certKey}.
@@ -1469,8 +1472,8 @@ def _parseSSL(
     )
 
     if cipher:
-        if isinstance(cipher, str):
-            cf.getContext().set_cipher_list(str.encode(cipher.replace(",", ":")))
+        cipher_bytes = cipher.replace(",", ":").encode("ascii")
+        cf.getContext().set_cipher_list(cipher_bytes)
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
 
 
@@ -1772,6 +1775,13 @@ def serverFromString(reactor, description):
     file is assumed to exist which contains the private key. If the certificate
     file name (C{certKey}) isn't provided, the private key file is assumed to
     contain the certificate as well.
+
+    For custom cipher list, you can make use of the 'cipher' keyword and a comma seperated
+    string as below::
+
+        serverFromString(
+            reactor, "ssl:443:privateKey=key.pem:certKey=crt.pem:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256"
+        )
 
     You may escape colons in arguments with a backslash, which you will need to
     use if you want to specify a full pathname argument on Windows::
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1786,7 +1786,7 @@ def serverFromString(reactor, description):
     string as below::
 
         serverFromString(
-            reactor, "ssl:443:privateKey=key.pem:certKey=crt.pem:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256"
+            reactor, "ssl:443:privateKey=key.pem:certKey=crt.pem:cipher=ALL,!ADH,@STRENGTH,+RSA,-DSA,SHA1+DES"
         )
 
     You may escape colons in arguments with a backslash, which you will need to
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1382,6 +1382,7 @@ def _parseSSL(
     privateKey="server.pem",
     certKey=None,
     sslmethod=None,
+    cipher=None,
     interface="",
     backlog=50,
     extraCertChain=None,
@@ -1466,6 +1467,10 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
+
+    if cipher:
+        if isinstance(cipher, str):
+            cf.getContext().set_cipher_list(str.encode(cipher.replace(",", ":")))
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
 
 
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -22,8 +22,6 @@ from unicodedata import normalize
 
 from zope.interface import directlyProvides, implementer, provider
 
-import OpenSSL
-
 from constantly import NamedConstant, Names
 from incremental import Version
 
@@ -1479,7 +1477,7 @@ def _parseSSL(
         try:
             cipherBytes = cipher.replace(",", ":").encode("ascii")
             cf.getContext().set_cipher_list(cipherBytes)
-        except OpenSSL.SSL.Error:
+        except SSLError:
             raise Exception("Invalid cipher list passed")
 
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1474,11 +1474,11 @@ def _parseSSL(
         **kw,
     )
     if cipher:
+        cipherBytes = cipher.replace(",", ":").encode("ascii")
         try:
-            cipherBytes = cipher.replace(",", ":").encode("ascii")
             cf.getContext().set_cipher_list(cipherBytes)
         except SSLError:
-            raise Exception("Invalid cipher list passed")
+            pass
 
     return ((int(port), factory, cf), {"interface": interface, "backlog": int(backlog)})
 
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1417,7 +1417,10 @@ def _parseSSL(
         constant in C{OpenSSL.SSL}.
     @type sslmethod: C{str}
 
-    @param cipher: A string containing a list of cipher values, formatted as comma-separated values.
+    @param cipher: A string containing a list of cipher values, formatted as comma-separated values,
+                    These values can be obtained from the OpenSSL documentation at
+                    https://www.openssl.org/docs/man1.0.2/man1/ciphers.html.
+                    This parameter is optional and can be used to restrict the ciphers supported by your server.
     @type cipher: C{str}
 
     @param extraCertChain: The path of a file containing one or more
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1415,7 +1415,7 @@ def _parseSSL(
         constant in C{OpenSSL.SSL}.
     @type sslmethod: C{str}
 
-    @param cipher: A string containing a list of cipher values, formatted as comma-separated values, 
+    @param cipher: A string containing a list of cipher values, formatted as comma-separated values,
                     These values can be obtained from the OpenSSL documentation at
                     https://www.openssl.org/docs/man1.0.2/man1/ciphers.html.
                     This parameter is optional and can be used to restrict the ciphers supported by your server.
--- a/src/twisted/internet/endpoints.py
+++ b/src/twisted/internet/endpoints.py
@@ -1415,7 +1415,10 @@ def _parseSSL(
         constant in C{OpenSSL.SSL}.
     @type sslmethod: C{str}
 
-    @param cipher: A string containing a list of cipher values, formatted as comma-separated values.
+    @param cipher: A string containing a list of cipher values, formatted as comma-separated values, 
+                    These values can be obtained from the OpenSSL documentation at
+                    https://www.openssl.org/docs/man1.0.2/man1/ciphers.html.
+                    This parameter is optional and can be used to restrict the ciphers supported by your server.
     @type cipher: C{str}
 
     @param extraCertChain: The path of a file containing one or more
--- a/src/twisted/internet/test/test_endpoints.py
+++ b/src/twisted/internet/test/test_endpoints.py
@@ -3266,13 +3266,11 @@ class ServerStringTests(unittest.TestCase):
         The colon (:) from the OpenSSL format is replaced with a comma (,).
         """
         reactor = object()
-        server = (
-            endpoints.serverFromString(
-                reactor,
-                "ssl:1234:backlog=12:privateKey=%s:"
-                "certKey=%s:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256,ALL,!ADH,"
-                % (escapedPEMPathName, escapedPEMPathName),
-            ),
+        server = endpoints.serverFromString(
+            reactor,
+            "ssl:1234:backlog=12:privateKey=%s:"
+            "certKey=%s:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256,ALL,!ADH,"
+            % (escapedPEMPathName, escapedPEMPathName),
         )
         self.assertIsInstance(server, endpoints.SSL4ServerEndpoint)
         self.assertEqual(server._sslContextFactory.method, TLSv1_2_METHOD)
--- a/src/twisted/internet/test/test_endpoints.py
+++ b/src/twisted/internet/test/test_endpoints.py
@@ -3285,14 +3285,15 @@ class ServerStringTests(unittest.TestCase):
         The colon (:) from the OpenSSL format is replaced with a comma (,).
         """
         reactor = object()
-        with self.assertRaises(Exception) as context:
+
+        self.assertRaises(
             endpoints.serverFromString(
                 reactor,
                 "ssl:1234:privateKey=%s:"
                 "certKey=%s:cipher=bla, blah"
                 % (escapedPEMPathName, escapedPEMPathName),
             )
-            self.assertTrue("Invalid cipher list passed" in str(context.exception))
+        )
 
     def test_unix(self):
         """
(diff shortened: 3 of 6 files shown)


KaviHarjani:
please review

NEWS.rst
```diff
@@ -4417,6 +4417,7 @@ Twisted Core 15.0.0 (2015-01-24)
 
 Features
 --------
+ - twisted.internet.endpoints adding support for cipher in `serverFromString`(#12134)
```
adiroiban: This needs to be moved into a "fragment" file.

We don't do direct edit of the release notes.

See more info here https://docs.twisted.org/en/stable/development/dev-process.html#release-notes-management

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1467,10 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
+
+    if cipher:
+        if isinstance(cipher, str):
```
adiroiban: I guess that it's ok if we don't do type checking for str.


```suggestion
    if cipher is not None:
```

What should happen if you pass a boolean or other types ?

I don't think that we need to worry.

What we need to worry, is to document the `cipher` arument and let people know that it should be `str`

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1467,10 @@ def _parseSSL(
...
+            cf.getContext().set_cipher_list(str.encode(cipher.replace(",", ":")))
```
adiroiban: I think is best to have explicit `ascii` encoding here.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,29 @@ def test_sslNoTrailingNewlinePem(self):
         ctx = server._sslContextFactory.getContext()
         self.assertIsInstance(ctx, ContextType)
 
+    @skipIf(skipSSL, skipSSLReason)
+    def test_sslWithCustomCipher(self):
+        """
+        Test adding custom cipher to the serverFromString function.
```
adiroiban: Maybe something like this.

See https://jml.io/test-docstrings/

```suggestion
        A cipher list is supported, using the OpenSSL format.
        The colon (:) from the OpenSSL format is replaced with a comman (').
```

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,29 @@ def test_sslNoTrailingNewlinePem(self):
...
+                "certKey=%s:sslmethod=TLSv1_2_METHOD:interface=10.0.0.1:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-ECDSA-CHACHA20-POLY1305,ECDHE-RSA-AES128-GCM-SHA256"
+                % (escapedPEMPathName, escapedPEMPathName),
+            ),
+        )
+        self.assertIsInstance(server, endpoints.SSL4ServerEndpoint)
+        self.assertIs(server._reactor, reactor)
+        self.assertEqual(server._port, 1234)
+        self.assertEqual(server._backlog, 12)
+        self.assertEqual(server._interface, "10.0.0.1")
+        self.assertEqual(server._sslContextFactory.method, TLSv1_2_METHOD)
```
adiroiban: ar these assertions needed for this test?

I guess that here we should assert that the ciphers were set...

I don't know how to get the configured cipher list for an SSL context :(

-------

For my code, to test it, I do an actual handshake...

but I feel that that it's too much here.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,29 @@ def test_sslNoTrailingNewlinePem(self):
...
+    def test_sslWithCustomCipher(self):
+        """
+        Test adding custom cipher to the serverFromString function.
+        """
+        reactor = object()
+        server = (
+            endpoints.serverFromString(
+                reactor,
+                "ssl:1234:backlog=12:privateKey=%s:"
+                "certKey=%s:sslmethod=TLSv1_2_METHOD:interface=10.0.0.1:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-ECDSA-CHACHA20-POLY1305,ECDHE-RSA-AES128-GCM-SHA256"
```
adiroiban: WOuld it make sense to keep the input for this test with a narrow scope?

Do we need backlog here ?
Do we need to care about sslmethod here ? 

I expect that we have separate tests for backlog or ssl method

```suggestion
                "certKey=%s:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256"
```

adiroiban (changes requested):
Many thanks. The changes look good and are a good start.

Things blocking the merge:

* dedicated release notes fragment
* handling of non-string arguments
* lack of documentation

How are people expected to know about this feature ?

I was expected to see it documented in the docstring of serverFromString  

similar to what we have for other rules https://docs.twistedmatrix.com/en/stable/api/twisted.internet.endpoints.html#serverFromString


there is also this documentation page https://docs.twisted.org/en/stable/core/howto/endpoints.html#endpoint-types-included-with-twisted

and I think that it would help to document this feature here.

----------


It would be nice to have a followup for clientFromString ... but this needs to be done in a separate ticket/PR

We should focus on the serverFromString for this PR.

Thanks again

src/twisted/internet/test/test_endpoints.py
adiroiban: Maybe it would help to have a cipher text like

```
ALL:!ADH:@STRENGTH:+RSA:-DSA:SHA1+DES
```

I hope that this provide a good cover of the value that can be found in a cipher definition

KaviHarjani:
please review

NEWS.rst
```diff
@@ -3873,7 +3873,7 @@ Twisted Core 15.4.0 (2015-09-04)
 ================================
 
 This is the last Twisted release where Python 2.6 is supported, on any
-platform. 
+platform.
```
glyph: This file should not be changed manually, please remove this change.

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1470,10 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
+
+    if cipher:
+        cipher_bytes = cipher.replace(",", ":").encode("ascii")
+        cf.getContext().set_cipher_list(cipher_bytes)
```
glyph: Twisted's coding standard is slightly different than PEP8; you should be using `camelCase` rather than `snake_case` for local variables.

```suggestion
        cipherBytes = cipher.replace(",", ":").encode("ascii")
        cf.getContext().set_cipher_list(cipherBytes)
```

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,27 @@ def test_sslNoTrailingNewlinePem(self):
...
+                "ssl:1234:backlog=12:privateKey=%s:"
+                "certKey=%s:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256,ALL,!ADH,"
+                "@STRENGTH,+RSA,-DSA,SHA1+DES"
+                % (escapedPEMPathName, escapedPEMPathName),
+            ),
+        )
+        self.assertIsInstance(server, endpoints.SSL4ServerEndpoint)
+        self.assertEqual(server._sslContextFactory.method, TLSv1_2_METHOD)
+        ctx = server._sslContextFactory.getContext()
+        self.assertIsInstance(ctx, ContextType)
```
glyph, line 3281: This coverage warning indicates that this test always fails.

glyph (dismissed):
Please get the tests passing and patch coverage to 100% before requesting another review so as to make efficient use of reviewer time; we can't integrate changes that don't pass our basic automated quality checks.

src/twisted/internet/test/test_endpoints.py
KaviHarjani: @glyph thankyou, I had to reinstall and run the test for this to show, it was actually passing all the tests locally and I was unable to figure

NEWS.rst
KaviHarjani: This was autoformatted

KaviHarjani:
please review

src/twisted/internet/endpoints.py
```diff
@@ -1414,6 +1415,9 @@ def _parseSSL(
         constant in C{OpenSSL.SSL}.
     @type sslmethod: C{str}
 
+    @param cipher: A string containing a list of cipher values, formatted as comma-separated values.
```
glyph: What is a "cipher value"? Presumably there is a list of these magic strings somewhere. Is it OpenSSL specific? How should the user discover the possible list of options? Why might they want to use this option?

glyph:
(Thanks a ton for getting those tests passing!)

glyph:
please review

glyph:
oops I did not mean to dismiss the review request with my one stray comment

src/twisted/internet/endpoints.py
KaviHarjani: @glyph yes, so if you do openssl ciphers -v 'ALL:eNULL', you can find all the cipher values 

The usecase is, I work as a developer in a bank and they want to limit cipher use, even in so basic requirement is TLS1.2 and even in there, they only want to enable a few ciphers

They are already using Daphne and Daphne is using `serverFromString` from Twisted 

They are averse of making big changes like using Twisted directly, so with this I can have them use only selected cipher for them

I was going to make this changes directly in Daphne but this could be used by a lot of people 
So doing it here made more sense to me

KaviHarjani: Should i mention from where to get the -cipher value from? 
I think people wanting to use this feature would actually know what they are looking for and will specifically try finding it, like I id haha

Also thanks for `Twisted` you guys <3

glyph: > The usecase is, I work as a developer in a bank and they want to limit cipher use, even in so basic requirement is TLS1.2 and even in there, they only want to enable a few ciphers

Sorry, I should have been clearer: I am asking these questions not because I don't know the answer (I do) but from the perspective of some random future developer reading about this in Twisted.  Twisted right now uses OpenSSL for TLS, but hopefully we will not always do that (I have a pretty short-term to-do item to look at pyrtls to see if we can get a more memory-safe option).

So, imagine that you are reading this as someone who knows what TLS is, and wants to do something with it, but doesn't know anything in particular about OpenSSL or how to use it.

glyph: So the questions are all things that need to be somehow answered within the text of the source code, not just in a comment reply :)

KaviHarjani: @glyph added it, let me know if its good enough or do I need to add more :)

src/twisted/newsfragments/12134.feature
```diff
@@ -0,0 +1 @@
+Added support for `cipher` in `serverFromString` function
```
adiroiban: The current test read more like a commit mesage.

The newsframgment, is for the public release notes.

It's important to clearly explain which API is targeted by this change, and what is this change all about.

The project convention for the release notes, it to start the message with the canonical name of the API that was changed

```suggestion
twisted.internet.endpoint.serverFromString now supports defining a cipher list. 
```

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,23 @@ def test_sslNoTrailingNewlinePem(self):
...
+    def test_sslWithCustomCipher(self):
+        """
+        A cipher list is supported, using the OpenSSL format.
+        The colon (:) from the OpenSSL format is replaced with a comma (,).
+        """
+        reactor = object()
+        server = endpoints.serverFromString(
+            reactor,
+            "ssl:1234:backlog=12:privateKey=%s:"
+            "certKey=%s:cipher=ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-RSA-AES128-GCM-SHA256,ALL,!ADH"
```
adiroiban: I see my previous comment was marked as "resolved" , but I don't see any info about why no change is required.

Do we need `backlog` in this test? Will the test fail if backlog is not provided ?

Is there any value in having both `ECDHE-ECDSA-AES128-GCM-SHA256` and `ECDHE-RSA-AES128-GCM-SHA256` suites in this list ?

I still think that we should look into a minimal list, that covers all the available OpenSSL options ... something like `ALL:!ADH:@STRENGTH:+RSA:-DSA:SHA1+DES`

I can see that this is valid

```
$ openssl ciphers 'ALL:!ADH:@STRENGTH:+RSA:-DSA:SHA1+DES'
```

-----------

I am note sure that this test provides good coverage.

It's important to have

For example, if I remove these lines from the branch, would this test fail ?

```
    if cipher:
        cipherBytes = cipher.replace(",", ":").encode("ascii")
        cf.getContext().set_cipher_list(cipherBytes)
```

--------

I still think that this test is valid.
I also think that the only way to check which ciphers were set, is to do a full handshake ... and this might be too much for this test.

But it's important to add a note about this limitation in the docstring of the test.

---------

What happens if I pass a string like `ssl:1234:cipher=bla,bla-bla`

I think that we need a test to check how the errors are reported.

I guess that the error here would be "SSL routines - no cipher match" ... but is best to have a test in which we describe that we don't have any custom error message and the OpenSSL low-level exception is raised.

I don't know if we want that.

I guess that it would be nice to handle the OpenSSL exception and show a better error message

adiroiban (changes requested):
Many thanks for the updates.

Happy to see the release notes and the test passing.

I can see that my comments were marked as resolved, but I don't see any reply to my comments. Not a big deal... but it's a bit confusing, especially since some change requests were not made.

I left another comment.

------------

It's ok to have a test for the "happy path", but I think that we should also have a test for error handling.

The test for error handling should also provide a better coverage to make sure the low-level OpenSSL API is called.

Please consider adding a test for the case in which an invalid cipher list  value is used.

Thanks again.

src/twisted/newsfragments/12134.feature
KaviHarjani: [#f991e5a](https://github.com/twisted/twisted/pull/12137/commits/f991e5a4aa8b557cb19c222f876940b12cc0c0b6)

src/twisted/internet/test/test_endpoints.py
KaviHarjani: [#fe8f46a](https://github.com/twisted/twisted/pull/12137/commits/fe8f46a3f91f6efc12a73f7c6fab3fb494e17c89), [#53b0d41](https://github.com/twisted/twisted/pull/12137/commits/53b0d41b6a8aa4b90edc2365739031c951e0db13)

KaviHarjani:
please review

NEWS.rst
KaviHarjani: [#6e30f](https://github.com/twisted/twisted/pull/12137/commits/6e30f9f83d118633e7be1c489d00073549c1b285)

src/twisted/internet/endpoints.py
KaviHarjani: [#692e540f](https://github.com/twisted/twisted/pull/12137/commits/692e540f85cac1146fb0a44a230138ca743ce5b0)

src/twisted/internet/endpoints.py
KaviHarjani: [#692e540f](https://github.com/twisted/twisted/pull/12137/commits/692e540f85cac1146fb0a44a230138ca743ce5b0)

KaviHarjani:
@adiroiban appreciate your reviews and apologise, I think I made most of the changes you requested except for the test one, not sure as you said custom handshake would be too much,  and that is why I marked those as resolved, I added comments with reference to the change

I have made changes and added an Exception, giving a better explanation of what is going on when cipher is wrong

Please let me know if it is good to go

Also, one more request @glyph @adiroiban if any one of you guys could help me with the patch, not sure why it says the code isn't covered and points to the uncovered test code

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,41 @@ def test_sslNoTrailingNewlinePem(self):
...
+        """
+        reactor = object()
+        with self.assertRaises(Exception) as context:
+            endpoints.serverFromString(
+                reactor,
+                "ssl:1234:privateKey=%s:"
+                "certKey=%s:cipher=bla, blah"
+                % (escapedPEMPathName, escapedPEMPathName),
+            )
+            self.assertTrue("Invalid cipher list passed" in str(context.exception))
```
adiroiban: This line is never called as the previous lines raises an exception.

This is why is best to avoid using the context manager

```python
exception = self.assertRaises(
    SomeException,
    endpoints.serverFromString,
    reactor=reactor,
    ...
)
```

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1473,13 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
+    if cipher:
+        try:
+            cipherBytes = cipher.replace(",", ":").encode("ascii")
```
adiroiban: can this line throw an SSLError ?

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1473,13 @@ def _parseSSL(
...
+            cf.getContext().set_cipher_list(cipherBytes)
+        except SSLError:
+            raise Exception("Invalid cipher list passed")
```
adiroiban: I don't think that raising a generic exception is a good idea.

And the SSLError details are lost.

-------

I have checked the other parts of twisted, and I see that they just let the low-level SSLError bubble to the caller.

No need for special handling of the error.

Just make sure that we have a test for the error case.

-------

I have checked the Twisted coding standard, but I could not find any guidence for how is best to handle the exceptions

https://docs.twisted.org/en/stable/development/coding-standard.html


--------

But I think that converting an SSLError into a generic Exception is not a good idea.

src/twisted/internet/test/test_endpoints.py
adiroiban: It's best to folow TDD

Write the test first, run the test and see if fail... then fix the code.

If you write the test first and it passes from the first run, it means that the test is not valid.

https://docs.twisted.org/en/stable/development/test-standard.html

src/twisted/internet/endpoints.py
```diff
@@ -1466,6 +1473,13 @@ def _parseSSL(
         dhParameters=dhParameters,
         **kw,
     )
+    if cipher:
+        try:
+            cipherBytes = cipher.replace(",", ":").encode("ascii")
+            cf.getContext().set_cipher_list(cipherBytes)
```
adiroiban, line 1479: I guess that this will work.

By understanding is that context factories might not cache the context.

This is why, we might need to change the API to something like this ... but this will break all existing context factory implementation.

So I guess that this is somehow ok for now.

We can refactor once someone nees to used this code with a custom context factory.

 
```
            cf.setOpenSSLCipherList(cipherStr)
```

If we delegate the cipher list setting to the context factory, then we no longer need the TLS handshake tests for endpoints... we will just rely on the context factory tests.

src/twisted/internet/test/test_endpoints.py
```diff
@@ -3259,6 +3259,41 @@ def test_sslNoTrailingNewlinePem(self):
         ctx = server._sslContextFactory.getContext()
         self.assertIsInstance(ctx, ContextType)
 
+    @skipIf(skipSSL, skipSSLReason)
+    def test_sslWithCustomCipher(self):
+        """
+        A cipher list is supported, using the OpenSSL format.
+        The colon (:) from the OpenSSL format is replaced with a comma (,).
+        Limitations to the test -> Not checked the ciphers returned.
```
adiroiban: I think that is best to add the full explanation/excuse of why we have a mediocre test here.

We have the discussion over the PR, but the information nees to be in the code.

```suggestion
        A cipher list is supported, using the OpenSSL format.
        The colon (:) from the OpenSSL format is replaced with a comma (,).
        
        This test does not provides much functionality coverage as there is no API to retrieve the current configured ciper list.
        Testing this functionality requires doing a TLS handshake.
```

adiroiban (changes requested):
Thanks for the changes.

I don't think that the new test is correct... the assertion is never called.

Also, I am not sure that converting SSLError to a generic Exception is a good idea.
I think that at this point is best to just raise the low-level SSLError.... as we don't have any other better Twisted specific exception.

I am still trying to find a good reason for not writing a proper test for the happy path.

I think that we should use iosim here and do a TLS handshake, to check that the code works as expections.

We use a ContextFactory here...but the cipher is set on a specific context and not an the context factory itself.

Later in the code execution, the ContextFactory might return a different context instance, and then the cipher list is not actually used in the TLS handshake.

src/twisted/internet/test/test_endpoints.py
KaviHarjani: [#e9308d](https://github.com/twisted/twisted/pull/12137/commits/6e30f9f83d118633e7be1c489d00073549c1b285)

src/twisted/internet/endpoints.py
KaviHarjani: [#e9308d](https://github.com/twisted/twisted/pull/12137/commits/6e30f9f83d118633e7be1c489d00073549c1b285)

src/twisted/internet/endpoints.py
KaviHarjani: [#e9308d](https://github.com/twisted/twisted/pull/12137/commits/6e30f9f83d118633e7be1c489d00073549c1b285)

src/twisted/internet/endpoints.py
KaviHarjani: @adiroiban what do you suggest?
Should we go ahead with making changes to the API ?

KaviHarjani:
> Thanks for the changes.
> 
> I don't think that the new test is correct... the assertion is never called.
> 
> Also, I am not sure that converting SSLError to a generic Exception is a good idea. I think that at this point is best to just raise the low-level SSLError.... as we don't have any other better Twisted specific exception.
> 
> I am still trying to find a good reason for not writing a proper test for the happy path.
> 
> I think that we should use iosim here and do a TLS handshake, to check that the code works as expections.
> 
> We use a ContextFactory here...but the cipher is set on a specific context and not an the context factory itself.
> 
> Later in the code execution, the ContextFactory might return a different context instance, and then the cipher list is not actually used in the TLS handshake.

Hi @adiroiban  I am having problems writing the test, I see there is a custom handshake done for `clientFromString` but I don't see any of those for `serverFromString`, so no reference point 
I tried the implementation with WrappingFactoryTests but just hitting roadblocks there, could you give hints


```
    @skipIf(skipSSL, skipSSLReason)
    def test_sslWithCustomCipher(self):
        """
        A cipher list is supported, using the OpenSSL format.
        The colon (:) from the OpenSSL format is replaced with a comma (,).
        Limitations to the test -> Not checked the ciphers returned.
        """
        server = endpoints.serverFromString(
            reactor,
            "ssl:1234:privateKey=%s:"
            "certKey=%s:cipher=ALL,!ADH,@STRENGTH,+RSA,-DSA,SHA1+DES"
            % (escapedPEMPathName, escapedPEMPathName),
        )

        factory = object()
        d = server.listen(factory)

        receivedHosts = []

        def checkPortAndServer(port):
            receivedHosts.append(port.getHost())

        d.addCallback(checkPortAndServer)

        self.assertIsInstance(server, endpoints.SSL4ServerEndpoint)
        ctx = server._sslContextFactory.getContext()
        self.assertIsInstance(ctx, ContextType)

```

```
class WrappingFactoryTests(unittest.TestCase):
    """
    Test the behaviour of our ugly implementation detail C{_WrappingFactory}.
    """

    def test_doStart(self):
        """
        L{_WrappingFactory.doStart} passes through to the wrapped factory's
        C{doStart} method, allowing application-specific setup and logging.
        """
        factory = ClientFactory()
        wf = endpoints._WrappingFactory(factory)
        wf.doStart()
        self.assertEqual(1, factory.numPorts)
```



I tried using ServerFactory instead of ClientFactory but that did not work either 
Could you please help me out?

Also with this change since it will not be cached to context, is there possibilities of the handshake to happen outside of the given cipher list

adiroiban:
You can start by writing an end to end test using the exact same steps and API as you would implement a TLS client->server connection outside of the testing code

-----------

Just listening for a port will not trigger any TLS handshake.

I think that you need to pass a real factory that uses a protocol that implements `IHandshakeListener`

Then make a client connection to trigger the handshake and check the negotiated protocol... or check the protocol negotiation failure.

The test can run over localhost with a randomly allocated port.

PR #12142, opened 2024-04-24
9713 conch types

### LINKED ISSUE #9713: IKnownHostEntry and PlainHost/HashedHost string type confusion
|[twm's avatar](https://github.com/twm)| @twm reported|
|-|-|
|Trac ID|trac#9713|
|Type|enhancement|
|Created|2019-10-14 04:31:25Z|

`twisted.conch.interfaces.IKnownHostEntry.matchesHost()`'s `hostname` argument is documented as type `str`, but the implementations disagree:

`twisted.conch.client.knownhosts.PlainHost` and `twisted.conch.client.knownhosts.HashedHost` both document the parameter as type `bytes`. The implementations accept both `str` _and_ bytes (`str` is encoded to `utf-8`j).

What string type does the actual client implementation pass? That is, if someone implements `IKnownHostEntry` according to the interface, will it work with the Conch client? The interface should document the type that promotes this type of compatibility. (If the interface is wrong and the type should be `bytes`, that's not a backwards-incompatible change, it's just a bug in the docs that probably occurred when porting to Python 3.)

```
trac-id__9713 9713
type__enhancement enhancement
reporter__twm twm
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1571027485492856 1571027485492856
changetime__1571027485492856 1571027485492856
version__None None
owner__None None

```
</details>

### LINKED ISSUE #12141: type confusion in KnownHosts.verifyHostKey traceback when adding ip key
If you have a known hosts entry for a host, but not its IP, `twisted.conch.client.knownhosts.KnownHostsFile.verifyHostKey` emits a traceback in `ConsoleUI.warn`.

The type confusion here is pervasive so this has a few related issues, #9713 among them

glyph:
## Scope and purpose

Fixes #12141 
Fixes #9713 

This fixes an interactive traceback with Conch, but also adds a bunch of type annotations and testing problems to ensure that other related issues are easier to fix.

### DIFF
--- a/src/twisted/conch/client/knownhosts.py
+++ b/src/twisted/conch/client/knownhosts.py
@@ -461,7 +461,7 @@ class KnownHostsFile:
         """
         hhk = defer.execute(self.hasHostKey, hostname, key)
 
-        def gotHasKey(result):
+        def gotHasKey(result: bool) -> bool | Deferred[bool]:
             if result:
                 if not self.hasHostKey(ip, key):
                     addMessage = (
--- a/src/twisted/conch/client/knownhosts.py
+++ b/src/twisted/conch/client/knownhosts.py
@@ -8,12 +8,14 @@ An implementation of the OpenSSH known_hosts database.
 @since: 8.2
 """
 
+from __future__ import annotations
 
 import hmac
 import sys
 from binascii import Error as DecodeError, a2b_base64, b2a_base64
 from contextlib import closing
 from hashlib import sha1
+from typing import IO, Callable, Literal
 
 from zope.interface import implementer
 
@@ -21,8 +23,10 @@ from twisted.conch.error import HostKeyChanged, InvalidEntry, UserRejectedKey
 from twisted.conch.interfaces import IKnownHostEntry
 from twisted.conch.ssh.keys import BadKeyError, FingerprintFormats, Key
 from twisted.internet import defer
+from twisted.internet.defer import Deferred
 from twisted.logger import Logger
 from twisted.python.compat import nativeString
+from twisted.python.filepath import FilePath
 from twisted.python.randbytes import secureRandom
 from twisted.python.util import FancyEqMixin
 
@@ -111,34 +115,33 @@ class PlainEntry(_BaseEntry):
     file.
 
     @ivar _hostnames: the list of all host-names associated with this entry.
-    @type _hostnames: L{list} of L{bytes}
     """
 
-    def __init__(self, hostnames, keyType, publicKey, comment):
-        self._hostnames = hostnames
+    def __init__(
+        self, hostnames: list[bytes], keyType: bytes, publicKey: Key, comment: bytes
+    ):
+        self._hostnames: list[bytes] = hostnames
         super().__init__(keyType, publicKey, comment)
 
     @classmethod
-    def fromString(cls, string):
+    def fromString(cls, string: bytes) -> PlainEntry:
         """
         Parse a plain-text entry in a known_hosts file, and return a
         corresponding L{PlainEntry}.
 
         @param string: a space-separated string formatted like "hostname
-        key-type base64-key-data comment".
-
-        @type string: L{bytes}
+            key-type base64-key-data comment".
 
         @raise DecodeError: if the key is not valid encoded as valid base64.
 
         @raise InvalidEntry: if the entry does not have the right number of
-        elements and is therefore invalid.
+            elements and is therefore invalid.
 
         @raise BadKeyError: if the key, once decoded from base64, is not
-        actually an SSH key.
+            actually an SSH key.
 
         @return: an IKnownHostEntry representing the hostname and key in the
-        input line.
+            input line.
 
         @rtype: L{PlainEntry}
         """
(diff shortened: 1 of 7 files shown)


adiroiban:
Thanks Glyph for the PR. I will be traveling in the next day... but I will try to get this review.
If somehow I forgot to review this by the start of next week, feel free to ping me and I will take a look.

Thanks again.

glyph:
Thanks Adi! I just had a fun time accidentally causing my machine to kernel panic with a test bug :) but this should be reviewable now...

glyph:
please review

src/twisted/conch/client/knownhosts.py
```diff
@@ -111,65 +115,61 @@ class PlainEntry(_BaseEntry):
...
 
         @raise BadKeyError: if the key, once decoded from base64, is not
-        actually an SSH key.
+            actually an SSH key.
 
         @return: an IKnownHostEntry representing the hostname and key in the
-        input line.
+            input line.
 
         @rtype: L{PlainEntry}
```
adiroiban, line 146: Minor comment. Maybe this docstring can be removed as pydoctor should be able to extract this from the method signature.

I see the `rtype` was removed in other methods.

src/twisted/conch/client/knownhosts.py
```diff
@@ -520,19 +529,15 @@ def addHostKey(self, hostname, key):
...
         p = self._savePath.parent()
         if not p.isdir():
             p.makedirs()
 
-        if self._clobber:
-            mode = "wb"
-        else:
-            mode = "ab"
-
+        mode: Literal["a", "w"] = "w" if self._clobber else "a"
```
adiroiban, line 540: Hm... Not sure how the test passed for this code.

I see the file is opend in text mode, but then we write bytes.

----------

```suggestion
        mode: Literal["ab", "wb"] = "wb" if self._clobber else "an"
```

src/twisted/conch/client/knownhosts.py
```diff
@@ -598,20 +601,19 @@ def body(ignored):
                     answer = f.readline().strip().lower()
                     if answer == b"yes":
                         return True
-                    elif answer == b"no":
+                    elif answer in {b"no", b""}:
```
adiroiban, line 604: Maybe update the docstring

```diff
-'yes' and L{False} when the user answers 'no'. It may errback if
-there were any I/O errors.
+'yes' and L{False} when the user answers 'no' or no answer.
+It may errback if there were any I/O errors.

File: src/twisted/conch/endpoints.py [hunk 4]
```diff
@@ -773,21 +743,21 @@ def __init__(
...
+    def _opener(self) -> IO[bytes]:
         """
         Open the tty if possible, otherwise give back a file-like object from
         which C{b"no"} can be read.
 
         For use as the opener argument to L{ConsoleUI}.
         """
         try:
-            return self.tty.open("rb+")
+            return self.tty.open("r+")
```
adiroiban, line 756: Will this open in bytes mode ?

src/twisted/conch/test/test_endpoints.py
```diff
@@ -1435,31 +1435,37 @@ def test_cleanupConnectionImmediately(self):
         helper.cleanupConnection(object(), True)
 
 
+class _WriteDiscarder(BytesIO):
+    def write(self, data: object, /) -> int:
+        """
+        Discard writes because we are emulating a console object, where they'd
+        go to the screen, not back into the buffer to be read like an a+ file.
+        """
+        return 0
```
adiroiban, line 1444: can we assert that we write str or bytes  ?

I thins that ByteIO will reject `str` for a write operation, but it looks _WriteDiscarder will accept anything.


If this is on purpose, I think that this needs to mentioned in the docstring of `_WriteDiscarder`

src/twisted/conch/test/test_endpoints.py
```diff
@@ -1435,31 +1435,37 @@ def test_cleanupConnectionImmediately(self):
...
+        If the mode is r+, return a file with the given contents as a line.
 
-        @return: A L{_ReadFile} instance
+        @return: a buffer of the given contents, which will discard any writes
+            given to it.
         """
-        if mode == "rb+":
-            return _ReadFile(self.contents)
-        raise OSError(ENOSYS, "Function not implemented")
+        assert mode == "r+"
```
adiroiban, line 1467: should't we be explicit about binary mode?

adiroiban (changes requested):
Thanks for the changes.

Maybe I got the changes wrong.

I am not sure the changes from `bytes` IO to `str` IO are ok.

There is no misisng code coverage ... and I was expecting the test to fail.

Please see the inline comments related to `mode="wb"` vs `mode="b"` 

Other than that, this looks good.

Thanks again.

src/twisted/conch/client/knownhosts.py
glyph: `FilePath` always opens files in binary mode; it doesn't have a text mode, so the `b` is superfluous. See the annotation here: https://docs.twistedmatrix.com/en/stable/api/twisted.python.filepath.FilePath.html#open and note that it unconditionally returns `IO[bytes]`.

src/twisted/conch/endpoints.py
glyph: Yep, see above about `FilePath`.

src/twisted/conch/test/test_endpoints.py
glyph: Nope, once again, `FilePath` is always binary mode.

src/twisted/conch/client/knownhosts.py
glyph: OK, I will get rid of it.

adiroiban (approved):
All good. Thanks for the explanations. Much appreciated.

glyph:
> All good. Thanks for the explanations. Much appreciated.

Quick approval appreciated as well!

PR #12144, opened 2024-04-30
#11893 Move iocp dependency to Windows specific soft dependencies.

### LINKED ISSUE #11893: twisted-iocpsupport should be part of the `windows-platform` extra, not a hard requirement
Periodically, people report issues like #11891 when trying to use Twisted on Windows.  There are two problems here.

The first is that our CI pipeline for `twisted-iocpsupport` should be more proactive and produce extension modules for all configurations that a Windows user might encounter.

However, we are never going to be able to be *completely* comprehensive there, and thus we should make it so that `pip install twisted` can function as a pure-python package on Windows as it does on other platforms. So while we should make `pip install twisted[all]` work well by doing the first thing, we should also do this.

adiroiban:
## Scope and purpose

Fixes #11893

This is done so that `pip install twisted` will no longer depend on our IOCP support.

The IOCP support package might get outdate, as we don't have an active maintainer for Twisted IOCP reactor.

This is a backward incompatible change... as those who want IOCP support, will now have to install it expliclitly.

I am not sure how to make this change in a backward compatible way.

What we can do, is to add a big warning at the top of the release notes and in the release email

### Manual checks without IOCP

On Windows, get this branch and then run `pip install .` 
Check that IOCP is not installed

You can run the same test on Linux, to make sure IOCP is not installed... just to be extra safe :)

### Manual checks with IOCP

On Windows, get this branch and then run `pip install .[windows-platform]` 
Check that IOCP is  installed

### CI checks

The iocp support is automatically installed in our CI via tox "-window-" environment marker

https://github.com/twisted/twisted/actions/runs/8895048158/job/24424587427?pr=12144#step:10:35


adiroiban:
I think that this is ready for a first round of reviews

needs-review

glyph (approved):
I agree, we should support the most minimal non-native-deps installation on Windows as we do on other platforms.

adiroiban:
Thanks for the review. 

I have enabled the auto-merge for this PR.

I have added the "release-blocker" tag to this PR, so that when we do the release, it can server as a reminder to add a top release note about explicit twisted-iocpsupport install

PR #11971, opened 2023-09-08
#8478 Document testing deprecated modules.

adiroiban:
## Scope and purpose

Fixes #8478

This document how to test deprecated modules.

it also contains a few changes from a very old branch  7 years old - https://github.com/twisted/twisted/tree/8478-deprecated-class-test

docs at https://twisted--11971.org.readthedocs.build/en/11971/development/compatibility-policy.html#testing-deprecation-code

PR #12149, opened 2024-05-01
#12133 Speed up twisted.web's HTTP/1.1 server by about 10%

### LINKED ISSUE #12133: Speed up twisted.web, part 1 of N
Going to try do the equivalent of what I did for twisted.web client: fix a bunch of small performance issues that add up. Initial investigation suggests there will be things to fix, e.g. various overhead from deprecated code paths.

Might be a series of issues, will see.

itamarst:
## Scope and purpose

Fixes #12133

Doing deprecations on module-level is expensive, turns out, because it adds lots of attribute-lookup overhead for _everything_, including non-deprecated code. And other speedups as well.

Before:

```
$ hyperfine --warmup 1 "python http_server_benchmark.py"
Benchmark 1: python http_server_benchmark.py
  Time (mean ± σ):      1.220 s ±  0.010 s    [User: 1.190 s, System: 0.030 s]
  Range (min … max):    1.207 s …  1.238 s    10 runs
```

After:

```
$ hyperfine --warmup 1 "python http_server_benchmark.py"
Benchmark 1: python http_server_benchmark.py
  Time (mean ± σ):      1.105 s ±  0.012 s    [User: 1.081 s, System: 0.024 s]
  Range (min … max):    1.096 s …  1.137 s    10 runs
```

## Contributor Checklist:

### DIFF
--- a/src/twisted/python/deprecate.py
+++ b/src/twisted/python/deprecate.py
@@ -47,7 +47,8 @@ To deprecate properties you can use::
             '''
 
 
-To mark module-level attributes as being deprecated you can use::
+While it's best to avoid this as it adds performance overhead to *any* usage of
+the module, to mark module-level attributes as being deprecated you can use::
 
     badAttribute = "someValue"
 
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -2498,8 +2498,6 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
                 self._networkProducer.pauseProducing()
             return
 
-        self.resetTimeout()
-
         try:
             self._transferDecoder.dataReceived(data)
         except _MalformedChunkedDataError:
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -412,7 +412,7 @@ def stringToDatetime(dateString):
         # Two digit year, yucko.
         day, month, year = parts[1].split("-")
         time = parts[2]
-        year = year
+        year = int(year)
         if year < 69:
             year = year + 2000
         elif year < 100:
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1363,19 +1363,15 @@ class Request:
             cookie += b"; SameSite=" + sameSite
         self.cookies.append(cookie)
 
-    def setResponseCode(self, code, message=None):
+    def setResponseCode(self, code: int, message: Optional[bytes] = None) -> None:
         """
         Set the HTTP response code.
 
         @type code: L{int}
         @type message: L{bytes}
         """
-        if not isinstance(code, int):
-            raise TypeError("HTTP response code must be int or long")
         self.code = code
-        if message:
-            if not isinstance(message, bytes):
-                raise TypeError("HTTP response status message must be bytes")
+        if message is not None:
             self.code_message = message
         else:
             self.code_message = RESPONSES.get(code, b"Unknown Status")
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -2425,12 +2425,7 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
         if not self._maybeChooseTransferDecoder(header, data):
             return False
 
-        reqHeaders = self.requests[-1].requestHeaders
-        values = reqHeaders.getRawHeaders(header)
-        if values is not None:
-            values.append(data)
-        else:
-            reqHeaders.setRawHeaders(header, [data])
+        self.requests[-1].requestHeaders.addRawHeader(header, data)
 
         self._receivedHeaderCount += 1
         if self._receivedHeaderCount > self.maxHeaders:
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -379,7 +379,7 @@ def timegm(year, month, day, hour, minute, second):
     return seconds
 
 
-def stringToDatetime(dateString: bytes) -> int:
+def stringToDatetime(dateString):
     """
     Convert an HTTP date string (one of three formats) to seconds since epoch.
 
@@ -412,7 +412,7 @@ def stringToDatetime(dateString: bytes) -> int:
         # Two digit year, yucko.
         day, month, year = parts[1].split("-")
         time = parts[2]
-        year = int(year)
+        year = year
         if year < 69:
             year = year + 2000
         elif year < 100:
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -2271,13 +2271,15 @@ class HTTPChannel(basic.LineReceiver, policies.TimeoutMixin):
         )
         self._networkProducer.registerProducer(self, True)
 
+    def dataReceived(self, data):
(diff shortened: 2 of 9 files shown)


src/twisted/web/http.py
```diff
@@ -3145,11 +3136,9 @@ def dataReceived(self, data):
         using.
         """
         if self._negotiatedProtocol is None:
-            try:
-                negotiatedProtocol = self._channel.transport.negotiatedProtocol
-            except AttributeError:
-                # Plaintext HTTP, always HTTP/1.1
-                negotiatedProtocol = b"http/1.1"
+            negotiatedProtocol = getattr(
```
itamarst, line 3139: It's possible there's some edge case where this is an incompatible change?

src/twisted/web/resource.py
```diff
@@ -355,7 +349,26 @@ def getChild(self, chnam, request):
         return self
 
 
-class _UnsafeNoResource(_UnsafeErrorPage):
+class _UnsafeErrorPage(_UnsafeErrorPageBase):
```
adiroiban, line 352: I think that we can just remove this code.

We have this code available until 23.4

Based on [The First One’s Always Free](https://docs.twisted.org/en/stable/development/compatibility-policy.html#the-first-one-s-always-free), we already had than one release with this warning, so we can remove it.

src/twisted/python/deprecate.py
```diff
@@ -47,7 +47,8 @@ def badProperty(self, value):
             '''
 
 
-To mark module-level attributes as being deprecated you can use::
+While it's best to avoid this as it adds performance overhead to *any* usage of
```
adiroiban, line 50: I think that we can have this here.

Maybe also add a note that API deprecated in this way should be removed at the N+1 release cycle.

There is a separate PR to look at the overall update of the deprecation docs https://github.com/twisted/twisted/pull/11971

src/twisted/web/http.py
```diff
@@ -385,7 +385,7 @@ def stringToDatetime(dateString):
 
     @type dateString: C{bytes}
     """
-    parts = nativeString(dateString).split()
+    parts = dateString.decode("ascii").split()
```
adiroiban, line 388: I think there are multiple places in which we use nativeString ... and other helpers for 2to3 migration.

I think that is ok to have this change here... but maybe have a separate task to completely remove the usage of nativeString and deprecate nativeString

src/twisted/web/http.py
```diff
@@ -1363,19 +1363,15 @@ def _sanitize(val):
...
 
-    def setResponseCode(self, code, message=None):
+    def setResponseCode(self, code: int, message: Optional[bytes] = None) -> None:
         """
         Set the HTTP response code.
 
         @type code: L{int}
         @type message: L{bytes}
         """
-        if not isinstance(code, int):
```
adiroiban, line 1373: Is this a backward compatible change in terms of how this public API behaves and handles error cases ?

--------

I know that mypy will help with such errors and I agree that we should not have this error handling at runtime

------


I don't know we can raise a warning to inform on this incompatible change.


I think that what we can do is to remove this code and
add a newsfragment/12133.removal release note to let others know about this change

src/twisted/web/http.py
```diff
@@ -2275,13 +2271,15 @@ def connectionMade(self):
         )
         self._networkProducer.registerProducer(self, True)
 
+    def dataReceived(self, data):
```
adiroiban, line 2274: thanks for the change.

I was expecting that this is already handled via the PolicyMixin ... byt it looks like it's not.

https://github.com/twisted/twisted/blob/10fc2f1cf0500f59497557d65b6751d473c40d85/src/twisted/protocols/policies.py#L620

src/twisted/web/http.py
```diff
@@ -2427,12 +2425,7 @@ def headerReceived(self, line):
         if not self._maybeChooseTransferDecoder(header, data):
             return False
 
-        reqHeaders = self.requests[-1].requestHeaders
-        values = reqHeaders.getRawHeaders(header)
-        if values is not None:
-            values.append(data)
-        else:
-            reqHeaders.setRawHeaders(header, [data])
+        self.requests[-1].requestHeaders.addRawHeader(header, data)
```
adiroiban, line 2428: Thanks for the cleanup.

I guess that addRawHeader API was added later, and this is why we had the previous usage with get and set

src/twisted/web/http.py
adiroiban: The change looks ok to me.

Was this just a drive-by clenup, or this has a performance impact?

src/twisted/web/server.py
```diff
@@ -53,20 +53,21 @@
 
 
 # backwards compatibility
-deprecatedModuleAttribute(
+@deprecated(
```
adiroiban: Can we remove this ?

adiroiban (approved):
Thanks for the changes.

They look good.

Things to consider before the merge:

* Add a release note that the TypeError is no longer raised by the public API

-------

We can have the deprecated code merged,
but I would say that we should just remove it.
I can do that in a separate PR.

src/twisted/web/http.py
itamarst: About a 2% performance improvement.

src/twisted/web/resource.py
itamarst: This is... pretty long-lived API I think? So I am reluctant to break existing applications that might rely on it.

itamarst: Specifically the bit in the docs about "Removal should not be undertaken just to follow a timeline. Twisted should strive, as much as practical, not to break applications relying on it."

src/twisted/web/http.py
itamarst: Yeah I did it here because it showed up in the profiling.

src/twisted/web/server.py
itamarst: I guess 12 years is a generous amount of time. I'll do that.

adiroiban:
@glyph @itamarst about performance testing.

I have approved the request for codspeed ... 

--------


I think that one easy way is to write performance tests, is to write them as trial test.

We can have a pluging to trial that will write a report about the test run.
The report can include the time it takes for the test to run with setup + test + teardown, and just the "inner" test.. all in xunit format.

This data can then be sent to Datadog and we can generate graphs and dashboards.
With datadog import tool, we can add extra attributes for each test.
Datadog offered a free account...as long as we put a banner on our website.

I have implemented this for my private project... but stopped using Datadog as it was too expensive.
I am now working to move the database handling to prometheus or influxdb .

--------

The problem that I can see, is having a reliable / reference run environemnt.
From my past experience, the GitHub Action VMs were not alwasy performing the same.

I can volunteer to host a bare-metal server, dedicated to performance testing, and we can guarantee to have the same results.

src/twisted/web/http.py
adiroiban: wow! I was not aware of this. Thanks again for the update

itamarst:
I think the point of codspeed.io is that they provide a reliable runtime environment. We'll see how accurate that claim is.

PR #12153, opened 2024-05-02
#12152 Fix ReverseProxyResource example

### LINKED ISSUE #12152: TwistedWeb ReverseProxyResource example has string not bytes as third argument
The sample code in [ReverseProxyResource](https://docs.twisted.org/en/latest/web/howto/using-twistedweb.html#reverseproxyresource)  is supplying a `str` object as the third argument. This fails with `builtins.TypeError: can only concatenate str (not "bytes") to str` when a request is sent to the endpoint. Assuming it was last tested with Python 2 where concatenating a `str` and a `bytes` object would work.

thormick:
## Scope and purpose

Fixes #12152

Changes third argument of example code from string to bytes object. With a string as third argument this fails with `builtins.TypeError: can only concatenate str (not "bytes") to str` when a request is sent to the endpoint. This would have worked with Python 2 where concatenating a `str` and a `bytes` object works, seems like the example is from then.


thormick:
please review

adiroiban (approved):
Thanks for the fix. It looks ok.

This PR is perfect.

I am stupid: 
~~To follow our development process, I have created a separate GitHub issue dedicated to the documentation bug.~~ 

----------

In an ideal case, we would have the example code as separate file, and then execute the example files as part of our continous tests.

In this way, we could detect any issues in the example code that might be generated by changes in the main Twisted code.

PR #12157, opened 2024-05-02
#12155 Speed up twisted.web server, part 2

### LINKED ISSUE #12155: Speed up twisted.web server, part 2 of N
Probably focusing on header writing, there are some obvious bottlenecks are.

itamarst:
## Scope and purpose

Fixes #12155

Headers passed to the channel were already (almost all) sanitized, so I made sure all headers were sanitized and added a fast path.

Before:

```
$ hyperfine --warmup 1 "python http_server_benchmark.py"
Benchmark 1: python http_server_benchmark.py
  Time (mean ± σ):      1.091 s ±  0.009 s    [User: 1.053 s, System: 0.038 s]
  Range (min … max):    1.081 s …  1.108 s    10 runs
```

After:

```
$ hyperfine --warmup 1 "python http_server_benchmark.py"
Benchmark 1: python http_server_benchmark.py
  Time (mean ± σ):      1.035 s ±  0.008 s    [User: 1.001 s, System: 0.033 s]
  Range (min … max):    1.024 s …  1.049 s    10 runs
```
 
## Contributor Checklist:

### DIFF
--- a/src/twisted/web/_http2.py
+++ b/src/twisted/web/_http2.py
@@ -1059,7 +1059,7 @@ class H2Stream:
             self._producerProducing = False
 
     # Methods called by the consumer (usually an IRequest).
-    def writeHeadersObject(self, version, code, reason, headers):
+    def writeHeaders(self, version, code, reason, headers):
         """
         Called by the consumer to write headers to the stream.
 
--- a/src/twisted/web/_http2.py
+++ b/src/twisted/web/_http2.py
@@ -1078,6 +1078,8 @@ class H2Stream:
         """
         self._conn.writeHeaders(version, code, reason, headers, self.streamID)
 
+    writeHeadersPresanitized = writeHeaders
+
     def requestDone(self, request):
         """
         Called by a consumer to clean up whatever permanent state is in use.
--- a/src/twisted/web/_http2.py
+++ b/src/twisted/web/_http2.py
@@ -1059,7 +1059,7 @@ class H2Stream:
             self._producerProducing = False
 
     # Methods called by the consumer (usually an IRequest).
-    def writeHeaders(self, version, code, reason, headers):
+    def writeHeadersObject(self, version, code, reason, headers):
         """
         Called by the consumer to write headers to the stream.
 
@@ -1073,12 +1073,15 @@ class H2Stream:
         @type reason: L{bytes}
 
         @param headers: The HTTP response headers.
-        @type headers: Any iterable of two-tuples of L{bytes}, representing header
-            names and header values.
+        @type headers: L{twisted.web.http_headers.Headers}
         """
-        self._conn.writeHeaders(version, code, reason, headers, self.streamID)
-
-    writeHeadersPresanitized = writeHeaders
+        self._conn.writeHeaders(
+            version,
+            code,
+            reason,
+            [(k, v) for (k, values) in headers.getAllRawHeaders() for v in values],
+            self.streamID,
+        )
 
     def requestDone(self, request):
         """
--- a/src/twisted/web/http.py
+++ b/src/twisted/web/http.py
@@ -1200,7 +1200,6 @@ class Request:
             version = self.clientproto
             code = b"%d" % (self.code,)
             reason = self.code_message
-            headers = []
 
             # if we don't have a content length, we send data in
             # chunked mode, so that we can support pipelining in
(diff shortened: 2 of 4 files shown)


src/twisted/web/http.py
```diff
@@ -1211,7 +1210,7 @@ def write(self, data):
                 and self.method != b"HEAD"
                 and self.code not in NO_BODY_CODES
             ):
-                headers.append((b"Transfer-Encoding", b"chunked"))
```
itamarst, line 1214: This change in design means all headers end up in `responseHeaders`, which is perhaps a semantics change?

But not sure why Etag being added but cookies not being added made sense.

src/twisted/web/_http2.py
```diff
@@ -1059,7 +1059,7 @@ def flowControlBlocked(self):
             self._producerProducing = False
 
     # Methods called by the consumer (usually an IRequest).
-    def writeHeaders(self, version, code, reason, headers):
+    def writeHeadersObject(self, version, code, reason, headers):
```
itamarst: The `H2Stream` class is private so non-backwards compatible changes seems OK.

src/twisted/web/http.py
```diff
@@ -2666,6 +2662,35 @@ def writeHeaders(self, version, code, reason, headers):
         headerSequence.append(b"\r\n")
         self.transport.writeSequence(headerSequence)
 
+    def writeHeadersObject(self, version, code, reason, headers):
```
itamarst: `HTTPChannel` is a public API sadly, so we need to add a new method and leave the old one for backwards compatibility.

itamarst:
please review

src/twisted/web/_http2.py
adiroiban: All good. Thanks.

My investigation on this api.

`H2Stream` is accesible via `H2Connection.streams{}`

`H2Connection` is accesible via
`twisted.web.http._GenericHTTPChannelProtocol._channel`

While `_GenericHTTPChannelProtocol` is made public via `twisted.web.http.HTTPFactory` the `_channel` member is still private and I could not find anywhere in the code where it might be exposed as public API.

So all good.

src/twisted/web/http.py
```diff
@@ -1211,7 +1210,7 @@ def write(self, data):
                 and self.method != b"HEAD"
                 and self.code not in NO_BODY_CODES
             ):
-                headers.append((b"Transfer-Encoding", b"chunked"))
+                self.responseHeaders.setRawHeaders("Transfer-Encoding", [b"chunked"])
```
adiroiban, line 1213: Maybe add a comment, that we overwrite any previous encoding that might have been explicitly set by the resource.

src/twisted/web/http.py
adiroiban: Thanks for the update.

I don't remember reading it, but I guess that for Twisted API, the `transfer-encoding` and `set-cookie` headers are kind of a private headers, as we have separate methods to set them.

I was thinking that at some point Twisted API would reject setting the `tranfer-encoding` or `set-cookie` headers , and this is why we have the separate headers here... but I could not find any ticket about this.

------

I guess that it would be nice to raise a warning is a resource / user tries to set the transfer-encoding header.... but this is something for another issue/PR

src/twisted/web/http.py
adiroiban: Can we add type annotation to this method definition ?


------

Can we have the existing `writeHeaders` method signature to accept both Iterable and `twisted.web.http_headers.Headers` ?


When a `Headers` instance is pass, we skip the sanitization process.

----------

Is this a high CPU usage ?

Maybe make `twisted.web.http_headers.Headers` an iterator ?

adiroiban (approved):
Thanks for the updated. They look good.

I left a few comments inline... not blockers.

src/twisted/web/http.py
itamarst: I can see what the performance overhead of those two options is like.

itamarst: Oh the iterator version won't work because the issue is the sanitization overhead of readding everything to a `Headers` instance, which we don't want. I can measure isinstance.

itamarst: The slow down is ... hard to measure? So I went with isinstance() version.

src/twisted/web/http.py
itamarst: Previously we would just have... both? It's not something users should set, either way.

adiroiban:
Thanks for the update. Happy that this could be done without a new API. great work!

PR #12161, opened 2024-05-03
#12154 Setup minimal benchmarks using codspeed.io

### LINKED ISSUE #12154: Setup benchmarks using Codspeed.io
I have a couple of HTTP benchmarks I've been using, would be good to make sure there's no regressions. codspeed.io would allow running benchmarks on every PR.

itamarst:
## Scope and purpose

Fixes #12154

## Contributor Checklist:


tox.ini
```diff
@@ -219,3 +219,9 @@ commands =
     cp -r {distdir} {toxinidir}/dist # copy the wheel built by tox-wheel
     {envpython} -m pep517.build --source --out-dir={toxinidir}/dist {toxinidir}
     twine check {toxinidir}/dist/*.*
+
+
+[testenv:benchmarks-ci]
+deps =
+    pytest-codspeed
+commands = pytest --codspeed benchmarks/
```
adiroiban: For future cleaner diff... empty line at the end of the file

```suggestion
commands = pytest --codspeed benchmarks/

```

benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
+# Benchmarks, to be run by codspeed.io in CI
```
adiroiban, line 1: Not sure if this should be Markdown or RST.

I think that is important to mention that the tests are executed using `pytest` and not using trial.

This is why the test files are outside of `src/twisted`

benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Running benchmarks
+
+Note that as of mid-2024, codspeed.io uses a simulated CPU (Cachegrind) to run tests, so the measures of performance are not suitable for optimizing low-level compiled code.
+
+In CI, we use `tox -e benchmarks-ci` to run the benchmarks.
```
adiroiban: Maybe add a note that normal GitHub Action VMs are used for CI ... there are no dedicated servers.

```suggestion
In CI, we use `tox -e benchmarks-ci` to run the benchmarks, using GitHub Actions.
The tests are executed on GitHub VMs and the reports are sent to the codspeed.io cloud.
```

benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+You can run benchmarks locally by installing `pytest-benchmark` and then running `pytest benchmarks/`.
+
+## Writing benchmarks
+
+See the [Codspeed documentation](https://docs.codspeed.io/benchmarks/python).
+
+Note that the `@pytest.mark.benchmark` style of benchmark doesn't work with `pytest-benchmark`, so you might want to avoid it.
```
adiroiban:
```suggestion
Note that the `@pytest.mark.benchmark` style of benchmark doesn't work with `pytest-benchmark`, so you should not use it.
```

benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
+# Benchmarks, to be run by codspeed.io in CI
+
+## Running benchmarks
+
+Note that as of mid-2024, codspeed.io uses a simulated CPU (Cachegrind) to run tests, so the measures of performance are not suitable for optimizing low-level compiled code.
+
+In CI, we use `tox -e benchmarks-ci` to run the benchmarks.
+
+You can run benchmarks locally by installing `pytest-benchmark` and then running `pytest benchmarks/`.
```
adiroiban, line 15: I know that this info is on codspeed ,
but maybe it's useful to have it here

```suggestion
You can run benchmarks locally by installing `pytest-benchmark` and then running `pytest benchmarks/`.

No reports are generated.
This is just to check the tests aren't failing.
```

benchmarks/test_web_client.py
```diff
@@ -0,0 +1,36 @@
...
+            request = Request(
+                b"GET", b"/foo/bar", Headers({b"Host": [b"example.com"]}), None
+            )
+            response = protocol.request(request)
+            protocol.dataReceived(RESPONSE)
+            result = []
+            response.addCallback(result.append)
+            assert result
+
+    benchmark(go)
```
adiroiban, line 40: I was expecting that `benchmark()` already handles running the test multiple time... including having the test rounds as a parameter.

I expect there will be a lot of boilerplate code for  the tests like

```
def test_something(benchmark): 
    def go():
        for _ in range(1000):
          ACTUAL TEST
   benchmark(go)
```

Maybe we can have a decorator like this.

```
@generate_benchmark_test(__name__)
def http_client_small_response():
   """
   Docstring for this test.
   """
   protocol = HTTP11ClientProtocol()
   assert protocol
   MORE TEST CODE
```

It will "inject" into the module scope, new functions via
```setattr(sys.modules[module_name], name, new_test)```

name will start with `test_` to be picked up the pytest test discovery.

Not sure if this will work with pytest

I am using something like this with stdlib unittest


Other than fixture and parameters, I couldn't find any other way to dynamicaly generate tests in pytest

File: benchmarks/test_web_server.py [hunk 7]
```diff
@@ -0,0 +1,53 @@
+"""Benchmarks for the C{twisted.web} server."""
```
adiroiban: I think that this is the minimal standard header for files in Twisted.

```suggestion
# Copyright (c) Twisted Matrix Laboratories.
# See LICENSE for details.

"""
Benchmarks for the C{twisted.web} server.
"""
```

benchmarks/test_web_client.py
```diff
@@ -0,0 +1,36 @@
+"""Benchmarks for C{HTTP11ClientProtocol}."""
```
adiroiban: Just standard header

```suggestion
# Copyright (c) Twisted Matrix Laboratories.
# See LICENSE for details.

"""
Benchmarks for C{HTTP11ClientProtocol}.
"""
```

adiroiban (approved):
Thanks. I think that we should give it a go.

I would like to have this merged in trunk, to have a baseline.

Later we can see PR diff reports and we can improve on best practices on writing the tests.

-------

too bad there is no generic API for codspeed so that we could write these tests using trial.

.github/workflows/test.yaml
```diff
@@ -290,7 +290,30 @@ jobs:
         name: ${{ matrix.python-version || env.DEFAULT_PYTHON_VERSION }}-${{matrix.job-name || 'default-tests' }}
         fail_ci_if_error: true
 
+  benchmarks:
```
adiroiban, line 293: Note that with the current configuration, if the benchmark tests are failing, the merge is not blocked.

I think that we should add the benchmark tests to  `all-successful` list

```
  all-successful:
    # Is very important to force running this always, as otherwise it will be
    #  skipped by default.
    if: always()
    runs-on: ubuntu-latest
    # Here should be the list of all the other jobs defined in this file.
    needs:
      - testing
      - narrativedocs
      - apidocs
      - static-checks
      - release-publish
```

.github/workflows/test.yaml
```diff
@@ -290,7 +290,30 @@ jobs:
...
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '${{ env.DEFAULT_PYTHON_VERSION }}'
 
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip tox~=3.0
+        tox -e benchmarks-ci
```
adiroiban: Does it make sense to yet tox create another virtual env here?

Maybe just use pip install something like
```pip install .[benchmark]``` to install the benchmark depnedencies

File: .github/workflows/test.yaml [hunk 11]
```diff
@@ -290,7 +290,30 @@ jobs:
...
+
+    - name: Run benchmarks
+      uses: CodSpeedHQ/action@v2
+      with:
+        token: ${{ secrets.CODSPEED_TOKEN }}
+        # codspeed runs this command under a CPU emulator, so it's super-slow.
+        # As such, runnig the full `tox -e benchmarks-ci` here is too slow, so
+        # we break it into two parts.
+        run: .tox/benchmarks-ci/bin/pytest --codspeed benchmarks/
```
adiroiban: And then without tox we would have 

```suggestion
        run: python -m pytest --codspeed benchmarks/
```

benchmarks/README.md
glyph: Would it make sense to generate relative comparisons by just running `time` or something? Seems like an unfortunate detail that there's no information at all provided in local development.

adiroiban: I would be nice to have some local results... but.

My understanding is that it's hard to create any meaningful comparison.

You can have a test-suite execute on trunk and then on this branch, and comparate the time... but I don't know if the error margin is good enough.

--------

so the remaining option is to run the test with valgrid in trunk and on the current branch... that this requires a lot of work... and I would say that such work is outside of the scope of this PR.

this PR is about getting some reports via codspeed.io

-----------

I would say that we should give codspeed.io a try and see if it help.

Maybe, we will find out that this is not very helpful and we need to implement our own tools to run and colect results from `valgrind --tool=cachegrind python -m twisted.trial`

glyph: > I would be nice to have some local results... but.
> 
> My understanding is that it's hard to create any meaningful comparison.
> 
> You can have a test-suite execute on trunk and then on this branch, and comparate the time... but I don't know if the error margin is good enough.

"good enough" for what?  If I'm developing a performance improvement, I don't want to push to CI and wait for 20 minutes to see if my changes made any impact at all.  If I'm running repeatedly in the exact same environment on my development workstation, the results are same enough to get a comparison.  I am not even talking about running on trunk, then a branch; I just mean being able to run a benchmark, make an edit, then run it again to see if the edit did anything.

glyph: > so the remaining option is to run the test with valgrid in trunk and on the current branch... that this requires a lot of work... and I would say that such work is outside of the scope of this PR.

I'm totally fine with the precision and rigor of valgrind being reserved for running in CI.

glyph: > I would say that we should give codspeed.io a try and see if it help.

Sure, I just don't want to structure our benchmarks so they *can't* be run locally, given how simple our requirements are at this point. It's a bit annoying to have to drag in all of pytest but I'm not going to complain if that's the API that a service that someone *else* maintains presents to us

glyph: To be clear I would not block this on this basis, I think "run function in a loop, print how long it took" should *not* be an unreasonable demand but there seem to be a lot of moving pieces with codspeed so if it is actually difficult then go ahead and land without it

itamarst: `pytest-benchmark` does actually output numbers, so if you use it instead of `pytest-codspeed` you're good for local development.

benchmarks/test_web_client.py
itamarst: The rounds at least can probably be omitted, yeah.

PR #12168, opened 2024-05-06
Update the compatibility policy

### LINKED ISSUE #12167: Deprecation docs should use NEXT placeholders
The [compatibility policy](https://docs.twisted.org/en/stable/development/compatibility-policy.html) contains a number of examples of how to deprecate APIs that hard-code specific Twisted versions. This isn't what a Twisted developer should actually write: new deprecations should be written with [incremental placeholders](https://github.com/twisted/incremental?tab=readme-ov-file#updating):

    Version("Twisted", "NEXT", 0, 0)

    Twisted NEXT

The documentation should match practice.

twm:
Fix various formatting issues and introduce NEXT placeholders where applicable.

## Scope and purpose

Fixes #12167.

I noticed that the reST syntax for notes was mangled, resulting in the note content being hidden. I restored these notes.

I also removed language about "new-style" classes because old-style classes are mercifully a thing of the past and reworded/rearranged some sections.

### DIFF
--- a/docs/development/compatibility-policy.rst
+++ b/docs/development/compatibility-policy.rst
@@ -34,7 +34,8 @@ Although a change may be broadly considered backward compatible, as long as it d
 The compatibility policy described here is 99% about changes to **interface**,
 not changes to functionality.
 
-..  note::
+.. note::
+
     Ultimately we want to make the user happy but we cannot put every possible thing that will make every possible user happy into this policy.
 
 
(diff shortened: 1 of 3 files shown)


docs/development/compatibility-policy.rst
```diff
@@ -298,99 +303,105 @@ When an application wants to be upgraded to a new version of Twisted, it can do
...
+    deprecatedModuleAttribute(
+        Version("Twisted", "NEXT", 0, 0),
+        "Use twisted.internet.ssl.DefaultOpenSSLContextFactory instead.",
+        __name__,
+        SSLContextFactory.__name__,
+    )
 
-Functions and methods
-^^^^^^^^^^^^^^^^^^^^^
+Pass ``Version("Twisted", "NEXT", 0, 0)`` `incremental placeholder <https://github.com/twisted/incremental#updating>`_ to the to indicate the upcoming release.
```
adiroiban, line 341: Shouldn't we recommand a single placeholder ?

I would go with ``Twisted NEXT`` as it should be valid for both Python or RST code.

docs/development/compatibility-policy.rst
```diff
@@ -298,99 +303,105 @@ When an application wants to be upgraded to a new version of Twisted, it can do
...
         """
-        Function deprecated using a decorator and which has no replacement.
+        Function deprecated using a decorator which has no replacement.
         """
         return bar * 3
 
+If you can't use the decorator add a call to :py:func:`warnings.warn()` at the beginning of the implementation.
+The warning should be of type :py:exc:`DeprecationWarning` and the stack level should be set so that the warning refers to the code which is invoking the deprecated function or method.
+The deprecation message must include the name of the function which is deprecated, the version of Twisted in which it was first deprecated, and a suggestion for a replacement.
+If the API provides functionality which it is determined is beyond the scope of Twisted or it has no replacement, then it may be deprecated without a replacement.
```
adiroiban, line 376: Maybe have this as a separate paragraph.

This looks like an important info. Maybe move it at the top of the document?

```suggestion

If the API provides functionality which it is determined is beyond the scope of Twisted or it has no replacement, then it may be deprecated without a replacement.
```

adiroiban (approved):
Looks good. Thanks!

docs/development/compatibility-policy.rst
twm: They're not interchangeable: `Version(...)` is what you pass to the APIs; `Twisted NEXT` is what you use in docstrings and/or assertions.

docs/development/compatibility-policy.rst
twm: A separate paragraph  makes sense to me.

twm:
Thanks Adi!

docs/development/compatibility-policy.rst
adiroiban: Thanks for the info.
I see why you might want to have an increment Version instance in Python... but I always felt that this is an overengineering ... at least for the deprecation warning message.

PR #12170, opened 2024-05-06
#12169 defer: Expand inline callbacks tests

### LINKED ISSUE #12169: Improve defer.inlineCallbacks tests
Currently there are only few tests for defer.inlineCallbacks.

A PR is going to be created soon.

p12tic:
## Scope and purpose

Fixes #12169

This PR significantly expands `@inlineCallbacks` tests. The code has been taken from https://github.com/twisted/twisted/pull/1090.


p12tic:
I have also added an additional commit https://github.com/twisted/twisted/pull/12170/commits/1debeec7a32fd0c1916aac288a0b59ef02998da7 and some further commits that moves some inline callbacks tests from other files to test_inlinecb.py. It can be regarded as adding more tests too as it certainly makes them more discoverable.

p12tic:
please review

p12tic:
Failing tests should be rerun - seems like intermittent issue with the infrastructure

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -197,6 +907,42 @@ def calling():
...
+        def ic(d: object) -> Generator[Any, Any, None]:
+            """
+            This is never called.
+            It is only used as the decorated function.
+            The resulting function is never called in this test.
+            This is used to make sure that if we wrap
+            an already failed deferred, inlineCallbacks
+            will not add any extra traceback frames.
+            """
+            yield d
```
adiroiban: If we know that a line is never called, is best to also let our coverage tools know about this :)

```suggestion
            yield d  # pragma: no cover
```

src/twisted/newsfragments/12169.bugfix
```diff
@@ -0,0 +1 @@
+Added more tests for ``twisted.internet.defer.inlineCallbacks`` decorator.
```
adiroiban: What is the bug fix here ?

What was not working before in the public API?

-------

It looks like this should be a 12168.misc file.

adiroiban:
Which tests were failing and required a re-run ?

I can see that all the tests have passed from the first run - https://github.com/twisted/twisted/actions/runs/8972762885/job/24641482445?pr=12170

----------

We are trying to reduce the numer of flaky/random failing test.
If you find a randomly failing tests, please add info about the test name or a link to the build

There are a few open issues for specific flaky tests.

Thanks!

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,332 @@
...
+    Make a L{Failure} of a divide-by-zero error.
+    """
+    try:
+        raise ZeroDivisionError(msg)
+    except BaseException:
+        f = Failure()
+    return f
+
+
+class TerminalException(Exception):
```
adiroiban, line 45: Can you please add a docstring explaining the purpose of this exception.

The descring should help other developers if it's ok to re-use this exception in other tests, or is best to create another exception.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,332 @@
...
+    pass
+
+
+class BasicTests(TestCase):
```
adiroiban, line 54: Can we have a docstring for these tests.

The docstring should help other developers know whether or not their new tests should be added here, or in another testcase.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,332 @@
...
+    # First provide all the generator impls necessary for BaseDefgenTests
+
+    @inlineCallbacks
+    def _genBasics(self):
```
adiroiban: What is the purpose of this helper?

Can you please add docstring?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,332 @@
...
+        return self._genBasics().addCallback(self.assertEqual, "WOOSH")
+
+    @inlineCallbacks
+    def _genBuggy(self):
+        yield getThing()
+        1 / 0
+
+    def testBuggy(self):
+        """
+        Ensure that a buggy generator properly signals a Failure
```
adiroiban: What is a "buggy generator" ? What "properly" means?

Can this docstring be more specific ?

----------

What is my understanding  of the purpose of the tests, by reading the test code


"""
Any exception raised by a generator is converted into a L{Failure}.
"""

Is my understanding correct ?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,332 @@
...
+
+    @inlineCallbacks
+    def _genNothing(self):
+        if False:
+            yield 1
+
+    def testNothing(self):
+        """Test that a generator which never yields results in None."""
+
+        return self._genNothing().addCallback(self.assertEqual, None)
```
adiroiban: Do we need to spin the reactor for these tests ?

This comment is valid for the other tests.

If spining the reactor is needed, I think that is important to add a comment.


Maybe something like this?

```suggestion
        """
        C{None} is returned by a generator that never yields.
        """
        deferred = self._genNothing()
        result = self.successResultOf(deferred)
        self.assertIsNone(result)

```

src/twisted/internet/test/test_inlinecb.py
adiroiban: I see that this code was jus moved... but is better to add the coverage pragma info now.

Thanks.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -49,6 +359,406 @@ def f(d):
...
+        @inlineCallbacks
+        def f2(x):
+            expectations.append(("f2 enter", x))
+
+            d = Deferred()
+            deferredList.append((d, MyException()))
+            x = yield d
+
+            expectations.append(("f2 exit", x))
+            returnValue(x)
```
adiroiban: It looks like something is no ok here.

These lines are never called.

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -49,6 +359,406 @@ def f(d):
...
+        self.assertEqual(
+            expectations,
+            [
+                ("f3 enter", 1),
+                ("f2 enter", 1),
+                ("f3 exit", 5),
+            ],
+        )
+
+    def test_raisesExceptionFromDeferredWithWaitingSecondCallback(self):
```
adiroiban, line 763: Can we have a docstring describing the purpose / scope of this test ?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -49,6 +359,406 @@ def f(d):
...
+                ("f1 exit", 2),
+                ("f1 enter", 2),
+                ("f1 exit", 3),
+                ("f2 enter", 3),
+                ("f2 exit", 5),
+                ("f3 exit", 9),
+            ],
+        )
+
+    def test_raisesExceptionFromDeferredWithWaitingFirstCallback(self):
```
adiroiban, line 713: Can we have a docstring describing the purpose and scope of this test?

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -49,6 +359,406 @@ def f(d):
...
 
 
+class StackedInlineCallbacksTests(TestCase):
+    """
+    We have an optimization that invokes generators directly when an
+    inlineCallbacks-decorated function yields value directly to yield of
+    another inlineCallbacks-decorated function.
+    """
+
+    def runCallbacksOnDeferreds(self, deferredList):
```
adiroiban, line 388: Can we have a docstring for this test helper?

adiroiban (commented):
Many thanks for the changes.

Happy to see more tests and cleanups.

Can you please add more docstring to the tests?

I am not sure what is the purpose of some of the tests and why for example some lines are not covered.

If a line is expected not to be covered, it should be marked with `# pragma: no cover` 

------------

Also, I am very happy to have the test move outside of the #1090 PR

Thanks!

adiroiban (changes requested):
Previous comment was a "request changes" comment.

p12tic:
> I am not sure what is the purpose of some of the tests

I will add docstrings where I can. Some of the tests were just moved, so I don't understand them either, but I will try.

>  why for example some lines are not covered

It seems weird that coverage considers tests as something that needs to be covered. Perhaps the configuration is broken. As I understand, tests are what creates coverage for other tests, they themselves shouldn't be covered. What I'm missing here?

p12tic:
>  If you find a randomly failing tests, please add info about the test name or a link to the build

It was not the tests themselves that were failing. The infrastructure was bad, e.g. failed to upload things and so on.

adiroiban:
> I will add docstrings where I can. Some of the tests were just moved, so I don't understand them either, but I will try.

Ok. We can leave moved tests without a docsting.

But, if we don't understand the test, maybe is best to "leave it alone".
That is, don't try to move it.
In this way, it should be easier to follow its blame history.

With the move, I thought that you know what was going on there :) 

-----

> The infrastructure was bad, e.g. failed to upload things and so on.

No problem.

If it happens to see such an error again,  Is best to leave a link so that we can take a look and see what can be done to improve the infrastructrure

---------

> It seems weird that coverage considers tests as something that needs to be covered. 

I consider that coverage for the test code should be always 100% mandatory... no exception.

It helps detect tests or part of the tests that are never called.

For example we have tests skipped under various conditions, but we should make sure that a test is always executed at least once...that is, it is newer always skipped.

------

I was not asking to have tests for the testing function / helpers.

I was asking to make sure that any testing function or helper is alwasys executed at least one.

Does it make sense ?

Regards

p12tic:
Thanks for explanation, it all makes sense.

src/twisted/newsfragments/12169.bugfix
p12tic: Fixed.

src/twisted/internet/test/test_inlinecb.py
p12tic: Done.

src/twisted/internet/test/test_inlinecb.py
p12tic: Added.

src/twisted/internet/test/test_inlinecb.py
p12tic: Done.

src/twisted/internet/test/test_inlinecb.py
p12tic: Correct. I've expanded the description, renamed the test and renamed _genBuggy in both inlinecb and defgen test suites.

src/twisted/internet/test/test_inlinecb.py
p12tic: I would be duplicating the docstring of the test that this helper is for. To avoid this duplication I just moved the test helpers to the tests they belong to. In this case all such helpers were used by a single test each. Accordingly, the test docstring now appropriately explains the test helpers too.

src/twisted/internet/test/test_inlinecb.py
p12tic: Added, thanks!

src/twisted/internet/test/test_inlinecb.py
p12tic: Done. It's just a separate exception type that is used within this file. There's nothing unique in it.

src/twisted/internet/test/test_inlinecb.py
p12tic: Applied, thanks.

src/twisted/internet/test/test_inlinecb.py
p12tic: It is important to directly exercise the API because inlineCallbacks and Deferred may be optimized to interact in specific ways depending on how they are invoked together. Therefore it is important to specifically use addCallback here.

It would be a more complete test if there was a sibling of this test that tested the case of something in _genNothing blocking.

p12tic: I added a note in a comment explaining the importance of usage of specific APIs.

src/twisted/internet/test/test_inlinecb.py
p12tic: It is expected that these lines are not exercised. I added an explanation and no-cover pragma.

p12tic:
I've applied all suggestions.

p12tic:
please review

src/twisted/test/test_defgen.py
```diff
@@ -96,100 +146,32 @@ def testStackUsage(self):
...
-    """
-    Calls L{deferredGenerator} while suppressing the deprecation warning.
-
-    @param f: Function to call
-    @return: Return value of function.
-    """
-    return runWithWarningsSuppressed(
-        [
-            SUPPRESS(
-                message="twisted.internet.defer.deferredGenerator was " "deprecated"
```
adiroiban, line 119: I see that this was deprecated in 2015 ... I think that is ok to just remove this code rather than keep maintaining it.

src/twisted/test/test_defgen.py
```diff
@@ -36,58 +28,116 @@ def CRAP():
...
+        return self.assertFailure(_genProduceException(), ZeroDivisionError)
 
     def testNothing(self):
         """Test that a generator which never yields results in None."""
 
-        return self._genNothing().addCallback(self.assertEqual, None)
+        @deprecatedDeferredGenerator
+        def _genNothing():
+            if False:
+                yield 1
```
adiroiban: It's best to explicitly inform the `coverage` tool that we don't expect this line to execute.

```suggestion
                yield 1  # pragma: no cover
```

src/twisted/internet/test/test_inlinecb.py
```diff
@@ -3,22 +3,351 @@
...
+
+        return self.assertFailure(_genProduceException(), ZeroDivisionError)
+
+    def testNothing(self):
+        """Test that a generator which never yields results in None."""
+
+        @inlineCallbacks
+        def _genNothing():
+            if False:
+                yield 1
```
adiroiban:
```suggestion
                yield 1  # pragma: no cover
```

adiroiban (approved):
Thanks for moving the inline tests, outside of the deferedGenerator.

I think that we can delete the deprecated code for waitForDeferred and deferedGenerator.

The only comment is about informing coverage about lines that are exected not to be executed.

adiroiban:
Thanks for the cleanup. I have merged this as I hope it will help with future cleanup work for deprecated code and with deferred performance improvements.

src/twisted/test/test_defgen.py
p12tic: Agreed.

PR #12165, opened 2024-05-03
#8840 re-enable PTY tests on macOS

glyph:
## Scope and purpose

Fixes #8840

Issue #8840, opened 2016-09-25
PTY tests on macOS are flaky


twisted-trac:
|[<img alt="hawkowl's avatar" src="https://avatars.githubusercontent.com/u/3307100?s=50" width="50" height="50">](https://github.com/hawkowl)| @hawkowl reported|
|-|-|
|Trac ID|trac#8840|
|Type|defect|
|Created|2016-09-25 11:56:27Z|

```
<glyph> if you do write() then close() real fast on one end of a pty, and the other end does read(), if close() happens first, sometimes the read() gives back the empty string
```

Affected tests:

twisted.internet.test.test_process.PTYProcessTestsBuilder
twisted.test.test_process.PosixProcessPTYTests.test_executionError

<details><summary>Searchable metadata</summary>

```
trac-id__8840 8840
type__defect defect
reporter__hawkowl hawkowl
priority__normal normal
milestone__None None
branch__ 
branch_author__ 
status__new new
resolution__None None
component__core core
keywords__None None
time__1474804587955678 1474804587955678
changetime__1475218799477390 1475218799477390
version__None None
owner__None None

```
</details>

glyph:
We've had these tests disabled in CI for a while now, but I just did `trial -u twisted.internet.test.test_process` and got up to Test Pass ~~50~~ ~~62~~ ~~154~~ ~~237~~ ~~331~~ 410 with no issues, and it's continuing to run.  I've opened a PR to re-enable these tests on macOS.

glyph:
(I think the relevant Darwin bug was fixed many years ago at this point)

glyph:
(At 1 hour and 585 test passes, I'm calling it, and I'm going to stop running the tests.)

adiroiban:
Reverted in https://github.com/twisted/twisted/pull/12173

adiroiban:
example test failure ... not adding link to the build as the build log will expire in 30 days

```
[FAIL]
Traceback (most recent call last):
  File "/Users/runner/work/twisted/twisted/.tox/macos-withcov-alldeps/lib/python3.12/site-packages/twisted/internet/test/test_process.py", line 523, in test_errorDuringExec
    self.assertIn("\N{SNOWMAN}".encode(), output.getvalue())
  File "/Users/runner/work/twisted/twisted/.tox/macos-withcov-alldeps/lib/python3.12/site-packages/twisted/trial/_synctest.py", line 509, in assertIn
[...]

PR #12173, opened 2024-05-07
Revert "#8840 re-enable PTY tests on macOS"

glyph:
Reverts twisted/twisted#12165

Reopens #8840


glyph:
Test failures on #986 suggests that this is still flaky and the kernel issue is not fixed.

glyph:
The newsfragment check is just broken and its failure should be ignored here. https://github.com/twisted/towncrier/issues/598

adiroiban (approved):
Thanks. Reopens will not work in GitHub... but I will reopen it manually.

This also needs to be "forced merged" as the newsfragment check is mandatory.

glyph:
Thanks!

PR #12175, opened 2024-05-08
#12174 Run benchmarks with with PyPy as well

### LINKED ISSUE #12174: Expand benchmarks to run on PyPy as well
Currently the codspeed.io run on CPython only, but PyPy is also important.

Running benchmarks on PyPy requires a slightly different runtime approach, insofar as it has a JIT (and future CPython may add JIT too, of course). In particular, my impression is that codspeed will only run the code once, and measure instruction count. This is fine for CPython, but for a JIT it will give misleading results. So the benchmarking framework should run code multiple times if it's on PyPy.

This can be done with a custom pytest fixture.

Separately, there's the issue of how to communicate different runs to codspeed, which I will also look into.

itamarst:
## Scope and purpose

Fixes #12174 

Also run benchmarks on PyPy.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -291,7 +291,7 @@ jobs:
         fail_ci_if_error: true
 
   benchmarks:
-    name: Benchmarks for Python $${{ matrix.python-version }}
+    name: Benchmarks for Python ${{ matrix.python-version }}
     strategy:
       matrix:
         python-version: ["3.12", "pypy3.10"]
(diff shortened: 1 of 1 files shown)


adiroiban:
Is there an official solution for this ?

I guess that we might want to run these tests with python 3.12 and later with 3.13 to see if there are any improvements or regressions

----------

What do you say about a helper that injects test functions in module namespace?

It can generate one for pypy and one for cpython ... with different names

-------

I don't have experience with pytest... but I am using this trick to generate Selenium tests that are executed with Firefox and Chrome.. so that I get 2 different names

itamarst:
Currently blocked on upstream support: https://github.com/CodSpeedHQ/pytest-codspeed/issues/34

glyph:
@itamarst thanks for tracking that

